PR #22, opened 2017-08-22
Fixed bootstrap.php for case with install by Composer as dependency and run from vendor/bin

corpsee:
Fixed `bootstrap.php` for case with install by Composer as dependency and run from `vendor/bin` (`./vendor/bin/infection`).

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -4,6 +4,32 @@
  *
  * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
  */
-require __DIR__ . '/../vendor/autoload.php';
+
+$files = [
+    __DIR__ . '/../../../autoload.php',
+    __DIR__ . '/../vendor/autoload.php',
+    __DIR__ . '/vendor/autoload.php'
+];
+
+foreach ($files as $file) {
+    if (file_exists($file)) {
+        define('INFECTION_COMPOSER_INSTALL', $file);
+
+        break;
+    }
+}
+unset($file);
+
+if (!defined('INFECTION_COMPOSER_INSTALL')) {
+    fwrite(STDERR,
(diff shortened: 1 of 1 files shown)


app/bootstrap.php
```diff
@@ -4,6 +4,32 @@
...
+];
+
+foreach ($files as $file) {
+    if (file_exists($file)) {
+        define('INFECTION_COMPOSER_INSTALL', $file);
+
+        break;
+    }
+}
+unset($file);
```
borNfreee: do we need a `unset()` function fo non-reference var?

borNfreee:
Thank you so much for this awesome addition. I've just tested and it works nice.
I have only one small question about code and ready to merge it.

app/bootstrap.php
theofidry: the variable has been defined just above with the `foreach()`, it doesn't get unset as you exit the loop. Not sure if it's worth to remove unset though, or if you do I would unset `files` as well or maybe do that in a closure to avoid having to unset any variable

app/bootstrap.php
```diff
@@ -4,6 +4,32 @@
  *
  * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
  */
-require __DIR__ . '/../vendor/autoload.php';
+
+$files = [
+    __DIR__ . '/../../../autoload.php',
```
theofidry: doesn't it requires `DIRECTORY_SEPARATOR` to work properly on Windows?

app/bootstrap.php
corpsee: @borNfreee You are right. Fixed.

app/bootstrap.php
corpsee: It does not matter. It works on Windows in both cases.

borNfreee:
Thank you @corpsee

Issue #23, opened 2017-08-29
SourceDirGuesser failure


3DFace:
Trying to run first time, some troubles with SourceDirGuesser:

```
D:\Projects\php\generic-storage>D:\Tools\php7\php.exe D:\Tools\infection.phar

Welcome to the Infection config generator

We did not find configuration file. The following questions will help us to generate it for you.


PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Infection\Config\Guesser\SourceDirGuesser::Infection\Config\Guesser\{closure}() must be of the type string, array given in phar://D:/Tools/infection.phar/src/Config/Guesser/SourceDirGuesser.php:42
Stack trace:
#0 [internal function]: Infection\Config\Guesser\SourceDirGuesser->Infection\Config\Guesser\{closure}(Array)
#1 phar://D:/Tools/infection.phar/src/Config/Guesser/SourceDirGuesser.php(45): array_map(Object(Closure), Array)
#2 phar://D:/Tools/infection.phar/src/Config/Guesser/SourceDirGuesser.php(33): Infection\Config\Guesser\SourceDirGuesser->getValues('psr-0')
#3 phar://D:/Tools/infection.phar/src/Config/ValueProvider/SourceDirsProvider.php(46): Infection\Config\Guesser\SourceDirGuesser->guess()
#4 phar://D:/Tools/infection.phar/src/Command/ConfigureCommand.php(50): Infection\Config\ValueProvider\SourceDirsProvider->get(Object(Symfony\Component\Console\Input\ArrayInput), Object(Symfony\Component\Console\Output\ConsoleOutput), Array)
#5 phar://D:/Tools/infection.p in phar://D:/Tools/infection.phar/src/Config/Guesser/SourceDirGuesser.php on line 42
```

sidz:
@3DFace could you please attach `composer.json` from the `generic-storage` folder?

sidz:
So, seems like you have invalid configuration of the `autoload` section and it looks like this:
```
"autoload": {
        "psr-4": [{
            "VendorName\\": "src/"
        }]
    },
```
but should looks like:
```
"autoload": {
        "psr-4": {
            "VendorName\\": "src/"
        }
    },
```

p.s. If I'm wrong could you please put `autoload` part from your composer.json file?

3DFace:
Here is my autoload section:
```
	"autoload": {
		"psr-0": {
			"": ["src/", "tests/"]
		}
	}
```
[...]

Issue #21, opened 2017-08-22
Dependency on sebastian/diff


sergekukharev:
Hi!

Wanted to try out Infection, but could not install it in my environment:
```
composer require infection/infection --dev
```

I know that recommended installation is through PHAR or global composer installation, but I would like to track my dependencies, especially if I'm adding `infection.json.dist` to my project.

The problem is with the `sebastian/diff` package, I got conflicts:
```
- Installation request for infection/infection ^0.5.1 -> satisfiable by infection/infection[0.5.1].
    - Conclusion: remove sebastian/diff 2.0.1
    - Conclusion: don't install sebastian/diff 2.0.1
    - infection/infection 0.5.1 requires sebastian/diff ^1.4 -> satisfiable by sebastian/diff[1.4.0, 1.4.1, 1.4.2, 1.4.3].
    - Can only install one of: sebastian/diff[1.4.0, 2.0.1].
    - Can only install one of: sebastian/diff[1.4.1, 2.0.1].
    - Can only install one of: sebastian/diff[1.4.2, 2.0.1].
    - Can only install one of: sebastian/diff[1.4.3, 2.0.1].
    - Installation request for sebastian/diff (locked at 2.0.1) -> satisfiable by sebastian/diff[2.0.1].
```

`composer why sebastian/diff` hints me that my PHPUnit version (6.3.0) requires sebastian/diff ^2.0.

Let's think together what can be done, cause simply updating to ^2.0 might break older PHPUnit version projects. It's weird how a 3rd package dependency makes your project dependent on specific PHPUnit version.

borNfreee:
Thank you for trying out Infection, Serge.

The problem you are talking about is not so easy. I've investigated it a bit, and here is what I found:

#### composer-bin-plugin

https://github.com/bamarni/composer-bin-plugin - "No conflicts for your bin dependencies"

This composer plugin allows you to install bin vendors in isolated locations. I recommend to read its description, there are very impressive examples.

This plugin will definitely solve an issue, but you will have to add this plugin as a dependency.

There are several other repositories/libs that try to solve conflicting classes issue:

#### PHPStan (static analyzer):

https://github.com/phpstan/phpstan-shim + https://github.com/fprochazka/phpstan-compiler. Sp they created a different composer package with just one phar `phpstan.phar` file, which has all files' namespaces prefixed in order to avoid conflicts.

In this case you will depend on composer package, but actually will run phar file

```bash
vendor/bin/phpstan.phar analyse src
```

[...]

PR #25, opened 2017-08-30
Fix autoload section from composer.json and allow to use multiple paths

sidz:
Fix for https://github.com/infection/infection/issues/23 issue.

I've added some rules + few exceptions.
Didn't find any valid way to parse composer.json without including JsonLint library. I think it'll be overhead.

### DIFF
--- a/tests/Config/Guesser/SourceDirGuesserTest.php
+++ b/tests/Config/Guesser/SourceDirGuesserTest.php
@@ -7,8 +7,7 @@
 
 declare(strict_types=1);
 
-
-namespace Infection\Tests\Guesser;
+namespace Infection\Tests\Config\Guesser;
 
 use Infection\Config\Guesser\SourceDirGuesser;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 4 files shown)


tests/Config/Guesser/SourceDirGuesserTest.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+namespace Infection\Tests\Config\Guesser;
+
+use Infection\Config\Guesser\SourceDirGuesser;
+use PHPUnit\Framework\TestCase;
+
+class SourceDirGuesserTest extends TestCase
```
borNfreee: Just found another `SourceDirGuesserTest` file at the incorrect path (`src/tests/Guesser`).

Could you please move tests from that file to your file and remove it?

borNfreee:
> Didn't find any valid way to parse composer.json without including JsonLint library. I think it'll be overhead.

You are right, I would avoid including a new dependency as well because of problems outlined in #21

PR looks good to me, thank you! Will be merged today if nobody has any other comments.

tests/Config/Guesser/SourceDirGuesserTest.php
sidz: @borNfreee done

borNfreee:
Thank you @sidz !

PR #32, opened 2017-09-06
Interface mutations fix

oxidmod:
- Skip all mutations for interfaces
- Add uncovered mutations to to file and console log

### DIFF
--- a/src/Mutator/InterfaceParentTrait.php
+++ b/src/Mutator/InterfaceParentTrait.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ * Copyright © 2017 Volodimir Melko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+declare(strict_types=1);
+
+namespace Infection\Mutator;
+
+use Infection\Visitor\ParentConnectorVisitor;
+use PhpParser\Node;
+
+/**
+ * Checks if given node belongs to Interface
+ */
+trait InterfaceParentTrait
+{
+    private function isBelongsToInterface(Node $node): bool
+    {
+        $parentNode = $node->getAttribute(ParentConnectorVisitor::PARENT_KEY);
+
+        if ($parentNode instanceof Node\Stmt\Interface_) {
+            return true;
+        }
+
+        if ($parentNode instanceof Node) {
+            return $this->isBelongsToInterface($parentNode);
+        }
+
+        return false;
+    }
+}
--- a/src/Mutator/InterfaceParentTrait.php
+++ b/src/Mutator/InterfaceParentTrait.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Copyright © 2017 Volodimir Melko
+ * Copyright © 2017 Maks Rafalko
  *
  * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
  */
@@ -13,6 +13,8 @@ use PhpParser\Node;
 
 /**
  * Checks if given node belongs to Interface
+ *
+ * @author Volodimir Melko <v.melko28@gmail.com>
  */
 trait InterfaceParentTrait
 {
--- a/src/Process/Listener/MutationTestingConsoleLoggerSubscriber.php
+++ b/src/Process/Listener/MutationTestingConsoleLoggerSubscriber.php
@@ -97,15 +97,11 @@ class MutationTestingConsoleLoggerSubscriber implements EventSubscriberInterface
         $this->outputFormatter->finish();
 
         if ($this->showMutations) {
-            $this->showMutations(
-                $this->metricsCalculator->getEscapedMutantProcesses(),
-                'Escaped'
-            );
+            $this->showMutations($this->metricsCalculator->getEscapedMutantProcesses(), 'Escaped');
+        }
 
-            $this->showMutations(
-                $this->metricsCalculator->getEscapedMutantProcesses(),
-                'Uncovered'
-            );
+        if ($this->showMutations && $this->output->getVerbosity() > OutputInterface::VERBOSITY_NORMAL) {
+            $this->showMutations($this->metricsCalculator->getNotCoveredMutantProcesses(), 'Not covered');
         }
 
         $this->showMetrics();
@@ -113,9 +109,12 @@ class MutationTestingConsoleLoggerSubscriber implements EventSubscriberInterface
 
     private function showMutations(array $processes, string $headlinePrefix)
     {
+        $headline = sprintf('%s mutants:', $headlinePrefix);
+
         $this->output->writeln([
             '',
-            sprintf('%s mutants:', $headlinePrefix),
+            $headline,
+            str_repeat('=', strlen($headline)),
             '',
         ]);
 
--- a/src/Process/Listener/MutationTestingConsoleLoggerSubscriber.php
+++ b/src/Process/Listener/MutationTestingConsoleLoggerSubscriber.php
@@ -98,10 +98,10 @@ class MutationTestingConsoleLoggerSubscriber implements EventSubscriberInterface
 
         if ($this->showMutations) {
             $this->showMutations($this->metricsCalculator->getEscapedMutantProcesses(), 'Escaped');
-        }
 
-        if ($this->showMutations && $this->output->getVerbosity() > OutputInterface::VERBOSITY_NORMAL) {
-            $this->showMutations($this->metricsCalculator->getNotCoveredMutantProcesses(), 'Not covered');
+            if ($this->output->getVerbosity() > OutputInterface::VERBOSITY_NORMAL) {
+                $this->showMutations($this->metricsCalculator->getNotCoveredMutantProcesses(), 'Not covered');
+            }
         }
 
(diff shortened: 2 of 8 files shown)


borNfreee:
For everyone who will read this, I want to explain why we don't need to mutate `Interface`s with `PublicVisibility` mutator.

1. First, it does not make sense, since interfaces can have only public methods
2. CodeCoverage *does not* have coverage data for interfaces - this is how xDebug+PHPUnit works actually, so, currently, infection thinks that mutant is `uncovered` because file with interface has empty coverage data. See the output for [this repository](https://github.com/oxidmod/rest-bundle/tree/repo/mutation-tests)

```bash
Uncovered Mutants
===============

1) Infection\Mutator\FunctionSignature\PublicVisibility
/Users/user/tmp/rest-bundle/src/Request/Modifier/RequestModifierInterface.php:19

2) Infection\Mutator\FunctionSignature\PublicVisibility
/Users/user/tmp/rest-bundle/src/Response/Modifier/ResponseModifierInterface.php:21

...
```

While I'm ok with disabling `PublicVisibility` mutator for interfaces, I am not sure about disabling **all** mutators for interfaces (like you do in PR). 

For example, we can have a new mutator that adds `extends \Countable` for interfaces, and if tests pass with such Mutant, then we will know that interface is unused. But this is arguable mutant since I'm not sure whether it is task of mutation testing framework or static analyzer tool to detect unused interfaces.

Alternative solution would be to update just one `PublicVisibility` mutator and return `false` from `shouldMutate` method if it's used inside interface. But this is hard to implement right now, because mutator does not have knowledge about context - where it is used (class, abstract class, interface, etc.). This is a big task to implement later.

So now I'm ok with removing interfaces from mutation process (if nobody sees potential issues), but later we will revise this.

src/Process/Listener/TextFileLoggerSubscriber.php
```diff
@@ -76,12 +81,15 @@ private function getLogParts(array $processes, string $headlinePrefix): array
         $logParts = [sprintf('%s mutants:', $headlinePrefix), ''];
 
         foreach ($processes as $index => $mutantProcess) {
+            $mutant = $mutantProcess->getMutant();
```
borNfreee: Just before your PR, I pushed a new branch with similar changes, but (from my point of view of course) with a bit prettier output

Please see https://github.com/infection/infection/compare/better-log?expand=1

Example of output for your repo:

```bash
Not covered mutants:
====================


1) Infection\Mutator\FunctionSignature\PublicVisibility
/Users/user/tmp/rest-bundle/src/Request/Modifier/RequestModifierInterface.php:19

2) Infection\Mutator\FunctionSignature\PublicVisibility
/Users/user/tmp/rest-bundle/src/Response/Modifier/ResponseModifierInterface.php:21

3) Infection\Mutator\Boolean\TrueValue
/Users/user/tmp/rest-bundle/src/Response/ResponseRenderer.php:71

4) Infection\Mutator\FunctionSignature\PublicVisibility
/Users/user/tmp/rest-bundle/src/Transformer/TransformerInterface.php:21

5) Infection\Mutator\FunctionSignature\PublicVisibility
/Users/user/tmp/rest-bundle/src/Transformer/TransformerInterface.php:28
```

It's better because

1. it has `file:lineNum` notation which perfectly works with `ctrl+click` and opens exact line in your editor
2. it does not have diff for uncovered mutation, this *significantly* reduces the log
3. it has two separate methods where not needed information is removed for uncovered tests (e.g. we don't need command line for it, we don't need process output (even `''`))

Besides that, there is some changes to make ouput more readable.

I suggest to use code from that branch, or merge it to your branch somehow

src/Process/Listener/MutationTestingConsoleLoggerSubscriber.php
```diff
@@ -95,18 +95,29 @@ public function onMutationTestingFinished(MutationTestingFinished $event)
...
-        $processes = $this->metricsCalculator->getEscapedMutantProcesses();
 
         if ($this->showMutations) {
-            $this->showMutations($processes);
+            $this->showMutations(
+                $this->metricsCalculator->getEscapedMutantProcesses(),
+                'Escaped'
+            );
+
+            $this->showMutations(
```
borNfreee: I'm -1 on this addition, because from my experiense almost every big enough project has thousands (literally) uncovered mutations. It will just pollute the terminal without possibility to scroll up (because of terminal limitations).

Such long text should be inside text file only - `infection-log.txt`.

src/Process/Listener/TextFileLoggerSubscriber.php
oxidmod: Thanks for reply!
I agree with 1 and 3 and i will try to merge your changes.
In my opinion the log should be detailed as possible. I have an example, but I can show it in the evening.

src/Process/Listener/MutationTestingConsoleLoggerSubscriber.php
oxidmod: I thought about it, too. We could adjust the detail of the output relying on the "-vvv" option.

borNfreee: Yes, let's have displayed escaped mutations by default with `-s`, and uncovered with a verbose level

oxidmod: Ok. I will fix this PR as soon as possible.

src/Process/Listener/TextFileLoggerSubscriber.php
oxidmod: There is an example of the last uncovered mutation:
```diff
1) Infection\Mutator\Boolean\TrueValue
/mmedia/projects/oxid-rest-bundle/src/Response/ResponseRenderer.php
exec /usr/bin/php7.1 /mmedia/projects/oxid-rest-bundle/vendor/phpunit/phpunit/phpunit --configuration /tmp/infection/phpunitConfiguration.1454e421d53db78e46101421c0968c75.infection.xml --stop-on-failure
--- Original
+++ New
@@ @@
     {
-        switch (true) {
+        switch (false) {
             case $data === null:
                 return new NullResource();
             case is_array($data):
             case $data instanceof \Traversable:
                 return new Collection($data, $this->transformer, $resourceKey);
             default:
```

Code coverage still 100%. It looks like false positive and with logs i can check it.

borNfreee: I've created an issue in PHPUnit_Coverage repository. Please read/track it here: https://github.com/infection/infection/issues/34

src/Process/Listener/MutationTestingConsoleLoggerSubscriber.php
```diff
@@ -95,18 +95,28 @@ public function onMutationTestingFinished(MutationTestingFinished $event)
     {
         // TODO [doc] write test -> run mutation for just this file. Should be 100%, 100%, 100%,
         $this->outputFormatter->finish();
-        $processes = $this->metricsCalculator->getEscapedMutantProcesses();
 
         if ($this->showMutations) {
```
borNfreee: would it better to move the second condition with verbosity under this one?

```php
if ($this->showMutations) {
    // do something

    if ($this->output->getVerbosity() > OutputInterface::VERBOSITY_NORMAL) {
       // do something else
    }
}
```

borNfreee:
Looks good 👍  Just left a small comment and ready to merge it

src/Process/Listener/MutationTestingConsoleLoggerSubscriber.php
oxidmod: You are right. One moment!

oxidmod:
I finished here. Check it again please.

src/Mutator/InterfaceParentTrait.php
```diff
@@ -0,0 +1,33 @@
+<?php
+/**
+ * Copyright © 2017 Volodimir Melko
```
borNfreee: I think this is an autogenerated mistake. It should be as in other files - holder's name.
Not a big deal, but will be overridden by my auto-license-generating tool.

You can (and actually deserve ;)) to add your name as `@author` annotation, like e.g. symfony contributors do: https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Command/AboutCommand.php#L26

src/Mutator/InterfaceParentTrait.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Infection\Mutator;
+
+use Infection\Visitor\ParentConnectorVisitor;
+use PhpParser\Node;
+
+/**
+ * Checks if given node belongs to Interface
+ */
+trait InterfaceParentTrait
```
borNfreee: why do we need trait here?

1. this code is not used anywhere else
2. maybe it should be just in mutator?

I'm not a big fan of traits, but probably you have good arguments?

oxidmod: I was thinking about other potential mutators that would not apply to the interface. In this case, duplicate code will appear

borNfreee:
Thank you @oxidmod! Very useful improvements

PR #40, opened 2017-09-12
Add Filesystem and add Directory Check for logs path.

sidz:
Implemented Filesystem class.

For example to prevent cases when user wants to put his logs into `./my-test/folder/infection-log.txt`.
I've implemented `mkdir` method only.

### DIFF
--- a/app/container.php
+++ b/app/container.php
@@ -15,6 +15,7 @@ use Infection\Mutant\MutantCreator;
 use Infection\Command\InfectionCommand;
 use Infection\Process\Runner\Parallel\ParallelProcessRunner;
 use Infection\EventDispatcher\EventDispatcher;
+use Infection\Filesystem\Filesystem;
 use Infection\Finder\Locator;
 use Infection\TestFramework\PhpUnit\Config\Path\PathReplacer;
 use Infection\TestFramework\Config\TestFrameworkConfigLocator;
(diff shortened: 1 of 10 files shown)


borNfreee:
There was a failed build for this PR because of a badly written test. I've refactored it to use test dirs instead of real project dirs (`Files` and newly create `Filesystem`).

Currently build still fails because of decreased MSI https://travis-ci.org/infection/infection/jobs/274793692#L652

I will look into it tomorrow, seems like new code/tests do not kill all the mutants.

borNfreee:
During writing more tests to increase MSI, I found a critical bug in the `IncludeInterceptor.php` file.
There is a parameter that contains a bit mask with options for wrapped stream, method `mkdir`, and it was used incorrectly

This caused the following escaped mutant:

```diff
1) Infection\Mutator\Boolean\TrueValue
/Users/user/tmp/sidz/src/Filesystem/Filesystem.php

--- Original
+++ New
@@ @@
         }
-        if (true !== @\mkdir($path, $mode, true)) {
+        if (true !== @\mkdir($path, $mode, false)) {
             $error = \error_get_last();
             if ($error) {
                 throw new IOException(\sprintf('Failed to create "%s": %s', $path, $error['message']), 0, null, $path);
             }
             throw new IOException(\sprintf('Failed to create "%s"', $path), 0, null, $path);
         }
```

This is because `$options` parameter is a bit mask (e.g. with value `469`) that was always converted to `true` and was passed as a `recursive` parameter even if it was set to `false` in the original code.

Now it's rewritten as `$isRecursive = (bool) ($options & STREAM_MKDIR_RECURSIVE);` and works as expected.

------

If you are ok @sidz with my changes in tests, I can merge it.

sidz:
@borNfreee fine for me.

borNfreee:
Thank you @sidz

Issue #38, opened 2017-09-11
Idea: Reimplement timeout as percentage value / float


JanPietrzyk:
Hi,

I am currently moving an existing project from humbug to infection and have an issue related to the current implementation of the timeout property.

My existing project has a few exceptional tests that take a lot more time then 10 seconds, but most tests are < 1 second of course. Now in order to get infection to execute the test suite initially and generate the coverage  I have to set a high value as timeout. This does only extend the time infection then takes to detect infinite loops in the actual mutation run.

My current solution is to exclude these tests through a custom config. Nevertheless I think there is room for some improvement in infection itself.

My Suggestion:
- Remove the timeout from the initial run alltogether and trust the users here.
- Refactor the timeout into a percentage value. Mark tests that take X% longer then when they were executed initially as timeout.

borNfreee:
Thanks for posting your idea Jan.

> Now in order to get infection to execute the test suite initially and generate the coverage I have to set a high value as timeout. This does only extend the time infection then takes to detect infinite loops in the actual mutation run. 
> [...]
> Remove the timeout from the initial run alltogether and trust the users here.

Let's think together about what the `timeout` property is.

https://infection.github.io/guide/usage.html#Configuration-settings

> `timeout` - the allowed timeout configured for Infection. Make sure to set it to higher value than your tests are executed in seconds to avoid false-positives.

So, if your tests take 1 minutes (does not matter that 1 test takes 59 seconds and all others just one second) - timeout should be at least 1 minute. 

Why? Because infection (and actually humbug too) runs all tests that executed on the mutated line. If you have e.g. 10 tests and all these 10 tests cover all lines, then each mutation process will run all tests, right?

So I don't understand why you want to have `timeout=1s` when your tests (the whole test suite) take much longer - 1+59=60s.

I agree that probably, we need to remove timeout from the initial tests run, but I don't agree that we need to have timeout less in seconds than the whole test suite takes.

Could you please post your thoughts about this?

-----

> Refactor the timeout into a percentage value. Mark tests that take X% longer then when they were executed initially as timeout.

I can suggest having both - timeout float value (current implementation) and a percentage (new functionality). This is exaclty what Pitest (Java mutation testing framework) has: http://pitest.org/quickstart/maven/ (see `timeoutFactor` and `timeoutConstant`)

Do you want to contribute ;) ?

JanPietrzyk:
hmm, this leaves a lot of ideas in the wild ;) Let me try to sort my opinion into this...

[...]

PR #41, opened 2017-09-12
remove timeout from initial test run

JanPietrzyk:
as discussed in #38 this removes the timeout from initial run

### DIFF
--- a/tests/Process/Builder/ProcessBuilderTest.php
+++ b/tests/Process/Builder/ProcessBuilderTest.php
@@ -1,6 +1,6 @@
 <?php declare(strict_types=1);
 
-namespace Infection\Tests\Process\Listener;
+namespace Infection\Tests\Process\Builder;
 
 use Infection\Mutant\Mutant;
 use Infection\Process\Builder\ProcessBuilder;
@@ -38,4 +38,8 @@ class ProcessBuilderTest extends TestCase
         $this->assertSame(100.0, $process->getTimeout());
     }
 
+    protected function tearDown()
+    {
+        \Mockery::close();
+    }
 }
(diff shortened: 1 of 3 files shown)


tests/Process/Builder/ProcessBuilderTest.php
```diff
@@ -0,0 +1,41 @@
+<?php declare(strict_types=1);
+
+namespace Infection\Tests\Process\Listener;
```
borNfreee: this namespace is incorrect and leads to failed build https://travis-ci.org/infection/infection/jobs/274783226#L603

It should be `Infection\Tests\Process\Builder` I guess

tests/Process/Builder/ProcessBuilderTest.php
```diff
@@ -0,0 +1,41 @@
...
+
+use Infection\Mutant\Mutant;
+use Infection\Process\Builder\ProcessBuilder;
+use Infection\TestFramework\AbstractTestFrameworkAdapter;
+use PHPUnit\Framework\TestCase;
+
+
+class ProcessBuilderTest extends TestCase
```
borNfreee: tests with Mockery should contain 

```php
protected function tearDown()
{
    Mockery::close();
}
```

as other tests do.

Unfortunately, we still can't use Mockery's [test listener](https://github.com/mockery/mockery/blob/master/library/Mockery/Adapter/Phpunit/TestListener.php) because it does not support PHPUnit 6+, so we have to do it manually

JanPietrzyk:
@borNfreee fixed

borNfreee:
Thank you @JanPietrzyk!

PR #42, opened 2017-09-14
Add option to pass test framework extra options

borNfreee:
This feature allows to pass additional options to the testing framework.

Use case: when you want to run only particular tests, you can run

```bash
infection.phar --test-framework-options="--filter=just/unit/tests"
```

`--filter` options for PHPUnit is a special one, because we can add it to executable command only for *initial* test run. For each mutation process, it can't be used because we already have filtered out list of executed tests which should not be overridden by this option.

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -44,6 +44,12 @@ class InfectionCommand extends Command
                 'phpunit'
             )
             ->addOption(
+                'test-framework-options',
+                null,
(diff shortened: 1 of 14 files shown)


theofidry:
How is the test framework selected? Those options would be passed to the currently configured test framework?

borNfreee:
> How is the test framework selected?

by default it's `phpunit`, but it can be changed by specifying option `--test-framework=phpspec`

> Those options would be passed to the currently configured test framework?

yes

PR #44, opened 2017-09-26
Fix performance bottleneck by introducing a simple instance-level object cache

borNfreee:
There was a performance issue with opening and parsings (XPath) `junit.xml` thousands of times instead of caching the results.

This update eliminates the issue.

Blackfire comparison before/after: https://blackfire.io/profiles/compare/a6ef3b53-93c5-44b4-bf16-9c2510b44e61/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=&callname=main()

On infection itself, the number of `DomDocument::loadXML` calls reduced by almost 8k times.

blackfire

### DIFF
--- a/app/container.php
+++ b/app/container.php
@@ -26,6 +26,7 @@ use Infection\Differ\DiffColorizer;
 use Infection\TestFramework\PhpUnit\Adapter\PhpUnitAdapter;
 use Infection\Config\InfectionConfig;
 use Infection\Utils\VersionParser;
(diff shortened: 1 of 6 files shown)


snapshotpl:
What are you think to make this change as decorator for `TestFileDataProvider` interface? It will be easly to to test and reuse with other implementations

borNfreee:
@snapshotpl great idea! Done. Would you mind reviewing the code?

PR #49, opened 2017-10-07
Compatibility with PHPUnit 6.4

morozov:
1. The `Differ` class in [sebastian/diff](https://packagist.org/packages/sebastian/diff) 2.0 is `final`, so using composition instead of inheritance. This is compatible with both versions 1.4 and 2.0.
2. The new `Differ` doesn't return trailing matching lines so there's no need to trim the diff (must be changed by https://github.com/sebastianbergmann/diff/pull/59).
3. ~_composer.lock_ is removed since infection can be used as a development dependency and shouldn't enforce its own installed package versions.~


borNfreee:
Thank for creating this PR @morozov. Let's discuss some points before merging.

1.  👍  for composition.

--------

2. Seems like new Differ strips the common lines from the diff.

As for me, having such diff (Differ ^1.4)
```diff
--- Original
+++ New
@@ @@
     public function foo()
     {
-         return 2 + 1;
+         return 2 + 1;
     }
 }
```

is much better than having this one (Differ ^2.0.1)

```diff
--- Original
+++ New
@@ @@
-         return 2 + 1;
+         return 2 + 1;
```

because it's more clear where the mutation was done - we know the context - function `foo()`.
In Differ 2.0.1 all common chunks (the same lines in diff) are now removed from the diff output.

* [removing common chunks before mutated line](https://github.com/sebastianbergmann/diff/blob/master/src/Output/UnifiedDiffOutputBuilder.php#L66) 
* [removing common chunks after mutated line](https://github.com/sebastianbergmann/diff/blob/master/src/Output/UnifiedDiffOutputBuilder.php#L94)

So if we want to save behavior and display the same diff output as for Differ 1.4, we need to implement our own `OutputBuilder` and pass it to `Differ` constructor. Am I right?
I'm not sure whether it worths efforts and would like to hear your/others opinion.

--------

3. It does not enforce own installed version, please look at the https://getcomposer.org/doc/02-libraries.md#lock-file

> [...] this lock file will not have any effect on other projects that depend on it. It only has an effect on the main project.

However, I've checked several other open source CLI libs that can be installed as a dependency and none of them have `composer.lock`. 

Are there any drawbacks of having/removing `composer.lock`?

theofidry:
>Are there any drawbacks of having/removing composer.lock?

In a library, it makes no difference for the consumer. It does for the library maintainer:

- if a `composer.lock` is present, when testing with different config in the CI, `composer update` must be used as `composer install` would defeat the purpose of trying different environments
- for a library published in a PHAR, a PHAR is like a regular application: if there is no `composer.lock`, you are not 100% sure of what will be shipped.

morozov:
@borNfreee when I run the modified test with sebastian/diff 2.0, the [$diff](https://github.com/morozov/infection/blob/2ecd4d37870ab553fbdbeb88f83d6f488617afb6/tests/Differ/DifferTest.php#L79) generated by the base `Differ` looks like:
```diff
--- Original
+++ New
@@ @@
 public function diff($from, $to, LongestCommonSubsequence $lcs = null)
 {
-    $diff = parent::diff($from, $to, $lcs);
+    $diff = parent::diff($from, $to, $lcs);;
```
So there is still some context provided. How did you build your diffs?

> It does not enforce own installed version, please look at the https://getcomposer.org/doc/02-libraries.md#lock-file

My understanding of why or why not commit the lock file is incomplete and is based on that it's a common practice to not commit the lock file for libraries.

If the code is distributed both as an application and a library, and in the first case the lock file is considered but in the second is ignored (according to the link above), then depending on the way distribution, composer will install either locked, or the newest dependencies which may be unexpected. Without the lock file committed, it will install the newest dependencies in all cases which is consistent.

> for a library published in a PHAR, a PHAR is like a regular application: if there is no `composer.lock`, you are not 100% sure of what will be shipped.

The newest versions of the dependencies as of the moment of build matching your requirements specified in `composer.json` will be shipped. According to the link above, it's the same set of dependencies if someone installs the app as a dependency, and the lock file is ignored.

borNfreee:
> How did you build your diffs?

I tried to run infection on itself on your branch. Please see

<details>

```diff
Escaped mutants:
================


1) Infection\Mutator\Boolean\FalseValue
/Users/user/tmp/infection64/src/Config/Guesser/PhpUnitPathGuesser.php

--- Original
+++ New
@@ @@
-            if (strpos($namespace, 'SymfonyStandard') !== false && trim($parsedPath, '/') === 'app') {
+            if (strpos($namespace, 'SymfonyStandard') !== true && trim($parsedPath, '/') === 'app') {


2) Infection\Mutator\Boolean\LogicalAnd
/Users/user/tmp/infection64/src/Config/Guesser/PhpUnitPathGuesser.php

--- Original
+++ New
@@ @@
-            if (strpos($namespace, 'SymfonyStandard') !== false && trim($parsedPath, '/') === 'app') {
+            if (strpos($namespace, 'SymfonyStandard') !== false || trim($parsedPath, '/') === 'app') {


3) Infection\Mutator\Boolean\TrueValue
/Users/user/tmp/infection64/src/Config/Guesser/SourceDirGuesser.php

--- Original
+++ New
@@ @@
-        if (in_array('src', $dirs, true)) {
+        if (in_array('src', $dirs, false)) {
```

</details>

morozov:
The reason why I decided to remove `composer.lock` as part of this PR is that once I modified `composer.json` and tried to `composer update phpunit/phpunit` or `composer update phpunit/phpunit --with-dependencies`, it didn't work:
```
cd infection # the master branch
git checkout phpunit64 -- composer.json
git diff --cached
composer update phpunit/phpunit

Loading composer repositories with package information
Updating dependencies (including require-dev)
Nothing to install or update
Writing lock file
Generating autoload files
```
This led me to the confustion that the PHPUnit (or other dependency) versions are locked.

At the same time, removing the lock file and the vendor directory solved the problem:
```
rm -r vendor/ composer.lock 
composer show | grep phpunit/phpunit
phpunit/phpunit                    6.4.0   The PHP Unit Testing framework.
phpunit/phpunit-mock-objects       4.0.4   Mock Object library for PHPUnit
```
I might have been doing something wrong. Could you explain what exactly?

morozov:
> I tried to run infection on itself on your branch.

@borNfreee thanks, I'll have to take a closer look and see why it's different from what's covered by the test.

theofidry:
>@theofidry the newest versions of the dependencies as of the moment of build matching your requirements specified in composer.json will be shipped

Precisely, this may be different from what you use locally though. Perfect example: PHP-Scoper. As a library, we allow `nikic/php-parser: ^3.0` because the library can work with it. This a perfectly valid constraint. However when shipping the PHAR, we ship isolated code which is a very brittle process cf. [that config](https://github.com/humbug/php-scoper/blob/master/scoper.inc.php#L46). Even a patch update can break things. This is a case where you want to have a 100% control on the deps shipped and this is what a lock file is for, be it a CLI PHAR or a regular application.

Now @borNfreee noted that a lot of lib providing a CLI distributed code do not include a `composer.lock`. I've got two reasons:

- laziness
- unawareness

Now to expand on the first one, I don't necessarily mean it in a bad way. I've got multiple examples:

- In the case of PHP-Scoper, we actually could do without because the code doesn't need to be isolated for its PHAR: it doesn't execute any code except for loading the config which is very unlikely to cause an issue given the Finder stability. It can, but it's very unlikely so potentially not worth caring. We do though because it's also a way for us to more robust end-to-end tests. Simply put we believe the effort was worth it but it was by no means a necessity.
- In a tool like phpstan, it doesn't execute code either (I'm lying. It does right now but there is now a way to avoid that and they are working on it). As a result, if it's possible that a peculiar combination of deps breaks the app, this is much more likely to happen in the CI with an update or to a user installing it as a Composer dep that happening to the code shipped to the PHAR. So the risk is so low, it's just not worth caring for it.
- In a tool like PHPUnit, this is because there is a bigger issue anyway so it doesn't really matter. Indeed one issue with PHPUnit is that it requires a non negligible amount of dependencies. In an application, this may be annoying as it can pollute your application (is it to the level you should really care about it? That's another story). In the case of a library, this is much more problematic. Indeed for a library, the more dev deps you have, the more likely there will be a difference between the deps you will test your library with and the deps your user will have when installing your library. A simple example would be imagine you have `symfony/yaml: ^2.8||^3.0` in your library `composer.json`, but a dev dependency (direct or not) requires `symfony/yaml: ^3.0`. The result is, unless you are aware of the issue and tweak your CI, you will never test your library with `symfony/yaml: [2.8 ; 3.0[`. Again, maybe this is so unlikely to happen you just don't care enough to actually watch out for that, but this is a problem that exists nonetheless. So now back to PHPUnit, the problem above is one of the reason why some people like Symfony like to use a PHAR instead of installing it as a dependency. Now however PHPUnit does execute your code, so you might run in a conflict of version. If `symfony/yaml 2.8` is shipped the PHAR and your application relies on some code specific to `symfony/yaml 3.0`, it will be the one shipped in the PHAR that will be used, hence it will break. In the case of Symfony however this problem is mitigated by their PHPUnit-simple which is a tweak of PHPUnit to reduce that likelihood, and the sheer number of contributors also guarantee a fast error reporting when the problem still occurs. So while that problem exists (there is plans to tackle, PHP-Scoper being a major piece in that plan), the issue that may cause the lack of use of the `composer.lock` is not even worth to care about.

TL:DR; It's ok not to care. Sometimes your prefer (legit reasons or not) to go through the risk of triggering a known bug than fixing it.

morozov:
I've reverted the removal of the lock file since it's not that straightforward as I thought and is not really a dependency for this patch. Thank you both for the valuable input.

The side effects of upgrading sebastian/differ still need some work.

borNfreee:
Let me know if you need any help here.

I want to release 0.6.0 verions soon and would like to have PHPUnit 6.4 compatibility as well.

Probably we can revert upgrading `composer.lock` to Differ ^2.0 version and merge it as is, and you will continue investigating differ in a separate PR.

Thus, users will be able to work with Infection+PHPUnit 6.4 but with one-line diff when they use the lib as a dependency.
PHAR will continue working with Differ 1.4 for now.

What do you think?

morozov:
> Probably we can revert upgrading composer.lock to Differ ^2.0 version and merge it as is, and you will continue investigating differ in a separate PR.

That sounds good to me. See the update.

> Thus, users will be able to work with Infection+PHPUnit 6.4 but with one-line diff when they use the lib as a dependency.
PHAR will continue working with Differ 1.4 for now.

I wanted to suggest the same if we don't find a better solution (which I didn't manage to). Right now, I see that `UnifiedDiffOutputBuilder` can add line numbers to the diff, so it may look like this:
```diff
--- Original
+++ New
@@ -20 +20 @@
-        $this->callsCount = 1;
+        $this->callsCount = 0;
```
As an end user, I wouldn't mind getting line numbers instead of surrounding lines. We can take it out of this issue.

borNfreee:
> As an end user, I wouldn't mind getting line numbers instead of surrounding lines.

Sounds good! Line number will definitely help. Let's sort it out in a separate issue so I can merge this one.

Thank you @morozov !

morozov:
> Line number will definitely help. Let's sort it out in a separate issue so I can merge this one.

The line numbers would be definitely nice to have, but given that the file content is reconstructed from the AST, they may be not enough precise and thus misleading (is it a valid concern?).

I've filed sebastianbergmann/diff#69 to see what the design behind the current implementation is. I'll create the new issue based on the outcome of that one.

borNfreee:
> is it a valid concern?

Yes, unfortunately, this is a valid concern. Let's wait for the reply from the issue above

morozov:
@borNfreee could you check if https://github.com/sebastianbergmann/diff/pull/75 makes the diff output better?

borNfreee:
@morozov 

Looks like it works much better now

^2.0:

```
1) src/Converter.php:23    [M] PublicVisibility
exec /bin/php -c /private/var/folders/zb/crskfnsx48l9jnrd33wvc1z80000gn/T/infectionmy4IW9 vendor/phpunit/phpunit/phpunit --con$
--- Original
+++ New
@@ @@
-    public function convert(\SplFileObject $file, string $outputFormat, string $outputFilePath)
+    protected function convert(\SplFileObject $file, string $outputFormat, string $outputFilePath)

PHPUnit 6.4.0 by Sebastian Bergmann and contributors.

E

Time: 31 ms, Memory: 4.00MB

```

dev-master:
```
1) src/Converter.php:23    [M] PublicVisibility
exec /bin/php -c /private/var/folders/zb/crskfnsx48l9jnrd33wvc1z80000gn/T/infectionR69hZg vendor/phpunit/phpunit/phpunit --con$
--- Original
+++ New
@@ @@
         $this->readerFactory = $readerFactory;
         $this->writerFactory = $writerFactory;
     }
-    public function convert(\SplFileObject $file, string $outputFormat, string $outputFilePath)
+    protected function convert(\SplFileObject $file, string $outputFormat, string $outputFilePath)
     {
         $extension = $this->extensionGuesser->guess($file);
         $reader = $this->readerFactory->create($extension);

PHPUnit 6.4.0 by Sebastian Bergmann and contributors.

E

```

However, infection supports PHP 7.0+ while `dev-master` of `sebastian/diff` requires `^7.1`

morozov:
Cool, thank you @borNfreee. I believe it will be fine if the users of PHP 7.1 get the newest diff version while the ones who use PHP 7.0 get the old one. You can call it progressive enhancement :)

PR #51, opened 2017-10-08
Add Break-Continue mutators

sidz:


borNfreee:
Nice mutators! Thank you.

One interesting thing I found is that with current implementation it replaces `break` with `continue` in `switch` statement (which surprisingly is a [valid syntax](http://php.net/manual/en/control-structures.switch.php)) and it leads to the _equivalent mutant_

> _Equivalent Mutant_ does not change the behavior of the system at all. [...] can't be killed.

This works exactly the same https://3v4l.org/7cLT1
```diff

switch (1) {
    case 1:
-        break;
+        continue;
}
```

I think the following test should be added:

```php
public function test_does_not_replace_break_to_continue_in_switch()
    {
        $code = <<<'CODE'
<?php

switch (1) {
    case 1:
        break;
}
CODE;

        $expectedCode = <<<'CODE'
<?php

switch (1) {
    case 1:
        break;
}
CODE;

        $this->assertSame($expectedCode, $this->mutate($code));
    }
```

sidz:
@borNfreee done

borNfreee:
Awesome!

Thank you @sidz

Issue #54, opened 2017-10-19
Hide killed mutants in output log?


eko3alpha:
This library is nothing short of amazing.  Thank you!

Is there an option to hide the killed mutants in the log?  When I run it, I'm only interested in everything but killed mutants and I find the killed mutants list to be a bit noisy when scrolling to find what methods need attention.

Thanks again!

borNfreee:
Hi, thank you for these kind words and for using Infection!

I'm afraid I didn't get you right. Do you mean you want to exclude information about Killed mutants from the `infection-log.txt`? If so, this is not possible at the moment.

Why is it noisy? Killed mutants are at the bottom of the file, *before* Escaped and Timeouts.

I can imagine that in a big project with thousands of mutants and a big MSI percentage those killed mutants just pollute the log file, so I'm ok with adding a new option to generate "lightweight" log file. 

Any ideas on the naming of such option?  Would you like to contribute?

theofidry:
@borNfreee what about simply "log-verbosity"?

borNfreee:
@theofidry great suggestion 👍 . I like it because we can control not just types of mutants in the log, but also additional meta information like executed commands, diffs, process outputs, etc.

Just one question - what will be a default verbosity level for our logs? I would like to have a very verbose level defaulted since sometimes killed mutants are killed because of unexpected [fatal/environment] errors and IMO user should detect it as soon as possible. Is it ok to have such default verbosity level? (e.g. symfony has default verbosity level `NONE`)

theofidry:
I'm up for verbose mode by default, although it might be worth mentioning somewhere _why_. For some people fatal/environment errors are acceptable as well. I don't have an example on top of my head but I remember coming across this case a few times

eko3alpha:
I like the idea of verbosity,  that would remove some of the noise in the log file.  My goal with this suggestion is to quickly identify the file and line number so more tests can be added or code rewritten.  The current log format is a firehose of information.  For example, in my case, I'm mainly interested in the type of mutation (with example), file, and line number.  I'll see if I can come up with some suggestions.

eko3alpha:
After looking at my 17,371 line infection-log.txt with 1586 mutants...

My goal is to remove as many of these mutants as possible.  I'm not particularly interested in killed mutants, especially since I can see it in the summary report while the test is running.  When trying to kill a mutant I need the following information:

- Filename (preferably the relative path)
- Line number
- Mutation type
- Mutation

[...]

PR #56, opened 2017-10-22
Add Log verbosity

sidz:
Added 2 levels:
1 - debug/full mode (as it works now)
2 - short mode will show short log format (https://github.com/infection/infection/issues/54#issuecomment-338206510). Like:
```
1) /var/www/infection/src/Utils/TempDirectoryCreator.php:32    [M] TrueValue

--- Original
+++ New
@@ @@
             $path = sprintf('%s/%s', sys_get_temp_dir(), $dirName ?: 'infection');
-            if (!@mkdir($path, 0777, true) && !is_dir($path)) {
+            if (!@mkdir($path, 0777, false) && !is_dir($path)) {
                 throw new \RuntimeException('Can not create temp dir');
             }
             $this->tempDirectory = $path;
         }
         return $this->tempDirectory;
     }
```

- I left old diff format. It looks fine with PublicVisibility mutator but looks weird with Break/Continue for example:
```
- break;
+ continue;
```
- all levels will show line number and mutation name

### DIFF
--- a/src/Console/LogVerbosityInterface.php
+++ b/src/Console/LogVerbosityInterface.php
@@ -1,0 +1,14 @@
+<?php
+/**
+ * Copyright © 2017 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+namespace Infection\Console;
+
+interface LogVerbosityInterface
+{
+    const DEBUG = 1;
+    const NORMAL = 2;
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
 script:
   - vendor/bin/phpunit --coverage-clover=clover.xml
   - composer analyze
-  - if [[ $INFECTION == 1 ]]; then bin/infection --threads=4 --min-msi=58 --min-covered-msi=91; fi
+  - if [[ $INFECTION == 1 ]]; then bin/infection --threads=4 --min-msi=59 --min-covered-msi=91; fi
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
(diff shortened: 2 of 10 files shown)


src/Console/LogVerbosityInterface.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/**
+ * Copyright © 2017 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+namespace Infection\Console;
+
+interface LogVerbosityInterface
```
borNfreee: just a question: should this "constants container" be an interface or a class?

Just didn't see such practice with interfaces that hold constants only without being implemented, but it looks interesting. Is there any read about this? :)

borNfreee:
Good stuff, thank you. 
Several things to consider before merging:

1. Do we need to document this new feature? I think at least this page needs to be updated https://infection.github.io/guide/command-line-options.html

2.  Would you mind adding tests for `TextFileLoggerSubscriber`?  Probably, `file_put_contents` should be moved to `Filesystem` as well since we already have it inside the class, but I'm not sure...

3. Also, I noticed that `--min-msi=58` can be increased to `59` here: https://github.com/infection/infection/blob/0.6.0/.travis.yml#L29

src/Console/LogVerbosityInterface.php
sidz: I won't give you anything to read as this is not a good practice :) (When you interface contains constants only).

I'll create final class instead.

sidz:
@borNfreee all 3 are done.

borNfreee:
Thank you very much @sidz !

PR #58, opened 2017-10-25
Fix hardcoded framework name

Landerstraeten:


borNfreee:
While this definitely makes sense, there are several more issues with `PhpUnitPathProvider` class.

Would you mind making it a little bit prettier? :)

1. `PhpUnitPathProvider` should be renamed to something more common, like `TestFrameworkConfigPathProvider` because it is used for any test framework infection run with, not just phpunit

2. There is a `new PhpUnitPathGuesser()` call inside this class that tries to guess PHPUnit's configuration file location based on `composer.json` content (just for Symfony projects at the moment). I think it should be instantiated only when PHPUnit is used as a test framework, but not with PHPSpec

Landerstraeten:
Sure. Will do!

Landerstraeten:
I'll add some additional tests to improve the msi.

.travis.yml
```diff
@@ -3,13 +3,10 @@ language: php
 sudo: false
 
 php:
-  - 7.0
```
borNfreee: According to http://php.net/supported-versions.php, PHP 7.0 is still maintained version and will get security fixes up to December, 2018.

IMO it's too early to drop it. Currently, there are no PHP 7.1 features that we can't live without. We will just lose many users who are still on PHP7.0.x.

Please revert all the code related to PHP7.1

src/Config/ValueProvider/TestFrameworkConfigPathProvider.php
```diff
@@ -0,0 +1,108 @@
...
+        $this->consoleHelper = $consoleHelper;
+        $this->questionHelper = $questionHelper;
+    }
+
+    public function get(InputInterface $input, OutputInterface $output, array $dirsInCurrentDir, string $testFramework): ?string
+    {
+        try {
+            return $this->testFrameworkConfigLocator->locate($testFramework);
+        } catch (\Exception $e) {
+            if ($testFramework !== TestFrameworkTypes::PHPUNIT) {
```
borNfreee: just a question: does it make sense to combine this condition with the next one?

```php
if ($testFramework !== TestFrameworkTypes::PHPUNIT || !file_exists('composer.json')) {
   // ...
```

Landerstraeten: If you combine them, you need to write additional tests to kill mutants.
And I think it's more readable to just split them and fast return.

borNfreee:
Thank you @Landerstraeten!

PR #66, opened 2017-10-28
Skip `composer config bin-dir` check if custom path exists

sidz:
Small improvement for cases when user added custom phpunit path.

No needed to run `composer config bin-dir` for getting `/vendor/bin` directory as user determined the path. We just need to check is file exists and can we execute it.


borNfreee:
Nice catch, thank you

Issue #63, opened 2017-10-27
Do not mutate function signature of inherited methods


borNfreee:
There is no need to mutate `public` to `protected` for methods from the implemented interface.

```php
interface Doable
{
    public function do(): void;
}

class Foo implements Doable
{
    public function do(): void {}
}
```

Incorrect mutation:

```diff
class Foo implements Doable
{
-    public function do(): void {}
+    protected function do(): void {}
}
```

Expected result: *no mutation*, because it always leads to `Fatal error: Access level to Foo::do() must be public (as in class Doable)...`

------

At this point, we don't have information about parent classes/interfaces. It requires implementing a new smart feature with native reflection or BetterReflection.

theofidry:
I common issue with mutation testing is redundant/useless mutants. We can apply a load of dumb mutants, but the less relevant they are the more of a waste of time (as more mutants means longer run time).

That said if I were you I wouldn't focus on performance first: I can help with that and I have some plans for it, I would focus first on usability, having something easy to work with and to debug

borNfreee:
> I would focus first on usability

Do you mean that such issue like this one (removing useless mutants) should be the first priority? If so, I agree, because currently with ~940 mutations on infection itself this particular mutator creates 134 (which is 15%) useless mutants that are immediately killed because of a Fatal Error.

Feel free to open another usability issues you are aware about.

> I can help with that and I have some plans for it

Sorry, you can help with what, performance or usability? Didn't get it. Anyway, any help from you would be very useful 👍  😸

theofidry:
>Sorry, you can help with what, performance or usability?

Actually I'm a bit confusing as performance does count in usability: if a tool is too slow it's less usable right?

So I mean what should be the main focus first is getting things right: the mutants, the mutant strategy, framework integration, debuggability (worst situation: tests pass with the test framework, they don't with the mutation testing framework and impossible to know what's going on without digging horribly deep in the internals).

Getting fast enough is later when the above has been pinned down right.

>Anyway, any help from you would be very useful 👍 😸

Worry not I've meant to sent you an email with some plans for quite a while now, I just really need to get the alice packages out (as get the stable release) first to have enough free time for that :)

PR #67, opened 2017-11-06
[WIP] Add reflection classes to mutators. Do no mutate public method visibility if parent has the same one

borNfreee:
This is an attempt to add native PHP Reflection to Mutator Operators. Fixes https://github.com/infection/infection/issues/63.

- [x] `\ReflectionClass` instance for each Node inside functions/methods
- [x] Move parsed source code from test files (string variables) to separate real files to make them autoloadable.
- [x] Try [PHP-Scoper](https://github.com/humbug/php-scoper) to prefix namespaces inside `infection.phar`. This is needed because now we load the source code of a target project (creating instance of `\ReflectionClass` loads the target class). ~BLOCKED by https://github.com/humbug/php-scoper/issues/127~
- [x] ~BLOCKED by https://github.com/nikic/PHP-Parser/pull/441~
- [x] Test as a project dependency.
- [x] Test on many different projects as a PHAR.

As a result:

* Many useless mutants are not created anymore for `PublicVisibility` and `ProtectedVisibility` mutators (-130 mutants for infection itself).
* Now we can start to think/implement new mutators that was not possible to implement before without Reflection.

Please note:

* Now Infection loads project's autoload.php (see `app/bootstrap.php`) in order to instantiate `
\ReflectionClass()` instances

### Ho to prefix:

```bash
cd infection
composer install --no-dev
/path/to/php-scoper/bin/php-scoper add-prefix --output-dir=build-prefixed --force --prefix=Mutant

composer dump-autoload --working-dir=build-prefixed/ --classmap-authoritative --no-dev

sudo chmod +x build-prefixed/bin/infection
```

Feedback/review is much appreciated.

---------

**UPD:** Finally, I was able to prefix Infection lib with PHP-Scoper without errors, got the same MSI/Covered MSI, number of mutations, mutated files with Prefixed PHAR as with the original source code.


borNfreee:
> Both humbug/php-scoper#127 and nikic/PHP-Parser#441 are fixed now.

great. Will continue working on this PR next week

> but guaranteed to find weird cases with PhpSpec & PHPUnit

it shouldn't affect prefixing Infection as I understand since we don't have such dependencies in `--no-dev` mode

borNfreee:
@infection/core I think this is ready for review.

I've updated initial comment, please see details. If you have some time - could you please check it with some of your project(s)? With prefixed PHAR and as a dependency.

Not sure why builds are failing on PHP 7.2 with `phpdbg`. Didn't dig in yet.

sidz:
@borNfreee I've run Infection from this PR under phpdbg (php7.2) and got the same issue with segfault. Potentially it's an issue with php itself. It'll be nice to submit an issue (if it's an php issue) to the php bug-tracker

borNfreee:
> Potentially it's an issue with php itself

I was able to reproduce it with PHP 7.2.2 today. It fails with `phpdbg -qrr vendor/bin/phpunit` but works with `php vendor/bin/phpunit`.

Will continue git-bisecting tomorrow.

src/Visitor/FullyQualifiedClassNameVisitor.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    private $namespace;
+
+    public function enterNode(Node $node)
+    {
+        if ($node instanceof Stmt\Namespace_) {
+            $this->namespace = $node->name;
+        } elseif ($node instanceof Stmt\ClassLike) {
+            if (null !== $node->name) {
+                $node->fullyQualifiedClassName = Name::concat($this->namespace, $node->name);
```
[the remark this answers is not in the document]
borNfreee: I'm not sure I got it. This visitor's goal is to build FQCL for the current class, not for other imported (via `use` statement) classes.

When we have:

```php

namespace Infection\Mutator;

use SomeClass;
use Some\OtherClass as Aliased;

class Plus {
}
```

this visitor should build a `Infection\Mutator\Plus` FQCN and save it to the node to `fullyQualifiedClassName` property.

It does not need to work with `SomeClass`, `OtherClass`, `Aliased`.

Then we use it to build reflection in ReflectionVisitor:

```php
if ($node instanceof Node\Stmt\ClassLike) {
       $this->reflectionClass = new \ReflectionClass($node->fullyQualifiedClassName->toString());
}
```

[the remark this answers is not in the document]
theofidry: Oh right, I had my head too much in php-scoper. Should be alright then :)

BackEndTea:
> Now Infection loads project's autoload.php (see app/bootstrap.php) in order to instantiate \ReflectionClass() instances

I don't know how relevant this is, but maybe we can allow users to define the location of the autoload file in the infection config. Similar to how phpunit does things.

borNfreee:
Finally, rebased on master, ready for review again.

Regarding PHPScoper config, I think we will use Theo's one from https://github.com/humbug/php-scoper/pull/172/files later (but here PHPParser has been updated, so not all patchers are needed)

> I don't know how relevant this is, but maybe we can allow users to define the location of the autoload file in the infection config. Similar to how phpunit does things.

what does it bring for us?

src/Visitor/ReflectionVisitor.php
```diff
@@ -12,21 +12,32 @@
...
+
     public function beforeTraverse(array $nodes)
     {
         $this->scopeStack = [];
+        $this->reflectionClass = null;
     }
 
     public function enterNode(Node $node)
     {
+        if ($node instanceof Node\Stmt\ClassLike) {
```
BackEndTea: What happens here if we declare an anonymous class within our class?
Will all the nodes after that still be seen as instances of that anonymous class?

BackEndTea: It crashes on anonymous classes in my test cases.
This could be fixed by

```php
    {
        if ($node instanceof Stmt\Namespace_) {
            $this->namespace = $node->name;
        } elseif ($node instanceof Stmt\ClassLike) {
            if (null !== $node->name) {
                $node->fullyQualifiedClassName = Name::concat($this->namespace, $node->name);
            } else {
                $node->fullyQualifiedClassName = null;
            }
        }
    }
```
For the `FullyQualifiedClassNameVisitor`

and 
```php
    
    {
        if ($node instanceof Node\Stmt\ClassLike && $node->fullyQualifiedClassName !== null) {
            $this->reflectionClass = new \ReflectionClass($node->fullyQualifiedClassName->toString());
        }
```

in the `ReflectionVisitor`

borNfreee: Thank you @BackEndTea, I've added a test case.

BackEndTea:
>what does it bring for us?

I realize now that it not that usefull for us. Phpunits version is for bootstrapping an application, not just autoloading

BackEndTea:
Looks good to me. Id just like to see a test for the reflection visitor that checks if the reflection class is correctly set.

src/Mutator/FunctionSignature/ProtectedVisibility.php
```diff
@@ -49,6 +50,33 @@ public function shouldMutate(Node $node): bool
             return false;
         }
 
+        if ($this->hasSameProtectedParentMethod($node)) {
```
BackEndTea: Just a small thing, but maybe its slightly faster to first check if the node is protected? before checking if it has a protected parent.

And ofcourse the same goes for public

borNfreee:
Ok, I think it's ready to be merged (this PR also unlocks Profiles implementation).

Thanks everyone for the review and help.

PR #68, opened 2017-11-09
Show correct type of error message for CI flags

dmecke:
Hi there!

First of all, thank you very much for the great tool!

When using both CI flags, --min-msi and --min-covered-msi together and only min-covered-msi fails, it shows a message that the minimum required min-msi percentage is not reached which is wrong.

This pull request checks boths values individually and shows the corresponding error message.

Please tell me if I need to improve something in this pull request.

### DIFF
--- a/src/InfectionApplication.php
+++ b/src/InfectionApplication.php
@@ -114,6 +114,12 @@ class InfectionApplication
                 return 1;
             }
 
+            if ($this->hasBadCoveredMsi($metricsCalculator)) {
+                $io->error($this->getBadCoveredMsiErrorMessage($metricsCalculator));
+
+                return 1;
+            }
+
             return 0;
         } catch (InfectionException $e) {
             $io->error($e->getMessage());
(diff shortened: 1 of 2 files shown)


src/InfectionApplication.php
```diff
@@ -248,6 +259,13 @@ private function getBadMsiErrorMessage(MetricsCalculator $metricsCalculator): st
             );
         }
 
+        throw new \LogicException('Seems like something is wrong with calculations and min-msi options.');
+    }
+
+    private function getBadCoveredMsiErrorMessage(MetricsCalculator $metricsCalculator): string
+    {
+        $baseMessage = 'The minimum required %s percentage should be %s%%, but actual is %s%%. Improve your tests!';
```
borNfreee: Could you please reduce code duplication a little bit:

1. This message probably can be moved to const and reused on line 251
2. and there is a duplicated exception message `throw new \LogicException('Seems like something is wrong with calculations and min-msi options.');`. I suggest to create a custom one with incapsulated message or reuse this message somehow in this class

borNfreee:
Hello, thank you for using this tool and for this PR. Nice catch!

Left a small comment about duplicated strings, besides that I'm happy to merge it.

src/InfectionApplication.php
dmecke: I've now extracted the `$baseMessage` into a constant and created a new `MsiCalculationException`.

borNfreee:
Thank you @dmecke !

PR #80, opened 2017-11-18
Console logger output format to be compatible with TextFile logger format

sidz:
When we did changes for text logger output format seems like we forgot to change console output format as well to be compatible with new structure:
- Filename (preferably the relative path)
- Line number
- Mutation type
- Mutation

This PR fixes it. So we will have the same format for both Console and Text File loggers.

```
1) /var/www/infection/src/Filesystem/Filesystem.php:60    [M] FalseValue

--- Original
+++ New
@@ @@
         }
-        if (false === @file_put_contents($filename, $content)) {
+        if (true === @file_put_contents($filename, $content)) {
             throw new IOException(sprintf('Failed to write file "%s".', $filename), 0, null, $filename);
         }
     }
 }

```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ install:
 script:
   - vendor/bin/phpunit --coverage-clover=clover.xml
   - composer analyze
-  - if [[ $INFECTION == 1 ]]; then bin/infection --threads=4 --min-msi=60 --min-covered-msi=91; fi
+  - if [[ $INFECTION == 1 ]]; then bin/infection --threads=4 --min-msi=61 --min-covered-msi=91; fi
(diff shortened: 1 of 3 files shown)


borNfreee:
Nice. Could you please make Travis happy? ;)

sidz:
@borNfreee done, but I had to set 89% in Covered Code MSI. Please let me know if I should back 91 and improve anything

borNfreee:
Thank you @sidz! Seems like not all mutants are killed for `MutationTestingConsoleLoggerSubscriber`, that's why MSI has been decreased, but I don't think this file worths adding more tests.

PR #81, opened 2017-11-18
Improve the order script execution for travis

sidz:
This PR is changing script execution order for travis as I think here is no a lot of sense to run phpunit tests if code style has any issues.


borNfreee:
Thank you @sidz

PR #84, opened 2017-11-22
Add comma separated files filter

Landerstraeten:
It's currently not possible to have a comma separated file list with relative paths as a _filter_ argument.

For example: `$ bin/infection --filter=src/AppBundle/Entity/A.php,src/AppBundle/Service/B.php,src/AppBundle/Service/C.php`.

This PR adds this feature.

### DIFF
--- a/src/Finder/SourceFilesFinder.php
+++ b/src/Finder/SourceFilesFinder.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Copyright © 2017 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+declare(strict_types=1);
+
+namespace Infection\Finder;
+
+use Symfony\Component\Finder\Finder;
+
+class SourceFilesFinder
+{
+    private $sourceDirectories;
+    private $excludeDirectories;
+
+    /**
+     * @param string[] $sourceDirectories
+     * @param string[] $excludeDirectories
+     */
+    public function __construct(array $sourceDirectories, array $excludeDirectories)
+    {
+        $this->sourceDirectories = $sourceDirectories;
+        $this->excludeDirectories = $excludeDirectories;
+    }
+
+    public function getSourceFiles(string $filter = ''): Finder
+    {
+        $finder = new Finder();
+
+        array_walk($this->excludeDirectories, function ($excludePath) use ($finder) {
+            $finder->notPath($excludePath);
+        });
+
+        if ('' === $filter) {
+            $finder->in($this->sourceDirectories)->files();
+
+            return $finder;
+        }
+
+        $finder->in('.')->files();
+
+        $filters = explode(',', $filter);
+        array_walk($filters, function ($fileFilter) use ($finder) {
+            $finder->path($fileFilter);
+        });
+
+        return $finder;
+    }
+}
--- a/tests/Finder/SourceFilesFinderTest.php
+++ b/tests/Finder/SourceFilesFinderTest.php
@@ -1,0 +1,108 @@
+<?php
+/**
+ * Copyright © 2017 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Tests\Finder;
+
+use Infection\Finder\SourceFilesFinder;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Finder\Finder;
+
+class SourceFilesFinderTest extends TestCase
+{
+    public function test_it_lists_all_files_without_a_filter()
+    {
+        $sourceFilesFinder = new SourceFilesFinder(['tests/Files/Finder'], []);
+
+        $files = $sourceFilesFinder->getSourceFiles();
+
+        $this->assertInstanceOf(Finder::class, $files);
+        $this->assertSame(2, $files->count());
+    }
+
+    public function test_it_can_filter_one_file_by_a_relative_path()
+    {
+        $sourceFilesFinder = new SourceFilesFinder(['tests/Files/Finder'], []);
+
+        $filter = 'tests/Files/Finder/FirstFile.php';
+        $files = $sourceFilesFinder->getSourceFiles($filter);
+
+        $iterator = $files->getIterator();
+        $iterator->rewind();
+        $firstFile = $iterator->current();
+
+        $this->assertInstanceOf(Finder::class, $files);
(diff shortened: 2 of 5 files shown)


borNfreee:
Hello,

1. could you please add some tests for `--filter` option to make sure everything works as expected?

2. could you also please create a PR for documentation https://github.com/infection/site/blob/master/src/guide/command-line-options.md?

borNfreee:
@Landerstraeten seems like now it works the opposite way than in `master`.

On `master`, when I run `infection --filter=Plus.php`, only for this one file mutations are generated:

```bash
Generate mutants...

Processing source code files: 1/1
Creating mutated files and processes: 8/8

SS..M..E                                             (8 / 8)
```

when I run the same command on this branch, all files are processed, except `Plus.php`.

The second BC break is that now we need to path the whole relative path: `Mutator/Arithmetic/Plus.php` instead of just `Plus.php`.

-------

I think instead of this new code`$finder->notName($fileFilter);` there should be `$finder->name($fileFilter);` as it was before.

Could you please address these 2 issues or explain why it works the opposite?
Currently, on `master` branch, it works the same way as for example PHPUnit's `--filter` option works - it runs the test only for provided file/class/action, not excludes them.

Landerstraeten:
Hi @borNfreee,

I've updated the test cases. This should fix my mistake and make sure there are no BC breaks.

src/Finder/SourceFilesFinder.php
```diff
@@ -0,0 +1,50 @@
...
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+declare(strict_types=1);
+
+namespace Infection\Finder;
+
+use Symfony\Component\Finder\Finder;
+
+/**
+ * @author Lander Vanderstraeten <lander_vanderstraeten@hotmail.com>
```
theofidry: @borNfreee do we really want to bother with the `@author` tags? In my experience the commit history is more than enough for this, those tags are just noisy and annoying to manage

src/Finder/SourceFilesFinder.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+class SourceFilesFinder
```
theofidry: I think this class can be made final as well

src/Finder/SourceFilesFinder.php
```diff
@@ -0,0 +1,50 @@
...
+{
+    private $sourceDirectories;
+    private $excludeDirectories;
+
+    public function __construct(array $sourceDirectories, array $excludeDirectories)
```
theofidry: I would add a phpdoc bloc to add the `string[]` type, otherwise it might not clear what we are dealing with

src/Finder/SourceFilesFinder.php
```diff
@@ -0,0 +1,50 @@
...
+
+    public function getSourceFiles(string $filter = ''): Finder
+    {
+        $finder = new Finder();
+
+        array_walk($this->excludeDirectories, function ($excludePath) use ($finder) {
+            $finder->notPath($excludePath);
+        });
+
+        if (empty($filter)) {
```
theofidry: `'' !== $filter`, I recommend against using `empty()` whenever possible, it's a footgun function (`'0' _is_ empty)

tests/Finder/SourceFilesFinderTest.php
```diff
@@ -0,0 +1,85 @@
...
+class SourceFilesFinderTest extends TestCase
+{
+    public function test_it_lists_all_files_without_a_filter()
+    {
+        $sourceFilesFinder = new SourceFilesFinder(['tests/Files/Finder'], ['vendor']);
+
+        $files = $sourceFilesFinder->getSourceFiles();
+
+        $this->assertInstanceOf(Finder::class, $files);
+        $this->assertEquals(2, $files->count());
```
theofidry: `$this->assertSame()`, it's safer in general and avoid surprises with unexpected casting  (same elsewhere)

theofidry: Should we check here what are those two files? (same elsewhere)

tests/Finder/SourceFilesFinderTest.php
```diff
@@ -0,0 +1,85 @@
...
+use Symfony\Component\Finder\Finder;
+
+/**
+ * @author Lander Vanderstraeten <lander_vanderstraeten@hotmail.com>
+ */
+class SourceFilesFinderTest extends TestCase
+{
+    public function test_it_lists_all_files_without_a_filter()
+    {
+        $sourceFilesFinder = new SourceFilesFinder(['tests/Files/Finder'], ['vendor']);
```
theofidry: AFAICS there is no `vendor` directory in `tests/Files/Finder` so this could be replaced by an empty array (same elsewhere)

src/Finder/SourceFilesFinder.php
```diff
@@ -0,0 +1,51 @@
...
+            $finder->notPath($excludePath);
+        });
+
+        if ('' === $filter) {
+            $finder->in($this->sourceDirectories)->files();
+
+            return $finder;
+        }
+
+        $finder->in('.')->files();
```
borNfreee: This is the last thing I would like to discuss (and probably update in PR). In master, we start to look from `sourceDirectories`, not from `.`. There are many libs where real sources are located much deeper than in the root. Example: [PHP-Parser](https://github.com/nikic/PHP-Parser/tree/master/lib/PhpParser), sources are in `./lib/PHP-Parser`. So relative path in master branch would start from `./lib/PHP-Parser`, in this PR it starts from `.`.

I made such changes to make it like in master.

```php
public function getSourceFiles(string $filter = ''): Finder
    {
        $finder = new Finder();

        array_walk($this->excludeDirectories, function ($excludePath) use ($finder) {
            $finder->notPath($excludePath);
        });

        $finder->in($this->sourceDirectories)->files();

        $filters = explode(',', $filter);
        array_walk($filters, function ($fileFilter) use ($finder) {
            $finder->path($fileFilter);
        });

        return $finder;
    }
```

Does it make sense? :) Do you think relative path should start from `.` instead of what is configured as a `src` in `infection.json.dist`?

theofidry:
If there is relative path I would assume either from where the command is
executed or the configuration file. In this peculiar case, for relative
paths inside the config I would say relative to the config file, but for
paths passed via the command then relative to the current working directory.

Also something I saw in box is adding a base-path option (relative to the
config file) to change the base of the relative paths used in the config
file.

borNfreee:
ok, then it makes sense to start it from `.`.

Thank you @Landerstraeten !

PR #93, opened 2017-12-02
Fix issue when custom path for test framework exists

sidz:
Quick fix for case when `customPath` exists in the config file.

Issue added by my previous PR: https://github.com/infection/infection/pull/85


borNfreee:
Thank you @sidz

PR #98, opened 2017-12-26
PHPSPEC -  If you had a custom bootstrap file in the phpspec.yml it would generate the autoload without the semicolon.

AliceIW:
Fixed:
  - src/TestFramework/PhpSpec/Config/Builder/MutationConfigBuilder.php

Cause:
  - If you had a custom bootstrap file in the phpspec.yml it would generate the autoload without the semicolon.


borNfreee:
Thank you @AliceIW

PR #102, opened 2017-12-31
Add credits

theofidry:
Add credits for Padraic.

I also updated Humbug to mention the library is no longer maintained as well as abandoned the package on Packagist with Infection as a replacement.

Hope hope that will pull out a few Humbug users here. It could be good to provide a migration path as well which I believe should be relatively easy. Maybe even better, IIRC the base config of Humbug  is working fine so we could support the humbug config file and just provide a deprecation notice recommending to rename the file. That way one wouldn't have to change anything but the package as a first step, can't get a smoother migration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,3 +13,7 @@ Infection - Mutation Testing framework
 Please read documentation here: http://infection.github.io
 
 Twitter: http://twitter.com/infection_php
+
+### Credits
+
+This project is originally a fork of Pádraic Brady ([@padraic](https://github.com/padraic))'s [Humbug library](https://github.com/humbug/humbug). Humbug has since then been discountinued in favour of this project.
--- a/README.md
+++ b/README.md
@@ -16,4 +16,4 @@ Twitter: http://twitter.com/infection_php
 
 ### Credits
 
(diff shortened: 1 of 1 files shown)


borNfreee:
👍 

For those who are not familiar who is Padraic, he is 

* Author of [Mockery](https://github.com/mockery/mockery) (we use it in Infection btw)
* Author of Mutation Testing Framework - Humbug (Infection was inspired by it)
* a man who did a lot for other projects related to PHAR, security and testing

Here is a link to his blog (currently available in google archive, but highly recommended)  that influenced to Infection development a lot

* [Securely Distributing PHARs: Pitfalls and Solutions](http://archive.is/hDwaA#selection-99.0-99.51)

I'm totally ok with this credits section, as we discussed earlier.

Just one thing - infection  technically is *not a fork* of Humbug, it's a package that was highly inspired by Humbug. Probably, it makes sense to reword it to avoid confusion. Besides that, I'm happy to merge it.

theofidry:
And also the man who started phpspec, great sum up :)

borNfreee:
Awesome

Issue #106, opened 2018-01-09
Infection not working with phpdbg


jyggen:
| Question    | Answer
| ------------| ---------------
| Infection version | 47c8350d11ea100184ac20fcd8eb85be2c78194a and later.
| Test Framework version | PHPUnit 6.1.0 and PHPUnit 6.5.5
| PHP version | 7.0.22 and 7.2.1
| Platform    | MacOS

It seems like running Infection using phpdbg (0.5.0) is broken in 0.7.0 and later.

I first tested it on an internal project using PHP 7.2.1, PHPUnit 6.5.5 and Infection 0.7.0 without success (and a colleague verified it with PHP 7.0.22), and then on the Infection codebase itself (69551ba366ca116e56178a8f3c10416e75e1b91f) using PHP 7.2.1 and PHPUnit 6.1.0. In both cases we get the same error message.

I've checked out 0.6.2 of Infection and that version does work. Running Infection on Infection commit by commit since 0.6.2 it seems like 47c8350d11ea100184ac20fcd8eb85be2c78194a is the version that breaks phpdbg, because that commit and every commit after that has the same issue.

I'll do some more digging later and maybe I can put together a failing test case for it.

<details>
 <summary>Output while running Infection on Infection itself:</summary>
 
 ```shell
$ phpdbg -qrr bin/infection
    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/

Running initial test suite...

Phpunit version: 6.1.0

    2 [============================] < 1 sec

Generate mutants...

Processing source code files: 166/166
Creating mutated files and processes:    0/1008

 [ERROR] Code Coverage does not exist. File
         /var/folders/1_/j6pgbhz546g4t92sckw7215c0000gn/T/infection/coverage-xml/index.xml is not found. Check phpunit
         version Infection was run with and generated config files inside
         /var/folders/1_/j6pgbhz546g4t92sckw7215c0000gn/T/infection.
 ```
</details>

sidz:
@jyggen thank you. I've fixed that stupid issue here: https://github.com/infection/infection/pull/107

Issue #75, opened 2017-11-17
[epic] Performance optimization ideas


borNfreee:
Nikic [has recently published](https://github.com/nikic/PHP-Parser/blob/73be076/doc/component/Performance.markdown) a small article about possible performance optimizations for projects that use PHP-Parser.

I would like to post them here, discuss and step by step try all of them for Infection.

- [x] [Object reuse](https://github.com/nikic/PHP-Parser/blob/73be076/doc/component/Performance.markdown#object-reuse) - [implemented](https://github.com/infection/infection/pull/76)
- [x] [Garbage Collection](https://github.com/nikic/PHP-Parser/blob/73be076/doc/component/Performance.markdown#garbage-collection) - [didn't give](https://github.com/infection/infection/issues/75#issuecomment-346150671) any performance boost. 
- [x] Reuse created mutants to avoid traversing/prettyprinting - [implemented](https://github.com/infection/infection/pull/184)
- [x]  Partially [disabling XDebug](https://github.com/nikic/PHP-Parser/blob/73be076/doc/component/Performance.markdown#xdebug) - [implemented](https://github.com/infection/infection/pull/85). We can disable xDebug for each Symfony Process when e.g. PHPUnit is run against mutated code. Actually, we can disable it even for the main Infection process, and only enable when the initial test suite is executed to generate Code Coverage (this is also a separate Symfony Process in the code).
- [ ] Partially disable `phpdbg`

sidz:
@borNfreee I can pick up "disabling XDebug". If you already have any thoughts feel free to write to me.

borNfreee:
@sidz great, thank you!

theofidry:
tbh I wouldn't touch the garbage collection: it's trivial for the user to disable it if he can afford it by doing `php -d zend.gc_enable=0 vendor/bin/infection`

borNfreee:
I've tested disabling Garbage Collector with different (not big though) projects and for all of them it does not give any performance boost. Sometimes it's even worse with `gc_disable()` - more memory consumption and more execution time.

Marking this point as "won't implement" for now.

sidz:
@borNfreee here is my PR for "Partially disabling XDebug" enhancement: https://github.com/infection/infection/pull/85

suggestions are welcome

borNfreee:
Great! I've also added a point about `phpdbg` to the initial comment to track it

borNfreee:
[...]

PR #107, opened 2018-01-09
Fix phpdbg issue caused by xdebug disabling feature

sidz:
This PR fixes https://github.com/infection/infection/issues/106 issue which was caused by switching `$includeArgs` variable from `true` to `false`. But I totally forgot that to run `phpdbg` we need to add  `-qrr` argument.

And that is why we have an issue when this parameter is equal to `false` then `-qrr` argument will be ignored.
proof: https://github.com/symfony/process/blob/3.4/PhpExecutableFinder.php#L39

p.s. I didn't find more elegant case how to avoid it. Any suggestions are welcome.

### DIFF
--- a/src/Process/Builder/ProcessBuilder.php
+++ b/src/Process/Builder/ProcessBuilder.php
@@ -41,11 +41,13 @@ final class ProcessBuilder
      */
     public function getProcessForInitialTestRun(string $testFrameworkExtraOptions = ''): Process
     {
+        $includeArgs = PHP_SAPI === 'phpdbg';
+
         return new Process(
             $this->testFrameworkAdapter->getExecutableCommandLine(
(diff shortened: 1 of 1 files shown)


theofidry:
Not sure about the specific solution but I don't have much better to suggest for now to be honest.

That said maybe we should make a point from it and have an e2e test with phpdbg and xdebug disabled?

theofidry:
I also think we should have a similar case with the PHAR but I'll work on that as soon as I got box ready for infection :)

sidz:
@theofidry I'd say that we can add tests for xdebug and phpdbg when we will implement the final goal from https://github.com/infection/infection/issues/75 epic

theofidry:
Fair enough 👍

PR #109, opened 2018-01-10
Allow install of 1.0 dev

theofidry:
This allow one to use `composer require --dev infection/infection:^1.0@dev`


borNfreee:
Could you please expand what advantages gives this change? Or point to some doc ;)

theofidry:
Actually forget about it, I though just giving `infection/infection` without the version was failing due to the version not being stable.

Issue #108, opened 2018-01-09
Infection not working if using custom printer


BackEndTea:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.7.0
| Test Framework version | PHPUnit 6.5
| PHP version | 7.1.12
| Platform    | MacOS
| Github Repo | https://github.com/opencfp/opencfp

Running `vendor/bin/infection` crashes if one uses a custom printer.
In the linked repo we have `printerClass="Codedungeon\PHPUnitPrettyResultPrinter\Printer"` in our `phpunit.xml.dist`
The current workaround is to add `--test-framework-options="--printer PHPUnit\\\TextUI\\\ResultPrinter"`. But it would be nice if infection could detect that on its own

theofidry:
An alternative is to have a dedicated phpunit config file for Infection.

@borNfreee I didn't check but maybe we could have a minor perf gain by registering a NullPrinter? We don't use the printer result do we?

Another thing to keep in mind is env variables as well as listeners; I'm not sure what's the status for them yet

borNfreee:
I think the fix is really simple - we just need to remove [`printerClass` attribute](https://github.com/opencfp/opencfp/blob/master/phpunit.xml.dist#L15) from the custom XML files Infection generates in `tmp` folder. There 2 types of files:

1. For initial test suite run (to make sure project's all tests are in a passing state)
2. For each mutant (this is where we create PHPUnit config [with particular tests to run](https://github.com/infection/infection/blob/master/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php#L64))

These are 2 places where we need to remove this attribute, allowing PHPUnit to  fallback to `PHPUnit\TextUI\ResultPrinter` default printer.

Would you mind creating a PR for that? If no, I will be able to pick it up at the end of the week.

* https://github.com/infection/infection/blob/master/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php#L64 
* https://github.com/infection/infection/blob/master/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php#L52

----------

> We don't use the printer result do we?

Actually, we do. We then parse printer's output to understand whether tests failed or not (see how it works here https://github.com/infection/infection/blob/master/tests/TestFramework/PhpUnit/Adapter/PhpUnitAdapterTest.php#L41). Do you have any recommendations here?

> Another thing to keep in mind is env variables as well as listeners; I'm not sure what's the status for them yet

[...]

PR #110, opened 2018-01-10
Fix: remove printer attributes from phpunit

BackEndTea:
See #108


borNfreee:
Thank you @BackEndTea

BackEndTea:
Thanks @borNfreee

Issue #97, opened 2017-12-25
Mutations not working well with function_exists


BackEndTea:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.7.0
| Test Framework version | PHPUnit/ 6.5.-
| PHP version | 7.1.12
| Platform    | MacOS
| Github Repo | https://github.com/BackEndTea/Array-Meta

<details>
 <summary>Output with issue</summary>
Given the following source code

```php
<?php

declare(strict_types=1);

if (!\function_exists('isValidArrayKey')) {
    function isValidArrayKey($key): bool
    {
        return \is_string($key) || \is_int($key);
    }
}
```

I expected these tests to cover everything
```php
<?php

declare(strict_types=1);

namespace BackEndTea\ArrayMeta\Test\Unit;

use BackEndTea\ArrayMeta\Test\Fixture\KeyExceptionFixture;

class HelpersTest extends \PHPUnit\Framework\TestCase
{
    public function testIsValidArrayKeyReturnsTrueWhenSuppliedWithInteger()
    {
        $this->assertTrue(isValidArrayKey(3));
    }

    public function testIsValidArrayKeyReturnsTrueWhenSuppliedWithString()
    {
        $this->assertTrue(isValidArrayKey('key'));
    }

    /**
     * @param $key
     *
     * @dataProvider provideInvalidKeysForIsValidArrayTest
     */
    public function testIsValidArrayKeyReturnsFalseWhenSuppliedWithInvalidKey($key)
    {
        $this->assertFalse(isValidArrayKey($key));
    }

    public function provideInvalidKeysForIsValidArrayTest(): array
    {
        return [
            [new \stdClass()],
            [2.5],
            [new KeyExceptionFixture()],
            [array()],
            [null],
        ];
    }
}
```

However, it tells me the following mutation does not get covered
```diff
1) /Users/Ibuildings/Projects/ArrayWrap/src/helpers.php:8    [M] LogicalOr
exec /usr/local/Cellar/php71/7.1.12_23/bin/php -c /private/var/folders/5y/g0yc55n12hjcc4zb0k3z2x_m0000gn/T/infectionY1rj3s /Users/Ibuildings/Projects/ArrayWrap/vendor/phpunit/phpunit/phpunit --configuration /var/folders/5y/g0yc55n12hjcc4zb0k3z2x_m0000gn/T/infection/phpunitConfiguration.9263c2c2e35e7efa3a1cebf954bdba92.infection.xml --stop-on-failure 
[...]

PR #111, opened 2018-01-10
Filter mutations to only cover php classes

BackEndTea:
This filters out files that don't have 'class ' in them.

Related to: #97 

I'm not 100% sure this is the best way to handle this, so i'm open for suggestions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,10 @@
         "psr-4": {
             "Infection\\Tests\\": "tests/"
         },
-        "files": ["tests/Helpers.php"]
+        "files": [
+            "tests/Helpers.php",
+            "tests/Files/Finder/test_functions.php"
+        ]
     },
     "config": {
         "platform": {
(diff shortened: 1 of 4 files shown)


theofidry:
There is mutants that could cover interfaces, traits, or you could also have more exotic files which don't have any class or dynamic classes as well.

So while I get the idea I just don't think that's the way to do it :) We should limit it to not mutating the content of functions IMO.

BackEndTea:
You're right, i forgot about those cases.

It looks like it might be beneficial to wait for #67, since reflection should make it a lot more easy to check if we are dealing with a function, rather than a method

borNfreee:
Depends on what you are trying to achieve. Currently, it's possible to understand it without reflection, see here

https://github.com/infection/infection/blob/f1ebd5b28744bf68a6eb4d5280e30748e55d1596/src/Visitor/WrappedFunctionInfoCollectorVisitor.php#L74-L82

Removing `$isFunction` prevents Infection from mutating functions (but still mutates callbacks and methods).

PR #112, opened 2018-01-10
Update gitattributes

BackEndTea:
This causes these files to not be exported when tagging a release on github.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,6 @@
 # Denote all files that are truly binary and should not be modified.
 *.png binary
 *.jpg binary
+
+#Files we do not want to distribute
+/tests export-ignore


theofidry:
Do we really want that? It's merely 500KB and in exchange we need to maintain a file which we have no way to test hence easy to forget to update

BackEndTea:
The bigest one is really  the `/tests`. Since those tend to contain fixtures for testing purposes, and it can cause your IDE to start 'hinting' to those classes as well.

theofidry:
Then we can just not export the tests only instead of keeping a curated list of the unnecessary files. But I was referring to tests in terms of size:

```
512K	/path/to/infection/tests
```

So I doubt it's more than that for all the files included

borNfreee:
> The bigest one is really the /tests. Since those tend to contain fixtures for testing purposes, and it can cause your IDE to start 'hinting' to those classes as well.

Makes sense. On the other side, I agree there is no need to exclude every single [config] file.

I would vote for ignoring just `tests` atm.

BackEndTea:
I have updated it to only include `/tests`

borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee

PR #113, opened 2018-01-10
Fix: mutate methods but not functions

BackEndTea:
See #97 and #111

### DIFF
--- a/tests/Mutator/ReturnValue/AbstractValueToNullReturnValueTest.php
+++ b/tests/Mutator/ReturnValue/AbstractValueToNullReturnValueTest.php
@@ -286,6 +286,5 @@ CODE;
 
         $mutatedCode = $this->mutate($code);
         $this->assertSame($code, $mutatedCode);
-
     }
 }
--- a/tests/Mutator/ReturnValue/AbstractValueToNullReturnValueTest.php
+++ b/tests/Mutator/ReturnValue/AbstractValueToNullReturnValueTest.php
@@ -20,9 +20,12 @@ abstract class AbstractValueToNullReturnValueTest extends AbstractMutator
         $code = <<<"CODE"
 <?php
 
-function test()
+class Test
 {
-    return {$this->getMutableNodeString()};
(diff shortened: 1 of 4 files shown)


tests/Mutator/ReturnValue/AbstractValueToNullReturnValueTest.php
```diff
@@ -137,26 +171,33 @@ public function test_it_does_not_mutate_when_function_contains_another_function_
     {
```
theofidry: Actually I'm not sure about this. Could we double check that if you have:

```php
class Test
{
    function __invoke($x): void {
        $foo = function ($y): bool {
            return true;
        }

        return $foo($x);
    }
}
```

Then _maybe_ the mutants changing the `return true` to `return false` does work. I think the issue with function is only named functions not anonymous functions like that. If that's the case then even the following should work:

```php
$x = function foo() {
    return true; 
}

// As opposed of:

function foo() {
    // the content of this function (namespaced or not) cannot not be mutated
}
```

borNfreee: Actually, this is not a valid syntax. Anonymous function can't have a name.
```php
$x = function foo() {
    return true; 
}
```

I checked your example in this branch and there is a correct mutant:

```diff
class Test
{
    function __invoke($x): void {
        $foo = function ($y): bool {
-            return true;
+            return false;
        }

        return $foo($x);
    }
}
```

borNfreee:
@theofidry do you still have any concerns about this PR (please see https://github.com/infection/infection/pull/113#discussion_r161058354).

If not, I think it can be merged.

borNfreee:
Could you please rebase before merging?

BackEndTea:
Rebased

borNfreee:
Thank you @BackEndTea

theofidry:
@borNfreee back from holidays :) No it looks ok, my concern was that anonymous functions like those should still be mutated

PR #116, opened 2018-01-11
fix small typos

teiling88:
During my work on a blog post about infection I found some small typos and fixed them.


borNfreee:
Thank you!

Could you please share a blog post when it's ready so we can retweet it in https://twitter.com/infection_php?

teiling88:
Sure, I will do that! :-)

Thank you for the quick response!

PR #118, opened 2018-01-12
Move Files folder into the Fixtures folder

BackEndTea:
Currently both folders serve the same purpose


borNfreee:
Makes sense, thank you @BackEndTea

Issue #103, opened 2018-01-01
Make Humbug's config compatible with Infection


borNfreee:
See #102 

Make Humbug's config compatible with Infection to transition users without any efforts (probably just by renaming the config)

sidz:
@theofidry, @borNfreee  As I can see humbug's full config is:
```
{
    "chdir": "tests",
    "timeout": 10,
    "source": {
        "directories": [
            "."
        ],
        "excludes": [
            "vendor",
            "Tests"
        ]
    },
    "logs": {
        "text": "humbuglog.txt",
        "json": "humbuglog.json"
    }
}
```
and infection full config looks like this:
```
{
    "source": {
        "directories": [
            "src"
        ],
        "exclude": [
            "Config",
            "Folder/with/File.php"
        ]
    },
    "timeout": 10,
    "logs": {
        "text": "infection-log.txt"
    },
    "phpUnit": {
        "configDir": "app",
        "customPath": "\/path\/to\/phpunit-6.1.phar"
    }
}
```
So I'm not sure that do we need to take care about `chdir` and `logs.json` options?  Does anybody use `logs.json` @theofidry ?

and there is a minor difference between `source.exclude` option in infection as in humbug's config this parameter calls `excludes`.

theofidry:
I never used it and I doubt it's used in the wild so maybe we could just ignore it. I think we could add a small bridge support with deprecation notices for a smooth transition.

ryan7273:
I used it, but my use case is no longer valid.

I was integrating with a CI environment and I had a script which parsed the JSON results to determine if it was okay to proceed. The --min-msi and --min-covered-msi switches are a vast improvement over this method.

PR #120, opened 2018-01-14
Make Humbug's config to be compatible with Infection

sidz:
As were discussed in the https://github.com/infection/infection/issues/103 I've added `excludes` option only. And I'd say that in the 0.8 we can remove `exclude` from the config file as there is no a lot of sense to support both of them

@borNfreee we can update documentation as well.


borNfreee:
Thank you @sidz !

theofidry:
👌

Issue #101, opened 2017-12-29
Default values of functions not being found by coverage


BackEndTea:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.7.0
| Test Framework version | PHPUnit/ 6.5.-
| PHP version | 7.1.12
| Platform    | MacOS
| Github Repo | https://github.com/BackEndTea/Array-Meta

<details>
 <summary>Output with issue</summary>
 Given the following code

```php
    public function search($value, bool $strict = false)
    {
        if (($return =  \array_search($value, $this->items, $strict)) === false) {
            throw ValueNotFoundException::valueNotFound($value);
        }
        return $return;
    }

```

And the following test

```php
    public function testSearchDefaultsToNonStrictSearch()
    {
        $array = ['0', '1', '2', '3'];
        $meta = new ArrayMeta($array);
        $key = $meta->search(3);
        $this->assertSame(3, $key);
    }
```

I expected the mutation of $strict to true to be covered. 

However it tells me the mutation was not covered, as the code coverage doesn't look at the function signature, but only to the body of the function. So mutating default values always says it was not covered

</details>

borNfreee:
Thank you for opening this issue.

Actually, this is how PHPUnit's Coverage works:

![coverage-bug](https://user-images.githubusercontent.com/3725595/34920123-99f151e4-f97e-11e7-8bc1-db33236073c8.png)

Notice that coverage does not have information about function signature line (it is white, not green), even if the function was executed. To leverage this issue and to make e.g. Public/Protected Visibility mutators work, we are searching any executed line inside the called function, and if they are found, we treat the line with function signature _as covered_.

<Details>

https://github.com/infection/infection/blob/0f06fb474d30cc978419b66ba34ec22d70684934/src/TestFramework/Coverage/CodeCoverageData.php#L87-L106

</Details>

Currently, we have two types of mutators:

* Function body mutator (almost all mutators)
* Function signature mutator (Public/Protected visibility mutators)

It looks like function body mutator in your repository that changes `true` to `false` also changes values in the method signature. And from the first look, we can just not allow mutating code in the function signature for this type of mutators. But, your example shows that this is a pretty valid case and we should support these mutations.

[...]

PR #123, opened 2018-01-15
Fix: Change is function signature check

BackEndTea:
This PR:

* [x] Removes `isFunctionBodyMutator` & `isFunctionSignatureMutator`
* [x] Removes the `FunctionBodyMutator` class & the `FunctionSignatureMutator` class
* [x] Uses the AST to determine if a node is on the function signature

This change causes the MSI to go up, but the Covered Code MSI to go down.

Related to: #101


src/TestFramework/Coverage/CodeCoverageData.php
```diff
@@ -199,6 +193,18 @@ private function addTestExecutionInfo(array $coverage): array
         return $newCoverage;
     }
 
+    public function isLineFunctionSignature(string $filePath, int $line): bool
```
BackEndTea: Since this gets called multiple times it might be beneficial to 'cache' its results somehow.

src/Visitor/MutationsCollectorVisitor.php
```diff
@@ -58,19 +58,20 @@ public function __construct(array $mutators, string $filePath, array $fileAst, C
     public function leaveNode(Node $node)
     {
         foreach ($this->mutators as $mutator) {
-            if ($mutator->isFunctionBodyMutator() &&
-                !$node->getAttribute(WrappedFunctionInfoCollectorVisitor::IS_INSIDE_FUNCTION_KEY)) {
-                continue;
+            if (!$this->codeCoverageData->isLineFunctionSignature($this->filePath, $node->getLine())) {
```
sidz: I'd say that we can save result into some variable as we're executing it at least 3 times

src/TestFramework/Coverage/CodeCoverageData.php
```diff
@@ -199,6 +193,18 @@ private function addTestExecutionInfo(array $coverage): array
...
+    {
+        $coverage = $this->getCoverage();
+        foreach ($coverage[$filePath]['byMethod'] as $method => $coverageInfo) {
+            if ($line == $coverageInfo['startLine']) {
```
sidz: could we use strict types here `$line === (int) $coverageInfo['startLine']` ?

tests/Mutant/Generator/MutationsGeneratorTest.php
```diff
@@ -58,26 +50,12 @@ public function test_it_can_skip_not_covered_on_file_level()
         $this->assertCount(0, $mutations);
     }
 
-    public function test_it_can_skip_not_covered_on_file_line_level()
```
borNfreee: could you please explain why did you remove it?

BackEndTea:
I have applied received feedback @sidz & @borNfreee 

Before this gets merged i'd like to add a few tests for `isLineFunctionSignature`.

An optimisation idea would be to mark all mutations that could *possibly* be on the function signature and only check if it is on a signature for those.

And instead of looping over the methods for every mutation it might be better to save that info as a hash-map or something like that.

tests/TestFramework/Coverage/CodeCoverageDataTest.php
```diff
@@ -174,6 +178,58 @@ public function test_it_throws_an_exception_when_no_coverage_found()
         $coverage->hasTests('/abc/def.php');
     }
 
+    /**
+     * @dataProvider providesFunctionSignatureLineNumbers
+     */
+    public function test_it_correctly_finds_that_line_is_signature($lineNumber)
```
borNfreee: `int $lineNumber`

tests/TestFramework/Coverage/CodeCoverageDataTest.php
```diff
@@ -174,6 +178,58 @@ public function test_it_throws_an_exception_when_no_coverage_found()
...
+     * @dataProvider providesNonFunctionSignatureLineNumbers
+     */
+    public function test_it_knows_when_lines_are_not_function_signature($lineNumber)
+    {
+        $coverageData = $this->getCodeCoverageData();
+        $filePath = '/tests/Fixtures/Files/phpunit/coverage-xml/FirstLevel/firstLevel.php';
+
+        //Known function signature lines
+        $this->assertFalse(
+            $coverageData->isLineFunctionSignature($filePath,$lineNumber),
```
borNfreee: space after comma.
I've added `tests` folder to PHPCSFixer config and fixed all issues, could you please rebase?

tests/TestFramework/Coverage/CodeCoverageDataTest.php
```diff
@@ -174,6 +178,58 @@ public function test_it_throws_an_exception_when_no_coverage_found()
...
+
+    /**
+     * @dataProvider providesNonFunctionSignatureLineNumbers
+     */
+    public function test_it_knows_when_lines_are_not_function_signature($lineNumber)
+    {
+        $coverageData = $this->getCodeCoverageData();
+        $filePath = '/tests/Fixtures/Files/phpunit/coverage-xml/FirstLevel/firstLevel.php';
+
+        //Known function signature lines
```
borNfreee: incorrect comment (should be opposite)

BackEndTea:
Rebased and addressed  issues

borNfreee:
Everything looks good.

FYI: I tried to run this branch on infection itself, and there are 918 mutations while master branch has 1010+, so 100 mutations difference. That's suspicious. 

I ran it on a small project and found that `PublicVisibility` mutator just ignored in some cases, didn't manage to find a bug yet as I'm a bit out of time today. Will continue debugging tomorrow.

BackEndTea:
>FYI: I tried to run this branch on infection itself, and there are 918 mutations while master branch has 1010+, so 100 mutations difference. That's suspicious.

>I ran it on a small project and found that PublicVisibility mutator just ignored in some cases, didn't manage to find a bug yet as I'm a bit out of time today. Will continue debugging tomorrow.

Thats strange, i'll take a look at that as well.

>Also, seems like tests are broken atm, `test_it_can_skip_not_covered_on_file_line_level` method should contain 14 instead of 13 because of added new empty line after CS fixes in master.

You're right, i fixed it now.

BackEndTea:
I think the issue lies somewhere in this function: 
https://github.com/infection/infection/blob/11c0f87f0f3061c955da91ba5b3c266281ba18bd/src/TestFramework/PhpUnit/Coverage/CoverageXmlParser.php#L57-L92

Right now if 'byMethod' is empty the `isLineFunctionSignature` thinks the function is not a signature, which may cause an issue in the `leaveNode` function

borNfreee:
Finally, found a bug. We can't rely on the code coverage in `isLineFunctionSignature()` function because if the method from the source code has coverage 0% (was not executed at all), PHPUnit's Code Coverage does not have information in xml reports about it and we can't determine if particular line is a line of the function signature.

Moreover, this is not a correct approach since such one-line function breaks the idea:

```php
public function foo() { doSomething(); return true }
```

In the example above, each node _"is on the functon signature line"_, but what we actually whant to know is whether the node is **part** of the function signature, not on its line.

So, I tried to get that information from the AST. My manual tests show everything works the same as in master.

Could you please test this branch with your code to see if it fixes the original issue @BackEndTea ?

src/Visitor/WrappedFunctionInfoCollectorVisitor.php
```diff
@@ -48,6 +50,24 @@ public function leaveNode(Node $node)
...
+            return true;
+        }
+
+        if (!$node->hasAttribute(ParentConnectorVisitor::PARENT_KEY)) {
+            return false;
+        }
+
+        $parent = $node->getAttribute(ParentConnectorVisitor::PARENT_KEY);
+
+        if ($parent instanceof Node\Param) {
```
sidz: `return $parent instanceof Node\Param;`

BackEndTea:
Your change was in the right direction, but it did not fix my issues.

This change should do the trick

src/Visitor/WrappedFunctionInfoCollectorVisitor.php
```diff
@@ -36,6 +36,8 @@ public function enterNode(Node $node)
         if ($this->isFunctionLikeNode($node)) {
             $this->scopeStack[] = $node;
             $node->setAttribute(self::IS_ON_FUNCTION_SIGNATURE, true);
+        }elseif ($this->isPartOfFunctionSignature($node)) {
```
borNfreee: Seems like all items that "are part of function signature", e.g. return value, will never get `FUNCTION_SCOPE_KEY` attribute (the next `elseif` section. This is a BC

BackEndTea:
I rebased against master since that had some changes in the `WrappedFunctionInfoCollectorVisitor`.

> Seems like all items that "are part of function signature", e.g. return value, will never get FUNCTION_SCOPE_KEY attribute (the next elseif section. This is a BC break

would 
```php
    {
        $isInsideFunction = $this->isInsideFunction($node);

        if ($isInsideFunction) {
            $node->setAttribute(self::IS_INSIDE_FUNCTION_KEY, true);
        }

        if ($this->isPartOfFunctionSignature($node)) {
            $node->setAttribute(self::IS_ON_FUNCTION_SIGNATURE, true);
        }

        if ($this->isFunctionLikeNode($node)) {
            $this->scopeStack[] = $node;
            $node->setAttribute(self::IS_ON_FUNCTION_SIGNATURE, true);
        } elseif ($isInsideFunction) {
            $node->setAttribute(self::FUNCTION_SCOPE_KEY, $this->scopeStack[count($this->scopeStack) - 1]);
        }
    }
```
Do the trick?
I'm not 100% on what the different attributes or the `scopeStack` do

borNfreee:
> Would [...] Do the trick?

yes. I've added more tests and fixed a couple of minor issues.

borNfreee:
Tested original issue on you Array-Meta repository. 100% for all MSI metrics 👍 

Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee & @sidz for merging and for all the feedback.

PR #128, opened 2018-01-16
Ignore fixtures from cs fixer

BackEndTea:
Currently this is not *needed* but we may want to have 'unfixed' files in our Fixtures to run tests against. 

For example it broke a test in #123


borNfreee:
Thank you @BackEndTea

PR #130, opened 2018-01-17
Give higher priority to custom config path

BackEndTea:
If one specifies a custom config location, you would expect that to have higher priority than the default config files.

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -122,14 +122,17 @@ class InfectionCommand extends Command
     {
         $this->container['infection.config'] = function (Container $c) use ($input): InfectionConfig {
             try {
-                $configPaths = [
-                    InfectionConfig::CONFIG_FILE_NAME,
-                    InfectionConfig::CONFIG_FILE_NAME . '.dist',
-                ];
+                $configPaths = [];
                 $customConfigPath = $input->getOption('configuration');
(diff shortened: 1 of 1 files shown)


src/Command/InfectionCommand.php
```diff
@@ -122,14 +122,15 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-                $configPaths = [
-                    InfectionConfig::CONFIG_FILE_NAME,
-                    InfectionConfig::CONFIG_FILE_NAME . '.dist',
-                ];
+                $configPaths = [];
                 $customConfigPath = $input->getOption('configuration');
                 if ($customConfigPath) {
                     $configPaths[] = $customConfigPath;
                 }
+                $configPaths = [
```
sidz: if I'm not mistaken `$configPaths` with `$customConfigPath` will be overridden by
```
 $configPaths = [
...
];
```
So in this case $customConfigPath won't work as expected as it'll be excluded.

borNfreee: just wanted to write the same ;)

BackEndTea:
@borNfreee, @sidz  I fixed the error

borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee @sidz

Issue #125, opened 2018-01-16
PHP DOM Extension not working when explicitly enabled twice


patrickfunke:
| Question | Answer
| ------------| ---------------
| Infection version | 0.7.0 (@package_version@) via Composer
| Test Framework version | PHPUnit 5.4.8, but should be independent from PHPUnit's version
| PHP version | 7.1
| Platform | Debian

Because Infection generates its own PHP config file and enables the DOM extension there explicitly (`extension=dom.so`), execution will fail with the below error if the extension is also globally enabled in the mods-available/dom.ini file.
Unfortunately, the problem is kind of hidden as the regular execution of the testcases uses the regular php.ini (with the extension not enabled twice).

To avoid this issue, Infection should check whether the DOM extension is already enabled and if so, not include it again in its separate PHP config.

```
Notice: Undefined property: DOMDocument::$documentElement in phpunit/src/Util/Configuration.php on line 557
Fatal error: Uncaught Error: Call to a member function hasAttribute() on null in phpunit/src/Util/Configuration.php on line 559
```

sidz:
hm, it shouldn't enable extensions twice. Infection generates it's own php.ini based on the global and also it's just tried to disable xdebug extension. So it takes all ini's includes from `mods-available` options and it generates one php.ini.

This feature at most was copied from the `composer` as it has the same XdebugHandler and I think you don't have the same issue with `composer`.

@patrickfunke  thx for your feedback anyway we'll try to investigate it. Right now I can't replicate that issue on my PC (with Ubuntu which has the same php configs structure as Debian).

If you will have a chance could you please try to create a vagrant box which has this issue?

sidz:
I've found what is going on! The issue was introduced in the symfony/process 3.4.2. (https://github.com/symfony/process/commit/d7c95ff7eec4722f85d7b0615ca32960e4c5c66f#diff-9a01fc0e340da4c3f1e4a16029a63977R1722)
this particular line was removed in the 3.4.3 (https://github.com/symfony/process/commit/977afa0c3658bdef76eb965476260042752d1759#diff-9a01fc0e340da4c3f1e4a16029a63977)

So to fix this issue we need to update this component. I've already checked and everything is working fine now.

Also I'm not sure why our CI did find this issue ¯\_(ツ)_/¯

@patrickfunke thank you.

patrickfunke:
Glad to see such a simple fix to the issue. Thanks for locating the fault!

borNfreee:
[...]

PR #131, opened 2018-01-17
Update symfony/process as it has an issue which introduced by 3.4.2

sidz:
Fix issue with symfony/process https://github.com/infection/infection/issues/125


borNfreee:
Thank you @sidz

PR #132, opened 2018-01-19
Update IOException

BackEndTea:
* Delete the unused interface
 * Delete the unused function
 * Delete the useless constructor
 * Delete a test for the getPath function
 * Move formatting of error message logic to the Exception
 * Add unit tests for the IOException


theofidry:
Thanks for the contrib @BackEndTea :)

The first thing I would say is why not using the Symfony FileSytem instead? There is likely to be a few edge cases and Symfony makes sure we won't run into those IMO.

sidz:
@theofidry when I implemented this code (which was copied from symfony ^_^) we with @borNfreee decided to avoid of adding new dependency :P

theofidry:
To avoid more conflicts? I think it's ok as we already have a few Symfony components and often people keep all of them at the same version. And frankly, compared to PHP-Parser it's really a small dep :P

That said I hope to be able to push forward PHP-Scoper soonish so that we don't have that problem anymore, but I need to take care of a couple of things first in order to do that

borNfreee:
> That said I hope to be able to push forward PHP-Scoper soonish so that we don't have that problem anymore

According to packagist, users downloads infection via composer [150-200 times a day](https://packagist.org/packages/infection/infection/stats) while PHAR distribution was downloaded only 200+ times after 0.7.0 release. It means people prefer much more to have a dependency in their project rather than downloading a prefixed PHAR.

To be honest, I don't know what is the reason ;) That's why I think even with the prefixed PHAR we will still have conflicts with packages.

theofidry:
@borNfreee the reason is probably the same as with PHPUnit, most static analysis tools and co.: it's easier.

The reality is that right now using PHARs is less convenient:

- you need to figure out how to download them & keep track of them
- it doesn't necessarily completely solve the issue as long as the dependencies are not bundled together

I'm working on it, but having PHP-Scoper applied to both Infection & PHPUnit will be a huge milestone for it I think

borNfreee:
We will consider using symfony filesystem next time when we will need a new function from it. At this moment - I don't see a strong reason to add it as a dependency in such refactoring-only PR.

The same is for assertion library (btw - great lib, like it) - good to have, but not necessary right now.

Thank you @BackEndTea

PR #133, opened 2018-01-19
Fix: Use actual name

localheinz:
This PR

* [x] uses the actual test framework name when rendering its version

💁‍♂️ Not sure where the name is of importance in other places, but I noticed that the names rendered aren't what the projects' maintainers use.

### DIFF
--- a/src/TestFramework/PhpSpec/Adapter/PhpSpecAdapter.php
+++ b/src/TestFramework/PhpSpec/Adapter/PhpSpecAdapter.php
@@ -10,7 +10,6 @@ declare(strict_types=1);
 namespace Infection\TestFramework\PhpSpec\Adapter;
 
 use Infection\TestFramework\AbstractTestFrameworkAdapter;
-use Infection\TestFramework\TestFrameworkTypes;
 
 class PhpSpecAdapter extends AbstractTestFrameworkAdapter
 {
@@ -41,6 +40,6 @@ class PhpSpecAdapter extends AbstractTestFrameworkAdapter
 
     public function getName(): string
     {
-        return TestFrameworkTypes::PHPSPEC;
+        return 'PhpSpec';
     }
 }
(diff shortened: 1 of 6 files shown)


tests/TestFramework/PhpUnit/Adapter/PhpUnitAdapterTest.php
```diff
@@ -46,4 +47,21 @@ public function passProvider()
             ['ERRORS!', false],
         ];
     }
+
+    private function getAdapter(): PhpUnitAdapter
```
localheinz: Extracted this method here, similar to [`PhpSpecAdapterTest::getAdapter()`](https://github.com/infection/infection/blob/0.7.0/tests/TestFramework/PhpSpec/Adapter/PhpSpecAdapterTest.php#L93-L102).

src/TestFramework/PhpSpec/Adapter/PhpSpecAdapter.php
```diff
@@ -41,6 +41,6 @@ public function testsPass(string $output): bool
 
     public function getName(): string
     {
-        return TestFrameworkTypes::PHPSPEC;
+        return 'PhpSpec';
```
localheinz: Actually, looks like different styles are used:

* http://www.phpspec.net/en/stable/manual/upgrading-to-phpspec-4.html
* http://www.phpspec.net/en/stable/manual/upgrading-to-phpspec-3.html

src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
```diff
@@ -40,6 +40,6 @@ public function testsPass(string $output): bool
 
     public function getName(): string
     {
-        return TestFrameworkTypes::PHPUNIT;
+        return 'PHPUnit';
```
localheinz: See https://phpunit.de:

screen shot 2018-01-19 at 23 55 50

borNfreee:
Thank you @localheinz

localheinz:
Thank you, @borNfreee and @theofidry!

PR #135, opened 2018-01-22
Feature: e2e tests

BackEndTea:
This PR:

* [x] Splits up File logging logic.
* [x] Excludes Fixtures from the test suite
* [x] Documentation

But more importantly it introduces a way to run e2e tests for infection.

From the project root running `./tests/e2e_tests` will start the e2e tests script.

Currently its designed to do the following for each sub folder of `tests/Fixtures/e2e`
1. Run `composer install` if there is a composer.json or composer.lock
2. Run  `run_test.bash` if it exists in the subfolder, otherwise it will run `standard_script.bash`
3. If the script exists with a non 0 exit code, log that in the `error.log` file

After it has handled all sub-folders the `e2e_tests` script exits with a 1 and displays error.log to the console if any of the earlier tests failed. If everything was successful, it simply exits with a 0.

If one wants to write a new e2e test case, there is most likely no need to write any bash scripts, since most of the basics are handeled by eiter `e2e_tests` or `standard_script.bash`.

I'd like to hear thoughts/opinions on this.


borNfreee:
I like the simplicity of this approach. Several questions though:

1. Should we run it (and how) on AppVeyor for windows?
2. Have you investigated how other projects run e2e tests (with or without special tools)?

Why I ask is because e.g. PhpSpec uses Behat for this purpose, they also run without problem e2e tests on Travis and AppVeyor.

theofidry:
@borNfreee may not be the best reference but I run e2e tests for any PHAR related project ([PHP-Scoper](https://github.com/humbug/php-scoper/blob/master/Makefile#L61) & [Box](https://github.com/humbug/box/blob/master/Makefile#L59)).

I think it's a good idea, although it might be better to do it in a separate PR

BackEndTea:
>Should we run it (and how) on AppVeyor for windows?

In its current form, no. However we could add windows versions of the 'standard' scripts, and run those on AppVeyor. This will require developers to write the same scripts twice if they need custom scripts to check everything.

>Have you investigated how other projects run e2e tests (with or without special tools)?

My approach is inspired by https://github.com/humbug/php-scoper. But i tried to make it really simple to create new tests for infection.

Using Behat could work, but it will probably require some refactoring. There are some references to global variables and the use of `getcwd()`. Which will make that a bit hairy. If we want to do that we should do some refactoring to handle those.

I also looked at setting up something like https://github.com/FriendsOfPHP/PHP-CS-Fixer, mainly the [Integration Test Case](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.10/tests/Test/AbstractIntegrationTestCase.php). But i couldn't get that to work well with infection.

borNfreee:
> If we want to do that we should do some refactoring to handle those.

I would stick to the current approach as it is very simple and straightforward until someone has any objections.

src/Config/InfectionConfig.php
```diff
@@ -209,9 +209,14 @@ private function getExcludes(): array
     /**
      * @return string|null
      */
-    public function getTextFileLogPath()
+    public function getLogPathInfoFor($log)
```
sidz: could we please add `string`  type hint?

src/Config/InfectionConfig.php
```diff
@@ -209,9 +209,14 @@ private function getExcludes(): array
...
     {
-        return $this->config->logs->text ?? null;
+        return $this->config->logs->$log ?? null;
+    }
+
+    public function getLogsTypes()
+    {
+        return $this->config->logs ?? null;
```
sidz: maybe it'd be better to remove empty array instead of null?

tests/Fixtures/e2e/Example_Test/phpunit.xml
```diff
@@ -0,0 +1,24 @@
...
+         convertErrorsToExceptions="true"
+         convertNoticesToExceptions="true"
+         convertWarningsToExceptions="true"
+         processIsolation="false"
+         stopOnFailure="false"
+         syntaxCheck="false"
+>
+    <testsuites>
+        <testsuite name="Test Suite">
+            <directory>./tests/</directory>
```
borNfreee: in composer.json `"Namespace_\\Test\\": "test/"`
but `tests` here

does it work?

src/Process/Listener/FileLoggerSubscriber/BaseFileLoggerSubscriber.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    public function onMutationTestingFinished(MutationTestingFinished $event)
+    {
+        $logType = $this->infectionConfig->getLogsTypes();
+
+        if ($logType === null) {
+            return;
+        }
+        $logType = (array) $logType;
```
sidz: if we will return only `arrays` from the `getLogsTypes` there is no need to cast this variable manually.

src/Process/Listener/FileLoggerSubscriber/BaseFileLoggerSubscriber.php
```diff
@@ -0,0 +1,99 @@
...
+
+        if (isset($logType['text'])) {
```
sidz: should we extract this strings (text, summary) into constants instead of harcoding?

src/Process/Listener/FileLoggerSubscriber/BaseFileLoggerSubscriber.php
BackEndTea: I cast it to array since the the config attribute is an instance of `'\stdClass`

sidz: fair enough. forgot about json_decode ^_^

tests/Fixtures/e2e/Example_Test/phpunit.xml
BackEndTea: For some reason it does, probably because phpunit deals with loading the tests, rather than using composer for that.

I fixed it to tests though to avoid confusion

src/Config/InfectionConfig.php
```diff
@@ -207,16 +207,16 @@ private function getExcludes(): array
     }
 
     /**
-     * @return string|null
+     * @return string
```
sidz: `string|null` should be :P

src/Process/Listener/FileLoggerSubscriber/DebugFileLogger.php
```diff
@@ -0,0 +1,77 @@
...
+                implode(
+                    array_merge($logs),
+                    "\n"
+                )
+            );
+        }
+    }
+
+    /**
+     * @param array|MutantProcess[] $processes
```
sidz: `array` annotation is redundant as you add `MutantProcess[]`

src/Config/InfectionConfig.php
BackEndTea: Woops went a bit overboard with the fixing here.

src/Process/Listener/FileLoggerSubscriber/DebugFileLogger.php
```diff
@@ -0,0 +1,77 @@
...
+
+class DebugFileLogger extends FileLogger
+{
+    public function writeToFile()
+    {
+        $logFilePath = $this->infectionConfig->getLogPathInfoFor('debug');
+
+        $logs = [];
+
+        $logs[] = "Total: " . $this->metricsCalculator->getTotalMutantsCount();
```
sidz: don't see any sense of using double quotes instead of `'` here :)

.gitignore
```diff
@@ -4,3 +4,4 @@ infection-log.txt
 build/*
 !build/.gitkeep
 .php_cs.cache
+tests/e2e/error.log
```
borNfreee: there is no such file

src/Process/Listener/FileLoggerSubscriber/BaseFileLoggerSubscriber.php
```diff
@@ -0,0 +1,126 @@
...
+use Infection\EventDispatcher\EventSubscriberInterface;
+use Infection\Events\MutationTestingFinished;
+use Infection\Filesystem\Filesystem;
+use Infection\Mutant\MetricsCalculator;
+
+class BaseFileLoggerSubscriber implements EventSubscriberInterface
+{
+    const TEXT_FILE = 'text';
+    const SUMMARY_FILE = 'summary';
+    const DEBUG_FILE = 'debug';
```
borNfreee: Could you please add documentation for these log types?

https://infection.github.io/guide/usage.html#Configuration-settings
https://github.com/infection/site/blob/master/src/guide/usage.md

borNfreee:
Thank you @BackEndTea 
Great addition!

BackEndTea:
Thanks everyone for the reviews

PR #136, opened 2018-01-22
Fix: Disable xdebug before composer analyze on travis

BackEndTea:
php-cs-fixer takes a major preformance hit when xdebug is enabled.
This pr disables xdebug on travis before that is ran, and enables it again when it is required


borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @sidz @borNfreee

PR #138, opened 2018-01-23
New file loggers

BackEndTea:
This PR is a rework of the new FileLoggers from #135. Although the end result is the same, they differ a bit in implementation/naming.

I separate it in this PR to reduce the diff of #135. Plus this also a separate feature/enhancement.

It might be nice to be able to be able to implement your own listeners in the future, but that something for another PR.


BackEndTea:
Closing this

Issue #129, opened 2018-01-16
Feature: Config setting for temp files


ryan7273:
For those on shared hosting or just wanting to separate multiple projects, it would be nice to have a setting in the config file for specifying the location of temp files. Using /tmp/infection as a default makes sense, but is not best in all cases.

borNfreee:
Thank you for suggesting this feature. As always, I encourage people to make a PR, this one should be easy. 

Otherwise, I will pick it up when I have a chance after 0.7.1 is released (not in this month I guess)

sidz:
https://github.com/infection/infection/pull/140

borNfreee:
Implemented

borNfreee:
@ryan7273 released under 0.7.1

ryan7273:
Thank you!!

PR #140, opened 2018-01-29
Config setting for temp files

sidz:
Add new `tmpDir` config option (feature request: https://github.com/infection/infection/issues/129):
```
...
"tmpDir": "/root/infecton",
...
```

I'll update documentation right after will get approval for this PR.

- [x] Documentation


tests/Utils/TempDirectoryCreatorTest.php
```diff
@@ -0,0 +1,47 @@
...
+        @unlink($this->tempDir);
+    }
+
+    public function test_it_creates_and_return_path()
+    {
+        $this->creator->createAndGet($this->tempDir);
+
+        $fullPath = sprintf('%s/%s', $this->tempDir, TempDirectoryCreator::BASE_DIR_NAME);
+
+        $this->assertTrue(is_dir($fullPath));
```
borNfreee: `$this->assertDirectoryExists($fullPath)`?

src/Config/InfectionConfig.php
```diff
@@ -243,4 +243,9 @@ function ($excludeDir) use ($srcDir) {
 
         return $excludedDirs;
     }
+
+    public function getTempDir(): string
```
theofidry: If we are not going to use `getTemporaryDirectory` I would rather go for `getTmpDir` which IMO is more standard. I does make sense with `sys_get_temp_dir()` but then PHP also has `upload_tmp_dir()` and I see `tmp` way more used in Unix, Java, Python, Ruby etc.

src/Utils/TempDirectoryCreator.php
```diff
@@ -9,29 +9,40 @@
...
+    /**
+     * @var Filesystem
+     */
+    private $fs;
+
     /**
      * @var string
      */
-    private $tempDirectory;
+    private $tempDirectoryPath;
```
theofidry: as per the comment above, `tmpDir`? eventually `*Path` but as there is no `Directory` there is no confusing IMO. I would as a result change the name of the class as well

src/Utils/TempDirectoryCreator.php
```diff
@@ -9,29 +9,40 @@
 
 namespace Infection\Utils;
 
+use Infection\Filesystem\Filesystem;
+
 class TempDirectoryCreator
 {
+    const BASE_DIR_NAME = 'infection';
```
theofidry: do we want it to be re-used? Otherwise we might as well add `/** @private */` here as we are not in PHP 7.1 hence cannot do `private const X`

src/Utils/TempDirectoryCreator.php
```diff
@@ -9,29 +9,40 @@
...
     /**
      * @var string
      */
-    private $tempDirectory;
+    private $tempDirectoryPath;
+
+    /**
+     * @param Filesystem $fs
+     */
+    public function __construct(Filesystem $fs)
```
theofidry: `$fileSystem`?

tests/Utils/TempDirectoryCreatorTest.php
```diff
@@ -0,0 +1,47 @@
...
+
+    protected function setUp()
+    {
+        $this->creator = new TempDirectoryCreator(new Filesystem());
+        $this->tempDir = sys_get_temp_dir();
+    }
+
+    protected function tearDown()
+    {
+        @unlink($this->tempDir);
```
theofidry: if we have the fileSystem instance here we might as well use its `remove()` function?

sidz: we don't have this method in our version of the Filesytem yet and I don't see any sense to create it just for tests.

src/Utils/TempDirectoryCreator.php
sidz: I'm using it in the tests to remove test directory. I'll add `@private` annotation anyway

sidz:
@borNfreee @theofidry updated

tests/Utils/TempDirectoryCreatorTest.php
theofidry: you could just use the Symfony one: 

```php
(new SymfonyFilesystem())->remove($this->tmpDir);
```

At least we would be sure the content of that directory is recursively removed as right now I'm not even sure this is working

theofidry:
So are we going with `tmp` or `temp`? I do prefer the former but you're free to stick to  `temp`. However let's pick one and apply it everywhere, right now there is a mixed of both

borNfreee:
> So are we going with tmp or temp?

I would stick to `tmp` because of reasons you described above

tests/Utils/TempDirectoryCreatorTest.php
sidz: we don't have `symfony/filesystem` as dependency but if you insist I'll backport `remove` method as well

theofidry: Then I would do so yes. You can easily double check that it's not working as it is:

```
$ mkdir tmp
$ touch tmp/foo

$ psysh
>>> unlink('tmp')
PHP Warning:  unlink(tmp): Operation not permitted on line 1
```

So we are actually not cleaning up at all here

sidz:
in this case maybe it'll be a good idea to add new dependency(`symfony/filesystem`) and remove our implementations? 
thoughts @borNfreee , @theofidry

borNfreee:
> in this case maybe it'll be a good idea to add a new dependency(symfony/filesystem) and remove our implementations?

This is something I can't decide in my head. My thoughts are:

- it is ok to have a very small own implementation of FS class
- at the same time if it is growing - time to think about using symfony's one
- but I still think that extra dependency (in particular SF Filesystem) is not needed, especially when we need only 3 methods of the whole package

I'm happy to merge it as is, but next time if we *really* need another method(s) from FS, we will add a new dependency

theofidry:
I don't feel strongly about that dependency so as you wish :)

borNfreee:
@sidz build fails. Could you please take a look? https://travis-ci.org/infection/infection/jobs/335193567#L682

I tried to restart it but it fails with the same error

sidz:
@borNfreee oops. has been fixed. Also `unlink` was replaced by `remove` as it doesn't work as we expect.

borNfreee:
Thank you!

PR #143, opened 2018-01-30
Added support for Codeception test framework

### LINKED ISSUE #45: Integrate Codeception testing framework
I'm using [codeception](http://codeception.com/) as my main testing framework and may be interested in providing the integration. Has anybody else already started working on codeception integration or have there been any blockers I don't know about yet?

tobiasstadler:
Fixes #45


DavertMik:
Any updates?

theofidry:
Will try to review that in the course of next week, I'm sure the rest of the team will too. Thanks for the hard work and sorry to be a bit slow on this one :)

src/TestFramework/Codeception/CommandLine/ArgumentsAndOptionsBuilder.php
```diff
@@ -0,0 +1,51 @@
...
+
+    public function build(string $configPath, string $extraOptions, Mutant $mutant = null): string
+    {
+        $options = [
+            'run',
+            '--no-colors'
+        ];
+
+        if ($mutant !== null) {
+            $options[] = '-o "paths: output: ' . $this->tempDir . '/' . $mutant->getMutation()->getHash() . '"';
```
borNfreee: could you please explain what is it used for?

src/TestFramework/Codeception/Config/Builder/InitialConfigBuilder.php
```diff
@@ -0,0 +1,30 @@
...
+    private $originalConfigPath;
+
+    public function __construct(string $originalConfigPath)
+    {
+        $this->originalConfigPath = $originalConfigPath;
+    }
+
+    public function build(): string
+    {
+        return $this->originalConfigPath;
```
borNfreee: Maybe it's better to create all needed for Infection parameters in a new, temp config like it is done for PHPUnit and PHPSpec rather that adding/overriding config in `ArgumentsAndOptionsBuilder`?

This will probably help to avoid passing `Mutant $mutant` parameter to `ArgumentsAndOptionsBuilder`

In other words, I suggest to move this code:

```php
$options[] = '-o "paths: output: ' . $this->tempDir . '"';
$options[] = '-o "coverage: enabled: true"';
$options[] = '--coverage-phpunit ' . CodeCoverageData::CODECEPTION_COVERAGE_DIR;
```

from `ArgumentsAndOptionsBuilder` and create temp config. It will be consistent with the other Test Framework and will help to avoid changing the interface

src/TestFramework/Codeception/Config/Builder/MutationConfigBuilder.php
```diff
@@ -0,0 +1,72 @@
...
+        return $this->originalConfigPath;
+    }
+
+    private function createCustomAutoloadWithInterceptor(Mutant $mutant): string
+    {
+        $originalFilePath = $mutant->getMutation()->getOriginalFilePath();
+        $mutatedFilePath = $mutant->getMutatedFilePath();
+        $interceptorPath = dirname(__DIR__, 4) . '/StreamWrapper/IncludeInterceptor.php';
+
+        // TODO change to what it was (e.g. app/autoload - see simplehabits)
```
borNfreee: please do not copy todo comments

src/TestFramework/Codeception/Config/Builder/MutationConfigBuilder.php
```diff
@@ -0,0 +1,72 @@
...
+    public function __construct(string $tempDirectory, string $projectDirectory, string $originalConfigPath)
+    {
+        $this->tempDirectory = $tempDirectory;
+        $this->projectDirectory = $projectDirectory;
+        $this->originalConfigPath = $originalConfigPath;
+    }
+
+    public function build(Mutant $mutant): string
+    {
+        $_SERVER['CUSTOM_AUTOLOAD_FILE_PATH'] = $customAutoloadFilePath = sprintf(
```
borNfreee: could you please expand on why do you need it?

src/TestFramework/Codeception/CustomAutoloadFilePath.php
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright © 2018 Tobias Stadler
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Infection\TestFramework\Codeception;
+
+use Codeception\Events;
```
borNfreee: 1. it will not work since `codeception/codeception` is in `require-dev` section and all files from this section are absent in PHAR distribution
2. we need to find a way to not use Codeception dependency at all. Infection should not depend on any test framework.

could you please explain what is the goal of this file, we will try to find another way to achieve the same goal

tests/TestFramework/Codeception/Adapter/CodeceptionAdapterTest.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        return [
+            ['OK, but incomplete, skipped, or risky tests!', true],
+            ['OK (5 tests, 3 assertions)', true],
+            ['FAILURES!', false],
+            ['ERRORS!', false],
+        ];
+    }
+
+    protected function tearDown()
```
borNfreee: please extend `Mockery\Adapter\Phpunit\MockeryTestCase` file. No need to use this tearDown function anymore

phpstan.neon
```diff
@@ -1,4 +1,5 @@
 parameters:
     ignoreErrors:
         - '#Access to an undefined property PhpParser\\Node(.*?)#'
-        - '#Call to an undefined method PhpParser\\Node(.*?)#'
\ No newline at end of file
+        - '#Call to an undefined method PhpParser\\Node(.*?)#'
+        - '#Call to an undefined method Mockery\\MockInterface(.*?)#'
```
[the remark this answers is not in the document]
tobiasstadler: phpstan complains about a call to an undefined method Mockery\MockInterface::getMutation() in tests/TestFramework/Codeception/CommandLine/ArgumentsAndOptionsBuilderTest.php

src/TestFramework/Codeception/Config/Builder/InitialConfigBuilder.php
tobiasstadler: unfortunately Codeception does not handle absolute paths in its config files, but maybe I can get it working using relative paths.

src/TestFramework/Factory.php
```diff
@@ -104,10 +108,22 @@ public function create(string $adapterName): AbstractTestFrameworkAdapter
...
+                new CodeceptionArgumentsAndOptionsBuilder($this->tempDir),
+                $this->versionParser
+            );
+        }
+
         throw new \InvalidArgumentException(
             sprintf(
                 'Invalid name of test framework. Available names are: %s',
-                implode(', ', [TestFrameworkTypes::PHPUNIT, TestFrameworkTypes::PHPSPEC])
+                implode(', ', [TestFrameworkTypes::PHPUNIT, TestFrameworkTypes::PHPSPEC, TestFrameworkTypes::CODECEPTION])
```
sidz: could you please use `TestFrameworkTypes::TYPES` constant here?

src/Command/InfectionCommand.php
```diff
@@ -69,7 +69,7 @@ protected function configure()
                 'test-framework',
                 null,
                 InputOption::VALUE_REQUIRED,
-                'Name of the Test framework to use (phpunit, phpspec)',
+                'Name of the Test framework to use (phpunit, phpspec, codeception)',
```
sidz: could you please use `TestFrameworkTypes::TYPES` constant here as we do in the ConfigureCommand?

tests/Fixtures/e2e/Codeception_Basic/run_test.bash
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+if [[ $PHPDBG = 1 ]]
```
sidz: @borNfreee told that he had an issue with this variant on his mac. Could you please change this line to the `if [ "$PHPDBG" = "1" ]` ?

tobiasstadler: I had an issue with if [[ $PHPDBG=1 ]] on my Mac.  if [[ $PHPDBG = 1 ]] works for me.

@borNfreee Does it also work for you?

borNfreee: yes, it does not work, see https://github.com/infection/infection/pull/151#discussion_r166079227

src/TestFramework/Codeception/Config/Builder/InitialConfigBuilder.php
DavertMik: I'd suggest creating Codeception's Extension which will transform the current config the way you need it. 
Extension can be published as a standalone package

src/TestFramework/Codeception/Config/Builder/MutationConfigBuilder.php
tobiasstadler: This is needed in order to autoload the mutated file. As far as I know Codeception does not support a custom bootstrap file (as e.g. PHPUnit)

src/TestFramework/Codeception/CustomAutoloadFilePath.php
tobiasstadler: I don't thin this is a problem, because the code path is only used when the test framework is codeception.

tests/e2e_tests
```diff
@@ -19,7 +19,7 @@ do
         composer install > /dev/null 2>&1
     fi
 
-    if [ -f "run_tests.bash" ]
+    if [ -f "run_test.bash" ]
```
BackEndTea: If you rebase on master please undo this change.

The intended filename is `run_tests.bash`, the `then` clause not running the correct file has been fixed on master.

tests/Fixtures/e2e/Codeception_Basic/run_test.bash
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
BackEndTea: Please rename the file to `run_tests.bash`

tests/Fixtures/e2e/Codeception_Basic/composer.json
```diff
@@ -0,0 +1,15 @@
...
+      "psr-4": {
+        "Codeception_Basic\\": "src/"
+      }
+    },
+    "autoload-dev": {
+      "psr-4": {
+        "Codeception_Basic\\Test\\": "tests/unit/"
+      }
+    }
+  }
```
BackEndTea: Missing a new line at the end of the file

tests/TestFramework/Codeception/Config/YamlConfigurationHelperTest.php
```diff
@@ -0,0 +1,144 @@
...
+
+namespace Infection\Tests\TestFramework\Codeception\Config;
+
+use Infection\Filesystem\Filesystem;
+use Infection\TestFramework\Codeception\Config\YamlConfigurationHelper;
+use Infection\Utils\TmpDirectoryCreator;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Yaml\Yaml;
+
+function rp(string $path) : string
```
BackEndTea: Could you move this function into the test class as a helper function

src/TestFramework/Codeception/Config/Builder/MutationConfigBuilder.php
sidz: please use `putenv` instead of touching global variables

src/TestFramework/AbstractTestFrameworkAdapter.php
```diff
@@ -74,12 +74,12 @@ public function __construct(
      *
      * @return string
      */
-    public function getExecutableCommandLine(string $configPath, string $extraOptions, bool $includePhpArgs = true): string
+    public function getExecutableCommandLine(string $configPath, string $extraOptions, bool $includePhpArgs = true, Mutant $mutant = null): string
```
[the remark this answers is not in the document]
borNfreee: this still needs to be fixed

src/TestFramework/Codeception/Config/Builder/MutationConfigBuilder.php
sidz: Infection runs frameworks via symfony console which knows how to work with envs. We already use it :P.

p.s. IMHO Don't like to touch global variables

theofidry: I'm not sure to follow, I don't see how using the Console component helps to make it thread-safe here.

Maybe for reference: https://github.com/symfony/recipes/pull/156

Also I don't like to touch global variables much, but that's exactly what put/getenv are doings and regardless of what we are using we should use a prefix here IMO

BackEndTea:
Could you run `composer cs:fix`?

BackEndTea:
I'll give this good look over after the weekend, and see if i can break anything.

src/TestFramework/Factory.php
```diff
@@ -111,10 +115,23 @@ public function create(string $adapterName, bool $skipCoverage): AbstractTestFra
             );
         }
 
+        if ($adapterName === TestFrameworkTypes::CODECEPTION) {
+            $codeceptionConfigPath = $this->configLocator->locate(TestFrameworkTypes::CODECEPTION);
+            $codeceptionConfigContent = file_get_contents($codeceptionConfigPath);
+
+            return new CodeceptionAdapter(
+                new TestFrameworkExecutableFinder(CodeceptionAdapter::EXECUTABLE),
+                new CodeceptionInitialConfigBuilder($this->tempDir, $this->projectDir, $codeceptionConfigContent, $this->infectionConfig->getSourceDirs()),
```
Fenikkusu: I was just trying this out on a project. A recent MR has changed the variable `$this->tempDir` to `$this->tmpDir`. I had to fix this before I could use the PR.

src/TestFramework/Factory.php
```diff
@@ -112,10 +116,23 @@ public function create(string $adapterName, bool $skipCoverage): AbstractTestFra
             );
         }
 
+        if ($adapterName === TestFrameworkTypes::CODECEPTION) {
+            $codeceptionConfigPath = $this->configLocator->locate(TestFrameworkTypes::CODECEPTION);
+            $codeceptionConfigContent = file_get_contents($codeceptionConfigPath);
+
+            return new CodeceptionAdapter(
+                new TestFrameworkExecutableFinder(CodeceptionAdapter::EXECUTABLE),
```
Fenikkusu: Not Found For Me. Changed It To `TestFrameworkFinder` and started working again.

Fenikkusu:
So far so good. I'm having issues running it through a docker image ( #254 ), but I think that is related more to infection than the codeception work.

sidz:
@Fenikkusu first of all this PR has issues which should be fixed. So @tobiasstadler could you please check Travis build https://travis-ci.org/infection/infection/builds/357027746?utm_source=github_status&utm_medium=notification

Fenikkusu:
I've come across an issue I didn't come across originally. If you are using params:env:.env, The system chokes because it can't find the .env file. I'm working on resolving the issue and have a partial fix at https://github.com/twistersfury/infection:codeception. However, this resulted in additional errors I have not yet had the time to resolve.

Fenikkusu:
I've figured out the additional errors. I was running into two issues. The first was I had an acceptance suite installed, but no actual tests installed. When coverage ran, codeception choked because it was trying to connect to the acceptance url, which didn't exist. This error was easy enough for me to resolve, just by disabling the suite. However, it brings up the point of should the ability be added as part of the configuration to be able to specify which suites are ran?

The next issue took me a bit to track down. This time, phpunit was choking on tokenization for coverage with phpunit. Basically I had files in the project that were empty, but weren't PHP files. I tracked down the command infection was running and tried to replicate the specifics. I ultimately realized that when the Codeception suite runs, it is setting the configuration of coverage files to '*' based off the infection config. IE:

```json
    "source": {
        "directories": [
            "src"
        ]
    }
```

Is translated into:
```yaml
coverage:
  enabled: true
  include:
    - src/*
```

Because it is '*', it is including all files, even non-php ones. I'm fairly certain if we just remove the '*' that will fix the issue. My question though, at this point is if there is a specific reason why * was used instead of just using the folder path?

Fenikkusu:
> So for those suites inflection should be disabled, that's for sure.

I agree that they should be disabled, but I still think the ability needs to be passed onto the user via configuration. I say that because the suites are not limited to unit, functional, and acceptance. For example, I've a project that has a suite called 'legacy' and another called 'coverage'. Both of these suites run under the pretense of unit testing and should be ran as part of the testing. I think in the end, an option in the json configuration needs to be added for suites. When not specified, it defaults to 'unit'. When specified, it runs all the suites specified. This could easily be a string that just directly injects into the command line arguments.

Fenikkusu:
Update: I tested the code using just the path and not the '*'. I still run into the same issue. The easiest solution is to just change it to be *.php. Done and done. The only problem I see with that is that the possibility of file names not ending in .php. IE: php5, php4, and phtml are all considered valid extensions, though rare (with phtml likely not being a class).

src/TestFramework/Codeception/CustomAutoloadFilePath.php
sanmai: This could work if this file won't be a part of Infection. Instead you could tell Codeception to include a file from the outside Infection. E.g. there's an `IncludeInterceptor` which [is added to the tested code at one point](https://github.com/infection/infection/blob/5bfd1be14bd3762a573717e118f6115a3426d77f/src/TestFramework/PhpSpec/Config/Builder/MutationConfigBuilder.php#L70). I'm not entirely sure you could do what you want there, but there's an opportunity.

Fenikkusu:
Noting Additional Issue: If codeception errors (IE: use db module without an accessible db), the error isn't getting properly echoed by infection.

borNfreee:
> the error isn't getting properly echoed by infection

that sucks. Could you please try set error hanlder as it is done in https://github.com/infection/infection/pull/275 and see if it helps?

Fenikkusu:
After a bit of digging, I was able to figure out how to get this to replicate consistently. If I'm not debugging, I never get any output from the error (again, use the db module with invalid db). If I put in a break point and force the codecept command to finish `before` the incept command has time to kick off the wait process, I get the proper output. AKA, break on line 303 of Symfony's Process.php file and step the codecept run to the next line. Let the codecept finish, then continue running infection. In this case, you will get the correct output.

Fenikkusu:
> that sucks. Could you please try to set an error hanlder as it is done in #275 and see if it helps?

@borNfreee, I tried that as well but saw no difference. At this point it appears to be a race condition in the underlying symfony framework.

BackEndTea:
@tobiasstadler  are you still working on this?

tobiasstadler:
Sorry, I am not working on this anymore.

BackEndTea:
> Sorry, I am not working on this anymore.

I'm gonna go ahead and close this PR then. Thank you for the work you put into it.
Even though this won't be merged it may be usefull if someone wants to pick it up at a later point.

Fenikkusu:
I've already done number of updates to this, though I haven't refreshed it in a couple of months. I'll open a new MR once I have a chance to rebase and clean up. I've been using it in a few applications, so I think it's about 80% there.

maks-rafalko:
Implemented in https://github.com/infection/infection/pull/800

PR #144, opened 2018-01-30
Tests: add a phpspec & phpunit comparison test

BackEndTea:
This PR:

* [x] Adds and e2e test that makes sure phpunit and phpspec give the same output
* [x] Fixes a typo in `tests/e2e_tests`


tests/Fixtures/e2e/Test_Frameworks/spec/SourceClassSpec.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace spec\Namespace_;
+
+use PhpSpec\ObjectBehavior;
+
+class SourceClassSpec extends ObjectBehavior
+{
+    public function it_returns_hello()
```
borNfreee: not a big deal to have mistakes in fixture classes, but still - `hello`? probably copy paste issue 

two lines below - `1, 2` - space after comma

tests/Fixtures/e2e/Test_Frameworks/README.md
```diff
@@ -0,0 +1,9 @@
+# Test Framework Equality
```
theofidry: is this README necessary?

tests/Fixtures/e2e/Test_Frameworks/phpspec.yml
```diff
@@ -0,0 +1,7 @@
+suites:
+  source_suite:
+    namespace: Namespace_
+    psr4_prefix: Namespace_
+
+extensions:
+  LeanPHP\PhpSpec\CodeCoverage\CodeCoverageExtension: ~
```
theofidry: missing line return

tests/Fixtures/e2e/Test_Frameworks/phpunit.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit backupGlobals="false"
+         backupStaticAttributes="false"
+         bootstrap="./vendor/autoload.php"
```
theofidry: `vendor/autoload.php` (all relative paths are necessarily relative to the config file)

tests/Fixtures/e2e/Test_Frameworks/phpunit.xml
```diff
@@ -0,0 +1,24 @@
...
+         colors="true"
+         convertErrorsToExceptions="true"
```
theofidry: I think a few of those options are either unnecessary or the default. If you open this in phpStorm and add the line `xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.4/phpunit.xsd"` (match the right version) and fetch the resource, you should be able to see which options are unnecessary

tests/Fixtures/e2e/Test_Frameworks/run_tests.bash
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+#Run test suite with phpunit
+php ../../../../bin/infection
+if [ $? != 0 ]
```
theofidry: you can put `set -euf -o pipefail` at the beginning of the file instead of having to rely on those return values each time

If you do, it also becomes unnecessary to do the whole checks at each `diff` calls as well, you can just chain things together and if an error occurs you should know where exactly from the error

Just to be clear: what you did is perfectly correct and valid, I think the above however would simplify this script

tests/Fixtures/e2e/Test_Frameworks/run_tests.bash
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+#Run test suite with phpunit
+php ../../../../bin/infection
```
theofidry: as you are re-using this I would define a readonly constant:

```bash
readonly INFECTION=../../../../bin/infection
```

tests/Fixtures/e2e/Test_Frameworks/run_tests.bash
theofidry: The same applies to `tests/e2e_tests`

tests/Fixtures/e2e/Test_Frameworks/run_tests.bash
theofidry: Actually maybe in order to run fully e2e tests we should use the PHAR for running those tests, what do you think?

tests/Fixtures/e2e/Test_Frameworks/run_tests.bash
BackEndTea: Great, i didn't know that was an option.

tests/Fixtures/e2e/Test_Frameworks/run_tests.bash
BackEndTea: >as you are re-using this I would define a readonly constant:
Sounds good to me

>Actually maybe in order to run fully e2e tests we should use the PHAR for running those tests, what do you think?

\We could, but that would require building the phar, and i'm not 100% sure on how to do that, so any help would be appreciated

tests/Fixtures/e2e/Test_Frameworks/spec/SourceClassSpec.php
BackEndTea: Ah my bad, that is indeed a coppy paste issue.

tests/Fixtures/e2e/Test_Frameworks/README.md
BackEndTea: Not 100%, but it makes it clear what the test is trying to prove

tests/Fixtures/e2e/Test_Frameworks/run_tests.bash
borNfreee: I think we should think about Phar outside of this PR

theofidry: 👍 I can do that when I integrate the new version of Box

BackEndTea:
@theofidry @borNfreee I applied your feedback, what do you think?

tests/Fixtures/e2e/Test_Frameworks/composer.json
```diff
@@ -0,0 +1,23 @@
+{
+  "config": {
```
sidz: we're using 4 spaces in the .json files. Could we please use the same style everywhere?

borNfreee: @theofidry left same comments in https://github.com/infection/infection/pull/151

leave it with me, I will fix it there

BackEndTea: @borNfreee Do you mean you will fix it in this PR, or do you want me to fix it?

borNfreee: I will fix it in my PR, let's get it merged ;)

borNfreee:
Thank you @BackEndTea

PR #145, opened 2018-01-31
Skip XdebugHandlerTest when infection is running via phpdbg

sidz:
We need to skip `XdebugHandlerTest` when infection is running via `phpdbg -qrr bin/infection`. So right now infection is not working with phpdbg as we expect.

p.s. Suggest to add phpdbg to travis to avoid issues like that in future. Please vote and I'll try to add it in the current PR

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,15 +5,26 @@ sudo: false
 matrix:
   include:
     - php: 7.0
-      env: INFECTION=1
+    - php: 7.0
+      env: PHPDBG=1
     - php: 7.1
-      env: INFECTION=1
     - php: 7.1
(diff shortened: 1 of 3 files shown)


theofidry:
👍  for adding a build with phpdbg

borNfreee:
I'm for it

vote

sidz:
done, but we have different MSI score for `xdebug` and `phpdbg` as we skip tests with phpdbg.
So should we have 2 different scores for both tools or one score would be enough?

theofidry:
I think one is enough. The score is bound to be different depending of the coverage tool (it can be even between xdebug versions)

PR #151, opened 2018-02-03
Allow relative path for `tmpDir` config setting. 

borNfreee:
- [x] Allow relative path for `tmpDir` config setting. 
- [x] Fix bug with e2e tests with checking $PHPDBG env var.
- [x] Update documentation with new relative feature (https://github.com/infection/site/commit/f5462d3e85ea1bf4c932f30f140dd769b236eab0)


src/Filesystem/Filesystem.php
```diff
@@ -102,4 +102,22 @@ public function remove($files)
...
     }
+
+    /**
+     * Returns whether the file path is an absolute path.
+     *
+     * @param string $file A file path
+     *
+     * @return bool
+     */
+    public function isAbsolutePath(string $file): bool
```
sidz: is this a time to add a new dependency? :D

borNfreee: lol, expected question. I still don't know.

sidz:
could you please check travis ^^?

src/Config/InfectionConfig.php
```diff
@@ -251,6 +258,12 @@ function ($excludeDir) use ($srcDir) {
 
     public function getTmpDir(): string
     {
-        return $this->config->tmpDir ?? sys_get_temp_dir();
+        $tmpDir = $this->config->tmpDir ?? sys_get_temp_dir();
+
+        if ($this->filesystem->isAbsolutePath($tmpDir)) {
```
sidz: I'd say that we don't need to execute this line for `sys_get_temp_dir`.

so it's actually only when `$this->config->tmpDir` is not empty

tests/Fixtures/e2e/standard_script.bash
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-if [[ $PHPDBG=1 ]]
+if [ "$PHPDBG" = "1" ]
```
sidz: what was the issue? on linux it works fine

borNfreee: on macos it was always `true`

borNfreee:
> could you please check travis ^^?

seems like something wrong with PHPCSFixer. New version has been downloaded on Travis and it wants to remove blank line after license, while we have

```
'header_comment' => [
            'commentType' => 'PHPDoc',
            'header' => $header,
            'location' => 'after_open',
            'separate' => 'bottom', <-------
        ],
```

Conflict between header style and `no_blank_lines_after_phpdoc` :|

src/Config/InfectionConfig.php
borNfreee: fixed

src/Config/InfectionConfig.php
```diff
@@ -251,6 +258,17 @@ function ($excludeDir) use ($srcDir) {
 
     public function getTmpDir(): string
     {
-        return $this->config->tmpDir ?? sys_get_temp_dir();
+
```
sidz: remove additional new line

src/Config/InfectionConfig.php
```diff
@@ -251,6 +258,17 @@ function ($excludeDir) use ($srcDir) {
...
+        if (empty($this->config->tmpDir)) {
```
sidz: maybe we need to throw an exception for case when `"tmpDir": ""` instead of use `sys_temp_dir`? what do you think?

borNfreee: it's better to continue working if we can rather than throw an exception.

I mean even if we throw an exception here, Infection should fall back to sys_temp_dir since this is not critical mistake and does not block mutation testing. WDYT?

sidz: imho to be user friendly we need to show warning message in this case if we don't want to throw an exception. (it's not easy to do right now)
so feel free to leave it as is

borNfreee:
Created an issue in PHPCSFixer repo about failures on Travis with the new downloaded 2.10.1 version: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3503

src/Config/InfectionConfig.php
borNfreee: forgot about it, will do, thank you

src/Filesystem/Filesystem.php
theofidry: I think it's time to, it's the 3rd method we are copy-pasting from it :) Besides it's not like filesystem is a huge dependency, it's quite slim and don't change much

tests/Fixtures/e2e/Custom_tmp_dir/composer.json
```diff
@@ -0,0 +1,20 @@
+{
```
theofidry: we're using 4 spaces in other files aren't we?

tests/Fixtures/e2e/Custom_tmp_dir/composer.json
```diff
@@ -0,0 +1,20 @@
...
+    "psr-4": {
+      "Namespace_\\": "src/"
+    }
+  },
+  "autoload-dev": {
+    "psr-4": {
+      "Namespace_\\Test\\": "tests/"
+    }
+  }
+}
```
theofidry: missing line return

tests/Fixtures/e2e/Custom_tmp_dir/phpunit.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit backupGlobals="false"
+         backupStaticAttributes="false"
+         bootstrap="./vendor/autoload.php"
+         colors="true"
+         convertErrorsToExceptions="true"
```
theofidry: aren't those the default? You can easily find out with phpstorm and adding `xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.4/phpunit.xsd"
 ` e.g. https://github.com/humbug/box/blob/master/phpunit.xml.dist#L3

tests/Fixtures/e2e/Custom_tmp_dir/tests/SourceClassTest.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Namespace_\Test;
+
+use Namespace_\SourceClass;
+
+class SourceClassTest extends \PHPUnit\Framework\TestCase
```
theofidry: Not a big deal but I find better to be consistent and import this class with a use statement?

tests/Fixtures/e2e/standard_script.bash
theofidry: I couldn't reproduce it, might be a different shell (I'm using zsh), but in any case it's better to compare strings in bash

src/Filesystem/Filesystem.php
sidz: @borNfreee let's do this. If you don't want to do it in the current PR let's do it as separate.

borNfreee: got it, I will do it in this PR

tests/Fixtures/e2e/standard_script.bash
borNfreee: `if [[ $PHPDBG=1 ]]` does not work on Terminal app in MacOS (while works on zsh). See image below. Also, `if [[ $PHPDBG=1 ]]` is not the same as `if [[ $PHPDBG = 1 ]]`


phpdbg

borNfreee:
All comments addressed, including documentation. Please review again 😋

tests/Fixtures/e2e/Custom_tmp_dir/infection.json
```diff
@@ -0,0 +1,12 @@
+{
+  "timeout": 25,
```
sidz: 4 spaces :P

borNfreee: fixed

PR #152, opened 2018-02-04
Enhancement: Implement integer increment and decrement mutators

localheinz:
This PR

* [x] implements `IntegerIncrement` and `IntegerDecrement` mutators
* [x] reduces the minimum required MSI to 58%

💁‍♂️ Not sure if something similar already exists and I missed it, but oftentimes tests do not verify boundaries, so maybe this helps to avoid off-by-one errors.

PR #153, opened 2018-02-04
Fix: Run 'composer cs:fix'

localheinz:
This PR

* [x] runs `composer cs:fix`

Somewhat related to #152.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/compare/v2.10.0...v2.10.2.

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -4,7 +4,6 @@
  *
  * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
  */
-
 use Infection\Php\XdebugHandler;
 use Infection\Php\ConfigBuilder;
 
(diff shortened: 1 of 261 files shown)


borNfreee:
[I faced the same issue ](https://github.com/infection/infection/pull/151#issuecomment-362852708) and I thought this is a bug so I create an [issue](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3503) in PHPCSFixer repo. 

At least, this is strange that patch version (2.10.0 -> 2.10.1) is not compatible with the previous one.

Let's get it merged and unblock 2 PRs.

Thank you @localheinz

PR #156, opened 2018-02-05
Fix appveyor build

BackEndTea:
Whatever bug was causing appveyor to fail beforehand appears to havebeen fixed. It was now failing due to tests not converting the string to a 'realpath', which caused some issues due to windowspaths. This change should fix all occurrences of this error

### DIFF
--- a/tests/Finder/LocatorTest.php
+++ b/tests/Finder/LocatorTest.php
@@ -99,14 +99,15 @@ class LocatorTest extends TestCase
         $found = $locator->locateAnyOf(['file.txt', 'secondfile.txt']);
         $this->assertStringEndsWith(
             'tests/Fixtures/Locator/file.txt',
-            $found,
+            p($found),
             'Found an incorrect file, orders may be broken.'
         );
 
         $found = $locator->locateAnyOf(['secondfile.txt', 'file.txt']);
+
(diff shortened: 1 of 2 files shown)


tests/Finder/LocatorTest.php
```diff
@@ -104,6 +104,8 @@ public function test_it_returns_first_possible_file()
         );
 
         $found = $locator->locateAnyOf(['secondfile.txt', 'file.txt']);
+        $found = realpath($found);
```
borNfreee: probably, you can wrap it with `p` function as in other tests in this file. It was created to transform window paths to unix-like

see for example:

https://github.com/infection/infection/blob/f59d66736ae089ffefd167e681382d3398655377/tests/Finder/LocatorTest.php#L28

borNfreee:
> Whatever bug was causing appveyor to fail beforehand appears to havebeen fixed.

Sometimes it requires completely removing the whole project and create it again in AppVeyor (oh windows!)

tests/Finder/LocatorTest.php
BackEndTea: Yeah i noticed that as well, should be fixed now.

borNfreee:
Thank you @BackEndTea 👍

PR #159, opened 2018-02-06
Enhancement: Enable ordered_imports fixer

localheinz:
This PR

* [x] enables the `ordered_imports` fixer
* [x] runs `php-cs-fixer`

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.10.2#usage:

>**ordered_imports**
>
>Ordering use statements.
>
>Configuration options:
>
>* `importsOrder` (`array`, `null`): defines the order of import types; defaults to `null`
>* `sortAlgorithm` (`'alpha'`, `'length'`): whether the statements should be sorted alphabetically or by length; defaults to `'alpha'`


borNfreee:
Thank you @localheinz

localheinz:
Thank you, @borNfreee, @sidz, and @theofidry!

PR #160, opened 2018-02-07
Fix: Reference schema relative to phpunit.xml

localheinz:
This PR

* [x] references `phpunit.xsd` relative to `phpunit.xml`

💁‍♂️ This way it will always reference the schema that is actually installed, and there will be no need to point to a new location when `phpunit/phpunit` is updated.


borNfreee:
Thank you @localheinz 👍

localheinz:
Thank you, @borNfreee and @theofidry!

PR #161, opened 2018-02-07
Enhancement: Use all the columns when running phpunit

localheinz:
This PR

* [x] uses all the columns available in the terminal when running `phpunit`

### Before

```

...............................................................  63 / 320 ( 19%)
............................................................... 126 / 320 ( 39%)
............................................................... 189 / 320 ( 59%)
............................................................... 252 / 320 ( 78%)
............................................................... 315 / 320 ( 98%)
.....                                                           320 / 320 (100%)

Time: 1.04 seconds, Memory: 18.00MB

OK (320 tests, 578 assertions)
```

### After

```

............................................................................................................................................................................................................................. 221 / 320 ( 69%)
...................................................................................................                                                                                                                           320 / 320 (100%)

Time: 1.05 seconds, Memory: 18.00MB

OK (320 tests, 578 assertions)
```

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -3,6 +3,7 @@
          xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/6.1/phpunit.xsd"
          bootstrap="./vendor/autoload.php"
          colors="true"
+         columns="max"
 >
     <testsuites>
         <testsuite name="Application Test Suite">


theofidry:
Not a big fan of that one tbh

localheinz:
@theofidry 

If `phpunit` cannot detect the width of the terminal, it will fall back to using 80 as width.

theofidry:
It's not really the problem, it's just that I do prefer when it doesn't take my whole screen :) I also find it more consistent with the other tools like Behat or phpspec. The only exception I have on top of my head is PHP-CS-Fixer

borNfreee:
Infection itself uses dot formatter with limited width as well, so for me it is also more natural to have 80 as width.

localheinz:
No problem, @borNfreee and @theofidry!

PR #162, opened 2018-02-07
Enhancement: Improve Mutator tests

BackEndTea:
This PR:

* [x] Simplifies Mutator tests
* [x] Fixes an issue with Full caps true or false not being mutated by their mutators
* [x] Does some general clean up for the Mutator tests (Copy paste errors)

The doTest function is a much simpler version of that used by PHP-CS-Fixer


tests/Mutator/Arithmetic/BitwiseNotTest.php
```diff
@@ -10,26 +10,40 @@
...
-        $mutatedCode = $this->mutate($code);
+        $this->doTest($input, $expected);
+    }
 
-        $expectedMutatedCode = <<<'CODE'
+    public function provideMutationCases(): array
+    {
+        return [
+            [
+                <<<'CODE'
```
theofidry: maybe instead of those `CODE` we could use `PHP`?

tests/Mutator/Arithmetic/BitwiseOrTest.php
```diff
@@ -10,26 +10,40 @@
...
+     */
+    public function test_mutator($input, $expected = null)
     {
-        $code = '<?php 1 | 2;';
-        $mutatedCode = $this->mutate($code);
+        $this->doTest($input, $expected);
+    }
 
-        $expectedMutatedCode = <<<'CODE'
+    public function provideMutationCases(): array
```
theofidry: I find cleaner to return a generator instead:

```php
 public function provideMutationCases(): \Generator
{
    yield [
        <<<'PHP'
...
PHP
    ];
}
```

tests/Mutator/Arithmetic/PlusTest.php
```diff
@@ -10,35 +10,79 @@
...
+$a - +1;
+CODE
+                ,
+            ],
+        ];
+    }
+
+    public function test_it_should_mutate_plus_expression()
+    {
+        $plusExpression = new \PhpParser\Node\Expr\BinaryOp\Plus(new LNumber(1), new LNumber(2));
```
theofidry: IMO we should import at least `Node` but not using the FQCN here

borNfreee:
Thank you @BackEndTea

PR #163, opened 2018-02-07
Allow mutation of full caps true of false

BackEndTea:
Separated from #162, since its not related to that change


borNfreee:
could you please add a test with `False` and `True`?

borNfreee:
ok, just found the test in #162.
Closing this as it will only produce a conflict. Will be merged in #162 

Thank you

PR #165, opened 2018-02-08
Make paths in config file relative to config file

### LINKED ISSUE #142: Paths in config should be relative to config file
Currently the paths in the config file need to be relative to the directory infection was executed from.

I think it would be better to have these paths be relative to the config file. Which is the same as for example phpunit does things.

BackEndTea:
This PR:

* [x] Makes all paths in the config file be relative to the config file.
* [x] Defaults to working directory if no config file is found

Fixes #142

### DIFF
--- a/src/Config/InfectionConfig.php
+++ b/src/Config/InfectionConfig.php
@@ -161,15 +161,10 @@ class InfectionConfig
      */
     private $configLocation;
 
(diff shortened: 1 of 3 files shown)


borNfreee:
Thank you @BackEndTea

Issue #166, opened 2018-02-08
Change to abstract function is never covered


BackEndTea:
Changing anything to an abstract function is never seen as covered, as proven in our own code base, and even if it was considered covered, in php you are free to make an overridden function more accessible. 

So i'd suggest not mutating abstract functions from public to protected, the protected to private mutator already does not mutate abstract functions.

<details>
 <summary>proof</summary>

 ```diff
104) /code/src/Mutator/Mutator.php:15    [M] PublicVisibility

--- Original
+++ New
@@ @@
 {
-    public abstract function mutate(Node $node);
+    protected abstract function mutate(Node $node);
     public abstract function shouldMutate(Node $node) : bool;
     public function getName() : string
     {
         $parts = explode('\\', static::class);
         return $parts[count($parts) - 1];
     }
```
</details>

BackEndTea:
Closed by #169

PR #169, opened 2018-02-09
Fix: Don't mutate abstract methods

BackEndTea:
This PR:

* [x] Stops mutations from `public` to `protected` on abstract functions
* [ ] ~Stops mutations of default values of abstract functions~

See #166 

Note: If #67 gets integrated before the next release, its probably better to do this with reflection.

I excluded stopping mutation of default values in abstract functions for now.

### DIFF
--- a/src/Mutator/FunctionSignature/PublicVisibility.php
+++ b/src/Mutator/FunctionSignature/PublicVisibility.php
@@ -55,6 +55,10 @@ class PublicVisibility extends Mutator
             return false;
         }
 
+        if ($node->isAbstract()) {
+            return false;
+        }
+
         return $node->isPublic();
     }
 
--- a/tests/Mutator/FunctionSignature/PublicVisibilityTest.php
+++ b/tests/Mutator/FunctionSignature/PublicVisibilityTest.php
@@ -215,6 +215,16 @@ interface TestInterface
 }
 PHP
             ],
+            'It does not mutate an abstract function' => [
+                <<<'PHP'
+<?php
+
+abstract class Test
+{
+    public abstract function foo(int $param, $test = 1) : bool;
+}
+PHP
+            ],
         ];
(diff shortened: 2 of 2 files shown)


src/Visitor/MutationsCollectorVisitor.php
```diff
@@ -58,6 +58,9 @@ public function leaveNode(Node $node)
     {
         foreach ($this->mutators as $mutator) {
             $isOnFunctionSignature = $node->getAttribute(WrappedFunctionInfoCollectorVisitor::IS_ON_FUNCTION_SIGNATURE, false);
+            if ($node->getAttribute(WrappedFunctionInfoCollectorVisitor::IS_ON_ABSTRACT_FUNCTION_SIGNATURE, false)) {
```
borNfreee: I would say that checking `$node->isAbstract()` in `PublicVisibility.php` is enough to fix theoriginal issue.

I mean, checking this flag for each node is slows down the things without a benefit since it isn't even used.

Also, after merging https://github.com/infection/infection/pull/67 it would be better to store `[\ReflectionMethod](http://php.net/manual/ru/class.reflectionmethod.php)` that "knows" everything about particular method rather than just flag that it is an abstract one

BackEndTea: The goal of `IS_ON_ABSTRACT_FUNCTION_SIGNATURE` is to stop mutation of default values in abstract function signatures, since those will also never be covered.

For example:
```diff
- abstract public function find(bool $includeArgs = true): string;
+ abstract public function find(bool $includeArgs = false): string;
```

Will never be covered, and is therefore a useless mutation. 

But as i said before, if #67 gets merged this isn't really needed

borNfreee:
Thank you @BackEndTea 👍

PR #170, opened 2018-02-09
Enhacement: native function invocation cs fixer rule

BackEndTea:
This PR:

* [x] Adds the `native_function_invocation` rule
* [x] Runs `composer cs:fix`

>* **native_function_invocation**
Add leading ``\`` before function invocation of internal function to speed up resolving.
*Risky rule: risky when any of the functions are overridden.*
Configuration options:
> - ``exclude`` (``array``): list of functions to ignore; defaults to ``[]``

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -26,6 +26,7 @@ return \PhpCsFixer\Config::create()
             'separate' => 'bottom',
         ],
         'is_null' => true,
+        'native_function_invocation' => true,
         'ordered_imports' => true,
         'phpdoc_align' => false,
         'phpdoc_summary' => false,
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -14,8 +14,8 @@ $files = [
 ];
 
 foreach ($files as $file) {
-    if (file_exists($file)) {
-        define('INFECTION_COMPOSER_INSTALL', $file);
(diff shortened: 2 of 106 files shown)


sidz:
No a lot of sense to prefix all php functions. Here is the list of functions where it may be useful:
https://github.com/php/php-src/blob/f2db305fa4e9bd7d04d567822687ec714aedcdb5/Zend/zend_compile.c#L3872

localheinz:
Also see https://wiki.php.net/rfc/fallback-to-root-scope-deprecation.

borNfreee:
> No a lot of sense to prefix all php functions.

as far as I understand from [Nikic's comment](https://github.com/symfony/symfony/pull/25854#issuecomment-359185316), it's not useless  to prefix all functions

theofidry:
👎  for me on that: it just makes the code way harder to read than it should.

IMO a more proper fix would be to import those functions/constants with use statements which for opcache or performance optimisations makes it strictly equivalent.

Now I'm aware there is no CS rule for this, but I think we should either wait or help for it, the perf. improvement we can get for this is too negligible for it right now at least.

Also note that this perf. improvement is only for non-PHAR usage as inside the PHAR PHP-Scoper will take care of this.

Issue #167, opened 2018-02-09
Phpunit "excludes" are not properly parsed


acelaya:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.7.1
| Test Framework version | PHPUnit 7.0.0
| PHP version | 7.1.0
| Platform    | Ubuntu
| Github Repo | -

Apparently, infection is not properly parsing the exlcudes from the original phpunit.xml(.dist) configuration file.

My phpunit.xml.dist file, includes this `filter` block.

```xml
<filter>
    <whitelist processUncoveredFilesFromWhitelist="true">
        <directory suffix=".php">./module/*/src</directory>

        <exclude>
            <directory suffix="Repository.php">./module/*/src/**</directory>
            <directory suffix="RepositoryInterface.php">./module/*/src/**</directory>
            <directory suffix=".php">./module/Api/src/Model</directory>
        </exclude>
    </whitelist>
</filter>
```

For some reason, the config file which infection generates is misinterpreting the first two excludes, which results in almost the whole project being excluded and therefore, the code coverage being almost 0%.

If I remove the first two excludes and leave only the third one, then everything works as expected.

acelaya:
I can see it is expanding the glob expression ` ./module/*/src/**`, which results in a `directory` entry for every actual directory.

However, those entries do not have the original `suffix` attribute (either with value Repository.php or RepositoryInterface.php), which is what causes everything to be excluded.

sidz:
@acelaya thank you. I hope https://github.com/infection/infection/pull/171 will solve this issue

acelaya:
Perfect. Thanks :)

sidz:
@acelaya PR has been merged. So could you please check Infection from master on your project and let us know if anything won't go well

acelaya:
Just tested it using dev-master, and I can confirm it works now.

Thanks! :-)

PR #171, opened 2018-02-10
Don't try to expand directories with `*` and `**`

sidz:
Stop expanding paths with `*`. Right now if phpunit.xm contains smth like:
```
<filter>
        <whitelist processUncoveredFilesFromWhitelist="true">
            <directory suffix=".php">./src</directory>

            <exclude>
                <directory>./src/*/*Bundle` </directory>
            </exclude>
        </whitelist>
</filter>
```
Infection will try to find all paths which are matching by this pattern and will expand it to the:
```
<filter>
        <whitelist processUncoveredFilesFromWhitelist="true">
            <directory suffix=".php">absoluteProjectPath/src</directory>

            <exclude>
                <directory>absoluteProjectPath/src/Dir1/AcmeBundle</directory>
                <directory>absoluteProjectPath/src/Dir2/LolBundle</directory>
            </exclude>
        </whitelist>
</filter>
```
I think we shouldn't care about it on the Infection side as PhpUnit will take care about it. So we still need to modify all relative paths to the absolute path.

should fix https://github.com/infection/infection/issues/167

@borNfreee if you have any reason to expand paths with `*` please give me shout.


borNfreee:
Thank you @sidz for fixing this issue 👍

sidz:
np @borNfreee when you will get first donate please give me 50% 😆

borNfreee:
LOL, deal!

so now it's time to be a hipster and create a [Patreon](https://www.patreon.com/) account

PR #172, opened 2018-02-11
Enhancement: More tests

BackEndTea:
This PR: 

* [x] Adds, improves & fixes tests
* [x] Adds 2 small changes the PublicVisibility and Plus Mutators
* Plus: No need to check if both values are arrays, if one is, the otherone should be, or there is an error, which means that mutation is useless.
* PublicVisibility: No need to make a strict in_array check, since the input is forced to be a string, see also https://github.com/BackEndTea/Benchmarks

And on a fun note, if this and #169 get merged, we have 100% infection coverage in the Mutator folder


src/Mutator/FunctionSignature/PublicVisibility.php
```diff
@@ -73,8 +73,7 @@ private function isBlacklistedFunction(string $name): bool
                 '__unset',
                 '__toString',
                 '__debugInfo',
-            ],
-            true
```
borNfreee: not sure about this one. I guess it is a good practice to use the third strict param here. 

IMO it's better to have escaped mutant rather than not using strictness.

What do you think?

PS in the future, when we implement profiles (or something like that), we can instruct infection not to mutate true/false strict param in native functions

tests/Config/Guesser/PhpUnitPathGuesserTest.php
```diff
@@ -13,9 +13,34 @@
...
-        $composerJson = <<<'CODE'
+        $guesser = new PhpUnitPathGuesser(json_decode($composerJson));
+
+        $this->assertSame($directory, $guesser->guess());
+    }
+
+    public function providesJsonComposerAndLocations(): \Generator
+    {
+        yield [
+            <<<'CODE'
```
theofidry: `'JSON'`? Often IDEs can pick it up for syntax highlighting/check

borNfreee:
Thank you @BackEndTea

Issue #137, opened 2018-01-23
Restart with xdebug


theofidry:
When we start infection without xdebug and phpdbg, you get:

```
You need to use phpdbg or install and enable xdebug in order to allow for code coverage generation.
```

1) I think the message could receive some pretty formatting with `SymfonyStyle::error()` to make it easier to read.

2) Enabling xdebug is usually easy or at least people are familiar with it. For phpdbg however people are likely not so I think adding the command on how to run it with phpdbg would be more user-friendly (even though it's already in the docs)

3) Which was actually the reason why I wanted to open this issue: as when xdebug is enabled we disable it later on for perf improvements, WDYT of enabling it if neither phpdbg is found and xdebug enabled instead of the error message above? We could then add an information "running with xdebug blabla" when running infection. If the user is not happy with it he can always cancel it and restart with phpdbg.

IMO that would likely be a smoother usage with Infection. I for one often enable/disable (it's dead easy with my shortcuts) for several reasons (perf, actual tests, etc.) so doing that would just works all the time instead.

thoughts?

sidz:
@theofidry here is my thoughts:

1. We can do this 👍 
2. The main reason why we chose this way: 
We can't guaranty that user has (installed) `xdebug` and can't even check if it was installed or not. Two potential cases here:
- extension was installed before but right now it's disabled. This situation is fine and we can enable xdebug on fly.
- extension is not installed. We can't check this on fly and with this situation if we will enable `xdebug.ini` on fly we will get error message (don't remember full message but there will be smth like "Can not load extension blablabla...") from PHP side and it will break infection at all.
Also instead of tell user that he doesn't have `xdebug` it will work while we will need to execute code coverage process.

This doesn't have any affect on `phpdbg` as to use it user will need to execute different command (`phpdbg -qr` instead `php bin/infection`).

3. this thought is good but we can't guaranty that xdebug installed.

theofidry:
Regarding 3 I think that’s fine as long Asnières we bail out with a IX
friendly’ message

borNfreee:
> Regarding 3 I think that’s fine as long Asnières we bail out with a IX
> friendly’ message

could you please clarify what does it mean?

theofidry:
Sorry too much autocorrect from the phone and no double check :P

PR #173, opened 2018-02-11
Pretty formatting for error message when xdebug and phpdbg are not used

sidz:
This PR is bringing pretty formatting for the error message (which @theofidry suggested to do in the https://github.com/infection/infection/issues/137) when phpdbg and xdebug are not used.

Also we will add notes when Infection is running via phpdbg or xdebug.

Note for phpdbg (the same will be for xdebug).
Image with phpdbg

When phpdbg and xdebug are not enabled
Image

### DIFF
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -10,12 +10,17 @@ namespace Infection\Console;
 
 use Infection\Command;
 use Infection\Config\InfectionConfig;
+use Infection\Php\ConfigBuilder;
+use Infection\Php\XdebugHandler;
 use Pimple\Container;
 use Symfony\Component\Console\Application as BaseApplication;
 use Symfony\Component\Console\Formatter\OutputFormatterStyle;
+use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\ConsoleOutput;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Console\Style\SymfonyStyle;
 
 class Application extends BaseApplication
 {
@@ -35,11 +40,58 @@ ASCII;
      */
     private $container;
 
+    /**
+     * @var SymfonyStyle
+     */
+    private $io;
+
+    /**
+     * @var bool
+     */
+    private $isXdebugLoaded;
+
+    /**
+     * @var bool
+     */
+    private $isDebuggerDisabled;
+
     public function __construct(Container $container, string $name = self::NAME, string $version = self::VERSION)
     {
         $this->container = $container;
+        $this->isDebuggerDisabled = empty((string) getenv(XdebugHandler::ENV_DISABLE_XDEBUG));
 
         parent::__construct($name, $version);
+    }
+
+    public function run(InputInterface $input = null, OutputInterface $output = null)
+    {
+        if (null === $input) {
+            $input = new ArgvInput();
+        }
+
+        if (null === $output) {
+            $output = new ConsoleOutput();
+        }
+
+        $this->io = new SymfonyStyle($input, $output);
+
+        if (PHP_SAPI === 'phpdbg') {
+            $this->io->note('You are running Infection with phpdbg enabled.');
+        } elseif ($this->isXdebugLoaded) {
+            $this->io->note('You are running Infection with xdebug enabled.');
+        }
+
+        $xdebug = new XdebugHandler(new ConfigBuilder(sys_get_temp_dir()));
+        $xdebug->check();
+        unset($xdebug);
+
+        if (PHP_SAPI !== 'phpdbg' && $this->isDebuggerDisabled && !\extension_loaded('xdebug')) {
+            $this->io->error('You need to use phpdbg or install and enable xdebug in order to allow for code coverage generation.');
+
+            die(1);
+        }
+
+        return parent::run($input, $output);
     }
 
     public function doRun(InputInterface $input, OutputInterface $output)
@@ -87,6 +139,11 @@ ASCII;
     public function getContainer(): Container
     {
         return $this->container;
+    }
+
+    public function getIO(): SymfonyStyle
+    {
+        return $this->io;
     }
 
     private function buildDynamicDependencies(InputInterface $input)
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -78,16 +78,19 @@ ASCII;
         $this->io = new SymfonyStyle($input, $output);
 
         if (PHP_SAPI === 'phpdbg') {
-            $this->io->note(sprintf(self::RUNNING_WITH_NOTE, PHP_SAPI));
+            $this->io->writeln(sprintf(self::RUNNING_WITH_NOTE, PHP_SAPI));
         } elseif ($this->isXdebugLoaded) {
-            $this->io->note(sprintf(self::RUNNING_WITH_NOTE, 'xdebug'));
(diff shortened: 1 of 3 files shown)


src/Console/Application.php
```diff
@@ -34,13 +40,60 @@ class Application extends BaseApplication
...
+
+    /**
+     * @var bool
+     */
+    private $isDebuggerDisabled;
+
     public function __construct(Container $container, string $name = self::NAME, string $version = self::VERSION)
     {
         $this->container = $container;
+        $this->isDebuggerDisabled = empty((string) getenv(XdebugHandler::ENV_DISABLE_XDEBUG));
```
theofidry: `'' !== trim((string) getenv(XdebugHandler::ENV_DISABLE_XDEBUG))`; Not strictly necessary but personally I prefer to avoid any usage of `empty()` which is gun footing

src/Console/Application.php
```diff
@@ -34,13 +40,60 @@ class Application extends BaseApplication
...
+        }
+
+        $xdebug = new XdebugHandler(new ConfigBuilder(sys_get_temp_dir()));
+        $xdebug->check();
+        unset($xdebug);
+
+        if (PHP_SAPI !== 'phpdbg' && $this->isDebuggerDisabled && !\extension_loaded('xdebug')) {
+            $this->io->error('You need to use phpdbg or install and enable xdebug in order to allow for code coverage generation.');
+
+            die(1);
```
theofidry: `return 1`  would be enough?

src/Console/Application.php
```diff
@@ -34,13 +40,60 @@ class Application extends BaseApplication
...
+
+        if (PHP_SAPI === 'phpdbg') {
+            $this->io->note(sprintf(self::RUNNING_WITH_NOTE, PHP_SAPI));
+        } elseif ($this->isXdebugLoaded) {
+            $this->io->note(sprintf(self::RUNNING_WITH_NOTE, 'xdebug'));
+        }
+
+        $xdebug = new XdebugHandler(new ConfigBuilder(sys_get_temp_dir()));
+        $xdebug->check();
+        unset($xdebug);
```
theofidry: why do you need to unset it?

src/Console/Application.php
```diff
@@ -88,6 +141,11 @@ public function getContainer(): Container
         return $this->container;
     }
 
+    public function getIO(): SymfonyStyle
```
theofidry: I'm not convinced this is very useful here, a `$io` variable looks like to be enough for the case above

src/Console/Application.php
```diff
@@ -34,13 +40,60 @@ class Application extends BaseApplication
...
+            $input = new ArgvInput();
+        }
+
+        if (null === $output) {
+            $output = new ConsoleOutput();
+        }
+
+        $this->io = new SymfonyStyle($input, $output);
+
+        if (PHP_SAPI === 'phpdbg') {
```
borNfreee: Could you please explain why do we need to show these messages? It looks like a warning, but it's not clear what to do with it for the final user.

sidz: it's implementation of the @theofidry's third proposal.

```
3. Which was actually the reason why I wanted to open this issue: as when xdebug is enabled we disable it later on for perf improvements, WDYT of enabling it if neither phpdbg is found and xdebug enabled instead of the error message above? We could then add an information "running with xdebug blabla" when running infection. If the user is not happy with it he can always cancel it and restart with phpdbg.
```

borNfreee: I updated the comment with the following

> As I understand from Theo's comment, it needs to be displayed when user has it disabled, but infection enables it automatically

src/Console/Application.php
sidz: SymfonyStyle is using in the InfectionCommand so it's preventing of creating 2 instances of the same object

src/Console/Application.php
sidz: @theofidry seems like we have a misunderstanding here. What did you mean ? 😆

src/Console/Application.php
```diff
@@ -34,13 +40,60 @@ class Application extends BaseApplication
...
+            $this->io->note(sprintf(self::RUNNING_WITH_NOTE, PHP_SAPI));
+        } elseif ($this->isXdebugLoaded) {
```
borNfreee: I don't quite understand how it works. Just checked - it always `null` and is not used/set anywhere. Could you please explain?

src/Console/Application.php
sidz: I hope it'll free memory. to be honest I've just moved it from the bootstrap.php as is

src/Console/Application.php
borNfreee: I got it like this: "if script is run without phpdbg and xdebug is *not* enabled - enable it automatically and show the message that it was enabled by Infection"

src/Console/Application.php
borNfreee: since this is a local var in the class method, I suppose it would be garbage collected automatically

src/Console/Application.php
sidz: good catch! Shame on me I pushed not final version ^_^

src/Console/Application.php
sidz: if so my comment from the issue is fully described why we can't do this and we can remove all these lines

borNfreee: @theofidry ?

theofidry: Right, we cannot enable xdebug or phpdbg for the user so instead we're sticking with bailing out with a pretty message to the user.

So my comment was:
- Adding a note to know which of xdebug or phpdbg is used
- Add a friendly note when bailing out if neither xdebug and phpdbg are used

Both are already done, @sidz made it more pretty if I understood that right.

borNfreee: Seems like I miss the point. It is still not clear for me why this orange (like warning) message is displayed and what value does it bring.

`! [NOTE] You are running Infection with phpdbg enabled.`

* If I have xdebug enabled globally - it's obvious that infection is run with Xdebug as well. 
* If I run `phpdbg -qrrr bin/infection ...` - it's obvious that it is run with phpdbg enabled.

-------

When you run let's say PHPCSFixer with Xdebug, it says "Xdebug has a major performance impact..." and it's clear that user should disable it. But what the use should do with the message above?

theofidry: Then maybe it should be displayed as plain text to not be interpreted as a warning. The idea is just to write this info somewhere.

It's not always obvious what Infection is running with especially if it's hidden behind an alias. For example in case of box running Infection is done with `make infection`, you don't pick xdebug or phpdbg neither do I know if xdebug is enabled or not (I have to enable/disable it very often so it's not always clear).

Also for the error message, maybe we should point to the doc or replicate with the doc how to run it with phpdbg:

```
[ERROR] Neither phpdbg or xdebug has been found. One of those is required by Infection in order to generate coverage data. Either:
- Enable xdebug and run infection again
- Use phpdbg: phpdbg -qrr infection [...]
```

borNfreee: Got it. My vote is for changing the colour to something more positive ;)

> Also for the error message, ...

👍

sidz: done.

p.s. maybe we should show these notes only under verbose mode? thoughts?
p.p.s. You know I start think that this message is really useless :)

borNfreee: > p.s. maybe we should show these notes only under verbose mode

like it

sidz:
@theofidry could you please re-review when you'll have a chance?

theofidry:
Will do tomorrow!

borNfreee:
Thank you @sidz

PR #175, opened 2018-02-11
Delete Result class

BackEndTea:
I couldn't find any instances of this class being used.


borNfreee:
Thank you @BackEndTea

BackEndTea:
Thanks you @borNfreee

Issue #94, opened 2017-12-02
Use existing coverage report


theofidry:
Unless I missed it, right now are running the tests which requires xdebug or phpdbg to get the coverage data to then be able to do our own stuff.

I think a typical user would already have tests with coverage somehow, so IMO it would make much more sense to provide the data coverage files which when done so would avoid us to have to run the tests once.

borNfreee:
Do you suggest to have an option for providing code coverage data for Infection?

I'm not sure skipping the "Initial test suite run" step is a good idea. 

1. We must run the whole test suite to make sure all tests are in a passing state. Without it, we can't guarantee the correctness of MSI.
2. When user changes the code and/or test, we must to re-generate code coverage.

I'm not sure I got your right though.

theofidry:
I think we're just describing different workflows.

Your workflow fits very well in the case where you run your tests without coverage for quick feedback and then run Infection to get a detailed report.

Another workflow would be to run the tests with coverage and then Infection for the report.

In the second case, running the tests again with coverage is a waste of time. And a use case for that second case is for example if you have to do a weird workaround for running the tests with coverage, if you want to display multiple coverage reports (XML + Text for example) or send the coverage report to another tool.

That said I don't think feature is a high priority one.

luispabon:
Currently, we're running our tests with coverage twice at CI, one with phpunit to get the reports, and another one with infection. 

Each step takes a few minutes as my current project has a number of end to end (eg drilling through the whole symfony stack) tests that are very slow when xdebug and coverage are enabled. Non coverage phpunit runs finish within the second.

If infection could take a coverage report file from a previous phpunit run it would cut our build times down substantially. Infection's initial test run would still run, but without coverage.

Infection could then print a warning that using an external coverage file could give misleading results if stale. It would be the developer's responsibility to ensure that doesn't happen.

luispabon:
An alternative way that would help in my use case would be for infection to honor coverage report generation settings on phpunit.xml including location. This would perhaps be cleaner and safer.

ryan7273:
This would cut down on my build times as well. I like the suggestion of using the `phpunit.xml` settings as it would allow Infection to be more easily integrated into a full CI setup without adding redundancy.

borNfreee:
I see your point guys.

[...]

PR #176, opened 2018-02-11
Use existing coverage reports

borNfreee:
Fixes https://github.com/infection/infection/issues/94
- [x] Allow to provide the path to folder with existing coverage report(s)
- [x] Disable Xdebug for Initial tests run if existing coverage provided
- [x] Documentation (https://github.com/infection/site/pull/20)
- [x] Reuse coverage on Travis to speed up builds

### DIFF
--- a/app/container.php
+++ b/app/container.php
@@ -31,6 +31,7 @@ use PhpParser\PrettyPrinter\Standard;
 use Pimple\Container;
 use SebastianBergmann\Diff\Differ as BaseDiffer;
 use Symfony\Component\Filesystem\Filesystem;
+use Infection\TestFramework\Coverage\CoverageDoesNotExistException;
 
 $c = new Container();
 
@@ -69,6 +70,12 @@ $c['coverage.dir.phpspec'] = function (Container $c) {
 };
 
 $c['phpunit.junit.file.path'] = function (Container $c) {
+    $filePath = sprintf('%s/%s', $c['coverage.path'], PhpUnitAdapter::JUNIT_FILE_NAME);
+
+    if (!file_exists($filePath)) {
+        throw CoverageDoesNotExistException::forJunit($filePath);
+    }
+
     return sprintf('%s/%s', $c['coverage.path'], PhpUnitAdapter::JUNIT_FILE_NAME);
 };
 
--- a/app/container.php
+++ b/app/container.php
@@ -31,7 +31,6 @@ use PhpParser\PrettyPrinter\Standard;
 use Pimple\Container;
 use SebastianBergmann\Diff\Differ as BaseDiffer;
 use Symfony\Component\Filesystem\Filesystem;
-use Infection\TestFramework\Coverage\CoverageDoesNotExistException;
 
 $c = new Container();
 
@@ -70,12 +69,6 @@ $c['coverage.dir.phpspec'] = function (Container $c) {
 };
 
 $c['phpunit.junit.file.path'] = function (Container $c) {
-    $filePath = sprintf('%s/%s', $c['coverage.path'], PhpUnitAdapter::JUNIT_FILE_NAME);
-
-    if (!file_exists($filePath)) {
-        throw CoverageDoesNotExistException::forJunit($filePath);
-    }
-
     return sprintf('%s/%s', $c['coverage.path'], PhpUnitAdapter::JUNIT_FILE_NAME);
 };
 
--- a/app/container.php
+++ b/app/container.php
@@ -56,12 +56,20 @@ $c['tmp.dir.creator'] = function (Container $c): TmpDirectoryCreator {
     return new TmpDirectoryCreator($c['filesystem']);
 };
 
-$c['coverage.dir.phpunit'] = function (Container $c): string {
-    return $c['tmp.dir'] . '/' . CodeCoverageData::PHP_UNIT_COVERAGE_DIR;
+$c['tmp.dir'] = function (Container $c): string {
+    return $c['tmp.dir.creator']->createAndGet($c['infection.config']->getTmpDir());
 };
 
-$c['coverage.dir.phpspec'] = function (Container $c): string {
-    return $c['tmp.dir'] . '/' . CodeCoverageData::PHP_SPEC_COVERAGE_DIR;
+$c['coverage.dir.phpunit'] = function (Container $c) {
+    return sprintf('%s/%s', $c['coverage.path'], CodeCoverageData::PHP_UNIT_COVERAGE_DIR);
+};
+
+$c['coverage.dir.phpspec'] = function (Container $c) {
+    return sprintf('%s/%s', $c['coverage.path'], CodeCoverageData::PHP_SPEC_COVERAGE_DIR);
+};
+
+$c['phpunit.junit.file.path'] = function (Container $c) {
+    return sprintf('%s/%s', $c['coverage.path'], PhpUnitAdapter::JUNIT_FILE_NAME);
 };
 
 $c['locator'] = function (Container $c): Locator {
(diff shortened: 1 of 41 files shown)


BackEndTea:
Could you add this option to travis? should save us some time during the build

src/TestFramework/Factory.php
```diff
@@ -27,7 +27,7 @@
     /**
      * @var string
      */
-    private $tempDir;
+    private $tmpDir;
```
BackEndTea: I like that we are making this consistent all over the code base, but the diff of this PR is already quite big, so maybe this should be moved to a separate PR. That would make it a bit easier to see exactly what changed.

borNfreee: I will spend more time creating separate PR for these 2 files than review requires :) leaving it here

borNfreee:
> Could you add this option to travis? should save us some time during the build

done

app/container.php
```diff
@@ -56,12 +56,20 @@
...
 
-$c['coverage.dir.phpunit'] = function (Container $c): string {
-    return $c['tmp.dir'] . '/' . CodeCoverageData::PHP_UNIT_COVERAGE_DIR;
+$c['tmp.dir'] = function (Container $c): string {
+    return $c['tmp.dir.creator']->createAndGet($c['infection.config']->getTmpDir());
 };
 
-$c['coverage.dir.phpspec'] = function (Container $c): string {
-    return $c['tmp.dir'] . '/' . CodeCoverageData::PHP_SPEC_COVERAGE_DIR;
+$c['coverage.dir.phpunit'] = function (Container $c) {
```
sidz: is this service using elsewhere? can't find any usage

p.s. the same for phpspec

src/Command/InfectionCommand.php
```diff
@@ -103,6 +104,13 @@ protected function configure()
                 InputOption::VALUE_REQUIRED,
                 'Custom configuration file path'
             )
+            ->addOption(
+                'coverage',
+                null,
+                InputOption::VALUE_REQUIRED,
```
sidz: I'd say that this option shouldn't be mandatory. Can we do it optional?

app/container.php
borNfreee: https://github.com/infection/infection/blob/master/src/Command/InfectionCommand.php#L278

src/Console/Application.php
```diff
@@ -106,18 +107,38 @@ private function buildDynamicDependencies(InputInterface $input)
...
+    private function getExistingCoveragePath(InputInterface $input, string $configLocation): string
+    {
+        $existingCoveragePath = $input->getOption('coverage');
+
+        if ($existingCoveragePath === '') {
+            return '';
+        }
+
+        /** @var Filesystem $fileSystem */
+        $fileSystem = $this->container['filesystem'];
```
sidz: can we add filesystem as argument?

src/TestFramework/PhpUnit/Coverage/CoverageXmlParser.php
```diff
@@ -134,7 +134,8 @@ private function getSourceFilePath(\DOMXPath $xPath, string $relativeCoverageFil
             return $realPath;
         }
 
-        throw new \Exception('Source file was not found');
+        throw new \Exception(
```
sidz: can we create smth like CoverageException? or put this message into CoverageDoesNotExistException?

src/Command/InfectionCommand.php
borNfreee: This parameter means that the value of the option is required, not for the option itself.

So `--coverage=dir` is allowed, but `--coverage` is not

src/Console/Application.php
```diff
@@ -106,18 +107,38 @@ private function buildDynamicDependencies(InputInterface $input)
...
+
+            return $existingCoveragePath ?: $c['tmp.dir'];
         };
     }
+
+    private function getExistingCoveragePath(InputInterface $input, string $configLocation): string
+    {
+        $existingCoveragePath = $input->getOption('coverage');
+
+        if ($existingCoveragePath === '') {
```
sidz: do we need to trim this value?

src/TestFramework/PhpSpec/Config/InitialYamlConfiguration.php
```diff
@@ -8,23 +8,68 @@
...
 use Symfony\Component\Yaml\Yaml;
 
 class InitialYamlConfiguration extends AbstractYamlConfiguration
 {
+    /**
+     * @var bool
+     */
+    private $skipCoverage;
+
+    public function __construct(string $tempDirectory, array $parsedYaml, bool $skipCoverage)
```
sidz: tmpDir?

src/Command/InfectionCommand.php
```diff
@@ -103,6 +104,13 @@ protected function configure()
                 InputOption::VALUE_REQUIRED,
                 'Custom configuration file path'
             )
+            ->addOption(
+                'coverage',
```
sidz: if it's path maybe we should call it `coverage-path` ?

borNfreee: I thought the same but made it consistent with the `--configuration` options https://infection.github.io/guide/command-line-options.html#configuration-or-c which also does not have `path` postfix

BackEndTea:
This may be outside of the scope of this PR, but if someone uses the --coverage option, there shouldn't be any need for them to have xdebug enabled, or run the command through phpdbg

For exmple, lets say i use phpdbg to get the coverage of my project, and i don't have xdebug installed.

I would then like to be able to run infection without phpdbg, since that slows it down. And as far as i know, once we have the original coverage, there is no need for xdebug or phpdbg.

src/Console/Application.php
```diff
@@ -106,18 +107,39 @@ private function buildDynamicDependencies(InputInterface $input)
...
-            return new InfectionConfig(json_decode($json), $c['filesystem'], $dir);
+            return new InfectionConfig(json_decode($json), $c['filesystem'], $configLocation);
         };
 
-        $this->container['tmp.dir'] = function (Container $c): string {
-            return $c['tmp.dir.creator']->createAndGet($c['infection.config']->getTmpDir());
+        $this->container['coverage.path'] = function (Container $c) use ($input): string {
+            $existingCoveragePath = $this->getExistingCoveragePath(
+                $input,
+                $c['infection.config']->getPhpUnitConfigDir(),
```
BackEndTea: Since the --coverage option is a command line option, its location should be relative to where the command is ran from, not the phpunit config location

src/Process/Builder/ProcessBuilder.php
```diff
@@ -36,12 +36,13 @@ public function __construct(AbstractTestFrameworkAdapter $testFrameworkAdapter,
      * Creates process with enabled debugger as test framework is going to use in the code coverage.
      *
      * @param string $testFrameworkExtraOptions
+     * @param bool $skipCoverage
```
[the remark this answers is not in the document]
borNfreee: we don't use useless PHPDoc `@param` since PHP7 already provides information *unless* there is a string, describing the method purpose as in this case `Creates process with enabled debugger as test framework is going to use in the code coverage.`

Without `@param` here, my IDE complains and underlines the method with warnings. I think this is a pretty common practice, and in this repository as well.

borNfreee:
> there shouldn't be any need for them to have xdebug enabled, or run the command through phpdbg

As I understand, it will be easier to do it when https://github.com/infection/infection/pull/173 gets merged since we will have information about passed options in `$input` object. 

@BackEndTea is there anything else to be addressed here?

sidz:
@borNfreee unfortunately we don't have necessary information about passed options in the Application.
At least we can't do `$input->getOption` because parameters binding happens a bit later.

borNfreee:
Thank you @sidz & @BackEndTea

Issue #79, opened 2017-11-18
Disable certain mutators for certain lines


GaryJones:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.6.1
| Test Framework version | PHPUnit 6.4.4
| PHP version | 7.1.7
| Platform    | MacOS
| Github Repo | -

For WordPress, callback methods hooked into the actions or filters need to be `public`.

The `Function Signature` mutator naturally tries to change the callback from `public` to `protected`, and flags it as a "not covered mutant".

Is there any way (to reduce known false positives) to specify that a method really should stay `public`, either by adding a line comment (e.g. `// infection:disable FunctionSignature`) or by adding an exclusion (file and line number? class and method name?) to `infection.json.dist`?

GaryJones:
Perhaps an obvious workaround - make sure you have a unit test that calls the `public` method. When this mutates to `protected`, the test will no longer be reported as "not covered".

borNfreee:
Hello,

> Disable certain mutators

This is really my one of the most wanted feature. Let's start discussing a possible implementation.

The first thought was the same - add something like other tools add in PHP and JS world:

```js
// tslint: disable-next-line
...
// infection:disable FunctionSignature
```

but now I don't like this idea because 

1. pollutes the code with 3rd-party tools related comments 
2. these comments can be outdated if a tool has been removed from particular project/not used anymore.

I like the idea of PHPStan -https://github.com/phpstan/phpstan#ignore-error-messages-with-regular-expresions.

They use patterns/regular expressions on the *config level* to skip/ignore some issues during analyzing.

We can use this idea and implement something similar, for example:

#### 1. Ignore mutators for some class

```js
// infection.json
{
    "mutators": {
        "ignore": {
             "PublicVisibility": [
                  "Ingore\For\Particular\Class",
                  "Ingore\For\Another\Class::method",
                  //  ...
                  "Ingore\For\**\*\Glob\Pattern\Or\Namespace"
             ]
        }
    }
}
```

Also, I think keeping the exact line number in the "ignore config" is a bad idea because line numbers change quite often

What I also like and want to have from PHPStan is the behaviour of notifying users when some of the ignored classes/methods for mutants are no longer exist (see https://github.com/phpstan/phpstan#ignore-error-messages-with-regular-expresions): for example if the `Some\Class::method` is ignored for a set of Mutation Operators and after some time is removed - Infection should notify develop that this method is no longer exists and the config should be updated

Issue #89, opened 2017-11-24
Define a level of strictness


theofidry:
One of the greatest point of mutation testing is that it provides a reliable way to know which part of your application is absolutely tested. A bad point is that it's horribly slow as it's a brute force technique. An easy way to get better performances is to simply play with the number of mutators applied. Indeed applying only 10 mutators is gonna result in a much faster execution time than 100 mutators.

IMO there is a lot of mutators that may not be useful depending of the context in which Infection is used. For example I could care less about the visibility mutator in my application and find it much more useful in my library.

Point is, I think it would be a good idea to provide a way for the user to define which mutators can be applied in the application and maybe providing different default sets. For example we could call that parameter `level` and depending of the value, we apply more or less mutators like for example PhpStan does with their analysis rules.

BackEndTea:
I was thinking about this a little bit, and i think a better way may be to provide `profiles`.

So you would have a `library` profile, which includes the `Visibility` mutants, and an `application` profile which does not.

This could also allow us to have Mutants that are turned off by default, and have some kind of 'I got a 30 minute lunch break' profile which would include all mutants, even the ones that are most likely duplicates, (Like those mentioned in #114)

An even better version of this is to have some way for others to define custom profiles, which they could re-use across projects, or even distribute themselves. Which is similar to how [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) does things.

theofidry:
Maybe `profile` would be a better terminology than `level` here 👍  A thing I like about it (`profile`) is that unlike `level`, we don't have to make a choice about what is stricter than another, so I think it will be easier to define

BackEndTea:
I have been playing around with profiles a bit, what is your opinion on a config like this:

```json
{
  "timeout": 25,
  "source": {
    "directories": [
      "src"
    ],
  },
  "mutators": {
    "@library": true,
    "Plus": false,
    "PublicVisibility": true,
    "TrueValue": false
  }
}
```

This would entail you use the `library` profile, but exclude the `Plus` & `TrueValue` mutators, but include the `PublicVisibility`.

Doing this would also allow us to add configs to mutators at a later stadium

theofidry:
that looks good to me 👍 The only thing is you would need to make sure that if there's a mutator in `@library` which is disabled later, it is actually disabled. For example if `@library` includes `TrueValue`, then with your example above `TrueValue` would be disabled

borNfreee:
[...]

PR #177, opened 2018-02-11
Feature:disable certain mutators

BackEndTea:
This PR is a first shot at disabling mutators

* [ ] ~~Disable based on file path (relative to config location)~~
* [x] Disable Based on class names/ namespaces
* [x] Disable for certain methods
* [x] Disable classes/namespaces with patterns ( eg. `App/Models/*`)
* [ ] Give a warning when disabled item is no longer available (file got deleted, method removed etc)

Fixes: #79 
Related: https://github.com/infection/infection/issues/89, https://github.com/infection/infection/issues/206


BackEndTea:
@infection/core I'd like some feedback on this.

I have dropped ignoring on filename for now, since i think it would be confusing to use both filenames and namespaces.

And it does not yet give a warnign if the disabled item is no longer available, as i'm not sure how to implement it.

src/Mutant/Generator/MutationsGenerator.php
```diff
@@ -125,6 +133,17 @@ private function getMutationsFromFile(SplFileInfo $file, bool $onlyCovered): arr
 
         $traverser = new NodeTraverser();
         $mutators = $this->getMutators();
+        foreach ($mutators as $mutator) {
```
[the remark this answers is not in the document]
borNfreee: Seems like we had (before this PR) a logic error here and now it's got worse: for each file, for each mutator - new `MutatorConfig` is created.

It means if we have 100 files, `$mutators = $this->getMutators();` will return 100 times *the same result* and 100 times `new MutatorConfig` will be created for particular mutator.

I suggest 

1. move `$mutators = $this->getMutators();` outside/before the loop `foreach ($files as $file) {`  in `generate()` method. It will allow calling this method once instead of X times, where X is the number of mutated files

2. since mutators will be outside the loop (no in `getMutationsFromFile` method), `MutatorConfig` also will be created only once for each mutator instead of X times

src/Mutator/MutatorConfig.php
```diff
@@ -0,0 +1,42 @@
...
+     * @var array
+     */
+    private $ignoreConfig;
+
+    public function __construct(array $config)
+    {
+        $this->ignoreConfig = $config['ignore'] ?? [];
+    }
+
+    public function isIgnored(string $class, string $method): bool
```
borNfreee: Seems like glob patter is not handled

src/Visitor/MutationsCollectorVisitor.php
```diff
@@ -66,6 +71,13 @@ public function leaveNode(Node $node)
                 }
             }
 
+            if ($mutator->getConfig()->isIgnored(
```
borNfreee: 1. Since `$mutator->getConfig()` must return an instance of `MutantConfig`, it's better to create Mutator with constructot injection: `new Mutator(new MutatorConfig())'`
Thus, it won't be possible to have invalid object.

2. Probably, we shouldn't expose `MutatorConfig` to client code. I mean, we can check whether Mutator is ignored or not like this

```php
class Mutator 
{
     public function __construct(MutatorConfig $mutatorConfig)
     {
          ...
     }

     public function isIgnored(): bool
     {
         return $this->config->isIgnored();
     }
}
```

so `$mutator->isIgnored()` instead of `$mutator->getConfig()->isIgnored()`

tests/Fixtures/e2e/IgnoredMutators/infection.json
```diff
@@ -0,0 +1,24 @@
...
+    "timeout": 25,
+    "source": {
+        "directories": [
+            "src"
+        ]
+    },
+    "logs": {
+        "summary": "infection-log.txt"
+    },
+    "mutators": {
```
borNfreee: Could you please add a documentation to for this feature?

BackEndTea:
@borNfreee I applied most of your feedback, could look at this again?

src/Mutator/MutatorConfig.php
```diff
@@ -0,0 +1,42 @@
...
+namespace Infection\Mutator;
+
+class MutatorConfig
+{
+    /**
+     * @var array
+     */
+    private $ignoreConfig;
+
+    public function __construct(array $config)
```
borNfreee: how will `array` type declaration work with the following config:

```json
{
    "mutators": {
        "Plus": false
    }
}
```
?

tests/Config/InfectionConfigTest.php
```diff
@@ -160,4 +160,31 @@ public function test_it_correctly_gets_config_logs()
 
         $this->assertSame(['text' => 'app', 'debug' => 'location'], $config->getLogsTypes());
     }
+
+    public function test_it_sets_ignored_mutators()
+    {
+        $config = <<<'JSON'
+{
+	"mutators": {
```
borNfreee: please use 4 spaces instead of tabs for indentation

src/Mutator/MutatorConfig.php
BackEndTea: Currently it would break on this. However this notation has no use until we implement profiles, which is outside of the scope for this PR.

My plan was to start working on profiles after this, which will involve some refactoring of this class anyways.

borNfreee: np. So am I right saying that we won't release 0.9.0 until profiles are implemented?

If so, we can skip creating documentation as well for now

tests/Fixtures/e2e/IgnoredMutators/src/OtherClass.php
```diff
@@ -0,0 +1,23 @@
...
+    {
+        return true;
+    }
+
+    private function isFalse()
+    {
+        return false;
+    }
+
+}
```
BackEndTea: Self reminder to add line break

src/Mutator/MutatorConfig.php
BackEndTea: Once this gets implemented profiles won't be that big a job, so i'm okay with delaying the doc pr until then.

borNfreee:
👏  Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee

Issue #182, opened 2018-02-15
Existing coverage: "cannot load zend opcache"


luispabon:
So I'm giving the new pre-existing coverage a spin (dev-master 1966cbb70). 

This is my coverage config at phpunit.xml

```
        <!-- Required by infection -->
        <log type="junit" target="reports/infection/phpunit.junit.xml"/>
        <log type="coverage-xml" target="reports/infection/coverage-xml"/>
        <log type="coverage-clover" target="reports/infection/coverage.xml"/>
```

I can see the files are generated correctly at the right folder.

Then I'm running infection like so:

```
~ bin/infection --coverage=./reports/infection/
    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____ 
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/
 
Running initial test suite...

PHPUnit version: 7.0.1

    1 [============================] < 1 sec
                                                                                                       
 [ERROR] Project tests must be in a passing state before running Infection.                            
                                                                                                       
         Phpunit reported an exit code of 143.                                                         
                                                                                                       
         Refer to the phpunit's output below:                                                          
                                                                                                       
         STDERR:                                                                                       
                                                                                                       
         Cannot load Zend OPcache - it was already loaded                                              
~
```

Tests are indeed in a passing stage, and I do have opcache available, although not enabled, in cli :

```
~ sudo phpdismod xdebug
~ bin/phpunit
PHPUnit 7.0.1 by Sebastian Bergmann and contributors.

Error:         No code coverage driver is available

...............................................................  63 / 210 ( 30%)
............................................................... 126 / 210 ( 60%)
............................................................... 189 / 210 ( 90%)
.....................                                           210 / 210 (100%)

Time: 367 ms, Memory: 22.00MB

OK (210 tests, 944 assertions)

~  php -i|grep opcache -i
Additional .ini files parsed => /etc/php/7.2/cli/conf.d/10-opcache.ini,
    with Zend OPcache v7.2.2-3+ubuntu17.10.1+deb.sury.org+1, Copyright (c) 1999-2018, by Zend Technologies
Zend OPcache

[ ...]
opcache.enable => On => On
opcache.enable_cli => Off => Off
[...]

```

sidz:
[...]

Issue #181, opened 2018-02-14
Do not require Xdebug/phpdbg when existing coverage is provided


borNfreee:
Extracted from https://github.com/infection/infection/pull/176

If someone uses the `--coverage` option, there shouldn't be any need for them to have `Xdebug` enabled, or run the command through `phpdbg`

For example, when user uses `phpdbg` to get the coverage of my project, he would then like to be able to run infection without `phpdbg`, since that slows it down.

PR #183, opened 2018-02-17
Do not require Xdebug/phpdbg when existing coverage is provided 

sidz:
Also 

* fixes https://github.com/infection/infection/issues/182 
* fixes https://github.com/infection/infection/issues/181

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -9,7 +9,6 @@ declare(strict_types=1);
 namespace Infection\Command;
 
 use Infection\Config\InfectionConfig;
-use Infection\Console\Application;
 use Infection\Console\Exception\InfectionException;
 use Infection\Console\Exception\InvalidOptionException;
 use Infection\Console\LogVerbosity;
@@ -44,9 +43,6 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Process\Process;
 
-/**
- * @method Application getApplication()
- */
 class InfectionCommand extends BaseCommand
 {
     const CI_FLAG_ERROR = 'The minimum required %s percentage should be %s%%, but actual is %s%%. Improve your tests!';
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -42,12 +42,10 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
-use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Process\Process;
 
 /**
  * @method Application getApplication()
- * @method SymfonyStyle getIO()
  */
 class InfectionCommand extends BaseCommand
 {
@@ -62,6 +60,11 @@ class InfectionCommand extends BaseCommand
      * @var EventDispatcher
      */
     private $eventDispatcher;
+
+    /**
+     * @var bool
+     */
+    private $skipCoverage;
 
     protected function configure()
     {
@@ -158,8 +161,7 @@ class InfectionCommand extends BaseCommand
     {
         $container = $this->getContainer();
         $testFrameworkKey = $input->getOption('test-framework');
-        $skipCoverage = strlen(trim($input->getOption('coverage'))) > 0;
-        $adapter = $container->get('test.framework.factory')->create($testFrameworkKey, $skipCoverage);
+        $adapter = $container->get('test.framework.factory')->create($testFrameworkKey, $this->skipCoverage);
 
         $metricsCalculator = new MetricsCalculator();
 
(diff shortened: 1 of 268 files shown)


sidz:
also I'd say that don't need to add $skipCoverage parameter to the InitialConfigBuilder at all.

src/Command/InfectionCommand.php
```diff
@@ -168,13 +170,20 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $processBuilder = new ProcessBuilder($adapter, $container->get('infection.config')->getProcessTimeout());
         $testFrameworkOptions = $this->getTestFrameworkExtraOptions($testFrameworkKey);
 
-        $initialTestsRunner = new InitialTestsRunner($processBuilder, $this->eventDispatcher);
-        $initialTestSuitProcess = $initialTestsRunner->run($testFrameworkOptions->getForInitialProcess(), $skipCoverage);
+        if (!$this->skipCoverage) {
+            $initialTestsRunner = new InitialTestsRunner($processBuilder, $this->eventDispatcher);
```
borNfreee: I'm afraid it's not correct to skip Initial Tests Run step. It's essential to know that *all* tests are in a passing state.

Running `./vendor/bin/phpunit` on the project actually differs from what Infection runs inside, because:

1. it's possible that some listeners are different (e.g. because of bug in infection like we had earlier)
2. it's possible that tests are green on usual PHPUnit run but fail on Infection run in a separate process

Btw, if you dump PHPUnit's output from initial tests suite run for infection itself, you can notice that some of the tests are skipped (while they are not skipped in usual PHPUnit run). This is because of different environment (STTY). In other words - several tests don't work when running in symfony process and they are skipped

https://github.com/infection/infection/blob/64e0596f027c23efc60440855064ae6ca6146603/tests/Config/ValueProvider/PhpUnitCustomExecutablePathProviderTest.php#L66-L68

So if we skip initial tests run, we can't be sure tests are green. As a result - many Mutant can be killed by that always-failed tests and it will lead to incorrect MSI

borNfreee:
> also I'd say that don't need to add $skipCoverage parameter to the InitialConfigBuilder at all.

could you please explain why? If I run infection with xdebug/phpdbg on this branch, I get a notice about performance impact. But since phpunit.xml does not contain coverage loggers, coverage is not generated. Without this flag, loggers will be added to XML files and coverege will be generated

sidz:
> What is the reason of #182 bug? I can't reproduce it on MacOS
Maybe you're running Infection via phpdbg because this issue is reproducible when Infection is running with xdebug.
Issue caused by https://github.com/infection/infection/blob/master/src/Process/Builder/ProcessBuilder.php#L45 as `$includeArgs` will be equal `true` for case while `$skipCoverage` is true.
You'll get command like this:
`exec /usr/bin/php7.2 -c /tmp/infectionvdYufd /var/www/infection/vendor/phpunit/phpunit/phpunit --configuration /tmp/infection/phpunitConfiguration.initial.infection.xml --stop-on-failure`

As we can see we have -c option with config which already has all extensions + disabled xdebug. But  as I can see `PHP_INI_SCAN_DIR` is not equal empty string and php will parse all extension configs from this folder again and we will get issue with already loaded extensions.

I'm not familiar with MacOS but in Linux we have /etc/php/php.ini (with main php config options) + /etc/php/conf.d/ for all extensions like xdebug, dom, cache, whatever else. And PHP_INI_SCAN_DIR usually has /etc/php/conf.d path which we need to reset to the empty string to avoid any issues like this.

> could you please explain why?
It depends on do we need to run Initial Tests Runner or not. So let's back to it a bit later if it will have a sense :P

src/Command/InfectionCommand.php
sidz: got it

sidz:
Here is the final version (I hope).

borNfreee:
Does it make sense to call `xdebug-disable` before infection is run (where `$PHPDBG == 1`)? Or, to cover more scenario, even create an additional item in the travis matrix with disabled xdebug run?

BackEndTea:
:+1: On adding another run to travis with xdebug disabled

sidz:
I thought that xdebug is enabled by default for each build so I don't think that we need use xdebug with phpdbg in the same time.

borNfreee:
Sorry if I was not clear, I meant to add additional run with disabled xdebug and disabled phpdbg

.travis.yml
```yml
matrix:
  include:
    - php: 7.0
      env: NO_DEBUGGER=1

...

script:
  - composer analyze
  - ...
  - if [[ $NO_DEBUGGER = 1 ]]; then xdebug-disable; fi
  - ...
  - if [[ $PHPDBG != 1 ]]; then bin/infection $INFECTION_FLAGS; else phpdbg -qrr bin/infection $INFECTION_FLAGS; fi
```

or smth like that

sidz:
any thoughts why e2e fail? xdebug should be enabled for them

sidz:
Now got and issue with php-cs-fixer but locally can't reproduce it. @borNfreee any thoughts?

borNfreee:
I can reproduce it after removing old `php-cs-fixer-v2.phar`

```
rm php-cs-fixer-v2.phar
composer cs:check
```

sidz:
thx, do we want to have a new line separator between header and `declare` block? as right now I need to apply it on every file?

BackEndTea:
That was how we had it before php cs fixer introduced a small bug

sidz:
Finally tests are green \o/. PR is ready for re-review

borNfreee:
Thank you @sidz

PR #184, opened 2018-02-18
Reuse created mutant files to avoid traversing and pretty printing

borNfreee:
Implemented [this idea](https://github.com/infection/infection/issues/75#issuecomment-357546689)

On my laptop, it saves 2s for infection itself. Not much, but still I think there is no need to traverse, pretty print files if we already have mutant on the filesystem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ matrix:
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


borNfreee:
Thank you @sidz and @theofidry

Issue #174, opened 2018-02-11
Add an option to pass additional parameters to the PHP binary


addiks:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.7.1
| Test Framework version | PHPUnit 6.5.6
| PHP version | PHP 7.1.14-1+ubuntu16.04.1+deb.sury.org+1
| Platform    | Ubuntu 16.04.3 LTS
| Github Repo | Not specific to one repo; It's specific to my machine

TL;DR: This is a feature-request for a new parameter "--php-arguments" for infection with that one can specify additional parameters to be given to the php executable that is used to start phpunit.

On my machine i enable xdebug on a per-process basis. Per default, xdebug is disabled. This is bebause xdebug can seriously slow down some processes. If i need xdebug to be enabled in a process, it pass the parameter "-d zend_extension=xdebug.so" to the process (mostly by using bash-aliases).

```bash
# ~/.bash_rc
...
alias phpd='php -d zend_extension=xdebug.so -d xdebug.remote_autostart=1 '
alias phpp='php -d zend_extension=xdebug.so -d xdebug.profiler_enable=1 '
alias phpx='php -d zend_extension=xdebug.so'
...
```
Xdebug is per default disabled in the php-module-config "/etc/php/7.1/cli/conf.d/20-xdebug.ini":
```ini
;zend_extension=xdebug.so

[xdebug]

xdebug.remote_enable=On
xdebug.remote_handler=dbgp
xdebug.remote_mode=req
xdebug.remote_port=9000
xdebug.remote_host=127.0.0.1
xdebug.remote_autostart=Off
xdebug.remote_log=/var/log/xdebug_remote.log

```
With this configuration infection does not work out of the box, instead it tells me to use phpdbg or xdebug:
```bash
$ infection
You need to use phpdbg or install and enable xdebug in order to allow for code coverage generation.
$
```
And when i try to run infection with xdebug (see above), this happens:
```bash
$ phpx ~/.config/composer/vendor/infection/infection/bin/infection
    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____ 
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/
 
Running initial test suite...

PHPUnit version: 6.5.6

   44 [============================] < 1 sec

Generate mutants...

[...]

PR #185, opened 2018-02-18
Add an option to pass additional parameters to the PHP binary

sidz:
PR is bringing `--initial-tests-php-options` option to cover https://github.com/infection/infection/issues/174 feature request

* [x] Add `--initial-tests-php-options`
* [x] Update documentation


src/Console/Application.php
```diff
@@ -98,6 +103,14 @@ public function run(InputInterface $input = null, OutputInterface $output = null
         return parent::run($input, $output);
     }
 
+    private function isInitialTestPhpOptionHasXdebug(InputInterface $input): bool
+    {
+        return (bool) preg_match(
+            '/(zend_extension\s*=.*xdebug.*)/mi',
+            (string) $input->getParameterOption('--initial-tests-php-options', true)
```
borNfreee: probably, the second parameter (default value) is not needed here as we have default value empty string in `configure()` method

sidz: parameters binding is not happening there.
To avoid manual type cast (with `(string)`) we can replace boolean value to the empty string instead

sidz:
I've found an issue. please don't merge it :)

sidz:
Ready for review.

borNfreee:
@sidz ready to merge?

sidz:
@borNfreee yes.

borNfreee:
Thank you @sidz

PR #188, opened 2018-02-19
Remove dead code & small refactor

BackEndTea:
* Replace array_walk wtih foreach
* Removed dead code from Locator class
* Rename filter to filters
* Add doc block type hints to fields


borNfreee:
Thank you @BackEndTea

PR #192, opened 2018-02-20
Add String_ Mutator

BackEndTea:
This PR:

- [x] Adds new feature: the string_ mutator
- [x] Covered by tests

Adds a new Mutator that replaces strings with an empty value, which for this project, adds roughly 600 mutations, and reduces MSI by about 10%, and covered code MSI by 2%.

### DIFF
--- a/src/Config/InfectionConfig.php
+++ b/src/Config/InfectionConfig.php
@@ -62,6 +62,7 @@ use Infection\Mutator\ReturnValue\IntegerNegation;
 use Infection\Mutator\ReturnValue\NewObject;
 use Infection\Mutator\ReturnValue\This;
 use Infection\Mutator\Sort\Spaceship;
+use Infection\Mutator\Value\String_;
 use Infection\Mutator\ZeroIteration\Foreach_;
 use Symfony\Component\Filesystem\Filesystem;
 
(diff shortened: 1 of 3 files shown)


borNfreee:
Could you please describe the main use cases where it is useful from your point of view?

> for this project, adds roughly 600 

this is about 50% of all generated mutations. I would say that this mutation operator will slow down infection on large projects a lot, which is not a good thing.

Maybe we should wait for Profiles feature and add it to the "Nightmare" profile? :)

---------

As in https://github.com/infection/infection/pull/154 - we need some confidence and analyzing before merging it

For example: we should skip this mutation

```diff
- preg_match('/\d/', '123');
+ preg_match('', '123');
```

because it generates a PHP Warning: https://3v4l.org/AovDJ and will always be killed. That said this particular mutation is useless.

theofidry:
For the record, I think the following is quite useful:

```php
$original = 'foo';

$mutated = oneOf([
    '',
    'null',
    ' ',
    ' foo',
    'foo ',
]);
```

I however think like @borNfreee that it's way too general to be useful. So I would keep the idea, but realistically if we start adding those kind of mutations without thinking twice, we are not going to transform a 5s unit tests suite into a 5min mutation test one, but into a 2h one in no time.

BackEndTea:
Closing this for now, until i have a better idea of what to do with this mutator.

PR #198, opened 2018-02-28
Ignore MSI violations with zero mutations

Landerstraeten:
- [x] Covered by tests
- [x] Doc PR https://github.com/infection/site/pull/22

We have a [task](https://github.com/phpro/grumphp/blob/master/doc/tasks/infection.md) for Infection in [Grumphp](https://github.com/phpro/grumphp). The Grumphp task uses the `--filter` option of infection so I only listens on changed files.

When Infection runs with a required MSI percentage and there are no mutations generated, Infection results in a failure since the required percentage is not reached.

This PR fast-exits Infection when no mutations are generated.

```
Running initial test suite...

PhpSpec version: 3.4.3



Generate mutants...



0 mutations were generated:
       0 mutants were killed
       0 errors were encountered
       0 time outs were encountered

Metrics:
         Mutation Score Indicator (MSI): 0%
         Mutation Code Coverage: 0%
         Covered Code MSI: 0%

 [ERROR] The minimum required MSI percentage should be 100%, but actual is 0%.
         Improve your tests!



Creating mutated files and processes:    0/0
```

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -205,8 +205,6 @@ class InfectionCommand extends BaseCommand
         );
 
         $mutations = $mutationsGenerator->generate($input->getOption('only-covered'), $input->getOption('filter'));
-        $ignoreMsi = (bool) $input->getOption('ignore-msi-with-no-mutations');
-        $hasGeneratedMutations = \count($mutations) > 0;
 
         $mutationTestingRunner = new MutationTestingRunner(
             $processBuilder,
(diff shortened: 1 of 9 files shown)


BackEndTea:
👍 For the idea and the PR.

I"d like to see a nice message to the user indicating that it exited without generating any mutations.

Also i'm not 100% on giving it a exit code of 0. What do you think @borNfreee

src/Command/InfectionCommand.php
```diff
@@ -200,6 +200,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         $mutations = $mutationsGenerator->generate($input->getOption('only-covered'), $input->getOption('filter'));
 
+        if (!((bool) \count($mutations))) {
```
borNfreee: `\count($mutations) === 0` if more general, without casting

borNfreee:
> Also i'm not 100% on giving it a exit code of 0.

This is what @Landerstraeten wants to achieve, since now Infection exits with non zero exit code.

my thoughts are:

it's better to have a parameter for this behaviour. Thus, it won't be a BC break. 

I saw different variations in libs about this case, like `--non-zero-exit-on-violation` on phpmnd (magic number detector), `--ignore-violations-on-exit` on PHPMD, etc.

So, we can add an option for such case, where _we want to ignore MSI violations with zero mutations_. I'm not sure about naming though

Landerstraeten:
I vote for `--ignore-msi-with-no-mutations` 😁

borNfreee:
then, could you please

* Create a documentation for this new option here https://github.com/infection/site/blob/master/src/guide/command-line-options.md
* Add a e2e test to make sure it works as expected (you can use https://github.com/infection/infection/blob/master/tests/add_new_e2e)

BackEndTea:
The travis error is due to this branch not being properly based on master. If it is rebased it should have the correct MSI etc.

I'm okay with merging this without an e2e test, as long as it gives some kind of notification to the user that there were 0 mutations, and that it exited succesfully because of that.

borNfreee:
> I'm okay with merging this without an e2e test

Sorry, no new functionality without tests. Just want to minimize sleepless nights after release dates 🤣 

I've added e2e test and simplified the code.

> as long as it gives some kind of notification to the user that there were 0 mutations

It shows the whole Metrics section to the console.

Ready for review.

borNfreee:
Thanks everyone!

Issue #55, opened 2017-10-19
Use PHP-Parser 4 to preserve mutated code formatting


morozov:
Currently, when displaying the difference between original and mutated code, the pretty-formatted representation of the AST is used:

https://github.com/infection/infection/blob/da36bf6a419d9dfc5450eba3d463712440c04023/src/Mutant/MutantCreator.php#L55-L59

Depending on how the original code is formatted, it may lead to the inconsistency between the diff and the original file contents. Additionally, given that `sebastian/diff 2.0` currently doesn't display common lines of code (sebastianbergmann/diff#69), in some cases, the information about how exactly the original code was mutated may be hard to understand.

PHP-Parser 4 introduces the ability to use the [formatting-preserving pretty printing](https://github.com/nikic/PHP-Parser/blob/v4.0.0alpha1/doc/component/Pretty_printing.markdown#formatting-preserving-pretty-printing). Using this kind of formatting together with displaying line numbers may eliminate the need for displaying common lines in the diff.

borNfreee:
Unfortunately, we can't use it right now because of additional formatting of [at least] empty lines.

Here is a result of using new `printFormatPreserving()`:

```diff
--- Original
+++ New
@@ @@
      */
-    public function changeTargetDate(\DateTimeInterface $newTargetDate)
+    protected function changeTargetDate(\DateTimeInterface $newTargetDate)
     {
         \Assert\that($newTargetDate)->greaterOrEqualThan(new \DateTimeImmutable());
-
         $this->targetDate = $newTargetDate;
     }
```

Look at the third line from the end - it is removed while it exists on the original source code.

This bug is currently reproducible for `PublicVisibility` and `ProtectedVisibility` mutators (seems like nested statements of the changed `ClassMethod_` node do not preserve formatting).

I've pushed changes to branch [php-parser-400](https://github.com/infection/infection/tree/php-parser-400) so we can back to it later or if someone wants to play with it. 

I've created an issue in PHP-Parser repository for this case https://github.com/nikic/PHP-Parser/issues/428

morozov:
Cool! I didn't hope it would work right off the bat as it's still in alpha. Thanks for taking a look.

borNfreee:
https://github.com/nikic/PHP-Parser/issues/428 is fixed, so we can continue testing with PHP-Parser 4.0

sidz:
PHP-Parser 4.0 is out. https://github.com/nikic/PHP-Parser/releases/tag/v4.0.0

[...]

PR #201, opened 2018-03-01
Upgrade to Php-parser 4.0

BackEndTea:
This PR:

- [x] Adds support for PHP-Parser 4.0
- [x] Covered by tests

### DIFF
--- a/app/container.php
+++ b/app/container.php
@@ -126,7 +126,7 @@ $c['version.parser'] = function (): VersionParser {
 };
 
 $c['lexer'] = function (): Lexer {
-    return new Lexer([
+    return new Lexer\Emulative([
         'usedAttributes' => [
             'comments', 'startLine', 'endLine', 'startTokenPos', 'endTokenPos', 'startFilePos', 'endFilePos',
         ],
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
     },
     "require": {
         "php": "^7.0",
-        "nikic/php-parser": "^3.1",
+        "nikic/php-parser": "^4.0",
         "padraic/phar-updater": "^1.0.4",
         "pimple/pimple": "^3.2",
         "sebastian/diff": "^1.4 || ^2.0 || ^3.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,28 +4,28 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "96811e05fc62914328c0e44673ec8afb",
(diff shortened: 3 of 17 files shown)


borNfreee:
fyi: there was some work in https://github.com/infection/infection/tree/php-parser-400 branch, I will dig into it later

sidz:
imo we should drop 3.0 support.

BackEndTea:
> imo we should drop 3.0 support.

We could, but i'd like to know how many people use infection as a dev dependency that also have a dependency on PHP-Parser

borNfreee:
https://twitter.com/infection_php/status/969309000994041858

theofidry:
@borNfreee if people are unhappy with it I think we could also mention bamarni/composer-bin-plugin and eventually [my article](https://medium.com/@tfidry/managing-your-dependencies-in-php-321d584441ab) about it

borNfreee:
> https://twitter.com/infection_php/status/969309000994041858

let's drop 3.x

theofidry:
Genuinely surprised by the amount of trollish answers :/

BackEndTea:
> If 4.0 is used, we should use printFormatPreserving() rather than prettyPrintFile().

I have been playing around with that, but it causes all kinds of small inconsistencies. 

When running it on infection all diffs are this:
```diff
--- Original
+++ New
@@ @@
 
-declare(strict_types=1);
+declare (strict_types=1);
```

borNfreee:
Then we need to post all such issues to PHP-Parser issue tracker.

Original reason of switching to 4.0 was saving formatting https://github.com/infection/infection/issues/55 (`Use PHP-Parser 4 to preserve mutated code formatting), because now files lose their code styles.

Of course, we can switch and don't use it, but what is our goal?

BackEndTea:
> Of course, we can switch and don't use it, but what is our goal?

Then our goal would be to support projects that want to use both PHP-parser and infection as dev-dependencies.

borNfreee:
I would suggest the following:

1. Move from 3.x to 4.x as it is new version, better performance, etc.
2. Drop support of 3.x as discussed above
3. do not use `printFormatPreserving()` at the moment because it's still not stable
4. from time to time post issues to PHP-Parser repository if we want eventually to activate `printFormatPreserving()`

BackEndTea:
I've changed this to upgrade to 4.0, and drop support for 3.0. And it currently does not implement `printFormatPreserving()` due to previously mentioned issues.

borNfreee:
Thank you @BackEndTea 👍

BackEndTea:
Thank you @borNfreee, @theofidry & @sidz

Sentoki:
https://github.com/infection/infection/pull/201/files#diff-3a47420a614944af649053250f50140fR102
We can have object in ast/Node name property? I have string with class name in this property and infection get error.

PR #202, opened 2018-03-01
Clean up the Mutant Creator

BackEndTea:
This PR:

- [x] Cleans up the `MutantCreator` and adds a test

No changes in logic, its mostly just moving around things into its own methods.

### DIFF
--- a/src/Mutant/MutantCreator.php
+++ b/src/Mutant/MutantCreator.php
@@ -34,6 +34,11 @@ class MutantCreator
      */
     private $prettyPrinter;
 
+    /**
+     * @var string[]
+     */
+    private $prettyPrintedCache = [];
+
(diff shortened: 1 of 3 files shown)


src/Mutant/MutantCreator.php
```diff
@@ -93,4 +78,28 @@ private function isCoveredByTest(Mutation $mutation, CodeCoverageData $codeCover
 
         return $codeCoverageData->hasTestsOnLine($filePath, $line);
     }
+
+    private function getMutatedCode(Mutation $mutation, string $mutatedFilePath): string
```
borNfreee: I would suggest to name it `createMutatedCode` as it more suite its content (file is created, content is generated)

BackEndTea:
@borNfreee I changed naming

borNfreee:
Thank you @BackEndTea

Issue #205, opened 2018-03-04
Display friendly message when infectin.json contains invalid JSON


borNfreee:
When `infection.json` contains invalid JSON, we have a very unfriendly fatal error message that leads to issues like this one https://github.com/infection/infection/issues/204:

```
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Infection\Config\InfectionConfig::__construct() must be an instance of stdClass, null given, called in infection/src/Console/Application.php on line 195 and defined in infection/src/Config/InfectionConfig.php:165
Stack trace:
```

This confuses users. Instead of displaying this error, we should display something like `infection.json contains invalid JSON. Please fix it and run Infection again`.

It can be fixed here: https://github.com/infection/infection/blob/0.8.1/src/Console/Application.php#L195

PR #209, opened 2018-03-05
Fix: Throw exception if config file contains invalid JSON

localheinz:
This PR

- [x] throws an exception in `Console\Application` when invalid JSON was found in a configuration file

Fixes https://github.com/infection/infection/issues/205.

💁‍♂️ Not sure, should I add an end-to-end test for this?

### DIFF
--- a/src/Config/Exception/InvalidConfigException.php
+++ b/src/Config/Exception/InvalidConfigException.php
@@ -1,0 +1,22 @@
+<?php
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Config\Exception;
+
+final class InvalidConfigException extends \RuntimeException
+{
+    public static function invalidJson(string $configFile, string $errorMessage): self
+    {
+        return new self(sprintf(
+            'The configuration file "%s" does not contain valid JSON: %s.',
+            $configFile,
+            $errorMessage
+        ));
+    }
+}
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace Infection\Console;
 
 use Infection\Command;
+use Infection\Config\Exception\InvalidConfigException;
 use Infection\Config\InfectionConfig;
 use Infection\Php\ConfigBuilder;
 use Infection\Php\XdebugHandler;
@@ -167,6 +168,11 @@ ASCII;
         return $this->io;
     }
 
+    /**
+     * @param InputInterface $input
+     *
+     * @throws InvalidConfigException
+     */
     private function buildDynamicDependencies(InputInterface $input)
     {
         $this->container['infection.config'] = function (Container $c) use ($input): InfectionConfig {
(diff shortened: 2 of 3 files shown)


BackEndTea:
> Not sure, should I add an end-to-end test for this?

I'm somewhat against adding an e2e test for every single thing. I'd prefer if we had a proper way to test classes in the `Console` and `Command` namespaces. But thats way outside the scope of this PR.

localheinz:
@BackEndTea 

Me, too, that's why I am asking - especially since I couldn't find any existing tests for `Console\Application`!

src/Console/Application.php
```diff
@@ -188,11 +193,21 @@ private function buildDynamicDependencies(InputInterface $input)
...
             } catch (\Exception $e) {
+                $infectionConfigFile = null;
                 $json = '{}';
                 $configLocation = getcwd();
             }
 
-            return new InfectionConfig(json_decode($json), $c['filesystem'], $configLocation);
+            $config = json_decode($json);
+
+            if (null === $config && 0 !== json_last_error() && $infectionConfigFile) {
```
BackEndTea: The `$infectionConfigFile` is somewhat useless, since it shouldn't be possible to have a json error with our default json.

src/Console/Application.php
```diff
@@ -188,11 +193,21 @@ private function buildDynamicDependencies(InputInterface $input)
...
+                throw new \RuntimeException(sprintf(
+                    'The configuration file "%s" does not contain valid JSON.',
+                    $infectionConfigFile
```
localheinz: Just want to make sure `$infectionConfigFile` is defined before I use it here.

src/Console/Application.php
```diff
@@ -188,11 +193,21 @@ private function buildDynamicDependencies(InputInterface $input)
...
             } catch (\Exception $e) {
+                $infectionConfigFile = null;
                 $json = '{}';
                 $configLocation = getcwd();
             }
 
-            return new InfectionConfig(json_decode($json), $c['filesystem'], $configLocation);
+            $config = json_decode($json);
+
+            if (null === $config && 0 !== json_last_error()) {
```
BackEndTea: Could you use `JSON_ERROR_NONE` so its more readable?

src/Console/Application.php
```diff
@@ -188,11 +193,21 @@ private function buildDynamicDependencies(InputInterface $input)
...
+                throw new \RuntimeException(sprintf(
```
sidz: We need to create our own exception based on the RuntimeException. As @theofidry usually tells - everything in the PHP can throw RuntimeException

src/Config/Exception/InvalidConfigException.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
```
sidz: missed header 😋

src/Console/Application.php
```diff
@@ -188,11 +194,21 @@ private function buildDynamicDependencies(InputInterface $input)
...
             } catch (\Exception $e) {
+                $infectionConfigFile = null;
                 $json = '{}';
                 $configLocation = getcwd();
             }
 
-            return new InfectionConfig(json_decode($json), $c['filesystem'], $configLocation);
+            $config = json_decode($json);
+
+            if (is_string($infectionConfigFile) && null === $config && 0 !== json_last_error()) {
```
sidz: could we please use `JSON_ERROR_NONE` instead of the magic number?

tests/Config/Exception/InvalidConfigExceptionTest.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
```
sidz: missed header

borNfreee:
Thank you @localheinz

localheinz:
Thank you, @BackEndTea, @borNfreee, and @sidz!

PR #210, opened 2018-03-05
Fix: Ignore infection.json

localheinz:
This PR

* [x] ignores `infection.json`

Spotted in #209.

💁‍♂️ A configuration file `infection.json.dist` is checked in - I believe we should ignore `infection.json`, then.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ build/*
 .php_cs.cache
 infection-cache
 .travis/infection-private.pem
+infection.json


.gitignore
```diff
@@ -5,4 +5,5 @@ build/*
 !build/.gitkeep
 .php_cs.cache
 infection-cache
-.travis/infection-private.pem
\ No newline at end of file
+.travis/infection-private.pem
+infection.json
```
localheinz: The order of entries in `.gitignore` appears to be a bit random; please let me know if there is a specific rule you would like me to follow, and then I will move this entry into the appropriate place.

borNfreee:
Thank you @localheinz

localheinz:
Thank you, @BackEndTea, @borNfreee, and @sidz!

PR #218, opened 2018-03-12
Fix: using custom path to TestFramework

### LINKED ISSUE #217: Custom path to Test Framework not used in 0.8.1
| Question    | Answer
| Infection version | 0.8.1 (`infection.phar --version`)

Patch to come.

tomtomau:
This PR:

- [x] Covered by tests

Fixes https://github.com/infection/infection/issues/217

* As a user, I want to use a custom path to the TestFramework (fixes #217)
* As a user, I want a descriptive error message if my custom path does not exist, rather than use a fallback

### DIFF
--- a/tests/Finder/TestFrameworkFinderTest.php
+++ b/tests/Finder/TestFrameworkFinderTest.php
@@ -1,0 +1,29 @@
+<?php
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Tests\Finder;
+
+use Infection\Finder\SourceFilesFinder;
+use Infection\Finder\TestFrameworkFinder;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Finder\Finder;
+
+class TestFrameworkFinderTest extends TestCase
+{
+    public function test_it_can_load_a_custom_path(): void
+    {
+        $filename = '/tmp/infection-test-framework-finder-test' . uniqid('', false);
+
+        touch($filename);
+
+        $frameworkFinder = new TestFrameworkFinder('not-used', $filename);
+
+        static::assertEquals($filename, $frameworkFinder->find(), 'Should return the custom path');
+    }
+}
--- a/tests/Finder/TestFrameworkFinderTest.php
+++ b/tests/Finder/TestFrameworkFinderTest.php
@@ -11,27 +11,49 @@ namespace Infection\Tests\Finder;
 
 use Infection\Finder\Exception\FinderException;
 use Infection\Finder\TestFrameworkFinder;
+use Infection\Utils\TmpDirectoryCreator;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Filesystem\Filesystem;
 
 class TestFrameworkFinderTest extends TestCase
 {
+    /**
+     * @var Filesystem
+     */
+    private $fileSystem;
+
+    /**
+     * @var string
+     */
+    private $workspace;
+
+    /**
+     * @var string
+     */
+    private $tmpDir;
+
+    protected function setUp()
+    {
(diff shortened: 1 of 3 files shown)


tests/Finder/TestFrameworkFinderTest.php
```diff
@@ -0,0 +1,42 @@
...
+
+use Infection\Finder\Exception\FinderException;
+use Infection\Finder\SourceFilesFinder;
+use Infection\Finder\TestFrameworkFinder;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Finder\Finder;
+
+class TestFrameworkFinderTest extends TestCase
+{
+    public function test_it_can_load_a_custom_path(): void
```
BackEndTea: Please remove the `: void`  since that breaks test on 7.0

BackEndTea:
👍 Its better to throw an error rather than using a fallback if the user specified a location.

I'd like to extend this behavior to other places, where the user specifies locations of files.

tests/Finder/TestFrameworkFinderTest.php
tomtomau: ah of course, habit :)

tests/Finder/TestFrameworkFinderTest.php
```diff
@@ -0,0 +1,40 @@
...
+
+use Infection\Finder\Exception\FinderException;
+use Infection\Finder\TestFrameworkFinder;
+use PHPUnit\Framework\TestCase;
+
+class TestFrameworkFinderTest extends TestCase
+{
+    public function test_it_can_load_a_custom_path()
+    {
+        $filename = '/tmp/infection-test-framework-finder-test' . uniqid('', false);
```
sidz: this test won't work on windows I think. Could you please use `sys_get_temp_dir` instead? also it'll be a good idea to clean up.

you can use symfony filesystem class for example. We already have couple tests which do similar things.

tomtomau:
Updated @sidz

tests/Finder/TestFrameworkFinderTest.php
```diff
@@ -0,0 +1,43 @@
...
+declare(strict_types=1);
+
+namespace Infection\Tests\Finder;
+
+use Infection\Finder\Exception\FinderException;
+use Infection\Finder\TestFrameworkFinder;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Filesystem\Filesystem;
+
+class TestFrameworkFinderTest extends TestCase
```
borNfreee: Please, use setUp and tearDown methods to create and remove temp folders:

https://github.com/infection/infection/blob/12bbf5ab87d840c8b64ec973de5e1506d27087b5/tests/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilderTest.php#L49-L64

tests/Finder/TestFrameworkFinderTest.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    public function test_it_can_load_a_custom_path()
+    {
+        $fileSystem = new Filesystem();
+        $filename = $fileSystem->tempnam(sys_get_temp_dir(), 'infection-test');
+        $fileSystem->touch([$filename]);
+
+        $frameworkFinder = new TestFrameworkFinder('not-used', $filename);
+
+        static::assertEquals($filename, $frameworkFinder->find(), 'Should return the custom path');
```
borNfreee: let's use `$this->assertEquals()` to be consistent across the project

tomtomau:
Updated! @borNfreee

djmattyg007:
Would you be able to create a new release (ie 0.8.2) that includes this patch?

borNfreee:
@djmattyg007 I will try to release it today or tomorrow

1ed:
Also affected by this. @borNfreee could you please release a new version if you have some time? Thanks!

borNfreee:
@djmattyg007 @1ed 

Released under [0.8.2](https://github.com/infection/infection/releases/tag/0.8.2)

Sorry for the delay

1ed:
That's grate! Thank you very much @borNfreee! There will be a phar too?

borNfreee:
> There will be a phar too?

@1ed sorry for this delay, uploaded the PHAR

PR #219, opened 2018-03-12
Fix test names

BackEndTea:
Since all other tests use `snake_case`.


borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee

Issue #90, opened 2017-11-24
Scoped configuration


theofidry:
Taking advantage of #88 and #89, there is one more point that I think would be awesome in terms of usability. The idea is that you don't have the same level of exigence for every part of your application. As such, I think it would be really good to be able to define scoped configuration, i.e. define options that would be applied to only a part of the code.

For example, assuming #89 is implemented, I could define a high level of strictness for a specific part of my application which is very important and a lower level of strictness for the rest.

BackEndTea:
If #89 and #79 are implemented, would that resolve this?

theofidry:
Well not exactly :) To be exact what has been implemented allow this but at a global level. What I'm actually thinking of is what we currently have but for smaller subsets. For example you could define that the directory `src/Domain` should be strictly tested so a lot of mutators enabled for it and for another part like `src/Infrastructure` reduce the strictness so apply a lot less mutators.

Right now this would be doable by applying different configuration files and merge the result sort of. But in any case, I think what we have is good enough for now. I'll re-open the issue with more details when I'll have time to dig deeper into this

BackEndTea:
What if, in addition to 'ignoring' mutators on certain namespaces/methods etc, we have a 'whitelist' option ( or another name), that says: only apply this mutator/profile on these namespescapes/methods?

theofidry:
I'm not sure yet tbh. I really prefer to dig into this with a real-life scenario first before going further in what we actually need to make it work :)

PR #220, opened 2018-03-13
Feature: Profiles

BackEndTea:
This PR:

- [x] Adds 'Profiles' to infection
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/24

BC Breaks:
* The --mutators options is no longer case insensitive
* A typo in the --mutators option will no longer just ignore the mutator, but it will stop infection and display an error message. The same goes for wrong keys in the config.

Fixes: #89, Fixes: #90, Fixes: #237 

Follows: #177 

The addition of profiles gives developers much more flexibility in configuring what to mutate and what not, especially in combination with the ignore functionality.

This PR opens the way for mutators like #154 & #192 in some kind of a 'nightmare' profile.

I'm not 100% sold on the current names of profiles, so ideas on that are especially welcome.


BackEndTea:
@borNfreee This PR does have some magic related to class names, so i'd like it if you could test it with a scoped phar

borNfreee:
I will be able to try it with php scoper not early than on the weekend. Hope to review sooner.

BackEndTea:
> I will be able to try it with php scoper not early than on the weekend. Hope to review sooner.

I'm thinking of changing it to use a 'translator list' rather than the namespace magic, so then that shouldn't have any impact.

BackEndTea:
We now keep a list of all mutators to have a translation from their name to the class.

There is a test to make sure all the combinations are right, but i'd like to have some kind of automated test that checks that all mutators are added to that list.

pinging @localheinz  since i know he has experience with `AutoReview` tests from PHP-CS-Fixer, and other projects

src/Mutator/MutatorGenerator.php
```diff
@@ -0,0 +1,137 @@
...
+     * @param string $mutator
+     * @param array|bool $setting
+     *
+     * @throws InvalidConfigException
+     */
+    private function registerFromName(string $mutator, $setting)
+    {
+        if (array_key_exists($mutator, MutatorProfile::FULL_MUTATOR_LIST)) {
+            $this->registerFromClass(MutatorProfile::FULL_MUTATOR_LIST[$mutator], $setting);
+        } else {
```
sidz: else is not needed here

src/Mutator/MutatorProfile.php
```diff
@@ -0,0 +1,207 @@
...
+    const SORT = [
+        Mutator\Sort\Spaceship::class,
+    ];
+
+    const ZERO_ITERATION = [
+        Mutator\ZeroIteration\Foreach_::class,
+    ];
+
+    const DEFAULT = [
+        '@arithmetic',
```
sidz: can we move these values to the constants?

src/Mutator/MutatorProfile.php
```diff
@@ -0,0 +1,207 @@
...
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Mutator;
+
+use Infection\Mutator;
+
+class MutatorProfile
```
sidz: final please

BackEndTea: We mark almost none of our other classes final. I'm okay with making this final, but it would not be in line with the rest of the code base.

src/Mutator/MutatorProfile.php
BackEndTea: I'm not sure what you mean.

Do you mean something like 
```php
const ARITHMETIC_PROFILE = '@arithmetic';
```

src/Mutator/MutatorGenerator.php
```diff
@@ -0,0 +1,137 @@
...
+    public function create(): array
+    {
+        if (empty($this->mutatorSettings)) {
+            $this->mutatorSettings = [
+                '@default' => true,
+            ];
+        }
+        $this->mutatorList = [];
+
+        foreach ($this->mutatorSettings as $mutator => $setting) {
```
borNfreee: in this loop variable `$mutator` can be either mutator or profile. Does it make sense to rename it to something like `$mutatorOrProfile` for better readability?

BackEndTea:
I've applied the feedback, and added a test to fix #237

BackEndTea:
> LGTM, except this one ...

Oops, forgot that one, fixed it now.

borNfreee:
Thank you @BackEndTea 🥇 🎉 🎂

BackEndTea:
Thank you @borNfreee  and @sidz

veewee:
Great job @BackEndTea !!!!

PR #207, opened 2018-03-05
Mutation badge

borNfreee:
This PR adds a Mutation Badge feature

How it looks like (for current branch): [Infection MSI](https://infection.github.io)

Image code is: `[Infection MSI](https://infection.github.io)`

- [x] Covered by tests
- [x] Change `mutation-badge` to `master` before merging 
- [x] Doc PR https://github.com/infection/site/pull/25

Fixes https://github.com/infection/infection/issues/168

To understand how it works, read this: http://stryker-mutator.io/blog/2018-02-08/get-your-mutation-score-badge-now.html

We will use Stryker's Dashboard API. Currently, it will work only for Travis, other CI servers will be implemented on demand.

Configuration is quite easy: 

1. Go to https://dashboard.stryker-mutator.io and sign with your github account.
2. Flip the switch next to your repository (activate it)
3. Configure your API key in your project. It is possible to ncrypt this variable using the encrypted environment variables. For example:
`$ travis encrypt STRYKER_DASHBOARD_API_KEY=89b99910-04d8-4xxx-9x91-23d709x828b4 --add`
4. Configure the badge logger in your `infection.json` file

```json
{
   "logs": {
       "badge": {"branch": "master"}
   }
}
```

PR #221, opened 2018-03-14
Add option to not log mutations

BackEndTea:
This PR:

- [x] Adds new feature: Log verbosity 3, which stops logging to files alltogether, which may be usefull for use on CI environments. 
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/26

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -148,7 +148,7 @@ class InfectionCommand extends BaseCommand
                 'log-verbosity',
                 null,
                 InputOption::VALUE_OPTIONAL,
-                'Log verbosity level. 1 - full logs format, 2 - short logs format.',
+                'Log verbosity level. 1 - full logs format, 2 - short logs format, 3 - no logs.',
                 LogVerbosity::DEBUG
             )
             ->addOption(
@@ -176,6 +176,9 @@ class InfectionCommand extends BaseCommand
         $metricsCalculator = new MetricsCalculator();
 
         $this->registerSubscribers($metricsCalculator, $adapter);
+        if (!$input->getOption('log-verbosity') === LogVerbosity::NONE) {
+            $this->registerFileLoggerSubscriber($metricsCalculator);
+        }
 
         $processBuilder = new ProcessBuilder($adapter, $container->get('infection.config')->getProcessTimeout());
         $testFrameworkOptions = $this->getTestFrameworkExtraOptions($testFrameworkKey);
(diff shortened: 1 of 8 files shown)


src/Command/InfectionCommand.php
```diff
@@ -176,6 +176,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $metricsCalculator = new MetricsCalculator();
 
         $this->registerSubscribers($metricsCalculator, $adapter);
+        if (!$input->getOption('log-verbosity') === LogVerbosity::NONE) {
```
borNfreee: Incorrect comparison. Should be `!==` instead of `! $... === LogVerbosity`

borNfreee:
👍 for this enhancement. Totally makes sense

BackEndTea:
I just realized this may conflict with #207, since users probably do want the badge generated on CI envs, but not the text logs.

borNfreee:
Right. Do we need to ignore just file loggers? At least it correlates with initial comment: `...stops logging to files alltogether...` ;)

BackEndTea:
@borNfreee I moved to logic of wheter or not to log to the `BaseFileLoggerSubscriber`.
This makes it testable, and also allows you to make a few changes to it so it still does the mutation badge in your PR.

src/Process/Listener/FileLoggerSubscriber/BaseFileLoggerSubscriber.php
```diff
@@ -73,7 +75,9 @@ public function onMutationTestingFinished(MutationTestingFinished $event)
         $logTypes = $this->filterLogTypes($logTypes);
 
         foreach ($logTypes as $logType => $directory) {
-            $this->useLogger($logType);
+            if ($this->logVerbosity !== LogVerbosity::NONE) {
```
sidz: I'd say that this condition can be moved to the top of this method and do smth like:

```
public function onMutationTestingFinished(MutationTestingFinished $event)
    {
         if ($this->logVerbosity !== LogVerbosity::NONE) {
             return false;
         }

        $logTypes = $this->infectionConfig->getLogsTypes();

        if (empty($logTypes)) {
            return;
        }

        $logTypes = $this->filterLogTypes($logTypes);

        foreach ($logTypes as $logType => $directory) {
            $this->useLogger($logType);
        }
    }
```

src/Command/InfectionCommand.php
```diff
@@ -148,7 +148,7 @@ protected function configure()
                 'log-verbosity',
                 null,
                 InputOption::VALUE_OPTIONAL,
-                'Log verbosity level. 1 - full logs format, 2 - short logs format.',
+                'Log verbosity level. 1 - full logs format, 2 - short logs format, 3 - no logs.',
```
theofidry: usually it's the opposite though: the higher the level the more complete the logs are

src/Process/Listener/FileLoggerSubscriber/BaseFileLoggerSubscriber.php
theofidry: 👍 except for the use of `empty()` :)

src/Command/InfectionCommand.php
BackEndTea: I know, but now it is this way. So i'd propose to keep it how it is right now, but add a notice to the end users that it will be changed.

And then change it in the next release.

BackEndTea:
I've rebased on masterand applied feedback.

It now tries to log the bade even with log verbosity of 'none'.

borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee

PR #224, opened 2018-03-18
Enforce coding standards on every build

sanmai:
This PR enforces coding standards on every build. 

A build will fail on any violation so maintainers won't have to do any of this. 

Travis CI build cache used, so in most cases php-cs-fixer will finish in no time.


composer.json
```diff
@@ -67,12 +68,10 @@
             "./phpstan.phar analyse src tests --level=2 -c phpstan.neon --no-interaction --no-progress"
         ],
         "cs:check": [
-            "wget -nc http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar",
```
borNfreee: Sorry but I don't like to depend on static analyzer tools. PHAR distribution is what is really needed here. There are many reason, at least:

- this is not a dependency of Infection, just tool that analyzes CS
- as a package, it can conflict with other dependencies
- as a package, it is locked to specific version

.travis.yml
```diff
@@ -37,11 +37,13 @@ env:
 cache:
   directories:
     - $HOME/.composer/cache
+    - build/cache
```
borNfreee: seems like it's not cached: https://travis-ci.org/infection/infection/jobs/354907585#L917

sanmai:
Good points. I've removed everything not needed, but kept a commit with removal of redundant options.

Separately I modified commands to run php-cs-fixer to include a build cache. I believe it should work, only not on a first build. [At least Travis stores the file.](https://travis-ci.org/infection/infection/jobs/354963155#L842) For example, [see here](https://travis-ci.org/sanmai/pipeline/jobs/354201230#L771) and [here](https://travis-ci.org/kalimatas/php-liquid/jobs/339277988#L728) where this scheme works.

.travis.yml
sanmai: Hard to tell why it does not in this specific build, but [see here](https://travis-ci.org/sanmai/pipeline/jobs/354201230#L771) and [here](https://travis-ci.org/kalimatas/php-liquid/jobs/339277988#L728) where the very same scheme works as expected.

theofidry: I think it needs to be an absolute path but the cache won't kick in with the PR, it needs to be merged to see it in effect

sanmai: Well, the opposite. [Docs say it must be relative](https://docs.travis-ci.com/user/caching#Arbitrary-directories) in our case.

PR #226, opened 2018-03-18
Enable SourceDirsProvider to accept a default value it guessed

### LINKED ISSUE #225: Infection config generator won't let a user input a default value it guessed
| Question    | Answer
| Infection version | dev-master

Infection config generator won't let a user input a default value it guessed from composer.json. 

 [For this composer.json](https://github.com/sanmai/pipeline/blob/master/composer.json):
```
"autoload": {
    "psr-4": {"Pipeline\\": "src/Pipeline/"}
},
```
It would look like this:
```
Which source directories do you want to include (comma separated)? [src/Pipeline]: 
  [0] .
  [1] build
  [2] src
  [3] tests
  [4] vendor
 > 
 Value "src/Pipeline" is invalid 
```

sanmai:
Fixes #225

### DIFF
--- a/src/Config/ValueProvider/SourceDirsProvider.php
+++ b/src/Config/ValueProvider/SourceDirsProvider.php
@@ -34,13 +34,11 @@ class SourceDirsProvider
         $this->questionHelper = $questionHelper;
     }
 
-    public function get(InputInterface $input, OutputInterface $output, array $dirsInCurrentDir): array
+    public function get(InputInterface $input, OutputInterface $output, array $dirsInCurrentDir, array $guessedSourceDirs = []): array
     {
         $output->writeln(['']);
 
-        $guessedSourceDirs = null;
-
-        if (file_exists('composer.json')) {
+        if (empty($guessedSourceDirs) && file_exists('composer.json')) {
             $content = json_decode(file_get_contents('composer.json'));
 
             if (json_last_error() !== JSON_ERROR_NONE) {
--- a/src/Config/ValueProvider/SourceDirsProvider.php
+++ b/src/Config/ValueProvider/SourceDirsProvider.php
@@ -28,26 +28,25 @@ class SourceDirsProvider
      */
     private $questionHelper;
 
-    public function __construct(ConsoleHelper $consoleHelper, QuestionHelper $questionHelper)
+    /**
+     * @var SourceDirGuesser
+     */
+    private $sourceDirGuesser;
+
+    public function __construct(ConsoleHelper $consoleHelper, QuestionHelper $questionHelper, SourceDirGuesser $sourceDirGuesser)
     {
         $this->consoleHelper = $consoleHelper;
         $this->questionHelper = $questionHelper;
+        $this->sourceDirGuesser = $sourceDirGuesser;
     }
 
-    public function get(InputInterface $input, OutputInterface $output, array $dirsInCurrentDir, array $guessedSourceDirs = []): array
+    public function get(InputInterface $input, OutputInterface $output, array $dirsInCurrentDir): array
     {
         $output->writeln(['']);
 
-        if (empty($guessedSourceDirs) && file_exists('composer.json')) {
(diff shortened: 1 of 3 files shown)


src/Config/ValueProvider/SourceDirsProvider.php
```diff
@@ -60,6 +58,11 @@ public function get(InputInterface $input, OutputInterface $output, array $dirsI
 
         $choices = array_merge(['.'], array_values($dirsInCurrentDir));
 
+        // if a default value is not any of current choices, include it as the first of allowed values
+        if (!in_array($defaultValues, $choices)) {
```
borNfreee: `$defaultValues` can contain more than just one guessed folders, e.g. `src/A,src/B`?

In this case, I guess it's not correct to insert this comma-separated value as a choice, because we will have

```
[ ] src/A,src/B
[ ] src
[ ] otherFolder
```

I think you should work here with `$guessedSourceDirs` array instead of string `$defaultValues`

Could you please add a test for this case?

src/Config/ValueProvider/SourceDirsProvider.php
```diff
@@ -34,13 +34,11 @@ public function __construct(ConsoleHelper $consoleHelper, QuestionHelper $questi
         $this->questionHelper = $questionHelper;
     }
 
-    public function get(InputInterface $input, OutputInterface $output, array $dirsInCurrentDir): array
+    public function get(InputInterface $input, OutputInterface $output, array $dirsInCurrentDir, array $guessedSourceDirs = []): array
```
borNfreee: Seems like this parameter `$guessedSourceDirs` is used only in unit test. Is it a mistake?

src/Config/ValueProvider/SourceDirsProvider.php
sanmai: Yes, I should have used an array. Fixed.

src/Config/ValueProvider/SourceDirsProvider.php
sanmai: That's correct. Right now there's no other way to give alternative guesses to the  SourceDirsProvider. They're tightly coupled with the SourceDirGuesser.

Do you think we should decouple them right this time?

borNfreee: Sorry, I still don't get it. The `$guessedSourceDirs` parameter is not used in the source code, only it the Unit Test. 

https://github.com/sanmai/infection/blob/0a320bcd2dbf4172821d1d7e1a8a9e1fe08e062f/src/Command/ConfigureCommand.php#L60

This is not correct. If it is not used, it should be removed.

> Right now there's no other way to give alternative guesses to the SourceDirsProvider

If you want to mock `SourceDirGuesser` in order to provide custom guessed values for unit tests - yes, it should be decoupled and injected. We should not create additional parameters just for unit tests.

src/Config/ValueProvider/SourceDirsProvider.php
```diff
@@ -60,6 +58,11 @@ public function get(InputInterface $input, OutputInterface $output, array $dirsI
 
         $choices = array_merge(['.'], array_values($dirsInCurrentDir));
 
+        // if a default value is not any of current choices, add its components to allowed values
+        if (!in_array($defaultValues, $choices)) {
```
borNfreee: shouldn't it be

```
foreach (guessedSourceDirs as $guessedSourceDir) {
    if (!in_array($defaultValues, $choices, true)) {
        $choices[] = $guessedSourceDir;
    }
}
```

?

What I mean here is we need to check all guessed values separately, not as concatenated string `src/A,src/B`

sanmai:
Got decoupled and injected. Is it any good now?

If there's too many commits, I'll merge them into a pair-you just tell me.

borNfreee:
> If there's too many commits, I'll merge them into a pair-you just tell me.

I always squash before merging

borNfreee:
Thank you @sanmai

PR #228, opened 2018-03-19
Add Assignment Mutator

sanmai:
Mutator replaces "+=", "*=", ".=", and similar with a plain "=". 

A very easy typo that may be hard to notice if there's not enough tests.

This PR:

- [x] Adds new feature: Assignment Mutator
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/28

PR #229, opened 2018-03-19
Add Assignment Equal Mutator

sanmai:
Replaces "==" with a "=".
 
It will only mutate if there's a variable on the left like here:
```php
if ($variable == BaseClass::CLASS_CONST) {
// mutates to... if ($variable = BaseClass::CLASS_CONST) {
```
This type of errors is main reason why this and other projects adopted Yoda notation for conditionals.

This PR:

- [x] Adds Assignment Equal Mutator
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/29

### DIFF
--- a/src/Mutator/Arithmetic/AssignmentEqual.php
+++ b/src/Mutator/Arithmetic/AssignmentEqual.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Mutator\Arithmetic;
+
+use Infection\Mutator\Mutator;
+use PhpParser\Node;
+
+class AssignmentEqual extends Mutator
+{
+    /**
+     * Replaces "==" with a "=".
+     *
+     * @param Node $node
+     *
+     * @return Node\Expr\Assign
+     */
+    public function mutate(Node $node)
+    {
+        return new Node\Expr\Assign($node->left, $node->right, $node->getAttributes());
+    }
+
+    public function shouldMutate(Node $node): bool
+    {
+        return $node instanceof Node\Expr\BinaryOp\Equal && $node->left instanceof Node\Expr\Variable;
+    }
+}
--- a/tests/Mutator/Arithmetic/AssignmentEqualTest.php
+++ b/tests/Mutator/Arithmetic/AssignmentEqualTest.php
@@ -48,7 +48,25 @@ if (1 == $a) {
 }
 PHP
                         ,
-                ],
+            ],
+            'It does not try to assign a variable to a class constant' => [
+                        <<<'PHP'
+<?php
+
+if (BaseClass::CLASS_CONST == $a) {
+}
+PHP
+                        ,
+            ],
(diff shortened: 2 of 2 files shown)


sanmai:
Should we add a separate subtype of mutators for this mutator and mutator from #228?

sanmai:
That's a good point. I made it so that it will only mutate if there's a variable on the left.

What we want to check that this mutation in a conditional is correctly caught. Should be more obvious from the test case now.

If that's okay now, I'll proceed to the docs.

sanmai:
@theofidry Have you been looking at the outdated changes?

This mutator won't mutate if there isn't a variable on the left. So it seems like it can't cause a straight on error in any case I can think of.

BackEndTea:
> This mutator won't mutate if there isn't a variable on the left. So it seems like it can't cause a straight on error in any case I can think of.

How does it handle checking against class constants? for example:

```php
if (BaseClass::CLASS_CONST == $variable) {
```

sanmai:
It won't mutate in this case. It will only mutate if there's a variable on the left like here:
```php
if ($variable == BaseClass::CLASS_CONST) {
```

sanmai:
There's such a case with a constant on the left where it does not mutate.

https://github.com/infection/infection/pull/229/commits/8900c17740264ebe95d9bafecbfa160698ac62f6#diff-31453075d710a632621d2f818f13e501R43

BackEndTea:
I've seen that, but i'd like to see two added tests.

One against a custom class constant, ` BaseClass::CLASS_CONST` for example, and a test case with a php built in constant like `PHP_EOL`.

I know it already works correctly, but it makes sure that our mutators keep this behavior, even if, for example, the `PhpParser` api would change.

sanmai:
Right, it's good to have a solid test suite if anyone ever wants to refactor this mutator. Done.

theofidry:
@sanmai no I just didn't see a test for it so I didn't assume this was checked :) If it's working all cool but I would add a test for it: it's very easy to miss a type of node so a quick refacto could easily break this without the tests failing

src/Mutator/Arithmetic/AssignmentEqual.php
```diff
@@ -0,0 +1,33 @@
...
+     * @return Node\Expr\Assign
+     */
+    public function mutate(Node $node)
+    {
+        return new Node\Expr\Assign($node->left, $node->right, $node->getAttributes());
+    }
+
+    public function shouldMutate(Node $node): bool
+    {
+        return $node instanceof Node\Expr\BinaryOp\Equal && $node->left instanceof Node\Expr\Variable;
```
theofidry: does that work if you have `$x->getFoo() == $a`? `Variable` doesn't look wide enough for it. Maybe we don't want to cover this case tbh, but I would find it interesting as this is where (im)mutability can important

BackEndTea: `$x->getFoo() = $a` will always error

sanmai: Here you go: 1f19d3a

theofidry: 🤔 yeah you're right. Now sure where I saw something like that I must be confusing with something else

borNfreee:
Thank you @sanmai 👍

PR #230, opened 2018-03-19
Add a For Loop mutator & test

jesperklippel:
This PR:

- [x] Adds new feature: For Loop Mutator
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/27

This mutator serves the same purpose as the `ForEach_` mutator, except for For Loops.

### DIFF
--- a/src/Mutator/ZeroIteration/For_.php
+++ b/src/Mutator/ZeroIteration/For_.php
@@ -1,0 +1,34 @@
+<?php
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Mutator\ZeroIteration;
+
+use Infection\Mutator\Mutator;
+use PhpParser\Node;
+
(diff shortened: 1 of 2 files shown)


src/Mutator/ZeroIteration/For_.php
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Mutator\ZeroIteration;
+
+use Infection\Mutator\Mutator;
+use PhpParser\Node;
+
+class For_ extends Mutator
```
theofidry: Do we need the trailing `_`?

jesperklippel: For is a reserved word by PHP, so it can't be used as a class name.

theofidry: `ForCondition`?

BackEndTea: `For_` is in line with all our other mutators

theofidry: Fair then :/

borNfreee:
Thank you @jesperklippel

PR #232, opened 2018-03-20
InfectionConfig::getLogsTypes() - force operator order

sanmai:
Else with a missing "logs" property in the config file a notice will come up:

> PHP Notice:  Undefined property: stdClass::$logs in src/Config/InfectionConfig.php on line 226

An example of a configuration causing the above error:

    {
        "source": {
            "directories": [
                "src/"
            ]
        }
    }

- [x] Covered by tests


borNfreee:
Thank you @sanmai

PR #235, opened 2018-03-20
Refactor MutationsCollectorVisitor

### LINKED ISSUE #234: Infection (?) marks covered code as not covered
| Question    | Answer
| Infection version | master
| Test Framework version | PHPUnit 6.5.7
| PHP version | PHP 7.0.27-1
| Platform    |  Linux
| Github Repo |  [sanmai/LazyLINQ](https://github.com/sanmai/LazyLINQ)

Steps to reproduce:

```bash
git clone https://github.com/sanmai/LazyLINQ 
cd LazyLINQ
composer install
```

Now run tests to verify 100% coverage. And run Infection.
```bash
$ php vendor/bin/phpunit

Summary:                   
  Classes: 100.00% (1/1)    
  Methods: 100.00% (43/43)  
  Lines:   100.00% (181/181)

$ php vendor/bin/infection --log-verbosity=2 --threads=$(nproc)

Processing source code files: 2/2
Creating mutated files and processes: 151/151

........SS............S.....S.....................   ( 50 / 151)
.......E....S..SS..E..............S...............   (100 / 151)
S......S..........S...S...........................   (150 / 151)
.                                                    (151 / 151)

151 mutations were generated:
     137 mutants were killed
      12 mutants were not covered by tests
       2 errors were encountered
       0 time outs were encountered

Metrics:
         Mutation Score Indicator (MSI): 92%
         Mutation Code Coverage: 92%
         Covered Code MSI: 100%
```
It'll show there were not covered mutants. Indeed we can see them in the log:

```bash
$ grep -A10 'Not covered mutants' infection-log.txt 
Not covered mutants:
====================

1) src/LazyLINQ/LazyCollection.php:265    [M] PublicVisibility

--- Original
+++ New
@@ @@
-    public function distinct(callable $comparer = null)
+    protected function distinct(callable $comparer = null)
```

But if we apply this mutant by hand:
```bash
sed -i 's/public function distinct/protected function distinct/g' src/LazyLINQ/LazyCollection.php
```
We'll get a failing test:

```bash
$ php vendor/bin/phpunit

There was 1 error:

1) LazyLINQ\LazyCollectionTest::testDistinct
Error: Call to protected method LazyLINQ\LazyCollection::distinct() from context 'LazyLINQ\LazyCollectionTest'
```

Clearly, this mutant isn't "not covered" by tests.

sanmai:
* Tag mutations as non-covered as early as we know it. Don't do it again later on.
* Exit early if a mutant can't mutate in MutationsCollectorVisitor.
* Tests updated to account for updated interface of a Mutation

- [x] Covered by tests

### DIFF
--- a/src/Visitor/MutationsCollectorVisitor.php
+++ b/src/Visitor/MutationsCollectorVisitor.php
@@ -63,6 +63,10 @@ class MutationsCollectorVisitor extends NodeVisitorAbstract
     public function leaveNode(Node $node)
     {
         foreach ($this->mutators as $mutator) {
+            if (!$mutator->shouldMutate($node)) {
+                continue;
+            }
+
             $isOnFunctionSignature = $node->getAttribute(ReflectionVisitor::IS_ON_FUNCTION_SIGNATURE, false);
 
             if (!$isOnFunctionSignature) {
@@ -80,29 +84,31 @@ class MutationsCollectorVisitor extends NodeVisitorAbstract
                 }
             }
 
-            if ($this->onlyCovered) {
-                if ($isOnFunctionSignature &&
-                    !$this->codeCoverageData->hasExecutedMethodOnLine($this->filePath, $node->getLine())) {
-                    continue;
-                }
+            $isCoveredByTest = $this->isCoveredByTest($isOnFunctionSignature, $node);
 
-                if (!$isOnFunctionSignature &&
-                    !$this->codeCoverageData->hasTestsOnLine($this->filePath, $node->getLine())) {
-                    continue;
-                }
+            if ($this->onlyCovered && !$isCoveredByTest) {
+                continue;
             }
 
-            if ($mutator->shouldMutate($node)) {
-                $this->mutations[] = new Mutation(
-                    $this->filePath,
-                    $this->fileAst,
-                    $mutator,
-                    $node->getAttributes(),
-                    get_class($node),
-                    $isOnFunctionSignature
-                );
-            }
+            $this->mutations[] = new Mutation(
+                $this->filePath,
+                $this->fileAst,
+                $mutator,
+                $node->getAttributes(),
+                get_class($node),
+                $isOnFunctionSignature,
+                $isCoveredByTest
+            );
         }
+    }
+
+    private function isCoveredByTest(bool $isOnFunctionSignature, Node $node): bool
+    {
+        if ($isOnFunctionSignature) {
+            return $this->codeCoverageData->hasExecutedMethodOnLine($this->filePath, $node->getLine());
+        }
+
+        return $this->codeCoverageData->hasTestsOnLine($this->filePath, $node->getLine());
     }
 
     /**
--- a/tests/Mutant/Generator/MutationsGeneratorTest.php
+++ b/tests/Mutant/Generator/MutationsGeneratorTest.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 
 namespace Infection\Tests\Mutant\Generator;
 
+use Infection\Config\Exception\InvalidConfigException;
 use Infection\EventDispatcher\EventDispatcher;
 use Infection\Mutant\Generator\MutationsGenerator;
 use Infection\Mutator\Arithmetic\Decrement;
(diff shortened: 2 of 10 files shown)


src/Visitor/MutationsCollectorVisitor.php
```diff
@@ -63,6 +63,10 @@ public function __construct(
     public function leaveNode(Node $node)
     {
         foreach ($this->mutators as $mutator) {
+            if (!$mutator->shouldMutate($node)) {
```
borNfreee: I would move it below the next 2 `if`s because `shouldMutate()` sometimes is quite expensive operation, for example see this one that uses reflection and iterates over parent classes/interfaces: https://github.com/infection/infection/blob/master/src/Mutator/FunctionSignature/PublicVisibility.php#L68

On the other side, things like  `isOnFunctionSignature` and `$node->getAttribute(ReflectionVisitor::IS_INSIDE_FUNCTION_KEY)` are just getters for already calculated values

tests/Mutant/Generator/MutationsGeneratorTest.php
```diff
@@ -28,7 +29,7 @@ public function test_it_collects_plus_mutation()
     {
         $codeCoverageDataMock = Mockery::mock(CodeCoverageData::class);
         $codeCoverageDataMock->shouldReceive('hasTestsOnLine')->andReturn(true);
-        $codeCoverageDataMock->shouldReceive('isLineFunctionSignature')->andReturn(false);
+        $codeCoverageDataMock->shouldReceive('hasExecutedMethodOnLine')->andReturn(false);
```
borNfreee: As I see, the method `isLineFunctionSignature()` does not exists at all, and test is still green

Let's add `->once()` or appropriate method for all such cases (here and in other test cases) to ensure methods are executed needed times.

For this line it will be `$codeCoverageDataMock->shouldReceive('hasExecutedMethodOnLine')->once()->andReturn(false);`

src/Visitor/MutationsCollectorVisitor.php
sanmai: I've run this change under `perf`, and I've run it again against the Infection codebase, and on five runs `perf` showed me average run time of 26.63 second for the currently implement variant, and 33.44 seconds for the variant you're suggestion. In both cases the standard deviation was about 5% or about 2 seconds.

Do you still think we have to move it below?

sanmai: We can micro-optimize here if we distinguish fast and slow mutators by their parent class or else. I'm not sure it's the right time, though.

tests/Mutant/Generator/MutationsGeneratorTest.php
sanmai: Done.

src/Visitor/MutationsCollectorVisitor.php
sanmai: If we are looking at those two suspects only:

```bash
vendor/bin/phpunit --coverage-xml=coverage/coverage-xml --log-junit=coverage/phpunit.junit.xml
perf stat -r 20 php bin/infection --quiet --coverage=coverage --threads=12 --mutators=ProtectedVisibility,PublicVisibility
```
For a case where `shouldMutate` comes first I get: 2.234926127 seconds ( +-  0,21% )
For a case where `shouldMutate` goes below next 2 `if`s: 2.292622426 seconds  ( +-  0,29% )

SD is about same in both cases, and difference is about 2%. Which makes me think FunctionSignature mutators aren't the bottleneck here...

borNfreee: Thanks for the explanation. Probably the bottleneck is `$mutator->isIgnored()` function. Have you tried to move `shouldMutate()` after the next `if` but before the `if` with `isIgnored()`?

Anyway, I'm happy to merge as is.

tests/Mutant/MutantCreatorTest.php
```diff
@@ -36,6 +36,7 @@ public function test_it_uses_avaialable_file_if_hash_is_the_same()
         $mutation->shouldReceive('getOriginalFileAst')->andReturn(['ast']);
         $mutation->shouldReceive('getAttributes')->andReturn(['startLine' => 1]);
         $mutation->shouldReceive('isOnFunctionSignature')->andReturn(true);
+        $mutation->shouldReceive('isCoveredByTest')->andReturn(true);
```
borNfreee: please add here `->once()` or something similar

sanmai: Done

src/Visitor/MutationsCollectorVisitor.php
sanmai: My poor man's instrumentation shows that there isn't a single true bottleneck. 

| Method                | ms per 1000  | 
| --------------------- | -------------|
| isCoveredByTest       | 1.08         | 
| isIgnored             | 0.99         | 
| isOnFunctionSignature | 0.49         |
| shouldMutate          | 0.348        |

What `leaveNode` does is a negative selection.

Therefore, only having average timings we can make a decision: the fastest function to exclude a given mutator should come first, and slowest the last. This way the negative selection will happen in the most efficient way.

It appears that by sheer luck or providence the current layout is optimal.

borNfreee: Great, thank you!

sanmai: What I also discovered is that `MutationsCollectorVisitor` gets constructed a lot of times. Probably once per each thread. Therefore we can get at least partial speed-up if we reuse result of MutationsCollectorVisitor in further invocations. It seems to remain a constant between invocations.

borNfreee:
> What I also discovered is that MutationsCollectorVisitor gets constructed a lot of times. 

Once for each mutated file.

> Probably once per each thread. 

Mutations are generated in the main thread, so I'm not sure I got you right.

PR #236, opened 2018-03-20
Fix a bug where new mutators weren't added to Config\InfectionConfig.php.

jesperklippel:
This PR fixes a bug where Assignment, AssignmentEqual and For Loop mutator weren't added to ```Config\InfectionConfig.php```

### DIFF
--- a/src/Config/InfectionConfig.php
+++ b/src/Config/InfectionConfig.php
@@ -9,6 +9,8 @@ declare(strict_types=1);
 
 namespace Infection\Config;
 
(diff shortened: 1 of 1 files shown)


borNfreee:
Can we have a test to avoid same in the future?

I'm not sure what is the best way to do it, but as for me - it would be ok traversing by Finder in all mutator folders and checn classes are in `DEFAULT_MUTATORS` constant.

borNfreee:
Merging this one, extracted my previous comment to https://github.com/infection/infection/issues/237

Thank you @jesperklippel

PR #240, opened 2018-03-21
ArrayItem Mutator

sanmai:
This PR:

- [x] Adds an ArrayItem Mutator
- [x] Covered by tests
- [x] Added to a relevant profile 
- [x] Doc PR: https://github.com/infection/site/pull/32

This mutator replaces `[$a->foo => $b->bar]` with `[$a->foo > $b->bar]` where one can reasonably expect side-effects to happen. 

Precisely, only those array items will be mutated which have either a function call, a method call, or a property fetch on either side of an item.

```php
[$a->foo => $b->bar];
[$a->foo() => bar()];
[foo() => $b->bar];
```
    
Will mutate to:

```php
[$a->foo > $b->bar];
[$a->foo() > bar()];
[foo() > $b->bar];
```

### Dogfooding
```
$ softlimit -m 500000000 bin/infection --threads=$(nproc) --mutators=ArrayItem

<skip>

S...M....M.M.M..M.M                                  (19 / 19)

19 mutations were generated:
      12 mutants were killed
       1 mutants were not covered by tests
       6 covered mutants were not detected
       0 errors were encountered
       0 time outs were encountered

Metrics:
         Mutation Score Indicator (MSI): 63%
         Mutation Code Coverage: 95%
         Covered Code MSI: 67%

```

That's 6 escaped mutants for the Infection itself.

### DIFF
--- a/src/Mutator/Boolean/ArrayItem.php
+++ b/src/Mutator/Boolean/ArrayItem.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Mutator\Boolean;
+
+use Infection\Mutator\Util\Mutator;
+use PhpParser\Node;
+
+class ArrayItem extends Mutator
+{
+    /**
+     * Replaces [$a->foo => $b->bar] with [$a->foo > $b->bar]
+     *
+     * @param Node $node
+     *
+     * @return Node\Expr\BinaryOp\Greater
+     */
+    public function mutate(Node $node)
+    {
+        return new Node\Expr\BinaryOp\Greater($node->key, $node->value, $node->getAttributes());
+    }
+
+    public function shouldMutate(Node $node): bool
+    {
+        return $node instanceof Node\Expr\ArrayItem && $node->key && ($this->isNodeWithSideEffects($node->value) || $this->isNodeWithSideEffects($node->key));
+    }
+
+    private function isNodeWithSideEffects(Node $node)
+    {
+        return
+            // __get() can have side-effects
+            $node instanceof Node\Expr\PropertyFetch ||
+            // these clearly can have side-effects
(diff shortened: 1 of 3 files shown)


borNfreee:
Thanks for creating this PR @sanmai.

I will ask a couple of questions to analyze this mutator:

1. Since we mutate `[$a => $b]` to `[%value%]`, we loose the previously existed key. This is great because when we have an explicit key in an array (`$a` can be a string or meaningful integer), it's created for some reason and tests must cover/check it. So mutating key `$a` to any integer value (by default it's `0`) is a good mutation.

2. The result of `$a => $b` is always a boolean: `true` or `false`, isn't it? The question here is why don't we mutate directly to let's say `false` here?

For example, if we have an alphabet map: `['a' => 1, 'b' => 2, ....,]`, we will end up with such mutation:

```diff
- ['a' => 1, 'b' => 2, ....,]
+ ['a' > 1, 'b' => 2, ....,]
```

which equals to `[false, 'b' => 2, ....,]`. So we can directly mutate to `false` instead of `'a' > 1`

3. Since (from my point of view) the main advantage of this mutator is that the key is being dropped, we can even mutate like `[$a => $b]` -> `[$b]`

What do you think?

sanmai:
The simplified mutations you're speaking of we're on my mind, but I decided against them, at least for now. We don't we mutate directly to `false` or `null` here because it is boring. Same goes for a key-removing mutation. It is boring because we'd loose all side effects. 

On the contrary, this mutation keeps all side-effects intact.

E.g. here:

```bash
- [$foo->getA() => $bar->getB()]
+ [$foo->getA() > $bar->getB()]
```

borNfreee:
> this mutation keeps all side-effects intact

very nice example, makes sense!

The last question: should we leave primitive mutation the same way `['string' => 1]` -> `['string' > 1]` or just mutate to `[false/true/null]` them? In this case, there were no side effects.

sanmai:
That's a good point. I think it would be best if this mutator would leave out the bulk of primitive mutations. I'll see to it.

sanmai:
Done. 

When we excluded primitive mutations, we removed one escaped mutation. So primitive mutations may still be worth considering, but without any great expectations.

PR #241, opened 2018-03-21
Removed hard-coded constants from Mutators generator test 

sanmai:
This PR:

- [x] Covered by tests

### DIFF
--- a/tests/Mutator/Util/MutatorsGeneratorTest.php
+++ b/tests/Mutator/Util/MutatorsGeneratorTest.php
@@ -13,17 +13,27 @@ use Infection\Config\Exception\InvalidConfigException;
 use Infection\Mutator\Arithmetic\Plus;
 use Infection\Mutator\Boolean\FalseValue;
 use Infection\Mutator\Boolean\TrueValue;
+use Infection\Mutator\Util\MutatorProfile;
 use Infection\Mutator\Util\MutatorsGenerator;
 use PHPUnit\Framework\TestCase;
 
 class MutatorsGeneratorTest extends TestCase
 {
+    private static $countDefaultMutators = 0;
+
+    public static function setUpBeforeClass()
+    {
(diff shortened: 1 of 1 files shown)


BackEndTea:
:+1: This is cleaner than my solution.

tests/Mutator/Util/MutatorsGeneratorTest.php
```diff
@@ -13,17 +13,29 @@
 use Infection\Mutator\Arithmetic\Plus;
 use Infection\Mutator\Boolean\FalseValue;
 use Infection\Mutator\Boolean\TrueValue;
+use Infection\Mutator\Util\MutatorProfile;
 use Infection\Mutator\Util\MutatorsGenerator;
 use PHPUnit\Framework\TestCase;
 
 class MutatorsGeneratorTest extends TestCase
 {
+    private function countMutatorsFromDefaultProfile()
```
BackEndTea: I'd move this to a `setUp` function.

sanmai: Done.

tests/Mutator/Util/MutatorsGeneratorTest.php
```diff
@@ -13,17 +13,29 @@
...
 use Infection\Mutator\Util\MutatorsGenerator;
 use PHPUnit\Framework\TestCase;
 
 class MutatorsGeneratorTest extends TestCase
 {
+    private $countDefaultMutators;
+
+    protected function setUp()
+    {
+        $this->countDefaultMutators = 0;
```
theofidry: from what I see the list doesn't change between each test so either you should initialize the value only when `$countDefaultMutators === null` or do that in the `beforeClass` something method which is called only once

sanmai: Indeed it doesn't. Done.

borNfreee:
Thank you @sanmai

PR #242, opened 2018-03-21
Yield Mutator

sanmai:
Replaces `yield $a => $b;` with `yield $a > $b;`

This PR:

- [x] Adds Yield Mutator
- [x] Covered by tests
- [x] Added to a relevant profile (blocked by #241)
- [x] Doc PR: https://github.com/infection/site/pull/31

### DIFF
--- a/src/Mutator/Boolean/Yield_.php
+++ b/src/Mutator/Boolean/Yield_.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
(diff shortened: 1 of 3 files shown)


BackEndTea:
Related to this: i'm also in favor of a `yield ...` to `return ...`, if it isn't typehinted to `\Generator`

sanmai:
The last part about not being typehinted seems difficult. But you should try to do it nevertheless.

borNfreee:
Thank you @sanmai

Issue #208, opened 2018-03-05
Not easy to get package version


lchrusciel:
| Question    | Answer
| ------------| ---------------
| Infection version | dev-master, local git distribution
| Test Framework version | doesn't matter
| PHP version | 7.2
| Platform    | MacOS
| Github Repo | -

Right now, checking a version of infection tells me, that I have `@package.version@` version of an app.
![zrzut ekranu 2018-03-05 o 20 40 54](https://user-images.githubusercontent.com/6213903/36995982-eb83ad78-20b5-11e8-8f2f-6ab7d45cb3a5.png)

BackEndTea:
@sidz @theofidry I think you guys created the part that deals with package version. Do you know what is  going on?

This isn't just for dev-master, it doesn't work on 0.8.1 either.

theofidry:
As far as I'm aware the version is simply not replaced when running infection from the package as opposed to the PHAR. The string `@package_version@` is replaced when building the PHAR.

BackEndTea:
@borNfreee how about changing `const VERSION = '@package_version@';` with `const VERSION = '0.9.0-dev';` and then removing the `-dev` for the release.

This does mean it has to be manually changed for every release.

theofidry:
I'm not really for managing it manually personally. That said we could replace the `@package_version` placeholder by something else when outside the PHAR. You can do something like in [Box](https://github.com/humbug/box/blob/master/src/Console/Application.php#L48-L60) or another way is just not using a placeholder at all like done in [PHP-Scoper](https://github.com/humbug/php-scoper/blob/master/src/functions.php#L49)

sidz:
Do we really need to take care about it? Composer doesn't do it. 

![image](https://user-images.githubusercontent.com/1302230/37273409-d2c781bc-25ea-11e8-8d98-15bfefe4b23f.png)

theofidry:
@sidz is 99% if not more installed as a PHAR, Infection is way more used as a library

PR #250, opened 2018-03-22
Show installed version when Infection has been installed as dependency

sidz:
This PR fixes https://github.com/infection/infection/issues/208

I've implemented new dependency to take the current installed version. If you're not happy with it we can implement the same solution like @theofidry did in [Box](https://github.com/humbug/box/blob/master/src/Console/Application.php#L48-L60) - just remove package version when Infection installed as dependency

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,6 +41,7 @@
     "require": {
         "php": "^7.0",
         "nikic/php-parser": "^4.0",
+        "ocramius/package-versions": "^1.2",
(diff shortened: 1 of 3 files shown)


theofidry:
@borNfreee by [not doing it](https://github.com/humbug/box/blob/master/src/Console/Application.php#L59) ^^

borNfreee:
Thank you @sidz

PR #252, opened 2018-03-23
Add a mandatory memory limit for PHPUnit sub-processes

### LINKED ISSUE #247: Mutant processes are not memory-bound
| Question    | Answer
| Infection version | master

```
```

## Steps to reproduce

        checkout a5e034f

    ```diff
    --- a/infection.json.dist
    +++ b/infection.json.dist
    @@ -1,5 +1,5 @@
     {
    -    "timeout": 25,
   +    "timeout": 2500,
    ```

        bin/infection -j4

## Workarounds

    ulimit -Sv 500000

sanmai:
This PR:

- [x] Adds a mandatory memory limit for PHPUnit sub-processes
- [x] Covered by tests
- [ ] Doc PR: ?

### Rationale

Infection is a command line program. Command line programs are expected to behave. Because they are expected to behave, there are usually no particular limits in terms of memory consumption. Command line programs are responsible for the commands they launch. So must be Infection. 

Currently there is no limits on how much memory a child process may consume other than by time. Since the main function of Infection is to cause unexpected bugs in programs it runs, there must be at least some limits in place to make sure new bugs do not cause more damage than they should.

There's a default time limit of 20 seconds. Current programs have no problem consuming many gigabytes of RAM in the allotted time limit. Therefore, it is not enough to have a time limit only.

### Methodology

Memory limit is introduced by altering `phpunit.xml` to include it. 

```
<php>
    <ini name="memory_limit" value="128M"/>
</php>
```

The memory limit is default to 128Mb, which is an accepted default for most web apps.

If there is a memory limit already defined, it won't be overridden. Therefore, this feature requires no configuration whatsoever. 

Fixes #247

### DIFF
--- a/src/Config/InfectionConfig.php
+++ b/src/Config/InfectionConfig.php
@@ -14,6 +14,7 @@ use Symfony\Component\Filesystem\Filesystem;
 class InfectionConfig
 {
     const PROCESS_TIMEOUT_SECONDS = 10;
+    const PROCESS_MEMORY_LIMIT = '128M';
     const DEFAULT_SOURCE_DIRS = ['.'];
     const DEFAULT_EXCLUDE_DIRS = ['vendor'];
     const CONFIG_FILE_NAME = 'infection.json';
--- a/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
+++ b/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
@@ -52,6 +52,7 @@ class MutationConfigBuilder extends ConfigBuilder
         $xmlConfigurationHelper->deactivateColours($this->xPath);
         $xmlConfigurationHelper->removeExistingLoggers($this->dom, $this->xPath);
         $xmlConfigurationHelper->removeExistingPrinters($this->dom, $this->xPath);
+        $xmlConfigurationHelper->addMemoryLimit($this->xPath, $this->dom);
     }
 
     public function build(Mutant $mutant): string
--- a/src/TestFramework/PhpUnit/Config/XmlConfigurationHelper.php
+++ b/src/TestFramework/PhpUnit/Config/XmlConfigurationHelper.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 
 namespace Infection\TestFramework\PhpUnit\Config;
 
+use Infection\Config\InfectionConfig;
 use Infection\TestFramework\PhpUnit\Config\Path\PathReplacer;
 
 class XmlConfigurationHelper
@@ -79,5 +80,26 @@ class XmlConfigurationHelper
         if ($nodeList->length) {
             $dom->documentElement->removeAttribute('printerClass');
         }
+    }
+
+    public function addMemoryLimit(\DOMXPath $xPath, \DOMDocument $dom)
+    {
+        if ($xPath->query('/phpunit/php/ini[@name="memory_limit"]')->length) {
+            return;
+        }
+
+        $nodeList = $xPath->query('/phpunit/php');
+
+        if ($nodeList->length) {
+            $node = $nodeList[0];
+        } else {
+            $node = $dom->createElement('php');
+            $xPath->query('/phpunit')[0]->appendChild($node);
+        }
+
+        $element = $dom->createElement('ini');
+        $element->setAttribute('name', 'memory_limit');
+        $element->setAttribute('value', InfectionConfig::PROCESS_MEMORY_LIMIT);
+        $node->appendChild($element);
     }
 }
--- a/tests/Fixtures/Files/phpunit/phpunit_with_ini_set.xml
+++ b/tests/Fixtures/Files/phpunit/phpunit_with_ini_set.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit>
+    <php>
+        <ini name="example" value="test"/>
+    </php>
+</phpunit>
--- a/tests/Fixtures/Files/phpunit/phpunit_with_memory_limit.xml
+++ b/tests/Fixtures/Files/phpunit/phpunit_with_memory_limit.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit>
+    <php>
+        <ini name="memory_limit" value="64M"/>
+    </php>
+</phpunit>
--- a/tests/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilderTest.php
+++ b/tests/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilderTest.php
@@ -113,18 +113,25 @@ class MutationConfigBuilderTest extends Mockery\Adapter\Phpunit\MockeryTestCase
         $this->assertSame($expectedCustomAutoloadFilePath, $resultAutoLoaderFilePath);
     }
 
+    private function builderFromFile(string $phpunitXmlPath): MutationConfigBuilder
+    {
+        $replacer = new PathReplacer($this->fileSystem, $this->pathToProject);
+        $xmlConfigurationHelper = new XmlConfigurationHelper($replacer);
+
+        return new MutationConfigBuilder(
+            $this->tmpDir,
+            file_get_contents($phpunitXmlPath),
+            $xmlConfigurationHelper,
+            $this->pathToProject
+        );
+    }
+
     public function test_it_sets_custom_autoloader_when_attribute_is_absent()
     {
         $this->mutant->shouldReceive('getCoverageTests')->andReturn([]);
         $phpunitXmlPath = __DIR__ . '/../../../../Fixtures/Files/phpunit/phpuit_without_bootstrap.xml';
-        $this->builder = new MutationConfigBuilder(
-            $this->tmpDir,
-            file_get_contents($phpunitXmlPath),
-            $this->xmlConfigurationHelper,
-            'project/dir'
-        );
 
-        $configurationPath = $this->builder->build($this->mutant);
+        $configurationPath = $this->builderFromFile($phpunitXmlPath)->build($this->mutant);
 
         $xml = file_get_contents($configurationPath);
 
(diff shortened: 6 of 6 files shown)


theofidry:
And this is were I'm unhappy :(

>Because they are expected to behave, there are usually no particular limits in terms of memory consumption

Here's a list of languages for which it is not the case out of the box (unless the distro tweaked them):

- PHP
- node.js
- Python does not have a limit per se, but it does accommodate based on the free resources so it won't make everything crash by being too greedy
- Java
- ...

So I neither feel this is a universal truth and in practice I also find it dangerous: you don't want a buggy program to be able to crash your whole computer because of a memory leak.

Now two other reasons why I'm against it:

- This feature already exists at the language level and that is `memory_limit`. I don't understand why we should ditch it and push more complexity at an application level to compensate.
- The current solution completely discard any physical capabilities of the machine used and enforced for every user of a given project. For example my CI machine is way more powerful than my computer, so we disable garbage collection and have a higher memory limit. With your configuration, you require us to either make it configurable via a CLI option (which is yet another config to add) or create another configuration file for the CI (which is yet another extra config to add and maintain as well)

However since it's the default on a lot of Ubuntu & Debian distrib, I think it warrants at the very least a warning message when no memory limit is used:

```
Warning: you are running Infection without a memory limit (`memory_limit=-1`). This can cause various
issues if your machine runs out of memory. You should either set the limit in your `php.ini` file or run
this process with `php -dmemory_limit=XM`
```

sanmai:
What are you caring about most? I care not about my principles, which are aligned with yours as far as I see, but more about general success of this project. Can you tell, will this project be more *successful* if we chose to implement this change (or #253), or if we chose to add a big fat warning?

Well, that seems to be a difficult question, but if we'll be putting forward logically valid arguments leading before our wants and needs, and make a sound cost-benefit analysis, we should be able to resolve this question with ease.

#### If we chose to implement this change...

Who is going to benefit from this change?
- All new users, who are used to 128Mb limit, common to most web apps. They would not have to read any warnings, and everything will work out of box not trying to smash their OS and work.
- General populace of software engineers caring about work being done, bugs being fixed, and so on, without a need to fidget with their configs, adjusting them especially for her majesty Infection.
- All users caring about speed. They wouldn't have to wait until their tests hit physical limits of their tech, because a test will fail as soon as it requests more than a configured amount of RAM.

Who will be at a disadvantage?
- Those rare software engineers who happen to need more than 128Mb of RAM for their tests to pass. They would have to spend fifteen minutes of research to add a line to a config file for this program. Just once.
- The list ends here.

#### If we chose to add a warning here and there, and be done...

Who is going to benefit from this change?
- Those rare software engineers who happen to need more than 128Mb of RAM for their tests to pass. They wouldn't have to adjust anything. Hooray!
- That's about it. Nobody else.

Who will be at a disadvantage?
- All new users. They would have to read warnings, adjust unnaturally their configs for PHP CLI, or risk finding Infection working hard to smash their OS and crash all very important programs they use with the OOM Killer.
- All other software engineers caring about work being done, bugs being fixed, and so on, will need to constantly care about changes they made to php.ini, sometimes stumbling upon then, unable to make a yearly report, that suddenly needs more than 128Mb we forced them to hardcode into php.ini with our big fat warnings. They are not going to be happy. They will blame Infection for all their trouble, that's sure.
- All people caring about speed. Sure, they have beefy tech with lots of RAM, and they will be punished. They will have to wait for each test to time out. Every time. No matter what. Well, at least their other programs like IDE will be in relative safety.

sanmai:
@theofidry BTW, this isn't helping:
    
    php -dmemory_limit=128M bin/infection

You can try by yourself on the HEAD, which is a5e034f.

theofidry:
@sanmai it's not about principle, it's about common sense. You can across this case with Infection, my point is this can happen with everything even just PHPUnit or pretty much any PHP CLI with those settings. And I hope if that happens to someone locally in dev which surely will piss the heck out of that person (I would be for sure as well), it's good enough of a lesson to 1. fix the issue and 2. ensure this is not the case on your production server especially if you have a queue system, cron jobs or running tests on the same server as the prod which can cause way more damage.

So while your solution is reasonable it looks to me only like a patch to a massive gaping hole which is way more threatening.

That said, if I'm the only one feeling that way fair enough. I don't think there is much more to express on my end regarding this matter so the following will be more on the technical aspect on this PR.

<hr/>

If accepted, this setting _must_ be settable via an option. Forcing to tweak the file for changing the default value is a no go as this would require some to handle to play with multiple config files depending of their machine resources as I mentioned above.

Maybe this should be exposed via an environment variable as well like done in Composer.

>The memory limit is default to 128Mb

I would put at least 256 or 512. If you have a few basic unit tests sure 129MB is plenty, but as soon as you have a few integration tests (e.g. the Behat integration tests on a big project I have take up to 2GB). I think beyond that is debatable although I see that [Composer is pushing it to 1.5GB](https://github.com/composer/composer/blob/master/bin/composer#L44). From your comment you seem to be thinking that 128MB is what people are familiar with. If most ubuntu users are used to the `-1` limit for CLI it's fair enough, but in my experience people put a much higher limit for the CLI 
 (although not `-1`). I don't doubt that 128MB remains quite common for the web config though.

>BTW, this isn't helping:
>php -dmemory_limit=128M bin/infection

Seems like a legit but to me :)

sanmai:
> If accepted, this setting must be settable via an option.

See #253, it has an adjustable setting there right in `infection.json.dist`. I think that #253 is way better because it works for everything, not just for PHPUnit.

I think whenever we set 128Mb or 256Mb is an open question. I don't mind it being set to 256Mb or else, but we have to start somewhere. It seems reasonable to start at 128Mb and see how it goes.

theofidry:
>See #253, it has an adjustable setting there right in infection.json.dist

My point is that it's not enough to have an acceptable DX either. But actually my original suggestion does not look right either.

What about the following: we apply that memory limit _only_ if the memory limit is `-1` and not forced. This would lead to the following scenarios:

```
memory_limit=-1
infection.json.dist ->nothing
$ infection  # limited to 128MB

memory_limit=-1
infection.json.dist: 512MB
$ infection  # limited to 512MB

memory_limit=-1
infection.json.dist: -1
$ infection  # unlimited

### Now would also be configurable that way:

memory_limit=X
infection.json.dist: Y
$ php -dmemory_limit=512M infection  # limited to 512MB

memory_limit=X
infection.json.dist: Y
$ php -dmemory_limit=-1 infection  # unlimited
```

*`X`, `Y` being any value, unlimited or not

>I think whenever we set 128Mb or 256Mb is an open question

[Infection is a few MB close](https://travis-ci.org/infection/infection/jobs/357431138#L735), [API-Platform is at 150MB](https://travis-ci.org/api-platform/core/jobs/357451903) just to name a few. Composer is pushing it to 1.5GB. So I think we can skip the 128MB step and go to the next one right away. The goal as you stated is to prevent a crash, not annoy the users with some extra config.

theofidry:
@borNfreee this whole thing also should be a warning for all of us. Imagine you run Infection on integration tests and a mutant change the path of a directory forcibly removed recursively that could result in serious data loses. I'm not certain this is possible yet but that's definitely something we should be wary of in the future and for that reason we might need to tag some mutants as risky or update the doc to recommend Infection to be run in a container instead.

sanmai:
> What about the following: we apply that memory limit only if the memory limit is -1 and not forced. 
That's ambiguous. You have a documented knob, and you can't tell if it'll work or now. Say, you added a limit of 128M to `infection.json.dist` but in php.ini you have it set to 8192M. What takes precedence?.. I think if there's a limit configured, it must be in effect without any if's 

Moreover, right now `php -dmemory_limit=512M infection` has no effect on child processes. And I don't think there's a point to limit memory allocations for the Infection itself. None of my two PRs add such limitations. 

We can do better. To remove all ambiguity I see no other way but adding a command line option. And, possibly, a question during initial configuration. How do you feel about this? @theofidry

sanmai:
>a mutant change the path of a directory forcibly removed recursively that could result in serious data loses

> recommend Infection to be run in a container

True, this must be the recommended default.

sanmai:
> Infection is a few MB close, API-Platform is at 150MB just to name a few.

Infection does not need that much memory because it runs only a subset of tests. Therefore I think that we should stick to 128Mb for mutant processes. That, unless there is a compelling evidence. I'd like to know if there is any.

Ocramius:
Tbh, should just pass down execution flags (including memory limit `-dmemory_limit`) to child processes exactly like phpunit does it with the PHPT test runner.

Otherwise, just take the default from `php.ini` and for Belgium's sake do not tamper with it! If a user set a memory limit of `-1` manually, then please leave it be and don't alter things any further.

theofidry:
>>What about the following: we apply that memory limit only if the memory limit is -1 and not forced.

>That's ambiguous. You have a documented knob, and you can't tell if it'll work or now. Say, you added a limit of 128M to infection.json.dist but in php.ini you have it set to 8192M. What takes precedence?.. I think if there's a limit configured, it must be in effect without any if's

I think the order of precedence should be the standard: command argument > app config > language config.

Do you find the scenarios I gave above confusing?

>Moreover, right now php -dmemory_limit=512M infection has no effect on child processes. And I don't think there's a point to limit memory allocations for the Infection itself. None of my two PRs add such limitations.
>We can do better.

Precisely we can do better. Don't make me require to have a dedicated `infection_ci.json`. There is two alternatives:

```
$ php -dmemory_limit=X infection
$ infection --memory-limit=X
```

The later with which you could achieve exactly the same thing as your current PR, i.e. applying the memory limit to the PHPUnit processes only, would be doable. I however thing the extra work compared to use the language feature is not worth it even if that one does apply the limit to Infection itself as well.

>Moreover, right now php -dmemory_limit=512M infection has no effect on child processes

Yes, but as I said, this is an actual bug and should be fixed.

sidz:
my 2 cents: 
- we shouldn't add these kind of options to `infection.json` and/or modify `phpunit.xml` as this is not a web application it's a cli command. What we really can do - copy memory_limit from the original phpunit.xml if this option will be there.
- this should be manipulated through `php -d` or `php.ini` (Agree with @Ocramius )

p.s. @Ocramius glad to see your interest on Infection ;)

theofidry:
Maybe we could include this check and setting this parameter in the
XdebugHandler. This way it would be test framework agnostic.

sanmai:
> Do you find the scenarios I gave above confusing?

That's confusing:

> we apply that memory limit only if the memory limit is -1 and not forced. 

How a user supposed to know if a memory limit will be applied or now? Should they check with the multitude `.ini` files every time to know that?

PR #255, opened 2018-03-23
Enhancement: test in multiple environments

BackEndTea:
This PR:

Adds the option to run tests in different environment with docker, with the following options:

* `test`: run everything
* `test-unit` run phpunit tests
* `test-infection-phpdbg` run infection with phpdbg
* ` test-e2e-phpdbg` run e2e tests with phpdbg
* ` test-infection-xdebug` run infection with xdebug
* `test-e2e-xdebug` run e2e tests with xdebug

For each of these options, except for plain `test` adding `-70`, `-71` or `-72` to it runs it ony for php 7.0, 7.1, or 7.2. So the following:
```bash
$ make test-infection-xdebug-70
```
will run infection with xdebug on php 7.0

It requires devs to be able to run `make ` and `docker`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,125 @@
+PHP_CS_FIXER_FUTURE_MODE=1
+
+.PHONY: all
+#Run all checks, default when running 'make'
+all: analyze test
+
+#Non phony targets for phars etc.
+vendor: composer.json composer.lock
+	composer install
+
+build/cache:
+	mkdir -p build/cache
+
+php-cs-fixer-v2.phar:
+	wget -nc http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar
+	chmod a+x php-cs-fixer-v2.phar
+
+phpstan.phar:
+	wget -nc https://github.com/phpstan/phpstan/releases/download/0.9.1/phpstan.phar
+	chmod a+x phpstan.phar
+
+
+#All tests, (infection itself, phpunit, e2e) for different php version/ environments (xdebug or phpdbg)
+.PHONY: test test-unit test-infection-phpdbg test-e2e-phpdbg test-infection-xdebug test-e2e-xdebug test-final-private
+test: test-unit test-infection-phpdbg test-e2e-phpdbg test-infection-xdebug test-e2e-xdebug test-final-private
+
+test-final-private:
+	./tests/final_private_test
+
+.PHONY: test-unit test-unit-70 test-unit-71 test-unit-72
+#php unit tests
+test-unit: test-unit-70 test-unit-71 test-unit-72
+
+test-unit-70: vendor
+	docker run -it --rm -v "$$PWD":/opt/infection -w /opt/infection php:7.0-cli php vendor/bin/phpunit
+
+test-unit-71: vendor
+	docker run -it --rm -v "$$PWD":/opt/infection -w /opt/infection php:7.1-cli php vendor/bin/phpunit
+
+test-unit-72: vendor
+	docker run -it --rm -v "$$PWD":/opt/infection -w /opt/infection php:7.2-cli php vendor/bin/phpunit
+
+
+.PHONY: test-infection-phpdbg test-infection-phpdbg-70 test-infection-phpdbg-71 test-infection-phpdbg-72
+#infection with phpdbg
+test-infection-phpdbg: test-infection-phpdbg-70 test-infection-phpdbg-71 test-infection-phpdbg-72
+
+test-infection-phpdbg-70: vendor
+	docker run -it --rm -v "$$PWD":/opt/infection -w /opt/infection php:7.0-cli phpdbg -qrr bin/infection --threads=4
+
+test-infection-phpdbg-71: vendor
+	docker run -it --rm -v "$$PWD":/opt/infection -w /opt/infection php:7.1-cli phpdbg -qrr bin/infection --threads=4
+
+test-infection-phpdbg-72: vendor
+	docker run -it --rm -v "$$PWD":/opt/infection -w /opt/infection php:7.2-cli phpdbg -qrr bin/infection --threads=4
+
+
+.PHONY: test-e2e-phpdbg test-e2e-phpdbg-70 test-e2e-phpdbg-71 test-e2e-phpdbg-72
+#e2e tests with phpdbg
+test-e2e-phpdbg: test-e2e-phpdbg-70 test-e2e-phpdbg-71 test-e2e-phpdbg-72
+
+test-e2e-phpdbg-70: vendor
+	docker run -it --rm -e PHPDBG=1 -v "$$PWD":/opt/infection -w /opt/infection php:7.0-cli ./tests/e2e_tests
+
+test-e2e-phpdbg-71: vendor
+	docker run -it --rm -e PHPDBG=1 -v "$$PWD":/opt/infection -w /opt/infection php:7.1-cli ./tests/e2e_tests
+
+test-e2e-phpdbg-72: vendor
+	docker run -it --rm -e PHPDBG=1 -v "$$PWD":/opt/infection -w /opt/infection php:7.2-cli ./tests/e2e_tests
+
+
+.PHONY: test-infection-xdebug test-infection-xdebug-70 test-infection-xdebug-71 test-infection-xdebug-72
+#infection with xdebug
+test-infection-xdebug: test-infection-xdebug-70 test-infection-xdebug-71 test-infection-xdebug-72
+
+test-infection-xdebug-70: build-xdebug-70
+	docker run -it --rm -v "$$PWD":/opt/infection -w /opt/infection infection_php70 php bin/infection --threads=4
+
+test-infection-xdebug-71: build-xdebug-71
+	docker run -it --rm -v "$$PWD":/opt/infection -w /opt/infection infection_php71 php bin/infection --threads=4
+
+test-infection-xdebug-72: build-xdebug-72
+	docker run -it --rm -v "$$PWD":/opt/infection -w /opt/infection infection_php72 php bin/infection --threads=4
+
+
+.PHONY: test-e2e-xdebug test-e2e-xdebug-70 test-e2e-xdebug-71 test-e2e-xdebug-72
+#e2e tests with xdebug
+test-e2e-xdebug: test-e2e-xdebug-70 test-e2e-xdebug-71 test-e2e-xdebug-72
+
+test-e2e-xdebug-70: build-xdebug-70
+	docker run -it --rm -v "$$PWD":/opt/infection -w /opt/infection infection_php70 ./tests/e2e_tests
+
+test-e2e-xdebug-71: build-xdebug-71
+	docker run -it --rm -v "$$PWD":/opt/infection -w /opt/infection infection_php71 ./tests/e2e_tests
+
+test-e2e-xdebug-72: build-xdebug-72
+	docker run -it --rm -v "$$PWD":/opt/infection -w /opt/infection infection_php72 ./tests/e2e_tests
+
+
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -0,0 +1,87 @@
...
+
+test-unit-71: deps
+	docker run -it --rm -v "$$PWD":/opt/infection -w /opt/infection php:7.1-cli php vendor/bin/phpunit
+
+test-unit-72: deps
+	docker run -it --rm -v "$$PWD":/opt/infection -w /opt/infection php:7.2-cli php vendor/bin/phpunit
+
+
+.PHONY:
+#infeciton with phpdbg
```
theofidry: `infection`

Makefile
```diff
@@ -0,0 +1,87 @@
+.PHONY: deps
```
theofidry: why are all those `.PHONY` needed?

sanmai: `.PHONY` needed for make to ignore any files names same as the target. If you happen have a file name `deps`, make will basically ignore `make deps`

Makefile
```diff
@@ -0,0 +1,80 @@
+deps: composer.json
+	composer install
```
sanmai: Did you mean composer.lock?

Then it must be like:
```
deps: composer.lock
composer.lock: composer.json
    composer update
```

Makefile
theofidry: Right but we don't have a `deps` directory do we?

sanmai: Right now we don't, but if would have it by any chance in future, you'll have a good time scratching your head. @BackEndTea did it all right with `.PHONY`

Makefile
```diff
@@ -0,0 +1,80 @@
+deps: composer.json
```
sanmai: Did you mean composer.lock?

Then it must be like:
```
deps: composer.lock
composer.lock: composer.json
    composer update
```
Else it will run `composer install` on every build.

Makefile
```diff
@@ -1,9 +1,11 @@
-deps: composer.json
+.PHONY: deps
+deps: composer.json composer.lock
```
sanmai: This way it will run `composer install` on every build.

If it'd be like this, it would only run `composer update` if `composer.json` is newer than `composer.lock`:
```
deps: composer.lock
composer.lock: composer.json
    composer update
```

Makefile
```diff
@@ -0,0 +1,87 @@
...
```
theofidry: why not depending on `vendor` directory and have a rule doing the `composer install`? Or is there a reason why you want to trigger a `composer install` multiple times?

Makefile
sanmai: That's how they do it in the make world. Can't help this.

Makefile
BackEndTea: I need to change that,as well as making sure either composer can be ran inside of the containers, or run composer install in the e2e tests beforehand

sanmai: If you just want to run `composer install` every time, then you don't need to depend on anything. This should work just fine:
```
deps:
    composer install
```
I believe compose can quickly infer if there's a work to be done. So that's a non-issue.

Makefile
theofidry: It's not a requirement here so it's about what we prefer.

Makefile
theofidry: Yes, so my question was about whether this was intended as from what I see only 1 `composer install` call is enough. Sure the overhead is small but it's not like removing that overhead is costing a lot of efforts here either

Makefile
BackEndTea: I meant for the e2e tests not the project itself

localheinz: Also see https://www.reddit.com/r/PHP/comments/7sltmh/makefile_for_lazy_developers/dt5rba0/.

sanmai:
Please let me stick for @BackEndTea here. Make is an industry standard. But what is more important is that every Makefile comes with a full dependency resolver. You just set the rules, and make decides which command needs to be run first. Comparing Composer commands with a Makefile is like comparing a simple screwdriver with a real-time bi-directional digital torque screwdriver. 

The other questions is we actually need a Makefile. Personally I would not wince because I know how make works, used it for many years, and so on. That makes me biased, so let me stop here.

theofidry:
@borNfreee I can feel you if you're not familiar with it, on another hand for the usage we are making I think it's gonna be relatively easy to grasp.

One thing I really like as well with Make as opposed to Composer scripts is that it's really easier to achieve the "things work feeling" e.g. clone the project &`make test` boom everything is taken care of. It's also easier to introduce a [help command](https://github.com/humbug/box/blob/master/Makefile#L4-L5) and you have the auto-completion so you benefit from a better discoverability

borNfreee:
I'm not against it, I wanted to know why @BackEndTea decided to use it while we have composer scripts and what does bring us. But you actually answered to this question.

If we now use Makefile, why do we need composer scripts?
Should we move `cs:check`, `analyze` and other stuff to Makefile?

theofidry:
>Should we move cs:check, analyze and other stuff to Makefile?

This is happening 😄

sanmai:
> Should we move cs:check, analyze and other stuff to Makefile?

Can second this more.

Makefile
localheinz: Also see https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html:

> There are two reasons to use a phony target: to avoid a conflict with a file of the same name, and to improve performance.

theofidry: @localheinz it's up to you, personally I would hate to always have to add a target to the .PHONY (and perf. is very unlikely to ever be an issue it's a micro-optim to say the least here) but whatever ways you guy swing I don't feel strongly about it

Makefile
```diff
@@ -0,0 +1,120 @@
...
+	docker build -t infection_php72 -f "$$PWD/devTools/Dockerfile-php72-xdebug" .
+
+#style checks/ static analysis
+.PHONY: analyze cs-fix cs-check phpstan validate update
+analyze: cs-check phpstan validate
+
+cs-fix: build/cache php-cs-fixer-v2.phar
+	./php-cs-fixer-v2.phar fix -v --cache-file=build/cache/.php_cs.cache
+
+cs-check: build/cache php-cs-fixer-v2.phar
```
borNfreee: please update this file with the new commands 

https://github.com/infection/infection/blob/master/.github/CONTRIBUTING.md#how-to-contribute

sanmai:
Just saying that we can have old aliases if there's a need:

```diff
-cs-fix: build/cache php-cs-fixer-v2.phar
+cs\:fix: build/cache php-cs-fixer-v2.phar
```

Makefile
```diff
@@ -0,0 +1,120 @@
+.PHONY: all
+#Run all checks, default when running 'make'
+all: analyze test
+
+#Non phony targets for phars etc.
+vendor: composer.json composer.lock
+	composer install
```
sanmai: Wouldn't it be nice if it would run `composer update` for us too, would it?

```make
# Do install if there's no 'vendor'
vendor/autoload.php:
    composer install

# If composer.lock is older than `composer.json`, do update,
# and touch composer.lock because composer not always does that
composer.lock: composer.json
    composer update && touch composer.lock

# We need both vendor/autoload.php and composer.lock being up to date
.PHONY: vendor
vendor: vendor/autoload.php composer.lock
```

Makefile
```diff
@@ -0,0 +1,120 @@
...
+
+build/cache:
+	mkdir -p build/cache
+
+php-cs-fixer-v2.phar:
+	wget -nc http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar
```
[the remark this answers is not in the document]
sanmai: And these has to depend on `vendor`: else make `make phpstan` will fail:

    Class ... was not found while trying to analyse it - autoloading is probably not configured properly.

sanmai:
Overall there's more work to be done, but that can wait.

E.g. it's a common convention to use variable placeholders for commands:

```make
PHPSTAN=./phpstan.phar
phpstan: $(PHPSTAN)
    $(PHPSTAN) analyse src tests --level=2 -c phpstan.neon --no-interaction --no-progress
```

Then, if you ever need to use a different version of a tool, you would't have to overwrite any files and risk forgetting to put them back. You could just replace any command for one run:
```bash
make phpstan PHPSTAN=./phpstan-2.phar
```
Or even disable it:

```bash
make phpstan PHPSTAN=/bin/true
```

[For more or less complete example see here.](https://github.com/sanmai/pipeline/blob/ca3f78dc614ecb0aaa29a7008cb83a9d194a5bcc/Makefile)

Makefile
```diff
@@ -9,13 +12,13 @@ vendor: composer.json composer.lock
 build/cache:
 	mkdir -p build/cache
 
-php-cs-fixer-v2.phar:
-	wget -nc http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar
-	chmod a+x php-cs-fixer-v2.phar
+$(PHP-CS-FIXER):
```
sanmai: That's the target for `php-cs-fixer-v2.phar`, you're dowloading it below all the same. No need to make it a variable. And more, it'll fail every time:
```
make PHP-CS-FIXER=php-cs-fixer-v3.phar php-cs-fixer-v3.phar 
# wget still trying to download php-cs-fixer-v2.phar
# make not seeing php-cs-fixer-v3.phar, fails
```

Makefile
```diff
@@ -9,13 +12,13 @@ vendor: composer.json composer.lock
...
+	wget http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar
+	chmod a+x $(PHP-CS-FIXER)
 
-phpstan.phar:
-	wget -nc https://github.com/phpstan/phpstan/releases/download/0.9.1/phpstan.phar
-	chmod a+x phpstan.phar
+$(PHPSTAN):
+	wget https://github.com/phpstan/phpstan/releases/download/0.9.1/phpstan.phar
+	chmod a+x $(PHPSTAN)
```
sanmai: Same here. That's the targets you define to download the default tools. If someone wants to use other version of a tool, it is their responsibility to download it.

Makefile
```diff
@@ -107,14 +110,14 @@ build-xdebug-72: vendor
...
 analyze: cs-check phpstan validate
 
-cs-fix: build/cache php-cs-fixer-v2.phar
-	./php-cs-fixer-v2.phar fix -v --cache-file=build/cache/.php_cs.cache
+cs-fix: build/cache $(PHP-CS-FIXER)
+	$(PHP-CS-FIXER) fix -v --cache-file=build/cache/.php_cs.cache
 
-cs-check: build/cache php-cs-fixer-v2.phar
-	./php-cs-fixer-v2.phar fix -v --cache-file=build/cache/.php_cs.cache --dry-run --stop-on-violation
+cs-check: build/cache $(PHP-CS-FIXER)
```
sanmai: You can't depend on a variable if you can't *make* it. There's a rule for `php-cs-fixer-v2.phar`, make knows how to download it. But it can't possibly know how to download `php-cs-fixer-beta-v5.phar`. Build will fail.

Fenikkusu:
A simple observation on my part. When I first pulled infection down to work on it, I wasn't aware of the composer scripts. Once it was brought to my attention, I ran it but learned I didn't have wget. A simple 'brew install wget' took care of this for me (OSX). I'm wondering if this can be handled by make as well.

sanmai:
And then there's no homebrew, so... Do we have to install it too? Indeed,

> that would be way outside the scope for this project

borNfreee:
Let's merge this one and improve everything step by step.

Thank you so much @BackEndTea - great work!

Thanks @sanmai, @localheinz, @theofidry for forcing makefile 🤣

PR #257, opened 2018-03-24
Use loaded ini settings

theofidry:
The original issue was reported by @sanmai in #252:

>right now `php -dmemory_limit=512M` infection has no effect on child processes

The issue is actually not specific to this peculiar parameter but any parameter overriden with the
`-d` option.

Indeed when building the temporary ini file, we are only looking at the settings coming from the
scanned files. Those settings however can be changed with the `-d` option. So provided you have:

```
// php.ini
param=X

// console
$ php -dparam=Y infection
```

We _should_ have in the temporary ini file `param=Y`.

<hr/>

Note that this issue is fixed by https://github.com/composer/xdebug-handler so we won't have this issue again when moving to it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
 
 env:
   global:
-    - INFECTION_FLAGS='--threads=4 --min-msi=57 --min-covered-msi=86 --coverage=coverage'
+    - INFECTION_FLAGS='--threads=4 --min-msi=56 --min-covered-msi=86 --coverage=coverage'
     - PHPUNIT_BIN='vendor/bin/phpunit --coverage-clover=clover.xml --coverage-xml=coverage/coverage-xml --log-junit=coverage/phpunit.junit.xml'
 
 cache:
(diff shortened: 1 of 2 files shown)


sanmai:
This PR is all good, but this isn't a proper fix for the original issue #247. (PR #252 was superseded by the proposal #253.)

"Solving" #247 with this PR is like throwing out the champagne along with the cork. For one reason there's no evidence the Infection ever needed any kind of memory limits. Infection is a fine well written program, there's no need to limit its memory use so drastically.

theofidry:
Yes it's not a patch for #247, I'm just saying this bug has been reported there nothing more :)

sidz:
@theofidry I want to integrate https://github.com/composer/xdebug-handler instead of our custom handler. Maybe today or tomorrow :)

borNfreee:
Since we have intergrated `xdebug-handler`, we don't have `src/Php/ConfigBuilder.php` anymore.

Is this PR still needed @theofidry ?

theofidry:
Yes that should be fixed now!

PR #258, opened 2018-03-24
Heuristics to set memory limit for mutants running with PHPUnit

### LINKED ISSUE #247: Mutant processes are not memory-bound

sanmai:
Since we know how much memory the initial test suite used, and only if we know, we can enforce a memory limit upon all mutation processes. Limit is set to be twice the known amount, because if we know that a normal test suite used X megabytes, if a mutant uses a lot more, this is a definite error.
    
Memory limit is introduced by altering a known temporary php.ini to include a directive to enable the limit as the very last line. We only apply a memory limit if there isn't one set.

This PR assumes no default memory limits.

This PR:

- [x] Adds heuristics to set memory limit for mutants running with PHPUnit ...
- [x] Covered by tests
- [x] Fixes #247 as far I concerned

I rather hope this PR will be more satisfactory to @theofidry and others who didn't like an idea of a default memory limit. We are not forcing a default upon anyone here.

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -30,6 +30,7 @@ use Infection\Process\Runner\InitialTestsRunner;
 use Infection\Process\Runner\MutationTestingRunner;
 use Infection\TestFramework\AbstractTestFrameworkAdapter;
 use Infection\TestFramework\Coverage\CodeCoverageData;
+use Infection\TestFramework\MemoryUsageAware;
 use Infection\TestFramework\PhpSpec\PhpSpecExtraOptions;
 use Infection\TestFramework\PhpUnit\Adapter\PhpUnitAdapter;
 use Infection\TestFramework\PhpUnit\Coverage\CoverageXmlParser;
@@ -194,7 +195,7 @@ class InfectionCommand extends BaseCommand
         }
 
         // We only apply a memory limit if there isn't one set
-        if ($adapter instanceof PhpUnitAdapter && ini_get('memory_limit') === '-1') {
+        if ($adapter instanceof MemoryUsageAware && ini_get('memory_limit') === '-1') {
             $this->applyMemoryLimitFromPhpUnitProcess($initialTestSuitProcess, $adapter);
         }
 
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -31,6 +31,7 @@ use Infection\Process\Runner\MutationTestingRunner;
 use Infection\TestFramework\AbstractTestFrameworkAdapter;
 use Infection\TestFramework\Coverage\CodeCoverageData;
 use Infection\TestFramework\PhpSpec\PhpSpecExtraOptions;
+use Infection\TestFramework\PhpUnit\Adapter\PhpUnitAdapter;
 use Infection\TestFramework\PhpUnit\Coverage\CoverageXmlParser;
 use Infection\TestFramework\PhpUnit\PhpUnitExtraOptions;
 use Infection\TestFramework\TestFrameworkExtraOptions;
@@ -192,6 +193,11 @@ class InfectionCommand extends BaseCommand
             return 1;
         }
 
+        // We only apply a memory limit if there isn't one set
+        if ($adapter instanceof PhpUnitAdapter && ini_get('memory_limit') === '-1') {
+            $this->applyMemoryLimitFromPhpUnitProcess($initialTestSuitProcess, $adapter);
+        }
+
         $codeCoverageData = $this->getCodeCoverageData($testFrameworkKey);
         $mutationsGenerator = new MutationsGenerator(
             $container->get('src.dirs'),
@@ -251,6 +257,37 @@ class InfectionCommand extends BaseCommand
         }
 
         throw new \InvalidArgumentException('Incorrect formatter. Possible values: "dot", "progress"');
+    }
+
+    private function applyMemoryLimitFromPhpUnitProcess(Process $process, PhpUnitAdapter $adapter)
+    {
+        $tempConfigPath = \php_ini_loaded_file();
+
+        if (empty($tempConfigPath) || !file_exists($tempConfigPath)) {
+            // Cannot add a memory limit when there's no php.ini
+            return;
+        }
+
+        $memoryLimit = $adapter->getMemoryUsed($process->getOutput());
+
+        if ($memoryLimit < 0) {
+            // Cannot detect memory used, not setting any limits
+            return;
+        }
+
+        /*
+         * Since we know how much memory the initial test suite used,
+         * and only if we know, we can enforce a memory limit upon all
+         * mutation processes. Limit is set to be twice the known amount,
+         * because if we know that a normal test suite used X megabytes,
+         * if a mutants uses a lot more, this is a definite error.
+         *
+         * By default we let a mutant process use twice as much more
+         * memory as an initial test suite consumed.
+         */
+        $memoryLimit *= 2;
+
+        file_put_contents($tempConfigPath, PHP_EOL . sprintf('memory_limit = %dM', $memoryLimit), FILE_APPEND);
     }
 
     private function registerSubscribers(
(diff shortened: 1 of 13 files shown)


src/Command/InfectionCommand.php
```diff
@@ -192,6 +194,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 1;
         }
 
+        // We only apply a memory limit if there isn't one set
+        if ($adapter instanceof PhpUnitAdapter && ini_get('memory_limit') === '-1') {
```
borNfreee: What about other test frameworks, e.g. PhpSpec? shouldn't we care about it as well?

I'm sure if we even end up with PhpUnitAdapter for now, we should introduce an interface like `MemoryUsageAware` (or whatever you name it) and check it rather than hardcoding PHPUnit related stuff in the very base command

```php
if ($adapter instanceof MemoryUsageAware && ini_get('memory_limit') === '-1') {
    $this->applyMemoryLimitFromTestFrameworkProcess();
}


class PhpUnitAdapter implement MemoryUsageAware...
```

sanmai:
If we could ignore other frameworks, then, yes, this is a bare-bones replacement for #253. 

PhpSpec isn’t reporting memory usage like PHPUnit does (I checked), so if we need to support it we have no choice by go by #253.

sanmai:
I believe PHP interpreter will not indicate by an exit code if it died because of a language error or not. Therefore, we could only reason about exit statuses if we capture actual error. Right now PHPUnit will not print any errors, but if it could be made to, it's a possibility.

sanmai:
I’m not sure if we can hack PhpSpec runner to output a number from *memory_get_peak_usage()*, and quick googling did not show anything. 

Did I address all you concerns, @borNfreee?

theofidry:
👍 I must say I much prefer this one to the others.

>I mean add a new letter to .: killed, M: escaped, S: uncovered, E: fatal error, T: timed out, for example, X: memory limit.

👍 👍 

IIRC you get an error along the lines of `Fatal error: Allowed memory size of` or a special code for this kind of error so we could know if the failure was caused by a memory exhaustion that way.

I think we should also make it configurable like we do for the time out; This can wait though.

>If we could ignore other frameworks

We could try to add a memory consumption on their end but I don't think it's a blocker for this PR.

sidz:
Is this approach really working as you expect? It will set `memory_limit` only for the main phpunit process :)
But PHPUnit runs sub-processes for each test case. And this approach won't work for them at all.

As Marco said yesterday we need to do it in the same way like phpunit does. So I investigated a bit yesterday and here is how phpunit works:

https://github.com/sebastianbergmann/phpunit/blob/7.0.2/src/Runner/PhptTestCase.php#L125 - here it takes ini section from the xml and puts all ini's to the main `$settings` variable

then `getCommand` https://github.com/sebastianbergmann/phpunit/blob/7.0.2/src/Util/PHP/DefaultPhpProcess.php#L95 method will generate `/usr/bin/php -d bla_bla=-1 file` for each test case.

So this PR will tweak only the main phpunit process.

src/Command/InfectionCommand.php
sanmai: - [x] MemoryUsageAware

sanmai:
@sidz, PHPUnit will pass trough all options we set in our custom php.ini even if we tell it to use process isolation:

    bin/infection --test-framework-options="--process-isolation"

If someone decides to shoot oneself in the foot by setting own limits with `<ini name="memory_limit" value="-1"/>`, they gonna be hit by it. That's par for the course.

sanmai:
@sidz we use `ConfigBuilder::ENV_TEMP_PHP_CONFIG_PATH` in `PhpExecutableFinder` also, and I don't see any replacement to it in the composer/xdebug-handler. It doesn't even export [a similar path it has](https://github.com/composer/xdebug-handler/blob/master/src/XdebugHandler.php#L35). And `PhpExecutableFinder` needs to know which `php.ini` to use. I don't see how you gonna do this.

sidz:
@sanmai https://github.com/infection/infection/pull/215#issuecomment-375773920

sanmai:
@sidz That's excellent! We would only have to change [this line](https://github.com/infection/infection/pull/258/commits/33ee1df0f71c2e8c612eeac93eb543269b1fa719#diff-a99db3f318d0547a301f216cb03b51a7R287) and [this line](https://github.com/infection/infection/blob/79c873663e088a3cbaf3a7a1cd7600f25845fb90/src/Process/ExecutableFinder/PhpExecutableFinder.php#L21) to use  `php_ini_loaded_file()` instead. It would be best if there will be an utility function. In any case this is trivial task and not a blocker to this PR.

sidz:
@sanmai you can change this line https://github.com/infection/infection/pull/258/commits/33ee1df0f71c2e8c612eeac93eb543269b1fa719#diff-a99db3f318d0547a301f216cb03b51a7R287 without any changes right now :) So please do it. thx

sanmai:
@sidz Done! 

I'm happy to say that my E2E test for this issue is passing with `php_ini_loaded_file()`.

tests/Fixtures/e2e/Memory_Limit/README.md
```diff
@@ -0,0 +1,9 @@
+# Title
```
borNfreee: please either remove this file or add correct information

src/Command/InfectionCommand.php
```diff
@@ -253,6 +260,38 @@ private function getOutputFormatter(): OutputFormatter
         throw new \InvalidArgumentException('Incorrect formatter. Possible values: "dot", "progress"');
     }
 
+    /**
+     * By default we let a mutant process use as much more memory as an initial test suite consumed.
+     *
+     * @var int
+     */
+    const DEFAULT_MEMORY_LIMIT_MULTIPLIER = 2;
```
sidz: imo all constants should be placed at the top of the class

src/Command/InfectionCommand.php
```diff
@@ -253,6 +260,38 @@ private function getOutputFormatter(): OutputFormatter
...
+        }
+
+        /*
+         * Since we know how much memory the initial test suite used,
+         * and only if we know, we can enforce a memory limit upon all
+         * mutation processes. Limit is set to be twice the known amount,
+         * because if we know that a normal test suite used X megabytes,
+         * if a mutants uses a lot more, this is a definite error.
+         */
+        $memoryLimit *= self::DEFAULT_MEMORY_LIMIT_MULTIPLIER;
```
sidz: could you please move this line under condition as we don't use it elsewhere?

borNfreee:
> What if I want or I need to set -1 for all sub-processes?

what is the use case when you want PHPUnit to consume unlimited (and much more than initial test run consumed) memory?

sidz:
can't give you any constructive example now but I have a stronger feeling that we're doing smth wrong.

borNfreee:
TBH I don't get it. This PR limits PHPUnit processes only with the twice bigger amount of memory than initial test suite consumes.

I'm just summing up it to myself as well to find any potential issues but nothing comes to my mind.

Anyway, we still have a plenty of time before 0.9.0, so we can see how it's going

src/Command/InfectionCommand.php
sanmai: 1. This isn't going to make the code more readable. 
2. PHP CS Fixer normally takes care of this if there's such policy, and it didn't.

When there's no policy, I side with *Code Complete* on minimizing live time for programming constructs.

src/Command/InfectionCommand.php
sanmai: Done.

src/Command/InfectionCommand.php
sidz: if you find any place in the infection code where we did it in the same way - feel free to leave it as is. if no - please move it to the top of the class. thank you

sanmai: However you wish. Done.

borNfreee:
Thanks for all your hard work on this topic @sanmai, much appreciated

Issue #251, opened 2018-03-22
Try to simplify Public API of Mutator class


borNfreee:
This is how `Mutator`'s public API [looks like](https://github.com/infection/infection/blob/master/src/Mutator/Util/Mutator.php) at the moment:

```php
abstract class Mutator
{
    abstract public function mutate(Node $node);

    abstract public function shouldMutate(Node $node): bool;

    public static function getName(): string {}

    public function isIgnored(string $class, string $method): bool { }
}
```

Note the `isIgnored` method we've added recently.

If you look at the [MutationsCollectorVisitor::leaveNode](https://github.com/infection/infection/blob/master/src/Visitor/MutationsCollectorVisitor.php#L63) method logic, you will see that `Mutator` *should not* mutate the code if *Mutator is ignored*, right?

What I want to say is `->isIgnored()` method should be an implementation detail of `->shouldMutate()` method. So I suggest the following change:

```php
public function shouldMutate(Node $node)
{
    if (!$this->someAbstractShouldMutate()) {
        return false;
    }

    return !$this->isIgnored($node);
}

private isIgnored() {} // it's private now (!)
```

#### Why does it bother me?

* Because it looks logical that we should not mutate when mutator is ignored

* One day, we will expose `Mutator` interface to the external world in order to allow users to implements their own mutators in the project that Infection will pick up, so we will promise some level of stability of this interface and every public method will be a pain for us.

theofidry:
👍

BackEndTea:
On that note, we should also make the `getName()` function final.

sanmai:
In regard of users from the outside world I think we better leave `shouldMutate()` as it is. Whatever reasons there are Infection thinks a mutator should be ignored is its own internal business. Therefore, `isIgnored` should stay separate from `shouldMutate()`. 

Similar is also true about `isIgnored()`. If an outside user want to ignore certain mutators for whatever principle they follow, it's their own business. We don't get to limit them anyhow, we can't hide `isIgnored()` because a user should be able to override it.

It doesn't seem like a problem right now, but just so you would know, [for performance reasons](https://github.com/infection/infection/pull/235#discussion_r176333456) it is better when `shouldMutate()` comes first.

borNfreee:
Disagree. Currently, we have not the best design of our classes (Mutator at lest) that brings us problems I mentioned in the description.

`isIgnored()` should not be public inside `Mutator` class, it's already public inside `MutatorConfig`.

[...]

PR #259, opened 2018-03-24
Remove `isIgnored` from the public API of Mutator class. Make methods final

borNfreee:
This PR:

- [x] Fixes https://github.com/infection/infection/issues/251
- [x] Covered by tests

The main idea is to reduce public API of `Mutator` class and improve future maintainability. Read more in #251

### DIFF
--- a/src/Mutator/Util/Mutator.php
+++ b/src/Mutator/Util/Mutator.php
@@ -47,6 +47,6 @@ abstract class Mutator
     {
         $parts = explode('\\', static::class);
 
-        return $parts[count($parts) - 1];
+        return end($parts);
     }
 }
--- a/src/Mutator/Util/Mutator.php
+++ b/src/Mutator/Util/Mutator.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 
 namespace Infection\Mutator\Util;
 
+use Infection\Visitor\ReflectionVisitor;
 use PhpParser\Node;
 
 abstract class Mutator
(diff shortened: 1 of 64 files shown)


src/Mutator/Util/Mutator.php
```diff
@@ -25,17 +26,27 @@ public function __construct(MutatorConfig $config)
...
+        return !$this->config->isIgnored($reflectionClass->getName(), $node->getAttribute(ReflectionVisitor::FUNCTION_NAME, ''));
     }
 
-    public function isIgnored(string $class, string $method): bool
+    final public static function getName(): string
     {
-        return $this->config->isIgnored($class, $method);
+        $parts = explode('\\', static::class);
+
+        return $parts[count($parts) - 1];
```
sanmai: While we are at it, this looks very much like 
```php
return end($parts);
```
Or even
```php
return substr(static::class, strrpos(static::class, '\\') + 1);
```

sanmai:
@BackEndTea  `shouldMutate` includes `mutatesNode` which [is the fastest method here](https://github.com/infection/infection/pull/235#discussion_r176333456); no order change needed.

But surely this can be confusing.

BackEndTea:
In that case i have no objections

borNfreee:
@sanmai @BackEndTea rebased and addressed comment

BackEndTea:
👍 for me

borNfreee:
Thank you @sanmai & @BackEndTea

PR #260, opened 2018-03-24
Do not generate file logs on CI for Infection itself

borNfreee:
We don't need file logs in Travis. Mutation scores for badge still will be sent.


borNfreee:
Thank you @sidz

PR #261, opened 2018-03-24
Add mutator that remove finally {} block

borNfreee:
This PR:

Adds new Mutator that removes `finally {}` block

- [x] Covered by tests
- [ ] Doc PR: TBD

Found one weird bug during testing this mutator on Infection itself

bug-finally

As you can see, when `try {}` or `catch {}` section has `return`, `break`, `continue`, then `finally` is not marked as covered. It can be either not covered (red line in HTML coverage report), or yellow (dead code) or white (N/A).

But it's not true as `finally` block is always executed: https://3v4l.org/ZX0Zh.

### DIFF
--- a/src/Mutator/Operator/Finally_.php
+++ b/src/Mutator/Operator/Finally_.php
@@ -1,0 +1,26 @@
+<?php
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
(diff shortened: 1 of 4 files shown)


borNfreee:
Thank you @localheinz, @sidz and @sanmai

djmattyg007:
Does this work if there is no catch block?

borNfreee:
@djmattyg007 omg, I didn't know it's even possible https://3v4l.org/Hse6E

PR #264, opened 2018-03-26
Update LICENSE

sanmai:
- [x] Updated years
- [x] GitHub now should recognize it as a valid *BSD 3-Clause License*


borNfreee:
Thank you @sanmai

Issue #244, opened 2018-03-21
Warn users about change to log-verbosity, change it for 0.10.0


BackEndTea:
Currently our highest verbosity is 1, and the lowest verbosity is 3. Which is backwards to how everyone else is doing it.

I suggest to change it to the following:
* 0 -> No logging
* 1 -> Hide killed mutations (default)
* 2 -> Show all mutations

For `0.9.0` Add a notice if the user is using log-verbosity, that tells the end user that it will be changed in the next update.

Extracted from: https://github.com/infection/infection/pull/221#discussion_r174892392

sanmai:
How about if we keep this option unchanged for backward compatibility in terms of numeric arguments, but instead introduce literal arguments. Say:

* `none` for no logging
* `default` to hide killed mutations
* `all` to show everything

This way no warnings will be needed. And we also gain ability to add more logging levels in future.

Handling for numeric arguments will be left in code, but a note about them will be removed from the docs etc.

sidz:
@BackEndTea The main idea of that verbosity order was that we could have more levels (and we've already added a new one) and with your approach you will break compatibility each time when you'll add a new level. For example you'll need to add a new level between 1 and 2.

I totally agree with @sanmai that the best approach in this case would be: use names instead of numbers.

PR #265, opened 2018-03-27
Use textual version of log verbosity

BackEndTea:
This PR:

- [x] Adds the textual versions of `log-verbosity`
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/34
 
 * Use `none`, `default`, and `all` instead of numbers
 * Doesn't drop support for using numbers
 * Defaults to not showing killed mutators instead of showing all in the log
 * Notifies the user if they use an unknown option

Fixes: #244 

As discussed in #244 this will 'deprecate' the numeric warnings, but it won't drop support for them. Documentation should no longer mention it, and once we do decide to drop support the version before it should explicitly warn end-users that in the next major it will de dropped. (`1.0.0` maybe?)


src/Command/InfectionCommand.php
```diff
@@ -451,4 +452,17 @@ protected function initialize(InputInterface $input, OutputInterface $output)
         $this->eventDispatcher = $this->getContainer()->get('dispatcher');
         $this->skipCoverage = \strlen(trim($input->getOption('coverage'))) > 0;
     }
+
+    /**
+     * @param int|string $loggingLevel
+     */
+    private function checkLoggingLevels($loggingLevel)
```
sanmai: That's what LogVerbosity should do especially if you want a test for this.

src/Process/Listener/MutationTestingResultsLoggerSubscriber.php
```diff
@@ -92,7 +92,7 @@ public function onMutationTestingFinished(MutationTestingFinished $event)
     private function filterLogTypes(array $logTypes): array
     {
         foreach ($logTypes as $key => $value) {
-            if ($this->logVerbosity === LogVerbosity::NONE) {
+            if ($this->logVerbosity === LogVerbosity::NONE || (int) $this->logVerbosity === LogVerbosity::NONE_INTEGER) {
```
sanmai: Would it be better if we convert a numeric verbosity to a string in the beginning, and avoid using deprecated constants in other code?

tests/Console/LogVerbosityTest.php
```diff
@@ -0,0 +1,100 @@
...
+
+        LogVerbosity::convertVerbosityLevel($input, $io);
+    }
+
+    /**
+     * @param string|int $inputVerbosity
+     *
+     * @return InputInterface|Mockery\MockInterface
+     */
+    private function set_input_expectations_when_it_does_not_change($inputVerbosity)
```
borNfreee: AFAIK, we use kebab_case only for test_* methods. 

* `protected function tearDown()`
* `private function setEnv(string $path)`
* `private function getPlusNode`

borNfreee:
> once we do decide to drop support the version before it should explicitly warn end-users that in the next major it will de dropped. (1.0.0 maybe?)

is it possible to do it right now (show deprecation warning)? any objectives?

BackEndTea:
> is it possible to do it right now (show deprecation warning)? any objectives?

Should not be a problem to add a notice

src/Console/LogVerbosity.php
```diff
@@ -9,9 +9,51 @@
...
+    public static function convertVerbosityLevel(InputInterface $input, SymfonyStyle $io)
+    {
+        $verbosityLevel = $input->getOption('log-verbosity');
+        if (in_array($verbosityLevel, self::ALLOWED_OPTIONS)) {
+            return;
+        }
+
+        if (array_key_exists((int) $verbosityLevel, self::ALLOWED_OPTIONS)) {
+            $input->setOption('log-verbosity', self::ALLOWED_OPTIONS[$verbosityLevel]);
+            $io->note('Numeric versions of log-verbosity have been deprecated, please use, ' . self::ALLOWED_OPTIONS[$verbosityLevel] . ' to keep the same result');
```
BackEndTea: We notify the end user that the numeric versions have been deprecated, but we still run the process likenormal

src/Console/LogVerbosity.php
```diff
@@ -9,9 +9,51 @@
...
+
+            return;
+        }
+
+        $io->note('Running infection with an unknown log-verbosity option, falling back to \'default\' option');
```
BackEndTea: Since the logging of the results does not change the output of msi/ what mutators are used etc we allow the process to continue, but we do give the end users a notice if they use an unknown value.

borNfreee:
travis failure is unrelated

Thank you @BackEndTea

BackEndTea:

PR #267, opened 2018-03-28
Error when composer fails to run

sanmai:
If a stock `composer` fails because of a PHP error (broken plugins...), a fallback wasn't used.


sanmai:
This does not seem to be testable unless `TestFrameworkFinder` is refactored to include `ComposerExecutableFinder` as an injected dependency.

borNfreee:
Thank you @sanmai

PR #269, opened 2018-03-28
Updating Dist Config File Locator

Fenikkusu:
This PR:

- [x] Adds new feature ...
- [x] Covered by tests

This PR is to compliment PR #143. Codeception doesn't use the .yml.dist format and instead uses .dist.yml. This PR updates the functionality such that the system will also look for a .dist.yml.

### DIFF
--- a/src/TestFramework/Config/TestFrameworkConfigLocator.php
+++ b/src/TestFramework/Config/TestFrameworkConfigLocator.php
@@ -13,6 +13,15 @@ use Infection\Finder\Exception\LocatorException;
 
 class TestFrameworkConfigLocator
 {
+    const DEFAULT_EXTENSIONS = [
+        'xml',
+        'yml',
+        'xml.dist',
+        'yml.dist',
+        'dist.xml',
+        'dist.yml',
+    ];
+
     /**
      * @var string
      */
@@ -28,7 +37,7 @@ class TestFrameworkConfigLocator
         $dir = $customDir ?: $this->configDir;
         $triedFiles = [];
 
-        foreach ($this->getDefaultExtensions() as $extension) {
+        foreach (static::DEFAULT_EXTENSIONS as $extension) {
             $conf = sprintf('%s/%s.%s', $dir, $testFrameworkName, $extension);
 
             if (file_exists($conf)) {
@@ -42,17 +51,5 @@ class TestFrameworkConfigLocator
             $dir,
             $triedFiles
         );
-    }
-
-    private function getDefaultExtensions(): array
-    {
-        return [
-            'xml',
-            'yml',
-            'xml.dist',
-            'yml.dist',
-            'dist.xml',
-            'dist.yml',
-        ];
     }
 }
--- a/src/TestFramework/Config/TestFrameworkConfigLocator.php
+++ b/src/TestFramework/Config/TestFrameworkConfigLocator.php
@@ -44,7 +44,7 @@ class TestFrameworkConfigLocator
         );
     }
 
-    private function getDefaultExtensions() : array
+    private function getDefaultExtensions(): array
     {
         return [
             'xml',
@@ -52,7 +52,7 @@ class TestFrameworkConfigLocator
             'xml.dist',
             'yml.dist',
             'dist.xml',
-            'dist.yml'
+            'dist.yml',
         ];
     }
 }
--- a/src/TestFramework/Config/TestFrameworkConfigLocator.php
+++ b/src/TestFramework/Config/TestFrameworkConfigLocator.php
@@ -9,6 +9,8 @@ declare(strict_types=1);
 
 namespace Infection\TestFramework\Config;
 
+use Infection\Finder\Exception\LocatorException;
+
 class TestFrameworkConfigLocator
 {
     /**
@@ -24,20 +26,33 @@ class TestFrameworkConfigLocator
     public function locate(string $testFrameworkName, string $customDir = null): string
     {
         $dir = $customDir ?: $this->configDir;
+        $triedFiles = [];
 
-        foreach (['xml', 'yml'] as $extension) {
+        foreach ($this->getDefaultExtensions() as $extension) {
             $conf = sprintf('%s/%s.%s', $dir, $testFrameworkName, $extension);
-            $altConf = sprintf('%s/%s.dist.%s', $dir, $testFrameworkName, $extension);
 
             if (file_exists($conf)) {
                 return realpath($conf);
-            } elseif (file_exists($conf . '.dist')) {
-                return realpath($conf . '.dist');
-            } elseif (file_exists($altConf)) {
-                return realpath($altConf);
             }
+
+            $triedFiles[] = sprintf('%s.%s', $testFrameworkName, $extension);
         }
 
-        throw new \RuntimeException(sprintf('Unable to locate %s(.dist).(xml|yml)(.dist) file.', $testFrameworkName));
+        throw LocatorException::multipleFilesDoNoExist(
+            $dir,
+            $triedFiles
+        );
+    }
+
+    private function getDefaultExtensions() : array
+    {
+        return [
+            'xml',
+            'yml',
+            'xml.dist',
+            'yml.dist',
+            'dist.xml',
+            'dist.yml'
+        ];
     }
 }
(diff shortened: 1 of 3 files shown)


src/TestFramework/Config/TestFrameworkConfigLocator.php
```diff
@@ -27,16 +27,17 @@ public function locate(string $testFrameworkName, string $customDir = null): str
...
         foreach (['xml', 'yml'] as $extension) {
             $conf = sprintf('%s/%s.%s', $dir, $testFrameworkName, $extension);
+            $altConf = sprintf('%s/%s.dist.%s', $dir, $testFrameworkName, $extension);
 
             if (file_exists($conf)) {
                 return realpath($conf);
-            }
-
-            if (file_exists($conf . '.dist')) {
+            } elseif (file_exists($conf . '.dist')) {
```
BackEndTea: No need for `elseif` over `if` since every `if` clause has a return statement.

src/TestFramework/Config/TestFrameworkConfigLocator.php
```diff
@@ -27,16 +27,17 @@ public function locate(string $testFrameworkName, string $customDir = null): str
...
                 return realpath($conf . '.dist');
+            } elseif (file_exists($altConf)) {
+                return realpath($altConf);
             }
         }
 
-        throw new \RuntimeException(sprintf('Unable to locate %s.(xml|yml)(.dist) file.', $testFrameworkName));
+        throw new \RuntimeException(sprintf('Unable to locate %s(.dist).(xml|yml)(.dist) file.', $testFrameworkName));
```
BackEndTea: This sentence is becoming a bit confusing, maybe we should have something like
`Unable to locate %s.(xml|yml) or dist version.` 

What do you think?

src/TestFramework/Config/TestFrameworkConfigLocator.php
```diff
@@ -27,16 +27,17 @@ public function locate(string $testFrameworkName, string $customDir = null): str
 
         foreach (['xml', 'yml'] as $extension) {
             $conf = sprintf('%s/%s.%s', $dir, $testFrameworkName, $extension);
+            $altConf = sprintf('%s/%s.dist.%s', $dir, $testFrameworkName, $extension);
```
sidz: Could we please do smth like:
```
$conf = sprintf('%s/%s.%s', $dir, $testFrameworkName, $extension);

if (file_exists($conf)) {
    return realpath($conf);
}

if (file_exists($conf . '.dist')) {
    return realpath($conf . '.dist');
}

$altConf = sprintf('%s/%s.dist.%s', $dir, $testFrameworkName, $extension);
if (file_exists($altConf)) {
    return realpath($altConf);
}
```
instead of if-elseif-elseif ?

src/TestFramework/Config/TestFrameworkConfigLocator.php
sidz: And also can we avoid of using RuntimeException? for we already have LocatorException class where we can specify our exception. I think it makes sense

sanmai:
```php
foreach (...) {
    //...
    $tried[] = $conf;
}

throw new X(sprintf('Unable to locate either of: %s', implode(', ', $tried));
```

src/TestFramework/Config/TestFrameworkConfigLocator.php
sanmai: no iffs and just loud and clear
```php
[
    'xml', 'yml',
    'xml.dist', 'yml.dist',
    'dist.xml', 'dist.yml',
]
```

src/TestFramework/Config/TestFrameworkConfigLocator.php
```diff
@@ -24,20 +26,33 @@ public function __construct(string $configDir)
...
         }
 
-        throw new \RuntimeException(sprintf('Unable to locate %s(.dist).(xml|yml)(.dist) file.', $testFrameworkName));
+        throw LocatorException::multipleFilesDoNoExist(
+            $dir,
+            $triedFiles
+        );
+    }
+
+    private function getDefaultExtensions() : array
```
sanmai: Can't this be a `const DEFAULT_EXTENSIONS = [ ... ];`? [Like it's done elsewhere.](https://github.com/infection/infection/blob/85ca245272a70db7ecd8e6db0de73fd3b3ec5a19/src/Mutator/Util/MutatorProfile.php#L16)

sidz: +1 to extract to the constant

Fenikkusu: It could be, but I'm thinking future proofing and maintaining solid coding. If the frameworks every overrode this class, the method is more effective. For example, since codeception uses .dist.yml, but not .yml.dist, the codeception testing code override this class and say specifically .yml.dist only. Const is too tightly coupled in my experience.

src/TestFramework/Config/TestFrameworkConfigLocator.php
```diff
@@ -24,19 +26,33 @@ public function __construct(string $configDir)
...
+    {
+        return [
+            'xml',
+            'yml',
+            'xml.dist',
+            'yml.dist',
+            'dist.xml',
+            'dist.yml'
```
sanmai: Missing comma here. Just run `composer cs:fix`

Probably the reason why Travis bailed

src/TestFramework/Config/TestFrameworkConfigLocator.php
sidz: If you want to override class like this you'll need to create your own Infection library as you don't have a chance to replace it via your own.

Also I'm pretty sure that we will drop 7.0 support at the end of this year and with 7.1 we can use constant visibility.

Fenikkusu: I'm actually referring specifically to the in-project frameworks (`./src/TestFramework`)

BackEndTea:
> (...Eww...Static...)

This is acceptable for throwing exceptions, since you don't want to create the exception, set its message, and then throw it. This way the exception itself is responsible for its formatting.

src/TestFramework/Config/TestFrameworkConfigLocator.php
borNfreee: > For example, since codeception uses .dist.yml, but not .yml.dist, the codeception testing code override this class and say specifically .yml.dist only. Const is too tightly coupled in my experience.

1. private method can't be overriden
2. it's possible to override const
3. since all such arrays we store in constants, let's store this one as well to be consistent across the project

Fenikkusu:
I've updated the code per requested, but I'm going on record to say I do not agree with the decision.

borNfreee:
@sanmai do you have an idea why e2e test with memory limit fails?

sanmai:
It looks like a Travis glitch, e.g. this suite run for over 6 minutes, while most others finished within 3 minutes. If you'll restart the build and it goes away, that was it.

For some reason I made Infection to timeout on this test super-quick. I'll make a PR to raise the timeout to the standard, it should be more reliable then.

borNfreee:
@sanmai It works after restart: https://travis-ci.org/infection/infection/jobs/359696861

Thank you @Fenikkusu 

_________

> I'm going on record to say I do not agree with the decision

We should stick to our agreements and coding standards, and every class should be consistent with the other classes in the package, even if it does not suit someone's CS/design choices. 

I hope this is not a big deal for you and I really appreciate your work!

Fenikkusu:
@borNfreee, I certainly agree that that one should follow the given guidelines and whatnot of a given project (hence having made the change despite my objection). However, I say that just as reference. I've been developing now for over a decade and I tend to follow Programming best practices, followed by PHP OOP best practices. As a result, my code structure does occasionally conflict with whatever framework and/or software I may be working in. One thing I've learned over the years though is that if I don't take the moment to voice my opinion, how is anyone else going to know the option exists.

For example, in this case, using the constants locks you down. You have no ability to expand it (specifically dynamically) outside of rewriting the code. With no more than 5 lines of code (I suspect), this could be easily updated to be overridden by the test framework itself (IE ./src/TestFrameworks/PHPUnit/ConfigLocator). If this were to be done, there is still no flexibility in the constants so the whole thing would have to altered anyways, hence using the method to begin with. 

Don't get me wrong, constants are an excellent source of resource, provided it's a resource that would/could never change. The file extensions, in my opinion, could very easily change. It's all about future-proofing the code. A mindset, if you will.

PR #270, opened 2018-03-28
Updated the PHP CS Fixer config

sanmai:
Updated the PHP CS Fixer config to follow established rules. No code change needed.

Avoids discussions like [this](https://github.com/infection/infection/pull/269#discussion_r177657790) and [this](https://github.com/infection/infection/pull/258#discussion_r177657524).

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -31,6 +31,11 @@ return \PhpCsFixer\Config::create()
         'phpdoc_align' => false,
         'phpdoc_summary' => false,
         'yoda_style' => false,
+        'ordered_class_elements' => [
+            'use_trait',
+            'constant_public',
+            'property_public',
+        ],
     ])
     ->setFinder($finder)
 ;
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -36,6 +36,7 @@ return \PhpCsFixer\Config::create()
             'constant_public',
             'property_public',
         ],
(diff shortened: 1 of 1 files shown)


.php_cs.dist
```diff
@@ -31,6 +31,12 @@ return \PhpCsFixer\Config::create()
         'phpdoc_align' => false,
         'phpdoc_summary' => false,
         'yoda_style' => false,
+        'ordered_class_elements' => [
+            'use_trait',
+            'constant_public',
+            'property_public',
```
borNfreee: Does it mean that other settings from this rules are left by default? I'm interested in `'method_public', 'method_protected', 'method_private'` - exactly in this order

sanmai: Other settings will be as before: unordered default. 

That's 103 files changed if you want them in this order. And say goodbye to `git blame` along the way. I wouldn't recommend this.

borNfreee: got it :(

BackEndTea: > That's 103 files changed if you want them in this order. And say goodbye to git blame along the way. I wouldn't recommend this.

If its something we really want, its better to just do it. The longer we wait, the bigger the impact of these things will be.

But if its not that important we can just skip it.

borNfreee:
Thank you @sanmai

PR #263, opened 2018-03-25
Replace custom XdebugHandler by composer/xdebug-handler package

sidz:
This PR is replacing our custom XdebugHandler by `composer/xdebug-handler` package

PR #275, opened 2018-03-29
Add a debug version of infection

BackEndTea:
This PR:

Adds a debug version of infection, which can be ran by `bin/infection-debug`. This is the most bare-bones version that will help with issues like this: https://github.com/infection/infection/pull/263#issuecomment-377037768

Another idea i played around with is some kind of PSR-3 logger which can be called statically, to help with debugging. But i'm not a fan of cluttering the source-code with tons of debug statements.

Running it on the current source code reveals this error right away:
`Not inheriting environment variables is deprecated since Symfony 3.3 and will always happen in 4.0. Set "Process::inheritEnvironmentVariables()" to true instead.`

### DIFF
--- a/bin/infection-debug
+++ b/bin/infection-debug
@@ -1,0 +1,8 @@
+#!/usr/bin/env php
+<?php
+
+set_error_handler(function($errno, $errstr) {
+    var_dump($errno, $errstr);
+}, E_ALL);
+
+require __DIR__ . '/infection';


theofidry:
What about adding a debug mode with the option `--debug`?

We could use [Symfony Debug](https://symfony.com/doc/current//components/debug.html). I however was not sure of how efficient it is and maybe having something closer to what the PR currently is but using [Symfony VarDumper](https://symfony.com/doc/current/components/var_dumper.html) to dump things which is way more user-friendly IMO

borNfreee:
> What about adding a debug mode with the option --debug?

Then it makes sense to think about `-vvv` as it states for debug in its description:

`-v|vv|vvv, --verbose                                       Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug`

>  Symfony VarDumper to dump things which is way more user-friendly IMO

While it's awesome, I think it's an overkill to use it just to dump error message

borNfreee:
can we merge it @BackEndTea ?

BackEndTea:
I think its good to merge as is, unless anyone is against it.

If we want to add more 'debug' functionality that can be done in a different PR.

borNfreee:
Agree. Thank you @BackEndTea !

BackEndTea:
Thank you @borNfreee @theofidry

PR #276, opened 2018-03-29
DX: use benefits of symfony/force-lowest

keradus:
(Sorry for removing template, I don't find it matching DX type of code changes)

Proposed change would bring 2 DX benefits:
- no need to list and manually maintain list of used Sf components in travis config file thanks to usage of meta-packages that already contains all Sf components (don't worry, this change won't make composer to install all components ;) )
- use `composer require` with `--no-update` will not trigger installation process, just change `composer.json` file. thanks to that we can first adjust composer file on the fly, and then install everything at once. (instead of currently installing version X, and then reinstalling to version Y)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,18 +15,18 @@ matrix:
     - php: 7.1
       env: PHPDBG=1
     - php: 7.1
-      env: SYMFONY_VERSION="^4.0"
+      env: SYMFONY_VERSION="v4.0"
     - php: 7.1
-      env: SYMFONY_VERSION="^4.0" PHPDBG=1
+      env: SYMFONY_VERSION="v4.0" PHPDBG=1
     - php: 7.2
     - php: 7.2
       env: NO_DEBUGGER=1
     - php: 7.2
       env: PHPDBG=1
     - php: 7.2
-      env: SYMFONY_VERSION="^4.0"
+      env: SYMFONY_VERSION="v4.0"
     - php: 7.2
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -45,16 +45,13 @@ before_install:
...
   - |
    if [ "${SYMFONY_VERSION}" != "" ]; then
       composer config --unset platform.php
-      composer require \
-        symfony/console:${SYMFONY_VERSION} \
-        symfony/process:${SYMFONY_VERSION} \
-        symfony/finder:${SYMFONY_VERSION} \
-        symfony/yaml:${SYMFONY_VERSION}
+      composer require --no-update symfony/force-lowest:$SYMFONY_VERSION
+      composer require --no-update symfony/lts:$(echo $SYMFONY_VERSION | grep -o 'v[0-9]\+') || true
```
sidz: the 4th version of the symfony is not a LTS and also https://packagist.org/packages/symfony/lts doesn't have `v4`. Did you check your approach locally?

p.s. tbh we've already tried to use `symfony/lts`

borNfreee:
> does `composer update` to update all deps makes sense to you?
> if not, let us kill the PR!

should I kill the PR? If not, what is the next step @keradus?
Don't want to keep dead PRs open

keradus:
I was waiting since march :D 
yeah, as I wrote, if proposed solution is not what you are willing to have, then I find no better way than current to handle deps.
for that, let us kill it.

borNfreee:
Thank you @keradus

PR #277, opened 2018-03-29
DX: PHP CS Fixer - groom usage

keradus:
DX only, vide commit msgs for reasoning of changes ;)

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -31,9 +31,11 @@ return \PhpCsFixer\Config::create()
         ],
         'no_useless_else'=> true,
         'ordered_class_elements' => [
-            'use_trait',
-            'constant_public',
-            'property_public',
+            'order' => [
+                'use_trait',
+                'constant_public',
+                'property_public',
+            ],
         ],
         'ordered_imports' => true,
         'phpdoc_align' => false,
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -17,16 +17,18 @@ return \PhpCsFixer\Config::create()
     ->setRiskyAllowed(true)
     ->setRules([
         '@Symfony' => true,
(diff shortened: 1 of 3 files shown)


.php_cs.dist
```diff
@@ -17,26 +17,30 @@ return \PhpCsFixer\Config::create()
     ->setRiskyAllowed(true)
     ->setRules([
         '@Symfony' => true,
+        '@Symfony:risky' => true,
         '@PHP70Migration' => true,
+        '@PHP70Migration:risky' => true,
 
         'concat_space' => ['spacing' => 'one'],
+        'declare_strict_types' => false,
```
borNfreee: could you please explain why `false`?

keradus: sure I can !
rule is not enabled on master, `@php7:risky` has it by default on, so I turned if off.
it is conflicting with header rule, ending up adding 2 headers (above and below declare statement), for that I didn't wanted to fix everything in single PR ;)

borNfreee: this is what @keradus is talking about:

```diff
<?php
/**
 * Copyright © 2017-2018 Maks Rafalko
 *
 * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
 */

declare(strict_types=1);
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
```

Is it a bug of PHP-CS-Fixer?

borNfreee:
Thank you @keradus

.php_cs.dist
keradus: this is known conflict between those 2 rules, we have raised issue for it for a while, nobody helped solving it so far

PR #278, opened 2018-03-29
Custom mutators

Stratadox:
This PR:

- Adds tests to make sure custom mutators can be used
  - Checks that the custom loader configuration gets loaded
  - Checks that custom mutator configuration adds a custom mutator instance

Because:

> As an enthusiastic infection tester,
> I want to add custom mutator classes,
> In order to unleash extra advanced mutation testing on my projects.

...and it turns out that yes, I can. This pull request makes that fact official by supplying the relevant tests.

### DIFF
--- a/tests/Config/InfectionConfigTest.php
+++ b/tests/Config/InfectionConfigTest.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace Infection\Tests\Config;
 
 use Infection\Config\InfectionConfig;
+use Infection\Tests\Fixtures\StubMutator;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Filesystem\Filesystem;
 use function Infection\Tests\normalizePath as p;
@@ -190,6 +191,22 @@ JSON;
                 ],
             ],
             (array) $config->getMutatorsConfiguration()['PublicVisibility']);
+    }
+
+    public function test_it_accepts_custom_mutators()
+    {
+        $config = <<<'JSON'
+{
+    "mutators": {
+        "Infection\\Tests\\Fixtures\\StubMutator": true
+    }
+}
+JSON;
+
+        $config = new InfectionConfig(json_decode($config), $this->filesystem, '/path/to/config');
+        $this->assertSame(
+            [StubMutator::class => true],
+            $config->getMutatorsConfiguration());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


theofidry:
I'm _really_ interested in this, I'm however not confident the current design is good enough. Here's a few concerns:

- I'm not sure the Mutants API is ready. While it's currently good enough for our mutators I'm still not sure about exposing it as an extension point to third-party. As Infection is still in dev (as didn't reach 1.0), we may also very easily break it which won't be pleasant for users
- It requires autoloading third-party code which is not trivial and may results in conflicts/issues. It requires at the very least a proper autoloading handling (being able to specify where the autoloader should be get from) and error handling
- As per the previous point, this is where PHP-Scoper is critical with the PHAR so we should finish that first and test it with an isolated config too

borNfreee:
> It requires autoloading third-party code which is not trivial and may results in conflicts/issues. It requires at the very least a proper autoloading handling (being able to specify where the autoloader should be get from) and error handling

Could you please elaborate on this? We already autoload project's code https://github.com/infection/infection/blob/master/app/bootstrap.php#L7-L11

If this is something we can work on - let's extract it to a separate issue

theofidry:
@borNfreee but that's infection own autoloading, not autoloading the tested project code isn't it?

Well, I need to check a couple of things related to it anyway so I'll come back for this

BackEndTea:
@theofidry I'm pretty sure that after 0.9.0 it does both.

borNfreee:
r

> @borNfreee but that's infection own autoloading, not autoloading the tested project code isn't it?

it autoloads project's code in order to create `\ReflectionClass`

BackEndTea:
> I'm not sure the Mutants API is ready. While it's currently good enough for our mutators I'm still not sure about exposing it as an extension point to third-party. As Infection is still in dev (as didn't reach 1.0), we may also very easily break it which won't be pleasant for users

This is the reason why i didn't add documentation for this feature when i added the posibility to do so. 

If we do add documentation for it, it should come  with a **BIG** disclaimer that it can break at any update.

borNfreee:
Ok guys, what do you propose with the API? 

Let's create concrete tasks (issues). I see no other way to move forward with it.

Anyway, regarding this PR. It still would be useful to create e2e test. I've remove documentation request for now

borNfreee:
rebased on master, resolved conflicts

merging it since these are just unit tests. Whe whole work of extension points will be done in separate PRs

maks-rafalko:
Full support of custom mutators has been released: https://github.com/infection/infection/releases/tag/0.29.0

PR #280, opened 2018-03-31
Function Scope Attribute Not Found

Fenikkusu:
Bug Fix - The file below resulted in the error 'Call to method getReturnType() on Non-Object'. I presume its just because this file is just a configuration/setup file. I've added the file to the excludes in my infection config, but wanted to go ahead and apply a JIC fix anyways.

`defines.const.php`
```php
<?php
/**
 * All Defines Enabled Here Are Intended To Be Non-Changing, Non-Dynamic Defines. They Shouldn't Be Overridden.
 */
namespace TwistersFury\Stuffs;

function define($constName, $constValue)
{
    if (!defined($constName)) {
        return \define($constName, $constValue);
    }

    return false;
}

/*
 * Debug Stuff
 */
define('TF_DEBUG_DISABLED', 0);
define('TF_DEBUG_ENABLED', 1);
define('TF_DEBUG_PHALCON', 2);
define('TF_DEBUG_SQL', 4);

define('TF_RUNNING_TESTS', false);

/*
 * Path Stuff
 */
define('DS', DIRECTORY_SEPARATOR);
```


src/Mutator/Util/AbstractValueToNullReturnValue.php
```diff
@@ -18,6 +18,9 @@ protected function isNullReturnValueAllowed(Node $node): bool
     {
         /** @var \PhpParser\Node\Stmt\Function_ $functionScope */
         $functionScope = $node->getAttribute(ReflectionVisitor::FUNCTION_SCOPE_KEY);
+        if (!$functionScope) {
```
borNfreee: Please add an empty line before if statement

Fenikkusu:
Soooooo...There is a file (tests/Mutator/ReturnValue/AbstractValueToNullReturnValueTestCase.php). This file does not appear to be used. Not only that, I suspect it should actually be in test/Mutator/Util to match the namespace of the original companion. My question at this point is A) Is it supped to be used. B) If not, can it be removed, or C) What needs to be done to resolve the lack there of?

Note: Was going to add logic for testing MR. Wasn't sure appropriate way to proceed.

borNfreee:
> There is a file (tests/Mutator/ReturnValue/AbstractValueToNullReturnValueTestCase.php). 

Ah.. this was a merge conflict. Checked - this file can be safely removed, it's not needed anymore.

Fenikkusu:
Pinging. Files updated and testing added in.

borNfreee:
Thank you @Fenikkusu

PR #281, opened 2018-03-31
Master config updates

Fenikkusu:
This PR:

- [x] Adds new feature ...
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/33

Most of my projects use Phalcon. Phalcon has it's own built in PSR-4 Autoloader that I tend to use over composer. When running infection against the system, I started getting errors because it couldn't find the classes. This PR does two things. The primary thing is it adds the ability to specify a bootstrap file that gets included on start. This took care of my class issue. Yes, I could have just updated composer, but I figure this is more useful.

The second thing this does is a side thought. I've added the ability to specify the test framework in config. This way you don't have to always type --test-framework= if using another framework.


src/Command/InfectionCommand.php
```diff
@@ -171,7 +172,17 @@ protected function configure()
...
+        $bootstrap = $container->get('infection.config')->getBootstrap();
+        if ($bootstrap) {
+            if (!file_exists($bootstrap)) {
+                throw LocatorException::fileOrDirectoryDoesNotExist($bootstrap);
+            }
+
+            require_once $bootstrap;
+        }
+
+        $testFrameworkKey = $input->getOption('test-framework') ?? $container->get('infection.config')->getTestFramework();
```
borNfreee: I believe `?:` instead of `??`

src/Command/InfectionCommand.php
```diff
@@ -171,7 +172,17 @@ protected function configure()
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $container = $this->getContainer();
-        $testFrameworkKey = $input->getOption('test-framework');
+
+        $bootstrap = $container->get('infection.config')->getBootstrap();
```
borNfreee: `$container->get('infection.config')` is used twice, could you please extract to variable?

tests/Fixtures/e2e/Config_Bootstrap/src/bootstrap.php
```diff
@@ -0,0 +1,2 @@
+<?php
+    file_put_contents(__DIR__ . './../infection-file.txt', 'Hello World!');
```
borNfreee: no indentation needed

tests/TestFramework/Config/InfectionConfigTest.php
```diff
@@ -0,0 +1,85 @@
+<?php
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+namespace Infection\Tests\TestFramework\Config;
```
borNfreee: Incorrect namespace. There is already a test file for `InfectionConfig` class: `tests/Config/InfectionConfigTest.php`

tests/TestFramework/Config/InfectionConfigTest.php
```diff
@@ -0,0 +1,85 @@
...
+    private $testSubject = null;
+
+    /** @var Filesystem */
+    private $mockFilesystem = null;
+
+    public function setUp()
+    {
+        parent::setUp();
+
+        $this->mockFilesystem = $this->getMockBuilder(Filesystem::class)
```
borNfreee: Should we use `Mockery` everywhere to be consistent?
Do we need a mock here?

tests/TestFramework/Config/InfectionConfigTest.php
```diff
@@ -0,0 +1,85 @@
...
+        $testSubject = new InfectionConfig(
+            $config,
+            $this->mockFilesystem,
+            ''
+        );
+
+        $this->assertEquals($result, $testSubject->{$methodName}());
+    }
+
+    public function dpConfig()
```
borNfreee: abbreviations is not a good practise, please use `configDataProvider` or smth similar

tests/TestFramework/Config/InfectionConfigTest.php
```diff
@@ -0,0 +1,85 @@
...
+    {
+        return [
```
borNfreee: 1. Let's use `yield`, we already started doing it.
2. Let's add a description for each dataset of provider

```php
public function configDataProvider(): \Generator
{
    yield 'It uses default test framework - phpunit' => [...];

    yield 'It returns configured test framework' => [...];
}

```

tests/Fixtures/e2e/Config_Framework/infection.json
```diff
@@ -0,0 +1,14 @@
...
+    "timeout": 25,
+    "source": {
+        "directories": [
+            "src"
+        ]
+    },
+    "logs": {
+        "debug": "infection-log.txt"
+    },
+    "system": {
```
borNfreee: I'm not sure we need a `system` key. It's ambiguous. Why `tmpDir` is on the root level, but `bootstrap` is under `system`? https://infection.github.io/guide/usage.html#Configuration

I suggest add `bootstrap` and `testFramework` to the root level to simplify json

src/Command/InfectionCommand.php
Fenikkusu: I updated the default value to be null (aka fall back to config/default). ?? is more appropriate in that situation.

tests/TestFramework/Config/InfectionConfigTest.php
Fenikkusu: The testSubject (InfectionConfig) requires a Filesystem object as a construct parameter. Haven't used Mockery before, but I presume it can be converted over easily enough. Personally, this is such a simple mock, I see no reason why to not just use PHPUnit's existing system.

tests/Fixtures/e2e/Config_Framework/infection.json
Fenikkusu: Personally, I think the opposite is more appropriate - move tmpDir (and others) to the system entry. In that way all the Infection specific settings are grouped together away from the testing framework configurations. Additionally, it might be better to name it 'config' instead of 'system' to follow the naming convention of composer, but that's neither here nor there.

tests/TestFramework/Config/InfectionConfigTest.php
Fenikkusu: Move to the other file. Since the other file was already using a Filesystem object, I just updated the testing to use it.

tests/Fixtures/e2e/Config_Framework/infection.json
Fenikkusu: After updating the site repo, changed my mind. I've converted it over to root.

Fenikkusu:
Documentation Update (Along With Original PR Desc), Applied Changes Per Requested

Fenikkusu:
I've updated the code, but wanted to point out something. While it is true the `??` was originally created as syntax sugar, it still serves other purposes. First, it works with functions, regardless of it is an array or not: https://3v4l.org/pfEG8 . Second, the primary difference between `??` and `?:` is that `??` only returns the second operand if the result of the first operation is null. If the result of the first operand is an empty string, for example, your result is an empty string. `?:` works differently in that in only returns the second operand when the first operand evaluates to false. If your function returns an empty string with `?:`, then you get the second operand. There are instances where `??` is move beneficial than `?:`. I felt this was one. If someone uses `--test-framework= --threads=4`, note the missing value, this, unless I'm mistaken, will return an empty string. This means when the tests run, they will get the error saying invalid framework. With ?:, it ends up falling back to the final 'phpunit' and run the testings using phpunit. While that may be what some intended, others may have just meant phpspec but forgot to complete the line. In such a case, they are going to start debugging a symptom of a bug, not the bug itself. Food for though.

borNfreee:
> Second, the primary difference between ?? and ?: is that ?? only returns the second operand if the result of the first operation is null. If the result of the first operand is an empty string, for example, your result is an empty string. ?: works differently in that in only returns the second operand when the first operand evaluates to false. If your function returns an empty string with ?:, then you get the second operand.

Good point btw

> If someone uses --test-framework= --threads=4, note the missing value, this, unless I'm mistaken, will return an empty string. This means when the tests run, they will get the error saying invalid framework. With ?:, it ends up falling back to the final 'phpunit' and run the testings using phpunit.

Just checked in this branch. Running infection as `infection --test-framework= --threads=4` throws an error

```
Invalid name of test framework. Available names are: phpunit, phpspec
```

so infection does not fallback to phpunit
--------

Thank you @Fenikkusu 👍

PR #249, opened 2018-03-22
Add some memory limits to the build to make sure CI works stable

sanmai:
This PR:

- [x] Introduces memory limits to the build.

Infection on Travis CI should give consistent results no matter what.

PR #282, opened 2018-03-31
Test that .travis.yml is valid

sanmai:
Since Travis CI [is known to silently fail if .travis.yml isn't valid](https://github.com/infection/infection/pull/249#issuecomment-375268364), we have a chance to check for that during routine testing, and at AppVeyor during normal CI.


theofidry:
I've just changed the Travis config from `Build only if .travis.yml is present`: `Off` => `On`. It doesn't look like that would fix it but that's the only parameter I see in the config that can remotely match what we are looking for. Could you try to see if that changes anything?

sanmai:
So Travis won't even build if .travis.yml is invalid. But good news is new test got it right and failed builds on AppVeyor.

sanmai:
Good idea. Done.

borNfreee:
Thank you @sanmai

PR #283, opened 2018-03-31
Making all non-extended classes final

sanmai:
This PR:

- [x] Makes most classes `final`, and some - `abstract`
- [x] Covered by tests

There was a discussion on if it is the right thing to [make constants instead of methods](https://github.com/infection/infection/pull/269#issuecomment-377230338), where it was decided it is proper so because every class should be consistent. Yet the entire discussion could have been avoided if the class in subject was unambiguously marked as not designed for inheritance.

This PR marks most classes (135 to be exact) that are not extended in this project as final. In some cases where classes were extensively mocked in tests, then interfaces were added to allow unrestricted mocking and future extension. In other cases these mocked classes were marked as `@internal`.

It is all not as bright. This change comes with a price:

- It will be slightly more difficult to refactor finalized classes.
- It will be slightly more difficult to navigate in code: given an abstract interface it won't be possible to jump to the concrete class with a single click right in the IDE.

Since there's work begun in #278 to allow externally provided mutators, this change also will declare what is not allowed to be extended, and what is more so, by implementing an interface, or by subclassing a known abstract class.

---

There's still more work to be done: a number of classes are still neither abstract nor final. It isn't possible to make them final outright because they're mocked in tests at times, or inherited from. Whenever it is the right thing to add an interface for each of these classes to allow mocking while making them final, is an open question. Probably best would be to address it in a separate PR. (All below were tagged as `@internal` for time being.)

| Mocked count | Class name |
|  ---- | ---- |
|13  |  ConsoleHelper           |
|9   |  CodeCoverageData        |
|3   |  VersionParser           |
|3   |  TestFrameworkFinder     |
|2   |  MutationConfigBuilder   |
|2   |  InitialConfigBuilder    |
|2   |  CoverageXmlParser       |
|1   |  XmlConfigurationHelper  |
|1   |  ProcessBuilder          |
|1   |  MetricsCalculator       |
|1   |  EventDispatcher         |
|1   |  Differ                  |
|1   |  BadgeApiClient          |
|0   |  XdebugHandler           |
|0   |  SourceDirGuesser        |
|0   |  InfectionException      |
|0   |  InfectionConfig         |
|0   |  DiffColorizer           |

```bash
# command for the report above
for class in $(git grep "^class" src/ | cut -f2 -d' ' | sort | uniq);
do echo "$(git grep $class::class | grep ::mock | wc -l) | $class |";
done | sort -rh | column -t | sed 's/^/|/'
```

### DIFF
--- a/src/Mutant/MutantInterface.php
+++ b/src/Mutant/MutantInterface.php
@@ -11,6 +11,9 @@ namespace Infection\Mutant;
 
 use Infection\MutationInterface;
 
+/**
+ * @see Mutant
+ */
 interface MutantInterface
 {
     public function getMutatedFilePath(): string;
--- a/src/Mutant/MutantInterface.php
+++ b/src/Mutant/MutantInterface.php
@@ -1,0 +1,25 @@
+<?php
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Mutant;
+
+use Infection\MutationInterface;
+
+interface MutantInterface
+{
+    public function getMutatedFilePath(): string;
+
+    public function getMutation(): MutationInterface;
+
+    public function getDiff(): string;
+
+    public function isCoveredByTest(): bool;
+
+    public function getCoverageTests(): array;
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,6 +61,7 @@ script:
   - if [[ $PHPDBG != 1 ]]; then xdebug-enable; fi
   - if [[ $PHPDBG != 1 ]]; then $PHPUNIT_BIN; else phpdbg -qrr $PHPUNIT_BIN; fi
   - ./tests/e2e_tests
+  - ./tests/final_private_test
   - if [[ $NO_DEBUGGER == 1 ]]; then xdebug-disable; fi
   - if [[ $PHPDBG != 1 ]]; then
       bin/infection $INFECTION_FLAGS;
--- a/composer.json
+++ b/composer.json
@@ -92,7 +92,8 @@
         ],
         "tests": [
             "./vendor/bin/phpunit",
-            "./tests/e2e_tests"
+            "./tests/e2e_tests",
+            "./tests/final_private_test"
         ]
     }
 }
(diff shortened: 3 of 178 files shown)


theofidry:
👍 on the move. I think having things locked on our end and open up extension points as we or others need them is a good default.

>In some cases where classes were extensively mocked in tests, then interfaces were added to allow unrestricted mocking and future extension.

In my experience interfaces could really be overkill in some of those cases. In which case I would propose we simply mark them as `@private` which hints the users that they shouldn't use it. If they are using an IDE like PhpStorm, PhpStorm will also refuse to provide the auto-completion for those marked as private classes. We can also eventually mention it in the doc like done [here](https://github.com/nelmio/alice#backward-compatibility-promise-bcp) if someone thinks this might not be enough.

sanmai:
Good point. I made the remaining classes `@internal`.

src/Mutant/MutantInterface.php
```diff
@@ -0,0 +1,25 @@
...
+
+declare(strict_types=1);
+
+namespace Infection\Mutant;
+
+use Infection\MutationInterface;
+
+interface MutantInterface
+{
+    public function getMutatedFilePath(): string;
```
sidz: suggestion: IMHO it'll be a good idea to add docblock with description on each method + class name with information about what each method usually do. In the same way like symfony has. (as example https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/Controller/ControllerResolverInterface.php)

p.s. feel free to leave as is if it doesn't make sense for u.

sanmai: Yes, that's sensible. I'll add a `@see` tag, but a full description is beyond the purpose of this PR.

sanmai: OK, done.

BackEndTea:
If we want to maintain this, we should have some kind of test for it. Maybe something like the php-cs-fixer auto review tests

sanmai:
This could work at CI:
```bash
for file in $(git grep -l ^class src/); do
    grep -q "@internal" $file || continue;
    sed -i~ 's/^class/final class/' $file;
    if php vendor/bin/phpunit 2>&1 > /dev/null; then
        echo "$(basename $file) should be made either final or internal";
        mv $file~ $file;
        exit 1;
    fi;
    mv $file~ $file;
done
```
Or even this:
```bash
for file in $(git grep -l ^class src/); do
    grep -q "@internal" $file || continue;
    echo "$file should be made either final or internal";
    exit 1;
done
```
That above if we decide to make disavow all inheritance. Currently 19 classes do no pass that test.

sanmai:
OK, so now there's a check in place. If there any non-private or non-final classes, the build will fail.

borNfreee:
👏 🥇 

Thank you @sanmai

PR #285, opened 2018-04-03
Run standard E2E tests from within PHPUnit

sanmai:
This PR:

- [x] (1) Runs most E2E tests from within PHPUnit
- [x] (2) Adds possibility to run Infection on itself within PHPUnit

Motivation: we have standalone e2e tests, but we can't tell what is their test coverage. Are they testing this or that? But if we run them from PHPUnit, we will know exactly which parts of code they cover, and which not. So we can write better tests.

Embedding E2E tests (1) adds 8 sec run time on warmed up E2E tests, bumps overall coverage to 75% (2191/2908 lines). Or even more: codecov/project thinks we get up to 79.8% (+18.88%). Suffice to say that's not the *solid* coverage, and Infections tells that right away. We had to lower requirement for Covered Code MSI from 86% to 72%. Still, [MSI normalized coverage](https://github.com/infection/infection/issues/290) is up 4% best-case.

With (2) the test suite runs well over one minute, for ~8% extra coverage. It is not enabled by default.

This PR considers a possibility that vendor dependencies from E2E tests may override stock vendor dependencies as explained in cc9d0e4. This was addressed in 287ec5e.


src/Console/Bootstrap.php
```diff
@@ -0,0 +1,186 @@
...
+use PhpParser\ParserFactory;
+use PhpParser\PrettyPrinter\Standard;
+use Pimple\Container;
+use SebastianBergmann\Diff\Differ as BaseDiffer;
+use Symfony\Component\Filesystem\Filesystem;
+
+/**
+ * @internal
+ */
+final class Bootstrap extends Container
```
sidz: any idea why it extends Container? Originally bootstrap is not a container.

sanmai: Not the best name, agreed.

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,156 @@
...
+    {
+        $process = new Process(sprintf('%s %s', (new ComposerExecutableFinder())->find(), 'install'));
+        $process->mustRun();
+
+        $this->previousLoader = include 'vendor/autoload.php';
+    }
+
+    private function runInfection(int $expectedExitCode = 0, array $argvExtra = []): string
+    {
+        if (!\extension_loaded('xdebug')) {
```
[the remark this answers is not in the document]
sanmai: Now that I could take a closer look at the failing builds, yes, we should.

sanmai:
Moving that here from the comments, cc @BackEndTea.

So far the fullest test suite [is failing on Travis with phpdbg](https://travis-ci.org/infection/infection/jobs/361660769#L735):
```
[PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 65536 bytes) 
in vendor/nikic/php-parser/lib/PhpParser/ParserAbstract.php on line 233]
```
1073741824 is about 1Gb. Probably this variant should not run on Travis. And not like I'm feeling brave to run the thing on my laptop with no memory limit. Dunno if we should keep it, at all.

src/Console/InfectionContainer.php
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * - Allow static analysis tools see what's behind an array item;
+ * - Localize all typing errors at their source;
+ * - And activate code completion, improved navigation.
+ *
+ * @internal
+ */
+final class InfectionContainer extends Container
+{
+    public static function getContainer(): Container
```
sidz: it looks like this method builds our container. Should it be `buildContainer` instead or even `build`?

src/Console/InfectionContainer.php
```diff
@@ -0,0 +1,193 @@
...
+
+        return $c;
+    }
+
+    private function getInfectionConfig(): InfectionConfig
+    {
+        return $this['infection.config'];
+    }
+
+    public static function prepareArgv($argv)
```
sidz: this method is not related with container. I'm not sure that this is the right place

app/container.php
```diff
@@ -7,144 +7,6 @@
...
-};
-
-$c['mutators'] = function (Container $c): array {
-    $mutatorConfig = $c['infection.config']->getMutatorsConfiguration();
-
-    return (new MutatorsGenerator($mutatorConfig))->generate();
-};
-
-return $c;
+return InfectionContainer::getContainer();
```
sidz: imo in this case this file is not needed and we can move this line right into `bootstrap.php`

src/Console/InfectionContainer.php
```diff
@@ -0,0 +1,193 @@
...
+        $c['locator'] = function (InfectionContainer $c): Locator {
+            return new Locator([$c['project.dir']], $c['filesystem']);
+        };
+
+        $c['path.replacer'] = function (InfectionContainer $c): PathReplacer {
+            return new PathReplacer($c['filesystem'], $c['phpunit.config.dir']);
+        };
+
+        $c['test.framework.factory'] = function (InfectionContainer $c): Factory {
+            return new Factory($c['tmp.dir'], $c['project.dir'], $c['testframework.config.locator'], $c['xml.configuration.helper'], $c['phpunit.junit.file.path'], $c->getInfectionConfig(), $c['version.parser']);
```
sidz: could you please do:
```
return new Factory(
    $dep1,
    $dep2,
    ...
);
```

src/Console/InfectionContainer.php
sanmai: I'm not sure too. That's just how it went. Any suggestions?

app/container.php
sanmai: I'm all in for its removal. Are you sure we can get rid of it?

src/Console/InfectionContainer.php
sanmai: Yes, it seems `build()` is a common name for similar methods.

.travis.yml
```diff
@@ -49,7 +50,7 @@ install:
   - |
    if [ "${SYMFONY_VERSION}" != "" ]; then
       composer config --unset platform.php
-      composer require \
+      travis_retry composer require \
```
theofidry: why is a `travis_retry` required here?

src/Command/ConfigureCommand.php
```diff
@@ -27,6 +27,8 @@
 
 final class ConfigureCommand extends BaseCommand
 {
+    const NONINTERACTIVE_MODE_ERROR = 'Infection config generator requires an interactive mode.';
```
theofidry: do we need to expose such messages via a constant? It feels a bit overkill plus right now they are exposed as public constants. As we're in 7.0 we cannot use `private const` but we could mark them with `@private`.

src/Command/InfectionCommand.php
```diff
@@ -276,8 +276,8 @@ private function applyMemoryLimitFromPhpUnitProcess(Process $process, PhpUnitAda
     {
         $tempConfigPath = \php_ini_loaded_file();
 
-        if (empty($tempConfigPath) || !file_exists($tempConfigPath)) {
-            // Cannot add a memory limit when there's no php.ini
+        if (empty($tempConfigPath) || !file_exists($tempConfigPath) || !is_writable($tempConfigPath)) {
+            // Cannot add a memory limit when there's no php.ini, or it isn't writeable
```
theofidry: `Cannot add a memory limit: there is no php.ini file or it is not writeable`?

bin/infection
```diff
@@ -10,27 +10,7 @@ require __DIR__ . '/../app/bootstrap.php';
...
-
-    if (!$found) {
-        array_splice($argv, 1, 0,'run');
-    }
-
-    return $argv;
-}
-
-$input = new ArgvInput(prepareArgv());
+$input = new ArgvInput(Application::prepareArgv($_SERVER['argv']));
```
theofidry: why do we need an `ArgvInput` at all? Can't we just leave it to the `Application::run()` to figure out which input to use and how to prepare it?

src/Console/InfectionContainer.php
```diff
@@ -0,0 +1,188 @@
...
+use PhpParser\Lexer;
+use PhpParser\Parser;
+use PhpParser\ParserFactory;
+use PhpParser\PrettyPrinter\Standard;
+use Pimple\Container;
+use SebastianBergmann\Diff\Differ as BaseDiffer;
+use Symfony\Component\Filesystem\Filesystem;
+
+/**
+ * InfectionContainer extending Container enables us
```
theofidry: I wonder if we need such phpdoc tbh

src/Console/InfectionContainer.php
```diff
@@ -0,0 +1,188 @@
...
+ *
+ * - Allow static analysis tools see what's behind an array item;
+ * - Localize all typing errors at their source;
+ * - And activate code completion, improved navigation.
+ *
+ * @internal
+ */
+final class InfectionContainer extends Container
+{
+    public static function build(): Container
```
theofidry: why a static `build()` instead of the regular `construct()`, do we need the container with a different configuration somewhere?

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,230 @@
...
+        }
+    }
+
+    protected function tearDown()
+    {
+        if ($this->previousLoader) {
+            $this->previousLoader->unregister();
+        }
+
+        \chdir($this->cwd);
```
theofidry: can we drop those leading backslashes? It's not a performance critical code and I think it just hurts readability in general. If we want to I recommend to import it via a `use` statement instead which can be configured in IDEs (Editor -> General -> Auto-Imports in PhpStorm). And without an automated step, like with a CS Fixer (and the one for the backslashes is only for a few opcache optimised functions and doesn't support the import), we're bound to miss some (like in the method just below)

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,230 @@
...
+
+        /*
+         * E2E tests are expected to follow PSR-4.
+         *
+         * We exploit this to autoload only classes belonging to the test,
+         * but not to vendored deps (so we don't need them here, but to run
+         * a testing tool from an E2E test we need them all the same).
+         */
+
+        $loader = new \Composer\Autoload\ClassLoader();
```
theofidry: this class can be imported via  a use statement

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,230 @@
...
+            copy('infection-log.txt', 'expected-output.txt');
+            $this->markTestSkipped('Saved golden output');
+        }
+
+        $this->assertFileEquals('expected-output.txt', 'infection-log.txt', sprintf('%s/expected-output.txt is not same as infection-log.txt', getcwd()));
+
+        return $output;
+    }
+
+    private function installComposerDeps()
```
theofidry: So if I understand correctly, here we install the deps + autoload the classes for the e2e test to make sure that if there's an auto-loading issue it's breaking here and not elsewhere because it would be harder to debug?

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,230 @@
...
+        // Leaving window open to negative tests (e.g. where Infection is expected to fail)
+        $this->assertEquals($expectedExitCode, $exitCode, "Unexpected exit code. Check with command's output for details.");
+
+        return $output->fetch();
+    }
+
+    private function requiresMemoryLimitSet()
+    {
+        if (ini_get('memory_limit') === '-1') {
+            // On a typical Linux with no default memory limit run:
```
theofidry: I feel like this comment could be in the skipped message :)

.travis.yml
sanmai: I had a failing build because composer could not install these here.

bin/infection
sanmai: It was that way before. I tried to keep most logic intact, since that’s beyond  the purpose of this PR

src/Command/ConfigureCommand.php
sanmai: How about if mark the entire class as private?

src/Command/InfectionCommand.php
sanmai: Yeah, when we don’t use XDebugHandler, there’s no temporary .ini file, and we can’t write to /etc/.../php.ini

src/Console/InfectionContainer.php
sanmai: I believe there was a question, but if that’s too much I’ll remove it

src/Console/InfectionContainer.php
sanmai: Well, I should have though about that! Let’s see if that’s doable.

tests/Console/E2ETest.php
sanmai: That’s right

sanmai: Well, that can be a problem. I can `include` all classes right away [instead of feeding them to the autoloader here](https://github.com/infection/infection/commit/287ec5e2e063643880e00b45d7828d2fbf5f15b6#diff-c50be878dbd21bc70fc3c4da7deedccfR144), but E2E tests could have (and they have I think) absolutely same classes, with the same content, and the same name. That's going to be an error. 

Infection sub-processes use actual files, but if it some point they happen to be different, this will be seen in different finals, and test will fail in this case.

tests/Console/E2ETest.php
sanmai: Good point. CS Fixer can add these leading backslashes all over the place if needed, but I don't think we need them here or anywhere, I'll remove them.

src/Command/InfectionCommand.php
sanmai: Oh. True, that's going to work better, thanks.

sanmai:
> what's the motivation for adding those e2e tests directly from PHPUnit

True, we had standalone e2e test, but we can't tell what is their test coverage. Are they testing this or that? But if we run them from PHPUnit, we would know exactly which parts of code they cover, and which not.

This already had a use. I found that there wasn't any coverage for the configure command, run it and found an issue #286, and then when I wasn't satisfied by the coverage, I tried to run configuration again, and hit another issue #288. Even this PR isn't merged yet, it is already helping fixing bugs.

.travis.yml
```diff
@@ -64,6 +65,7 @@ install:
 script:
   - composer analyze
   - if [[ $PHPDBG != 1 ]]; then xdebug-enable; fi
+  - if [[ $PHPDBG != 1 ]]; then $PHPUNIT_E2E; else phpdbg -qrr $PHPUNIT_E2E; fi
```
BackEndTea: I'm probably missing something here, but how is coverage generated from this, when it is ran separately from the other phpunit tests and without --coverage options

sanmai: That’s correct. Is there a way to join coverage from this run and the next? Is that going to make any sense at all? 

If not, I’ll just remove this and another line. We can have them both in one suite for some reason. It takes way over ten minutes to run on Travis.

BackEndTea: Maybe https://github.com/sebastianbergmann/phpcov may work

sanmai: Well, it appears I better spend time making actual tests work than making phpcov show kind-of total coverage where it can't be used by later Infection run anyway.

BackEndTea:
>It takes way over ten minutes to run on Travis.

Just the initial run, or does it cause the initial infection run to take that much longer?

Because if that is the case, we can run infection without the `--coverage` option, as to not include the e2e tests in the infection run itself.

sanmai:
> Just the initial run, or does it cause the initial infection run to take that much longer?

Infection on itself runs fine, the problem is when we run it on itself from under PHPUnit. [It looks like this.](https://travis-ci.org/infection/infection/jobs/361660768#L737) Probably because of XDebug making things a lot slower. Or, since this test alone runs fine, there could be other reason. Anyway, there should be better and quicker tests.

bin/infection
sanmai: $this->setDefaultCommand('run');

sanmai: Well, we can get rid of it with a single line of code, but then another issue appears. The issue can be fixed #294, but I more wary on why that's needed in the first place. I think it deserves a closer look in a separate PR, how do you think?

sanmai: OK never mind, I found the problem. Could you see if b02e5b4 makes any sense?

phpunit.xml
```diff
@@ -20,4 +20,11 @@
     <listeners>
         <listener class="\Mockery\Adapter\Phpunit\TestListener"></listener>
     </listeners>
+
+    <groups>
+        <exclude>
+            <group>large</group>
+        </exclude>
+    </groups>
+
```
sanmai: This is to exclude the full Infection-on-Infection test by default.

src/Command/ConfigureCommand.php
```diff
@@ -25,8 +25,13 @@
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
+/**
+ * @internal
+ */
```
sanmai: [Suggested here](https://github.com/infection/infection/pull/285#discussion_r179679765)

src/Command/ConfigureCommand.php
```diff
@@ -25,8 +25,13 @@
...
 final class ConfigureCommand extends BaseCommand
 {
+    const NONINTERACTIVE_MODE_ERROR = 'Infection config generator requires an interactive mode.';
+
```
sanmai: [Used in the test](https://github.com/infection/infection/pull/285/files#diff-c50be878dbd21bc70fc3c4da7deedccfR80) to confirm output of failing configure command.

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,243 @@
...
+    private $previousLoader;
+
+    protected function setUp()
+    {
+        // Without overcommit this test fails with `proc_open(): fork failed - Cannot allocate memory`
+        if (strpos(PHP_OS, 'Linux') === 0 &&
+            is_readable('/proc/sys/vm/overcommit_memory') &&
+            file_get_contents('/proc/sys/vm/overcommit_memory') == 2) {
+            $this->markTestSkipped('This test needs copious amounts of virtual memory. It will fail unless it is allowed to overcommit memory.');
+        }
```
sanmai: I've spend several hours trying to find the culprit only to find that I changed this setting while debugging a Redis install long time ago.

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,243 @@
...
+     */
+    public function test_it_runs_on_itself()
+    {
+        if (ini_get('memory_limit') === '-1') {
+            $this->markTestSkipped(implode("\n", [
+                'Refusing to run Infection on itself with no memory limit set: it is dangerous.',
+                'To run this test with a memory limit set please use:',
+                'php -dmemory_limit=128M vendor/bin/phpunit --group=large',
+            ]));
+        }
```
sanmai: There is a couple of places where a mutation triggers an infinite loop.

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,243 @@
...
+    {
+        $this->assertTrue(is_dir($path));
+        chdir($path);
+
+        $this->installComposerDeps();
+        $output = $this->runInfection(self::EXPECT_SUCCESS);
+
+        $this->assertRegExp('/You are running Infection with \w+ enabled./', $output);
+        $this->assertRegExp('/\d+ mutations were generated/', $output);
+        $this->assertRegExp('/\d+ mutants were killed/', $output);
```
sanmai: With this a test it'll fail early if there's an error in the output.

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,243 @@
...
+
+        if (isset($_SERVER['GOLDEN'])) {
+            copy('infection-log.txt', 'expected-output.txt');
+            $this->markTestSkipped('Saved golden output');
+        }
+
+        $this->assertFileEquals('expected-output.txt', 'infection-log.txt', sprintf('%s/expected-output.txt is not same as infection-log.txt (if that is OK, run GOLDEN=1 vendor/bin/phpunit)', getcwd()));
```
sanmai: Save some time copying files from here to where when adding a new mutator: running `GOLDEN=1 vendor/bin/phpunit` should be enough to cover most bases.

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,243 @@
...
+    private function installComposerDeps()
+    {
+        if (!file_exists('composer.json')) {
+            // Tests may have no deps to install
+            return;
+        }
+
+        if (!file_exists('vendor/autoload.php')) {
+            // Install deps only if there's none
+            $this->assertNotEmpty(getenv('PATH') ?: getenv('Path'), 'E2E tests need a system composer installed, but it could not be found without a PATH set');
```
sanmai: If we're not checking for PATH here, we get a similar puzzling TypeError as in #295

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,243 @@
...
+         */
+
+        $loader = new ClassLoader();
+
+        $map = require 'vendor/composer/autoload_psr4.php';
+
+        // $vendorDir is normally defined inside autoload_psr4.php, but PHPStan
+        // can't see there, so have to both tell it so, and verify that too
+        $vendorDir = $vendorDir ?? null;
+        $this->assertNotEmpty($vendorDir, 'Unexpected autoload_psr4.php found: please confirm that all dependincies are installed correctly for this fixture.');
```
sanmai: E.g. `composer install` could glitch leaving an empty autoload_psr4.php

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,243 @@
...
+
+    public function e2eTestSuiteDataProvider(): \Generator
+    {
+        $directories = Finder::create()
+            ->depth('== 0')
+            ->in(__DIR__ . '/../Fixtures/e2e/')
+            ->directories();
+
+        foreach ($directories as $dirName) {
+            if (file_exists($dirName . '/run_tests.bash')) {
```
borNfreee: Probably, it would be better to create an array of the tests we want to skip her, because just checking custom `run_tests.bash` file existence is too dirty check. It skips too many cases, even if the test is not "very slow to fail".

what do you think?

tests/Console/E2ETest.php
borNfreee: could you please explain why `GOLDEN`? ;) does it have special meaning in other technical places, just curious?

Also, I would say we should add this info to https://github.com/infection/infection/blob/master/.github/CONTRIBUTING.md (we will move dev documentation to the separate file or event to `infection/site` but that's definitely out of the scope of this PR)

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,243 @@
...
+        if (!file_exists('composer.json')) {
+            // Tests may have no deps to install
+            return;
+        }
+
+        if (!file_exists('vendor/autoload.php')) {
+            // Install deps only if there's none
+            $this->assertNotEmpty(getenv('PATH') ?: getenv('Path'), 'E2E tests need a system composer installed, but it could not be found without a PATH set');
+
+            $process = new Process(sprintf('%s %s', (new ComposerExecutableFinder())->find(), 'install'));
```
borNfreee: Probably, it makes sense to cache the result of `(new ComposerExecutableFinder())->find()` between test cases as it does not change

tests/Console/E2ETest.php
borNfreee: `dependincies` -> `dependencies`, typo

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,243 @@
...
+        $map = require 'vendor/composer/autoload_psr4.php';
+
+        // $vendorDir is normally defined inside autoload_psr4.php, but PHPStan
+        // can't see there, so have to both tell it so, and verify that too
+        $vendorDir = $vendorDir ?? null;
+        $this->assertNotEmpty($vendorDir, 'Unexpected autoload_psr4.php found: please confirm that all dependincies are installed correctly for this fixture.');
+
+        foreach ($map as $namespace => $paths) {
+            foreach ($paths as $path) {
+                if (strpos($path, $vendorDir) !== false) {
```
borNfreee: Could you please explain it with a bit more details why do we register just src/test classes here, but not classes from the e2eTest/vendor directory?

tests/Console/E2ETest.php
sanmai: We assume there's always an `expected-output.txt` present in the test directory. When a non-standard `run_tests.bash` is used we cannot be sure there's that file. I think it is better to follow a convention than add another knob to configure a test suite. One more thing to remember, and one more thing to overlook. If we can do without it, why not?

tests/Console/E2ETest.php
```diff
@@ -0,0 +1,243 @@
...
+     * @var \Composer\Autoload\ClassLoader|null
+     */
+    private $previousLoader;
+
+    protected function setUp()
+    {
+        // Without overcommit this test fails with `proc_open(): fork failed - Cannot allocate memory`
+        if (strpos(PHP_OS, 'Linux') === 0 &&
+            is_readable('/proc/sys/vm/overcommit_memory') &&
+            file_get_contents('/proc/sys/vm/overcommit_memory') == 2) {
```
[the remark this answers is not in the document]
sanmai: It is never going to be used anywhere, and we don't yet have `private const`

tests/Console/E2ETest.php
sanmai: Let's say a fixture uses a different version of PHPUnit than we are using for Infection. And if that versions happens to be API incompatible with our version, and if we autoload a class from a different version that has the *same* name, but *different* in regard of API, we're in trouble. 

[Here's how Theo explained it.](https://github.com/infection/infection/pull/285#discussion_r179682879)

tests/Console/E2ETest.php
borNfreee: We can do something like 

```php
if (file_exists('expected-output.txt')) {
    $this->assertFileEquals('expected-output.txt', 'infection-log.txt', ...);
}
```

then.

What I want to say is when we skip so many e2e tests (8 of 15 atm), we lose code coverage.

sanmai: If we're not asserting anything, like against known `expected-output.txt`, for what kind of coverage will be these tests for then? We already had to lower MSI requirement from 86% to 72%.

OK, we'll be testing against exit code, but these tests are testing against it too. Not sure if that's going to give us any coverage bump, this looks very much redundant.

tests/Console/E2ETest.php
sanmai: We can do it, but this part is usually done by `./tests/e2e_tests`. In most cases we never have to install any dependencies right from the test, but failing a test because of missing deps will be silly. In other words, we almost never have to run `(new ComposerExecutableFinder())->find()` from here.

Do you still think we need to optimize it here?

tests/Console/E2ETest.php
sanmai: OK, done

tests/Console/E2ETest.php
sanmai: I think `CONTRIBUTING.md` could be left alone in this case because if you would scroll to the right a failing test would say exactly what you'd want to do:

`$this->assertFileEquals('expected-output.txt', 'infection-log.txt', sprintf('%s/expected-output.txt is not same as infection-log.txt (if that is OK, run GOLDEN=1 vendor/bin/phpunit)', getcwd()));`

sanmai:
@borNfreee I think I've addressed all you comments. Type fixed. Is there anything else to be done, can you see?

tests/Console/E2ETest.php
borNfreee: Seems like I don't understand the code. Could you please help me understand what is going on in `installComposerDeps()` by answering to these quesitons:

1. We run `composer install` if needed - that's clear
2. We register a new `ClassLoader`, but it loads just files from `src` and `spec`/`tests` folder of each e2e test. Why just from these 2 folders? Why do we need to autoload them?
3. If we don't load classes from e2eTests/vendor, how does it work at all? :|

sanmai: Infection has no use for any classes from `vendor` folder. As for `src` files, we need to autoload them because Infection wants to see what's inside of them PHP-wise from the ReflectionVisitor.

For example, if you leave out this part (by not autoloading anything), Infection would fail with:
```
Generate mutants...

Processing source code files: 0/1
In ReflectionVisitor.php line 45:
                                               
  [ReflectionException (-1)]                   
  Class Namespace_\SourceClass does not exist  
                                               
Exception trace:
 ReflectionClass->__construct() at src/Visitor/ReflectionVisitor.php:45
 Infection\Visitor\ReflectionVisitor->enterNode() at vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:179
 PhpParser\NodeTraverser->traverseArray() at vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:108
 PhpParser\NodeTraverser->traverseNode() at vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:197
 PhpParser\NodeTraverser->traverseArray() at vendor/nikic/php-parser/lib/PhpParser/NodeTraverser.php:85
 PhpParser\NodeTraverser->traverse() at src/Mutant/Generator/MutationsGenerator.php:144
 Infection\Mutant\Generator\MutationsGenerator->getMutationsFromFile() at src/Mutant/Generator/MutationsGenerator.php:107
 Infection\Mutant\Generator\MutationsGenerator->generate() at src/Command/InfectionCommand.php:228
 Infection\Command\InfectionCommand->execute() at vendor/symfony/console/Command/Command.php:252
 Symfony\Component\Console\Command\Command->run() at vendor/symfony/console/Application.php:936
```

tests/Console/E2ETest.php
borNfreee: > If we're not asserting anything, like against known expected-output.txt, for what kind of coverage will be these tests for then?

Let's take `Ignore_MSI_Zero_Mutations` and `Test_Frameworks`: both folders has custom `run_tests.bash`, but they also have `expected-output.txt`. Why do we skip them? Just because some other test folder with custom `run_tests.bash` *probably* does not contain `expected-output.txt`? That's strange as for me.

In this case it's better to replace `if (file_exists($dirName . '/run_tests.bash')) { continue }` with `if (! file_exists($dirName . '/expected-output.txt')) { continue }`

tests/Console/E2ETest.php
borNfreee: no, thanks

tests/Console/E2ETest.php
sanmai: That'd be a bummer if we ever happen to add a *negative* test which would also have `expected-output.txt`. You can't say from `expected-output.txt` alone if you're looking at positive or negative test. Yet from missing run_tests.bash you can tell for certain - this a positive test, that should not err.

tests/Console/E2ETest.php
borNfreee: thanks, no problem

sanmai:
For some reason GitHub isn't showing all comments and replies here right on this page, but [they're all seen from the full diff](https://github.com/infection/infection/pull/285/files).

tests/Console/E2ETest.php
borNfreee: Ok, no I see.

So, just to double check with you: 

1. We autoload classes from src because Infection uses reflection and reflection should know how to autoload files, right?
2. We skip all files from e2e `vendor` folder **by purpose** because the can conflict with Infection's vendor dependencies, right?

If everything is correct above, could you please add (or update) your comments in `installComposerDeps()` method with this information about reflection and why we skip e2e `vendor` autoloading?

tests/Console/E2ETest.php
sanmai: If we'd like you say, a present `expected-output.txt` would mean that this *must* be a positive test. And if you would ever need to design a negative test, you would be forced to use a different name, like `expected-output2.txt`. That's silly, don't you think?

borNfreee: Yes, that's silly. But it is not very good to skip good e2e tests just because they have custom `run_tests.bash` while they are positive and have `expected-output.txt`. 

Any new positive test with `expected-output.txt` but with custom `run_tests.bash` will be never run by PHPUnit - this is the biggest concern I have.

borNfreee:
> Is there anything else to be done, can you see?

LGTM, except my concern about skipped current and future tests with custom `run_tests.bash` script.

tests/Console/E2ETest.php
sanmai: I'm not as concerned here because a *custom* test probably will require a *proper customized* procedure to run, which is [a piece of cake within the current architecture](https://github.com/infection/infection/pull/285/files#diff-c50be878dbd21bc70fc3c4da7deedccfR80). It is the right thing to do to a add a specialized testing procedure for a specialized non-standard test, don't you think?

sanmai: So my point is: 

- If you're adding a standard test, no worries everything is taken care for you.
- If you're adding a custom test, please be a good fellow and code a custom procedure for it.

borNfreee:
Thank you @sanmai

PR #287, opened 2018-04-04
Fix test framework detection in ConfigureCommand

### LINKED ISSUE #286: Initial Infection run fails with an error
| Question    | Answer
| Infection version | master

Initial Infection run on master fails with an error instead of starting a configuration master:

```
  The "--test-framework" option requires a value.  
```

git bisect points at commit f9bb8730af

sanmai:
This PR:

- [x] Adds an E2E test for ConfigureCommand in a non-interactive mode
- [x] Fixes #286

### DIFF
--- a/tests/Fixtures/e2e/Unconfigured/run_tests.bash
+++ b/tests/Fixtures/e2e/Unconfigured/run_tests.bash
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -e
+
+run () {
+    local INFECTION=${1}
+    local PHPARGS=${2}
+
+    if [ "$PHPDBG" = "1" ]
+    then
+        phpdbg $PHPARGS -qrr $INFECTION
+    else
+        php $PHPARGS $INFECTION
+    fi
+}
+
+test -x $(which tput) && tput setaf 2 # green
+if run "../../../../bin/infection" "" < /dev/null 2>&1 | grep -s 'Infection config generator requires an interactive mode.'; then
+	exit 0;
+fi
+
+test -x $(which tput) && tput setaf 1 # red
+echo "Infection configuration master did not start."
+
+exit 1;
+
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -74,7 +74,7 @@ final class InfectionCommand extends BaseCommand
                 null,
                 InputOption::VALUE_REQUIRED,
                 'Name of the Test framework to use (phpunit, phpspec)',
-                null
+                ''
             )
             ->addOption(
                 'test-framework-options',


tests/Fixtures/e2e/Unconfigured/run_tests.bash
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -e
+
+run () {
+    local INFECTION=${1}
+    local PHPARGS=${2}
```
borNfreee: I'm not sure it's needed here. Probably copy pasted?

tests/Fixtures/e2e/Unconfigured/run_tests.bash
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
borNfreee: the name of `Unconfigured` folder conflicts with the neighbor PR: https://github.com/infection/infection/pull/285/files#diff-c50be878dbd21bc70fc3c4da7deedccfR90

As I understand, it should have "_" to work properly

sanmai: It's not going to conflict, at least that's what `git merge` says to me. But I'm planning to rebase #285 over master once this merges anyway. Nothing to be concerned here.

tests/Fixtures/e2e/Unconfigured/run_tests.bash
sanmai: Yes, it is, but I decided to leave this function `run` as it is because it should go to a common with all tests library. With these left you can see that it is [the same function as in e.g. here](https://github.com/infection/infection/blob/master/tests/Fixtures/e2e/Memory_Limit/run_tests.bash#L5), so you would change this file to use the common library without any second thoughts. 

If that's unacceptable, that's not problem, I'll remove these.

borNfreee: thanks, that's ok

tests/Fixtures/e2e/Unconfigured/run_tests.bash
borNfreee: I mean not a git conflict, but logic conflict `Finder::create()->name('*_*')`. As far as I understand it searches all directories with `_` inside, thus `Unconfigured` will be skipped.

Anyway, let's merge it and adress (if it is a case) it in #285

borNfreee:
Thank you @sanmai

tests/Fixtures/e2e/Unconfigured/run_tests.bash
sanmai: Good point, I will drop name requirement.

(Since this is a negative test, Infection fails with an error on it, #285 addresses this test [with a separate test case](https://github.com/infection/infection/pull/285/files#diff-c50be878dbd21bc70fc3c4da7deedccfR66) just for it. During normal E2E tests Infection is not expected to fail.)

PR #293, opened 2018-04-06
Add Protected and Public Visiblity tests

BackEndTea:
This PR:
Adds a test to prove that non-inherited function visibility still gets mutated.


borNfreee:
Thank you @BackEndTea 👍

BackEndTea:
Thank you @borNfreee

PR #296, opened 2018-04-07
Problem with a custom PATH

sanmai:
Subprocesses may need to have a custom PATH to run commands of their choice: e.g. one of our subjects may want to run a bash script. Infection should pass through `PATH` (and `Path`) as it is, at very least.


borNfreee:
Tested locally, works well.
@sidz do you have any objections here as you are more familiar with this part?

borNfreee:
Thank you @sanmai

Issue #271, opened 2018-03-28
Infection can't report on effectiveness of mutators


sanmai:
| Question    | Answer
| ------------| ---------------
| Infection version | <0.9

Infection can't report on effectiveness of mutators it use. Effectiveness in a sense of number of mutations created, killed, escaped, and errorred. Such report must be mandatory for every PR with a new mutator, and can have other uses.

For example, from #243:

| Mutator          | Mutations     | Killed        | Escaped       | Errors       | 
| ---------------- | ------------- | ------------- | ------------- |------------- |
| Assign           | 437           | 281           | 27            | 1            |
| MethodCallFalse  | 1275          | 583           | 116           | 3            |
| MethodCallNull   | 1275          | 582           | 117           | 3            |
| MethodCallTrue   | 1275          | 563           | 135           | 4            |

BackEndTea:
What about a markdown log type, which prints info like that?

sanmai:
I'm good with markdown, and if there would be JSON too, that'd be best.

PR #297, opened 2018-04-08
[Mutator] Remove type cast operators

borNfreee:
Here is an RFC for a new set of mutators that remove type casting. I find them useful because for example these mutators found several places in Infection itself where type casting is redundant and not needed

- [x] Covered by tests
- [x] Other mutators from http://php.net/manual/en/language.types.type-juggling.php#language.types.typecasting `(object)`, `(unset)`, etc.
- [x] Doc PR: https://github.com/infection/site/commit/02b017c01cbd780a43f6ebf49d8b013c634ed5dc

To kill several mutant, a couple of new tests were added that cover type casting logic.

I would like to hear your feedback (objections if any) before moving forward

### DIFF
--- a/src/Mutator/Cast/CastArray.php
+++ b/src/Mutator/Cast/CastArray.php
@@ -12,7 +12,7 @@ namespace Infection\Mutator\Cast;
 use Infection\Mutator\Util\Mutator;
 use PhpParser\Node;
 
-class CastInt extends Mutator
+class CastArray extends Mutator
 {
     public function mutate(Node $node)
     {
@@ -21,6 +21,6 @@ class CastInt extends Mutator
 
     protected function mutatesNode(Node $node): bool
     {
-        return $node instanceof Node\Expr\Cast\Int_;
+        return $node instanceof Node\Expr\Cast\Array_;
     }
 }
--- a/src/Mutator/Cast/CastArray.php
+++ b/src/Mutator/Cast/CastArray.php
@@ -9,16 +9,10 @@ declare(strict_types=1);
 
 namespace Infection\Mutator\Cast;
 
-use Infection\Mutator\Util\Mutator;
 use PhpParser\Node;
 
-class CastArray extends Mutator
+final class CastArray extends AbstractCastMutator
 {
-    public function mutate(Node $node)
-    {
-        return $node->expr;
-    }
-
     protected function mutatesNode(Node $node): bool
     {
         return $node instanceof Node\Expr\Cast\Array_;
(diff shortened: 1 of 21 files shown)


src/Mutator/Util/MutatorsGenerator.php
```diff
@@ -99,7 +99,7 @@ private function registerFromClass(string $mutator, $setting)
                 $this->mutatorList[$mutator] = [];
             }
         } elseif (!array_key_exists($mutator, $this->mutatorList) || $this->mutatorList[$mutator] === []) {
-            $this->mutatorList[$mutator] = (array) $setting;
```
borNfreee: we support `true`, `false` and an `array`. First two are handled in the code above, that's why this type casting is not needed

src/TestFramework/PhpUnit/Coverage/CoverageXmlParser.php
```diff
@@ -153,7 +153,7 @@ private function getCoveredLinesData(\DOMNodeList $lineCoverageNodes): array
 
         foreach ($lineCoverageNodes as $lineCoverageNode) {
             /** @var \DOMNode $lineCoverageNode */
-            $lineNumber = (int) $lineCoverageNode->getAttribute('nr');
```
borNfreee: `$lineNumber` is used as a key in the array `$array[$lineNumber]`. PHP automatically converts all numeric strings to integer keys so no type casting is needed

src/Mutator/Util/MutatorsGenerator.php
sanmai: Hands down type conversion here is beyond MutatorsGenerator responsibilities. No wonder you wanted to remove i. At some point it should be refactored out of here, but probably not right now.

src/Mutator/Cast/CastArray.php
```diff
@@ -0,0 +1,26 @@
...
+use PhpParser\Node;
+
+class CastArray extends Mutator
+{
+    public function mutate(Node $node)
+    {
+        return $node->expr;
+    }
+
+    protected function mutatesNode(Node $node): bool
```
sanmai: This gets me thinking in the line of...

```php
abstract class AbstractCastMutator
{

    final public function mutate(Node $node) { ... }

    protected function mutatesNode(Node $node): bool
    {
        return $node instanceof Node\Expr\Cast;
    }
}

class CastArray extends AbstractCastMutator
{
    protected function mutatesNode(Node $node): bool
    {
         parent::mutatesNode(Node $node) && ....;
    }
}
```
WDYT?

sanmai:
There's seven cast-type nodes in total, including `(null)` cast. Guess there could be a single mutator to cover them all, but I find it interesting to know if there are some casts that are needlessly overused. If mutators are kept separate, this will be a piece of cake within #271.

This mutator will offend and enrage type purists without any doubt, I already love it just for that.

borNfreee:
Thanks @sanmai for the review

Issue #262, opened 2018-03-25
Syntax errors are not accounted for, not counted


sanmai:
As of writing Infection does not distinguish between different types of errors, of which there are at least several groups:

- Common language errors (`Error: Call to protected method`)
- Memory errors (`Fatal error: Allowed memory size of`, see #258)
- Syntax errors (we should not have them, but how do we know if we do not account for them)
- File errors (e.g. `flock(): Illegal operation argument` could be a temporary glitch)
- Other errors

That's just some types that I think worth counting and reporting separately. There could be more.

[First reported here.](https://github.com/infection/infection/pull/258#pullrequestreview-106702392)

Related: #222

borNfreee:
> Syntax errors (we should not have them, but how do we know if we do not account for them)

I thought PHP-Parser should always generate a valid PHP code, but can't find a proof in its documentation anymore. Probably I mixed up something.

Also related https://github.com/infection/infection/issues/222#issue-305355702 (see proposed report): 

```
6 mutations were generated:
       5 mutants were killed:
             - 5 tests failures
             - 0 PHP errors
             - 0 time outs
```

So we can add those groups to the report above

sanmai:
> I thought PHP-Parser should always generate valid PHP code

Well, we can safely say that it is proven that it doesn't always generate a correct code.

Errors from mutators from #243 have:

```diff
-        $updater->getStrategy()->setPackageName(self::PACKAGE_NAME);
+        false->setPackageName(self::PACKAGE_NAME);
```

And that is a parse error: https://3v4l.org/kZNGg
```
PHP Parse error:  syntax error, unexpected '->' (T_OBJECT_OPERATOR)
```

Therefore we must account for syntax errors.

theofidry:
I think we should try to be smarter and account for useless/equivalent mutants at some point.

To expand a bit: creating a mutant is relatively easy and requires little resources. However evaluating a mutant is way more expansive (we need to run the tests). There is a report somewhere mentioning about ~30% of the generated mutants are redundant hence not useful. I still need to check the effective strategies we could employ, but meanwhile checking for a PHP validity would be a good start and removing a bunch of mutants that provide no value.

BackEndTea:
What if we could lint the 'new' PHP before writing it to a file? And if it contains a syntax error ignore it.

sanmai:
There’s even php_check_syntax

borNfreee:
Well, as @sanmai already pointed

> we should not have them (syntax errors)

[...]

PR #301, opened 2018-04-09
Ensure mutated code is always valid

borNfreee:
This PR adds a check *in unit tests* that mutated code is always valid.

Extracted from https://github.com/infection/infection/issues/262 and [this](https://github.com/infection/infection/issues/262#issuecomment-379547689) comment in particular

For reference: 

> Created Mutant with invalid syntax must be considered as a bug in Infection.

Concerns about this PR:

* It slows down Mutators tests (from 5s to 15s)
* I'm not sure it's the best way of PHP linting, but this is the simplest one that I was able to find

With this new assertion we are sure that at least tested cases produce valid code for each Mutator.

Thoughts?

### DIFF
--- a/tests/Mutator/AbstractMutatorTestCase.php
+++ b/tests/Mutator/AbstractMutatorTestCase.php
@@ -12,7 +12,6 @@ namespace Infection\Tests\Mutator;
 use Infection\Mutator\Util\Mutator;
 use Infection\Mutator\Util\MutatorConfig;
 use Infection\Tests\Fixtures\SimpleMutatorVisitor;
-use Infection\Utils\TmpDirectoryCreator;
 use Infection\Visitor\CloneVisitor;
 use Infection\Visitor\FullyQualifiedClassNameVisitor;
 use Infection\Visitor\ParentConnectorVisitor;
@@ -22,7 +21,6 @@ use PhpParser\NodeTraverser;
 use PhpParser\ParserFactory;
 use PhpParser\PrettyPrinter\Standard;
 use PHPUnit\Framework\TestCase;
-use Symfony\Component\Filesystem\Filesystem;
 
 abstract class AbstractMutatorTestCase extends TestCase
 {
@@ -30,21 +28,6 @@ abstract class AbstractMutatorTestCase extends TestCase
      * @var Mutator
      */
     protected $mutator;
-
-    /**
-     * @var Filesystem
-     */
-    private $fileSystem;
-
-    /**
-     * @var string
-     */
-    private $workspace;
-
-    /**
-     * @var string
-     */
-    private $tmpDir;
 
     public function doTest(string $inputCode, string $expectedCode = null)
     {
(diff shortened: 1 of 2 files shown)


tests/Mutator/Boolean/Yield_Test.php
```diff
@@ -27,13 +27,19 @@ public function provideMutationCases(): \Generator
             <<<'PHP'
 <?php
 
-(yield $a => $b);
+function test()
```
borNfreee: Before this change, this mutator was failing with `PHP Fatal error:  The "yield" expression can only be used inside a function in /tmp/XXX`

sanmai:
IMO we should not do this *unless there's a particular need* because this is a slow procedure, and because all files will be linted by phpunit/etc nevertheless.

Speaking of particular needs, this should be done for our own tests.

Other than our own needs, I'm sure PHPUnit can be made to report fatal errors, and we can parse them just like we're doing it now. (For that we would need a "broken" mutator or two, that will always produce a code with a syntax error. Next is E2E test case with this mutator, and so on with detection.)

sanmai:
Because of the slowdowns IMO we should not do this in general and by default. By request and during CI - that's sure.

Another way to lint is to use `fork` - `eval($code)` - `waitpid` cycle. And then look at the exit status 
`pcntl_wexitstatus` to see if there was an error or now. Since this is our code, we can guarantee lack of
side effects, and with no side effects we can expect no infinite cycles. This will be as expensive as a `fork` call. I believe `exec` is more costly. Please disregard my previous comment.

Scheme with fork could be even faster if we won't wait for a single process to finish, but rather collect all exit statuses later at some point.

theofidry:
I'm not a big fan :/

While I do agree it would be better to keep generating valid code, there is two things I'm not fond of in that PR:

- It is slower. The point of avoiding to generate redundant or invalid mutants would be to be faster because evaluating them would be more expansive than checking if they should be created in the first place; It doesn't look like to be the case here as highlighted by @sanmai and your perf results
- It relies on the file-system. This is less of a concern right now but in the future it would be nice if we could do some monkey-patching via the autoloading rather than dumping the file and be dependent on the FS which is harder to parallelise properly and slower. That said we're not there yet so it's not the main point here

sanmai:
Here's a PoC of forking linter:
```php
function eval_fork($code)
{
    $pid = pcntl_fork();

    if ($pid == 0) {
        // Child
        error_reporting(0);
        eval($code);
        exit();
    }

    if ($pid == -1) {
        return false;
    }

    // we are the parent
    pcntl_wait($status);

    return 0 == $status;
}

var_dump(eval_fork('$a = 1;')); // true

var_dump(eval_fork('$a = 1 /* ; */')); // false
```

tests/Mutator/AbstractMutatorTestCase.php
```diff
@@ -81,4 +82,18 @@ protected function mutate(string $code)
 
         return $prettyPrinter->prettyPrintFile($mutatedNodes);
     }
+
+    private function assertSyntaxIsValid(string $realMutatedCode)
+    {
+        exec(sprintf('echo %s | php -l', escapeshellarg($realMutatedCode)), $output, $returnCode);
```
sanmai: `proc_open` seems like a good fit here

theofidry:
@sanmai should we start to investigate solutions like [amphp](amphp.org)? I used them for Box to process stuff in parallel and so far it was pretty good. Much more performant that whatever was done with the Symfony Process which has a design flaw for parallel processing

sanmai:
> I'm pretty sure there can be some difficulties with it, e.g. with autoloading.

We can no-op the code with `if (false) { ... code ... }` and/or with `function() { ... code ... };`, or like so.

theofidry:
@borNfreee nevermind my comment; I should have checked the diff :)

borNfreee:
@sanmai with `proc_open_linter` I have the same `100ms` as with `exec()`, it has the same performance.

I would stick to `exec()` then, because:

* it's a one-liner
* it takes the same time
* it works even on windows

Regarding `pcntl_fork()`: it requires additional extension to be installed. Neither my local PHP setup nor our Docker images have it (and I pretty much sure many of the developers as well). So this is an additional dependency that we can avoid

borNfreee:
Thank you @sanmai for your help

PR #304, opened 2018-04-16
Use our own Docker images for all testing

sanmai:
- Install expect because it is needed to e2e testing
- Skip XDebugHandler tests if xdebug isn't used
- Permit parallel execution e.g. with `make -j -Otarget test`

TODO:

- [x] blocked by #329
- [x] `make test-infection-xdebug-70` is failing with `Code Coverage does not exist` for some reason.
- [x] `make test-e2e-xdebug-72` finds previously undiscovered mutants

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ build/cache:
 #All tests, (infection itself, phpunit, e2e) for different php version/ environments (xdebug or phpdbg)
 .PHONY: test test-unit test-infection-phpdbg test-e2e-phpdbg test-infection-xdebug test-e2e-xdebug
 test: test-unit test-infection-phpdbg test-e2e-phpdbg test-infection-xdebug test-e2e-xdebug
+	# All tests finished without errors
 
 .PHONY: test-unit test-unit-70 test-unit-71 test-unit-72
 #php unit tests
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ PHPSTAN=./phpstan.phar
 PHP-CS-FIXER=./php-cs-fixer-v2.phar
 PHPUNIT=vendor/bin/phpunit
 
-DOCKER_RUN=docker run -t --rm -e PHPDBG=1 -v "$$PWD":/opt/infection -w /opt/infection
+DOCKER_RUN=docker run -t --rm -v "$$PWD":/opt/infection -w /opt/infection
 DOCKER_RUN_70=$(DOCKER_RUN) infection_php70
 DOCKER_RUN_71=$(DOCKER_RUN) infection_php71
 DOCKER_RUN_72=$(DOCKER_RUN) infection_php72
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,6 +1,12 @@
 PHP_CS_FIXER_FUTURE_MODE=1
 PHPSTAN=./phpstan.phar
 PHP-CS-FIXER=./php-cs-fixer-v2.phar
+PHPUNIT=vendor/bin/phpunit
+
+DOCKER_RUN=docker run -t --rm -e PHPDBG=1 -v "$$PWD":/opt/infection -w /opt/infection
```
sanmai: Why do we need `-e PHPDBG=1` here?

sanmai: Changing `-it` to `-t` allows parallel execution.

Makefile
```diff
@@ -33,82 +39,88 @@ test-final-private:
...
-
-build-xdebug-71: vendor
-	docker build -t infection_php71 -f "$$PWD/devTools/Dockerfile-php71-xdebug" .
-
-build-xdebug-72: vendor
-	docker build -t infection_php72 -f "$$PWD/devTools/Dockerfile-php72-xdebug" .
+build-xdebug-70: vendor devTools/Dockerfile-php70-xdebug.json
+devTools/Dockerfile-php70-xdebug.json: devTools/Dockerfile-php70-xdebug
+	docker build -t infection_php70 -f devTools/Dockerfile-php70-xdebug .
+	docker image inspect infection_php70 >> devTools/Dockerfile-php70-xdebug.json
```
sanmai: Make needs something to pin on. Therefore these .json placeholders: if a source file changes, it'll trigger a new `docker build`

Makefile
BackEndTea: ` -e PHPDBG=1` causes the e2e tests to execute with PHPDBG

Makefile
BackEndTea: Never knew you could do that :+1:

BackEndTea:
Should we also install composer on these docker images? This means we can completely run the tests within the docker environment.

sanmai:
Well, that's a problem. Our Docker images use Debian jessie, and you can't `apt-get install composer` there. Sure, this can be without `apt-get` in an old-fashioned way, but I think this better be a separate PR, what do you think?

sanmai:
OK, I think we're done here.
```
$ time make -j -Otarget

# All tests finished without errors

real	9m7,887s
user	0m16,331s
sys	0m13,912s
```

Makefile
```diff
@@ -1,6 +1,12 @@
 PHP_CS_FIXER_FUTURE_MODE=1
 PHPSTAN=./phpstan.phar
 PHP-CS-FIXER=./php-cs-fixer-v2.phar
+PHPUNIT=vendor/bin/phpunit
+
+DOCKER_RUN=docker run -t --rm -v "$$PWD":/opt/infection -w /opt/infection
+DOCKER_RUN_70=flock devTools/*php70*.json $(DOCKER_RUN) infection_php70
```
borNfreee: could you please explain why we need `flock` here?

devTools/Dockerfile-php70-xdebug
```diff
@@ -1,4 +1,7 @@
 FROM php:7.0-cli
 
 RUN pecl install xdebug && docker-php-ext-enable xdebug
+RUN apt-get -y update && apt-get install -y expect
+RUN useradd --shell /bin/bash infection
```
borNfreee: am I right we need it to automatically use `bash` when we are on docker?

sanmai: We just follow the principle of the least surprise here, because people are used to bash, and the default shell in most Linux'es these days is `/bin/sh`, which is, well, quite ugly outside of scripts. That is to say there's no particular need to use `/bin/bash` as a shell here, just a matter of future convenience.

Makefile
sanmai: `make -j` will start `docker run` for all targets at once, it will be like if you start `phpunit` multiple times at once. I believe we can't run tests in parallel now because of some issues, which may not be worth fixing. Hence we have to restrict Make from running PHP-version specific targets, which use the same VM, in parallel.

borNfreee:
Thank you @sanmai

PR #307, opened 2018-04-17
InfectionContainer now implements PSR-11 ContainerInterface

sanmai:
This PR:

- [x] InfectionContainer made implement ContainerInterface, one wrapper less
- [x] InfectionCommand directly uses InfectionContainer to access InfectionConfig
- [x] Minor refactoring in Application: check agains phpdbg like it is done against xdebug
- [x] Min MSI 57 -> 62, min covered MSI 86 -> 87
- [ ] Address review comments

### DIFF
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -64,12 +64,12 @@ ASCII;
      */
     private $isDebuggerDisabled;
 
-    public function __construct(Container $container, string $name = self::NAME, string $version = self::VERSION)
+    public function __construct(InfectionContainer $container, string $name = self::NAME, string $version = self::VERSION)
     {
         $this->container = $container;
-        $this->isDebuggerDisabled = ('' === trim((string) getenv(XdebugHandler::ENV_DISABLE_XDEBUG)));
-        $this->isXdebugLoaded = \extension_loaded('xdebug');
-        $this->isUnderPhpdbg = PHP_SAPI == 'phpdbg';
+        $this->isDebuggerDisabled = $this->container['environment.debugger.disabled'];
+        $this->isXdebugLoaded = $this->container['environment.xdebug.loaded'];
+        $this->isUnderPhpdbg = $this->container['environment.phpdbg.used'];
 
         parent::__construct($name, $version);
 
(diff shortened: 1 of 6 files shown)


src/Console/InfectionContainer.php
```diff
@@ -170,10 +174,47 @@ public function __construct(array $values = [])
...
+            return PHP_SAPI == 'phpdbg';
+        };
+    }
+
+    public function setInfectionConfigInitializer(\Closure $initializer)
+    {
+        $this[self::KEY_INFECTION_CONFIG] = $initializer;
+    }
+
+    public function getInfectionConfig(): InfectionConfig
```
borNfreee: what is so special with infection config that it has its own method?

If we are implementing `Psr\Container\ContainerInterface`, I would stick to its public interface without any additional methods, preserving `Psr\Container\ContainerInterface` typehints rather than our `InfectionContainer`. Otherwise, why do we need to implement `Psr\Container\ContainerInterface`?

src/Console/Application.php
```diff
@@ -54,16 +54,22 @@
...
      */
     private $isDebuggerDisabled;
 
-    public function __construct(Container $container, string $name = self::NAME, string $version = self::VERSION)
+    public function __construct(InfectionContainer $container, string $name = self::NAME, string $version = self::VERSION)
     {
         $this->container = $container;
-        $this->isDebuggerDisabled = ('' === trim((string) getenv(XdebugHandler::ENV_DISABLE_XDEBUG)));
-        $this->isXdebugLoaded = \extension_loaded('xdebug');
+        $this->isDebuggerDisabled = $this->container['environment.debugger.disabled'];
```
borNfreee: Since our container is a `Psr\Container\ContainerInterface` (and no more), we should use `$this->container->get('environment.debugger.disabled')` instead. I guess not all PSR-11 containers implement `ArrayAccess`

Otherwise, why do we need to implement `Psr\Container\ContainerInterface`?

BackEndTea:
If we are implementing the PSR-11 interface in our own code, we should add to our dependencies.

PR #308, opened 2018-04-18
Remove and add PHP_INI_SCAN_DIR at the right places

### LINKED ISSUE #306: Initial test suite may fail to deliver coverage, e.g. under Docker

### Steps to reproduce

```bash
```

### Expected result 

### Actual result

```
                                                                    
There was 1 error:                                                                                            
                                                                                                              
```

### Discussion

```patch
index 3b96bba..b589beb 100644
 
 {
+    {
+    }
+
     {
```
Outputs:
```
                                                                                          
STDOUT:                                                                                   
                                                                                          
F                                                                                         
                                                                                          
Time: 773 ms, Memory: 12.00MB                                                             
                                                                                          
There was 1 failure:                                                                      
                                                                                          
FAILURES!                                                                                 
```

### Guilty as charged

sanmai:
Fixes #306

This tangled mess is because we're trying to disable xdebug when we would need it for sub-processes, and there's no right way to disable it for this process *only*, leaving sub-processes with xdebug enabled. If we use a custom php.ini, we have to include all .ini files in it, but then we have to ignore all those .ini files for us, for which we need to set envvar `PHP_INI_SCAN_DIR=`, which naturally propagates to our subprocesses, which brings us where we are.

A more proper solution to a general xdebug-avoidance problem would be to identify parts where xdebug is actually harmful, and run those parts in a subprocess with disabled xdebug.

Removing all xdebug workarounds right now would make Infection about 20% slower.

### Is it all about xdebug?

Not, not at all. My point is that we *should not* pollute any other process environment with `PHP_INI_SCAN_DIR=` but our own anyway. `PHP_INI_SCAN_DIR=` really breaks things. 

Say, a program we're mutating wants to start another PHP program that's going to read an XML file and do something with what it found. But it would not know that in needs to use a different `php.ini`, and with `PHP_INI_SCAN_DIR=` set that other program would not be able to read XML file like at all, because the extension it needs won't be loaded. We would be thinking: oh great, another mutant erred, but actually it won't be a mutant, but a bug in Infection that caused that error.

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -274,6 +274,11 @@ final class InfectionCommand extends BaseCommand
 
     private function applyMemoryLimitFromPhpUnitProcess(Process $process, PhpUnitAdapter $adapter)
     {
+        if (PHP_SAPI == 'phpdbg') {
+            // Under phpdbg we're using a system php.ini, can't add a memory limit there
+            return;
+        }
+
         $tempConfigPath = \php_ini_loaded_file();
 
         if (empty($tempConfigPath) || !file_exists($tempConfigPath) || !is_writable($tempConfigPath)) {
--- a/src/Php/ConfigBuilder.php
+++ b/src/Php/ConfigBuilder.php
@@ -31,6 +31,11 @@ final class ConfigBuilder
         $this->tempDir = $tempDir;
     }
 
+    public static function hasBuiltTempPhpConfig(): bool
+    {
+        return getenv(self::ENV_TEMP_PHP_CONFIG_PATH) !== false;
+    }
+
     /**
      * @return string|null config path
      *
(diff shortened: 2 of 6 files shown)


src/Command/InfectionCommand.php
```diff
@@ -274,6 +274,11 @@ private function getOutputFormatter(): OutputFormatter
 
     private function applyMemoryLimitFromPhpUnitProcess(Process $process, PhpUnitAdapter $adapter)
     {
+        if (PHP_SAPI == 'phpdbg') {
+            // Under phpdbg we're using a system php.ini, can't add a memory limit there
```
theofidry: maybe we could by still doing a server restart but not disabling xdebug; Might not be easy with Composer XdebugHandler though

sanmai:
> Yes, we need Xdebug for initial run tests process only. Other processes (like running Test Framework for Mutants) don't need it.

Don't get me wrong, but we're having issue #306 precisely because they need it.

src/Command/InfectionCommand.php
sanmai: phpdbg has a default memory limit even on Debian, so this isn't a big concern

sanmai:
Answering my own question, this seems to be a sub-optimal solution. I am going to close this for now, but if no other solution is found, we'll have to get back to this.

sanmai:
@sidz it appears we're still using older version of xdebug-handler

```
$ composer show composer/xdebug-handler | grep ver
versions : * 1.0.0
```

PR #309, opened 2018-04-18
ParallelProcessRunner: use a common procedure for new processes

sanmai:
Fixes a bug where uncovered mutations would actually run if they happen to be in the initial batch.


sanmai:
OK, done.

borNfreee:
Awesome, thank you @sanmai

PR #310, opened 2018-04-19
Disable xdebug for all but the initial test process

### LINKED ISSUE #306: Initial test suite may fail to deliver coverage, e.g. under Docker

sanmai:
This PR disables xdebug for all but the initial test process. 

Some may consider this a breaking change because earlier xdebug may have been used for some test runs in uncertain circumstances. I'm not going to go looking for these circumstances because #308 shown that there is a consensus that xdebug must not be used for anything but the initial test run.

[PHPRC](http://php.net/manual/en/configuration.php) is used to force our php.ini to all subprocesses.

Fixes #306

Supersedes #308

#263 is a definitely a better way to go, only if it could fix #306 and let us disable xdebug for sub-processes we do not control. ~~Which it wont.~~


borNfreee:
We don't have our xdebug handler implementation anymore. Should we close this one @sanmai?

sanmai:
We definitely should once I scavenge the good parts from here.

PR #311, opened 2018-04-23
Memory_Limit reliability fixes

sanmai:
- PHP must be able to load other .ini files
- Memory limit cannot be enforced from our custom php.ini under PHPDBG


borNfreee:
Thank you @sanmai

PR #312, opened 2018-04-23
Build PHAR with Humbug Box

theofidry:
Note that the PHAR is not isolated _yet_. This IMO should be part of another PR.

Meanwhile I would suggest we automate the deploy of the secured PHAR (can be done in another PR).


Makefile
```diff
@@ -125,3 +124,10 @@ phpstan: vendor $(PHPSTAN)
 
 validate:
 	composer validate --strict
+
+bin/infection.phar: app bin src vendor box.json.dist scoper.inc.php box.phar
```
borNfreee: There was a `build` folder created for generated PHAR since https://github.com/infection/infection/pull/67

IMO it would be consistent to have all non-commited files in one folder to not pollute the source code, what do you think?

theofidry: I'm fine with that

BackEndTea:
>Meanwhile I would suggest we automate the deploy of the secured PHAR (can be done in another PR).

I like this idea, however we should have a 'fall-back' in case travis has a failure or something, so we aren't left with releases without phars.

theofidry:
I think a release without PHAR is better. You can always fix it and do an empty patch release, for the people using the PHAR with the `self-upate` it's not gonna make a difference

theofidry:
I've updated it. I'll push for the isolated PHAR in a couple of weeks, but it would be nice if someone could add an e2e test for the PHAR meanwhile. I need to fix a couple of things in PHP-Scoper first

borNfreee:
Thank you @theofidry

Issue #295, opened 2018-04-07
Infection does not fail gracefully if PATH isn't set


sanmai:
| Question    | Answer
| ------------| ---------------
| Infection version | master

With no PATH set Infection fails [in AbstractExecutableFinder](https://github.com/infection/infection/blob/master/src/Finder/AbstractExecutableFinder.php#L22) with a vague TypeError:

```
$ PATH= /usr/bin/php ./bin/infection
 
PHP Fatal error:  Uncaught TypeError: explode() expects parameter 2 to be string, boolean given in src/Finder/AbstractExecutableFinder.php:25
Stack trace:
#0 src/Finder/AbstractExecutableFinder.php(25): explode(':', false)
#1 src/Finder/ComposerExecutableFinder.php(38): Infection\Finder\AbstractExecutableFinder->searchNonExecutables(Array, Array)
#2 src/Finder/TestFrameworkFinder.php(90): Infection\Finder\ComposerExecutableFinder->find()
#3 src/Finder/TestFrameworkFinder.php(73): Infection\Finder\TestFrameworkFinder->findComposer()
#4 src/Finder/TestFrameworkFinder.php(46): Infection\Finder\TestFrameworkFinder->addVendorFolderToPath()
#5 src/TestFramework/AbstractTestFrameworkAdapter.php(101): Infection\Finder\TestFrameworkFinder->find()
#6 src/Process/Builder/ProcessBuilder. in src/Finder/AbstractExecutableFinder.php on line 25
```
`PATH` (or `Path`) environment variable is crucial to Infection, but its absence isn't handled. Alternatively Infection can be made to continue as usual, if possible without PATH.

(Should we use `$_ENV['PATH']` instead of `getenv('PATH')`?)

sidz:
> (Should we use $_ENV['PATH'] instead of getenv('PATH')?)

Don't think that this is a good idea because of this explanation - https://stackoverflow.com/a/8798318

and also we can disable every super global by this option http://php.net/manual/ru/reserved.variables.environment.php#98113  . I'd prefer to leave `getenv`

theofidry:
`$_ENV['PATH']` is safer to use though: unlike `getenv()` it's thread safe

PR #313, opened 2018-04-23
If PATH is not set, continue infection execution with fallback

borNfreee:
If PATH/Path is not set, fall back to `vendor/bin` in `ComposerExecutableFinder`
and continue Infection execution instead of displaying a weird error

Fixes https://github.com/infection/infection/issues/295

### DIFF
--- a/src/Finder/AbstractExecutableFinder.php
+++ b/src/Finder/AbstractExecutableFinder.php
@@ -21,17 +21,23 @@ abstract class AbstractExecutableFinder
      */
     protected function searchNonExecutables(array $probableNames, array $extraDirectories = [])
     {
+        $path = getenv('PATH') ?: getenv('Path');
+
+        if (!$path) {
+            return null;
+        }
+
         $dirs = array_merge(
-            explode(PATH_SEPARATOR, getenv('PATH') ?: getenv('Path')),
+            explode(PATH_SEPARATOR, $path),
             $extraDirectories
         );
 
         foreach ($dirs as $dir) {
             foreach ($probableNames as $name) {
-                $path = sprintf('%s/%s', $dir, $name);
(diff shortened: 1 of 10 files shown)


BackEndTea:
Is there any way to mimic this behaviour (not having a path set). Because if so it might be nice to add a test for it.

sanmai:
The only objection is hanging empty `Empty_Path/README.md`.

borNfreee:
Unfortunately, on Travis, setting `PATH=` [breaks the build](https://travis-ci.org/infection/infection/jobs/370291735#L785). After setting empty value to `PATH`, Travis does not know any command (bash, which, php, etc.).

Tested on Travis VM (connected by ssh to the build):

```bash
cd tests/Fixtures/e2e/Empty_Path

PATH=  run_tests.bash

/usr/bin/env: bash: No such file or directory
```

then 

```bash
PATH= /bin/bash run_tests.bash

which: No such file or directory
/usr/bin/env: php: No such file or directory
```

I have no idea how to resolve this.

borNfreee:
> what about https://github.com/infection/infection/blob/master/src/Finder/TestFrameworkFinder.php#L103 line?

at the moment of line 103, `PATH` will always have the value even if it was empty, because of https://github.com/infection/infection/blob/master/src/Finder/TestFrameworkFinder.php#L84 (`addVendorFolderToPath()`). So I see no way that this line can cause an issue.

sanmai:
What about this:
```bash
phpdbg=$(phpdbg)
PATH= $phpdbg -qrr $INFECTION
```

BackEndTea:
Could we have a bootstrap file which simply does a `putenv` of path.

borNfreee:
@BackEndTea could you please explain it in more details ?

borNfreee:
@BackEndTea it works!

PR #316, opened 2018-04-26
Fix: Add vendor and temporary files to e2e gitignore

BackEndTea:
We do not want to commit these files to the repo, so it makes sense to ignore them. This was not an issue before as they were ignored in the main gitignore. That one was changed to only ignore those files in the root folder.


borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee @sidz

PR #319, opened 2018-04-26
Make the directory the name instead of a parameter

BackEndTea:
By passing the directory as a name we still get it in the error message if it fails, (so instead of data set 0 its called dataset <directory name>).  But we don't have to pass it as a parameter.

### DIFF
--- a/tests/Console/E2ETest.php
+++ b/tests/Console/E2ETest.php
@@ -90,7 +90,7 @@ class E2ETest extends TestCase
      * @dataProvider e2eTestSuiteDataProvider
      * @group e2e
      */
-    public function test_it_runs_an_e2e_test_with_success(string $e2eTestName, string $fullPath)
+    public function test_it_runs_an_e2e_test_with_success(string $fullPath)
     {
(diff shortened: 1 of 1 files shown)


tests/Console/E2ETest.php
```diff
@@ -90,7 +90,7 @@ public function test_it_runs_configure_command_if_no_configuration()
      * @dataProvider e2eTestSuiteDataProvider
      * @group e2e
      */
-    public function test_it_runs_an_e2e_test_with_success(string $e2eTestName, string $fullPath)
+    public function test_it_runs_an_e2e_test_with_success(string $fullPath)
```
borNfreee: `: \Generator`

BackEndTea: This is a test function, not a dataprovider, so no return type needed (except maybe void in the future)

borNfreee: sorry, my mistake

borNfreee:
Thank you @BackEndTea

Issue #318, opened 2018-04-26
Make Crash Course


Fenikkusu:
Someone mind giving me a make crash-course? I've read up on make syntax and can otherwise read the instructions, but am still having troubles. 

My basic understanding at this point is that any `<target>:<files>` line can be called using `make <target>` and any `.PHONY` line is there to say 'always run this regardless of if `<files>` exist or not.

For example, I want to build the phar on my local comp. Looking at the make file, I thought I would just run `make build/bin/infection.phar` and be good to go. However, running that gives me the error `make: *** No rule to make target `build/bin/infection.phar'.  Stop.`

BackEndTea:
`make build/bin/infection.phar` works fine for me. Are you sure you are up to date with master?

`.PHONY` means that its not a real file, and should not be checked, so its command should always be executed. 

> My basic understanding at this point is that any `<target>:<files>` line can be called using `make <target>` and any `.PHONY` line is there to say 'always run this regardless of if `<files>` exist or not.

That is correct, `<target>:<files>` is `<target>:<dependencies>` 
So to run the `target` it first checks all `dependencies`. If it is a PHONY target it will always run its command, if its a file it only runs its command if needed.

Fenikkusu:
> make build/bin/infection.phar works fine for me. Are you sure you are up to date with master?

Looks like I was on the wrong project actually xX my bad. Question though, how do I generate the pem file, and should this be added to the make file.

Fenikkusu:
Got It, Closing

theofidry:
>Question though, how do I generate the pem file, and should this be added to the make file.

You should not :/

I don't like this because it's making it harder for contributors. I worked around that in PHP-Scoper by creating a dev `box.json` via the `Makefile` to remove the signing algorithm, but I think it should be integrated directory in Box rather

PR #321, opened 2018-04-27
Make Tweaks

Fenikkusu:
This PR:

- [x] Adds new feature ...
- [ ] Covered by tests

Fixes https://github.com/infection/infection/issues/318

I made a few tweaks to the make process, just to make it a little easier to work with phars locally.

1. Added a target for the .travis/infection-private.pem - I presume travis will add this automatically during a normal run, if it is so needed.
1. Added a `link` target - just a quick helper to link to /usr/local/bin/infection
1. Added a `clean` target - removes a previously created phar and link.
1. Updated .gitignore - If you run box.phar multiple times (IE: -B make flag), you phar file will become box.phar.1, box.phar.2, etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,10 +133,11 @@ box.phar:
 	chmod a+x box.phar
 
 .travis/infection-private.pem:
-	openssl req -newkey rsa:2048 -new -nodes -x509 -days 3650 -keyout .travis/infection-private.pem -subj "/C=GB/ST=London/L=London/O=Global Security/OU=IT Department/CN=example.com"
+	openssl req -newkey rsa:2048 -new -nodes -x509 -days 3650 -keyout .travis/infection-private.pem -out /dev/null -subj "/C=GB/ST=London/L=London/O=Global Security/OU=IT Department/CN=example.com"
 
 .PHONY: clean
 clean:
+	rm -rf box.phar*
 	rm -rf build/bin/infection.phar
 	rm -rf /usr/local/bin/infection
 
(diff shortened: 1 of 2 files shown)


theofidry:
As per https://github.com/infection/infection/issues/318#issuecomment-385000490, I don't think we should require the signing algorithm for the dev used only for dev purposes. But for the real PHAR (which can be build and used in the CI and then deployed automatically), I think it would be better if @borNfreee would do it as described in [create and deploy secure PHARs](https://medium.com/@tfidry/create-and-deploy-secure-phars-c5572f10b4dd). I think however it should be a GPG key for the Infection project (there's a link about that in the article as well)

Makefile
```diff
@@ -125,9 +125,22 @@ phpstan: vendor $(PHPSTAN)
...
+	openssl req -newkey rsa:2048 -new -nodes -x509 -days 3650 -keyout .travis/infection-private.pem -out /dev/null -subj "/C=GB/ST=London/L=London/O=Global Security/OU=IT Department/CN=example.com"
+
+.PHONY: clean
+clean:
+	rm -rf box.phar*
+	rm -rf build/bin/infection.phar
+	rm -rf /usr/local/bin/infection
+
+.PHONY: link
+link: build/bin/infection.phar
```
theofidry: Hm you're sure about that?

Fenikkusu:
I do have to agree that signing for local development may not be necessary.

Makefile
Fenikkusu: Sure about the linking? I just see it as a quick helper for development. Probably wouldn't do anything for windows users.

borNfreee:
this is not relevant anymore, use `make build/bin/infection.phar` to build not signed local PHAR for development/test

Issue #126, opened 2018-01-16
Feature: Include bootstrap.php entirely


patrickfunke:
For some projects, PHPUnit's bootstrap.php may contain a very customized setup (e.g. additional autoloading or global test data). Currently, Infection ignores everything but the autoloader generated by composer.

To enable the framework for such complex applications, I'd propose the possibility of including the bootstrap.php entirely as-is to ensure test execution works as intended.

I already tried this out and it seems the further analysis works just fine with such a complete custom bootrap file.

theofidry:
Shouldn't the autoloading of that bootstrap be delegated to PHPUnit?

patrickfunke:
The correct bootstrap.php is defined in phpunit.xml, but it seems to me that Infection builds its own bootstrap.php wich is then referenced in the configuration file created for each mutant. When looking up how Infection builds this bootstrap.php for the mutants, most of the content of the original one is not included.

Issue #322, opened 2018-04-27
all mutants escape, even if some should be caught


yivi:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.8.1 
| Test Framework version | 7.1.4
| PHP version | 7.2.4
| Platform    | MacOS
| Github Repo | - https://github.com/yivi/demu_mutant_fail

Created a very basic proof-of-concept project to play around with mutation testing.
There is a single class (`FooZero`), with a single method:

```php
public function zero(int $a): int

    {
        return $a * 0;
    }
```

And a single test covering it:
```php
 public function testGiveMeZero()
    {
        $a = new FooZero();
        $this->assertSame($a->zero(10), 0);
    }
```

The test passes normally: 

> vendor/bin/phpunit --testdox
> PHPUnit 7.1.4 by Sebastian Bergmann and contributors.
>
>Runtime:       PHP 7.2.4 with Xdebug 2.6.0
>Configuration: /Users/ivan.yivoff-ext/www_personal/demo_mutant_fail/phpunit.xml.dist
>
>FooMin
> ✔ Give me zero [7.49 ms]
>
> Time: 68 ms, Memory: 4.00MB

But when I run infection, it tells me that all mutants have escaped... 

```
vendor/bin/infection --test-framework-options="--bootstrap=bootstrap.php"
You are running Infection with xdebug enabled.
    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/

Running initial test suite...

PHPUnit version: 7.1.4

    7 [============================] < 1 sec

Generate mutants...

Processing source code files: 1/1
Creating mutated files and processes: 4/4
.: killed, M: escaped, S: uncovered, E: fatal error, T: timed out

MMMM                                                 (4 / 4)

4 mutations were generated:
       0 mutants were killed
       0 mutants were not covered by tests
       4 covered mutants were not detected
       0 errors were encountered
       0 time outs were encountered
```
But I think that all four mutations should fail the test:

**PublicVisibility** should naturally fail.
**Multiplication** changes the zero multiplication for a zero division... which should logically fail.
**OneZeroInteger** would make the test fail, as a the method would return the number passed, and not zero. **DecrementInteger**  should fail similarly.

[...]

Issue #320, opened 2018-04-26
infection ignores phpunit.xml bootstrap file


yivi:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.8.1 (`infection.phar --version`)
| Test Framework version | PHPUnit 
| PHP version | 7.2.4
| Platform    |  MacOS
| Github Repo | -

I'm able to run my tests normally by just invoking phpunit and reading my phpunit.xml configuration file.

But when I run infection, all my tests fail by failing to instantiate my tested class:
> 1) DummyClassTest::testOne
> Error: Class 'DummyClass' not found

It's pretty clear that my _bootstrap.php_ file is not being read and executed (it consists only of a couple `requires`, one for `vendor/autoload.php` and the other to the file where `DummyClass` has been defined.

Is there a way to force infection to run the bootstrap defined in phpunit.xml?

<details>
 <summary>phpunit.xml</summary>
 
 ```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/6.0/phpunit.xsd"
         bootstrap="./bootstrap.php"
         verbose="true"
         colors="true"
         backupGlobals="false"
>
    <testsuite name="dummy">
        <directory suffix="Test.php">tests</directory>
    </testsuite>
</phpunit>
 ```
</details>

<details>
 <summary>Output with issue</summary>
 
 ```
1) /Users/user/www_personal/anagram_testing/src/DummyClass.php:11    [M] DecrementInteger
exec /usr/local/Cellar/php/7.2.4_1/bin/php -c /private/var/folders/vm/4njclq0551gfx1q36d_syk7sfhkp5k/T/infectionjdFypN  /Users/user/www_personal/anagram_testing/vendor/phpunit/phpunit/phpunit --configuration /var/folders/vm/4njclq0551gfx1q36d_syk7sfhkp5k/T/infection/phpunitConfiguration.13e9de1857302da54b54d93cf82c5602.infection.xml --stop-on-failure 
--- Original
+++ New
@@ @@
         // if ( $empireValue <= 500 || true === $isBegginer ) {
-        if ($aburdValue < 500 || true === $trivialToggle) {
+        if ($aburdValue < 499 || true === $trivialToggle) {
             return $trivialIndex / 2;
         }
         return $trivialIndex * 1.5;
     }
 }

[...]

PR #323, opened 2018-04-27
Use project's bootstrap file in PHPUnit sub-processes rather than hardcoded `vendor/autoload.php`

borNfreee:
This PR

- [x] Loads `bootstrap` file from project's `phpunit.xml` config for each Mutant proces
- [x] Covered by tests
- [x] Fixes https://github.com/infection/infection/issues/126 https://github.com/infection/infection/issues/320 https://github.com/infection/infection/issues/322
- [x] Docs: https://github.com/infection/site/pull/36

Since `ReflectionVisitor` uses `new \ReflectionClass()` code, it must know how to autoload project's classes. That's why we need to require custom project's bootstrap file not only for PHPUnit processes, but for Infection as well.

This is done by using `bootstrap` setting in `infection.json`, see added e2e test.

### DIFF
--- a/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
+++ b/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
@@ -38,6 +38,11 @@ class MutationConfigBuilder extends ConfigBuilder
      */
     private $originalXmlConfigContent;
 
+    /**
+     * @var \DOMDocument
+     */
+    private $dom;
+
     public function __construct(string $tempDirectory, string $originalXmlConfigContent, XmlConfigurationHelper $xmlConfigurationHelper, string $projectDir)
     {
         $this->tempDirectory = $tempDirectory;
@@ -45,14 +50,17 @@ class MutationConfigBuilder extends ConfigBuilder
 
         $this->xmlConfigurationHelper = $xmlConfigurationHelper;
         $this->originalXmlConfigContent = $originalXmlConfigContent;
+
+        $this->dom = new \DOMDocument();
+        $this->dom->preserveWhiteSpace = false;
+        $this->dom->formatOutput = true;
+        $this->dom->loadXML($this->originalXmlConfigContent);
     }
 
     public function build(MutantInterface $mutant): string
     {
-        $dom = new \DOMDocument();
-        $dom->preserveWhiteSpace = false;
-        $dom->formatOutput = true;
-        $dom->loadXML($this->originalXmlConfigContent);
+        // clone the dom document because it's mutated later
+        $dom = clone $this->dom;
 
         $xPath = new \DOMXPath($dom);
 
--- a/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
+++ b/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
@@ -65,10 +65,12 @@ class MutationConfigBuilder extends ConfigBuilder
             $mutant->getMutation()->getHash()
         );
 
-        $this->setCustomAutoLoaderPath($customAutoloadFilePath);
+        $originalAutoloadFile = $this->getOriginalBootstrapFilePath();
+
+        $this->setCustomBootstrapPath($customAutoloadFilePath);
         $this->setFilteredTestsToRun($mutant->getCoverageTests());
 
-        file_put_contents($customAutoloadFilePath, $this->createCustomAutoloadWithInterceptor($mutant));
+        file_put_contents($customAutoloadFilePath, $this->createCustomAutoloadWithInterceptor($mutant, $originalAutoloadFile));
 
         $path = $this->buildPath($mutant);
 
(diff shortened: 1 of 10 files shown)


tests/Fixtures/e2e/Save_PHPUnit_Bootstrap_File/infection.json
```diff
@@ -0,0 +1,12 @@
...
+    "timeout": 25,
+    "source": {
+        "directories": [
+            "src"
+        ]
+    },
+    "logs": {
+        "summary": "infection-log.txt"
+    },
+    "bootstrap": "./custom-autoload.php"
```
borNfreee: should we add this case to documentation? (When project has custom autoloaded, it should be added to `bootstrap` section of infection config)

BackEndTea:
This will also fix #126

tests/Fixtures/e2e/Save_PHPUnit_Bootstrap_File/infection.json
BackEndTea: Isn't the goal of this PR to make sure that we use the custom bootstrap file in phpunit, rather than hard coding it. 

So if we add it here to the custom infection bootstrap file that does not prove that it is correctly included/used.

borNfreee: Sorry, I didn't get what you mean

> Isn't the goal of this PR to make sure that we use the custom bootstrap file in phpunit, rather than hard coding it.

Yes, this is a goal of this PR. Previously we always included `vendor/autoload` in PHPUnit process, now we include that file, that was in the original `phpunit.xml`

> So if we add it here to the custom infection bootstrap file that does not prove that it is correctly included/used.

Here, in `infection.json` we add custom autoload to make Infection work itself (not PHPUnit process that infection runs). Without it, our ReflectionVisitor won't be able to autoload classes that are not autoloaded by composer's autoloader. You can run added e2e test to see what I mean and remove this line

BackEndTea: Ah i understand whats going on now.

In that case, yes, we should add it to the documentation.

src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
```diff
@@ -29,68 +29,70 @@ class MutationConfigBuilder extends ConfigBuilder
...
-        $xmlConfigurationHelper->deactivateColours($this->xPath);
-        $xmlConfigurationHelper->removeExistingLoggers($this->dom, $this->xPath);
-        $xmlConfigurationHelper->removeExistingPrinters($this->dom, $this->xPath);
+        $this->xmlConfigurationHelper = $xmlConfigurationHelper;
+        $this->originalXmlConfigContent = $originalXmlConfigContent;
     }
 
     public function build(MutantInterface $mutant): string
     {
+        $dom = new \DOMDocument();
```
BackEndTea: I have no idea how expensive the creation of these objects are, but maybe we can only do the steps that are different per mutant, every single time. While working from a base dom object

tests/Fixtures/e2e/Save_PHPUnit_Bootstrap_File/infection.json
borNfreee: added documentation https://github.com/infection/site/pull/36

src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
borNfreee: Initially, I moved this object creation from the constructor by purpose, because there was another bug: https://travis-ci.org/infection/infection/jobs/372222312#L761

When this object was created in constructor, the same instance was used in `build()` method, and each time [it was modified](https://github.com/infection/infection/pull/323/files#diff-5780b7fc216dd1e72bb1cdde6558f339R125).

This led to side effect, when each new `phpunit.xml` file has previous one as an `bootstrap` file path (because \DOMDocument is not immutable).

So I decided to move it to `build()` method.

Now, with [this new commit](https://github.com/infection/infection/pull/323/commits/fab5839f74e19bc0c51229e56330b0cd919d1929), it's a little bit faster to clone already instantiated object rather than create it from the scratch. So I moved it back, but instead of the full reusing - it is being cloned.

Perfomance:

| Operation | Time |
|--|--|
|Instantiating `\DOMDocument`| Clone `\DOMDocument` |
| `8.0108642578125E-5ms` | from `9.0599060058594E-6ms` to `2.1934509277344E-5ms` |

So, cloning from 4 to 10 times faster.

BackEndTea: Pretty cool, how did you measure those times? just with `microtime()`, or do you use something else?

borNfreee: Just microtime

borNfreee:
Does anybody have any objections before merging? @infection/core

BackEndTea:
:+1: for me

borNfreee:
Thank you @BackEndTea and @sidz

PR #326, opened 2018-04-28
Error on faulty configuration

### LINKED ISSUE #231: Infection should error when provided with a non existent config file
When a configuration file is provided through the `--configuration` (`-c`) option, and that file is not found, infection tries to fall back on the default config file.

Just like #218 infection should notify the user that the specified config file does not exist and exit.

BackEndTea:
This PR:

- [x] Makes infection error when provided with a non-existent file in `--configuration` 
- [x] Covered by tests

Fixes #231

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -481,18 +481,16 @@ final class InfectionCommand extends BaseCommand
     protected function initialize(InputInterface $input, OutputInterface $output)
     {
         parent::initialize($input, $output);
-
+        $locator = $this->getContainer()->get('locator');
         if ($customConfigPath = $input->getOption('configuration')) {
-            $this->getContainer()->get('locator')->locate($customConfigPath);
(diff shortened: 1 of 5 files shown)


src/Command/InfectionCommand.php
```diff
@@ -482,15 +482,18 @@ protected function initialize(InputInterface $input, OutputInterface $output)
     {
         parent::initialize($input, $output);
 
-        $customConfigPath = $input->getOption('configuration');
-        $configExists = $customConfigPath && file_exists($customConfigPath);
+        if ($customConfigPath = $input->getOption('configuration')) {
+            $this->getContainer()->get('locator')->locate($customConfigPath);
```
borNfreee: does it make sense to extract `$this->getContainer()->get('locator')` to variable? used twice

borNfreee:
Thank you @BackEndTea

PR #327, opened 2018-04-28
Move bootstrap logic into infection file

BackEndTea:
Now that bootstrap.php has become much smaller its more logical to have that logic within the infection entry-point.

We also take advantage of php not un setting variables after a foreach, so we no longer need a define, but instead break out of the loop if we found the file, and simply check if file is set, and error out otherwise.

We now also include the `bin` folder in php-cs-fixer and the infection file as well.


bin/infection
```diff
@@ -1,13 +1,49 @@
...
+
+if (!isset($file)) {
+    fwrite(
+        STDERR,
+        'You need to set up the project dependencies using Composer:' . PHP_EOL . PHP_EOL .
+        '    composer install' . PHP_EOL . PHP_EOL .
+        'You can learn all about Composer on https://getcomposer.org/.' . PHP_EOL
+    );
+
+    die(1);
```
borNfreee: let's use `exit()` since we already use it in `src` and it would be consistent with `exit` in e2e tests

borNfreee:
Thank you @BackEndTea

PR #328, opened 2018-04-29
Cache the results of the getHash function

BackEndTea:
This PR:

- [x] Caches the hash value of a mutation

It gets called 3 times per mutation, so it makes sense to cache its results instead of recreating it

### DIFF
--- a/src/Mutation.php
+++ b/src/Mutation.php
@@ -49,6 +49,11 @@ final class Mutation implements MutationInterface
      */
     private $isCoveredByTest;
 
+    /**
+     * @var string
+     */
+    private $hash;
+
     public function __construct(
         string $originalFilePath,
         array $originalFileAst,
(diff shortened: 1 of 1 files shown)


src/Mutation.php
```diff
@@ -89,21 +94,25 @@ public function getMutatedNodeClass(): string
...
-            $attributes['startTokenPos'],
-            $attributes['endTokenPos'],
-            $attributes['startFilePos'],
-            $attributes['endFilePos'],
-        ];
-
-        $hashKeys = array_merge([$this->getOriginalFilePath(), $mutatorClass], $attributeValues);
-
-        return md5(implode('_', $hashKeys));
+        if (!isset($this->hash)) {
```
localheinz: What do you think about comparing with `null` instead?

BackEndTea: I intended to do that at first. But i found isset being used elsewhere in the code.

borNfreee:
Thank you @BackEndTea

PR #329, opened 2018-04-30
Use a transparent xdebug-free environment for all but the initial test process

### LINKED ISSUE #306: Initial test suite may fail to deliver coverage, e.g. under Docker

### LINKED ISSUE #332: Weird behaviour in Docker container
| Question    | Answer
| Infection version | 0.8.1
| Test Framework version | PHPUnit 7.1.5
| PHP version | 7.1.16
| Platform    | Docker `php:7.1`
| Github Repo | https://github.com/trinet-at/docker-infection

I use GitLab and the gitlab-ci task runner for testing my projects. The runner uses Docker to execute the tasks, having `php:7.1` as base image. When I added infection to my project I noticed a weird behavior when infection runs in the container. Locally everything works fine.
Infection complains that a required extension, in this test case the `intl` extension is not loaded, although it is.

To replicate the problem I created a small test scenario which uses the `Locale` class from `intl` from which the problem can be reproduced.

```
git clone https://github.com/trinet-at/docker-infection.git
cd docker-infection
composer install
docker build -t docker-infection .
```

Locally, infection runs fine:
`vendor/bin/infection`

If I run infection in the Docker container the error occurs (`vendor/bin/infection` is the default `CMD` in the image)

`docker run --rm -it -v $PWD:/app docker-infection`

However, running PHPUnit completes without a problem:
`docker run --rm -it -v $PWD:/app docker-infection vendor/bin/phpunit`

Also, running the `infection.phar` (installed as executable in `/usr/local/bin` in the image) succeeds:
`docker run --rm -it -v $PWD:/app docker-infection infection.phar`

The "solution" is to combine all of the `.ini` files of PHP into one `php.ini`.
There's a script in the repo which does exactly that and runs `vendor/bin/infection` afterwards:
`docker run --rm -it -v $PWD:/app docker-infection ./convert-ini.sh`

So it seems to me that infection somehow runs PHPunit in a non-standard way.

Does anybody have a solution for that without having to use the `.phar` file or combining PHP's `.ini` files? Maybe it's somehow related to issue #306.

Thanks!

sanmai:
Fixes #306
Fixes #332

(`make test` still fails because of missing `expect`, #304 should fix this problem too)

### DIFF
--- a/src/Console/Util/PhpProcess.php
+++ b/src/Console/Util/PhpProcess.php
@@ -51,6 +51,15 @@ final class PhpProcess extends Process
             return;
         }
 
+        /*
+         * Vanilla processes are expected to run in a vanilla, or the most original, environment.
+         *
+         * For that, we need to remove or reset all environment variables we set to setup our custom
+         * xdebug-free environment, later setting them back so we won't have to think again about it
+         * for the bulk of other processes, which not require xdebug to function, and work better
+         * without it.
+         */
+
         self::restoreVanillaEnvironment();
 
         parent::start($callback, $env);
--- a/src/Console/Util/PhpProcess.php
+++ b/src/Console/Util/PhpProcess.php
@@ -32,7 +32,7 @@ final class PhpProcess extends Process
      * - PHP_INI_SCAN_DIR should be made blank because our php.ini has all we need
      *   (a previous value can be found in XdebugHandler::getRestartSettings()["scanDir"])
      */
-    public static function setupEnvironment()
+    public static function setupXdebugFreeEnvironment()
     {
         if (!isset(self::$phprc)) {
             self::$phprc = getenv('PHPRC');
@@ -55,7 +55,7 @@ final class PhpProcess extends Process
 
         parent::start($callback, $env);
 
-        self::setupEnvironment();
+        self::setupXdebugFreeEnvironment();
     }
 
     private static function restoreVanillaEnvironment()
--- a/src/Console/Util/PhpProcess.php
+++ b/src/Console/Util/PhpProcess.php
@@ -46,7 +46,7 @@ final class PhpProcess extends Process
     {
         // Xdebug wasn't skipped, running as is
         if ('' == XdebugHandler::getSkippedVersion()) {
-            parent::start($callback, $env);
+            parent::start($callback, $env ?? []);
 
             return;
         }
@@ -62,7 +62,7 @@ final class PhpProcess extends Process
 
         self::restoreVanillaEnvironment();
 
-        parent::start($callback, $env);
+        parent::start($callback, $env ?? []);
 
         self::setupXdebugFreeEnvironment();
     }
--- a/src/Console/Util/PhpProcess.php
+++ b/src/Console/Util/PhpProcess.php
@@ -1,0 +1,77 @@
+<?php
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Console\Util;
+
+use Composer\XdebugHandler\XdebugHandler;
+use Symfony\Component\Process\Process;
+
+/**
+ * Vanilla PHP process and utility functions
+ *
+ * @internal
+ */
+final class PhpProcess extends Process
+{
+    private static $phprc;
+
+    /**
+     * Setups a default Xdebug-free environment for all subprocesses:
+     *
+     * - PHPRC should point to our temporary php.ini (we need to save a previous value)
+     *
+     * - PHP_INI_SCAN_DIR should be made blank because our php.ini has all we need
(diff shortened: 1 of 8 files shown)


src/Console/Util/PhpProcess.php
```diff
@@ -0,0 +1,77 @@
...
+        self::setupEnvironment();
+    }
+
+    private static function restoreVanillaEnvironment()
+    {
+        self::putenv('PHPRC', self::$phprc);
+        self::putenv('PHP_INI_SCAN_DIR', XdebugHandler::getRestartSettings()['scanDir']);
+    }
+
+    private static function putenv($name, $value)
```
BackEndTea: `$value` can be a string or a bool, but `$name` should always be a string.
Could we change this to `private static function putenv(string $name, $value)`, and add a doc block to help IDE's understand that value can be either

src/Console/Util/PhpProcess.php
```diff
@@ -0,0 +1,77 @@
...
+use Symfony\Component\Process\Process;
+
+/**
+ * Vanilla PHP process and utility functions
+ *
+ * @internal
+ */
+final class PhpProcess extends Process
+{
+    private static $phprc;
```
BackEndTea: Can we add the following doc block here?
```php
/**
 * @var string|bool
 */
```

src/Process/Builder/ProcessBuilder.php
```diff
@@ -49,24 +50,25 @@ public function getProcessForInitialTestRun(
...
                 $this->testFrameworkAdapter->buildInitialConfigFile(),
                 $testFrameworkExtraOptions,
                 $includeArgs,
                 $phpExtraOptions
             ),
             null,
-            $includeArgs ? array_replace($_ENV, $_SERVER) : null,
+            null,
             null,
             null
```
BackEndTea: Instead off `null, null, ...` what about skiping those and doing a `$process->setTimeout(null)` afterwards?

I'm not sure which one is cleaner, what do you think?

sanmai: Yes, that's definitely better than `null, null, null, null`

BackEndTea:
This looks ready to merge for me.

src/Console/Util/PhpProcess.php
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Setups a default Xdebug-free environment for all subprocesses:
+     *
+     * - PHPRC should point to our temporary php.ini (we need to save a previous value)
+     *
+     * - PHP_INI_SCAN_DIR should be made blank because our php.ini has all we need
+     *   (a previous value can be found in XdebugHandler::getRestartSettings()["scanDir"])
+     */
+    public static function setupEnvironment()
```
borNfreee: I suggest to rename it to smth like `setupXdebugFreeSubprocessesEnvironment()` or `setupXdebugFreeEnvironmentForSubprocesses()` or at least `setupXdebugFreeEnvironment()`.

It confused me each time I saw this method call because of the too general name.

What do you think?

src/Console/Util/PhpProcess.php
```diff
@@ -0,0 +1,84 @@
...
+    public function start(callable $callback = null, array $env = null)
+    {
+        // Xdebug wasn't skipped, running as is
+        if ('' == XdebugHandler::getSkippedVersion()) {
+            parent::start($callback, $env);
+
+            return;
+        }
+
+        self::restoreVanillaEnvironment();
```
borNfreee: 1. Am I right that we need to restore original env here because it was set up by calling `->setupEnvironment()` on `Application.php`?

2. what do you think about `restoreOriginalEnvironment()`? feel free to ignore

3. would you mind adding some comments here about why we restore and then setup xdebug-free env again? I would highly simplify understanding of the whole feature

sanmai: 1. That's correct, only we have no choice of *not* calling `setupEnvironment()` in `Application.php`.

2. "Vanilla" is very concrete, "original" is not as much. Where vanilla means the very-very original environment, just an original environment could be some other environment.

3. I'll see to it.

src/Console/Util/PhpProcess.php
sanmai: That's good to go.

sanmai:
@borNfreee I've got this rebased over master, Travis repors all green after some fixing

borNfreee:
Thank you @sanmai

Issue #325, opened 2018-04-28
Xdebug/ phpdbg check should be later in the infection process


BackEndTea:
| Question    | Answer
| ------------| ---------------
| Infection version|  dev/master
| Test Framework version | n/a
| PHP version | 7.1
| Platform    | Fedora
| Github Repo | n/a

Running `bin/infection --help`, i expect to see the help overview, instead i see this

```
 [ERROR] Neither phpdbg or xdebug has been found. One of those is required by Infection in order to generate coverage   
         data. Either:                                                                                                  
                                                                                                                        
         - Enable xdebug and run infection again                                                                        
         - Use phpdbg: phpdbg -qrr infection                                                                            
         - Use --coverage option with path to the existing coverage report                                              
         - Use --initial-tests-php-options option with `-d zend_extension=xdebug.so` and/or any extra php parameters  
```

The same goes for running `bin/infection configure`. I expect to get into the configure modues, but instead i'm greeted by this error message.

Current work-around is to add the `--coverage=xyz` flag, so i get through the initial xdebug check.

sidz:
>.<

sidz:
Just enable xdebug :rofl: 

I'll take care about it

BackEndTea:
> Just enable xdebug

True, but if i'd normally run things with phpdbg i'd expect to be able to run `--help` etc without it.

sidz:
@BackEndTea it was a joke

sanmai:
Isn't that just got fixed by #330?

borNfreee:
yes, thanks @sanmai

PR #330, opened 2018-05-01
Xdebug/phpdbg check should be later in the infection process

sidz:
This PR:

Allow to run --help, --version, list commands without debugger

Issue: https://github.com/infection/infection/issues/325

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -175,7 +175,15 @@ final class InfectionCommand extends BaseCommand
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        if (!$this->isDebuggerOrAdditionalParametersExist()) {
+        if (!$this->hasDebuggerOrCoverageOption()) {
+            $this->io->error([
+                'Neither phpdbg or xdebug has been found. One of those is required by Infection in order to generate coverage data. Either:',
+                '- Enable xdebug and run infection again' . PHP_EOL .
+                '- Use phpdbg: phpdbg -qrr infection' . PHP_EOL .
+                '- Use --coverage option with path to the existing coverage report' . PHP_EOL .
+                '- Use --initial-tests-php-options option with `-d zend_extension=xdebug.so` and/or any extra php parameters',
+            ]);
+
             return 1;
         }
 
@@ -417,22 +425,14 @@ final class InfectionCommand extends BaseCommand
         $this->eventDispatcher = $this->getContainer()->get('dispatcher');
     }
 
-    public function isDebuggerOrAdditionalParametersExist(): bool
+    public function hasDebuggerOrCoverageOption(): bool
     {
         if (!$this->skipCoverage
             && PHP_SAPI !== 'phpdbg'
             && !\extension_loaded('xdebug')
             && !XdebugHandler::getSkippedVersion()
-            && !$this->isInitialTestPhpOptionHasXdebug()
+            && !$this->isXdebugIncludedInInitialTestPhpOptions()
         ) {
-            $this->io->error([
-                'Neither phpdbg or xdebug has been found. One of those is required by Infection in order to generate coverage data. Either:',
-                '- Enable xdebug and run infection again' . PHP_EOL .
-                '- Use phpdbg: phpdbg -qrr infection' . PHP_EOL .
-                '- Use --coverage option with path to the existing coverage report' . PHP_EOL .
-                '- Use --initial-tests-php-options option with `-d zend_extension=xdebug.so` and/or any extra php parameters',
-            ]);
-
             return false;
         }
 
@@ -460,7 +460,7 @@ final class InfectionCommand extends BaseCommand
         }
     }
 
-    private function isInitialTestPhpOptionHasXdebug(): bool
+    private function isXdebugIncludedInInitialTestPhpOptions(): bool
     {
         return (bool) preg_match(
             '/(zend_extension\s*=.*xdebug.*)/mi',
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -425,18 +425,13 @@ final class InfectionCommand extends BaseCommand
         $this->eventDispatcher = $this->getContainer()->get('dispatcher');
     }
 
-    public function hasDebuggerOrCoverageOption(): bool
+    private function hasDebuggerOrCoverageOption(): bool
     {
-        if (!$this->skipCoverage
-            && PHP_SAPI !== 'phpdbg'
-            && !\extension_loaded('xdebug')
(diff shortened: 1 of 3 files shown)


src/Command/InfectionCommand.php
```diff
@@ -52,6 +52,7 @@
 final class InfectionCommand extends BaseCommand
 {
     const CI_FLAG_ERROR = 'The minimum required %s percentage should be %s%%, but actual is %s%%. Improve your tests!';
+    const COMMAND_NAME = 'run';
```
BackEndTea: If we are gonna do this here, should we do it for all our commands?

src/Console/Application.php
```diff
@@ -124,13 +109,32 @@ private function isInitialTestPhpOptionHasXdebug(InputInterface $input): bool
         );
     }
 
-    public function doRun(InputInterface $input, OutputInterface $output)
+    protected function doRunCommand(Command $command, InputInterface $input, OutputInterface $output)
     {
+        if ($command->getName() === InfectionCommand::COMMAND_NAME) {
```
sanmai: Clearly this is what an InfectionCommand *should* do. It it their responsibility to check that we're running with xdebug or not. Not Application's.

src/Console/Application.php
```diff
@@ -96,23 +98,6 @@ public function run(InputInterface $input = null, OutputInterface $output = null
         $xdebug = new XdebugHandler(self::INFECTION_PREFIX, '--ansi');
         $xdebug->check();
 
-        if (PHP_SAPI !== 'phpdbg'
-            && !$this->isXdebugLoaded
-            && !XdebugHandler::getSkippedVersion()
-            && !$input->hasParameterOption('--coverage', true)
-            && !$this->isInitialTestPhpOptionHasXdebug($input)
-        ) {
-            $this->io->error([
```
sanmai: We can set a flag in the containter  (here or in doRun)  intead of outputting an error here, and InfectionCommand could check it and err. This error text is also only important for the InfectionCommand, and only. Outputting this error is beyond resposibilities of the Application.

src/Command/InfectionCommand.php
sanmai: There's `$command->getName()`

sidz: because I don't want to compare it with a string?

src/Console/Application.php
sidz: make sense

BackEndTea: If the check comes later in the process then i think we can remove the `COMMAND_NAME` again, since we don't need to check it from the outside.

src/Command/InfectionCommand.php
```diff
@@ -481,38 +487,68 @@ private function getTestFrameworkExtraOptions(string $testFrameworkKey): TestFra
...
+            $this->runConfigurationCommand($locator);
+        }
 
-            return;
+        $this->io = $this->getApplication()->getIO();
+        $this->skipCoverage = \strlen(trim($input->getOption('coverage'))) > 0;
+        $this->eventDispatcher = $this->getContainer()->get('dispatcher');
+    }
+
+    public function isDebuggerOrAdditionalParametersExist(): bool
```
sidz: not sure about name of this function :)

src/Command/InfectionCommand.php
```diff
@@ -481,38 +487,68 @@ private function getTestFrameworkExtraOptions(string $testFrameworkKey): TestFra
...
 
             if ($result !== 0) {
                 throw InfectionException::configurationAborted();
             }
         }
-        $this->doInitialize($input);
     }
 
-    private function doInitialize(InputInterface $input)
+    private function isInitialTestPhpOptionHasXdebug(): bool
```
sanmai: it should be either *is* or *has* 

e.g. `initialTestPhpOptionsHaveXdebug()` or `isXdebugIncludedInInitialTestPhpOptions()` 

(I'd side with the first for English-readability, but the latter is fine too)

src/Console/Application.php
sanmai: Yeah, it is gone as I see

src/Command/InfectionCommand.php
sanmai: What about simply `hasDebugger`

src/Command/InfectionCommand.php
```diff
@@ -174,6 +176,10 @@ protected function configure()
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        if (!$this->isDebuggerOrAdditionalParametersExist()) {
+            return 1;
```
sanmai: Should not we output the error message here directly, but not as a side effect of calling `isDebuggerOrAdditionalParametersExist`?

src/Command/InfectionCommand.php
sidz: nope. because this method is not only about debugger

src/Command/InfectionCommand.php
```diff
@@ -481,38 +495,60 @@ private function getTestFrameworkExtraOptions(string $testFrameworkKey): TestFra
...
 
-            return;
+        $this->io = $this->getApplication()->getIO();
+        $this->skipCoverage = \strlen(trim($input->getOption('coverage'))) > 0;
+        $this->eventDispatcher = $this->getContainer()->get('dispatcher');
+    }
+
+    public function hasDebuggerOrCoverageOption(): bool
+    {
+        if (!$this->skipCoverage
```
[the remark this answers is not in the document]
borNfreee: or

```php
return $this->skipCoverage
    || PHP_SAPI === 'phpdbg'
    || \extension_loaded('xdebug')
    || XdebugHandler::getSkippedVersion()
    || $this->isXdebugIncludedInInitialTestPhpOptions();
```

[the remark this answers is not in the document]
sidz: @BackEndTea please don't do like you mentioned :) never! I believe that phpstorm gave you that idea :)

will do  like @borNfreee suggested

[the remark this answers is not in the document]
BackEndTea: After i wrote it, it did feel quite dirty, so yeah, the version by @borNfreee should be used :sweat_smile:

borNfreee:
could you please rebase @sidz ?

src/Command/InfectionCommand.php
```diff
@@ -174,6 +176,18 @@ protected function configure()
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        if (!$this->hasDebuggerOrCoverageOption()) {
+            $this->io->error([
```
BackEndTea: We can move the logging to the ConsoleOutput class now

borNfreee:
Thank you @sidz !

PR #331, opened 2018-05-01
Replace mock creation function from getMockBuilder by createMock

sidz:
Very small "improvement" which replaces `getMockBuilder` by `createMock`. It a bit reduces lines of code


borNfreee:
Thank you @sidz

PR #334, opened 2018-05-04
Enforce declare strict_types everywhere

BackEndTea:
This PR:

- [x] Enables the `declare_strict_types` php-cs-fixer rule, and fixes the errors caused by enabling it
- [x] Covered by tests

Follows #277


borNfreee:
Thank you @BackEndTea

PR #335, opened 2018-05-04
Apply max level phpstan to src folder

BackEndTea:
This PR:

- [x] Sets the phpstan level for the `src/` folder to max, and fixes all issues found by phpstan.
- [x] Covered by tests

Most fixes are changing docblocks to the Interface it is hinted against, rather than the class.

We now also throw an error when the phpexectuable finder returns false, instead of failing on a harder to debug TypeError.

### DIFF
--- a/devTools/phpstan-src.neon
+++ b/devTools/phpstan-src.neon
@@ -1,8 +1,7 @@
 parameters:
-    excludes_analyse:
-        - %currentWorkingDirectory%/tests/Fixtures/*
-
     ignoreErrors:
         - '#Access to an undefined property PhpParser\\Node(.*?)#'
         - '#Call to an undefined method PhpParser\\Node(.*?)#'
         - '#Cannot access property \$name on string.#'
+        - '#Accessing property \$expr on possibly null value of type PhpParser\\Node(.*?)#'
+        - '#expects PhpParser\\Node\\Identifier, string given.#'
--- a/src/Mutator/Util/MutatorsGenerator.php
+++ b/src/Mutator/Util/MutatorsGenerator.php
@@ -88,7 +88,7 @@ final class MutatorsGenerator
 
     /**
      * @param string $mutator
-     * @param array|bool $setting
+     * @param array|bool|\stdClass $setting
      */
     private function registerFromClass(string $mutator, $setting)
     {
--- a/tests/Visitor/ReflectionVisitorTest.php
+++ b/tests/Visitor/ReflectionVisitorTest.php
@@ -138,7 +138,7 @@ class ReflectionVisitorTest extends TestCase
         ];
     }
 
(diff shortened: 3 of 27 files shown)


devTools/phpstan-src.neon
```diff
@@ -1,8 +1,7 @@
 parameters:
-    excludes_analyse:
-        - %currentWorkingDirectory%/tests/Fixtures/*
-
     ignoreErrors:
         - '#Access to an undefined property PhpParser\\Node(.*?)#'
         - '#Call to an undefined method PhpParser\\Node(.*?)#'
         - '#Cannot access property \$name on string.#'
+        - '#Accessing property \$expr on possibly null value of type PhpParser\\Node(.*?)#'
+        - '#expects PhpParser\\Node\\Identifier, string given.#'
```
BackEndTea: This error is most likely caused due to phpstan still using v3 of the parser, which causes it to compare it against the class they use instead of our version. (probably)

src/Mutator/Util/MutatorsGenerator.php
```diff
@@ -99,7 +99,7 @@ private function registerFromClass(string $mutator, $setting)
                 $this->mutatorList[$mutator] = [];
             }
         } elseif (!array_key_exists($mutator, $this->mutatorList) || !$this->mutatorList[$mutator]) {
-            $this->mutatorList[$mutator] = (array) $setting;
+            $this->mutatorList[$mutator] = $setting;
```
borNfreee: just curious, what was the issue reported by PHPStan here?

I removed this `(array)` cast and then reverted during `@cast` mutators developing, but don't remember why

BackEndTea: It reported a useless to array cast, but it looks like i messed up some other things, so i'll go ahead and fix that first.

BackEndTea: As it turns out, this wasn't a useless cast, because the input can be a `\stdClass` as well.

tests/Visitor/ReflectionVisitorTest.php
```diff
@@ -138,7 +138,7 @@ public function isPartOfSignatureFlagProvider(): array
         ];
     }
 
-    private function getSpyVisitor(string $nodeClass): NodeVisitorAbstract
```
borNfreee: could you please explain why it was removed?

BackEndTea: We are type-hinting against `NodeVisitorAbstract`, but then calling the `isPartOfSignature` function, which is not part of the `NodeVisitorAbstract` api. 

For me by removing the return type here, PhpStorm is also capable of detecting the `isPartOfSignature` function, which it previously wasn't.

borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee

PR #337, opened 2018-05-06
Use our own ConsoleOutput class over SymfonyStyle

BackEndTea:
This PR:

Extracts console output logic from the `InfectionCommand` class into its own class.

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace Infection\Command;
 
 use Infection\Config\InfectionConfig;
+use Infection\Console\ConsoleOutput;
 use Infection\Console\Exception\InfectionException;
 use Infection\Console\Exception\InvalidOptionException;
 use Infection\Console\LogVerbosity;
@@ -18,7 +19,6 @@ use Infection\Console\OutputFormatter\OutputFormatter;
 use Infection\Console\OutputFormatter\ProgressFormatter;
 use Infection\EventDispatcher\EventDispatcher;
 use Infection\Finder\Exception\LocatorException;
-use Infection\Mutant\Exception\MsiCalculationException;
 use Infection\Mutant\Generator\MutationsGenerator;
 use Infection\Mutant\MetricsCalculator;
 use Infection\Process\Builder\ProcessBuilder;
@@ -43,7 +43,6 @@ use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Process\Process;
 
 /**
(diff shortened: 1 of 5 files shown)


src/Console/ConsoleOutput.php
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        throw MsiCalculationException::create('min-msi');
+    }
+
+    public function logBadCoveredMsiErrorMessage(MetricsCalculator $metricsCalculator, float $minCoveredMsi)
+    {
+        if ($minCoveredMsi) {
+            $this->io->error(
+                    sprintf(
```
borNfreee: incorrect indentation

src/Console/ConsoleOutput.php
```diff
@@ -0,0 +1,100 @@
...
+        $this->io->note('Running infection with an unknown log-verbosity option, falling back to ' . $default . ' option');
+    }
+
+    public function logInitialTestsDoNotPass(Process $initialTestSuitProcess, string $testFrameworkKey)
+    {
+        $lines = [
+            'Project tests must be in a passing state before running Infection.',
+            sprintf(
+                '%s reported an exit code of %d.',
+                ucfirst($testFrameworkKey),
```
borNfreee: while we at it, could you please change it to `AbstractTestFrameworkAdapter::getName()`

src/Command/InfectionCommand.php
```diff
@@ -511,8 +427,53 @@ protected function initialize(InputInterface $input, OutputInterface $output)
...
     }
+
+    /**
+     * Checks if the infection run passed the constraints like min msi & covered msi.
+     *
+     * @param MetricsCalculator $metricsCalculator
+     *
+     * @return int
+     */
+    private function didRunPassConstraints(MetricsCalculator $metricsCalculator): int
```
sidz: should it be smth like `hasRunPassConstraints` or `isRunPassConstraints` ?

src/Console/ConsoleOutput.php
```diff
@@ -0,0 +1,104 @@
...
+            $this->io->error(
+                sprintf(
+                    self::CI_FLAG_ERROR,
+                    'MSI',
+                    $minMsi,
+                    $metricsCalculator->getMutationScoreIndicator()
+                )
+            );
+
+            return;
```
sidz: maybe it'd be better to do smth like:
```
if (!$minMsi) {
    throw new exception...
}

$this->io->error...
```
It'll remove useless `return`

src/Console/ConsoleOutput.php
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        throw MsiCalculationException::create('min-msi');
+    }
+
+    public function logBadCoveredMsiErrorMessage(MetricsCalculator $metricsCalculator, float $minCoveredMsi)
+    {
+        if ($minCoveredMsi) {
```
sidz: the same here

src/Command/InfectionCommand.php
BackEndTea: How about `hasRunPassedConstraints` ?

sanmai:
I guess now we can have a test suite for `ConsoleOutput`. Do you think this should go to another PR?

borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee @sanmai

PR #339, opened 2018-05-06
Enhacement: AutoReview tests

BackEndTea:
This PR:

- [x] Adds `AutoReview` tests

Currently it checks the following:
* `bin/infection` must be executable
* All source classes/interfaces/traits should be marked `@internal`, unless they are part to be used by our users.
* All source classes should be final or abstract, unless part of the 'Legacy' list of classes that are extensively mocked in our tests.
* Source classes should not expose public properties
* All tests should be marked internal, this is for people who use `--prefer-source` when running composer.
* All tests should be final or abstract.
* Mutators do not declare public methods
* All Mutators have tests

The goal of these tests is to help (new) contributors in finding style errors that would otherwise have to be mentioned in a PR review. Therefore all messages should be informative about the error, and how it can/should be fixed.

This adds roughly 1500 tests to our code base, which run in about 1 seconds. Its not part of the default phpunit tests, but it is included in the `make analyze` command, which is now mentioned in the `CONTRIBUTING.md`.

This is almost a copy paste of the AutoReview tests in php-cs-fixer: https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/master/tests/AutoReview


BackEndTea:
@infection/core what do you think about these tests?

tests/AutoReview/BuildConfigYmlTest.php
```diff
@@ -0,0 +1,53 @@
...
+use Symfony\Component\Yaml\Yaml;
+
+/**
+ * @internal
+ *
+ * @group auto-review
+ *
+ * @coversNothing
+ */
+final class BuildConfigYmlTest extends TestCase
```
BackEndTea: Now that we have an `AutoReview` namespace, this test makes more sense here.

tests/AutoReview/MutatorTest.php
```diff
@@ -0,0 +1,162 @@
...
+                $className,
+                $testClassName
+            )
+        );
+    }
+
+    public function provideConcreteMutatorClasses()
+    {
+        return array_map(
+            static function ($item) {
```
borNfreee: interesting, didn't see it before. 

just curious - any benefits you are aware of except class is not bound to anonymous fn anymore?

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -0,0 +1,404 @@
...
+    {
+        $rc = new \ReflectionClass($className);
+        $docBlock = $rc->getDocComment();
+
+        if (in_array($className, self::$extensionPoints)) {
+            if ($docBlock === false) {
+                $this->markTestSkipped(
+                    sprintf(
+                        'The "%s" class is an extension point, but does not have a doc-block.' .
+                        'Consider adding a doc-block to improve usability.',
```
borNfreee: please add a space before `Consider`, now it looks like `...a doc-block.Consider...`

BackEndTea:
> I vote for adding this group of tests at least to the Travis builds, if not to the default test suite

It is part of the Travis build, because travis runs `make analyze --keep-going`.

However it may be better to run it within the standard php-unit run, as 99% of people knows how to use it, and expects to run it during development, but not everyone may be familiar with `make` or read the contributing.md

.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -1,44 +1,44 @@
----
```
BackEndTea: These changes are a bit of a side-effect of another auto-review i was experimenting with, as these files used the wrong line-ending.

tests/AutoReview/ProjectCodeTest.php
BackEndTea: :+1:  fixed it.

tests/AutoReview/MutatorTest.php
BackEndTea: The function can't be bound to a class, later on. I mostly copy pasted it this way from php-cs-fixer, so we can make it non static if we want to.

sanmai: `static function` is slightly faster (I don't think this will be noticeable on a number of invocations we're having here, probably added out of a habit)

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -88,6 +88,10 @@
 
     public function test_infection_bin_is_executable()
     {
+        if (DIRECTORY_SEPARATOR === '\\') {
```
borNfreee: we have such check for windows:

```
if (stripos(PHP_OS, 'WIN') === 0) {
```

would it be consistent to use the same? not a big deal though

borNfreee:
Thank you @BackEndTea

BackEndTea:

keradus:
> This is almost a copy paste of the AutoReview tests in php-cs-fixer: https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/master/tests/AutoReview

ha, kerad is proud !

PR #341, opened 2018-05-07
Replace our CloneVisitor with the one provided by PhpParser

BackEndTea:
This PR:

- [x] Replaces our own CloneVisitor with the PhpParser one.


borNfreee:
Was it closed because of massive difference in memory usage?

BackEndTea:
I closed this as on re-evaluation it seems we have a bit of a different goal with our clone visitor than the one provided by php-parser. 

Plus in reality this is only a few line of code to maintain, so its not providing us with a real advantage of using the one provided by php-parser.

PR #343, opened 2018-05-10
Run a test framework executable as it is

### LINKED ISSUE #300: Infection does not correctly detect whether phpunit is php-executable
| Question    | Answer
| Infection version | 0.8.1
| Test Framework version | PHPUnit 7.1.1
| PHP version | 7.2.4
| Platform    | Ubuntu
| Github Repo | -

In my current system, the `phpunit` executable that's reachable through my $PATH is neither a php file nor is it a phar file, but a bash script:

```bash
cat (which phpunit)
#!/bin/bash
###
# Copyright (C) 2018 Luis Pabon
#
# This software may be modified and distributed under the terms
# of the MIT license.  See the LICENSE file for details.
###
# This script will try to run first phpunit installed in your project (assumes standard vendor bin location, modify if otherwise)
# then fall back to a phar installed version.
# Usage: create and chmod +x an empty file somewhere in your path (eg ~/bin/phpunit) then tweak PHPUNIT_PHAR below for your system.
###

PHPUNIT_PROJECT="vendor/bin/phpunit"
PHPUNIT_PHAR="${HOME}/bin/phpunit-phar/phpunit-latest"
if [[ -f ${PHPUNIT_PROJECT} ]]; then
    ${PHPUNIT_PROJECT} $@
else
    ${PHPUNIT_PHAR} $@
fi
```
Infection isn't currently checking php can run this file, but it is piping it into php anyway with the following result:

image

`infection.log` contains (many instances of) the following:

```diff
47) /home/luis/EDITED/EDITED/backend/EDITED/src/EDITED/Helper/Combinations.php:39    [M] Throw_
exec /usr/bin/php7.2 -c /tmp/infectioniA9A4I  /home/luis/bin/phpunit --configuration /tmp/infection/phpunitConfiguration.7e482421e74a43edd340f041a81894ad.infection.xml --stop-on-failure 
--- Original
+++ New
@@ @@
     public function getNumberOfCombinations(int $numItems, int $groupSize) : int
     {
         if ($numItems < 0) {
-            throw new \InvalidArgumentException('$numItems must greater or equal than 0');
+            new \InvalidArgumentException('$numItems must greater or equal than 0');
         }
         if ($groupSize < 1) {
             throw new \InvalidArgumentException('$groupSize must greater or equal than 1');

###
# Copyright (C) 2018 Luis Pabon
#
# This software may be modified and distributed under the terms
# of the MIT license.  See the LICENSE file for details.
###
# This script will try to run first phpunit installed in your project (assumes standard vendor bin location, modify if otherwise)
# then fall back to a phar installed version.
# Usage: create and chmod +x an empty file somewhere in your path (eg ~/bin/phpunit) then tweak PHPUNIT_PHAR below for your system.
###

PHPUNIT_PROJECT="vendor/bin/phpunit"
PHPUNIT_PHAR="${HOME}/bin/phpunit-phar/phpunit-latest"
if [[ -f ${PHPUNIT_PROJECT} ]]; then
    ${PHPUNIT_PROJECT} $@
else
    ${PHPUNIT_PHAR} $@
fi

```

I cannot work around this problem on 0.8.x because it lacks 59ed180c53dd94cd80eb.

sanmai:
Run a test framework executable as it is if we're using a standard CLI, and there's a standard interpreter available on PATH.

This lets folks use, say, a bash wrapper over phpunit.

Fixes #300

- [x] E2E test added

### DIFF
--- a/src/TestFramework/AbstractTestFrameworkAdapter.php
+++ b/src/TestFramework/AbstractTestFrameworkAdapter.php
@@ -136,6 +136,23 @@ abstract class AbstractTestFrameworkAdapter
             );
         }
 
+        /*
+         * Run an executable as it is if we're using a standard CLI and
+         * there's a standard interpreter available on PATH.
+         *
+         * This lets folks use, say, a bash wrapper over phpunit.
+         */
+        if ('cli' == PHP_SAPI && empty($phpExtraArgs) && is_executable($frameworkPath) && `command -v php`) {
+            return sprintf(
+                '%s %s',
+                'exec',
+                $frameworkPath
+            );
+        }
+
+        /*
+         * In all other cases run it with a choosen PHP interpreter
+         */
         return sprintf(
             '%s %s %s %s',
             'exec',
--- a/src/TestFramework/AbstractTestFrameworkAdapter.php
+++ b/src/TestFramework/AbstractTestFrameworkAdapter.php
@@ -151,7 +151,7 @@ abstract class AbstractTestFrameworkAdapter
         }
 
         /*
-         * In all other cases run it with a choosen PHP interpreter
+         * In all other cases run it with a chosen PHP interpreter
          */
         return sprintf(
             '%s %s %s %s',
--- a/src/Finder/TestFrameworkFinder.php
+++ b/src/Finder/TestFrameworkFinder.php
@@ -96,7 +96,12 @@ class TestFrameworkFinder extends AbstractExecutableFinder
             return $this->customPath;
         }
 
-        $candidates = [$this->testFrameworkName, $this->testFrameworkName . '.phar'];
+        $candidates = [
+            $this->testFrameworkName,
+            $this->testFrameworkName . '.phar',
+            $this->testFrameworkName . '.bat',
+        ];
+
         $finder = new ExecutableFinder();
 
         foreach ($candidates as $name) {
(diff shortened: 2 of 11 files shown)


src/TestFramework/AbstractTestFrameworkAdapter.php
```diff
@@ -136,6 +136,23 @@ private function makeExecutable(string $frameworkPath, bool $includeArgs = true,
             );
         }
 
+        /*
+         * Run an executable as it is if we're using a standard CLI and
+         * there's a standard interpreter available on PATH.
+         *
+         * This lets folks use, say, a bash wrapper over phpunit.
+         */
+        if ('cli' == PHP_SAPI && empty($phpExtraArgs) && is_executable($frameworkPath) && `command -v php`) {
```
BackEndTea: Could you explain what  `command -v php` means/does?

sanmai:
Bash script from #300 isn't as correct and proper as I would have made it, which makes a little point in using exactly such script, but I'll see if I can add an E2E test nevertheless.

src/TestFramework/AbstractTestFrameworkAdapter.php
BackEndTea: Ah, i never knew using backticks like that was possible.

http://php.net/manual/en/language.operators.execution.php

BackEndTea:
Could you combine this by updating
```diff
- $candidates = [$this->testFrameworkName, $this->testFrameworkName . '.phar']
+ $candidates = [$this->testFrameworkName .'.bat', $this->testFrameworkName, $this->testFrameworkName . '.phar']
```

In the `TestFrameworkFinder`.

That should fix windows builds, since the current patch does not yet fix it.

borNfreee:
Thank you @sanmai

PR #347, opened 2018-05-11
Add php unit cs fixer rules

BackEndTea:
Adding these rules will cut down the number of inconsistencies within
our test suite.

This PR:

- [x] Adds new php-cs-fixer rules for phpunit tests.

The following rules got added:
* `php_unit_set_up_tear_down_visibility` 
> Changes the visibility of the `setUp()` and `tearDown()` functions of PHPUnit to protected, to match the PHPUnit TestCase.

* `php_unit_strict`
> PHPUnit methods like `assertSame` should be used instead of `assertEquals`.

* `php_unit_test_annotation`   Configured to prefix methods with `@test` and to use snake case.
> Adds or removes `@test` annotations from tests, following configuration.

```bash
$ ./php-cs-fixer-v2.phar describe @PHPUnit60Migration:risky 
Description of @PHPUnit60Migration:risky set.

 * php_unit_dedicate_assert risky
   | PHPUnit assertions like "assertInternalType", "assertFileExists", should be used over "assertTrue".
   | Configuration: ['target' => '5.6']

 * php_unit_expectation risky
   | Usages of `->setExpectedException*` methods MUST be replaced by `->expectException*` methods.
   | Configuration: ['target' => '5.6']

 * php_unit_mock risky
   | Usages of `->getMock` and `->getMockWithoutInvokingTheOriginalConstructor` methods MUST be replaced by `->createMock` or `->createPartialMock` methods.
   | Configuration: ['target' => '5.5']

 * php_unit_namespaced risky
   | PHPUnit classes MUST be used in namespaced version, eg `\PHPUnit\Framework\TestCase` instead of `\PHPUnit_Framework_TestCase`.
   | Configuration: ['target' => '6.0']

 * php_unit_no_expectation_annotation risky
   | Usages of `@expectedException*` annotations MUST be replaced by `->setExpectedException*` methods.
   | Configuration: ['target' => '4.3']
```

This means we are using all phpunit rules of php-cs-fixer, except those related to covers annotation.

### DIFF
--- a/tests/AutoReview/ProjectCodeTest.php
+++ b/tests/AutoReview/ProjectCodeTest.php
@@ -162,7 +162,8 @@ final class ProjectCodeTest extends TestCase
         }
 
         $this->assertInternalType(
-            'string', $docBlock,
+            'string',
+            $docBlock,
             sprintf(
                 'The "%s" class is not an extension point, and should be marked as internal.',
                 $className
@@ -302,7 +303,8 @@ final class ProjectCodeTest extends TestCase
         $docBlock = $rc->getDocComment();
 
         $this->assertInternalType(
-            'string', $docBlock,
+            'string',
+            $docBlock,
             sprintf(
                 'Test class  "%s" must be marked internal.',
                 $className
--- a/tests/AutoReview/ProjectCodeTest.php
+++ b/tests/AutoReview/ProjectCodeTest.php
@@ -161,8 +161,8 @@ final class ProjectCodeTest extends TestCase
             return;
         }
 
-        $this->assertTrue(
-            is_string($docBlock),
+        $this->assertInternalType(
+            'string', $docBlock,
             sprintf(
                 'The "%s" class is not an extension point, and should be marked as internal.',
                 $className
@@ -301,8 +301,8 @@ final class ProjectCodeTest extends TestCase
         $rc = new \ReflectionClass($className);
         $docBlock = $rc->getDocComment();
 
-        $this->assertTrue(
-            is_string($docBlock),
+        $this->assertInternalType(
+            'string', $docBlock,
             sprintf(
                 'Test class  "%s" must be marked internal.',
                 $className
--- a/tests/Config/Guesser/SourceDirGuesserTest.php
+++ b/tests/Config/Guesser/SourceDirGuesserTest.php
@@ -92,15 +92,14 @@ JSON;
         $this->assertSame(['sources', 'libs'], $guesser->guess());
     }
 
-    /**
-     * @expectedException \LogicException
-     * @expectedExceptionMessage autoload section does not match the expected JSON schema
-     */
     public function test_it_throw_invalid_autoload_exception()
     {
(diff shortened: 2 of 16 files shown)


tests/Config/Guesser/SourceDirGuesserTest.php
```diff
@@ -92,16 +92,15 @@ public function test_it_returns_list_if_contains_array_of_paths_without_src()
...
-     * @expectedException \LogicException
-     * @expectedExceptionMessage autoload section does not match the expected JSON schema
-     */
     public function test_it_throw_invalid_autoload_exception()
     {
         $guesser = new SourceDirGuesser(
             json_decode('{"autoload":{"psr-4": [{"NameSpace\\//": ["sources", "libs"]}]}}')
         );
 
+        $this->expectException(\LogicException::class);
```
sidz: imo annotations look better :)

borNfreee: https://phpunit.de/manual/6.5/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.exceptions

PHPUnit documentation does not say that annotations worse or better than code, this is just an alternative. So I am also curious why PHPCSFixer forces to use the code?

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -303,8 +303,8 @@ public function test_all_test_classes_are_marked_internal(string $className)
         $rc = new \ReflectionClass($className);
         $docBlock = $rc->getDocComment();
 
-        $this->assertTrue(
-            is_string($docBlock),
+        $this->assertInternalType(
+            'string', $docBlock,
```
borNfreee: probable autofixed?

please use one parameter per line, here and above

tests/Config/Guesser/SourceDirGuesserTest.php
localheinz: See [Questioning PHPUnit Best Practices](https://thephp.cc/news/2016/02/questioning-phpunit-best-practices).

BackEndTea:
> PHPUnit documentation does not say that annotations worse or better than code, this is just an alternative. So I am also curious why PHPCSFixer forces to use the code?

I'm not 100% sure on their reasons, but a few that i like

* You can specify *when* the exception is to happen. If one of your set up steps could also cause an exception than you won't know that it actually failed in the arrange step.
* Not having to use the FQN. You can just import the class with a use statement.

A bit of googling also pointed me to this article https://thephp.cc/news/2016/02/questioning-phpunit-best-practices which basically states the same.

tests/Config/Guesser/SourceDirGuesserTest.php
borNfreee: nice explanation. That moment when I don't understand why it's not in the documentation
thank you

borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee  @sidz @localheinz

PR #348, opened 2018-05-11
Simplify the EventDispatcher

BackEndTea:
This PR:

- [x] Removes dead code from the EventDispatcher

Most of its public functions were not being used within the code base(anymore).
Removed the unused functions, the now useless interface,and updated tests to show this behaviour.


src/EventDispatcher/EventDispatcher.php
```diff
@@ -31,41 +31,24 @@ public function dispatch($event)
...
-     * {@inheritdoc}
+     * @param string $eventName
+     *
+     * @return callable[]
      */
-    public function getListeners($eventName)
+    private function getListeners($eventName)
     {
-        if (!$this->hasListeners($eventName)) {
+        if (!isset($this->listeners[$eventName])) {
```
sidz: can be simplified as well: 
`return $this->listeners[$eventName] ?? [];`

src/EventDispatcher/EventDispatcher.php
```diff
@@ -31,41 +31,24 @@ public function dispatch($event)
...
     }
 
     /**
-     * {@inheritdoc}
+     * @param string $eventName
+     *
+     * @return callable[]
      */
-    public function getListeners($eventName)
+    private function getListeners($eventName)
```
sidz: seems like we can add `string` type hinting here

borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee @sidz

PR #352, opened 2018-05-16
Run Infection on Travis with xdebug

### LINKED ISSUE #344: Big difference between MSI for phpdbg and Xdebug
In master, there is a big difference in the builds between Covered MSI reported for infection itself

* `phpdbg` ([91%](https://travis-ci.org/infection/infection/jobs/377589047#L1083) CMSI)
* `XDebug` ([74%](https://travis-ci.org/infection/infection/jobs/377589048#L1060) CMSI)

Need to investigate the reason. Probably different set of tests, but just in case..

sanmai:
Fixes #344

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,7 @@ script:
   - if [[ $NO_DEBUGGER == 1 ]]; then xdebug-disable; fi
   - |
     if [[ $PHPDBG != 1 ]]; then
-      bin/infection $INFECTION_FLAGS;
+      INFECTION_ALLOW_XDEBUG=1 bin/infection $INFECTION_FLAGS;
     else
       phpdbg -qrr bin/infection $INFECTION_FLAGS;
     fi


.travis.yml
```diff
@@ -82,7 +82,7 @@ script:
   - if [[ $NO_DEBUGGER == 1 ]]; then xdebug-disable; fi
   - |
     if [[ $PHPDBG != 1 ]]; then
-      bin/infection $INFECTION_FLAGS;
+      INFECTION_ALLOW_XDEBUG=1 bin/infection $INFECTION_FLAGS;
```
sidz: could you please explain why do we need to run xdebug? 
if you check `$INFECTION_FLAGS` you will see that we put coverage path there. We did it on purpose to speed up travis builds.

sanmai: See issue #344 and comments starting with this https://github.com/infection/infection/issues/344#issuecomment-388598387

It is not like that's the only way to go. Another way is to ignore issue #344 completely, since we're getting valid metrics from phpdbg runs.

BackEndTea:
*If* we want to run travis with xdebug enabled it should be on a different run, since its not the standard use case.

Also letting our users set `INFECTION_ALLOW_XDEBUG` seems like a bad idea. We should probably use a commandline flag that does this for us.

sanmai:
Well, we only run Infection with xdebug enable in a single distinct case: when running upon itself. So we are not using `INFECTION_ALLOW_XDEBUG` as a standard for Travis.

It appear to me that using a command line flag will be marginally better than `INFECTION_ALLOW_XDEBUG` since most of the time Infection spends on running tests, and if we're using a command line flag we'll be still spending the same time running tests. I'm not sure it is worth it to add a specialized flag, this needs a benchmark to decide.

BackEndTea:
> It appear to me that using a command line flag will be marginally better than INFECTION_ALLOW_XDEBUG since most of the time Infection spends on running tests, and if we're using a command line flag we'll be still spending the same time running tests. I'm not sure it is worth it to add a specialized flag, this needs a benchmark to decide.

My concern isn't necessarily with the speed, its about user expectations. All other infection configuration is done through the config file, or as a cli flag. Setting environment variables isn't something i'd expect as an end user.

I'd even be okay if we add a `--allow-xdebug` flag, and all that did was set `INFECTION_ALLOW_XDEBUG=1` since at least its in line with other infection settings, from the users point of view.

Plus, if we ever want to change away from the current xdebug handler, we can do that a bit easier.

BackEndTea:
> That's certainly true if we expect a user to be able to turn xdebug on and off, now and then. Do we?

Its a bit of an edge case. But our own test suite fits that edge case, as we need xdebug to run some tests, and we'd like to run as much tests as possible when calculating our MSI etc.

BackEndTea:
@infection/core I'd like to close this and #344 as a wontfix for now. Unless we find users that actually require xdebug in their test-suite.

What do you think?

borNfreee:
+1 for wontfix

sidz:
@BackEndTea just do it :)

Issue #356, opened 2018-05-17
Infection infects /tmp


Majkl578:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.8.2
| Test Framework version | PHPUnit 7
| PHP version | 7.2
| Platform    | Linux

Infection pollutes /tmp with thousands of files.
```
$ du -hs /tmp/infection
270M	/tmp/infection
$ ll /tmp/infection | wc -l
18706
```

This should be cleaned up after run. Otherwise it eats space and RAM (`/tmp` is typically `tmpfs` which is RAM-based).

borNfreee:
Probably, it should be configurable since we reuse created mutants to avoid extra IO operations

https://github.com/infection/infection/blob/2276560c4886c79fd6b7175a6039f5c6650d8a93/src/Mutant/MutantCreator.php#L71-L76

On the other side, I'm not sure many people will run infection many times in a row

sidz:
> Once thing that could be done however is create the temporary files in a infection directory of the temporary directory and clean it up at the start of the run (to be careful that it won't affect/collide with the one used for XdebugHandler)

We can't guarantee that user didn't change `tmpDir` between infection execution. IMO we need to clean `tmpDir` after a run.

theofidry:
I don't think the user should change it neither rely that it won't change.

That said it might be an issue with concurrent runs

borNfreee:
fixed https://github.com/infection/infection/pull/357

PR #357, opened 2018-05-17
Clean up after Mutation testing has been finished

sidz:
Remove all files from the temporary dir after infection finished mutation testing.

Cover: https://github.com/infection/infection/issues/356

p.s. Doesn't cover cleanup when interrupting the command

### DIFF
--- a/tests/Fixtures/e2e/Custom_tmp_dir/infection.json
+++ b/tests/Fixtures/e2e/Custom_tmp_dir/infection.json
@@ -8,5 +8,5 @@
     "logs": {
         "summary": "infection-log.txt"
     },
-    "tmpDir": "/tmp/infection-cache"
+    "tmpDir": "."
 }
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -29,6 +29,7 @@ use Infection\Process\Listener\InitialTestsConsoleLoggerSubscriber;
 use Infection\Process\Listener\MutantCreatingConsoleLoggerSubscriber;
 use Infection\Process\Listener\MutationGeneratingConsoleLoggerSubscriber;
 use Infection\Process\Listener\MutationTestingConsoleLoggerSubscriber;
+use Infection\Process\Listener\CleanUpAfterMutationTestingFinishedSubscriber;
 use Infection\Process\Listener\MutationTestingResultsLoggerSubscriber;
 use Infection\Process\Runner\InitialTestsRunner;
 use Infection\Process\Runner\MutationTestingRunner;
@@ -355,6 +356,10 @@ final class InfectionCommand extends BaseCommand
                 $metricsCalculator,
                 $this->getContainer()->get('filesystem'),
                 $this->input->getOption('log-verbosity')
+            ),
+            new CleanUpAfterMutationTestingFinishedSubscriber(
+                $this->getContainer()->get('filesystem'),
+                $this->getContainer()->get('tmp.dir')
             ),
         ];
     }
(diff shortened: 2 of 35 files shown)


BackEndTea:
If we implement this we can remove the logic to check if the mutated files already exist.

sidz:
@BackEndTea yeah, that's what @borNfreee  was talking about in the issue.

tests/Fixtures/e2e/Custom_tmp_dir/infection.json
```diff
@@ -8,5 +8,5 @@
     "logs": {
         "summary": "infection-log.txt"
     },
-    "tmpDir": "/tmp/infection-cache"
+    "tmpDir": "."
```
sanmai: Have you tried running `make test` with this change?

borNfreee: could you please explain why `.`?

borNfreee:
I'm still wonder whether we should to do it configurable.

During debugging, I do a lot of checking in this `tmpDir` folder, finding temp phpunit configs or Mutant files.

Moreover, our `infection-log.txt` contains the whole executable command line with path to tmp `php.ini`, with path to custom `phpunit.xml`, which is pretty useful to just copy and past to terminal, running the test suite like it was run by Infection.

Now, it will be impossible and will complicate debugging (in this case we would have to disable substciber manually to keep `tmpDir` during debugging).

I suggest to review the following ideas: 

* do not remove `tmpDir` when using debug mode
  * we can add `--debug` option as was suggested by @theofidry https://github.com/infection/infection/pull/275#issuecomment-377485355
  * we can add `APP_DEBUG` env var like mentioned here https://symfony.com/doc/current/console.html#the-console-app-env-app-debug 
* add something like `--keep-tmp-dir
* etc.

Personally, I would prefer `--debug` option as it would be useful not only for this particular case, but for other similar situation in the future.

ryan7273:
I agree with the --debug option. When I first switched from Humbug to Infection, it took me a week of sifting through those files to figure out why it wasn't working correctly. If tmpDir had been cleaned I wouldn't have had anything to work with. It also seems reasonable that there may be other changes/logging which you would want when debugging but would not need during normal usage.

tests/Fixtures/e2e/Custom_tmp_dir/infection.json
sidz: @borNfreee because temporary folder will be erased when you'll run infection on itself. e2e will clean up temporary folder.

sanmai: I doubt this is going to work on Docker since a current directory is non-writable there.

theofidry:
I would add a `--debug`. Unlike `-vvv` we are talking of a debug _mode_ and not verbosity. As an example, for Box you can run the `compile` command with `--debug`  with sets the verbosity to debug and dump the code shipped into the PHAR in a `.box_dump` directory with some extra info

sidz:
already implemented, but didn't push yet :yum:

tests/Fixtures/e2e/Custom_tmp_dir/infection.json
sidz: @sanmai locally I'm getting only
```
Phpunit_Bat_Wrapper/


In FinderException.php line 33:
                                                                                                          
  Unable to locate a phpunit executable on local system. Ensure that phpunit is installed and available. 
```

ubuntu 18.04

sanmai:
Why do we have to use `--debug ` if we already have a debugging `--log-verbosity`?

borNfreee:
@sanmai see https://github.com/infection/infection/pull/357#issuecomment-390945317

because `--debug` is *not only* about logging, and this example shows it (we do not clear `tmpDir` when it is a debug mode).

While `--log-verbosity` is for our users for normal usage, `--debug` will be useful for those who, well, debug infection ;)

borNfreee:
Could you please add a documentation for this new option @sidz ?

sidz:
@borNfreee doen

borNfreee:
Thank you @sidz

PR #359, opened 2018-05-23
Add/Update/Improve unit tests 

BackEndTea:
This PR:

- [x] Covered by tests

### DIFF
--- a/src/Events/MutantProcessFinished.php
+++ b/src/Events/MutantProcessFinished.php
@@ -26,9 +26,6 @@ final class MutantProcessFinished
         $this->mutantProcess = $mutantProcess;
     }
 
-    /**
-     * @return MutantProcessInterface
-     */
     public function getMutantProcess(): MutantProcessInterface
     {
         return $this->mutantProcess;
--- a/src/Mutator/FunctionSignature/ProtectedVisibility.php
+++ b/src/Mutator/FunctionSignature/ProtectedVisibility.php
@@ -74,7 +74,7 @@ final class ProtectedVisibility extends Mutator
                     return true;
                 }
             } catch (\ReflectionException $e) {
-                continue;
+                return false;
             } finally {
                 $parent = $parent->getParentClass();
             }
--- a/tests/Fixtures/Autoloaded/ProtectedVisibility/pv-final.php
+++ b/tests/Fixtures/Autoloaded/ProtectedVisibility/pv-final.php
@@ -4,7 +4,7 @@ namespace ProtectedVisibilityFinal;
 
 class Test
 {
-    protected final function foo(int $param, $test = 1) : bool
+    protected final function &foo(int $param, $test = 1) : bool
     {
         echo 1;
         return false;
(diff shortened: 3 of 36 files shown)


src/Events/MutantProcessFinished.php
```diff
@@ -26,9 +26,6 @@ public function __construct(MutantProcessInterface $mutantProcess)
         $this->mutantProcess = $mutantProcess;
     }
 
-    /**
-     * @return MutantProcessInterface
-     */
```
BackEndTea: These doc blocks don't add any value since we have the return type

src/Mutator/FunctionSignature/ProtectedVisibility.php
```diff
@@ -74,7 +74,7 @@ private function hasSameProtectedParentMethod(Node $node): bool
                     return true;
                 }
             } catch (\ReflectionException $e) {
-                continue;
```
BackEndTea: We get a reflection exception when the method doesn't exist on that class. If it doesn't exists on that class it also doesn't exist on a parent class. Since inherited, not overridden functions are also found by this. So if we get an error we know the paren't doesn't have a same protected method

tests/Fixtures/Autoloaded/ProtectedVisibility/pv-final.php
```diff
@@ -4,9 +4,9 @@
 
 class Test
 {
-    protected final function foo(int $param, $test = 1) : bool
```
BackEndTea: This way we test if we don't mutate whether a method passes by reference. As this could cause confusion in the diff.

tests/Mutator/AbstractMutatorTestCase.php
```diff
@@ -34,12 +34,14 @@
 
     public function doTest(string $inputCode, string $expectedCode = null)
     {
+        $inputCode = rtrim($inputCode, "\n");
```
BackEndTea: We trim any trailing new line for both the input and expected code, so we don't have to worry about that any more.

tests/Mutator/ReturnValue/ThisTest.php
```diff
@@ -27,7 +27,7 @@ public function test_mutator($input, $expected = null)
 
     public function provideMutationCases(): \Generator
     {
-        yield 'It does not mutate with not nullable return typehint' => [
```
BackEndTea: This was probably a copy paste error

tests/Mutator/ZeroIteration/Foreach_Test.php
```diff
@@ -24,28 +24,44 @@ public function test_mutator($input, $expected = null)
...
 
 $array = [1, 2];
 foreach (array() as $value) {
 }
 PHP
-                ,
-            ],
+            ,
+        ];
+        yield 'It does not change whether items were passed by reference' => [
```
BackEndTea: Once again, to make sure the diffs stay relevant of what we are doing, and not to cause confusion.

borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee

PR #361, opened 2018-05-29
Stop mutation of abstract methods default parameters

BackEndTea:
This PR:

- [x] Stops default parameters of abstract methods from being mutated. These are never covered by tests, so they never result in a killed mutant.
- [x] Covered by tests

```php
 abstract public function search($value, bool $strict = false);
```
Should not be mutated in any way, as it is never covered by tests

### DIFF
--- a/src/Visitor/MutationsCollectorVisitor.php
+++ b/src/Visitor/MutationsCollectorVisitor.php
@@ -72,10 +72,17 @@ final class MutationsCollectorVisitor extends NodeVisitorAbstract
 
             $isOnFunctionSignature = $node->getAttribute(ReflectionVisitor::IS_ON_FUNCTION_SIGNATURE, false);
 
-            if (!$isOnFunctionSignature) {
-                if (!$node->getAttribute(ReflectionVisitor::IS_INSIDE_FUNCTION_KEY)) {
-                    continue;
-                }
+            if (!$isOnFunctionSignature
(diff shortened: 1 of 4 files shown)


src/Visitor/MutationsCollectorVisitor.php
```diff
@@ -72,10 +72,17 @@ public function leaveNode(Node $node)
...
-                if (!$node->getAttribute(ReflectionVisitor::IS_INSIDE_FUNCTION_KEY)) {
-                    continue;
-                }
+            if (!$isOnFunctionSignature
+                && !$node->getAttribute(ReflectionVisitor::IS_INSIDE_FUNCTION_KEY)
+            ) {
+                continue;
+            }
+
+            if ($isOnFunctionSignature
```
borNfreee: hm..

[we have a test](https://github.com/infection/infection/blob/master/tests/Fixtures/Autoloaded/PublicVisibility/pv-same-method-abstract.php) that proofs it should not be mutated on the mutator level:

https://github.com/infection/infection/blob/638e06296271fdb10193159cec0a8139af1ce9d8/tests/Mutator/FunctionSignature/PublicVisibilityTest.php#L178-L180

Probably, something is broken in another place

borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee

Issue #364, opened 2018-06-01
False positive: count(-1) (decrement integer mutator)


luispabon:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.8.2
| Test Framework version | phpunit 7.1
| PHP version | php 7.2
| Platform    | Ubuntu

I have the following code being mutated by DecrementInteger

```php
 if (\count($diff) === 0) {
    // stuff
}
```

into
```php
 if (\count($diff) === -1) {
    // stuff
}
```

This is clearly a false positive as `count` can't possibly return 0.

borNfreee:
this should now be fixed in https://github.com/infection/infection/pull/365

luispabon:
:D

Issue #206, opened 2018-03-05
Allow settings for Mutators


BackEndTea:
With the proposal of #79, implemented like https://github.com/infection/infection/issues/89#issuecomment-365725972, we could extend the config file yet again to pass 'settings' to the Mutators.

Given the following config file:
```js
{
    "timeout": 25,
    "mutators": {
        "@library": true,
        "Minus": true, // included
        "Plus": false, // completely ignored
        "TrueValue": { // ignored for particular cases
            "ignore": [
                "Ignore\\For\\Particular\\Class",
                "Ignore\\For\\Another\\Class::method",
                "Ignore\\For\\**\\*\\Glob\\Pattern\\Or\\Namespace"
            ],
            "settings": {
                "in_array": false,
            }
        },
        "DecrementInteger": { // ignored for particular cases
            "settings": {
                "symfonty_event_subscriber": false,
            }
        }
    }
}
```

The `TrueValue` mutator could have a setting that it will not mutate for the in_array function. The `DecrementInteger` could have a setting that it does not mutate within the `getSubscribedEvents()` function, as a lot of time a +1 or -1 on those doens't change anything.

All settings should have a reasonable default value, and the config file is used to overwrite those.

At this point we should also look into some kind of json validation, so if invalid keys are passed, that are still valid json it should give some kind of notification to the user, and either quit entirely or continue if possible.

borNfreee:
Good idea!

> At this point we should also look into some kind of json validation, so if invalid keys are passed, that are still valid json it should give some kind of notification to the user, and either quit entirely or continue if possible.

Yeah, with such a complex config file we will definitely need some kind of validation, as @localheinz already [mentioned](https://github.com/infection/infection/issues/204#issuecomment-370335823). 

Probably, if we are going to implement it, we can already add a new dependency (json validation) for https://github.com/infection/infection/issues/205 as both features will benefit from it.

PR #365, opened 2018-06-01
Do not decrement integer 0 when it is being compared with the result of count()

borNfreee:
Do not decrement integer 0 when it is being compared with the result of `count()`

Incorrect mutation:

```diff
- if (count($a) === 0) {
+ if (count($a) === -1) {
```

This PR:

- [x] Fixes https://github.com/infection/infection/issues/364
- [x] Covered by tests

### DIFF
--- a/src/Mutator/Number/DecrementInteger.php
+++ b/src/Mutator/Number/DecrementInteger.php
@@ -18,7 +18,7 @@ use PhpParser\Node;
  */
 final class DecrementInteger extends Mutator
 {
-    private const COUNT_NAME = 'count';
+    const COUNT_NAME = 'count';
 
     /**
      * Decrements an integer by 1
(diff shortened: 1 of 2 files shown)


borNfreee:
nice cases. I've updated the code

src/Mutator/Number/DecrementInteger.php
```diff
@@ -18,6 +18,8 @@
  */
 final class DecrementInteger extends Mutator
 {
+    private const COUNT_NAME = 'count';
```
BackEndTea: Can't have private const in php 7

BackEndTea:
I like that we are reducing false positives, but as i mentioned, if count is run on a user defined object implementing the `Countable` interface, then -1 might be a valid option.

Its outside of the scope of this PR, but it might be time to start looking into #206 and turn this into a setting.

borNfreee:
As for me, returning -1 as a result of `count()` is a code smell. It must be 0+

At least this is how it works in real life :)

borNfreee:
Thank you @BackEndTea

PR #367, opened 2018-06-02
Run unit tests on PHP 7.0 on pre-push

borNfreee:
We should run unit tests on PHP 7.0 rather than PHP intalled on the host machine of the developer.

It will help to avoid situations like [this](https://github.com/infection/infection/pull/365#discussion_r192553842)

### DIFF
--- a/devTools/pre-push
+++ b/devTools/pre-push
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-make analyze && ./vendor/bin/phpunit
+make analyze && make test-unit-70


BackEndTea:
There can also be edge cases that would only be covered on php 7.1 or 7.2, and im somewhat against running all tests pre-push, as that is what ci is checking for us.

borNfreee:
I am not sure I got you. This update just runs unit tests on 7.0 instead of host machine's PHP version. No new tests were added to the hook.

Could you explain what do you suggest?

BackEndTea:
Personally i'm okay with running it against the users php version, and finding errors that only exist for specific versions of php in ci.

Testing specifically against 7.0 would not catch 7.1 ony errors, and the same goes for other versions.

borNfreee:
1. If it works on PHP 7.0, it is likely works on 7.1 and other higher versions.
2. If it works on PHP 7.1, it is not guaranteed that it works on PHP 7.0

In [this](https://github.com/infection/infection/pull/365#discussion_r192553842) example I've added `private` to the `const`. `pre-push` code executed unit tests on my host machine's PHP 7.1 without any error. But it failed on CI for 7.0.

So the goal is just to reduce the possibility of pushing invalid code to the remote.

BackEndTea:
> If it works on PHP 7.0, it is likely works on 7.1 and other higher versions.

True, but we do test features that are 7.1+, which are skipped on 7.0 (nullable returns etc)

borNfreee:
@BackEndTea do you have any objections/concern before merging?

borNfreee:
thank you @BackEndTea and @sanmai

PR #368, opened 2018-06-02
Check `flock` exists on current OS. If not - skip it

borNfreee:
Now it's impossible to run for example `make test-unit-70` on MacOS because it doesn't have `flock` function

`-bash: flock: command not found`

This update checks if it exists, and if not - continues execution instead of failing with the error above.

I'm newbie in bash, so advice on how to do it better would be very useful.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,9 @@ PHP-CS-FIXER=./php-cs-fixer-v2.phar
 PHPUNIT=vendor/bin/phpunit
 
 DOCKER_RUN=docker run -t --rm -v "$$PWD":/opt/infection -w /opt/infection
-DOCKER_RUN_70=flock devTools/*php70*.json $(DOCKER_RUN) infection_php70
-DOCKER_RUN_71=flock devTools/*php71*.json $(DOCKER_RUN) infection_php71
-DOCKER_RUN_72=flock devTools/*php72*.json $(DOCKER_RUN) infection_php72
+DOCKER_RUN_70=type -t flock && flock devTools/*php70*.json & $(DOCKER_RUN) infection_php70
+DOCKER_RUN_71=type -t flock && flock devTools/*php71*.json & $(DOCKER_RUN) infection_php71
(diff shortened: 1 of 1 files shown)


sanmai:
It appears there's a drop-in replacement:
```
brew tap discoteq/discoteq
brew install flock
```

Makefile
```diff
@@ -4,9 +4,9 @@ PHP-CS-FIXER=./php-cs-fixer-v2.phar
 PHPUNIT=vendor/bin/phpunit
 
 DOCKER_RUN=docker run -t --rm -v "$$PWD":/opt/infection -w /opt/infection
-DOCKER_RUN_70=flock devTools/*php70*.json $(DOCKER_RUN) infection_php70
-DOCKER_RUN_71=flock devTools/*php71*.json $(DOCKER_RUN) infection_php71
-DOCKER_RUN_72=flock devTools/*php72*.json $(DOCKER_RUN) infection_php72
+DOCKER_RUN_70=type -t flock && flock devTools/*php70*.json & $(DOCKER_RUN) infection_php70
+DOCKER_RUN_71=type -t flock && flock devTools/*php71*.json & $(DOCKER_RUN) infection_php71
+DOCKER_RUN_72=type -t flock && flock devTools/*php72*.json & $(DOCKER_RUN) infection_php72
```
sanmai: This single ampersand here makes `flock` useless in this command. In other words, Docker will always run no matter if `flock` is found or not.

sanmai:
`flock` was added here because it wasn't possible to run the whole test suite without it.

Best way to fix this problem is not by adding some workarounds, but by fixing the root cause, which is improper use of temporary directories, if I recall correctly.

sanmai:
I confirm that tests are still failing if being run simultaneously. E.g. try removing flock and run:

    make -j test-unit

`flock` is essential to the current mode of operation if one ever wants to run multiple tests in parallel.

Makefile
borNfreee: > Docker will always run no matter if flock is found or not

that was exactly what I wanted to achieve

borNfreee:
thank you @sanmai 

closing in favor of https://github.com/infection/infection/pull/373

PR #370, opened 2018-06-03
Run a unit test in a separate process

BackEndTea:
I ran tests with https://github.com/fiunchinho/phpunit-randomizer and found that `test_it_throws_an_exception_if_not_configured` had to be ran first in order to pass its test. By running it in a separate process it isn't dependent on other tests.

After the fix i ran the whole test suite 100 times with different seeds, and the include interceptor tests specifically 1000 times, but no other errors were found.

### DIFF
--- a/tests/StreamWrapper/IncludeInterceptorTest.php
+++ b/tests/StreamWrapper/IncludeInterceptorTest.php
@@ -48,13 +48,17 @@ final class IncludeInterceptorTest extends \PHPUnit\Framework\TestCase
          * if any of our tests fail for any reason.
          *
          * Silenced a warning here: stream_wrapper_restore(): file:// was never changed, nothing to restore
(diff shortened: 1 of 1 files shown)


tests/StreamWrapper/IncludeInterceptorTest.php
```diff
@@ -34,6 +34,7 @@
 
     public static function setUpBeforeClass()
     {
+        var_dump('Settign up');
```
sanmai: Should this be here?

BackEndTea: Woops, forgot to remove that during some debugging.

borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee  & @sanmai

PR #371, opened 2018-06-03
Run cs fixer

BackEndTea:
The latest version of php-cs-fixer updated the symfony profile, whihc now includes the `native_constant_invocation` rule with the following settings:
```
* native_constant_invocation risky
   | Add leading `\` before constant invocation of internal constant to speed up resolving. Constant name match is case-sensitive, except for `null`, `false` and `true`.
   | Configuration: ['fix_built_in' => false, 'include' => ['DIRECTORY_SEPARATOR', 'PHP_SAPI', 'PHP_VERSION_ID']]
```

### DIFF
--- a/bin/infection
+++ b/bin/infection
@@ -39,7 +39,7 @@ if (!isset($file)) {
 require $file;
 unset($file);
 
-if (PHP_SAPI !== 'cli' && PHP_SAPI !== 'phpdbg') {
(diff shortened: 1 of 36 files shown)


sanmai:
It appears there's a consensus that [only most used constants should receive this treatment](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3127), and that's what you're proposing exactly. Seems reasonable to me.

BackEndTea:

PR #372, opened 2018-06-03
Enhancement: enable native function invocation cs rule

BackEndTea:
This PR:

- Adds the `native_function_invocation` php-cs-fixer rule, with the `@compiler_optimized` setting.

This means that only functions which are cached as opcodes will be prefixed. As these are the functions that will see the most benefit from doing so.

```bash
$ ./php-cs-fixer-v2.phar describe native_function_invocation
Description of native_function_invocation rule.
Add leading `\` before function invocation to speed up resolving.

Fixer applying this rule is risky.
Risky when any of the functions are overridden.

Fixer is configurable using following options:
* exclude (array): list of functions to ignore; defaults to []
* include (array): list of function names or sets to fix. Defined sets are `@internal` (all native functions), `@all` (all global functions) and `@compiler_optimized` (functions that are specially optimized by Zend); defaults to ['@internal']
* scope ('all', 'namespaced'): only fix function calls that are made within a namespace or fix all; defaults to 'all'
```

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -30,6 +30,9 @@ return \PhpCsFixer\Config::create()
             'location' => 'after_open',
             'separate' => 'bottom',
         ],
+        'native_function_invocation' => [
+            'include' => ['@compiler_optimized']
+        ],
         'no_useless_else'=> true,
         'ordered_class_elements' => [
             'order' => [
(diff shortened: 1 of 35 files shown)


borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee, @sanmai & @sidz

PR #373, opened 2018-06-04
Allow easy Docker testing on Mac OS

sanmai:
Testing if flock exists or not needs somewhat more complicated logic than one wants to add to single line in a Makefile. Therefore this wrapper. Should fix the problem from #368 

This PR also includes two small fixes to make Docker testing run more smoothly in most circumstances.

### DIFF
--- a/tests/Fixtures/e2e/Parameters_Coverage/composer.json
+++ b/tests/Fixtures/e2e/Parameters_Coverage/composer.json
@@ -11,5 +11,11 @@
         "psr-4": {
             "ParamCoverage\\Test\\": "tests/"
         }
+    },
+    "config": {
+        "platform": {
+            "php": "7.0.8"
+        },
+        "sort-packages": true
     }
 }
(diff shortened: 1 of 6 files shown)


sanmai:
travis-ci build is failing because of a new rule `native_constant_invocation`, which is going to be rectified by  #371

tests/Fixtures/e2e/Parameters_Coverage/composer.json
```diff
@@ -11,5 +11,11 @@
         "psr-4": {
             "ParamCoverage\\Test\\": "tests/"
         }
+    },
+    "config": {
```
BackEndTea: Is this related to the mac os builds? or is it something else?

sanmai: Nah, that's just a random bug not worthy a separate PR: test dependencies are installed just once, and without an explicit platform definition if the first install happens not from PHP 7.0, all future tests on earlier versions will go broke.

borNfreee:
thank you @sanmai

PR #374, opened 2018-06-04
Look for .bat on all platforms

sanmai:
For Phpunit_Bat_Wrapper test to work we have to look for .bat on all platforms.

### DIFF
--- a/src/Finder/TestFrameworkFinder.php
+++ b/src/Finder/TestFrameworkFinder.php
@@ -99,6 +99,7 @@ class TestFrameworkFinder extends AbstractExecutableFinder
         $candidates = [
             $this->testFrameworkName,
             $this->testFrameworkName . '.phar',
+            $this->testFrameworkName . '.bat', // always looking for .bat for testing with .bat not on Windows
         ];
 
         /*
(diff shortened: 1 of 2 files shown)


sanmai:
Well, it turns out I was using a cached build of infection.phar... I'll check one more time, but it appears these changes are not needed. Closing for now.

src/Finder/TestFrameworkFinder.php
```diff
@@ -99,6 +99,7 @@ private function findTestFramework(): string
         $candidates = [
             $this->testFrameworkName,
             $this->testFrameworkName . '.phar',
+            $this->testFrameworkName . '.bat', // always looking for .bat for testing with .bat not on Windows
```
BackEndTea: Now we are looking for the `.bat` version twice.

sanmai: We could be like:
```
if ('\\' == \DIRECTORY_SEPARATOR) {
    array_unshift($candidates, $this->testFrameworkName . '.bat');
} else {
    $candidates[] = $this->testFrameworkName . '.bat';
}
```
Do you think this would be better? I'm not a big fan of `else` it it can be avoided.

borNfreee: IMO, in this case `else` is better than finding 2 times `.bat` file

sanmai: OK, done.

borNfreee:
Thank you @sanmai

Issue #366, opened 2018-06-01
False positive when calculation is in multiple lines


rieschl:
| Question    | Answer
| ------------| ---------------
| Infection version | dev-master#2e24f03
| Test Framework version | PHPUnit 7.2.2
| PHP version | 7.1.17
| Platform    | Debian (also tested in WSL)
| Github Repo | https://github.com/trinet-at/infection-multiline

When a calculation spans across multiple lines and infection does a mutation on it, it reports it as false positive, e.g. the original code:
```
$result = $value
    / 100
    * 100;

return $result;
```
and mutation:
```
$result = $value
    * 100
    * 100;

return $result;
```
This results in a false positive:
```
Not covered mutants:
====================

1) /mnt/e/PhpStormProjects/infection-bug/src/Calculator.php:9    [M] Division

--- Original
+++ New
@@ @@
 {
     public function calculateInMultipleLines(float $value) : float
     {
-        $result = $value / 100 * 100;
+        $result = $value * 100 * 100;
         return $result;
     }
     public function calculateInSingleLine(float $value) : float
```

When the calculation is in a single line, it works as expected:
```
$result = $value / 100 * 100;

return $result;
```
and mutation:
```
$result = $value * 100 * 100;

return $result;
```

I've created the [test repo ](https://github.com/trinet-at/infection-multiline), it's easier to see the code than to explain it :)

BackEndTea:
This is most likely due to how coverage is calculated for php. In the same way that for a multi line array, only the first value is considered covered (except when it assigned to a class property, then the assignment is considered covered)

sanmai:
It really does not report line 9 as covered...
```xml
<coverage>
  <line nr="10">
    <covered by="TrinetTest\InfectionMultiline\CalculatorTest::testCalculateInMultipleLines"/>
    <covered by="TrinetTest\InfectionMultiline\CalculatorTest::testCalculateInMultipleLinesFloat"/>
    <covered by="TrinetTest\InfectionMultiline\CalculatorTest::testCalculateInSingleLineFloat"/>
  </line>
  <line nr="11">
  ...
```

rieschl:
But is line 9 something that can be covered? It's just `$result = $value`. If it can PHOUnit seems to treat it lie whitespace because of the 100% coverage output. Line 11 and 12, which are mutated, are covered though.
Btw. looking at your coverage report I noticed that I used the wrong method call in one of the tests. I already fixed that in the latest commit.

[...]

PR #375, opened 2018-06-04
Account for statements spanning multiple lines

sanmai:
- [x] MutationsCollectorVisitor now accounts for statements spanning multiple lines
- [x] Covered by tests

Fixes https://github.com/infection/infection/issues/366

### DIFF
--- a/src/TestFramework/Coverage/CodeCoverageData.php
+++ b/src/TestFramework/Coverage/CodeCoverageData.php
@@ -110,22 +110,47 @@ class CodeCoverageData
             return [];
         }
 
-        $filePath = $mutation->getOriginalFilePath();
-        $line = $mutation->getAttributes()['startLine'];
-
         if ($mutation->isOnFunctionSignature()) {
+            return iterator_to_array($this->getTestsForMutationOnFunctionSignature($mutation), false);
+        }
+
+        return iterator_to_array($this->getTestsForMutation($mutation), false);
+    }
+
+    private function getTestsForMutationOnFunctionSignature(MutationInterface $mutation): \Generator
+    {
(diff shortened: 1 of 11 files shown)


BackEndTea:
I'm not 100% sure, but i thought that function signatures were 100% covered, e.g. for the following signature, if it is called, all lines are considered covered

```php
public function foo(
    $a,
    $b,
    $c,
    $d
    ) {
}
```

sanmai:
I'm not sure about that too, but even if that is so, it shouldn't matter. 

Anyway, I'm going to be adding tests, so I'd test for this case too.

sanmai:
@BackEndTea you were right: there's no need to check for each line on function signatures

BackEndTea:
Won't we run into issues on this line
https://github.com/infection/infection/blob/3518545a6bab661f33a47750809dfd8d487ed7b4/src/TestFramework/Coverage/CodeCoverageData.php#L124

Because there it will still look at the original line, rather than the one we found tests on.

sanmai:
Considering it reports all mutants properly caught, and no uncovered mutants... OK, let me try to think of some extra test cases which may see Infection err on that line.

sanmai:
Actually that line is important. If Infection only looks for tests that cover only the very first line of our mutation, it may naturally miss other tests that may actually also cover this mutation, this all leading to false positives. 

Yet we have to way to verify that Infection has found all tests covering specific mutation since we're not logging the number of tests found. Neither were bugs reported in this regard.

borNfreee:
I've changed base branch from `master` to `0.9`. Details are in the Slack

borNfreee:
Thank you @sanmai

PR #376, opened 2018-06-05
Makefile updated to optionally run without a .phar 

sanmai:
- [x] infection.phar should be rebuild if any of our source files change (not yet including vendor)
- [x] It should be possible to choose an executable used for e2e; .phar is the default, but testing with a standard executable is as simple as `make INFECTION=bin/infection`
- [x] Moved URL to download Box to the head of Makefile, also configurable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,11 @@ PHPSTAN=./phpstan.phar
 PHP-CS-FIXER=./php-cs-fixer-v2.phar
 PHPUNIT=vendor/bin/phpunit
 INFECTION=build/bin/infection.phar
+
+# URLs to download all tools
 BOX_URL="https://github.com/humbug/box/releases/download/3.0.0-alpha.5/box.phar"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -139,12 +141,12 @@ validate:
...
-build/bin/infection.phar: bin src vendor box.phar box.json
+build/bin/infection.phar: $(shell find bin/ src/ -type f) box.phar box.json .git/HEAD
 	php box.phar compile
 
 box.json: box.json.dist
 	cat box.json.dist | sed -E 's/\"key\": \".+\",//g' | sed -E 's/\"algorithm\": \".+\",//g' > box.json
 
 box.phar:
-	wget https://github.com/humbug/box/releases/download/3.0.0-alpha.5/box.phar
+	wget $(BOX_URL)
```
BackEndTea: What is the reason that we use a variable here, but not for the other phars?

sanmai: That's a good point. We should.

sanmai: OK, done in 343c410

borNfreee:
Thank you @sanmai

Issue #377, opened 2018-06-05
Config_Framework E2E test fails on AppVeyor


sanmai:
| Question    | Answer
| ------------| ---------------
| Infection version | master
| Test Framework version | PhpSpec
| PHP version | any
| Platform    | Windows + AppVeyor

Config_Framework E2E test sporadically fails on AppVeyor with the following explanation:

```
1) Infection\Tests\Console\E2ETest::test_it_runs_an_e2e_test_with_success with data set "Config_Framework" (Symfony\Component\Finder\SplFileInfo Object (...))
Unexpected exit code. Command output wasYou are running Infection with xdebug enabled.
 [WARNING] Infection cannot control exit codes and unable to relaunch itself.                                           
           It is your responsibility to disable xdebug/phpdbg unless needed.                                            
    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____ 
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/
 
Running initial test suite...
PhpSpec version: 4.3.0
    0 [>---------------------------] < 1 sec
    1 [->--------------------------] < 1 sec
    2 [-->-------------------------] < 1 sec [ERROR] Project tests must be in a passing state before running Infection.                                             
                                                                                                                        
         PhpSpec reported an exit code of 1.                                                                            
                                                                                                                        
         Refer to the PhpSpec's output below:                                                                           
                                                                                                                        
         STDOUT:                                                                                                        
                                                                                                                        
         TAP version 13                                                                                                 
                                                                                                                        
                                                                                                                        
         STDERR:                                                                                                        
                                                                                                                        
         Cannot load Xdebug - it was already loaded

```

PR #362, opened 2018-05-29
Only add xdebug if we dont have it set already on appveyor

BackEndTea:
This should fix the appveyor failures

Fixes #377

PR #378, opened 2018-06-05
Ignore Config_Framework E2E test on Windows as it's a known bug

sanmai:
Ignore Config_Framework E2E test on Windows as it's a known bug #377.

There's little point in tagging most PR with a failed build where they fail not because a given PR is buggy, but because of  #377. Surely when time comes this workaround has to be removed.

Either #362 or this should be merged, if any. I'm in favor of a proper fix if there is.

### DIFF
--- a/tests/Console/E2ETest.php
+++ b/tests/Console/E2ETest.php
@@ -95,6 +95,11 @@ final class E2ETest extends TestCase
      */
     public function test_it_runs_an_e2e_test_with_success(string $fullPath)
     {
+        if ('\\' == \DIRECTORY_SEPARATOR && 'Config_Framework' == basename((string) $fullPath)) {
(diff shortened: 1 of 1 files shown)


BackEndTea:
The issue isn't windows, its that phpspec errors out when xdebug attempts to load twice, while phpunit simply gives out a warning, but doesn't change its exit code.

sanmai:
Well, we should have a similar error on Travis then, shouldn't we?

BackEndTea:
No, the error is in the appveyor set up. It caches php, and its inis for future runs, but tries to apply the xdebug settings every time.

sanmai:
If that's what you're trying to fix in #362, I'll be happy to see that work.

If #362 is really fixing #377, which it appears to, then this PR would be unnecessarily.

sanmai:
Superseded by #362

PR #379, opened 2018-06-05
Extract functionality out of infection command

BackEndTea:
The command class is kind of a god class, this PR removes some functionality from the command and places it into its own sub classes, notably:

* Creation of event subscribers
* Checking if we can generate or have coverage
* Checking if mutation testing run passed its constraints

Meaning this functionality can possibly be unit tested, and it works towards testability of the command class

### DIFF
--- a/src/Console/ConsoleOutput.php
+++ b/src/Console/ConsoleOutput.php
@@ -11,6 +11,7 @@ namespace Infection\Console;
 
 use Infection\Mutant\Exception\MsiCalculationException;
 use Infection\Mutant\MetricsCalculator;
+use Infection\Process\Runner\TestRunConstraintChecker;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Process\Process;
 
@@ -20,6 +21,8 @@ use Symfony\Component\Process\Process;
 final class ConsoleOutput
 {
     const CI_FLAG_ERROR = 'The minimum required %s percentage should be %s%%, but actual is %s%%. Improve your tests!';
+
+    const RUNNING_WITH_DEBUGGER_NOTE = 'You are running Infection with %s enabled.';
 
     /**
      * @var SymfonyStyle
@@ -69,7 +72,7 @@ final class ConsoleOutput
         $this->io->error($lines);
     }
 
-    public function logBadMsiErrorMessage(MetricsCalculator $metricsCalculator, float $minMsi)
+    public function logBadMsiErrorMessage(MetricsCalculator $metricsCalculator, float $minMsi, string $type)
     {
         if (!$minMsi) {
             throw MsiCalculationException::create('min-msi');
@@ -78,25 +81,12 @@ final class ConsoleOutput
         $this->io->error(
             sprintf(
                 self::CI_FLAG_ERROR,
-                'MSI',
+                ($type === TestRunConstraintChecker::MSI_FAILURE ? 'MSI' : 'Covered Code MSI'),
                 $minMsi,
-                $metricsCalculator->getMutationScoreIndicator()
-            )
-        );
-    }
-
-    public function logBadCoveredMsiErrorMessage(MetricsCalculator $metricsCalculator, float $minCoveredMsi)
-    {
-        if (!$minCoveredMsi) {
-            throw MsiCalculationException::create('min-covered-msi');
-        }
-
-        $this->io->error(
-            sprintf(
-                self::CI_FLAG_ERROR,
-                'Covered Code MSI',
-                $minCoveredMsi,
-                $metricsCalculator->getCoveredCodeMutationScoreIndicator()
+                ($type === TestRunConstraintChecker::MSI_FAILURE ?
+                    $metricsCalculator->getMutationScoreIndicator() :
+                    $metricsCalculator->getCoveredCodeMutationScoreIndicator()
+                )
             )
         );
     }
@@ -109,6 +99,19 @@ final class ConsoleOutput
             '- Use phpdbg: phpdbg -qrr infection' . PHP_EOL .
             '- Use --coverage option with path to the existing coverage report' . PHP_EOL .
             '- Use --initial-tests-php-options option with `-d zend_extension=xdebug.so` and/or any extra php parameters',
+        ]);
+    }
+
+    public function logRunningWithDebugger(string $debugger)
+    {
+        $this->io->writeln(sprintf(self::RUNNING_WITH_DEBUGGER_NOTE, $debugger));
+    }
+
+    public function logNotInControlOfExitCodes()
+    {
+        $this->io->warning([
+            'Infection cannot control exit codes and unable to relaunch itself.' . PHP_EOL .
+            'It is your responsibility to disable xdebug/phpdbg unless needed.',
         ]);
     }
 }
--- a/src/Console/InfectionContainer.php
+++ b/src/Console/InfectionContainer.php
@@ -253,7 +253,7 @@ final class InfectionContainer extends Container
             );
         };
 
-        $this['subscriber.creator'] = function () use ($input): SubscriberBuilder {
+        $this['subscriber.builder'] = function () use ($input): SubscriberBuilder {
             return new SubscriberBuilder(
                 $input,
                 $this['metrics'],
--- a/src/Console/InfectionContainer.php
+++ b/src/Console/InfectionContainer.php
@@ -16,9 +16,13 @@ use Infection\Differ\Differ;
 use Infection\EventDispatcher\EventDispatcher;
 use Infection\EventDispatcher\EventDispatcherInterface;
 use Infection\Finder\Locator;
+use Infection\Mutant\MetricsCalculator;
 use Infection\Mutant\MutantCreator;
 use Infection\Mutator\Util\MutatorsGenerator;
+use Infection\Process\Builder\SubscriberBuilder;
+use Infection\Process\Coverage\CoverageChecker;
 use Infection\Process\Runner\Parallel\ParallelProcessRunner;
+use Infection\Process\Runner\TestRunConstraintChecker;
 use Infection\TestFramework\Config\TestFrameworkConfigLocator;
 use Infection\TestFramework\Coverage\CachedTestFileDataProvider;
 use Infection\TestFramework\Coverage\CodeCoverageData;
@@ -173,6 +177,10 @@ final class InfectionContainer extends Container
 
             return (new MutatorsGenerator($mutatorConfig))->generate();
         };
+
+        $this['metrics'] = function (): MetricsCalculator {
+            return new MetricsCalculator();
+        };
     }
 
     private function getInfectionConfig(): InfectionConfig
@@ -227,6 +235,34 @@ final class InfectionContainer extends Container
             return $this['filesystem']->isAbsolutePath($existingCoveragePath)
                 ? $existingCoveragePath
                 : sprintf('%s/%s', getcwd(), $existingCoveragePath);
+        };
+
+        $this['coverage.checker'] = function () use ($input): CoverageChecker {
+            return new CoverageChecker(
+                \strlen(trim($input->getOption('coverage'))) > 0,
+                $input->getOption('initial-tests-php-options')
+            );
+        };
+
+        $this['test.run.constraint.checker'] = function () use ($input): TestRunConstraintChecker {
+            return new TestRunConstraintChecker(
+                $this['metrics'],
+                $input->getOption('ignore-msi-with-no-mutations'),
+                (float) $input->getOption('min-msi'),
+                (float) $input->getOption('min-covered-msi')
+            );
+        };
+
+        $this['subscriber.creator'] = function () use ($input): SubscriberBuilder {
+            return new SubscriberBuilder(
+                $input,
+                $this['metrics'],
+                $this['dispatcher'],
+                $this['diff.colorizer'],
+                $this['infection.config'],
+                $this['filesystem'],
+                $this['tmp.dir']
+            );
         };
     }
 }
--- a/src/Console/InfectionContainer.php
+++ b/src/Console/InfectionContainer.php
@@ -20,7 +20,7 @@ use Infection\Mutant\MetricsCalculator;
 use Infection\Mutant\MutantCreator;
 use Infection\Mutator\Util\MutatorsGenerator;
 use Infection\Process\Builder\SubscriberBuilder;
-use Infection\Process\Coverage\CoverageChecker;
+use Infection\Process\Coverage\CoverageRequirementChecker;
 use Infection\Process\Runner\Parallel\ParallelProcessRunner;
 use Infection\Process\Runner\TestRunConstraintChecker;
 use Infection\TestFramework\Config\TestFrameworkConfigLocator;
@@ -237,8 +237,8 @@ final class InfectionContainer extends Container
                 : sprintf('%s/%s', getcwd(), $existingCoveragePath);
         };
 
-        $this['coverage.checker'] = function () use ($input): CoverageChecker {
-            return new CoverageChecker(
+        $this['coverage.checker'] = function () use ($input): CoverageRequirementChecker {
+            return new CoverageRequirementChecker(
                 \strlen(trim($input->getOption('coverage'))) > 0,
                 $input->getOption('initial-tests-php-options')
             );
--- a/src/Process/Coverage/CoverageChecker.php
+++ b/src/Process/Coverage/CoverageChecker.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Process\Coverage;
+
+use Composer\XdebugHandler\XdebugHandler;
+
+/**
+ * @internal
+ */
+final class CoverageChecker
+{
+    /**
+     * @var bool
+     */
+    private $skipCoverage;
+
+    /**
+     * @var string
+     */
+    private $initialTestPhpOptions;
+
+    public function __construct(bool $skipCoverage, string $initialTestPhpOptions)
+    {
+        $this->skipCoverage = $skipCoverage;
+        $this->initialTestPhpOptions = $initialTestPhpOptions;
+    }
+
+    public function hasDebuggerOrCoverageOption(): bool
+    {
+        return $this->skipCoverage
+            || \PHP_SAPI === 'phpdbg'
+            || \extension_loaded('xdebug')
+            || XdebugHandler::getSkippedVersion()
+            || $this->isXdebugIncludedInInitialTestPhpOptions();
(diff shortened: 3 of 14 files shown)


src/Console/InfectionContainer.php
```diff
@@ -173,6 +177,10 @@ public function __construct(array $values = [])
 
             return (new MutatorsGenerator($mutatorConfig))->generate();
         };
+
+        $this['metrics'] = function (): MetricsCalculator {
```
BackEndTea: `Pimple` caches the values the container provides, meaning that it is created on the first run, and then every time it is requested again it returns the same object.

src/Console/InfectionContainer.php
```diff
@@ -228,5 +236,33 @@ public function buildDynamicDependencies(InputInterface $input)
...
+        $this['test.run.constraint.checker'] = function () use ($input): TestRunConstraintChecker {
+            return new TestRunConstraintChecker(
+                $this['metrics'],
+                $input->getOption('ignore-msi-with-no-mutations'),
+                (float) $input->getOption('min-msi'),
+                (float) $input->getOption('min-covered-msi')
+            );
+        };
+
+        $this['subscriber.creator'] = function () use ($input): SubscriberBuilder {
```
sidz: shouldn't it be `subscriber.builder` instead of `subscriber.creator`?

src/Process/Coverage/CoverageChecker.php
```diff
@@ -0,0 +1,51 @@
...
+declare(strict_types=1);
+
+namespace Infection\Process\Coverage;
+
+use Composer\XdebugHandler\XdebugHandler;
+
+/**
+ * @internal
+ */
+final class CoverageChecker
```
sidz: tbh it's not a coverage checker :) should have a bit different name but can't give you any variant right now

BackEndTea: Yeah youre right, but i can't think of a better name either.

It checks if it is provided with coverage, or if it can generate coverage (it has a debugger), so it was teh best i could come up with.

sanmai: Coverage Existence Checker?

tests/Console/ConsoleOutputTest.php
```diff
@@ -135,4 +125,27 @@ public function test_log_missed_debugger_or_coverage_option()
...
+    {
+        $io = $this->createMock(SymfonyStyle::class);
+        $io->expects($this->once())->method('writeln')
+            ->with('You are running Infection with foo enabled.');
+
+        $consoleOutput = new ConsoleOutput($io);
+        $consoleOutput->logRunningWithDebugger('foo');
+    }
+
+    public function test_log_not_in_controll_of_exit_codes()
```
sanmai: controll <- typo

src/Process/Coverage/CoverageChecker.php
BackEndTea: I don't think thats good either, as the coverage doesn't have to exist, as it is also okay if we can create it during our run.

src/Process/Runner/TestRunConstraintChecker.php
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    public function getMinRequiredValue(): float
+    {
+        return $this->failureType === self::MSI_FAILURE ? $this->minMsi : $this->minCoveredMsi;
+    }
+
+    private function hasBadMsi(MetricsCalculator $metricsCalculator): bool
+    {
+        return $this->minMsi && ($metricsCalculator->getMutationScoreIndicator() < $this->minMsi);
```
borNfreee: we can use `$this->metricsCalculator` and remove the useless parameter

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -119,6 +120,7 @@
         ResultsLoggerTypes::class,
         SummaryFileLogger::class,
         TextFileLogger::class,
+        SubscriberBuilder::class,
```
borNfreee: I guess this list should not be appended

borNfreee:
Please let me know if it's going to be in 0.9.0, so I will wait for it before creating `0.9.0-beta` tag.

If so, please change the base branch from `master` to `0.9`

BackEndTea:
Im at the dpc right now so no way i can change anything the next few days

borNfreee:
@sidz & @sanmai, are you happy with the last changes?

src/Console/ConsoleOutput.php
```diff
@@ -78,25 +81,12 @@ public function logBadMsiErrorMessage(MetricsCalculator $metricsCalculator, floa
         $this->io->error(
             sprintf(
                 self::CI_FLAG_ERROR,
-                'MSI',
+                ($type === TestRunConstraintChecker::MSI_FAILURE ? 'MSI' : 'Covered Code MSI'),
```
sanmai: This a string v. string comparison, `===` is unnecessary here since no type juggling could happen here.

src/Console/ConsoleOutput.php
```diff
@@ -78,25 +81,12 @@ public function logBadMsiErrorMessage(MetricsCalculator $metricsCalculator, floa
...
-            throw MsiCalculationException::create('min-covered-msi');
-        }
-
-        $this->io->error(
-            sprintf(
-                self::CI_FLAG_ERROR,
-                'Covered Code MSI',
-                $minCoveredMsi,
-                $metricsCalculator->getCoveredCodeMutationScoreIndicator()
+                ($type === TestRunConstraintChecker::MSI_FAILURE ?
```
sanmai: Same as above. This actually lowers our MSI score.

src/Process/Coverage/CoverageRequirementChecker.php
```diff
@@ -0,0 +1,51 @@
...
+    public function __construct(bool $skipCoverage, string $initialTestPhpOptions)
+    {
+        $this->skipCoverage = $skipCoverage;
+        $this->initialTestPhpOptions = $initialTestPhpOptions;
+    }
+
+    public function hasDebuggerOrCoverageOption(): bool
+    {
+        return $this->skipCoverage
+            || \PHP_SAPI === 'phpdbg'
```
sanmai: This and two places above also.

(Not sure if this specific line would affect MSI.)

src/Process/Runner/TestRunConstraintChecker.php
```diff
@@ -0,0 +1,100 @@
...
+    ) {
+        $this->metricsCalculator = $metricsCalculator;
+        $this->ignoreMsiWithNoMutations = $ignoreMsiWithNoMutations;
+        $this->minMsi = $minMsi;
+        $this->minCoveredMsi = $minCoveredMsi;
+    }
+
+    public function hasTestRunPassedConstraints(): bool
+    {
+        if ($this->ignoreMsiWithNoMutations && $this->metricsCalculator->getTotalMutantsCount() === 0) {
```
sanmai: `getTotalMutantsCount()` returns strictly `int`s

src/Process/Runner/TestRunConstraintChecker.php
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    public function getErrorType(): string
+    {
+        return $this->failureType;
+    }
+
+    public function getMinRequiredValue(): float
+    {
+        return $this->failureType === self::MSI_FAILURE ? $this->minMsi : $this->minCoveredMsi;
```
sanmai: Here

src/Console/ConsoleOutput.php
theofidry: `===` is unnecessary but I would argue it's safer to use `===` everywhere and use `==` only when strictly necessary

sanmai: I oppose, since any use of `===` is inherent sign of a *bad* design. There only a small number of cases where one really has to use `===`, all other cases can be avoided by setting explicit return types, or by other means like not using equality comparison in the first place. 

Should we get used to bad design and use `===` to work around it, or should we fight it with all our might? I don't see any another choice here.

theofidry: Woa. I must admit I did not expect that PoV. We might have different opinions about what bad & good design is (and that's fine, it's a subject there will always be disagreement on). I would however argue that using `===` is _safer_. It provides less surprises and risks, so rather than trying to figure out when `==` is sufficient, it is easier to use `===`. There is only two trade-offs that I see:

- Sometimes `==` _is_ necessary, notably with value objects
- It requires one more character to type

Which IMO we can cope with for the benefit brought by it. 

>Should we get used to bad design and use === to work around it, or should we fight bad designs with all our might? I don't see any another choice here.

I would argue the bad design here is not about using `===` or `==`, but about PHP itself. But that doesn't really help since we are in PHP and PHP is not gonna change this feature anytime soon if ever. So I would rather be pragmatic and play it safe, especially if it costs no extra efforts.

sanmai: Sure `===` is safer, but *smells* funny all the same. One thing is when you can't change how `strpos` works, another thing where that's you who are building an API method that may return either `null` or `false` or `0`. These kinds of API are bad is so many ways I wouldn't even try to describe. They are bad for commercial products because every bad API is time and money in the end, spend on fighting it. For open source these things are even more important to avoid since almost nobody is getting paid here.

Good news we are not having to deal with those nasty styles of API, we all good in this regard, so why should we even start to prepare for them?

sanmai: When I see `===` anywhere in the code I immediately start questioning myself: why is this here? What for? Against what is it supposed to guard against? Am I missing something? Why would if be needed? What was that other person thinking?.. And here I am, spending several minutes of my unpaid time on an unimportant yet unobvious peculiarity just because someone decided to play it safe. 

This is not the first time I'm arguing against `===`, so before I asked around, and it turns out almost everyone has to spend time researching every use `===` up to a point where they became desensitized to any use of `===` and just think of them like of regular equals comparison. That's where tide turns for worse because if you ignore `===`, you're missing signals, you're not seeing bad API at the right time. Ignoring `===` makes things worse.

theofidry: I can't help to say but I had a chuckle when reading this, because this is _precisely_ my feeling but when facing `==` instead of `===`.

At the core the issue is this one: do you prefer static typing or not. If yes, you default to strict types and when you omit them, it's for a very specific reason and vice-versa if you prefer dynamic typings. Now we can have fun debating which side is better and why are we debating this in PHP which is at its core a hybrid organic language, but it would be derailing a bit of the original discussion.

So at the core I think it's a question of which side we want to pick for this project. I admit and never hid that I tend to prefer strict types and guards like https://github.com/beberlei/assert or https://github.com/webmozart/assert. It's not pretty to fail with an `InvalidArgumentException`, but I find more efficient to bail out early with an error easy to report which helps to fix the issue fast. But I also worked on several loosely typed code bases so I'm fine and respect the other approach as well. I just think we need to agree on which one to pick here :)

@infection/core thoughts?

BackEndTea: Personally i'm in the defaul to `===` and only use `==` when absolutely necessary camp, but i can understand the other side as well.

This does mean that infection users may be part of camp `===` or camp `==`. So i'd like to propose to add a `==` => `===` and `!=` => `!==` mutator as well, move all 4 of them to two new profiles, and disable them both by default.

This means we can cater to people who prefer `===` and those who prefer `==` with minimal configuration.

For our code base, i'm with @theofidry on this one.

sanmai: You can't really test `===` in most cases, that's what Infection shows with a low MSI score. If you really care about MSI score, you should consider using `==`, else just move on. All in all I don't think anyone's annoyance by these mutators is valid reason to disable them. We are about technical quality of software, not about ethical, aren't we?

sanmai: As for static typing, I think that for any API is best to stick to specific return/input types as much as possible. This way a user would never ever need to use `===` just because all types are known beforehand.

@theofidry Is that what you assume under static typing?

BackEndTea: > If you really care about MSI score, you should consider using `==`

I care about MSI in the same way i care about code coverage. I care about it when it helps me understand/track how my test suite is doing. If it becomes lower due to reasons i disagree with, e.g. `==` instead of `===` it becomes a useless metric.

> I don't think anyone's annoyance by these mutators is valid reason to disable them.

When i use a tool, infection or anything else, i want that tool to work for me. I want it to tell me the things i want to know, without too much of a hassle. I completely disagree with changing my code base, simply because my tool tells me i should, especially on opinionated things. (e.g. `==` vs `===`). I'm perfectly fine if those opinionated things align with how the project was already ran/how i want to run the project.

> We are about technical quality of software, not about ethical, aren't we?

Of course, and for mutators like `+` => `-` its obvious what we are testing etc, but `===` => `==` and vice versa, becomes quite opinionated, or ethical if you will.

sanmai: Opinions aren't going well with logical arguments, that's understood. But they are also respectable arguments at times like these, simply because this should be fun and joy, but when your opinions are not respected, it is not anymore.

In this regard, what would you think if we would keep status quo ante? That would mean that I will retract my requests on this PR, and I would not ask anyone else in the future to replace redundant `===`, and I expect that all `===` mutators will be left alone. They're a killer feature for me, just so you would know.

theofidry: >For one because of listed above, for another because PHP is a dynamic language, which is a great power. Best is to embrace and understand these powers, not just blindly follow workarounds, invented by someone else.

But the argument is as valid the other way around: PHP can allow some strictness which is a great power.

>Best is to embrace and understand these powers, not just blindly follow workarounds, invented by someone else. Otherwise put, if you prefer ===, then there should be some reasoning behind this decision.

My reasoning is the following: when I compare `'1'` and `1`, I expect the result to be `false` not `true`. I am not looking for a loose comparison by default, it's not a workaround, this is what I want. Your point is from what I understand whether `===` is _necessary_ to compare it, i.e. if you can safely compare the two values without loose comparison, but IMO even when that's the case it shows the wrong intent. So it's not just a workaround to me: it expresses better the intent I have besides being safe enough to be a sensible default.

>As for static typing, I think that for any API is best to stick to specific return/input types as much as possible. This way a user would never ever need to use === just because all types are known beforehand.

"Strict typing" may be wrongly used my bad. I mean strictness to the code: strict comparison, strict declaration, types whenever possible and in my case I would also add some guards via an assertion library as well. It's not 100% as there is valid cases for duck typing or loose comparisons, but this is rare enough for me to loosen the strictness when necessary.

>All in all I don't think anyone's annoyance by these mutators is valid reason to disable them. We are about technical quality of software, not about ethical, aren't we?

I agree with the message. But mutation testing is expansive so we need to considerate what mutator brings the best value and have a sensible default to have a reasonable running time (i.e. make Infection as fast as possible) and report the least number of false positive. The further one can go the more exigent the profile can be, but if it turns out `===` -> `==` has too many false positives in most code bases I would argue its place is in more exigent profiles, not the default one

sanmai: > when I compare `'1'` and `1`

I see that if you *ever* need to compare these, it's an API problem or just a bug. Bugs should be properly fixed, but not worked around with `===` or else.

BackEndTea: > Opinions aren't going well with logical arguments, that's understood. But they are also respectable arguments at times like these, simply because this should be fun and joy, but when your opinions are not respected, it is not anymore.

My apologies if i came off harsh or disrespectfull, that wasn't my intention. I simply wanted to state that tools should cater to the users needs, and if there isn't a clear 'better' alternative that it should be configurable.

theofidry: >I see that if you ever need to compare these, it's an API problem or just a bug. Bugs should be properly fixed, but not worked around with === or else.

I agree I usually make sure I do not compare those, but this is PHP: it is neither a strongly typed language neither do we have a compiler to warn us about those. Also mistakes can happen, so I would rather stick to `===` which is safer, is more correct to my intent and has absolutely 0 drawback rather than trying to be more clever than I am and figure out each time if it's really necessary or not.

Ocramius: `'0.' == '0.0'` and `==` makes no sense (ever) anymore.

sanmai: That's a nice example where one would actually need to use `===`, but if you need to do it just like than, you're doing it wrong in the first place. It hardly makes any sense to do it just like that be it with `==` or with `===`. Floats must be floats, not strings.

Ocramius: Just don't use `==` please. Like, never ever.

sanmai: @Ocramius would you care to explain your position?

Ocramius: https://github.com/infection/infection/pull/379#discussion_r198437386

sanmai: Very well, so, where exactly in this PR do you see this type of flawed comparison?

Ocramius: I simply exclude these issues upfront, because you just drop more and more time bombs in the codebase, as they become regressions later on. Every `==` removed is an overall betterment that should be welcome (and tbh I don't even understand how this thread got this long by now).

Or you can just install something like `doctrine/coding-standard`, which overall deny the usage of this operator (among multiple flawed PHP features), and stop having these discussions overall. PHPStan strict rules would probably help too.

sanmai: Do you think you could open a PR with these additions, e.g. with PHPStan? Just to end this argument once and for all.

(Actually I'd prefer if PHP-CS-Fixer would do that, but it can't.)

Ocramius: Not sure if I can fit it within my schedule, but I'll keep it in mind 👌

sanmai: If you could point me at the right example, that'd be just about same as if you would make a PR.

BackEndTea:
@sanmai @borNfreee Is this good to merge?

borNfreee:
Well, I have already approved this PR and think it can be merged.

Regarding `==` vs `===`: I must say nobody here gave a good explanation why `count($a) === 0` is better than `count($a) == 0` and as for me - there *can be* cases where we can safely use `==`.

However, 

* since many static code analysis tools (like PHPStan, EA Extended) reports `==` as a bad code
* to be in line with community standards
* to be consistent in our code base
* to not have such discussions in the future anymore

we should go with `===` and merge https://github.com/infection/infection/pull/398 as well, once and for all.

I don't agree that this is the only one correct way (please see the [reduced MSI](https://github.com/infection/infection/pull/398/files#diff-354f30a63fb0907d4ad57269548329e3R37) and [useless](https://github.com/infection/infection/pull/398/files#diff-820567f26d24a995e8bfe293afec8fd6R76) `===` in `count($a)`), but this is a better choice from two we have.

Also, I will merge it to `master` rather than to `0.9` since I spent too much time testing freezed `0.9` branch before release.

borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee and everyone else

PR #383, opened 2018-06-08
Introduce PregMatchMatches mutator

BackEndTea:
This PR:

- [x] Adds new feature, the `PregMatchMatches` mutator.
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/53

```diff

- preg_match('/a/', 'b', $foo);
+ $foo = array();
```

cc: @addono

### DIFF
--- a/src/Mutator/Regex/PregMatchMatches.php
+++ b/src/Mutator/Regex/PregMatchMatches.php
@@ -37,28 +37,27 @@ final class PregMatchMatches extends Mutator
             return false;
         }
 
-        if ($node->name instanceof Node\Expr\Variable ||
+        if (!$node->name instanceof Node\Name ||
             strtolower((string) $node->name) !== 'preg_match') {
             return false;
         }
 
-        return $this->isAllowedByReturnType($node) && \count($node->args) >= 3;
+        return \count($node->args) >= 3 && $this->isAllowedByReturnType($node);
     }
 
     private function isAllowedByReturnType(Node $node): bool
     {
-        if (($parent = $node->getAttribute(ParentConnectorVisitor::PARENT_KEY)) instanceof Node\Stmt\Return_) {
-            $functionScope = $parent->getAttribute(ReflectionVisitor::FUNCTION_SCOPE_KEY);
-            $returnType = $functionScope->getReturnType();
-            if ($returnType instanceof Node\Identifier) {
-                $returnType = $returnType->name;
-            }
-            // no return value specified
-            if (null !== $returnType) {
-                return false;
-            }
+        if (!(($parent = $node->getAttribute(ParentConnectorVisitor::PARENT_KEY)) instanceof Node\Stmt\Return_)) {
+            return true;
         }
 
-        return true;
+        $functionScope = $parent->getAttribute(ReflectionVisitor::FUNCTION_SCOPE_KEY);
+        $returnType = $functionScope->getReturnType();
+
+        if ($returnType instanceof Node\Identifier) {
+            $returnType = $returnType->name;
+        }
+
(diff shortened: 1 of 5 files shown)


src/Mutator/Regex/PregMatchMatches.php
```diff
@@ -0,0 +1,64 @@
...
+
+        return $this->isAllowedByReturnType($node) && \count($node->args) >= 3;
+    }
+
+    private function isAllowedByReturnType(Node $node): bool
+    {
+        if (($parent = $node->getAttribute(ParentConnectorVisitor::PARENT_KEY)) instanceof Node\Stmt\Return_) {
+            $functionScope = $parent->getAttribute(ReflectionVisitor::FUNCTION_SCOPE_KEY);
+            $returnType = $functionScope->getReturnType();
+            if ($returnType instanceof Node\Identifier) {
```
borNfreee: could you please add empty lines before `if` statements like you did it in `mutatesNode()`? IMO, with new lines it's more readable

src/Mutator/Regex/PregMatchMatches.php
```diff
@@ -0,0 +1,64 @@
...
+                $returnType = $returnType->name;
+            }
+            // no return value specified
```
borNfreee: is it a correct comment?

if `null !== $returnType`, then we have a specified return *type*

tests/Mutator/Regex/PregMatchMatchesTest.php
```diff
@@ -0,0 +1,110 @@
...
+            <<<'PHP'
+<?php
+
+namespace RegexMatchMatches;
+
+class ReturnTypes
+{
+    public function foo()
+    {
+        return $matches = array();
```
sanmai: Should you return `0` here but not an empty array, what do you think?

sanmai: Shouldn't it be like 
```diff
- return preg_match('/a/', 'b', $matches);
+ $matches = array();
+ return 0;
```

tests/Mutator/Regex/PregMatchMatchesTest.php
```diff
@@ -0,0 +1,110 @@
...
+PreG_maTch('/a/', 'b', $foo);
+PHP
+            ,
+            <<<'PHP'
+<?php
+
+$foo = array();
+PHP
+        ];
+        yield 'It does not mutate if there are less than 3 arguments' => [
```
sanmai: Shouldn't there be a test case with more than 3 args? preg_match can have as much as 5 args.

tests/Mutator/Regex/PregMatchMatchesTest.php
BackEndTea: This is kinda difficult in with the current php-parser api. As we would flip between statements and expressions.

borNfreee:
@sanmai do you have any other comment before merging?

tests/Mutator/Regex/PregMatchMatchesTest.php
sanmai: Well, if not switching between statements and expressions, then it could be like this:
```php
return (int) $matches = array();
```

borNfreee:
I'm preparing `0.10.0` and this is the last not merged PR. 

Merging  as @sanmai has approved it, however, I wonder if this [comment](https://github.com/infection/infection/pull/383#discussion_r208064660) waits for some reaction.

borNfreee:
thank tou @BackEndTea

BackEndTea:

Issue #382, opened 2018-06-08
Tolerant failures (Recoverable fatal error in PregQuote mutator)


theofidry:
| Question    | Answer
| ------------| ---------------
| Infection version | master (`2499d3d`)

When an unexpected error occurs, for an incorrect assumption in Infection leading to a fatal error like this:

```
PHP Recoverable fatal error:  Object of class PhpParser\Node\Expr\PropertyFetch could not be converted to string in /home/travis/build/humbug/box/vendor/infection/infection/src/Mutator/Regex/PregQuote.php on line 36
```

I think a more contextual error would be helpful, for example giving the problematic file name

BackEndTea:
Do you know exactly what is causing the error?

theofidry:
Well looking at the code I have a rough explanation but the interesting bit that I'm missing is what actual code ended up generating that issue. You can try yourself it happens on humbug/box master, I couldn't point it yet because it is so slow to run

BackEndTea:
Does #385 fix your error @theofidry ?

borNfreee:
ping @theofidry

theofidry:
Will try to check again on Box and open the issue again if the issue is still there

PR #384, opened 2018-06-10
Improve PregQuote check

BackEndTea:
If the node is anything other than the direct function call we can't determine what it is, so we should skip it.

Should fix the error in #382


borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee & @sidz

PR #385, opened 2018-06-11
Improve errors that could be generated while mutating or parsing

### LINKED ISSUE #382: Tolerant failures (Recoverable fatal error in PregQuote mutator)

BackEndTea:
This PR:

- [x] Improves error messages that face the user
- [x] Covered by tests

Fixes #382 

In cases like #382 it isn't clear what file caused the error, so this pr makes sure the user gets told what the error is.
If we encounter a parse error its also good to know what file caused that as well.


tests/Fixtures/Autoloaded/Mutator/ErrorMutator.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Infection\WrongMutator;
+
+
+use Infection\Mutator\Util\Mutator;
+use PhpParser\Node;
+
+class ErrorMutator extends Mutator
+{
+
```
borNfreee: please remove this empty line

tests/Mutator/Util/Exception/MutatorExceptionTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Infection\Tests\Mutator\Util\Exception;
+
+use Infection\Mutator\Arithmetic\Plus;
+use Infection\Mutator\Util\Exception\MutatorException;
+use Infection\Mutator\Util\MutatorConfig;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @internal
```
theofidry: why marking the test cases as internal?

src/Mutant/Exception/ParserException.php
```diff
@@ -0,0 +1,28 @@
...
+declare(strict_types=1);
+
+namespace Infection\Mutant\Exception;
+
+/**
+ * @internal
+ */
+final class ParserException extends \Exception
+{
+    public static function fromInvalidFile(\SplFileInfo $file, \Throwable $original): self
```
theofidry: subjective but I find importing symbols instead of leading `\` more readable

src/Mutator/Util/Exception/MutatorException.php
```diff
@@ -0,0 +1,47 @@
...
+declare(strict_types=1);
+
+namespace Infection\Mutator\Util\Exception;
+
+use Infection\Mutator\Util\Mutator;
+
+/**
+ * @internal
+ */
+final class MutatorException extends \Exception
```
theofidry: why is this class in the `Util` namespace?

src/Mutator/Util/Exception/MutatorException.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    private $mutator;
+
+    public static function errorWhileMutating(Mutator $mutator, \Throwable $previous): self
```
theofidry: maybe `couldNotMutate()` instead?

src/Mutator/Util/Mutator.php
```diff
@@ -30,8 +31,12 @@ public function __construct(MutatorConfig $config)
 
     final public function shouldMutate(Node $node): bool
     {
-        if (!$this->mutatesNode($node)) {
-            return false;
+        try {
```
theofidry: I'm a bit confused: why is it necessary here?

src/Mutant/Generator/MutationsGenerator.php
```diff
@@ -128,8 +130,12 @@ public function generate(bool $onlyCovered, string $filter = ''): array
      */
     private function getMutationsFromFile(SplFileInfo $file, bool $onlyCovered, array $mutators): array
     {
-        /** @var Node[] $initialStatements */
-        $initialStatements = $this->parser->parse($file->getContents());
+        try {
+            /** @var Node[] $initialStatements */
+            $initialStatements = $this->parser->parse($file->getContents());
+        } catch (\Throwable $t) {
+            throw ParserException::fromInvalidFile($file, $t);
```
theofidry: I'm not sure of the flow but if the file here is "just" a PHP file, it's _valid_ for the file to not be a valid PHP file. There can be various reasons for this so instead of bailing out hard like that, I would just ignore the file, eventually log it somewhere

src/Mutant/Generator/MutationsGenerator.php
```diff
@@ -146,7 +152,11 @@ private function getMutationsFromFile(SplFileInfo $file, bool $onlyCovered, arra
         $traverser->addVisitor(new FullyQualifiedClassNameVisitor());
         $traverser->addVisitor(new ReflectionVisitor());
 
-        $traverser->traverse($initialStatements);
+        try {
+            $traverser->traverse($initialStatements);
+        } catch (MutatorException $e) {
```
theofidry: catching a mutator exception to thrown another mutator exception is a bit of a smell to me. Since it's a hard error (i.e. we are stopping the process on that error), I think an `InvalidArgumentException` exception is valid, we just need a comprehensive error message

src/Mutant/Exception/ParserException.php
BackEndTea: That is what we have been doing in other classes as well.

src/Mutator/Util/Exception/MutatorException.php
BackEndTea: It felt like the best place for it, but i'm open to suggestions.

src/Mutant/Generator/MutationsGenerator.php
BackEndTea: In my opinion, if you have an invalid php file, you don't intend infection to make use of that file, so you should add it to an ignore list.

Before, we would error on it as well, the message was just really unclear.

src/Mutator/Util/Mutator.php
BackEndTea: This is meant to catch errors like the one in #382. Infection *should* never error during that step, so if it does we want to let the user know about it, and give the details that user can hand to us.

src/Mutant/Generator/MutationsGenerator.php
BackEndTea: I agree it is a bit of a smell, but i'm not sure what other option would be the best.

I don't agree with changing it to an `InvalidArgumentException`, as this could cause confusion to our end user. I'd rather name it something like `InternalInfectionException`

src/Mutator/Util/Exception/MutatorException.php
theofidry: The `Exception` namespace but without `Util`

src/Mutant/Generator/MutationsGenerator.php
theofidry: Well it's a choice to be made but personally I would much prefer to just have to include `src` and not having to whitelist the couple of invalid files in it than having to whitelist all of them which can be really annoying

src/Mutant/Generator/MutationsGenerator.php
theofidry: IMO just throwing a meaningful exception is enough: we don't need two of them especially since you don't really enrich the message. So either would throw the child one and remove this catch block remove the child one and throw the exception here.

>I don't agree with changing it to an InvalidArgumentException, as this could cause confusion to our end user. I'd rather name it something like InternalInfectionException

I don't know what's confusing: `InvalidArgumentException` is a `LogicException`  which means there is something that requires a fix and not something that should be catch. Having a dedicated exception here brings no benefit and don't make it clearer if the exception message is clear enough

src/Mutator/Util/Exception/MutatorException.php
BackEndTea: I don't want to use the `Exception` namespace within `Mutator` for actual Exceptions. The Throw and Finally_ mutator would be better suited to live in that namespace.

theofidry: Yes indeed, sorry I was thinking of `src/Exception`

src/Mutant/Generator/MutationsGenerator.php
theofidry: @infection/core what's your opinion on this? We have the choice of:

- We keep it as it is: if the file cannot be parsed (e.g. it is an invalid PHP file), we bail out (i.e. interrupt the whole process) on this error: no PHP file cannot be invalid
- We accept that a PHP file cannot be valid so in case of failure, we just skip this file

In my experience, there is several cases where I had to deal with invalid PHP files precisely to test something, so the first option is more annoying to use IMO as it requires the user the whitelist all the invalid files. The second solution however wouldn't require the user any extra action.

Second element which is a bit lost in the collapsed discussion: if we go with the second solution, `ParserException` can be completely removed. If we go with the first solution, I do not thing we should have `ParserException` neither:

- It does not bring anything itself: the user should never catch it, so the only thing needed is a self-explanatory message, not a new exception
- Would having an exception preferred, this exception should be an `\InvalidArgumentException` not an `\Exception`

tests/Fixtures/Autoloaded/Mutator/ErrorMutator.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    public function mutate(Node $node)
+    {
+        return $node;
+    }
+
+    protected function mutatesNode(Node $node): bool
+    {
+        /*
+         * This is intended to cause an error.
```
theofidry: `//` instead?

src/Exception/MutatorException.php
```diff
@@ -0,0 +1,39 @@
...
+ * @internal
+ */
+final class MutatorException extends \Exception
+{
+    /**
+     * @param string $filePath
+     * @param Mutator $mutator
+     * @param \Throwable $previous
+     *
+     * @return MutatorException
```
theofidry: `@return self` here would be more correct, but since in this case there is absolutely no additional information brought by the phpdoc I would just remove it: it is nothing but noise in this case

src/Exception/MutatorException.php
```diff
@@ -0,0 +1,39 @@
...
+declare(strict_types=1);
+
+namespace Infection\Exception;
+
+use Infection\Mutator\Util\Mutator;
+
+/**
+ * @internal
+ */
+final class MutatorException extends \Exception
```
theofidry: What about calling this class `InvalidMutator` instead of `internalErrorWhileTraversing()` `create()`? IMO this is more readable and explicit:

```php
throw InvalidMutator::create();
```

than:

```php
throw MutatorException:: internalErrorWhileTraversing();
```

src/Mutant/Generator/MutationsGenerator.php
borNfreee: I like zero-config style of CLI tools, where you don't need to configure anything to get started.

However, are there any *real* examples of projects where they need to have invalid files?

> I had to deal with invalid PHP files precisely to *test* something

If you had to *test* something, probably it had to be in `./tests` dir, not `./src`? As I understand production-ready project should have only valid files in `src`. Otherwise, how would it work?

----

My 2 cents: 

* IMO, it would be a new enhancement and out of this PR since this one just improves the error message
* if we really need to handle cases for projects with invalid PHP files (but first we should understand how many of them, if any), I'm for ignoring exception and logging/displaying it to the console rather than re-throwing

Also, the original issue (#382) [does not have a clear example](https://github.com/infection/infection/issues/382#issuecomment-396090725) which makes me think it's probably not a real case. Did I miss something?

sanmai: How are that tests passing for such projects while using the very same invalid files?

theofidry: >How are that tests passing for such projects while using the very same invalid files?

It depends a lot on how those invalid files are used. One example could be you expect a certain failure of your application if you try to load it.

https://travis-ci.org/humbug/box/jobs/391429814

>Also, the original issue (#382) does not have a clear example which makes me think it's probably not a real case

It is, I indeed forgot to link the original project which is [Box](https://travis-ci.org/humbug/box/jobs/391429814). Clone and `make tm` will result in this error.

>However, are there any real examples of projects where they need to have invalid files?

I don't have one on top of my head, but I remember facing this issue with PHP-Scoper. That said maybe it was a slightly different case which could be irrelevant here

sanmai: > One example could be you expect a certain failure of your application if you try to load it.

I wonder why would we mutate such files in the first place...

theofidry: >I wonder why would we need to mutate such files in the first place...

We shouldn't, but the question is in terms of usability: do we simply ignore it (not extra action required for the user) or we require the user to whitelist them.

However I think @borNfreee has a point:

>If you had to test something, probably it had to be in ./tests dir, not ./src?

So I guess whitelisting it is fine since it is likely to not affect the users ever.

So that being said, I would come back to the original proposal:

> if we go with the second solution, `ParserException` can be completely removed. If we go with the first solution, I do not thing we should have `ParserException` neither:
>
>- It does not bring anything itself: the user should never catch it, so the only thing needed is a self-explanatory message, not a new exception
>- Would having an exception preferred, this exception should be an `\InvalidArgumentException` not an `\Exception`

BackEndTea: > It does not bring anything itself: the user should never catch it, so the only thing needed is a self-explanatory message, not a new exception

My reasoning for the `ParserException` is as follows:
* Its 'easier' to place the logic for the creation of the exception in a seperate exception, and
not make it the responsibility of the Generator.
* Being presented with a `InvalidArgumentException` may give the user the idea that they gave some wrong input somewehre.

borNfreee: Guys, let's move forward with this simple PR. 

As I understand, we agreed that invalid code should not be in `./src` in most of the cases, in other cases (I guess it is < 1%) users will add an exclusion).

So the open question is just with `ParserException` vs `InvalidArgumentException`, right? 

> Being presented with a `InvalidArgumentException` may give the user the idea that they gave some wrong input somewehre.

I find this reasonable. If anyone still has objections regarding current state of the PR - please comment. If not - let's merge it.

src/Exception/InvalidMutatorException.php
```diff
@@ -0,0 +1,32 @@
...
+ * @internal
+ */
+final class InvalidMutatorException extends \Exception
+{
+    public static function create(string $filePath, Mutator $mutator, \Throwable $previous): self
+    {
+        return new self(
+            sprintf(
+                'Encountered an error with the "%s" mutator in the "%s" file. ' .
+                'This is most likely a bug in infection, so please report this in our issue tracker.',
```
sanmai: `Infection` maybe?

src/Mutant/Exception/ParserException.php
```diff
@@ -0,0 +1,28 @@
...
+    {
+        return new self(
+            sprintf(
+                'Unable to parse file "%s", most likely due to syntax errors.',
+                $file->getRealPath()
+            ),
+            0,
+            $original
+        );
```
sanmai: Should we pass a previous exception here?

src/Mutant/Generator/MutationsGenerator.php
sanmai: Being presented with any kind of exception is bad, no matter how aptly named it is. 

ParserException is certainly better named, and best the enemy of the good, so I'm all in for the new ParserException if it can include the original exception it was caused by. It'll help debugging problems.

borNfreee:
ok, I think we agreed on this one (more or less). Let's get it merged

Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee and everyone else

PR #386, opened 2018-06-12
Fix: No need to update composer twice

localheinz:
This PR

- [x] avoids running `composer self–update` twice on Travis CI

💁‍♂️ For reference, see

* https://travis-ci.org/infection/infection/jobs/390874802#L509
* https://travis-ci.org/infection/infection/jobs/390874802#L544


borNfreee:
Thank you @localheinz

localheinz:

PR #389, opened 2018-06-18
Coalesce mutator

adeptofvoltron:
This PR:

- [x] Adds new feature ...
- [x] Covered by tests

### DIFF
--- a/src/Mutator/Operator/Coalesce.php
+++ b/src/Mutator/Operator/Coalesce.php
@@ -26,7 +26,7 @@ final class Coalesce extends Mutator
      */
     public function mutate(Node $node)
     {
-        return new Node\Expr\BinaryOp\Coalesce(new Node\Expr\Cast\Unset_($node->left), $node->right);
+        return $node->right;
     }
 
     protected function mutatesNode(Node $node): bool
--- a/src/Mutator/Operator/Coalesce.php
+++ b/src/Mutator/Operator/Coalesce.php
@@ -10,7 +10,6 @@ declare(strict_types=1);
 namespace Infection\Mutator\Operator;
 
 use Infection\Mutator\Util\Mutator;
-use Infection\Visitor\ParentConnectorVisitor;
 use PhpParser\Node;
 
 /**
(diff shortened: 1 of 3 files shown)


adeptofvoltron:
Last days travis is so slow...even on magento2 project travis do not work at all :/

src/Mutator/Operator/Coalesce.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Replaces "'someValue' ?? 'otherValue';" with "(unset) 'someValue' ?? 'otherValue'"
+     *
+     * @param Node $node
+     *
+     * @return Node\Expr\BinaryOp\Coalesce
+     */
+    public function mutate(Node $node)
+    {
+        return new Node\Expr\BinaryOp\Coalesce(new Node\Expr\Cast\Unset_($node->left), $node->right);
```
adeptofvoltron: it ought to be
`new Node\Expr\ConstFetch(new Node\Name('null'))`
instead 
`new Node\Expr\Cast\Unset_($node->left)`

Another thing to handle tommorow.

borNfreee:
Travis reasonably fails with the following deprecation notice on PHP 7.2+:

```
Deprecated: The (unset) cast is deprecated
```

Why `(unset)` has been deprecated? Answer is here: https://wiki.php.net/rfc/deprecations_php_7_2

> The `(unset)` cast casts a value to `null`. This means that `(unset)` expr is simply an expression that always returns `null` and has no other side effects. Apart from being useless, this behavior is also confusing, as many people reasonably assume that `(unset) $a` will behave similarly to `unset($a)`, while in truth it does no such thing.

So, this is not a correct mutator. Moreover, it is the same as

```diff
- $a = $b ?? $c;
+ $a = null ?? $c;
```

which in its turn is the same as

```diff
- $a = $b ?? $c;
+ $a = $c;
```

src/Mutator/Operator/Coalesce.php
sanmai: Why not simply `$node->right`?

adeptofvoltron: good point.

sanmai:
It seems to me that AppVeyor build failures are unrelated.

adeptofvoltron:
added. Still same troubles with AppVeyor

adeptofvoltron:
added. Still some issues with AppVeyor...maybe they did something wrong during:
image

borNfreee:
Thank you @adeptofvoltron

Issue #222, opened 2018-03-14
[RFC] Improve CLI report


theofidry:
After playing a bit for a demo I found a few things that were either bugging me or that I think are a bit confusing.

This discussion was on Slack initially but I think putting it here is more appropriate. So right now here's the current CLI report:

<details>
<summary>Current report</summary>

```
You are running Infection with xdebug enabled.
    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/

Running initial test suite...

PHPUnit version: 7.0.2

    8 [============================] < 1 sec

Generate mutants...

Processing source code files: 2/2
Creating mutated files and processes: 6/6
.: killed, M: escaped, S: uncovered, E: fatal error, T: timed out

M.....                                               (6 / 6)

6 mutations were generated:
       5 mutants were killed
       0 mutants were not covered by tests
       1 covered mutants were not detected
       0 errors were encountered
       0 time outs were encountered

Metrics:
         Mutation Score Indicator (MSI): 83%
         Mutation Code Coverage: 100%
         Covered Code MSI: 83%

Please note that some mutants will inevitably be harmless (i.e. false positives).
```
</details>

So here's the list of my nitpicks from a potential a first-time user:

- What is used to run the initial test suite? How long does it take? There is nothing to help to debug here
- I left the command run for some time and looked away. When I came back I had the result but no idea how long it took
- Where is the log file generated?
- I only know mutant `killed` and `escaped`: what is `not covered by tests`, `not detected`, `errors` and `time outs`?
- I know `Mutation Score`, why do I have 3 of them?

Also although I think it would be cool at some point to be able to say "execute _all_ the tests" or specific tests for non-covered mutants, I think it's too early for this extension point and for now way too expansive. As such I would just suggest to drop the notion of "not covered" and completely ignore non-covered code for now.

But because complaining is easy, here's a suggestion for improvements:

<details>
<summary>Suggested report</summary>

```
# You are running Infection with xdebug enabled.: removed as displayed in the initial run block
    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
[...]

PR #390, opened 2018-06-26
Show time and memory info to the console output

borNfreee:
This PR:

- [x] Adds time and memory usage to the console: `Time: 4s. Memory: 20.00MB`
- [x] Covered by tests
- [x] ~blocked by https://github.com/infection/infection/pull/379~

<details>
<summary>Full output with time and memory output</summary>

```bash
You are running Infection with phpdbg enabled.
    ____      ____          __  _

Running initial test suite...

PHPUnit version: 5.5.7

   24 [============================]  1 sec

Generate mutants...

Processing source code files: 50/50
Creating mutated files and processes: 301/301

SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS   ( 50 / 301)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSM.MSSSSSSSSSSSS   (100 / 301)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS.SSSS   (150 / 301)
SS..SSSSSSSSSSSSSSSSSSM.MM.SSSSSSSSSSSSSSSSMM...M.   (200 / 301)
M..M...M..SSSSSSSSSS.......M...........SSSS.MSSSSS   (250 / 301)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSMM   (300 / 301)
M                                                    (301 / 301)

301 mutations were generated:
      36 mutants were killed
     249 mutants were not covered by tests
      16 covered mutants were not detected
       0 errors were encountered
       0 time outs were encountered

Metrics:
         Mutation Score Indicator (MSI): 12%
         Mutation Code Coverage: 17%
         Covered Code MSI: 69%

Dashboard report has not been sent: it is not a Travis CI

Time: 4s. Memory: 20.00MB
```

</details>

------

Originally proposed in https://github.com/infection/infection/issues/222 by @theofidry


tests/Performance/Time/TimerIsAlreadyStartedExceptionTest.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+final class TimerIsAlreadyStartedExceptionTest extends TestCase
+{
+    public function test_creation()
+    {
+        $exception = TimerIsAlreadyStartedException::create();
+
+        $this->assertInstanceOf(\Exception::class, $exception);
+        $this->assertInstanceOf(TimerIsAlreadyStartedException::class, $exception);
+        $this->assertSame($exception->getMessage(), 'Timer can not be started again without stopping.');
```
BackEndTea: The two arguments should be swapped here.

src/Performance/Memory/MemoryFormatter.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Infection\Performance\Memory;
+
+/**
+ * @internal
+ */
+final class MemoryFormatter
+{
+    const BYTES_IN_MEGABYTE = 1024 * 1024;
+
+    public function toHumanReadableString(float $bytes): string
```
BackEndTea: I wonder if we can extract some kind of `Formatter` interface here.

theofidry:
👍 on the feature

Do we really need +19 classes for it thought? While it's not the prettiest, the [Box one](https://github.com/humbug/box/blob/master/src/Console/Command/Compile.php#L839-L846) is only 7 lines long

BackEndTea:
Shouldn't the target of this be `0.10.0`? Since its a new feature, rather than a bug-fix.

borNfreee:
> Do we really need +19 classes for it thought? While it's not the prettiest, the Box one is only 7 lines long

While I also had this feeling during development, I still think it is worth it. SOLID, you know ;)

The big number of files is because each new class requires an additional test-file (otherwise tests fail). So the actual number of new classes that do the job `2x` less.

Also, with this implementation we can easily test it. And it continues event-based architecture we already have, which IMO is also a good thing to decouple classes.

src/Performance/Memory/MemoryFormatter.php
theofidry: What about having only one formatter here? Even better having it in `PerformanceLoggerSubscriber`. Unless we have a need for it elsewhere, it doesn't really make sense to separate them and would reduce the number of classes needed

src/Performance/Memory/MemoryUsageProvider.php
```diff
@@ -0,0 +1,21 @@
...
+namespace Infection\Performance\Memory;
+
+/**
+ * @internal
+ */
+final class MemoryUsageProvider
+{
+    public function getPeakUsage(): int
+    {
+        return memory_get_peak_usage(true);
```
theofidry: For me this is a _big_ smell, not of bad architecture or design, but of over-engineering. I'm not arguing this is not SOLID, but IMO this is overkill: are we ever going to need another memory usage provider? The only case I can think of is having a `FixedMemoryUsageProvider` which could be used for some tests of the Infection command, but in this case I would also introduce a `MemoryProvider` interface

src/Performance/Time/TimerIsAlreadyStartedException.php
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Performance\Time;
+
+/**
+ * @internal
+ */
+final class TimerIsAlreadyStartedException extends \Exception
```
theofidry: I would rather extend `\LogicException`: this would be a wrong usage not an exception you would like to catch

src/Performance/Time/Timer.php
```diff
@@ -0,0 +1,42 @@
...
+final class Timer
+{
+    /**
+     * @var float|null
+     */
+    private $microTime;
+
+    public function start()
+    {
+        if (null !== $this->microTime) {
```
theofidry: What about using a guard with an assert library here instead? Here's an example with https://github.com/webmozart/assert:

```php
Assert::null($this->microTime, 'Cannot start a timer that has already been started');
```

Since those are exceptions that are not meant to be catch, I think an assertion library is more concise: in case of error the message will be as comprehensible and easy to spot & report and would remove the need to introduce 2 new exception classes for it

borNfreee: agree. This is a subject to discuss for @infection/core whether to include this new lib or not.

I have no objections since this lib has no dependencies and really useful, so I have added it to the PR. If anyone agains - let me know so we can revert it.

src/Performance/Memory/MemoryUsageProvider.php
borNfreee: I see your point. While I don't think it is a smell, I agree we can simplify this particular code.

src/Performance/Memory/MemoryFormatter.php
borNfreee: > I wonder if we can extract some kind of Formatter interface here.

TBH I see no reason to extract the interface, because:

1. it is assumed to have only 1 implementation here
2. classes are final and internal, they are not supposed to be replaced by reimplementing the abstraction (interface) by users that use Infection
3. IMO memory and time formatter has nothing in common. It is a coincidence that they have method with the same name. `Formatter` does not suite here because it is too general thing, also memory and time formatters has different signature (`int`/`float` for parameter)
4. as stated before - it would be overkill here, keeping in mind that classes are internal and has 1 implementation

borNfreee: > What about having only one formatter here? Even better having it in PerformanceLoggerSubscriber. 

Disagree. `PerformanceLoggerSubscriber` must not know how to format memory and time strings, this is violation of SRP from my POV. After moving this functionality to subscriber, it will be impossible to unit test it without making the methods `public`

borNfreee:
Rebased on `master`, addressed/answered to comment. Could you please review it again @theofidry & @BackEndTea ?

borNfreee:
Rebased again. Merging it due to no review activity

PR #391, opened 2018-06-27
Add counterparts to identical mutator and remove them from default

BackEndTea:
This PR:

- [x] Removes the identical mutators from the default list and adds their equal counterparts.
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/45

See also: https://github.com/infection/infection/pull/379#discussion_r198386216


sanmai:
Why are you proposing to remove them from the default set? Is there a measured harm from them being present on the list? I don't understand.

It's okay to remove useless mutators which are most likely to be caught by tests (like, if 99 mutations of 100 were caught), but these mutations are the opposite. If they're uncaught by the tests, they show that you have weak or inefficient tests, and that's what Infection supposed to do.

src/Mutator/Util/MutatorProfile.php
```diff
@@ -63,13 +65,11 @@
     const BOOLEAN = [
         Mutator\Boolean\ArrayItem::class,
         Mutator\Boolean\FalseValue::class,
-        Mutator\Boolean\IdenticalEqual::class,
```
sanmai: These are particulary useful mutators. E.g. it is very easy to forget to test an example below on the inclusion of a substring at the beginning of the string:
```
false !== strpos($string, $substring)
```
When this example is mutated, then it could skip under the radar if a test coverage is insufficient:
```
false != strpos($string, $substring)
```
Otherwise put, if this mutation lives on, your tests are not good enough.

Removal of these mutators from the default profile is unwarranted.

tests/Mutator/Boolean/EqualIdenticalTest.php
```diff
@@ -0,0 +1,50 @@
...
+        $this->doTest($input, $expected);
+    }
+
+    public function provideMutationCases(): \Generator
+    {
+        yield 'It mutates equal operator into identical operator' => [
+            <<<'PHP'
+<?php
+
+$a == $b;
```
sanmai: Exactly under which circumstances these mutations are expected to fly under the radar? 

* If `$a == $b` is false, so is `$a === $b` false too. 
* If `$a != $b` is true, so is `$a !== $b` is true.
* If `$a != $b` is false, so is `$a !== $b` false too.
* If `$a == $b` is true, so is `$a === $b` true too, in most circumstances. Moreover, you can't possibly write a test for this mutation.

These new mutations bring nothing to the table unless proven otherwise.

src/Mutator/Util/MutatorProfile.php
BackEndTea: This comes down to how you write your code. If you default to `==`, and only use `===` when strictly needed, than this mutator is rather useful.

However, if you default to `===` and only use `==` when strictly needed, 90+% of these mutations can't be caught.

For a lot of projects, the default is `===`. For this reason PHP-CS-Fixer has the `strict_comparison` fixer, that changes `==` to `===`, and coding standards like the doctrine one disallow usage of `==`.

tests/Mutator/Boolean/EqualIdenticalTest.php
BackEndTea: This mutator is intended for those who use `===` at all times, and only use `==` when strictly needed. Which is also why its not a default mutator, as its not usefull for all our users, and may produce a bunch of false positives if you have a different code style for your project.

src/Mutator/Util/MutatorProfile.php
sanmai: This mutator was created specifically to cover practical and useful cases like above and similar. It has nothing to do with coding styles.

If you think we should be using `===` at all time, best course of action is to propose adding `strict_comparison` fixer. Removal of an essential mutator solves nothing.

tests/Mutator/Boolean/EqualIdenticalTest.php
sanmai: Well, what kind of faulty *tests* do you think these mutators are supposed to find?

BackEndTea: Lets say your application runs some other cli programs, and needs to check it return types. Because some of them have their own adapters for php you use that to run these programs.

But as there is no standard for it, some return the int 0 on success, while others return the sting 0 on success.

So you'd have a check somewhere that looks like this:
```php
if ($cliAdapter->getExitCode() != 0) {
    throw new NotSuccessfulException();
}
```

Now, if your tests only check here with the int 0, changing the `!=` to `!==` won't be caught. Which cold be improved by also adding a check against the string 0.

BackEndTea: The reason it should not be in the default set is for those who use `==` and `!=` almost exclusively. For them this mutator will give them a ton of mutators that can't be caught. 

Which is the same reason why i propose to make the `===` => `==` mutator not part of default, as people who use `===` exclusively will get a ton of mutators that can't be caught.

sanmai: Where an `==` &rarr; `===` mutator (and another) has a distinct and very specific purpose to work against very well known test deficiencies, which I tried to demonstrate in #394, and even if these mutators may cause some grief at times, I don't see the grief only as a valid reason for removal. I myself found some quite embarrassing testing bugs with an `==` &rarr; `===` mutator in my own code, and that's the whole purpose of mutation testing, by the way.

On the other hand these counterpart mutators are really quite vague, which I'm not a big fan of tbh. If you, yourself, do not wish to sponsor them, I don't see why we should have them, like, at all.

sanmai:
@BackEndTea are you going to pursue this PR any further? If not, I'm ready to take over.

BackEndTea:
Feel free to take over

BackEndTea:
Gonna close this for now

sanmai:
Took liberty to reopen this in the light of #489.

BackEndTea:
Rebased on master, and ready for review

borNfreee:
Thank you @BackEndTea

BackEndTea:

PR #393, opened 2018-06-28
Test against nightly php build

BackEndTea:
This PR:

- [x] Tests our build against the nightly php build, but allows it to fail, as the build might not be stable.

This can help us prepare for future php releases, and pick up on deprecations etc.


borNfreee:
Thank you @BackEndTea

PR #398, opened 2018-07-01
Enable strict_comparison cs fixer rule

BackEndTea:
```
$ ./php-cs-fixer-v2.phar describe strict_comparison
Description of strict_comparison rule.
Comparisons should be strict.

Fixer applying this rule is risky.
Changing comparisons to strict might change code behavior.

Fixing examples:
 * Example #1.
   ---------- begin diff ----------
   --- Original
   +++ New
   @@ -1,2 +1,2 @@
    <?php
   -$a = 1== $b;
   +$a = 1=== $b;
   
   ----------- end diff -----------
```

### DIFF
--- a/tests/Console/E2ETest.php
+++ b/tests/Console/E2ETest.php
@@ -40,7 +40,7 @@ final class E2ETest extends TestCase
         // Without overcommit this test fails with `proc_open(): fork failed - Cannot allocate memory`
         if (strpos(PHP_OS, 'Linux') === 0 &&
             is_readable('/proc/sys/vm/overcommit_memory') &&
-            file_get_contents('/proc/sys/vm/overcommit_memory') == 2) {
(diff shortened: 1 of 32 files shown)


borNfreee:
Thank you @BackEndTea

tests/Console/E2ETest.php
```diff
@@ -40,7 +40,7 @@ protected function setUp()
         // Without overcommit this test fails with `proc_open(): fork failed - Cannot allocate memory`
         if (strpos(PHP_OS, 'Linux') === 0 &&
             is_readable('/proc/sys/vm/overcommit_memory') &&
-            file_get_contents('/proc/sys/vm/overcommit_memory') == 2) {
+            file_get_contents('/proc/sys/vm/overcommit_memory') === 2) {
```
sanmai: This won't work, you have to cast to int here because `file_get_contents` always returns string.

PR #402, opened 2018-07-01
Enhancement: Configure blank_line_before_statement fixer

localheinz:
This PR

- [x] configures the `blank_line_before_statement` fixer
- [x] runs `make cs-fix`

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.12.1#usage:

>**blank_line_before_statement** [`@Symfony`]
>
>An empty line feed must precede any configured statement.
>
>Configuration options:
>
>* `statements` (a subset of `['break', 'case', 'continue', 'declare', 'default', 'die', 'do', 'exit', 'for', 'foreach', 'goto', 'if', 'include', 'include_once', 'require', 'require_once', 'return', 'switch', 'throw', 'try', 'while', 'yield']`): list of statements which must be preceded by an empty line; defaults to `['break', 'continue', 'declare', 'return', 'throw', 'try']`


borNfreee:
Thank you @localheinz

localheinz:
Thank you, @BackEndTea, @borNfreee and @sanmai!

PR #404, opened 2018-07-05
Fix issue when the quiet mode show some message

### LINKED ISSUE #403: Quiet mode is as quiet as one can expect it to be
| Question    | Answer
| Infection version | 9.0

Quiet mode is as quiet as it should be. Steps to reproduce:

1. Run `vendor/bin/infection --quiet`
2. Observe being printed out: `You are running Infection with xdebug enabled.`

sidz:
This PR fixes #403 issue and fixes the quiet mode

### DIFF
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -98,7 +98,7 @@ ASCII;
 
     private function logRunningWithDebugger(InputInterface $input)
     {
-        if (!$input->hasParameterOption(array('--quiet', '-q'), true)) {
+        if (!$input->hasParameterOption(['--quiet', '-q'], true)) {
             if (\PHP_SAPI === 'phpdbg') {
                 $this->consoleOutput->logRunningWithDebugger(\PHP_SAPI);
             } elseif (\extension_loaded('xdebug')) {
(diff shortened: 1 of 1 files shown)


BackEndTea:
What if we move this message into the InfectionCommand. That way Symfony will handle whether or not it should output, based on `--quiet`

sidz:
we can't as we show this message before _xdebugHandler_ will do its dirty job :)

src/Console/Application.php
```diff
@@ -100,6 +96,17 @@ public function run(InputInterface $input = null, OutputInterface $output = null
         return parent::run($input, $output);
     }
 
+    private function logRunningWithDebugger(InputInterface $input)
+    {
+        if (!$input->hasParameterOption(['--quiet', '-q'], true)) {
```
sanmai: Shouldn't it be like this instead, what do you think?
```php
    public function logRunningWithDebugger(string $debugger)
    {
        // --- $this->io->writeln(sprintf(self::RUNNING_WITH_DEBUGGER_NOTE, $debugger));
        $this->io->writeln(sprintf(self::RUNNING_WITH_DEBUGGER_NOTE, $debugger), OutputInterface::VERBOSITY_QUIET);
    }
```

sidz: @sanmai yeah sure. Thought about it. 

but symfony will do it a bit later here https://github.com/symfony/console/blob/master/Application.php#L845

borNfreee:
Thank you @sidz

PR #405, opened 2018-07-06
BadgeLogger now reports exact errors

sanmai:
This PR:

- [x] Makes BadgeLogger report detailed explanation for an error
- [x] To avoid confusion, BadgeLogger now will also look for `STRYKER_DASHBOARD_API_KEY` since [that's what the official manual for the Stryker Dashboard calls for](https://stryker-mutator.io/blog/2018-02-08/get-your-mutation-score-badge-now.html).
- [x] BadgeApiClient now check for HTTP code on success
- [x] Covered by tests

### DIFF
--- a/src/Logger/BadgeLogger.php
+++ b/src/Logger/BadgeLogger.php
@@ -19,6 +19,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 final class BadgeLogger implements MutationTestingResultsLogger
 {
     const ENV_INFECTION_BADGE_API_KEY = 'INFECTION_BADGE_API_KEY';
+    const ENV_STRYKER_DASHBOARD_API_KEY = 'STRYKER_DASHBOARD_API_KEY';
 
     /**
      * @var BadgeApiClient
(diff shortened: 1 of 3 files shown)


src/Logger/BadgeLogger.php
```diff
@@ -66,26 +67,60 @@ public function log()
...
         }
+
+        $apiKey = getenv(self::ENV_INFECTION_BADGE_API_KEY);
+
+        /*
+         * Original Stryker Dashboard manual warrants a different API key:
+         * fall back to theirs if our isn't present
+         */
+        if ($apiKey === false) {
+            $apiKey = getenv(self::ENV_STRYKER_DASHBOARD_API_KEY);
```
borNfreee: In this case, please update the documentation. Now it says:

`Make sure to use INFECTION_BADGE_API_KEY environment variable name instead of STRYKER_DASHBOARD_API_KEY.`

Both should be allowed.

sanmai: Done: https://github.com/infection/site/pull/49

sanmai:
@borNfreee I've added some extra checks to the BadgeApiClient, please see if they are to your liking

borNfreee:
still LGTM

borNfreee:
Thank you @sanmai

PR #406, opened 2018-07-06
Fix a bug from PR #398

sanmai:
This PR:

- [x] Fixes a typing bug from PR #398

I wonder why isn't there a static analyzer to catch these bugs.

### DIFF
--- a/tests/Console/E2ETest.php
+++ b/tests/Console/E2ETest.php
@@ -40,7 +40,7 @@ final class E2ETest extends TestCase
         // Without overcommit this test fails with `proc_open(): fork failed - Cannot allocate memory`
         if (strpos(PHP_OS, 'Linux') === 0 &&
             is_readable('/proc/sys/vm/overcommit_memory') &&
-            intval(file_get_contents('/proc/sys/vm/overcommit_memory')) === 2) {
+            (int) file_get_contents('/proc/sys/vm/overcommit_memory') === 2) {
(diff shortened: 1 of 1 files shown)


tests/Console/E2ETest.php
```diff
@@ -40,7 +40,7 @@ protected function setUp()
         // Without overcommit this test fails with `proc_open(): fork failed - Cannot allocate memory`
         if (strpos(PHP_OS, 'Linux') === 0 &&
             is_readable('/proc/sys/vm/overcommit_memory') &&
-            file_get_contents('/proc/sys/vm/overcommit_memory') === 2) {
+            intval(file_get_contents('/proc/sys/vm/overcommit_memory')) === 2) {
```
borNfreee: `(int) X` is more performant that `intval(X)` (reported by `EA Extended Ultimate`)

intval

sanmai: That part is very far from being a hot code, but whatever.

BackEndTea: We have the `modernize_types_casting ` php-cs fixer rule ( part of `@Symfony:risky`) which should auto fix this.

borNfreee:
Thank you @sanmai

tests/Console/E2ETest.php
borNfreee: @sanmai we are already using it as part of `@Symfony:risky`

PR #408, opened 2018-07-10
PHPStan upgraded to 0.10.1, new errors and warnings fixed

sanmai:
All these new `assert()`s should probably start throwing exceptions at some point.

### DIFF
--- a/devTools/phpstan-src.neon
+++ b/devTools/phpstan-src.neon
@@ -2,4 +2,3 @@ parameters:
     ignoreErrors:
         - '#Access to an undefined property PhpParser\\Node(.*?)#'
         - '#Call to an undefined method PhpParser\\Node(.*?)#'
-        - ':Parameter #1 $key of function array_key_exists expects int|string, float|int given.:'
--- a/devTools/phpstan-src.neon
+++ b/devTools/phpstan-src.neon
@@ -2,6 +2,4 @@ parameters:
     ignoreErrors:
         - '#Access to an undefined property PhpParser\\Node(.*?)#'
         - '#Call to an undefined method PhpParser\\Node(.*?)#'
-        - '#Cannot access property \$name on string.#'
-        - '#Accessing property \$expr on possibly null value of type PhpParser\\Node(.*?)#'
-        - '#expects PhpParser\\Node\\Identifier, string given.#'
+        - ':Parameter #1 $key of function array_key_exists expects int|string, float|int given.:'
--- a/devTools/phpstan-tests.neon
+++ b/devTools/phpstan-tests.neon
@@ -1,3 +1,5 @@
 parameters:
     excludes_analyse:
         - %currentWorkingDirectory%/tests/Fixtures/*
+    ignoreErrors:
+        - '#Call to an undefined method Mockery#'
--- a/src/Http/BadgeApiClient.php
+++ b/src/Http/BadgeApiClient.php
@@ -45,6 +45,8 @@ class BadgeApiClient
 
         $ch = curl_init();
 
+        \assert(\is_resource($ch));
+
         curl_setopt($ch, CURLOPT_URL, self::STRYKER_DASHBOARD_API_URL);
         curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
         curl_setopt($ch, CURLOPT_POST, true);
@@ -61,7 +63,9 @@ class BadgeApiClient
             $this->output->writeln(sprintf('Stryker dashboard returned an unexpected response code: %s', $responseCode));
         }
 
-        $this->output->writeln('Dashboard response:', OutputInterface::VERBOSITY_VERBOSE);
-        $this->output->writeln($response, OutputInterface::VERBOSITY_VERBOSE);
+        if (\is_string($response)) {
+            $this->output->writeln('Dashboard response:', OutputInterface::VERBOSITY_VERBOSE);
+            $this->output->writeln($response, OutputInterface::VERBOSITY_VERBOSE);
+        }
     }
 }
--- a/src/Mutant/Generator/MutationsGenerator.php
+++ b/src/Mutant/Generator/MutationsGenerator.php
@@ -107,7 +107,7 @@ final class MutationsGenerator
         $this->eventDispatcher->dispatch(new MutationGeneratingStarted($files->count()));
 
         foreach ($files as $file) {
-            if (!$onlyCovered || ($onlyCovered && $this->hasTests($file))) {
+            if (!$onlyCovered || $this->hasTests($file)) {
                 $allFilesMutations[] = $this->getMutationsFromFile($file, $onlyCovered, $mutators);
             }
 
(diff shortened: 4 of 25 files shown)


devTools/phpstan-tests.neon
```diff
@@ -1,3 +1,5 @@
 parameters:
     excludes_analyse:
         - %currentWorkingDirectory%/tests/Fixtures/*
+    ignoreErrors:
+        - '#Call to an undefined method Mockery#'
```
sanmai: Too many of these, all bogus, easier to ignore.

src/Http/BadgeApiClient.php
```diff
@@ -61,7 +63,9 @@ public function sendReport(
             $this->output->writeln(sprintf('Stryker dashboard returned an unexpected response code: %s', $responseCode));
         }
 
-        $this->output->writeln('Dashboard response:', OutputInterface::VERBOSITY_VERBOSE);
-        $this->output->writeln($response, OutputInterface::VERBOSITY_VERBOSE);
+        if (\is_string($response)) {
+            $this->output->writeln('Dashboard response:', OutputInterface::VERBOSITY_VERBOSE);
+            $this->output->writeln($response, OutputInterface::VERBOSITY_VERBOSE);
+        }
```
sanmai: `$response` could be a boolean, no point in outputting a response then

src/Mutant/Generator/MutationsGenerator.php
```diff
@@ -107,7 +107,7 @@ public function generate(bool $onlyCovered, string $filter = ''): array
         $this->eventDispatcher->dispatch(new MutationGeneratingStarted($files->count()));
 
         foreach ($files as $file) {
-            if (!$onlyCovered || ($onlyCovered && $this->hasTests($file))) {
+            if (!$onlyCovered || $this->hasTests($file)) {
                 $allFilesMutations[] = $this->getMutationsFromFile($file, $onlyCovered, $mutators);
             }
```
sanmai: Equivalent code. [See this discussion for details.](https://github.com/vimeo/psalm/issues/735)

src/Mutant/MutantCreator.php
```diff
@@ -70,8 +70,11 @@ public function create(MutationInterface $mutation, CodeCoverageData $codeCovera
 
     private function createMutatedCode(MutationInterface $mutation, string $mutatedFilePath): string
     {
-        if (file_exists($mutatedFilePath)) {
-            return file_get_contents($mutatedFilePath);
+        if (is_readable($mutatedFilePath)) {
+            $mutatedCode = file_get_contents($mutatedFilePath);
+            \assert(\is_string($mutatedCode));
```
sanmai: `file_get_contents()` can return `false` even if we check for `is_readable()`

src/Mutator/Util/AbstractValueToNullReturnValue.php
```diff
@@ -19,10 +19,10 @@
 {
     protected function isNullReturnValueAllowed(Node $node): bool
     {
-        /** @var \PhpParser\Node\Stmt\Function_ $functionScope */
-        $functionScope = $node->getAttribute(ReflectionVisitor::FUNCTION_SCOPE_KEY);
+        /** @var \PhpParser\Node\Stmt\Function_|null $functionScope */
+        $functionScope = $node->getAttribute(ReflectionVisitor::FUNCTION_SCOPE_KEY, null);
 
-        if (!$functionScope) {
+        if (null === $functionScope) {
```
sanmai: `getAttribute()` received the default `null` as the second argument

src/StreamWrapper/IncludeInterceptor.php
```diff
@@ -95,6 +95,8 @@ public function stream_open($path, $mode, $options)
 
     public function dir_closedir()
     {
+        \assert(\is_resource($this->fp));
```
sanmai: Another way to handle this would be declaring `$fp` as `@var resource` instead of `@var resource|bool`, but it could actually be `false` so best is to use actual asserts.

src/TestFramework/PhpSpec/Config/Builder/InitialConfigBuilder.php
```diff
@@ -45,7 +45,7 @@ public function build(): string
 
         $yamlConfiguration = new InitialYamlConfiguration(
             $this->tempDirectory,
-            Yaml::parse(file_get_contents($this->originalYamlConfigPath)),
+            Yaml::parseFile($this->originalYamlConfigPath),
```
sanmai: `Yaml::parseFile` throws up if it can't read a file.

src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
```diff
@@ -180,6 +180,8 @@ function (array $a, array $b) {
             $testSuite->appendChild($file);
         }
 
+        \assert($nodeToAppendTestSuite instanceof \DOMNode);
+
```
sanmai: Are we really checking for validity of a `phpunit.xml`? I'm not sure.

src/TestFramework/PhpUnit/Coverage/PhpUnitTestFileDataProvider.php
```diff
@@ -24,7 +24,7 @@
     private $jUnitFilePath;
 
     /**
-     * @var \DOMXPath
+     * @var \DOMXPath|null
```
sanmai: [See this issue.](https://github.com/phpstan/phpstan/issues/1192)

devTools/phpstan-src.neon
```diff
@@ -2,6 +2,4 @@ parameters:
     ignoreErrors:
         - '#Access to an undefined property PhpParser\\Node(.*?)#'
         - '#Call to an undefined method PhpParser\\Node(.*?)#'
-        - '#Cannot access property \$name on string.#'
-        - '#Accessing property \$expr on possibly null value of type PhpParser\\Node(.*?)#'
-        - '#expects PhpParser\\Node\\Identifier, string given.#'
+        - ':Parameter #1 $key of function array_key_exists expects int|string, float|int given.:'
```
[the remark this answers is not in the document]
sanmai: Well, apparently you have to add this typehint before a variable, not after, what I did.

theofidry:
I would replace the usage of those `assert()` by `Webmozart\Assert::*()` otherwise I'm +1 for the guards

borNfreee:
As I know, `Webmozart\Assert::*()` requires an additional plugin (https://github.com/phpstan/phpstan-webmozart-assert) wich we can't use as a PHAR (correct me if I'm wrong). 

So we will have to use phpstan as a composer dependency and install plugins as dependencies as whell.

sanmai:
Standard `assert()` calls are benign by default in PHP 7+, therefore you can put there anything you want like:
```php
assert($this->veryExpensiveCheck());
```
I doubt this is true about `Webmozart\Assert::*()`. So they're quire different even if named similarly.

Better yet, most if not all of subjects of this PR are of very rare occurrence. If they happen to be not so rare, if would be trivial to covert them to actual exceptions. That would be the right thing to do.

borNfreee:
Let's move forward with this one.

Next time we need an `assert()`, we will replace it with the assertion library and add phpstan plugins.

Let's do it step by step and don't stuck with frozen PRs.

Thank you @sanmai

PR #411, opened 2018-07-15
Explicitly require the DOM extension

BackEndTea:
This PR:

* Explicitly requires the dom extension

A few things to note:

* We only require this when running phpunit, we don't use any DOM related things for phpspec as far as i know.
* We also require the curl extension, but only when dealing with the badge api.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,6 +58,7 @@
     },
     "require": {
         "php": "^7.0",
+        "ext-dom": "*",
         "composer/xdebug-handler": "^1.1",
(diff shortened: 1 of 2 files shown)


sanmai:
> The lock file is not up to date with the latest changes in composer.json, it is recommended that you run `composer update`.

borNfreee:
thank you @BackEndTea

BackEndTea:
Thank you @borNfreee @sanmai

PR #423, opened 2018-07-30
Fix exclude to excludes for generated config file

### LINKED ISSUE #422: Generated infection config uses old exclude key rather than excludes
| Question    | Answer
| Infection version | 0.9.0
| Test Framework version | PHPUnit/ 6.5
| PHP version |7.2
| Platform    | MacOs
| Github Repo | -

Discovered this while setting up a new repo.

Reproduce:
1. run `vendor/bin/infection` without a config file present
2. add an exclude option
3. Check configuration file

I expected the key to be `excludes`, but it is `exclude`, which no longer works.

BackEndTea:
This PR:

- [x] changes the generated config to use `excludes` instead of `exclude`
- [x] Fixes #422

### DIFF
--- a/src/Command/ConfigureCommand.php
+++ b/src/Command/ConfigureCommand.php
@@ -141,7 +141,7 @@ final class ConfigureCommand extends BaseCommand
         }
 
         if ($excludedDirs) {
-            $configObject->source->exclude = $excludedDirs;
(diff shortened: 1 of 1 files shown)


borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee @sidz

PR #424, opened 2018-07-30
Sort logs

### LINKED ISSUE #421: Mutations appear in a 'random' order in the log
Currently mutations appear in the order they were created (or killed?) in in logs. This doesn't make a lot of sense, so it might be better to at least group them per file, preferably going from top to bottom.

BackEndTea:
Sort by filename, with a fall back to line number.
Also add tests for that.
This PR:

- [x] Sorts the logs based on filename, and then by line number.
- [x] Covered by tests
- [x] Fixes #421


borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee @sidz

PR #430, opened 2018-08-02
Use latest version of dependencies & test on travis against lowest

BackEndTea:
This PR:

- [x] Updates dependencies to their latest version for the phar
- [x] Tests travis against our lowest dependencies as well
- [x] Updates minimum symfony versions to 3.4

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,7 @@
     },
     "config": {
         "platform": {
-            "php": "7.1.19"
+            "php": "7.1.3"
         },
         "sort-packages": true
     },
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,7 @@
     },
     "config": {
         "platform": {
-            "php": "7.1"
+            "php": "7.1.19"
         },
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -49,7 +49,7 @@
     },
     "config": {
         "platform": {
-            "php": "7.1"
+            "php": "7.1.19"
```
sanmai: Why do we need this?

This is a real bummer, e.g. because I'm still on earlier version at places, where it is very cumbersome to upgrade.

BackEndTea: We need at least 7.1.3 for symfony 4

sanmai: I'm fine with 7.1.3 if you ask, but I think we'll be fine with just `7.1` in this field.

BackEndTea: It wouldnt update symfony for me with 7.1 in the config. Ill update it to 7.1.3

BackEndTea:
Closing this in favor of the other PR

PR #431, opened 2018-08-03
Round MSI down to the smallest integer, instead of nearest

### LINKED ISSUE #426: 100% MSI reported even if a small number of uncovered mutations is found
Infection may report 100% MSI and won't fail on `--min-msi` if there was a sufficiently large number of mutations. [A quote from our Slack channel:](https://symfony-devs.slack.com/archives/C78M7A4BY/p1533189199000112)

> I guess it’s something like 99.99%  and this value is rounded to 100 https://github.com/infection/infection/blob/master/src/Mutant/MetricsCalculator.php#L137, so it passes --min-msi

sanmai:
Round metrics down to the smallest integer, instead of nearest, which is the worst case for a metric.

Fixes #426

### DIFF
--- a/src/Mutant/MetricsCalculator.php
+++ b/src/Mutant/MetricsCalculator.php
@@ -134,7 +134,7 @@ class MetricsCalculator
         $defeatedTotal = $this->killedCount + $this->timedOutCount + $this->errorCount;
 
         if ($this->totalMutantsCount) {
-            $detectionRateAll = round(100 * ($defeatedTotal / $this->totalMutantsCount));
+            $detectionRateAll = floor(100 * ($defeatedTotal / $this->totalMutantsCount));
         }
 
         return $detectionRateAll;
--- a/src/Mutant/MetricsCalculator.php
+++ b/src/Mutant/MetricsCalculator.php
@@ -134,7 +134,7 @@ class MetricsCalculator
         $defeatedTotal = $this->killedCount + $this->timedOutCount + $this->errorCount;
 
         if ($this->totalMutantsCount) {
-            $detectionRateAll = floor(100 * ($defeatedTotal / $this->totalMutantsCount));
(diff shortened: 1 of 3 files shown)


src/Mutant/MetricsCalculator.php
```diff
@@ -134,7 +134,7 @@ public function getMutationScoreIndicator(): float
         $defeatedTotal = $this->killedCount + $this->timedOutCount + $this->errorCount;
 
         if ($this->totalMutantsCount) {
-            $detectionRateAll = round(100 * ($defeatedTotal / $this->totalMutantsCount));
+            $detectionRateAll = floor(100 * ($defeatedTotal / $this->totalMutantsCount));
```
borNfreee: I believe at least Covered MSI (line 167) should also be updated since we used it in `--min-coverered-msi` - the same bug will be here.

And what about `getCoverageRate()`? Probably having the same logic there would be consistent?

sanmai: Yes, I think that's going to be more proper to account for the worst case for all metrics.

tests/Mutant/MetricsCalculatorTest.php
```diff
@@ -41,42 +41,40 @@ public function test_it_shows_zero_values_by_default(): void
         $this->assertSame(0.0, $calculator->getCoveredCodeMutationScoreIndicator());
     }
 
+    private function addMutantProcess(MetricsCalculator $calculator, int $resultCode, int $count = 1): void
```
BackEndTea: Should we move this below the test method?

sanmai: It is in reading order and in close proximity. I don't think it would be any better if it would be below.

borNfreee:
Thank you @sanmai

Issue #134, opened 2018-01-20
Add infection to PHIVE (PHAR Installation and Verification Environment)


shochdoerfer:
I'd love to be able to install infection via [PHIVE](https://phar.io). Steps needed to enable PHIVE support:
1. Deploy a phive.xml file to your webserver, see [here](https://github.com/phar-io/phar.io/blob/master/data/repository.xsd) for the XSD or [here](https://github.com/sebastianbergmann/phar-site-generator) for a project generating such a file
2. Add the phive.xml url to list of [PHIVE repositories](https://github.com/phar-io/phar.io/blob/master/data/repositories.xml)

theofidry:
You can already do it by doing `phive install infection/infection`. Registering a `phive.xml` to the PHIVE repositories is for aliases only :)

It could be nice to have the `infection` alias though to only do `phive install infection`

shochdoerfer:
I was actually referring to the aliases, sorry for making this not clear in the issue description.

pxgamer:
Interesting, when I try installing through Phive it fails to find a `.asc` or `.sig` file.

sidz:
@borNfreee will we add infection to PHIVE?

borNfreee:
Yes, we will. As I understand [from this tutorial](https://phar.io/howto/sign-and-upload-to-github.html), we need to upload `.asc` file to make it compatible with GitHub releases.

I don't want to do it manually each time I release infection, cause we need also to generate `infection.phar.pubkey` using open ssl private key and this GPG `.asc` file. I would like to automate all this process, but currently I'm out of time.

We will do it later with automating the whole process, involving Travis API to store private keys.

theofidry:
@borNfreee when you'll want to tackle this you can follow the [article that I wrote](https://medium.com/@tfidry/create-and-deploy-secure-phars-c5572f10b4dd) detailing how to automate it (this is what I used for PHP-Scoper). There's also a link on how to deal with GPG signatures instead of OpenSSL

djmattyg007:
Is there anything that can be done to help this process?

borNfreee:
> Is there anything that can be done to help this process?

That's not a priority one, at least for me. If you guys want to speed this process up, you can start helping with the whole build process, that I described in [this comment](https://github.com/infection/infection/issues/134#issuecomment-364693747)

[...]

PR #436, opened 2018-08-05
Add GPG signing for PHAR

borNfreee:
This PR:

- [x] Removes built-in OpenSSL PHAR singing
- [x] Adds GnuPG (GPG) signing
- [x] Doc PR: https://github.com/infection/site/pull/54

Fixes https://github.com/infection/infection/issues/134

After this one gets merged, I will create a test release (e.g. `0.10.0-rc.1`) and then will create a PR for adding an alias to Phive.


sidz:
> After this one gets merged, I will create a test release (e.g. 0.10.0-rc.1) and then will create a PR for adding an alias to Phive.

Probably we need to merge it into 0.9 branch as well. What do you think?

borNfreee:
> Probably we need to merge it into 0.9 branch as well. What do you think?

could you please give some details about why we need this on 0.9? I'm going to release 0.10.0 this week

-------

Thanks for the review

sidz:
@borNfreee thought that we will support 0.9 till the EOL php 7.0 :)

Issue #435, opened 2018-08-05
Infection 0.9 isnt compatible with symfony/console <3.4


BackEndTea:
Spotted here: https://github.com/Roave/DoctrineSimpleCache/pull/27#issuecomment-410521494

Not sure what is causing the issue, but best course of action would be to simply require 3.4. Which i already targeted for `0.10` in #430.

Should we also release a new 0.9 with that minimum requirement?

sidz:
I'd say that we shouldn't drop the support of 3.2 for Infection 0.9. 

build with fail: https://travis-ci.org/Roave/DoctrineSimpleCache/jobs/412320302

will double check on the next week

borNfreee:
Fixed and released as 0.9.3

PR #437, opened 2018-08-07
Fix symfony/console 3.2 compatibility

sidz:
This PR fixes https://github.com/infection/infection/issues/435

### DIFF
--- a/src/Console/InfectionContainer.php
+++ b/src/Console/InfectionContainer.php
@@ -218,7 +218,11 @@ final class InfectionContainer extends Container
         };
 
         $this['coverage.path'] = function () use ($input): string {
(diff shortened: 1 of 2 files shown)


BackEndTea:
How does this work when using the coverage option with symfony/console 3.2?

sidz:
@BackEndTea I didn't investigate to much but the answer somewhere here: https://github.com/symfony/console/commit/0572b233ec2b40d3e51d9e4f021c0da7d9c2113d

BackEndTea:
> Seems to be in line with upstream changes, although it is somewhat surprising BC break with unclear reasons.

Reading the comment i think it has something to do with behaviour that changed in php 7

borNfreee:
Thank you @sidz

PR #439, opened 2018-08-10
PregMatchMatches mutator made to 'return' ints

sanmai:
This PR:

- [x] Updates PregMatchMatches with a simplified operation
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/55

Instead of a return type check we can use a simple cast:
```php
return (int) $matches = array();
```
Therefore, our replacement expression will always result in an int of zero. [This is a totally valid cast in PHP.](https://3v4l.org/S2BIj) 

On a side note, `preg_match()` [may return `false` at times](http://php.net/manual/en/function.preg-match.php#refsect1-function.preg-match-returnvalues), but I don't think anyone in their right mind would assume such return type alone. Outside of that when implemented like above we can remove all return type checks.


borNfreee:
Thank you @sanmai !

PR #440, opened 2018-08-10
Improve compatibility with framework based applications

patrickfunke:
When using a custom bootstrap/autoloader, this file is directly required by the InfectionCommand. Applications to be tested may also use the variable names $config and $container for their own configuration and dependency injection. This will cause either the application or Infection to use the wrong config/container. With a unique name, this collision is avoided.

This PR:

- [x] Renames config and container to avoid naming collisions.

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -170,29 +170,29 @@ final class InfectionCommand extends BaseCommand
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $container = $this->getContainer();
+        $infectionContainer = $this->getContainer();
 
-        if (!$container->get('coverage.checker')->hasDebuggerOrCoverageOption()) {
+        if (!$infectionContainer->get('coverage.checker')->hasDebuggerOrCoverageOption()) {
             $this->consoleOutput->logMissedDebuggerOrCoverageOption();
 
             return 1;
         }
 
-        $config = $container->get('infection.config');
+        $infectionConfig = $infectionContainer->get('infection.config');
 
-        $this->includeUserBootstrap($config);
+        $this->includeUserBootstrap($infectionConfig);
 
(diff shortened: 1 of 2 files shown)


src/Command/InfectionCommand.php
```diff
@@ -170,29 +170,29 @@ protected function configure(): void
 
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $container = $this->getContainer();
+        $infectionContainer = $this->getContainer();
```
theofidry: I don't think the renaming is necessary here

src/Command/InfectionCommand.php
```diff
@@ -259,9 +259,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
         return $statusCode;
     }
 
-    private function includeUserBootstrap(InfectionConfig $config): void
+    private function includeUserBootstrap(InfectionConfig $infectionConfig): void
     {
-        $bootstrap = $config->getBootstrap();
+        $bootstrap = $infectionConfig->getBootstrap();
```
theofidry: maybe instead of changing names here the bootstrap file should be included in its own scope:

```php
$bootstrap = $config->getBootstrap();

(function ($_file) {
    // Include the file in a separate scope to avoid variable shadowing
    require_once $_file;
})($bootstrap);
```

sanmai: Or you can `unset($config)` before the inclusion.

theofidry: I think a closure will be safer to use: you don't have any change to do if you add a variable or parameter; but yeah ultimately both works so let's pick one and be done with it :)

borNfreee:
thank you @patrickfunke

patrickfunke:
Please excuse that I didn't react - I was on vacation. I definitely like the solution, thanks for fixing the issue so quickly!

borNfreee:
No worries. I'm so glad you didn't check github notifications during vacation and didn't do any programming. This is something I should also try when I have one :)

Issue #442, opened 2018-08-11
Infection may apply a memory limit to the normal php ini


BackEndTea:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.9.3 (`infection.phar --version`)
| Test Framework version | PHPUnit 6.5
| PHP version | 7.2.7
| Platform    Ubuntu
| Github Repo | this one

When running infection with `INFECTION_ALLOW_XDEBUG=1`, we use the 'default' php ini, as we don't use xdebug handler. When you run cli from an account that has write access to the php ini, a memory limit will be set

sanmai:
> When you run cli from an account that has write access to the php ini

One shouldn't do that in the first place.

BackEndTea:
> One shouldn't do that in the first place.

I must admit that I ran this as root in order to trigger this behavior. It is, however, something that could happen if a user is not careful.

I was also able to reproduce this on my mac, without needing to sudo

PR #443, opened 2018-08-11
Do not change system php ini with xdebug allow

BackEndTea:
This PR:

- [x] Makes sure we don't change the system php ini.

Fixes: #442

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -274,8 +274,9 @@ final class InfectionCommand extends BaseCommand
 
     private function applyMemoryLimitFromPhpUnitProcess(Process $process, MemoryUsageAware $adapter): void
     {
-        if (\PHP_SAPI === 'phpdbg') {
+        if (\PHP_SAPI === 'phpdbg' || \extension_loaded('xdebug')) {
             // Under phpdbg we're using a system php.ini, can't add a memory limit there
+            // If xdebug is enabled, that means we aren't using xdebug handler, and are also using the system php.ini
             return;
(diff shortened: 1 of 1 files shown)


src/Command/InfectionCommand.php
```diff
@@ -274,8 +274,9 @@ private function includeUserBootstrap(InfectionConfig $config): void
 
     private function applyMemoryLimitFromPhpUnitProcess(Process $process, MemoryUsageAware $adapter): void
     {
-        if (\PHP_SAPI === 'phpdbg') {
+        if (\PHP_SAPI === 'phpdbg' || \extension_loaded('xdebug')) {
```
sidz: So with this change we will change memory limit only when we use `--coverage` option. Iis that what @sanmai wanted to achieve when he developed that part?

sanmai: Normally you're not allowed to change the system php.ini (there are checks to taht). Normally means not running Infection under a root user.

sanmai:
Normally you're not allowed to change the system php.ini (there are checks to that). Normally means not running Infection under a root user. So this naturally makes me think about under which circumstances is this going on?

src/Command/InfectionCommand.php
BackEndTea: If xdebug handler is active, then xdebug won't be active at this point of the application.

But as per https://github.com/infection/infection/pull/443#issuecomment-412418591, there may be better ways of doind this check

BackEndTea:
This is now part of #448

PR #445, opened 2018-08-11
Do not require dev packages on Deploy stage

borNfreee:
Seems like `humbug/box` removes dev dependencies automatically because I don't see them in the built PHAR, however, I see no reason to install dev dependencies on Deploy stage.


borNfreee:
Thank you

PR #446, opened 2018-08-11
Fix symfony dependency issues

BackEndTea:
Fixed some dependency related issues:
* Yaml::parseFile() wasn't present untill 3.4
> Solution: use parse() with file_get_contents()
* Process didn't accept array untill 3.3
> Solution: Update to 3.3+ as 4.0+ needs array
* Console had a bug issue in with empty defaults prior to 3.3.6  & 3.2.13
> Solution: Update minimum to 3.13 and add a conflict rune for between 3.3.0 and 3.3.5
See: https://github.com/symfony/symfony/pull/22244

Also:
* Update dependencies for phar
* Test against low dependencies on travis.


BackEndTea:
Running into an issue with composer, opened up an issue here: https://github.com/composer/composer/issues/7542

BackEndTea:
@infection/core I'd like a review on this.

composer.json
```diff
@@ -67,16 +67,16 @@
         "padraic/phar-updater": "^1.0.4",
         "pimple/pimple": "^3.2",
         "sebastian/diff": "^1.4 || ^2.0 || ^3.0",
-        "symfony/console": "^3.2 || ^4.0",
-        "symfony/filesystem": "^3.2 || ^4.0",
+        "symfony/console": "~3.2.13|| ^3.3.6 || ^4.0",
```
sidz: why do you bump up other symfony dependencies to 3.3 but `symfony/console` stay with 3.2? IMO it will be better to lock all dependencies with one version

sanmai: What if we get them all up to 3.4?

BackEndTea:
@infection/core Whats your opinion? Can this be merged after a rebase, or does anything need to be changed?

borNfreee:
I will try to look into this PR as soon as possible

BackEndTea:
> What are the issues?

* Yaml::parseFile() wasn't present untill 3.4Yaml::parseFile() wasn't present untill 3.4
* Process didn't accept array until 3.3Process didn't accept array until 3.3
* Console had a bug issue in with empty defaults prior to 3.3.6 & 3.2.13Console had a bug issue in with empty defaults prior to 3.3.6 & 3.2.13
* Filesystem::appendFile wasn't present until 3.3

> The build is broken on Travis after rebase (in the worst case we can still merge it as it won't be broken in master.

This appears to be due to the fact that infection command doesn't have good MSI, and it being the only class affected.

src/Command/InfectionCommand.php
```diff
@@ -104,7 +104,7 @@ protected function configure(): void
                 'j',
                 InputOption::VALUE_REQUIRED,
                 'Threads count',
-                1
+                '1'
```
BackEndTea: The input hints on this method changed in one of the later version of symfony to no longer allow ints.

BackEndTea:
Thank you @borNfreee, @sanmai & @sidz

Issue #432, opened 2018-08-04
Family of rounding functions mutators


sanmai:
PR #431 shows that we can't tell if a test is bound to account all rounding errors or not. 

For this job would really help having function call mutators, that would change `floor` to `round`, and `round` to `floor`, to `floor` to `ceil`, `ceil` to `round`, and any other combinations.

E.g. this mutation being caught would mean that tests are really testing that we're rounding down at all times:
```diff
- return floor($this->a / $this->b);
+ return round($this->a / $this->b);
```

Should be very straightforward to make.

deleugpn:
Started working on this today. If nobody comes up with a PR for this, I'll try to open one in the next few days. 
I already learned a little about how the tests and the mutators work, however since there's no example of mutators replacing one `Node\Name\FuncCall` with another, I'm still trying to figure out how PhpParser works.

One thing I might get early feedback though: should there be one mutator for each combination of changes? I'm currently building one `Infection\Mutator\Rounding\Round` that is capable of swapping both `floor` and `ceil` with `round`. That way, the end result is probably 3 new mutators.

borNfreee:
Sorry, missclicked from mobile phone.

> One thing I might get early feedback though: should there be one mutator for each combination of changes?

Oops, I have already started to do it as one mutator that `yield`s many mutations in branch [`yield-mutations`](https://github.com/infection/infection/tree/yield-mutations?files=1) 

I always happy when people contribute to open source, so if you want, you can continue there in that branch, I will push it as soon as I will be at home.

The main idea was not to create N mutators, but create only one, and this will be possible only in that branch. It will be a good example how powerful interface with Generators is.

deleugpn:
Hey @borNfreee I'm not sure I understood what you meant by one mutator. I was thinking that we would have 3 new Mutator classes (Round, Float and Ceil), each one capable of mutating the function call into the other two.
I used the preg_match test case and I noticed it uses Generators for the test code, not sure if that's what you meant. I opened PR https://github.com/infection/infection/pull/447 to help you guide me towards the right path if it makes it easier.
I'm not working on it right now, but I'm expecting to have a couple of hours tomorrow to try and make the mutations work.

borNfreee:
Ok, let me explain it with more details.

I'm working on a new feature that will allow each Mutator produce N mutations instead of just one. It's better to see it with the code example:

instead of this:

```php
# Arithmetic/Plus.php

public function mutate(Node $node)
{
[...]

PR #447, opened 2018-08-11
[WIP] [Feature] Introducing Round/Float/Ceil Mutator

deleugpn:
This PR:

- [x] Adds new feature ...
- [ ] Covered by tests

Closes https://github.com/infection/infection/issues/432

This is my first contribution to infection and I have no experience with PHP-Parser whatsoever. In order to help out in any conversation on issue #432 I decided to open this PR early so I can get any guidelines in order to quickly make progress on the development of this mutation.
I hope to have some time to continue working on this PR in the following couple of days and hopefully it will be ready for final review before next weekend.

Thanks.

### DIFF
--- a/tests/Mutator/Rounding/RoundTest.php
+++ b/tests/Mutator/Rounding/RoundTest.php
@@ -73,5 +73,115 @@ $float = '2.5';
 round($float);
 PHP
         ];
+
+        yield 'It mutates variables on floor to round' => [
+            <<<'PHP'
+<?php
+
+$float = '2.5';
+ceil($float);
+PHP
+            ,
+            <<<'PHP'
+<?php
+
+$float = '2.5';
+round($float);
+PHP
+        ];
+
+        yield 'It does not mutate other functions' => [
+            <<<'PHP'
+<?php
+
+strtolower('lower');
+PHP
+        ];
+
+        yield 'It mutates \ceil to round' => [
+            <<<'PHP'
+<?php
+
+\ceil(9.9);
+PHP
+            ,
+            <<<'PHP'
+<?php
+
+round(9.9);
+PHP
+        ];
+
+        yield 'It mutates \floor to round' => [
+            <<<'PHP'
+<?php
+
+\floor(9.9);
+PHP
+            ,
+            <<<'PHP'
+<?php
+
+round(9.9);
+PHP
+        ];
+
+        yield 'It mutates when ceil is being assigned to a variable' => [
+            <<<'PHP'
+<?php
+
+$result = ceil(9.9);
+PHP
+            ,
+            <<<'PHP'
+<?php
+
+$result = round(9.9);
+PHP
+        ];
+
+        yield 'It mutates when floor is being assigned to a variable' => [
+            <<<'PHP'
+<?php
+
+$result = floor(9.9);
+PHP
+            ,
+            <<<'PHP'
+<?php
+
+$result = round(9.9);
+PHP
+        ];
+
+        yield 'It mutates when the function is incorrectly cased' => [
+            <<<'PHP'
+<?php
+
+$result = CeiL(9.9);
+PHP
+            ,
+            <<<'PHP'
+<?php
+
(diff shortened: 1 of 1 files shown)


BackEndTea:
Thank you for your contribution.

Could you run 
`make cs-fix` to fix any code style issues. It also looks like you have an IDE that doesn't use the `.editorconfig` styles. If you use PHP storm you can find the plugin [here
](https://plugins.jetbrains.com/plugin/7294-editorconfig)

src/Mutator/Rounding/Round.php
```diff
@@ -0,0 +1,41 @@
...
+    public function mutate(Node $node)
+    {
+        $node->name->parts = ['round'];
+
+        return $node;
+    }
+
+    protected function mutatesNode(Node $node): bool
+    {
+        return $node->name instanceof Node\Name && (
```
BackEndTea: Not all nodes may have a name property. If we check that the current node is a function call we can be more sure that we are mutation the correct nodes, like so : ```$node instanceof Node\Expr\FuncCall &&```, what do you think?

src/Mutator/Rounding/Round.php
```diff
@@ -0,0 +1,41 @@
...
+ * @internal
+ */
+final class Round extends Mutator
+{
+    /**
+     * Replaces "floor();" and "ceil();" with "round();"
+     *
+     * @param Node|Node\Name $node
+     *
+     * @return Node\Expr\Cast\Int_
```
BackEndTea: This method doesn't return an `Node\Expr\Cast\Int_`, how about changing it to just `Node` for now, or `Node|Node\Expr\FuncCall`

BackEndTea:
How about adding a few tests that make sure of the following cases:

* Other functions are not mutated (any function call will do)
* It doesn't crash when a variable is called as a function, or a direct string e.g.
```php
$a= 'ceil';
$a(3.5);
('ceil')(3.5);
```
* It mutates when it is called with a backslash e.g. `\round($a);`
* It mutates when the output is assigned to a variable
* It mutates when part of an if/while/for (any will do) statement.

src/Mutator/Util/MutatorProfile.php
```diff
@@ -59,6 +59,7 @@
         Mutator\Arithmetic\PowEqual::class,
         Mutator\Arithmetic\ShiftLeft::class,
         Mutator\Arithmetic\ShiftRight::class,
+        Mutator\Rounding\ChangeRoundFunctionCall::class,
```
BackEndTea: This needs to be updated to the current class name.

Since this is a new group(profile), please add it to a new const. See other groups as an example

tests/Mutator/Rounding/RoundTest.php
```diff
@@ -0,0 +1,76 @@
...
+
+declare(strict_types=1);
+
+namespace Infection\Tests\Mutator\Rounding;
+
+use Infection\Tests\Mutator\AbstractMutatorTestCase;
+
+/**
+ * @internal
+ * @group f
```
BackEndTea: This line probably needs to be removed

borNfreee:
Thanks for your work! 

I left a comment about these mutators that are being developed in a separate branch as part of new `Mutator`'s interface with `yield`ing instead of `return`ing the value.

Please read https://github.com/infection/infection/issues/432#issuecomment-412299662

src/Mutator/Rounding/Round.php
```diff
@@ -0,0 +1,43 @@
...
+        return new Node\Expr\FuncCall(
+            new Node\Name('round'),
+            $node->args,
+            $node->getAttributes()
+        );
+    }
+
+    protected function mutatesNode(Node $node): bool
+    {
+        return $node instanceof Node\Expr\FuncCall && (
```
BackEndTea: If we apply the $node->name instanceof Node\Name here as well, then we can be sure that it is a name. (Which makes sure that we don't error when dealing with variable functions

PR #448, opened 2018-08-11
Extract two methods out of infection command

BackEndTea:
This PR:

- [x] Extracts memory usage limiting and parsing of mutators out of the infection command 
- [x] Covered by tests

Fixes: #442 

Mostly copy pasting things into their own classes, and adding tests.


tests/Performance/Limiter/MemoryLimiterTest.php
```diff
@@ -0,0 +1,90 @@
...
+        $adapter = $this->createMock(AbstractTestFrameworkAdapter::class);
+        $adapter->expects($this->never())->method($this->anything());
+
+        $memoryLimiter = new MemoryLimiter($fs, 'foo/bar');
+        $memoryLimiter->applyMemoryLimitFromProcess($process, $adapter);
+    }
+
+    public function test_it_applies_memory_limit_if_possible(): void
+    {
+        $filename = sys_get_temp_dir() . '/fake-ini' . microtime() . '.ini';
```
sidz: Could we please prepare a workspace for files which we will use as part of this test. For example `infection-test` or whatever else. And on `tearDown` we will need to drop that folder.

an example you can find in the MutationConfigBuilderTest.

Why it's should be done?
Because our users (and me) may mount temporary folder on RAM :)

src/Console/InfectionContainer.php
```diff
@@ -196,6 +198,10 @@ public function __construct(array $values = [])
         $this['memory.formatter'] = function (): MemoryFormatter {
             return new MemoryFormatter();
         };
+
+        $this['memory.limit.applier'] = function (): MemoryLimiter {
+            return new MemoryLimiter($this['filesystem'], \php_ini_loaded_file());
```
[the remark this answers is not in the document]
BackEndTea: By passing it into the class we make this class testable, because otherwise there is no proper way to test this

tests/Mutant/Generator/MutationsGeneratorTest.php
```diff
@@ -173,8 +160,8 @@ public function test_it_throws_correct_exception_when_mutator_is_invalid(): void
 
     private function createMutationGenerator(
         CodeCoverageData $codeCoverageDataMock,
-        array $whitelistedMutatorNames = [],
-        MutatorConfig $mutatorConfig = null,
+        ?string $whitelistedMutatorNames = null,
```
borNfreee: since it's not an array, please rename to singular

borNfreee:
thank you @BackEndTea

PR #450, opened 2018-08-13
[RFC] Yield mutations

borNfreee:
This PR:

- [x] Updates interface of Mutator class to return multiple mutations instead of one
- [x] Adds `RoundingFamily` mutator as an example of new interface usage. Fixes https://github.com/infection/infection/issues/432
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/57 / https://github.com/infection/site/pull/56 (but needs to be updated)
- [x] Test with different projects to understand if anything is broken

This is a feature that allows each Mutator producing N mutations instead of just one. It's better to see it with the code example:

instead of this:

```php
# Arithmetic/Plus.php

public function mutate(Node $node)
{
    return new Node\Expr\BinaryOp\Minus($node->left, $node->right, $node->getAttributes());
}
```

we will have this:

```php
# Arithmetic/Plus.php

public function mutate(Node $node): \Generator
{
    yield new Node\Expr\BinaryOp\Minus($node->left, $node->right, $node->getAttributes());
}

```

Note the `\Generator` return type and `yield` keyword instead of `return`. 

#### Why is it needed?

This is needed exactly for mutators like `RoundingFamily`, where **one** Mutator can produce **many** mutations instead of creating N separate Mutator classes.

Also, this will allow us to produce as many mutations as we want for one `Node` instance: example here is a mutator that will remove groups from Regular Expression:

```diff
- /several(.*)groups(.*)/
+ /severalgroups(.*)/
+ /several(.*)groups/ 
```

Note that there are 2 mutations from only 1 node (string). This is impossible to do with the current (in master branch) Mutators implementation.

----

`RoundingFamily` mutator produces 2 mutations for one `Node` when `Traverser` leaves the node, in particular:

1. it mutates `floor()` to `ceil()` AND `round()`
2. it mutates  `ceil()` to `floor()` AND `round()`
3. it mutates `round()` to `ceil()` AND `floor()`

All this logic is inside one class thanks to `yield`ing many mutations in a `foreach` loop.

This PR is also targeted to stabilize `Mutator` interface since this is one of the future extension points of Infection and much more powerful than the current implementation.


src/Mutation.php
```diff
@@ -57,14 +57,26 @@
...
         string $originalFilePath,
         array $originalFileAst,
         Mutator $mutator,
         array $attributes,
         string $mutatedNodeClass,
         bool $isOnFunctionSignature,
-        bool $isCoveredByTest
+        bool $isCoveredByTest,
+        Node $mutatedNode,
+        int $mutationByMutatorIndex
```
BackEndTea: This class now needs 9 parameters to instantiate. Its probably outside the scope of this PR, but we can probably refactor this to reduce that number, and keep it maintanable.

src/MutationInterface.php
```diff
@@ -33,4 +33,6 @@ public function getOriginalFileAst(): array;
     public function isOnFunctionSignature(): bool;
 
     public function isCoveredByTest(): bool;
+
+    public function getMutatedNode();
```
BackEndTea: How about adding a return type here?

src/Mutation.php
borNfreee: From the first look, I don't see how we can reduce the number of parameters (I'm not speaking about grouping them to some value objects which is the same basically).

Almost all of the public methods are just pure getters and. If you have any particular ideas, please advice.

src/MutationInterface.php
borNfreee: added, but had to revert. See how NodeTraverser works: https://github.com/nikic/PHP-Parser/blob/7f96481c80ecfe0fd8d37506064d3a25ed9cad23/lib/PhpParser/NodeTraverser.php#L209-L229

We can have many return types, including `Node`, `Node[]`, `int`, `null`.

I've added PHPDoc with all of them instead.

tests/Mutator/ReturnValue/IntegerNegationTest.php
```diff
@@ -29,34 +29,58 @@ public function test_mutator($input, $expected = null): void
     public function provideMutationCases(): array
     {
         return [
-            'It mutates negative int return to positive' => [
+            'It mutates negative -1 int return to positive' => [
```
BackEndTea: I think this has to be `-2`

tests/Mutator/AbstractMutatorTestCase.php
```diff
@@ -32,22 +35,31 @@
...
-            $this->assertSame($expectedCode, $realMutatedCode);
-            $this->assertSyntaxIsValid($realMutatedCode);
+        if ($expectedCode === null) {
+            $this->assertCount(0, $mutants);
         } else {
-            $this->assertSame($inputCode, $realMutatedCode);
+            foreach ($mutants as $realMutatedCode) {
+                $expectedCodeSample = array_shift($expectedCodeSamples);
+
+                if ($expectedCodeSample === null) {
```
BackEndTea: I think this will not error if there are more expected mutations than ones actual given. 
What about adding an assertSame on count of `$mutants` and `$expectedCodeSamples`

tests/Mutator/ReturnValue/IntegerNegationTest.php
borNfreee: mutated code should have -1 in this case, fixed

tests/Mutator/AbstractMutatorTestCase.php
borNfreee: fixed

src/Mutator/Arithmetic/RoundingFamily.php
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    protected function mutatesNode(Node $node): bool
+    {
+        if (!$node instanceof Node\Expr\FuncCall) {
+            return false;
+        }
+
+        if (!$node->name instanceof Node\Name ||
+            !\in_array($this->getNormalizedFunctionName($node->name), self::MUTATORS_MAP, true)) {
```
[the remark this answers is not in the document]
borNfreee: ok. Are you aware of any PHP-CS-Fixer rule that checks that? didn't find

.travis.yml
```diff
@@ -4,7 +4,7 @@ sudo: false
 
 env:
   global:
-    - INFECTION_FLAGS='--threads=4 --min-msi=62 --min-covered-msi=72 --coverage=coverage --log-verbosity=none'
+    - INFECTION_FLAGS='--threads=4 --min-msi=61 --min-covered-msi=69 --coverage=coverage --log-verbosity=none'
```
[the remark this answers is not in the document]
borNfreee: Rebased on master and checked random mutated folders side by side in `infection.log` - I didn't notice any difference. It's not 100% evidence there is no issue, but now it looks ok after rebasing

borNfreee:
any brave reviewers?

src/Mutator/Util/Mutator.php
```diff
@@ -24,7 +24,7 @@ public function __construct(MutatorConfig $config)
         $this->config = $config;
     }
 
-    abstract public function mutate(Node $node);
+    abstract public function mutate(Node $node): \Generator;
```
snapshotpl: What are you think to type it as `iterable`?

borNfreee: Great suggestion! Done. It's really more powerful

> It accepts any array or object implementing the Traversable interface. Both of these types are iterable using foreach and can be used with yield from within a generator.

BackEndTea: The downside here is that returning an array is also okay.
And the php-parser api allows to return an array of nodes to replace a set of nodes. Which may cause confusion.

borNfreee: booo :(

you are right, returning an array has special meaning and will definitely lead to an error.

https://github.com/nikic/PHP-Parser/blob/7f96481c80ecfe0fd8d37506064d3a25ed9cad23/lib/PhpParser/NodeTraverser.php#L212

I think the last commit should be reverted then

borNfreee: If someone curious, with generator we still can return an array to replace one node with several nodes, and we have this example https://github.com/infection/infection/blob/c71d6dd1976d44d5fb932e68cb8d0b83fdc020ff/src/Mutator/ReturnValue/FunctionCall.php#L29-L34

borNfreee: Updated my comment. We *want* to return an *array* in some cases, see example above. So technically it's ok to 

* return generator
* return an array if we want to replace one node with several nodes

We just need to cover this in documentation

borNfreee: Decided to revert.

With `\Generator`, if we want to replace 1 node by 1 node, we do

```php
yield new Node\...;
```

1 node by 1 node with different cases:

```php
yield new Node\...\1;
yield new Node\...\2;
```

1 node by many new nodes:

```php
yield [new Node\...\1, new Node\...\2];
```

With `iterable`, if we want to replace 1 node by 1 node, we do

```php
return  [new Node\...];
```

1 node by 1 node with different cases:

```php
return [new Node\...\1, new Node\...\2];
```

1 node by many new nodes:

```php
return [[new Node\...\1, new Node\...\2]]; // confusing
```

And with `iterable` we will also able to use `yield`. I agree, this more confusing than powerful.

src/Mutator/Arithmetic/BitwiseNot.php
```diff
@@ -22,11 +22,11 @@
...
      * @param Node $node
      *
-     * @return mixed
+     * @return \Generator
      */
-    public function mutate(Node $node)
+    public function mutate(Node $node): \Generator
     {
-        return $node->expr;
+        yield $node->expr;
```
sanmai: We're loosing flexibility here. E.g. if it would be allowed to return either simple values, or a generator, not only this upgrade would not require so many changes all around, some future mutators would be easier.

E.g. try refactoring this to use generators without a whole bunch of else/elseif:
```
if ($a) {
    return $k;
}

if ($b) {
    return $c;
}

if ($d) {
    return $k;
}
```
In the same time detecting generators from simple values is super easy:
```
if ($input instanceof Generator) {....
```

borNfreee: Not use I got what you mean and what you suggest.
Regarding code example, with generator it would be:

```php
if ($a) {
    yield $k;
    return;
}

if ($b) {
    yield $c;
    return;
}

if ($d) {
    yield $k;
    return;
}
```

No `else`/`elseif` if needed.

And we have such case: https://github.com/infection/infection/pull/450/files#diff-9908f23125ee671a485bd3a39b546da0R32

sanmai: Yeah, that's my point. Instead of a simple and clear `return` one has to use yield-return combo which could make everything very hard to understand because, for example, if you `yield` something twice, even if you return, you got to keep two values, and that's kind of expected for `yield`, but that's not necessarily easy to see from the code.

May I suggest we only use `yield` only where it is actually needed?

src/Visitor/MutationsCollectorVisitor.php
```diff
@@ -98,15 +98,19 @@ public function leaveNode(Node $node): void
...
-            $this->mutations[] = new Mutation(
-                $this->filePath,
-                $this->fileAst,
-                $mutator,
-                $node->getAttributes(),
-                \get_class($node),
-                $isOnFunctionSignature,
-                $isCoveredByTest
-            );
+            foreach ($mutator->mutate($node) as $mutationByMutatorIndex => $mutatedNode) {
```
sanmai: this could be like:

```
$mutations = $mutator->mutate($node);
if (!$mutations instanceof \Generator) {
   $mutations = [$mutations];
}
foreach ($mutations as $mutationByMutatorIndex => $mutatedNode) {

}
```
So you can see that these just three-four lines make everything way simpler elsewhere.

tests/Fixtures/StubMutator.php
```diff
@@ -8,7 +8,7 @@
 
 class StubMutator extends Mutator
 {
-    public function mutate(Node $node)
+    public function mutate(Node $node): \Generator
```
sanmai: This can't be called in the current form. [Example.](https://3v4l.org/A3I6V)
```
TypeError: Return value of StubMutator::mutate() must be an instance of Generator, none returned
```
So you see generators are tricky.

src/Mutator/Arithmetic/BitwiseNot.php
sanmai: Oh, and there's a fun part about returning from a generator. [It isn't exactly doing what you can expect.](https://3v4l.org/JLN73)

borNfreee:
@sanmai Would it work for you if I merge this one and will start splitting interfaces as a separate PR? Because it's already a very big diff

src/Mutation.php
```diff
@@ -132,4 +149,12 @@ public function isCoveredByTest(): bool
     {
         return $this->isCoveredByTest;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getMutatedNode()
+    {
+        return $this->mutatedNode;
```
theofidry: how can the node be an `int`, `array` or `null`?

src/Mutator/Arithmetic/BitwiseNot.php
theofidry: Generators are indeed a bit funny but I would either return a generator yielding the mutations or an array. Having mixed single value + generator/array looks more confusing than anything.

Maybe when you are using a specific mutator and expects only one mutation returning a single value would be a bit counter-intuitive, but I would argue 99% you do not care that it's a single value or multiple ones.

Wether we pick generators or arrays however is how you guys prefers it, given the usage, even if I have a personal preference for generators, I do think arrays will be simpler and less error-prone

src/Mutator/Arithmetic/RoundingFamily.php
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Infection\Mutator\Arithmetic;
+
+use Infection\Mutator\Util\Mutator;
+use PhpParser\Node;
+
+/**
+ * @internal
+ */
+final class RoundingFamily extends Mutator
```
theofidry: this one shouldn't be in this PR should it?

src/Visitor/MutationsCollectorVisitor.php
theofidry: point taken but then you can't have the return typehint and is there really any case where we care about returning a single value as opposed to an array? Is doing `return [$mutatedNode]` really more of an issue?

tests/Fixtures/StubMutator.php
theofidry: I would argue it's more of an example of forgetting to refactor a specific class rather than dealing with the generator trickiness here though

theofidry:
@sanmai sorry I replied to your comments during the review but it doesn't look like they were linked properly so reading them might be a bit annoying

src/Mutation.php
borNfreee: see https://github.com/infection/infection/pull/450#discussion_r209751722

in PHP-Parser API, you can replace node with another node, with array of nodes, you can return int like `NodeTraverser::STOP_TRAVERSAL`, etc.

src/Mutator/Arithmetic/RoundingFamily.php
borNfreee: It's the best example of how the new approach works, and it shouldn't be implemented with 3 classes like it was in another PR.

Of course, I can extract it, but this is extra work and I don't see why we should do it at this moment ;)

sanmai:
@theofidry no prob. Using arrays is the same deal with making everyone use generators. `BitwiseNot` returns just one mutation, why should it care that someone somewhere else wants to return more than one mutation?..

sanmai:
> if I merge this one and will start splitting interfaces as a separate PR? 

I'm fine with that. 

If you could use a feature-branch for this, I can give you a hand with that. Plus, then we could see what was exactly changed compared with master.

sanmai:
I think you can make this PR a lot shorter if you adopt a flexible approach, with a single interface, [like from here](https://github.com/infection/infection/pull/450#discussion_r216157923). Then it could go to the master.

(I'm not a big fan of huge changesets, since they pollute history, and make things like git blame output much less usable. As a saying goes, better a diamond with a flaw than a pebble without.)

theofidry:
>returns just one mutation, why should it care that someone somewhere else wants to return more than one mutation?..

I just don't really see a reason to tweak the interface signature for this peculiar case. The same way I would return an empty collection for no mutation I would return a collection of one element instead of changing the signature and callee to support single values. IMO, it's a question of where you want the "pain" (it's a bit of a strong for what it is hence the quote) be: in the implementation, i.e. force `[$node]`, `[]` or `$nodes` or in the callee which true for now is a single place so in the end, I think it doesn't matter all too much.

That said I would argue if you support multiple values (no value, single and array), I would also support generators...

borNfreee:
I'm a little be lost with what you suggest guys. We decided to separate interfaces to not confuse people, now we want to support generators as well as other 4 return types, ending up with 5 return types in general: `\Generator|Node|Node[]|int|null`

see my comment above:

> Such code assumes that the mutate() method won't have a return type, which is what I not even don't like, but want to avoid. And this is exactly why I reverted previous commit - because this type widening leads to confusion in the client code and helps users to make a mistake.

Probably this is what we will have at the end, but I want to understand what we agree with and what to do next to not waste time.

> you can make this PR a lot shorter if you just for a time being...

what do you mean by 'for a time being'? Would it be a temporary solution? For how long? What should we do with splitting interfaces, and should we split it at all?

> That said I would argue if you support multiple values (no value, single and array), I would also support generators...

just to make it clear: It is dictated by PHP-Parser, it was not our choice

sanmai:
> what do you mean by 'for a time being'? Would it be a temporary solution? 

Until your next PR where you plan to introduce this new scheme with two interfaces.

> should we split it at all?

I would rather not to, but if you absolutely feel like we *must* have two separate interfaces for mutators (traditional, and with a generator), I'm fine with that too. (Basically it boils down to removal of `mutate(Node $node)` from `Mutator`, since you can't have another abstract extending class [because it'll break the LSP](https://3v4l.org/FNjsO).)

borNfreee:
@sanmai I've reverted changes in mutators for now and made it to return `Node|Node[]|\Generator`. Could you please have a look at the diff again?

borNfreee:
Thank you for the review @BackEndTea, @sanmai, @theofidry

PR #452, opened 2018-08-14
Add check for Node\Expr\Closure in MutationsCollectorVisitor

pierredup:
In one of my projects, I'm getting an error `PHP Fatal error:  Uncaught Error: Call to undefined method PhpParser\Node\Expr\Closure::isAbstract() in .../src/Visitor/MutationsCollectorVisitor.php:88`. This is with a simple function and test (no objects). This PR fixes the error, but I'm not completely sure how to add a test for it. Any guidance would be appreciated

#### Reproducer:

file.php

```php
<?php

function map($values, $callable): array
{
    return array_map(function ($value) use ($callable) {
        return $callable($value);
    }, $values);
}
```

test.php

```php
class MapTest extends TestCase
{
    public function testMap()
    {
        $square = function (int $n) {
            return $n * $n;
        };

        $this->assertSame([16, 64], map([4, 8], $square));
    }
}
```

### DIFF
--- a/src/Visitor/MutationsCollectorVisitor.php
+++ b/src/Visitor/MutationsCollectorVisitor.php
@@ -85,7 +85,7 @@ final class MutationsCollectorVisitor extends NodeVisitorAbstract
 
             if ($isOnFunctionSignature
                 && ($methodNode = $node->getAttribute(ReflectionVisitor::FUNCTION_SCOPE_KEY))
-                && $methodNode->isAbstract()
+                && (!$methodNode instanceof Node\Expr\Closure && $methodNode->isAbstract())
             ) {
                 continue;
             }


borNfreee:
Thanks for opening this PR. I'me pretty sure this is a bug and functions must not be mutated (see corresponded PR and test https://github.com/infection/infection/pull/113/files#diff-ed91c9c161449c5a13dd2bad202c1fceR276)

I didn't have a chance to investigate this particular code/case, but seems like I believe we should return earlier than `$methodName->abstract()` line

Btw, do you have a public repository with reproducible code? I would like to exclude all other mistakes (like mutating the test code and so on).

pierredup:
I'm getting this on [lodash-php/lodash-php](https://github.com/lodash-php/lodash-php). I'll see if I can create a minimal repository to reproduce the error

sanmai:
Indeed I was able to reproduce the issue on lodash-php. Suffice to say we could actually have a closure exactly at that place, [as ReflectionVisitor treats them like functions].(https://github.com/infection/infection/blob/32a399d214e384feaed79775ce460b9ada6da952/src/Visitor/ReflectionVisitor.php#L124)

pierredup:
Closed in favour of #456

borNfreee:
Thank you @pierredup 

Release [0.10.3](https://github.com/infection/infection/releases/tag/0.10.3)

PR #453, opened 2018-08-15
Move PHP-CS-Fixer from `analyze` to PrettyCI

borNfreee:
Move PHP-CS-Fixer from `analyze` to PrettyCI for CI. 

It will speed up feedback about broken builds a little bit, read more here https://prettyci.com/


borNfreee:
Thank you

PR #454, opened 2018-08-17
Rename infection-log.txt -> infection.log

borNfreee:
This PR:

- [x] Renames `infection-log.txt` -> `infection.log`
- [x] Doc PR: https://github.com/infection/site/pull/58

#### Why?

Because this is how other frameworks, tools, etc. name their log files (symfony, nginx, apache and so on).

There is no BC break because each project should have already existing `infection.json` with the name of the config file. Only new users are affected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 /.idea
 /vendor
-/infection-log.txt
+/infection.log
 /build/*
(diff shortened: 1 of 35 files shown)


borNfreee:
thank you

PR #456, opened 2018-08-20
Verify that MutationsCollectorVisitor can work around closures

sanmai:
ReflectionVisitor treats closures just like plain functions, where they're different objects. This isn't a problem for ReflectionVisitor, but MutationsCollectorVisitor should take a care of this. Sadly, we don't have a test suite for the latter, therefore a test for MutationsGenerator was used to cover this issue.

Fixes #452

### DIFF
--- a/tests/Fixtures/Files/Mutation/OneFile/OneFile.php
+++ b/tests/Fixtures/Files/Mutation/OneFile/OneFile.php
@@ -15,6 +15,8 @@ class OneFile
 
     public function foo($value = true): int
     {
-        return 33 + 44;
+        return (function ($input = 33) {
+            return $input;
+        })() + 44;
     }
 }
--- a/src/Visitor/MutationsCollectorVisitor.php
+++ b/src/Visitor/MutationsCollectorVisitor.php
@@ -83,11 +83,13 @@ final class MutationsCollectorVisitor extends NodeVisitorAbstract
                 continue;
             }
 
-            if ($isOnFunctionSignature
-                && ($methodNode = $node->getAttribute(ReflectionVisitor::FUNCTION_SCOPE_KEY))
-                && $methodNode->isAbstract()
-            ) {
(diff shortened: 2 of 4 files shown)


tests/Fixtures/Files/Mutation/OneFile/OneFile.php
```diff
@@ -15,6 +15,8 @@ class OneFile
 
     public function foo($value = true): int
     {
-        return 33 + 44;
+        return (function ($input = 33) {
+            return $input;
+        })() + 44;
```
sanmai: A mutation on a closure's signature needed to trigger the issue.

borNfreee:
@sanmai could you please target it to `0.10`?

sanmai:
I see why we don't mutate plain functions, but I don't see how that's concerns closures in any way.

We cannot mutate plain functions because we can't override them in the standard PHP. But if one can get APD installed on a CI server, theoretically [we can do whatever there is we need with functions](http://php.net/manual/en/function.rename-function.php). But then, again, there are no such problems with closures, since they require not autoloading whatsoever.

borNfreee:
> But if one can get APD installed on a CI server

I don't think it worth it, at least it sounds scary.

> But then, again, there are no such problems with closures, since they require not autoloading whatsoever.

Well, from what I understand, we shouldn't mutate closures that are **inside** plain functions

```php
function property($path): callable
{
    return function ($value, $index = 0, $collection = []) use ($path, $propertyAccess) {
          // should NOT be mutated
    };
}
```

On the other side, we can mutate closures that are on the global scope:

```php
// global scope

$filtered = array_filter([1, 2, 3], function ($value) {
    return $value % 2; // we CAN mutate here
});
```

Does it make sense @sanmai ?

borNfreee:
Yes, you are right. 

Thank you @sanmai

PR #458, opened 2018-08-21
Validate that infection.json contains valid (writable) file paths for loggers

borNfreee:
When for example `logs.text` key in `infection.json` file contains not writable path, infection fails with a correct message, but does that only *after* mutation logic.

not_writable_bug

This PR:

- [x] Validates logger file paths right after `InfectionConfig` is instantiated.
- [x] Covered by tests

With this update, user will get validation error immediately:

fixed_bug

### DIFF
--- a/src/Config/Validator.php
+++ b/src/Config/Validator.php
@@ -29,17 +29,19 @@ final class Validator
         $logTypes = $config->getLogsTypes();
 
         foreach ($logTypes as $logType => $file) {
-            if ($logType !== ResultsLoggerTypes::BADGE) {
-                $dir = \dirname($file);
+            if ($logType === ResultsLoggerTypes::BADGE) {
+                continue;
+            }
 
-                if (is_dir($dir) && !is_writable($dir)) {
-                    throw new IOException(
-                        sprintf('Unable to write to the "%s" directory. Check "logs.%s" file path in infection.json.', $dir, $logType),
-                        0,
-                        null,
-                        $dir
-                    );
-                }
+            $dir = \dirname($file);
+
+            if (is_dir($dir) && !is_writable($dir)) {
+                throw new IOException(
+                    sprintf('Unable to write to the "%s" directory. Check "logs.%s" file path in infection.json.', $dir, $logType),
+                    0,
+                    null,
+                    $dir
+                );
             }
         }
     }
--- a/src/Config/Validator.php
+++ b/src/Config/Validator.php
@@ -1,0 +1,46 @@
+<?php
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+declare(strict_types=1);
+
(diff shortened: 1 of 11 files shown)


src/Config/Validator.php
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Infection\Config;
+
+use Infection\Logger\ResultsLoggerTypes;
+use Symfony\Component\Filesystem\Exception\IOException;
+
+/**
+ * @internal
+ */
+final class Validator
```
BackEndTea: What about making this a `ConfigValidator`, which takes the infection config as its input, and does the validation for all other things as well?

src/Config/Validator.php
```diff
@@ -0,0 +1,46 @@
...
+
+        return true;
+    }
+
+    private function validateLogFilePaths(InfectionConfig $config): void
+    {
+        $logTypes = $config->getLogsTypes();
+
+        foreach ($logTypes as $logType => $file) {
+            if ($logType !== ResultsLoggerTypes::BADGE) {
```
BackEndTea: what if we do 

```php
if ($logType === ResultsLoggerTypes::BADGE) {
    continue;
}
```
to cut down the indentation a bit.

src/Config/Validator.php
borNfreee: > which takes the infection config as its input, and does the validation for all other things as well?

it does exactly that. I'm not sure I got what you mean. It takes `InfectionConfig` instance and does validation (only `logs` key is validated at the moment)

If you are speaking about renaming `Validator` to `ConfigValidator` - I'm not sure, because it is already in `Config` namespace, so it would be `Infection\Config\ConfigValidator`.

BackEndTea: My bad, i didn't think about simply adding more statements to `validate` in the future, which works the same as having it in a constructor.

sanmai:
There's more to it. If a given file is writable now, does not necessarily it will still be writable an hour later. So I think it would be best if Infection would fail gracefully on a failed write in the end, with, like, a warning.

src/Console/InfectionContainer.php
```diff
@@ -196,6 +197,10 @@ public function __construct(array $values = [])
         $this['memory.formatter'] = function (): MemoryFormatter {
             return new MemoryFormatter();
         };
+
+        $this['infection.config.validator'] = function (): Validator {
+            return new Validator();
```
sanmai: Why can't it be initialized right here? E.g. just like it's done above for other dependencies with `$this->getInfectionConfig()`.

sanmai: Oh, nevermind. It'll be more difficult to pull out.

borNfreee:
> There's more to it. If a given file is writable now, does not necessarily it will still be writable an hour later. So I think it would be best if Infection would fail gracefully on a failed write in the end, with, like, a warning.

Done. Could you please guys review it again?

borNfreee:
thank you

PR #460, opened 2018-08-23
[Removal] Don't remove calls to assert()

duncan3dc:
The new `FunctionCallRemoval` mutator is removing calls to `assert()` and expecting the tests to fail, but the `assert()` function is a special case that it doesn't make sense to remove.

This PR prevents calls to `assert()` from being removed, and adds a test to ensure they aren't removed

### DIFF
--- a/src/Mutator/Removal/FunctionCallRemoval.php
+++ b/src/Mutator/Removal/FunctionCallRemoval.php
@@ -28,6 +28,22 @@ final class FunctionCallRemoval extends Mutator
             return false;
         }
 
-        return $node->expr instanceof Node\Expr\FuncCall;
+        if (!$node->expr instanceof Node\Expr\FuncCall) {
+            return false;
+        }
+
+        $name = $node->expr->name;
+
+        if (!$name instanceof Node\Name) {
+            return true;
+        }
+
+        $string = strtolower((string) $name);
+
+        if ($string === 'assert') {
+            return false;
+        }
+
+        return true;
     }
 }
--- a/tests/Mutator/Removal/FunctionCallRemovalTest.php
+++ b/tests/Mutator/Removal/FunctionCallRemovalTest.php
@@ -43,6 +43,28 @@ PHP
             ,
         ];
 
+        yield 'It removes dynamic function calls' => [
+            <<<'PHP'
+<?php
+
+$start = true;
+$foo();
+('foo')();
+$end = true;
+
+PHP
+            ,
+            <<<'PHP'
+<?php
+
+$start = true;
+
+
+$end = true;
+PHP
+            ,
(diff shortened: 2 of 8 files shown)


duncan3dc:
I've made those changes now and the build is green :tada:

src/Mutator/Removal/FunctionCallRemoval.php
```diff
@@ -28,6 +29,22 @@ protected function mutatesNode(Node $node): bool
...
 
-        return $node->expr instanceof Node\Expr\FuncCall;
+        if (!$node->expr instanceof Node\Expr\FuncCall) {
+            return false;
+        }
+
+        $name = $node->expr->name;
+
+        if (!$name instanceof Node\Name) {
+            return false;
```
BackEndTea: I think this can be `return true;`. Even if we don't know what function is called, because the name is a variable or a string for example, we can still remove it and see if it has an effect.

tests/Mutator/Removal/FunctionCallRemovalTest.php
```diff
@@ -82,6 +82,17 @@ public function provideMutationCases(): \Generator
 
 $this->foo();
 $a = 3;
+PHP
+        ];
+
+        yield 'It does not remove an assert() call' => [
```
BackEndTea: Please add a test where the function name is a variable or a string to make sure it doesn't crash there.

The `PregQuote` test is an example of how to do that.

src/Mutator/Removal/FunctionCallRemoval.php
```diff
@@ -11,6 +11,7 @@
 
 use Infection\Mutator\Util\Mutator;
 use PhpParser\Node;
+use function strtolower;
```
BackEndTea: We don't do this anywhere within the code base, so please revert this change.

src/Mutator/Removal/FunctionCallRemoval.php
duncan3dc: Yep, that was a mistake, fixed and added a test :+1:

src/Mutator/Removal/FunctionCallRemoval.php
borNfreee: on the other side - we can discuss it and start using this approach, at least understand core developers think about it.

Personally, I find it more readable as it removes all those `\`,

`\assert(\count($a) > 0)` vs `assert(count($a) > 0)`

@infection/core

borNfreee:
Nice work @duncan3dc ! Thank you for your contribution

borNfreee:
Released https://github.com/infection/infection/releases/tag/0.10.1

PR #457, opened 2018-08-21
WIP: Codeception

Fenikkusu:
This PR:

- [x] Adds Codeception Testing Framework
- [x] Covered by tests
- [ ] Doc PR: https://github.com/infection/site/pull/XXX

Fixes https://github.com/infection/infection/pull/457

Since Tobias is no longer maintaining the Codeception work he did, I've cleaned up my butchering of his work and am resubmitting it. As of now, the changes are working. However, I suspect there are a few pieces that may need to be altered due to a change in practice with the base system. I have not had the time to verify this. I am currently trying to get the changes working in the GitLab Runner, but am having some hard times with it (there is an error occurring, but no error output is given). A number of the changes I have added in are for the purpose of debugging the issues in the CI, but will likely be fine to maintain as additional features to the code.

I'm looking for any comments/thoughts on cleaning this up to get it fully implemented. On a side note, I'm aware a number of the checks will be invalid at this time. I've not fully cleaned up everything yet.

PR #461, opened 2018-08-24
WIP: Updating Factory To Be More Generic (Not Framework Specific).

Fenikkusu:
This PR:

- [x] Adds new feature ...
- [ ] Covered by tests

Preliminary PR For Factory Clean Up Per Disccusions In PR #457. This PR is to clean up the TestFramework factory to not be framework specific. This will allow new frameworks to be added without the need to update the factory with the new framework.


Fenikkusu:
Does it make sense to standardize class names as part of this. I see no reason to repeat the framework name in the class name when the namespace is already doing that. For example:

```
\Infection\TestFramework\PhpUnit\ExtraOptions
\Infection\TestFramework\PhpSpec\ExtraOptions
\Infection\TestFramework\Codeception\ExtraOptions
```

instead of

```
\Infection\TestFramework\PhpUnit\PhpUnitExtraOptions
\Infection\TestFramework\PhpSpec\PhpSpecExtraOptions
\Infection\TestFramework\Codeception\CodeceptionExtraOptions
```

Fenikkusu:
Looking to get some feedback on this most recent run. I've gotten the factory and name spacing cleaned up. One thing I noticed is that the factory is being initialized with params specifically for PhpUnit. This doesn't make sense to me, so I started to strip those out, with the intention of initialization in the actual relative classes. This begs the question of what makes the most sense to access them. They were originally in the container, so I can pull them from there as well, but I'd either have to pass it through or create a new instance, neither of which seems appropriate. Thoughts?

Fenikkusu:
Looking to get some feedback on this most recent run. I've gotten the factory and name spacing cleaned up. One thing I noticed is that the factory is being initialized with params specifically for PhpUnit. This doesn't make sense to me, so I started to strip those out, with the intention of initialization in the actual relative classes. This begs the question of what makes the most sense to access them. They were originally in the container, so I can pull them from there as well, but I'd either have to pass it through or create a new instance, neither of which seems appropriate. Thoughts?

borNfreee:
Could you please give more details, probably with code examples? I'm not sure I understand the problem you are trying to solve.

> This doesn't make sense to me, so I started to strip those out, with the intention of initialization in the actual relative classes.

What do you mean by `actual relative classes`?

As for me, it's ok that different test framework adapters require different dependencies. Keeping this in mind I don't know how to make the factory class framework independent. We can leave dependency management for Container itself, but in this case, we will still need to create additional container service when new test framework is implemented.

If this is going to be so complex, let's answer to the question: does it worth refactoring at all? We are adding test frameworks not so often ;)

Fenikkusu:
Well, keep in mind that a portion of the refactor comes more from the FQNS changes. If I were to just move all the creations directly to the container, that would eliminate a chunk of the complexity of the change.

By `actual relative classes`, I mean, primarily, the PhpUnit Initial Builder. It has dependencies on the xml configuration helper. Rather than passing this through the to the Factory, I think it makes more sense to let the Builder just directly pull from the container. If that makes sense.

src/TestFramework/Factory.php
```diff
@@ -67,60 +48,73 @@ public function __construct(
...
+        $configPath = $this->configLocator->locate($adapterName);
 
-            return new PhpSpecAdapter(
-                new TestFrameworkFinder(TestFrameworkTypes::PHPSPEC),
-                new PhpSpecInitialConfigBuilder($this->tmpDir, $phpSpecConfigPath, $skipCoverage),
-                new PhpSpecMutationConfigBuilder($this->tmpDir, $phpSpecConfigPath, $this->projectDir),
-                new PhpSpecArgumentsAndOptionsBuilder(),
-                $this->versionParser
-            );
+        $baseNamespace = __NAMESPACE__ . '\\' . $adapterName . '\\';
```
BackEndTea: I prefer the old if, else statements, it does add mroe boilerplate code, but in all reality there won't be much test frameworks being added. This is too 'magical', and makes is harder to reason about what is going on.

Fenikkusu:
Okie Dokie

PR #462, opened 2018-08-25
Use default test framework if neither CLI nor config has the value for it

borNfreee:
During the installation process of Infection for the new project, I noticed a wierd question from `ConfigureCommand`:

```
Where is your .(xml|yml)(.dist) configuration located?:
```

Note the absence of `phpunit` ot `phpspec` before `.(xml|yml)(.dist)` (sprintf pattern is `Where is your %s.(xml|yml)(.dist) configuration located?`

Previously, we had a default value for `--test-framework` option in `InfectionCommand`, but it was removed in https://github.com/infection/infection/pull/281

Now, the value of test framework can be set from `infection.json.dist` or cli option. So this update checks cli option as a priority and then checks config. If nothing is found, it defaults to `phpunit`.

After this update, no question is raised at all, infection automatically finds `phpunit.xml[.dist]`.


borNfreee:
Thank you

PR #463, opened 2018-08-26
Adding Additional CLI Settings To Config

Fenikkusu:
This PR:

- [x] Adds new feature ...
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/60

Per discussions in PR #457, Created separate PR For Config Updates. Added additional cli setting to config file for initial-test-php-options and test-framework-options. Needed the changes because I was trying to debug in a Gitlab Runner and couldn't get the normal command to run correctly.

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -196,7 +196,7 @@ final class InfectionCommand extends BaseCommand
         $initialTestSuitProcess = $initialTestsRunner->run(
             $testFrameworkOptions->getForInitialProcess(),
             $this->skipCoverage,
-            explode(' ', $input->getOption('initial-tests-php-options'))
+            explode(' ', $input->getOption('initial-tests-php-options') ?: $config->getInitialTestsPhpOptions())
         );
 
         if (!$initialTestSuitProcess->isSuccessful()) {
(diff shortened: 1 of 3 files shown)


src/Command/InfectionCommand.php
```diff
@@ -196,7 +196,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $initialTestSuitProcess = $initialTestsRunner->run(
             $testFrameworkOptions->getForInitialProcess(),
             $this->skipCoverage,
-            explode(' ', $input->getOption('initial-tests-php-options'))
```
borNfreee: what about checking it with `hasOption()`?

```php
$input->hasOption('initial-tests-php-options') ? $input->getOption('initial-tests-php-options') : $config->getInitialTestsPhpOptions()
```

This will work correcty when `""` is passed as a value of the option. Same below

borNfreee: @Fenikkusu are you going to finish this PR or can we take over it?

Fenikkusu: Honestly, I didn't see this come through and hadn't had a chance to circle back to it since. If I get the chance, I can update it tonight. Otherwise you are welcome to run with it.

Fenikkusu:
Should be Good Now

Issue #459, opened 2018-08-23
Premature Process Termination.


Fenikkusu:
| Question    | Answer
| ------------| ---------------
| Infection version | #457 
| Test Framework version | Codeception 2.4.2
| PHP version | 7.1.17
| Platform    | MacOS
| Github Repo | https://github.com/twistersfury/infection

InitialTestsRunner::57-58 needs to be revised. When codeception runs, it echos to the error output in pieces. This creates a race condition where all the output of STDERR does not have a chance to echo/process before the parent infection run calls the stop command. This ends up resulting in error code 143 because the pipe gets killed prematurely. I'm still looking for the best way to fix this, but thought someone might have an answer. At the very least, the stop command should allow the echo to finish. My suggestion is to only call the stop after the timeout command. That should give the main command to successfully finish in full. Will test when I get the chance.

Fenikkusu:
I have applied a preliminary/suggested fix to the MR.

maks-rafalko:
Closing as no one complains about this, and Codeception was redeveloped from scratch

PR #464, opened 2018-08-26
Resolving Premature Exit On Error

Fenikkusu:
This PR:

- [x] Adds new feature ...
- [x] Covered by tests

Fixes https://github.com/infection/infection/issues/459

When running the test framework, if infection detects any errors, it immediately kill the process. If this happens and the process has additional output, you may not get the proper error. Instead you get an error code 143 because the pipe was closed early as a result of a race condition.

This fixes this by adding a 10 second delay to processing to allow the framework to fully exit. In most cases, it will likely take less than 1 second.

This fix relates to PR #457 in that the bug was originally detected when running codeception testing through GitLab Runner.


sanmai:
CI fails with `SYMFONY_VERSION="^4.0"`

src/Process/Runner/InitialTestsRunner.php
```diff
@@ -31,16 +31,22 @@
...
+    private $errorTimeout;
+
     /**
      * InitialTestsRunner constructor.
      *
      * @param ProcessBuilder $processBuilder
      * @param EventDispatcherInterface $eventDispatcher
      */
-    public function __construct(ProcessBuilder $processBuilder, EventDispatcherInterface $eventDispatcher)
+    public function __construct(ProcessBuilder $processBuilder, EventDispatcherInterface $eventDispatcher, int $errorTimeout = 10)
```
sanmai: I guess this should be a `private const` somewhere above, what do you think?

Fenikkusu:
Anyone want to provide suggestions on why appveyor is not passing...?...'Command exited with code -1073741819' I'm on an OSX machine so not sure how to replicate that currently...

Fenikkusu:
Must be an appveyor issue. I'm getting the same error on PR #465 where none of the code and/or tests were changed.

Fenikkusu:
> Why do we need to wait for the whole STDERR output from Codeception, but not just the first bits? To get the full error message?

It's not so much about getting the full error message so much as giving the process the chance to actually exit properly. Otherwise, infections kills the pipe prematurely, and when codeception tries to write to that pipe, the process as a whole errors out (IE: symfony dies) because it tries to write to the closed pipe. This results in error code 143 from codeception, which typically just exists with code 1 and the error response. Since all of this happens within microseconds, infections ends up getting the 143 exit code, and a blank message.

 > If the answer to the previous question is "yes", I would say we need to do this new logic only for Codeception and leave previous logic as it was, because it adds extra 1000ms when it should not.

It's true it adds a second, but I can easily alter that in two ways the first is to change it to a `while` instead of a `do` loop, so the first thing it does is check is running. By that time, it may of been exited already. Second, I can alter it from 1000ms to say 100ms, which honestly is probably more appropriate, as I suspect that in most cases, 1000ms is overkill.

theofidry:
@Fenikkusu any chance to get a rebase to merge this one?

Fenikkusu:
Somehow managed to screw up the rebase. Reopening after changes.

Fenikkusu:
Rebased, Cleaned Up Commits. Also had a test run done for a project and it took care of the specific issue (related to codeception failure when pre-running testing).

src/Process/Runner/InitialTestsRunner.php
```diff
@@ -47,9 +47,14 @@
...
 final class InitialTestsRunner
 {
+    private const ERROR_TIMEOUT = 10;
+
     private $processBuilder;
     private $eventDispatcher;
 
+    /**
+     * InitialTestsRunner constructor.
+     */
```
theofidry, line 57:
```suggestion
```

I don't think this comment has any value

src/Process/Runner/InitialTestsRunner.php
```diff
@@ -66,16 +71,35 @@ public function run(
         array $phpExtraOptions,
         bool $skipCoverage
     ): Process {
+        /** @var Process $process */
```
theofidry, line 74:
```suggestion
```

The process type is documented in `ProcessBuilder` so this should not be necessary

src/Process/Runner/InitialTestsRunner.php
```diff
@@ -47,9 +47,14 @@
...
+    private const ERROR_TIMEOUT = 10;
+
     private $processBuilder;
     private $eventDispatcher;
 
+    /**
+     * InitialTestsRunner constructor.
+     */
     public function __construct(
         InitialTestsRunProcessFactory $processFactory,
```
theofidry, line 59: hm looks like the property `processBuilder` has not been renamed to `processFactory`, I'll check that out in a separate PR

src/Process/Runner/InitialTestsRunner.php
```diff
@@ -66,16 +71,35 @@ public function run(
...
     ): Process {
+        /** @var Process $process */
         $process = $this->processBuilder->createProcess(
             $testFrameworkExtraOptions,
             $phpExtraOptions,
             $skipCoverage
         );
 
+        //Tracking Process Error Exit
+        $expirationData = (object) [
```
theofidry, line 82: is there any reason why to use an object here? It feels like this could just be `$expirationTime = null` and then pass the variable by reference in the callback

src/Process/Runner/InitialTestsRunner.php
```diff
@@ -66,16 +71,35 @@ public function run(
...
         bool $skipCoverage
     ): Process {
+        /** @var Process $process */
         $process = $this->processBuilder->createProcess(
             $testFrameworkExtraOptions,
             $phpExtraOptions,
             $skipCoverage
         );
 
+        //Tracking Process Error Exit
```
theofidry, line 81: CS nitpick, we used to leave a space between the `//` and the text, i.e. `// Tracking`

src/Process/Runner/InitialTestsRunner.php
```diff
@@ -47,9 +47,14 @@
  */
 final class InitialTestsRunner
 {
+    private const ERROR_TIMEOUT = 10;
```
theofidry, line 50: I think a comment would be appreciated here, at the very least to specify the unit of that number

src/Process/Runner/InitialTestsRunner.php
```diff
@@ -66,16 +71,35 @@ public function run(
...
+        $process->run(function (string $type) use ($process, $expirationData): void {
             if ($process::ERR === $type) {
+                //If already started, do not start again and let parent one run - prevent infinite loop.
+                //->isRunning calls the callback again every time there is any updated output.
+                if ($expirationData->time !== null) {
+                    return;
+                }
+
+                //Give The Error Processing Time To Fully Output
+                $expirationData->time = time() + self::ERROR_TIMEOUT;
```
theofidry, line 97: an import is missing for the `time()` function

src/Process/Runner/InitialTestsRunner.php
```diff
@@ -66,16 +71,35 @@ public function run(
...
+        ];
+
         $this->eventDispatcher->dispatch(new InitialTestSuiteWasStarted());
 
-        $process->run(function (string $type) use ($process): void {
+        $process->run(function (string $type) use ($process, $expirationData): void {
             if ($process::ERR === $type) {
+                //If already started, do not start again and let parent one run - prevent infinite loop.
+                //->isRunning calls the callback again every time there is any updated output.
+                if ($expirationData->time !== null) {
```
theofidry, line 92: After checking, I'm under the impression this could be done by passing the `$timeout` property to `Process::stop()`. It's however hard to tell for me if it's enough since I can't reproduce the issue

tests/e2e/Delayed_Error_Output/tests/SourceClassTest.php
```diff
@@ -0,0 +1,22 @@
...
+namespace Namespace_\Test;
+
+use Namespace_\SourceClass;
+use PHPUnit\Framework\TestCase;
+
+class SourceClassTest extends TestCase
+{
+    public function test_hello()
+    {
+        $outHandle = fopen('php://stderr', 'w');
```
theofidry, line 12:
```suggestion
        $stderrHandle = fopen('php://stderr', 'w');
```
?

composer.lock
```diff
@@ -3167,6 +3167,5 @@
     },
     "platform-overrides": {
         "php": "7.3.12"
-    },
-    "plugin-api-version": "1.1.0"
+    }
```
theofidry, line 3170: is that update necessary?

src/Process/Runner/InitialTestsRunner.php
Fenikkusu: I took a look at this and it is unfortunately a little different. By default, the timeout is 10 when calling `stop`. The problem is that `stop` immediately sends the SIGTERM signal and then waits for it to shutdown, which makes sense. This in effect causes the issue we are seeing. If the SIGTERM wasn't there, then this would be fine. BTW, the e2e test is able to replicate the issue easily enough.

src/Process/Runner/InitialTestsRunner.php
Fenikkusu: Removed

src/Process/Runner/InitialTestsRunner.php
Fenikkusu: Bad past experiences is the only reason I do this. I've had too many issues with forced by-ref causing issues that I'd rather rely on the natural by-ref of objects. 

Updated

src/Process/Runner/InitialTestsRunner.php
Fenikkusu: Updated

tests/e2e/Delayed_Error_Output/tests/SourceClassTest.php
Fenikkusu: Updated

composer.lock
Fenikkusu: Not sure how that happened. I've reverted the changes to composer.lock completely.

src/Process/Runner/InitialTestsRunner.php
Fenikkusu: Updated

src/Process/Runner/InitialTestsRunner.php
Fenikkusu: Updated

src/Process/Runner/InitialTestsRunner.php
Fenikkusu: Updated

Fenikkusu:
Changes Applied.

maks-rafalko:
Closing as no one complains about this, and Codeception was redeveloped from scratch

PR #465, opened 2018-08-29
Tweaking Makefile For Easy Symfony Switch

Fenikkusu:
This PR:

- [x] Adds new feature ...

While working on PR #464 I was getting errors with the testing in travis for Symfony 4.0. It took me a bit of digging to figure out how to do the switch locally. After figuring it out, I went ahead and added some helper methods to the Makefile to switch between 3.0 and 4.0 more easily.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,3 +137,20 @@ build/bin/infection.phar: $(shell find bin/ src/ -type f) box.phar box.json.dist
 box.phar:
 	wget $(BOX_URL)
 	chmod a+x box.phar
+
+.PHONY: symfony-40 symfony-30
+symfony-40:
+	composer config --unset platform.php
+	composer require symfony/console:^4.0 \
+		symfony/filesystem:^4.0 \
+		symfony/process:^4.0 \
+		symfony/finder:^4.0 \
+		symfony/yaml:^4.0
+
+symfony-30:
+	composer config platform.php 7.1
+	composer require symfony/console:^3.0 \
+		symfony/filesystem:^3.0 \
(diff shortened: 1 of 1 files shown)


borNfreee:
Thanks for making this PR.

I'm personally -1 for this PR, because it's a very specific task. I have never needed it during infection development and think you will never need it again, so probably it will be a dead code.

As an option, you can create your own makefile locally which would be in `.gitignore` and extend the current functionality (https://www.gnu.org/software/make/manual/html_node/Overriding-Makefiles.html)

@infection/core thoughts?

sanmai:
Mixed feelings about this. In a such an extensive Makefile we have these useful bits will be forgotten with certainty. So they should be mentioned [in the `CONTRIBUTING.md`](https://github.com/infection/infection/blob/master/.github/CONTRIBUTING.md). But on the other hand these are so small so they could go to `CONTRIBUTING.md` in their entirety. So, what do we all think about adding them there?

borNfreee:
When it will be a time to update our `.travis.yml`

https://github.com/infection/infection/blob/af67763a108c03779f13efea2493aa8d3f4f2564/.travis.yml#L40-L47

we will likely forget to update `CONTRIBUTING.md` with the similar changes, because nobody cares about this file in general. Thus, we will have 2 places with different, but similar instructions.

I'm not against such things, only want to avoid unneeded maintenance for us in the future.

borNfreee:
I appreciate your work @Fenikkusu, however, the future maintenance and possible inconsistencies do not worth this small addition.

Thank you

Issue #477, opened 2018-09-05
File permission tests fail on Docker for Windows


johnstevenson:
| Question    | Answer
| ------------| ---------------
| Infection version | master
| Test Framework version | irrelvant
| PHP version | 7.1.18
| Platform    | Docker for Windows
| Github Repo | -

When running tests on Docker for Windows, file permissions cannot be set due to the limitations of the host operating system. This currently makes the following tests fail:

- Config/ValidatorTest.php
- Logger/SummaryFileLoggerTest.php

PR #478, opened 2018-09-05
Ensure file permission is really set in tests

johnstevenson:
This PR:

- [ ] Marks tests skipped if the required file permission cannot be set

Fixes https://github.com/infection/infection/issues/477

This is noticeable on Docker for Windows, where file permissions cannot be set due to the host (Windows) operating system.


johnstevenson:
No worries (I only wanted to update xdebug-handler, but this other stuff started to irritate me!)

PR #485, opened 2018-09-22
Fix Decrement integer

BackEndTea:
This PR:

- [x] Fixes some issues surrounding DecrementInteger
- [x] Uses `toLowerString` for php-parser name nodes
- [x] Covered by tests

The checks for the decrement integer mutator did not consider the `sizeOf` function, which is an alias of `count`. It did not behave correctly if the count function was on the right side of the equation. Which needs to have greater than and lesser than flipped.

I also changed our own `strtolower` for node names to use php-parser's `toLowerString`.

### DIFF
--- a/src/Mutator/Number/DecrementInteger.php
+++ b/src/Mutator/Number/DecrementInteger.php
@@ -18,7 +18,7 @@ use PhpParser\Node;
  */
 final class DecrementInteger extends Mutator
 {
-    private const COUNT_NAME = ['count', 'sizeof'];
+    private const COUNT_NAMEs = ['count', 'sizeof'];
 
     /**
      * Decrements an integer by 1
@@ -55,8 +55,8 @@ final class DecrementInteger extends Mutator
 
         if ($parentNode->left instanceof Node\Expr\FuncCall
             && \in_array(
-                $this->getLowercaseMethodName($parentNode, 'left'),
-                self::COUNT_NAME,
+                $parentNode->left->name->toLowerString(),
+                self::COUNT_NAMEs,
                 true)
         ) {
             return $this->isAllowedLeftSideCountComparison($parentNode);
@@ -64,8 +64,8 @@ final class DecrementInteger extends Mutator
 
         if ($parentNode->right instanceof Node\Expr\FuncCall
             && \in_array(
-                $this->getLowercaseMethodName($parentNode, 'right'),
-                self::COUNT_NAME,
+                $parentNode->right->name->toLowerString(),
+                self::COUNT_NAMEs,
                 true)
         ) {
             return $this->isAllowedRightSideCountComparison($parentNode);
@@ -104,10 +104,5 @@ final class DecrementInteger extends Mutator
     {
         return $parentNode instanceof Node\Expr\BinaryOp\Smaller
             || $parentNode instanceof Node\Expr\BinaryOp\SmallerOrEqual;
-    }
-
-    private function getLowercaseMethodName(Node $node, string $part): string
-    {
-        return $node->{$part}->name->toLowerString();
     }
 }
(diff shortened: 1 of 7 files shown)


src/Mutator/Number/DecrementInteger.php
```diff
@@ -18,7 +18,7 @@
  */
 final class DecrementInteger extends Mutator
 {
-    private const COUNT_NAME = 'count';
+    private const COUNT_NAME = ['count', 'sizeof'];
```
borNfreee: `COUNT_NAMES`?

tests/Mutator/Number/DecrementIntegerTest.php
```diff
@@ -122,6 +132,54 @@ public function provideMutationCases(): array
...
+                <<<'PHP'
+<?php
+
+if (0 == count($a)) {
+    echo 'bar';
+}
+PHP
+                ,
+            ],
+            'It does decrement zero when it is compared as greater than count() on the right side' => [
```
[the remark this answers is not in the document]
sanmai: Yeah, it does not. It only makes sense to decrement positive integers.

Ah, and I see that it won't decrement `1` in that case too.

src/Mutator/Number/DecrementInteger.php
```diff
@@ -38,37 +38,76 @@ protected function mutatesNode(Node $node): bool
...
+                $this->getLowercaseMethodName($parentNode, 'left'),
+                self::COUNT_NAME,
+                true)
+        ) {
+            return $this->isAllowedLeftSideCountComparison($parentNode);
+        }
+
+        if ($parentNode->right instanceof Node\Expr\FuncCall
+            && \in_array(
+                $this->getLowercaseMethodName($parentNode, 'right'),
```
sanmai: I get this method was before, but how is it better than just `$node->right->name->toLowerString();`... That's going to be much easier to understand tbh.

tests/Mutator/Number/DecrementIntegerTest.php
```diff
@@ -162,24 +162,6 @@ public function provideMutationCases(): array
 if (0 == count($a)) {
     echo 'bar';
 }
-PHP
-                ,
-            ],
-            'It does decrement zero when it is compared as greater than count() on the right side' => [
```
borNfreee: seems like the code still decrements `0` to `-1`, but now we don't have a test that proves it.

do we need to update the code to not decrement in this case?

src/Mutator/Number/DecrementInteger.php
```diff
@@ -18,7 +18,7 @@
  */
 final class DecrementInteger extends Mutator
 {
-    private const COUNT_NAME = 'count';
+    private const COUNT_NAMEs = ['count', 'sizeof'];
```
borNfreee: `COUNT_NAMES`, not `COUNT_NAMEs`

tests/Mutator/Number/DecrementIntegerTest.php
```diff
@@ -150,6 +190,24 @@ public function provideMutationCases(): array
 if (count($a) < -1) {
     echo 'bar';
 }
+PHP
+                ,
+            ],
+            'It does decrement when compared against a variable function' => [
```
borNfreee: could you please restore test removed in the previous commit and make sure that it does **not** decrement `0` to `-1` in this case:

```php
if (0 > count($a)) {	
    echo 'bar';	
}
```

BackEndTea: In that case the following should also not be allowed:

```diff
<?php

- if (count($a) < 0) {
+ if (count($a) < -1) {
    echo 'bar';
}
```

I'm thinking we can probably simplify this whole thing and just say: 'if count is compared to 0 in any way, don't mutate it.'

borNfreee: > if count is compared to 0 in any way, don't mutate it

yes, since result of `count()` is strictly not negative

borNfreee:
Thank you @BackEndTea

PR #487, opened 2018-09-22
Validate phpunit.xml

### LINKED ISSUE #409: Infection does not fail gracefully on an invalid phpunit.xml
For example, for this `phpunit.xml`:
```xml
<invalid>
</invalid>
```
Infection fails with:

```
PHP Fatal error:  Uncaught Error: Call to a member function setAttribute() on null in 
src/TestFramework/PhpUnit/Config/XmlConfigurationHelper.php:62
```

borNfreee:
This PR:

- [x] Fixes #409
- [x] Add `phpunit.xml` validation: by checking the root node and then by XSD schema if needed/possible. XSD can be validated with local and remote (http[s]) paths to files
- [x] Adds `ext-libxml` to the list of required extensions. This is *not a new requirement*, just made it explicit by adding to `composer.json`

### DIFF
--- a/src/TestFramework/PhpUnit/Config/Exception/InvalidPhpUnitXmlConfigException.php
+++ b/src/TestFramework/PhpUnit/Config/Exception/InvalidPhpUnitXmlConfigException.php
@@ -1,0 +1,26 @@
+<?php
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Infection\TestFramework\PhpUnit\Config\Exception;
+
+/**
+ * @internal
+ */
+final class InvalidPhpUnitXmlConfigException extends \RuntimeException
+{
+    public static function byRootNode(): self
+    {
+        return new self('phpunit.xml does not contain a valid PHPUit configuration.');
+    }
+
+    public static function byXsdSchema(string $libXmlErrorsString): self
+    {
+        return new self(sprintf('phpunit.xml file does not pass XSD schema validation. %s', $libXmlErrorsString));
+    }
+}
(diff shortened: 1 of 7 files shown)


localheinz:
Also see https://github.com/sebastianbergmann/phpunit/commit/d4484be1a9c56c73353517510ce3b0690f6d738e.

src/TestFramework/PhpUnit/Config/Exception/InvalidPhpUnitXmlConfigException.php
```diff
@@ -0,0 +1,26 @@
...
+namespace Infection\TestFramework\PhpUnit\Config\Exception;
+
+/**
+ * @internal
+ */
+final class InvalidPhpUnitXmlConfigException extends \RuntimeException
+{
+    public static function byRootNode(): self
+    {
+        return new self('phpunit.xml does not contain a valid PHPUit configuration.');
```
localheinz: Note that there is an `n`missing here.

tests/TestFramework/PhpUnit/Config/XmlConfigurationHelperTest.php
```diff
@@ -232,6 +233,77 @@ public function test_it_removes_existing_printers(): void
         );
     }
 
+    public function test_it_validates_xml_by_root_node(): void
+    {
+        $this->expectException(InvalidPhpUnitXmlConfigException::class);
+        $this->expectExceptionMessage('phpunit.xml does not contain a valid PHPUnit configuration.');
```
localheinz: How about moving this block to right before performing the action instead?

For reference, also see https://thephp.cc/news/2016/02/questioning-phpunit-best-practices.

borNfreee: Done

borNfreee:
Thank you @localheinz and @sidz

Issue #407, opened 2018-07-07
Do not mutate code that is ignored from code coverage


sebastianbergmann:
Code that is ignored from code coverage (`@codeCoverageIgnore` annotation on a class or method; `@codeCoverageIgnoreStart` and `@codeCoverageIgnoreEnd` blocks; `@codeCoverageIgnore` on a single line) should not be mutated.

sebastianbergmann:
Thanks for pointing out the workaround!

borNfreee:
partially implemented (for class and method level). Thank you @sebastianbergmann

PR #491, opened 2018-09-25
@codeCoverageIgnore annotations support

borNfreee:
This PR Fixes https://github.com/infection/infection/issues/407:

- [x] Covered by tests

There were requested 4 cases regarding PHPUnit's annotation:

- [x] `@codeCoverageIgnore` on the class level
- [x] `@codeCoverageIgnore` on the method level
- [ ] `@codeCoverageIgnore` on the line level
- [ ] `@codeCoverageIgnoreStart` and `codeCoverageIgnoreEnd` to ignore any lines of code.

This PR considers only the first two cases. Here is an explanation of why didn't I implement 3rd and 4th:

## `@codeCoverageIgnore` on the line level

There are 2 reasons why I think we should not support this case:

1. It's very tricky to implement with PHP-Parser. Look at this simple code:

```php
# test.php file
<?php

exit; // @codeCoverageIgnore
```

And this is what the AST looks like:

```php
./vendor/bin/php-parse test.php

4: Stmt_Expression[3:1 - 3:5](
    expr: Expr_Exit[3:1 - 3:4](
        expr: null
    )
)
5: Stmt_Nop[4:1 - 3:5](
    comments: array(
        0: // @codeCoverageIgnore
    )
)
```

The comment is not assigned to the `Expr_Exit` node, instead, it is assigned to the `Stmt_Nop` node that is completely another node in the tree.

And here is a `--var-dump` output of the `Stmt_Nop`:

```php
object(PhpParser\Node\Stmt\Nop)#1084 (1) {
    ["attributes":protected]=>
    array(5) {
      ["startLine"]=>
      int(4)
      ["startFilePos"]=>
      int(35)
      ["comments"]=>
      array(1) {
        [0]=>
        object(PhpParser\Comment)#1083 (4) {
          ["text":protected]=>
          string(23) "// @codeCoverageIgnore"
          ["line":protected]=>
          int(3)
          ["filePos":protected]=>
          int(12)
          ["tokenPos":protected]=>
          int(4)
        }
      }
      ["endLine"]=>
      int(3)
      ["endFilePos"]=>
      int(11)
    }
  }
```

Note that this method has `startLine`=4, `endLine`=3, and has comment which also starts on the line 3, which is a mess or PHP-Parser's bug.

2. Second reason is I don't think it's worth to write some not stable code to support this little feature. I have literally never seen a piece of code that uses `@codeCoverageIgnore` on the line level to ignore just one particular line of code.

## `@codeCoverageIgnoreStart` and `codeCoverageIgnoreEnd` 

This case is more difficult than the previous one. 

For such code:

```php
<?php

$a = 1;

// @coverageIgnoreStart
$b = 2;
$c = 3;

// @coverageIgnoreEnd

1;
```

We have the following AST:

<details>
<summary>AST dump (click me)</summary>

```php
array(
    0: Stmt_Expression[3:1 - 3:7](
        expr: Expr_Assign[3:1 - 3:6](
            var: Expr_Variable[3:1 - 3:2](
                name: a
            )
            expr: Scalar_LNumber[3:6 - 3:6](
                value: 1
            )
        )
    )
    1: Stmt_Expression[6:1 - 6:7](
        expr: Expr_Assign[6:1 - 6:6](
            var: Expr_Variable[6:1 - 6:2](
                name: b
                comments: array(
                    0: // @coverageIgnoreStart
                )
            )
            expr: Scalar_LNumber[6:6 - 6:6](
                value: 2
            )
            comments: array(
                0: // @coverageIgnoreStart
            )
        )
        comments: array(
            0: // @coverageIgnoreStart
        )
    )
    2: Stmt_Expression[7:1 - 7:7](
        expr: Expr_Assign[7:1 - 7:6](
            var: Expr_Variable[7:1 - 7:2](
                name: c
            )
            expr: Scalar_LNumber[7:6 - 7:6](
                value: 3
            )
        )
    )
    3: Stmt_Expression[11:1 - 11:2](
        expr: Scalar_LNumber[11:1 - 11:1](
            value: 1
            comments: array(
                0: // @coverageIgnoreEnd
            )
        )
        comments: array(
            0: // @coverageIgnoreEnd
        )
    )
)
```

</details>

Note that `@coverageIgnoreEnd` is assign to the `1;` line of code, and the number of code comment is 9 while the number of the `1;` is 11, which is again a little bit confusing.

Of course, we can collect all comments of all nodes when each node is visited, and then somehow understand where `@coverageIgnoreStart` and `@coverageIgnoreEnd` were placed, but this looks for me too difficult with the little value.

If you think we still should implement last two cases, I would do it as a separate PR to not make this one overcomplicated.

### DIFF
--- a/src/Mutant/Generator/MutationsGenerator.php
+++ b/src/Mutant/Generator/MutationsGenerator.php
@@ -18,6 +18,8 @@ use Infection\Mutant\Exception\ParserException;
 use Infection\Mutation;
 use Infection\Mutator\Util\Mutator;
 use Infection\TestFramework\Coverage\CodeCoverageData;
+use Infection\Visitor\CodeCoverageClassIgnoreVisitor;
+use Infection\Visitor\CodeCoverageMethodIgnoreVisitor;
 use Infection\Visitor\FullyQualifiedClassNameVisitor;
 use Infection\Visitor\MutationsCollectorVisitor;
 use Infection\Visitor\ParentConnectorVisitor;
@@ -133,9 +135,11 @@ final class MutationsGenerator
             $onlyCovered
         );
 
+        $traverser->addVisitor(new CodeCoverageClassIgnoreVisitor());
         $traverser->addVisitor(new ParentConnectorVisitor());
         $traverser->addVisitor(new FullyQualifiedClassNameVisitor());
         $traverser->addVisitor(new ReflectionVisitor());
+        $traverser->addVisitor(new CodeCoverageMethodIgnoreVisitor());
         $traverser->addVisitor($mutationsCollectorVisitor);
 
         $traverser->traverse($initialStatements);
--- a/src/Mutant/Generator/MutationsGenerator.php
+++ b/src/Mutant/Generator/MutationsGenerator.php
@@ -26,6 +26,7 @@ use Infection\Visitor\ParentConnectorVisitor;
 use Infection\Visitor\ReflectionVisitor;
 use PhpParser\Node;
 use PhpParser\NodeTraverser;
+use PhpParser\NodeVisitorAbstract;
 use PhpParser\Parser;
 use Symfony\Component\Finder\SplFileInfo;
 
@@ -83,10 +84,11 @@ final class MutationsGenerator
     /**
      * @param bool $onlyCovered mutate only covered by tests lines of code
      * @param string $filter
+     * @param NodeVisitorAbstract[] $extraNodeVisitors
      *
      * @return Mutation[]
      */
-    public function generate(bool $onlyCovered, string $filter = ''): array
+    public function generate(bool $onlyCovered, string $filter = '', array $extraNodeVisitors = []): array
     {
         $sourceFilesFinder = new SourceFilesFinder($this->srcDirs, $this->excludeDirsOrFiles);
         $files = $sourceFilesFinder->getSourceFiles($filter);
@@ -96,7 +98,7 @@ final class MutationsGenerator
 
         foreach ($files as $file) {
             if (!$onlyCovered || $this->hasTests($file)) {
-                $allFilesMutations[] = $this->getMutationsFromFile($file, $onlyCovered, $this->mutators);
+                $allFilesMutations[] = $this->getMutationsFromFile($file, $onlyCovered, $extraNodeVisitors);
             }
 
             $this->eventDispatcher->dispatch(new MutableFileProcessed());
@@ -110,11 +112,11 @@ final class MutationsGenerator
     /**
      * @param SplFileInfo $file
      * @param bool $onlyCovered mutate only covered by tests lines of code
-     * @param array $mutators
+     * @param NodeVisitorAbstract[] $extraNodeVisitors extra visitors to influence to mutation collection process
      *
      * @return Mutation[]
      */
-    private function getMutationsFromFile(SplFileInfo $file, bool $onlyCovered, array $mutators): array
+    private function getMutationsFromFile(SplFileInfo $file, bool $onlyCovered, array $extraNodeVisitors): array
     {
         try {
             /** @var Node[] $initialStatements */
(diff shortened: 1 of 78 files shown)


src/Visitor/CodeCoverageMethodIgnoreVisitor.php
```diff
@@ -0,0 +1,54 @@
...
+
+        $docComment = $method->getDocComment();
+
+        if ($docComment === false) {
+            return null;
+        }
+
+        if (strpos($docComment, '@codeCoverageIgnore') !== false) {
+            /*
+             * This is a workaround to "disable" this Node from mutation.
```
borNfreee: See https://github.com/nikic/PHP-Parser/issues/534 for a feature request to support this case

src/Mutant/Generator/MutationsGenerator.php
```diff
@@ -126,17 +128,33 @@ private function getMutationsFromFile(SplFileInfo $file, bool $onlyCovered, arra
...
-        $traverser->addVisitor(new ReflectionVisitor());
-        $traverser->addVisitor($mutationsCollectorVisitor);
+        $orderedVisitors = [
+            40 => new ParentConnectorVisitor(),
+            30 => new FullyQualifiedClassNameVisitor(),
+            20 => new ReflectionVisitor(),
+            10 => $mutationsCollectorVisitor,
+        ];
+
+        $visitorsQueue = new \SplPriorityQueue();
```
BackEndTea: I think we should move this logic out of this class. The ordering that is.

I think we can extend the NodeTraverser, and maybe opt to open an issue or PR on php-parser to get that functionality in there.

borNfreee: > I think we should move this logic out of this class.

could you please elaborate on this? move to to which class? how would it look like? what are the benefits? where to store priorities?

> I think we can extend the NodeTraverser

again - what for? I'm not a big fan of extending (read - overriding) lib's classes, so I would like to understand the purpose here.

borNfreee: > My idea would be to move this to a new class.

Done https://github.com/infection/infection/pull/491/commits/623906c13be7b5e85fb38f58d3b3e417c415a614

> That is because it is not the responsibility of this class to determine the order in which node visitors should be used.

The order of visitors was determined before this PR (in master) in this class, by ordering `->addVistor()` method calls

```php
$traverser->addVisitor(new ParentConnectorVisitor());
$traverser->addVisitor(new FullyQualifiedClassNameVisitor());
$traverser->addVisitor(new ReflectionVisitor());
$traverser->addVisitor($mutationsCollectorVisitor);
```

and I think this is a correct place to do that, because it is how PHP-Parser NodeTraverser's API works - we create an object and adds ordered visitors to influence to the AST.

If you mean we should move `->addVisitor()` method calls to one more class - I would say this would be an overkill

sanmai:
I would really enjoy having `@codeCoverageIgnore` just on the class/method level right about in the next version.  `@codeCoverageIgnoreStart` and `@codeCoverageIgnoreEnd` are not as wanted. Just saying.

Maybe this task could be merged and released in parts, what do you think?

borNfreee:
@BackEndTea I've fixed your comment, could you please review again?

src/Visitor/CodeCoverageMethodIgnoreVisitor.php
BackEndTea: With 4.1.0 released we can bump the minimum version and use that.

borNfreee:
Shit, PHPStan fails with https://travis-ci.org/infection/infection/jobs/439746807#L656 because it uses the class autoloaded from `phpstan.phar` (4.0.*` rather than from `vendor` folder.

I would ignore this by adding to `phpstan-src.neno` - `ignoreErrors` unless they either upgrade PHPParser to 4.1 or prefix their phar

src/Visitor/CodeCoverageMethodIgnoreVisitor.php
borNfreee: @BackEndTea done

PR #492, opened 2018-09-25
Replace ExampleTest namespace with unique one, same as E2E dirname

borNfreee:
Since many namespaces under e2e tests are not unique, build started to fail after I added `\ReflectionMethod` in one of the visitor.

https://travis-ci.org/infection/infection/jobs/432793693#L697-L700

This is because of class names collision.

This PR replaces an `ExampleTest` namespace with a dir name, which is unique.

Could somebody please check it on Linux? because I'm on MacOS.


borNfreee:
> 👍 How about changing all the name fixes in this PR as well?

it's already done in https://github.com/infection/infection/pull/491

This PR is targeted to that branch

PR #494, opened 2018-09-27
Change mutator API to return an object we control

BackEndTea:
This PR:

- [x] Adds the `MutatorResult` interface, which is returned by the `mutate` method.
- [x] Adds the  `GeneratorResult` & `SingleMutationResult` classes, which are implementations of `MutatorResult`
- [x] Fixes type hints on mutate doc-blocks
- [x] Moves the `Mutator` class the the `Infection` namespace.
- [x] Deleted the `InterfaceParentTrait` and moved the method into the `PublicVisibility` class

This is the implementation of the idea mentioned in #450.


src/Mutator/Util/Result/SingleMutationResult.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * @var Node|Node[]
+     */
+    private $node;
+
+    /**
+     * @param Node|Node[] $node
+     */
+    public function __construct($node)
```
BackEndTea: This doesn't have a typehint, as a 'single' mutation can also return an array of nodes to replace a node.

I could also split it up in an `ArrayMutationResult` or something like that.

BackEndTea:
> Thank you for contributing to Infection. We noticed you didn't add any tests. Could you please add them to make sure everything works as expected?

Weird, i did add some tests, and the bot didn't pick up on it.

src/Mutator/Util/Result/MutatorResult.php
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types=1);
+
+namespace Infection\Mutator\Util\Result;
+
+use PhpParser\Node;
+
+/**
+ * @internal
+ */
+interface MutatorResult
```
sidz: My 5 cents:
What about to call an Interface like `MutatedNode` and create same class. Because `MutatorResult` it's smth abstract and doesn't reflect what is it and what for.

borNfreee: I also agree that `MutatorResult` is too common name, but `MutatedNode` is also confusing when `\Generator` is used, even when `Node[]` is used (because it's not singular but plural, while `MutatedNode` is strictly singular)

What if we rename it in a way as it is a strategy, not the result of something:

```php
interface NodeReplacer {
    public function replaceWithNodes(): \Generator; // or getReplacedNodes(): \Generator
}
```

#### For a single Node 

```php
class SingleNodeReplacer implements NodeReplacer {
    public function __construct(Node $node) { ... }

    public function replaceWithNodes(): \Generator {
        yield $this->node;
    }
}
```

#### for array of Nodes

```php
class ArrayNodeReplacer implements NodeReplacer {
    public function __construct(Node[] $node) { ... }

    public function replaceWithNodes(): \Generator {
        yield $this->nodes;
    }
}
```
#### for Generator

```php
class GeneratorNodeReplacer implements NodeReplacer {
    public function __construct(\Generator $nodes) { ... }

    public function replaceWithNodes(): \Generator {
        return $this->nodes;
    }
}
```

client coude would be

```php
...
$nodeReplacer = $mutator->mutate($node);

foreach ($nodeReplacer->replaceWithNodes()  as $node) { ... }

// or

foreach ($nodeReplacer->getReplacedNodes()  as $node) { ... }

File: src/Exception/InvalidMutatorException.php [hunk 3]
```diff
@@ -9,7 +9,7 @@
 
 namespace Infection\Exception;
 
-use Infection\Mutator\Util\Mutator;
```
borNfreee: any reason behind it?

src/Mutator/Util/Result/MutatorResult.php
borNfreee: Probably, the previous comment is too much.

Another ideas:

* `MutatedAst` as an interface name (which implies singular/plural)

Or, if we continue PHP-Parser wording (see [`NodeTraverser`](https://github.com/nikic/PHP-Parser/blob/4fb48c0e18ed69bb60ec27137e0875b0e23637e9/lib/PhpParser/NodeTraverser.php#L100-L162)), we can name it `NodeReplacement`:

```php
abstract public function mutate(Node $node): NodeReplacement;
```

This one also implies singular and plural.

* `SingleNodeReplacement implements NodeReplacement`
* `ArrayNodeReplacement implements NodeReplacement`
* ...

src/Mutator/Util/Result/MutatorResult.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    /**
+     * @return \Generator|Node[]
+     */
+    public function getResults(): \Generator;
```
sanmai: Why not `iterable`?

src/Mutator/Util/Result/SingleMutationResult.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $this->node = $node;
+    }
+
+    public function getResults(): \Generator
+    {
+        yield $this->node;
+    }
```
sanmai: With an `iterable` typehint this could return a simple array.

borNfreee:
> because one would have to learn two or three more things more than they used to learn to just add a new mutator.

this is a price for avoiding `mixed` return type. As I said in the previous PR, seeing `@return Node|Node[]|\Generator` is like "what the hell the author was doing..". Though this is of course just my subjective opinion.

Also, I doubt that anyone "used to learn" anything about mutators. Mutators are closed at the moment, we don't expose public interfaces to write custom mutators, and 90%+ of mutators are written by core team. So for the majority of new developers this will be just new thing, no need to re-learn.

> With this change that will be ever so slightly more difficult

We can create abstract classes to return that simplicity back again, example for the `SingleMutator`:

```php
abstract class SingleMutator extends Mutator
{
    abstract protected function getMutatedNode(Node $node): Node;

    public function mutate(Node $node): NodeReplacement
    {
        return SingleNodeReplacement($this->getMutatedNode($node);
    }
}
```

And then adding new mutator will be easy:

```php
final class Plus extends SingleMutator
{
    /**
     * Replaces "+" with "-"
     *
     * @param Node $node
     *
     * @return Node\Expr\BinaryOp\Minus
     */
    protected function getMutatedNode(Node $node): Node
    {
        return new Node\Expr\BinaryOp\Minus($node->left, $node->right, $node->getAttributes());
    }

    protected function mutatesNode(Node $node): bool
    {
        // ...
    }
}
```

We will end up with 3 new abstract classes with clear goals, and the main method `mutate()` will finally have a return type.

sanmai:
> seeing `@return Node|Node[]|\Generator` is like "what the hell the author was doing.."

Yeah, that's kind of funny return declaration, I'd have a similar reaction to it if I wouldn't know what means what. I see the point, that's going to make things easier to maintain in the long-medium term.

(Mutators are _not_ closed at the moment in a sense there are folks who may be planning new mutators right about now, like you see with #500.)

src/Exception/InvalidMutatorException.php
sanmai: It seems like the logic behind this is that this is a major class/interface that deserves to be at the root of the similarly named namespace.

sanmai:
I wonder if we can move forward with this.

src/Exception/InvalidMutatorException.php
borNfreee: > deserves to be at the root of the similarly named namespace

sounds reasonable

> Maybe moving them all one directory level down would be a better bet?

I'm fine with it, the only one note here is to update these lines

https://github.com/infection/infection/blob/ef45b8e6f6eec00821b385ba670a4ca95c2a9d20/tests/Mutator/Util/MutatorProfileTest.php#L42-L43

probably, it's out of the scope of this PR

BackEndTea:
@infection/core sorry about the delay on this PR, i've updated it to use 3 abstract mutator classes that can be used depending on what implementation is needed.

src/Mutator/Arithmetic/Assignment.php
```diff
@@ -9,26 +9,24 @@
...
     /**
      * Replaces "+=", "*=", ".=", and similar with a plain "="
      *
      * @param Node|Node\Expr\AssignOp $node
      *
-     * @return MutatorResult
+     * @return Node
      */
-    public function mutate(Node $node): MutatorResult
+    public function getMutatedNode(Node $node): Node
```
borNfreee: This should be `protected`, not `public`. 

`abstract protected function getMutatedNode(Node $node): array;` in the parent

Same for all classes.

BackEndTea:
After some reconsideration i've decided to abandon this PR, it doesn't seem worth it.

PR #496, opened 2018-10-01
Enable lean releases

raphaelstolt:
Keeps the release packages lean.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,4 +5,24 @@
 *.jpg binary
 
 #Files we do not want to distribute
-/tests export-ignore
+.ci/                 export-ignore
+.editorconfig        export-ignore
+.gitattributes       export-ignore
+.github/             export-ignore
+.gitignore           export-ignore
+.php_cs.dist         export-ignore
+.travis.yml          export-ignore
+appveyor.yml         export-ignore
+box.json.dist        export-ignore
+build/               export-ignore
+CHANGELOG.md         export-ignore
+codecov.yml          export-ignore
+composer.lock        export-ignore
+devTools/            export-ignore
+infection.json.dist  export-ignore
+LICENSE              export-ignore
+Makefile             export-ignore
+phpunit.xml.dist     export-ignore
+README.md            export-ignore
+tests/               export-ignore
+setup_environment.sh export-ignore


.gitattributes
```diff
@@ -5,4 +5,24 @@
...
+codecov.yml          export-ignore
+composer.lock        export-ignore
+devTools/            export-ignore
+infection.json.dist  export-ignore
+LICENSE              export-ignore
+Makefile             export-ignore
+phpunit.xml.dist     export-ignore
+README.md            export-ignore
+tests/               export-ignore
+setup_environment.sh export-ignore
```
localheinz: Maybe sort

- by type (directories first)
- then alphabetically 

?

borNfreee: @raphaelstolt sorry for bothering you, are you going to finish this PR or can we take over?

raphaelstolt: Feel free to sort the ignored artifacts to your liking and commit them in this PR.

raphaelstolt:
@theofidry Sorry I can't really follow your argumentation. Maybe this [tweet](https://twitter.com/raphaelstolt/status/914850874811396096) sums up the intention of maintained `.gitattributes` files. Adding it as an requirement to the `CONTRIBUTING.md` might also reduce the cost of maintenance and there's even tooling support for enforcing lean packages via Travis CI.

After all it's your project so feel free to reject or modify the PR.

theofidry:
What I'm saying is, unless there is tooling to help to maintain `.gitattributes` files (in which case they would be more than welcome), having to list _everything_ is just tedious and realistically is not gonna happen and the `gitattributes` file will always be out of sync. So my suggestion is, instead of trying to be exhaustive, the `gitattributes` should be limited to the important directories, e.g. `devTools` and `tests`. You're free to do numbers but I would wager all the small files listed are likely to be 1-5% of the total weight you would remove with the directories mentioned above.

In other words: I'm all for the idea, but due to the lack of tooling we got to stay pragmatic and keep it simple.

borNfreee:
@raphaelstolt what exact tools are you talking about? probably we can integrate them to this repository

raphaelstolt:
@theofidry and @borNfreee Here's a shameless plug for such a [tool](https://github.com/raphaelstolt/lean-package-validator) I wrote myself, which might be integrated into your `.gitattributes` maintenance workflow. It should work pretty good with [PDS](https://github.com/php-pds/skeleton) conform packages.

borNfreee:
nice tool btw! 

I'm for adding it to our pipeline in Travis. However, what about creating a PHAR archive, personally I'm not a big fan of adding such tools to `require/require-dev` sections of `composer.json`, and all existing tools we use are also PHARs.

borNfreee:
Closing since https://github.com/infection/infection/pull/532 has been merged (with sorted folders/files)

Feel free to open a PR to run `lean-package-validator ` as PHAR on CI.

Anyway, thank you for your work @raphaelstolt.

Issue #498, opened 2018-10-02
Public to Private Visibility Mutator


sanmai:
We have mutators that...
- Replaces "public function..." with "protected function ..."
- Replaces "protected function..." with "private function ..."

Yet we don't have a mutator that would change "public function..." into "private function..." that could just tell that a public function can have a private visibility.

For example, if there is such a function, one has to go though loops first changing a public function into protected only to hear from Infection, again, that this function can be made private.

PR #500, opened 2018-10-02
Mutate public to private methods

deleugpn:
This PR:

- [x] Adds new feature ...
- [x] Covered by tests
- [x] Doc PR https://github.com/infection/site/pull/68

Fixes https://github.com/infection/infection/issues/498

This PR adds the ability to mutate from `public` straight to `private` methods.

### DIFF
--- a/src/Mutator/FunctionSignature/PublicVisibility.php
+++ b/src/Mutator/FunctionSignature/PublicVisibility.php
@@ -24,19 +24,31 @@ final class PublicVisibility extends Mutator
     use InterfaceParentTrait;
 
     /**
-     * Replaces "public function..." with "protected function ..."
+     * Replaces "public function..." with "protected function ..." and "private function ..."
      *
      * @param Node $node
      *
-     * @return ClassMethod
+     * @return \Generator
      */
     public function mutate(Node $node)
     {
         /* @var ClassMethod $node */
-        return new ClassMethod(
+        yield new ClassMethod(
             $node->name,
             [
                 'flags' => ($node->flags & ~Class_::MODIFIER_PUBLIC) | Class_::MODIFIER_PROTECTED,
+                'byRef' => $node->returnsByRef(),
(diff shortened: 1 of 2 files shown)


sanmai:
1. CI is failing because you need to run E2E tests and update fixtures for them (as now they see more mutations). The latter is typically done with `./tests/e2e_tests`.
2. I'd make a `private function makeClassMethod($visiblity): ClassMethod` and avoid copy'n'pasting a bunch of code, what do you think?

deleugpn:
Sorry, I did this PR at GitHub opening event in Amsterdam and forgot to check back the status. I'll look into the e2e test.
I swear I tried making a private method as you said and it was breaking the mutation.
I'll try again =D

sanmai:
Please do. I'm here if you need a hand.

src/Mutator/FunctionSignature/PublicVisibility.php
```diff
@@ -24,16 +24,16 @@
...
      * @param Node $node
      *
-     * @return ClassMethod
+     * @return \Generator
      */
     public function mutate(Node $node)
     {
         /* @var ClassMethod $node */
-        return new ClassMethod(
+        yield new ClassMethod(
```
miguelgf: What improvement adds the yield here instead of the return?

borNfreee: this is a new feature (not yet released). It allows to return `N` mutations from one mutator. Note the second mutation on line 48 below.

Also, this is not the same as `reutnr [new ClassMethod(...), new ClassMethod(...)]`. Such code replaces 1 `Node` with an array of `Node`s, which is completely different behaviour.

borNfreee:
Are you going to finish this PR @deleugpn or can we take over it?

deleugpn:
Sorry, you can take it over. I hope to be back at some point.

PR #502, opened 2018-10-12
Update ProtectedVisibility/PublicVisibility to guard against missing reflection

### LINKED ISSUE #501: PublicVisibility mutator failing due to missing reflection
| Question    | Answer
| Infection version | 0.10.5 (as a Composer package)
| Test Framework version | PHPUnit 7.4.0
| PHP version | PHP 7.2.2 (cli)
| Platform    | OSX Sierra
| Github Repo | https://github.com/humbug/php-scoper/pull/265

<details>
<summary>Complete output</summary>

```bash
$ php fixtures/set020-infection/vendor/infection/infection/bin/infection --coverage=dist/infection-coverage
You are running Infection with xdebug enabled.
    ____      ____          __  _

Running initial test suite...

PHPUnit version: 7.4.0

  723 [============================] 4 secsProcessing source code files:  0/44

Generate mutants...

Processing source code files:  8/44
In InvalidMutatorException.php line 21:

  Encountered an error with the "PublicVisibility" mutator in the "/Users/tfidry/Project/Humbug/php-scoper/src/functions.php" file. This is most likely a bug in Infection, so please report this in our
  issue tracker.


In PublicVisibility.php line 103:

  Call to a member function getInterfaces() on null


run [--test-framework TEST-FRAMEWORK] [--test-framework-options TEST-FRAMEWORK-OPTIONS] [-j|--threads THREADS] [--only-covered] [-s|--show-mutations] [-c|--configuration CONFIGURATION] [--coverage COVERAGE] [--mutators MUTATORS] [--filter FILTER] [--formatter FORMATTER] [--min-msi MIN-MSI] [--min-covered-msi MIN-COVERED-MSI] [--log-verbosity LOG-VERBOSITY] [--initial-tests-php-options INITIAL-TESTS-PHP-OPTIONS] [--ignore-msi-with-no-mutations] [--debug]
```
</details>

The issue seems to be caused by that following piece of code:

```php
// src/functions.php from humbug/php-scoper

function create_scoper(): Scoper
{
    return (new class() extends ApplicationFactory {
        public static function createScoper(): Scoper
        {
            return parent::createScoper();
        }
    })::createScoper();
}
```

sanmai:
This happens because `ReflectionVisitor` adds a class reflection only for non-anonymous, named, classes. For an anonymous class we need to create an actual instance of that class, which seems to be infeasible for now.

https://github.com/infection/infection/blob/c4ebe3eeb1f0e4f435b43131b799be025e217f6a/src/Visitor/ReflectionVisitor.php#L47-L49

PHP Parser has nothing to do with that issue.

This PR:

- [x] Fixes #501
- [x] Covered by tests

### DIFF
--- a/src/Mutator/FunctionSignature/ProtectedVisibility.php
+++ b/src/Mutator/FunctionSignature/ProtectedVisibility.php
@@ -65,6 +65,11 @@ final class ProtectedVisibility extends Mutator
         /** @var \ReflectionClass $reflection */
         $reflection = $node->getAttribute(ReflectionVisitor::REFLECTION_CLASS_KEY);
 
+        if (!$reflection instanceof \ReflectionClass) {
+            // assuming the worst where interface has the same method
+            return true;
+        }
+
         $parent = $reflection->getParentClass();
 
         while ($parent) {
--- a/src/Mutator/FunctionSignature/ProtectedVisibility.php
+++ b/src/Mutator/FunctionSignature/ProtectedVisibility.php
@@ -62,11 +62,11 @@ final class ProtectedVisibility extends Mutator
 
     private function hasSameProtectedParentMethod(Node $node): bool
     {
-        /** @var \ReflectionClass $reflection */
+        /** @var \ReflectionClass|null $reflection */
         $reflection = $node->getAttribute(ReflectionVisitor::REFLECTION_CLASS_KEY);
 
         if (!$reflection instanceof \ReflectionClass) {
-            // assuming the worst where interface has the same method
+            // assuming the worst where a parent class has the same method
             return true;
         }
 
--- a/src/Mutator/FunctionSignature/PublicVisibility.php
+++ b/src/Mutator/FunctionSignature/PublicVisibility.php
@@ -92,11 +92,11 @@ final class PublicVisibility extends Mutator
 
     private function hasSamePublicParentMethod(Node $node): bool
     {
-        /** @var \ReflectionClass $reflection */
+        /** @var \ReflectionClass|null $reflection */
         $reflection = $node->getAttribute(ReflectionVisitor::REFLECTION_CLASS_KEY);
 
         if (!$reflection instanceof \ReflectionClass) {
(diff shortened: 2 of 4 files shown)


src/Mutator/FunctionSignature/ProtectedVisibility.php
```diff
@@ -65,6 +65,11 @@ private function hasSameProtectedParentMethod(Node $node): bool
         /** @var \ReflectionClass $reflection */
         $reflection = $node->getAttribute(ReflectionVisitor::REFLECTION_CLASS_KEY);
 
+        if (!$reflection instanceof \ReflectionClass) {
+            // assuming the worst where interface has the same method
+            return true;
+        }
```
sanmai: PHPStan says that a strict comparison using `===` between `ReflectionClass` and `null` will always evaluate to false.

sanmai:
@theofidry np, please see updated description; PHP Parser is not at fault here.

src/Mutator/FunctionSignature/ProtectedVisibility.php
theofidry: I think this is because you have `/** @var \ReflectionClass $reflection */` instead of `/** @var \ReflectionClass|null $reflection */`.

I'm not against the instanceof check, however I admit I would prefer `!($reflection instanceof \ReflectionClass)`. That said it's personal preference over CS so pay no mind to it if you disagree

theofidry:
>Not sure if there's anything else can be done here, what do you all think?

Your patch looks good to me

src/Mutator/FunctionSignature/ProtectedVisibility.php
sanmai: I'm not the one to disagree! This is definitely more proper because leaves no window for error.

sanmai:
Is it okay that we're being pessimistic here? 

Being optimistic can result in extra erroneous mutations, but we may miss interesting and useful mutations.

PR #507, opened 2018-10-15
Explicitly add the default profile to the list of mutators

sanmai:
As even if the list is seemingly empty by default, it is not. If it would actually contain the default profile by default, it would make the profiles feature more obvious to configure.

Right now [it is not immediately obvious from the documentation](https://infection.github.io/guide/profiles.html#The-Profiles) that one has to add the default profile to the list, you have to actually read what is written. With this small change a user could start adding and changing mutators right away, just after skimming the example configuration provided.

- [x] Simplifies usage of Infection
- [x] Covered by tests


borNfreee:
thank you @sanmai

PR #508, opened 2018-10-15
Do not travers plain functions unless they are in the method or closures

### LINKED ISSUE #466: Do not mutate the code inside plain functions
Related https://github.com/infection/infection/pull/456

Example: when there is a closure inside a plain function, current Infection mutates it. It does not make any sense.

```php
function property($path): callable
{
    return function ($value, $index = 0, $collection = []) use ($path, $propertyAccess) {
          // should NOT be mutated
    };
}
```

On the other side, we can mutate closures that are on the global scope:

```php
// global scope

$filtered = array_filter([1, 2, 3], function ($value) {
    return $value % 2; // we CAN mutate here
});
```

borNfreee:
This PR prevents Infection from mutating any code inside plain function, unless function is in the method or closure.

- [x] Fixes #466
- [x] Covered by tests

https://github.com/infection/infection/issues/466#issuecomment-419304479

> On the other hand, we should definitely mutate functions that do not exist in the global scope at the time of a mutation. E.g. if function_exists($functionName) returns false, then our mutations will affect tests, and will be useful. Although I'm not entirely sure that is going to happen any often.

I'm not sure about this, because Infection can be unaware of project's bootstrap file that can be used to load functions. I would rather not traverse plain function at all, simpler and less confusing. However, feel free to discuss it.

### DIFF
--- a/src/Visitor/ReflectionVisitor.php
+++ b/src/Visitor/ReflectionVisitor.php
@@ -10,6 +10,7 @@ declare(strict_types=1);
 namespace Infection\Visitor;
 
 use PhpParser\Node;
+use PhpParser\NodeTraverser;
 use PhpParser\NodeVisitorAbstract;
 
 /**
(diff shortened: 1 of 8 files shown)


borNfreee:
thank you

PR #513, opened 2018-10-17
Implement UnwrapArrayMap and UnwrapArrayFilter mutators

### LINKED ISSUE #490: Function call unwrapping mutators
[A fine example from our Slack channel](https://symfony-devs.slack.com/archives/C78M7A4BY/p1537651009000100) by @borNfreee:

```diff
- $lowercasedStrings = array_map(['L', 'O', 'L'], function ($v) {
-    return strtolower($v);
- });
+ $lowercasedStrings = ['L', 'O', 'L'];
```
And just as such for `array_filter`. There could be more similar mutators which all will be looking like:

```diff
- $output = standardFunction($input, $foo, $bar);
+ $output = $input;
```

akondas:
This PR:

- [x] Create `AbstractUnwrapMutator`
- [x] Implement `UnwrapArrayFilter` and `UnwrapArrayMap` mutators
- [x] Covered by tests
- [ ] Doc PR: see https://github.com/infection/site/pulls for inspiration

Inspiration in #490 

Fixes #490

### DIFF
--- a/src/Mutator/Unwrap/AbstractUnwrapMutator.php
+++ b/src/Mutator/Unwrap/AbstractUnwrapMutator.php
@@ -1,0 +1,42 @@
+<?php
+/**
+ * Copyright © 2017-2018 Maks Rafalko
+ *
+ * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Mutator\Unwrap;
+
+use Infection\Mutator\Util\Mutator;
+use PhpParser\Node;
+
+/**
+ * @internal
+ */
+abstract class AbstractUnwrapMutator extends Mutator
+{
+    abstract protected function getFunctionName(): string;
+
+    abstract protected function getParameterIndex(): int;
+
+    /**
+     * Replaces "$a = function(arg1, arg2);" with "$a = arg1;"
+     *
+     * @return Node\Param;
+     */
+    public function mutate(Node $node)
+    {
+        return $node->args[$this->getParameterIndex()];
+    }
+
+    protected function mutatesNode(Node $node): bool
+    {
+        if (!$node instanceof Node\Expr\FuncCall) {
+            return false;
+        }
+
+        return $node->name->toString() === $this->getFunctionName();
+    }
+}
--- a/src/Mutator/Unwrap/AbstractUnwrapMutator.php
+++ b/src/Mutator/Unwrap/AbstractUnwrapMutator.php
@@ -26,17 +26,17 @@ abstract class AbstractUnwrapMutator extends Mutator
      *
      * @return Node\Param;
      */
-    public function mutate(Node $node)
+    final public function mutate(Node $node)
     {
         return $node->args[$this->getParameterIndex()];
     }
 
-    protected function mutatesNode(Node $node): bool
+    final protected function mutatesNode(Node $node): bool
     {
         if (!$node instanceof Node\Expr\FuncCall) {
             return false;
         }
 
-        return $node->name->toString() === $this->getFunctionName();
+        return $node->name->toLowerString() === strtolower($this->getFunctionName());
     }
 }
--- a/src/Mutator/Util/MutatorProfile.php
+++ b/src/Mutator/Util/MutatorProfile.php
@@ -33,6 +33,7 @@ final class MutatorProfile
         '@sort' => self::SORT,
         '@zero_iteration' => self::ZERO_ITERATION,
         '@cast' => self::CAST,
+        '@unwrap' => self::UNWRAP,
 
         //Special Profiles
         '@default' => self::DEFAULT,
@@ -159,6 +160,11 @@ final class MutatorProfile
         Mutator\Cast\CastString::class,
     ];
 
+    public const UNWRAP = [
+        Mutator\Unwrap\UnwrapArrayMap::class,
+        Mutator\Unwrap\UnwrapArrayFilter::class,
+    ];
+
     public const DEFAULT = [
         '@arithmetic',
         '@boolean',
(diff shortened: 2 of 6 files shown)


sanmai:
These seem like fine examples for our new generating mutatiors:

https://github.com/infection/infection/blob/68f70ad62ffd14af62781d989e4230c45c235010/src/Mutator/Arithmetic/RoundingFamily.php#L42-L48

That way you can also cover `array_flip()`, `array_reverse()`, `array_replace()`, and others, without much extra effort. Sure, tests will be longer, but that's unavoidable.

src/Mutator/Unwrap/AbstractUnwrapMutator.php
```diff
@@ -0,0 +1,42 @@
...
+    abstract protected function getFunctionName(): string;
+
+    abstract protected function getParameterIndex(): int;
+
+    /**
+     * Replaces "$a = function(arg1, arg2);" with "$a = arg1;"
+     *
+     * @return Node\Param;
+     */
+    public function mutate(Node $node)
```
sanmai:
```suggestion
    final public function mutate(Node $node)
```

tests/Mutator/Unwrap/UnwrapArrayFilterTest.php
```diff
@@ -0,0 +1,59 @@
...
+
+$a = ['A', 1, 'C'];
+PHP
+        ];
+
+        yield [
+            <<<'PHP'
+<?php
+
+$a = array_filter(['A', 1, 'C'], function($var): bool {
```
borNfreee: could you please add more tests (like parameter as a variable, function with lower/upper cased characters and so on.)

See the list of cases in https://github.com/infection/infection/blob/master/tests/Mutator/Regex/PregQuoteTest.php

same for `array_map`

src/Mutator/Unwrap/AbstractUnwrapMutator.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return $node->args[$this->getParameterIndex()];
+    }
+
+    protected function mutatesNode(Node $node): bool
```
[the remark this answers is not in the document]
akondas: Definietly :)

tests/Mutator/Unwrap/UnwrapArrayFilterTest.php
akondas: Nice catch, wrongly capitalized was the problem :)

sanmai: There's `$node->name->toLowerString()` of use.

tests/Mutator/Unwrap/UnwrapArrayFilterTest.php
```diff
@@ -0,0 +1,153 @@
...
+<?php
+
+$a = ['A', 1, 'C'];
+if ($a === $a) {
+    return true;
+}
+PHP
+        ];
+
+        yield 'It mutates correctly when array_map is wrongly capitalized' => [
```
sanmai:
```suggestion
        yield 'It mutates correctly when array_filter is wrongly capitalized' => [
```

tests/Mutator/Unwrap/UnwrapArrayFilterTest.php
```diff
@@ -0,0 +1,153 @@
...
+PHP
+            ,
+            <<<'PHP'
+<?php
+
+$a = ['A', 1, 'C'];
+PHP
+        ];
+
+        yield 'It mutates correctly when array_map uses another function as input' => [
```
sanmai:
```suggestion
        yield 'It mutates correctly when array_filter uses another function as input' => [
```

akondas:
Can we restart CI jobs? `Failed loading C:\tools\php\ext\xdebug.dll` it looks like an unreliable error.

borNfreee:
Thank you @akondas, nice work.

The build is green now

borNfreee:
@akondas can I ask you to update documentation, please? See https://github.com/infection/site/issues/70

borNfreee:
I just realized that with this new base abstract class, we can implement not only array functions mutators as mentioned [here](https://github.com/infection/infection/issues/490#issuecomment-426157145), but for any other function, for example

```diff
- return strtolower($string);
+ return $string;
```

👍

sanmai:
Yeah, and things like:
```
- return array_reduce($array, $func, $initial);
+ return $initial;
```

Should we make a list within an issue with ideas?

borNfreee:
good idea, done https://github.com/infection/infection/issues/514

src/Mutator/Util/MutatorProfile.php
```diff
@@ -159,6 +160,11 @@
         Mutator\Cast\CastString::class,
     ];
 
+    public const UNWRAP = [
+        Mutator\Unwrap\UnwrapArrayMap::class,
+        Mutator\Unwrap\UnwrapArrayFilter::class,
```
localheinz: @akondas @borNfreee @sanmai 

Is the order important here?

If not, maybe we can sort alphabetically?

src/Mutator/Util/MutatorProfile.php
```diff
@@ -278,5 +284,9 @@
         'CastInt' => Mutator\Cast\CastInt::class,
         'CastObject' => Mutator\Cast\CastObject::class,
         'CastString' => Mutator\Cast\CastString::class,
+
+        // Unwrap
+        'UnwrapArrayMap' => Mutator\Unwrap\UnwrapArrayMap::class,
+        'UnwrapArrayFilter' => Mutator\Unwrap\UnwrapArrayFilter::class,
```
localheinz: @akondas @borNfreee @sanmai 

Is the order important here?

If not, maybe we can sort alphabetically?

src/Mutator/Util/MutatorProfile.php
sanmai: Same order [as in the manual](http://php.net/manual/en/book.array.php) should work best, IMO.

Issue #396, opened 2018-06-29
[RFC] Run the project test suite with random order


borNfreee:
Extracted from Slack and originally raised by @BackEndTea

#### Problem

There can be a situation when unit test can implicitly depend on other unit test (by human mistake or any other reason) and changing the order of executed tests leads to fail. 

Infection runs only several unit tests for each Mutant, with the order of execution time (fastest - first), so if the tests fails because of incorrect (not original) order, we have a killed Mutant which is obviously not correct.

#### Proposal

We can run unit tests in a random order during our "Run initial test suite" stage in Infection. By doing it, we can check that each test is independent and if the test suite fails with random order, we can notify user about this issue with a detailed error message.

Since PHPUnit 7.2, we can do it out of the box ([related PR](https://github.com/sebastianbergmann/phpunit/pull/3092), [documentation](https://github.com/epdenouden/phpunit/wiki/PHPUnit-test-running-order-management))

What do you think about adding this feature?

BackEndTea:
👍 As long as its properly documented.

For its implementation i'd do the following

Checks (in order)
* PHPUnit  7.3 or higher, use `--order-by=random` flag
* PHPUnit 7.2 use the `--random-order` flag, which is deprecated in 7.3
* Otherwise: check if `fiunchinho/phpunit-randomizer` is installed, and run that
* Run normal order

Or, we could create our own version of this so we can always run in random order. This does mean that we have to work with actual phpunit code, instead of just its cli. Meaning we will have more issues in supporting different versions.

I'm not sure if this is possible or relevant for phpspec.

sanmai:
Great idea! We should do it, but not without a proper explanation for a user.

deleugpn:
If I understood this correctly, people providing code coverage to infection would skip this feeature and perhaps should be warned about it.

PR #519, opened 2018-10-23
Run project's tests in a random order for InitialTestRun process

borNfreee:
This PR changes the way Infection run the whole project's test suite. Now, if PHPUnit version is `7.2+`, tests will be executed in the random order.

- [x] Covered by tests

The order of the tests are controlled by `phpunit.xml`'s attributes:

* `executionOrder="random"`
* `resolveDependencies="true"` 

#### Why do we need `resolveDependencies`?

If some tests depend on each other through `@depends` annotation, then running them in a random order will lead to [skipped broken tests](https://github.com/epdenouden/phpunit/wiki/PHPUnit-test-running-order-management#demo-3-handling-dependencies). To avoid reordering such dependent tests and still use randomness, we should use `resolveDependencies="true"` (or `--resolve-dependencies` option)

Read more: https://github.com/epdenouden/phpunit/wiki/PHPUnit-test-running-order-management#demo-3-handling-dependencies

Fixes https://github.com/infection/infection/issues/396


tests/Fixtures/e2e/PhpUnit_Depends_Order/tests/SourceClassTest.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * If one of the tests with `@depends` annotation is reordered, and this reorder is caused by
+     * incorrectly used `executionOrder="random"` flag (or `--order=random` option), then PHPUnit automatically
+     * skips such test case.
+     *
+     * When one of the test is skipped, we get self::$counter < 4, which will trigger an error.
+     *
+     * Expected result: random order of these tests should respect `@depends` annotation and do not
+     * produce skipped tests, thanks to `resolveDependencies="true"`flag (or `--resolve-dependencies` option)
```
borNfreee:
```suggestion
     * produce skipped tests, thanks to `resolveDependencies="true"` flag (or `--resolve-dependencies` option)
```

src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
```diff
@@ -21,7 +21,6 @@ public function build(string $configPath, string $extraOptions): array
         return array_filter([
             '--configuration',
             $configPath,
-            '--stop-on-failure',
```
sanmai: Why do we need to continue after a failure?

borNfreee: we don't, this flag is added on XML level (covered by tests btw), and this is just a duplication that I found

src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
```diff
@@ -170,4 +171,26 @@ private function createNode(\DOMDocument $dom, string $nodeName): \DOMElement
...
         return $node;
     }
+
+    private function addRandomTestsOrderAttributes(string $version, \DOMXPath $xPath): void
+    {
+        if (!version_compare($version, '7.2', '>=')) {
+            return;
+        }
+
+        $this->updateOrAddAttribute('executionOrder', 'random', $xPath);
```
sanmai: I feel like this may be suboptiomal for some projects because people can get used to hidden dependencies (e.g. tests executed in order of definition, communicating through instance's variable; heck, even PHPUnit has such nasty example [right in the manual](https://phpunit.readthedocs.io/en/7.4/fixtures.html)!). Therefore I feel like it will be more appropriate to check if there's already an `executionOrder` attribute with any value (it could be `default` or `reverse`), not overwriting it if it is set.

borNfreee:
> I feel like this may be suboptiomal for some projects because people can get used to hidden dependencies

And that's the problem. Hidden dependencies, relying on execution order - all of these should be considered as a bad practise and `unresolved design problem` as [PHPUnit documentation states](https://phpunit.readthedocs.io/en/7.4/fixtures.html).

> even PHPUnit has such funny example right in the manual!

Not sure what example you are referring to. Example 4.1 does not share instance variable, it's reset for each test case.

--------

What I want to say is that **random execution order is a must for Infection**. Why? Because we run subset of the tests in a random order for each Mutant (not really random, but sorted by fastest first).

https://github.com/infection/infection/blob/6b800242c88187b72e7bd6690071933ddb9e0877/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php#L167-L173

What does it mean? It means that if there are hidden dependencies, then such tests will fail in any case. Thus, hidden dependencies should be fixed OR author should clearly understand and agree with such failures.

In other words, running tests in a random order should be opt-out (current implementation), not opt-in (what is suggested above).

I like the idea of adding an option to disable  random order for those who

* have a large codebase and don't want or is not able to fix hidden dependencies
* clearly understand and agree with incorrectly killed Mutants

> Therefore I feel like it will be more appropriate to check if there's already an executionOrder attribute with any value (it could be default or reverse), not overwriting it if it is set.

I'm not sure we need to keep e.g. `reverse` order. What is the point of `reverse` order for Infection? Infection will run all those tests in a random (fastest - first) order for the very first Mutant.

### Summary

I suggest to

* keep random order as a default behaviour for Infection, because it is a must for proper mutation process
* add an option to *keep original order*, be it `default`, `reverse`, whatever
* force people to fix hidden dependencies in their test suites because this is the only right way

What do you think?

sanmai:
Section around Example 4.1 does not even mention hidden dependencies. This is one argument that our users are not aware about this problem. 

What you say is a sound argument. I wonder why wouldn't PHPUnit itself run tests in random order?.. It appear to me it is their responsibility to ensure good test writing practice. 

> I'm not sure we need to keep e.g. reverse order.

I would not even guess, but if that's what our user wants, so be it!

Not sure what you say about an option. If that's about command line option, I feel like this solution if by far inferior to actually checking for a relevant property in `phpunit.xml`. Most recent PHPUnit is still very new, so we can guess that most folks not even using this option. Therefore, in most cases Infection will run tests in random order.

src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
```diff
@@ -65,4 +65,20 @@ public function getName(): string
...
+
+    public function getInitialTestsFailRecommendations(string $commandLine): string
+    {
+        $recommendations = parent::getInitialTestsFailRecommendations($commandLine);
+
+        if (version_compare($this->getVersion(), '7.2', '>=')) {
+            $recommendations = sprintf(
+                "%s\n\n%s",
+                "Infection runs the test suite in a RANDOM order. Make sure your tests do not have hidden dependencies.\n\n" .
+                'You can add these attributes to `phpunit.xml` to check it: <phpunit executionOrder="random" resolveDependencies="true" ...',
```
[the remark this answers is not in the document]
borNfreee: done

src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
borNfreee: done

borNfreee:
> So I think it must be like this:
> 
> * keep random order as a default behaviour for Infection, because it is a must for proper mutation process
> * keep original order, be it default, reverse, whatever, but only if it was set

done, could you please review it again?

borNfreee:
> I'd add a test here that really depends on a predefined order, by a hidden dependency. Thus we will know for sure that Infection does not overwrite the order tag

done

borNfreee:
Thank you @borNfreee for the review

sanmai:
No problem. This new e2e test seems all right.

PR #512, opened 2018-10-17
Moving mutated php code to it's own file

deleugpn:
Hey guys, first I'd like to apologize for my PR #500 which is quite late already. I'm trying to better organize my time to finish that one.
Today I got a spare minute and tried to fix the conflict that it has, but it was quite hard for me. Not only that, but also when I was actually working on adding the new mutations there I noticed the difficulty of adding new mutated code into the test files.

This PR is a Proof Of Concept proposal to use the `getFileContent` method not only for the original code, but also for the mutated one. The mutated code would be suffixed by `mx` where `x` is the number of the mutation (e.g. `pv-final-m0.php`, `pv-final-m1.php`, etc...). The reason for the suffix is that with the `yield` system we might generate more than 1 mutation per original code.

If this gets a positive feedback, I'll finish moving the entire file to this system and then rework PR #500 which would be a lot easier.

PR #521, opened 2018-10-25
Move ProtectedVisibility Mutators to dedicated file

deleugpn:
Carrying on the work started on #512


sanmai:
This all can be even better if a test would scan for files in an automatically detected directory, then run tests against those files. Adding a news test will be a matter of adding a file or two. And if it could see that there is only `m0.php` file, of if there's a pair of `m0.php` `m1.php` then it could also automatically check for correct mutations. What do you think?

borNfreee:
> And if it could see that there is only `m0.php` file, of if there's a pair of `m0.php` `m1.php` then it could also automatically check for correct mutations.

I think this will not be clear for new contributors, and even for us in a month or two. As for me - it's better to leave as is - explicitly and without magic.

deleugpn:
I also thought about how to make it less brittle, but as borNfreee mentioned it would be 'too magical' at this point. I was thinking about just keeping it like this for now and as we get more into  this routine, perhaps another refactoring could come in and improve the way of writing these tests even more.

borNfreee:
Closing as the very similar PR was also closed https://github.com/infection/infection/pull/512

Feel free to reopen if you find some time and still thinks it's needed.

PR #522, opened 2018-10-26
Prevent using symfony/console with broken exit code handling

duncan3dc:
There's a [bug](https://github.com/symfony/symfony/issues/28666) in symfony/console 3.4.16 and 4.1.5 where the exit code is reported as success if it failed.

This obviously affects this tool in a major way rendering it useless in a CI environment, so this PR ensures this bug cannot be installed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,6 +54,7 @@
         "sort-packages": true
     },
     "conflict": {
+        "symfony/console": "3.4.16,4.1.5",
         "symfony/process": "3.4.2"
     },
     "require": {
--- a/composer.lock
+++ b/composer.lock
@@ -1,10 +1,10 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "07ac6eb17366576ae5ce9758f0421c17",
+    "content-hash": "6138f69ed34077ab4485a8d1781e44f4",
     "packages": [
         {
             "name": "composer/ca-bundle",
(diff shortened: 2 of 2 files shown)


sanmai:
That’s impressive. We definitely need an E2E test for a case of failing Testing program. (Not sure if we had one.)

borNfreee:
Thanks for creating this PR!

1. Is this bug in `3.4.16` only or in all versions prior it?
2. We have `composer.lock` committed to this repo, and we are locked with `3.4.10` version (see https://github.com/infection/infection/blob/master/composer.lock#L530) but as far as I see we are not affected. Do you have an idea why?

duncan3dc:
It was introduced in 3.4.16 and 4.1.5 so doesn't affect prior versions

borNfreee:
What if we just add

```
"conflict": {
        "symfony/console": "3.4.16,4.1.5"
},
```

to `composer.json? We already have this exception for `symfony/process`

https://github.com/infection/infection/blob/10f36ab2e7040442ccd4aae8806515da4fe8b89b/composer.json#L56-L58

With suchapproach, we can install any version except those two, and stay with the current one which works fine.

duncan3dc:
I've switched it to use `conflicts` now and updated `composer.lock` for you

borNfreee:
I'm sorry for bothering you again, but now build fails because PHPStan complains about type mismatches.

This is because of this change https://github.com/symfony/console/compare/v3.4.10..v3.4.17#diff-0fbcb6c63b7c2e0022d2b94ab7762465R125

I would recommend either

* revert `composer.lock` changes since we decided to use `conflict` section
* or fix PHPStan error, which probably what you want to do ;)

Also, `conflict: "3.4.16,4.1.5"` does not work. I experimented a little bit with it, and got a working solution:

```
"symfony/console": "=3.4.16 || =4.1.5"
```

borNfreee:
Merged in https://github.com/infection/infection/pull/523

Thank you @duncan3dc for your help!

PR #506, opened 2018-10-14
Update TestFrameworkFinder to always look for a .bat first.

sanmai:
Fixes #503

PR #523, opened 2018-10-27
Restrict installation with broken versions of symfony/console

sanmai:
- [x] Use `conflict` approach from #522
- [x] Also ignore bogus PHPStan warnings. (You can't avoid updating composer.lock, and these are quite cumbersome to fix.)
- [x] Backport #506 as it is required for some builds to complete.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,6 +54,7 @@
         "sort-packages": true
     },
     "conflict": {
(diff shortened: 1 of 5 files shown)


borNfreee:
Thank you @sanmai

Issue #451, opened 2018-08-13
Add json-schema validation for `infection.json.dist` config file


borNfreee:
Related:

* https://github.com/infection/infection/issues/206
* https://github.com/infection/infection/issues/419

Also https://github.com/infection/infection/issues/204#issuecomment-370335823

> How about providing a JSON Schema and then using [justinrainbow/json-schema](https://github.com/justinrainbow/json-schema) to validate it, after ensuring we actually have valid JSON?

sidz:
As we will implement schema validation for our config file it looks like that it shouldn't be possible to allow Infection to continue processed if the configuration file doesn't exist (like we do currently)

https://github.com/infection/infection/blob/master/src/Console/InfectionContainer.php#L233-L240

this `catch` generates valid empty json which will pass the next condition and then InfectionCommand will trigger `configure` command for case when configuration file doesn't exist

https://github.com/infection/infection/blob/master/src/Command/InfectionCommand.php#L298-L313

Because an empty config file shouldn't be valid for us and at least `source.directories` option should be required (if you thing that we need to add more required fields please give me a nudge) suggest to stop doing this and start think that an empty json is invalid for us.

With this idea it will mean that we should show a message for case when Infection can't find a config file and show a message like psalm, composer do.

An example from psalm:
```
Could not locate a config XML file in path /var/www/infection/. Have you run 'psalm --init' ?
```
Composer:
```
Composer could not find a composer.json file in /var/www/composer
```

thoughts?

borNfreee:
> it shouldn't be possible to allow Infection to continue processed if the configuration file doesn't exist

This is a huge DX feature from my point of view and this is very convinient to run infection for the first time and let it automaticallu run `configure` command for you, and then continue without any other steps.

So, my vote is to save this functionality despite any technical decisions.

We can generate `{"source": {"directories": ["src"]}}` instead of `{}` and tweak it during the `configure` phase by our Guessers, or use Guessers at this point. Does it make sense?

-----

PS shitty permalink-bot, I have uninstalled it because it breaks line numbers in URLs

borNfreee:
implemented

PR #524, opened 2018-10-29
Implement schema validation for the config file

sidz:
This PR implements schema validation for the config file which was discussed before in the https://github.com/infection/infection/issues/451

I created JsonFile class with the hard coded schema file path constant because imo it will be only the one json file in the Infection for the nearest future. If it will be required we may refactor it. (Give me a shout if we need to do it more flexible)

Also I didn't add validation for the mutators profiles and particular classes. For case like: 
```
"mutators": {
        "@default": true,
        "@function_signature": false,
        "TrueValue": {
            "ignore": [
                "NameSpace\\*\\SourceClass::method",
                "Full\\NameSpaced\\Class"
            ]
        }
    }
```

### DIFF
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -1,0 +1,104 @@
+{
+    "$schema": "https://json-schema.org/draft-07/schema#",
+    "type": "object",
+    "additionalProperties": false,
+    "required": [
+        "timeout",
+        "source"
+    ],
+    "properties": {
+        "timeout": {
+            "type": "integer",
+            "description": "The allowed timeout in seconds configured for Infection."
+        },
+        "source": {
+            "type": "object",
+            "required": [
+                "directories"
+            ],
+            "additionalProperties": false,
+            "properties": {
+                "directories": {
+                    "type": "array",
+                    "minItems": 1,
+                    "description": "Contains all folders with source code you want to mutate.",
+                    "items": {
+                        "type": "string"
+                    }
+                },
+                "excludes": {
+                    "type": "array",
+                    "minItems": 1,
+                    "description": "Contains all folders or files you want to exclude withing your source folders.",
+                    "items": {
+                        "type": "string"
+                    }
+                }
+            }
+        },
+        "logs": {
+            "type": "object",
+            "minProperties": 1,
+            "additionalProperties": false,
+            "properties": {
+                "text": {
+                    "type": "string",
+                    "description": "Human readable text log file."
+                },
+                "summary": {
+                    "type": "string",
+                    "definition": "Summary log file, which display the amount of mutants per category, (Killed, Errored, Escaped, Timed Out & Not Covered)."
+                },
+                "debug": {
+                    "type": "string",
+                    "description": "Debug log file, which displays what mutations were found on what line, per category."
+                },
+                "perMutator": {
+                    "type": "string",
+                    "description": "Markdown file which will give a break-down of the effectiveness of each mutator."
+                },
+                "badge": {
+                    "type": "object",
+                    "additionalProperties": false,
+                    "properties": {
+                        "branch": {
+                            "type": "string",
+                            "description": "Mutation score badge for your github project."
+                        }
+                    }
+                }
+            }
+        },
+        "tmpDir": {
+            "type": "string",
+            "description": "Folder where Infection creates its configs, caches and other stuff."
+        },
+        "phpUnit": {
+            "type": "object",
+            "description": "PHPUnit additional settings.",
+            "additionalProperties": false,
+            "properties": {
+                "configDir": {
+                    "type": "string",
+                    "description": "Path to directory containing PHPUnit configuration file."
+                },
+                "customPath": {
+                    "type": "string",
+                    "description": "Custom path to PHPUnit executable."
+                }
+            }
+        },
+        "mutators": {
+            "type": "object",
+            "description": "Contains the settings for different mutations and profiles"
+        },
+        "testFramework": {
+            "type": "string",
+            "description": "Sets the framework to use for testing. Defaults to phpunit."
+        },
+        "bootstrap": {
+            "type": "string",
+            "description": "Use to specify a file to include as part of the startup to pre-configure the infection environment. Useful for adding custom autoloaders not included in composer."
+        }
+    }
+}
--- a/src/Console/InfectionContainer.php
+++ b/src/Console/InfectionContainer.php
@@ -35,9 +35,8 @@ declare(strict_types=1);
 
 namespace Infection\Console;
 
-use Infection\Config\Exception\InvalidConfigException;
+use Infection\Config\ConfigCreatorFacade;
 use Infection\Config\InfectionConfig;
-use Infection\Config\Validator;
 use Infection\Differ\DiffColorizer;
 use Infection\Differ\Differ;
 use Infection\EventDispatcher\EventDispatcher;
@@ -226,10 +225,6 @@ final class InfectionContainer extends Container
             return new MemoryFormatter();
         };
 
-        $this['infection.config.validator'] = function (): Validator {
-            return new Validator();
-        };
-
         $this['memory.limit.applier'] = function (): MemoryLimiter {
             return new MemoryLimiter($this['filesystem'], \php_ini_loaded_file());
         };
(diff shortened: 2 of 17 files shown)


resources/schema.json
```diff
@@ -0,0 +1,105 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
```
BackEndTea: How about making this https?

src/Json/Exception/JsonValidationException.php
```diff
@@ -0,0 +1,33 @@
...
+
+use JsonSchema\Exception\ValidationException;
+
+/**
+ * @internal
+ */
+final class JsonValidationException extends ValidationException
+{
+    /**
+     * @param string $path
```
BackEndTea: I think we can safely remove this doc, what do you think?

src/Json/Exception/ParseException.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Infection\Json\Exception;
+
+/**
+ * @internal
+ */
+final class ParseException extends \RuntimeException
+{
+    /**
+     * @param string $path
```
BackEndTea: Same for this doc block

src/Console/InfectionContainer.php
```diff
@@ -217,46 +212,9 @@ private function getInfectionConfig(): InfectionConfig
...
-                );
-            }
-
-            // getcwd() may return false in rare circumstances
-            \assert(\is_string($configLocation));
-
-            $infectionConfig = new InfectionConfig($config, $this['filesystem'], $configLocation);
-
-            $this['infection.config.validator']->validate($infectionConfig);
+            $facade = new ConfigCreatorFacade($this['locator'], $this['filesystem']);
```
BackEndTea: :+1: For Moving this kinda stuff to its own class

src/Json/Exception/JsonValidationException.php
localheinz: @BackEndTea @sidz 

Perhaps we should enable the `no_superfluous_phpdoc_tags` fixer?

sidz: agree :) This came from the my current project where doc block is required everytime :)

src/Json/JsonFile.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @param string $path
+     */
+    public function __construct(string $path)
+    {
+        $this->path = $path;
+    }
+
+    private function validateSchema(): void
```
borNfreee: Suggest to move all private methods to the bottom of the file

src/Finder/LocatorInterface.php
```diff
@@ -0,0 +1,34 @@
...
+    public function locate(string $name): string;
+
+    /**
+     * Determine the realpath for the bunch of file path.
+     *
+     * @param array $fileNames
+     *
+     * @return string
+     */
+    public function locateAnyOf(array $fileNames): string;
```
localheinz: Perhaps `locateOneOf()` would be better? After all, the method returns a single path, not an arbitrary number.

src/Json/JsonFile.php
localheinz: @borNfreee 

There's actually the `ordered_class_elements ` fixer that does that, but it will of course also apply all of these changes to existing classes.

resources/schema.json
```diff
@@ -0,0 +1,105 @@
...
+            "description": "Folder where Infection creates its configs, caches and other stuff."
+        },
+        "phpUnit": {
+            "type": "object",
+            "description": "PhpUnit additional settings.",
+            "additionalProperties": false,
+            "properties": {
+                "configDir": {
+                    "type": "string",
+                    "description": "Custom directory path with phpunit.xml.dist file."
```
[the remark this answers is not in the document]
sidz: make sense. All descriptions have been copied from the official site: https://infection.github.io/guide/usage.html

probably will need to change description there as well

src/Json/JsonFile.php
sidz: I did in this PR but yeah we may enable `ordered_class_elements` in the separate PR like @localheinz suggests

localheinz: @borNfreee @sidz 

It's already enabled, actually, see:

https://github.com/infection/infection/blob/6b800242c88187b72e7bd6690071933ddb9e0877/.php_cs.dist#L58-L64

We could further configure it, though!

src/Finder/LocatorInterface.php
BackEndTea: Maybe even `locateFirstOf()`?

sidz: let stay with `oneOf`

src/Finder/LocatorInterface.php
```diff
@@ -0,0 +1,34 @@
...
+     * @param string $name
+     *
+     * @return string
+     */
+    public function locate(string $name): string;
+
+    /**
+     * Determine one realpath for the bunch of file paths.
+     *
+     * @param array $fileNames
```
borNfreee: `string[]`?

sidz:
@borNfreee done. Also squashed 😎

src/Finder/LocatorInterface.php
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Infection\Finder;
+
+/**
+ * @internal
+ */
+interface LocatorInterface
+{
+    /**
+     * Determine the realpath of the file by the file path.
```
BackEndTea: Should this have an `@throws LocatorException` doc comment?

And the same for `locateOneOf`?

tests/Config/ConfigCreatorFacadeTest.php
```diff
@@ -0,0 +1,141 @@
...
+        $this->filesystemMock = $this->createMock(Filesystem::class);
+
+        $this->creatorFacade = new ConfigCreatorFacade($this->locator, $this->filesystemMock);
+
+        $this->filesystem = new Filesystem();
+        $this->workspace = sys_get_temp_dir() . \DIRECTORY_SEPARATOR . 'infection-test' . \microtime(true) . \random_int(100, 999);
+        $this->tmpDir = (new TmpDirectoryCreator($this->filesystem))->createAndGet($this->workspace);
+    }
+
+    protected function tearDown(): void
```
BackEndTea: I'm not 100% sure, but is `tearDown` always called after every test? Even if the test failed?

tests/Config/ConfigCreatorFacadeTest.php
```diff
@@ -0,0 +1,141 @@
...
+            ->willReturn($customConfigPath);
+
+        $this->filesystem->dumpFile($customConfigPath, $json);
+
+        $infectionConfig = $this->creatorFacade->createConfig($customConfigPath);
+
+        self::assertSame($infectionConfig->getProcessTimeout(), 25);
+    }
+
+    public function test_it_creates_empty_config_class_successfully(): void
```
BackEndTea: If we are testing the creation of an empty config class, why is  there json dumped into the config file?

src/Json/JsonFile.php
```diff
@@ -0,0 +1,106 @@
...
+    {
+        $validator = new Validator();
+
+        $schemaFile = self::SCHEMA_FILE;
+
+        // Prepend with file:// only when not using a special schema already (e.g. in the phar)
+        if (false === strpos($schemaFile, '://')) {
```
BackEndTea: I don't think i've seen this before, could you explain why it is needed for this piece, but not for other times where we interact with the filesystem?

tests/Config/ConfigCreatorFacadeTest.php
sanmai: It should be, unless there was a fatal error of sorts.

tests/Config/ConfigCreatorFacadeTest.php
sidz: will change the names of these methods. Locator should return smth like it found some configuration file.

src/Json/JsonFile.php
sidz: good question. don't remember why I did this

sidz: https://travis-ci.org/infection/infection/jobs/451027361 here is the reason :smile:

sidz:
@borNfreee could you please kick off the Appveyor build again?

borNfreee:
done

borNfreee:
Thank you @sidz

sidz:
:hugs:

PR #526, opened 2018-10-29
Fix: Article

localheinz:
This PR

* [x] fixes articles in keys for data provider sets

Follows #513.


localheinz:

PR #528, opened 2018-10-30
Include the complete license in headers

sanmai:
- [x] Modify PHP-CS-Fixer configuration to include the license verbatim in headers as required by the terms of the license. Specifically, the license says that "redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer."

Some licenses like Apache's [permit to include only a predefined header](https://www.apache.org/licenses/LICENSE-2.0#apply), where [this license requires full text to be included to comply](https://github.com/infection/infection/blob/10f36ab2e7040442ccd4aae8806515da4fe8b89b/LICENSE#L9-L10) with the terms of the license in case of re-distribution, e.g. when using files from a project in another project. So it is also a matter of courtesy to not to force any re-distributors to copy and paste the license manually if they wish to use a file from this project in theirs.

We're using PHPDoc-style header comment, and modern IDE can hide it by default, so it should not be a problem to include the whole terms. Eclipse can do that [at least from 2011](https://github.com/eclipse/pdt/commit/04f64af9101f18b8d5c5a0958aa8c0b37effe745).

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,11 +1,8 @@
 <?php
 
-$year = date('Y');
-$header = <<<TXT
-Copyright © 2017-$year Maks Rafalko
+$header = trim('This code is licensed under the BSD 3-Clause License.'.substr(file_get_contents('LICENSE'), strlen('BSD 3-Clause License')));
 
-License: https://opensource.org/licenses/BSD-3-Clause New BSD License
-TXT;
+
 $finder = \PhpCsFixer\Finder::create()
     ->in('src')
     ->in('tests')
--- a/src/Command/BaseCommand.php
+++ b/src/Command/BaseCommand.php
@@ -1,8 +1,34 @@
 <?php
 /**
- * Copyright © 2017-2018 Maks Rafalko
+ * This code is licensed under the BSD 3-Clause License.
  *
- * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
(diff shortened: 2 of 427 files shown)


.php_cs.dist
```diff
@@ -1,11 +1,8 @@
 <?php
 
-$year = date('Y');
```
borNfreee: shouldn't we retain this behavior to force update the year in 2019?

.php_cs.dist
```diff
@@ -1,11 +1,8 @@
...
-$header = <<<TXT
-Copyright © 2017-$year Maks Rafalko
+$header = trim('This code is licensed under the BSD 3-Clause License.'.substr(file_get_contents('LICENSE'), strlen('BSD 3-Clause License')));
```
localheinz: 👌

.php_cs.dist
localheinz: @borNfreee 

Ha, when the time comes, only this needs to be updated:

https://github.com/infection/infection/blob/6b800242c88187b72e7bd6690071933ddb9e0877/LICENSE#L3

borNfreee:
could you please rebase on master so we can merge it immediately? there are new files in master that will not pass fixer's check

sanmai:
@borNfreee  done

borNfreee:
Thank you @sanmai

PR #532, opened 2018-10-30
Update gitattributes file

gmponos:
This PR:

- When you do a composer install the package downloads also these files. By adding them to gitattributes it will not download them anymore.

On an unrelated issue, is it correct to have composer.lock submitted on a "package" like this. The policy most of the open source packages is not to submit the composer.lock file and leave the developer to be able to install in some cases even the latest versions.


sanmai:
Seemingly related to #496

.gitattributes
```diff
@@ -5,4 +5,16 @@
...
+/.codecov.yml         export-ignore
+/.editorconfig        export-ignore
+/.gitattributes       export-ignore
+/.github              export-ignore
+/.gitignore           export-ignore
+/.php_cs.dist         export-ignore
+/.travis.yml          export-ignore
+/infection.json.dist  export-ignore
+/phpunit.xml.dist     export-ignore
+/tests                export-ignore
```
localheinz: Seeing https://github.com/infection/infection/blob/bd4cd3208186e14e67857a29a59901b76221b739/.editorconfig#L7 I can recommend using an IDE which is equipped with https://editorconfig.org/#download. Alternatively, if you are using PhpStorm, the corresponding setting to ensure a line feed at end of file can be manually enabled:

screen shot 2018-10-31 at 08 42 12

.gitattributes
```diff
@@ -5,4 +5,16 @@
 *.jpg binary
 
 #Files we do not want to distribute
-/tests export-ignore
\ No newline at end of file
+/.appveyor            export-ignore
```
localheinz: What do you think about sorting by

* kind (directories, files)
* name

?

Also note that some of the entries do not match existing files or directories.

```diff
diff --git a/.gitattributes b/.gitattributes
index dc46e75..7044b87 100644
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,17 +4,21 @@
 *.png binary
 *.jpg binary

-#Files we do not want to distribute
-/.appveyor            export-ignore
-/.build               export-ignore
-/.ci                  export-ignore
-/.codecov.yml         export-ignore
+# Directories and files we do not want to distribute
+/.build/              export-ignore
+/.ci/                 export-ignore
+/.github/             export-ignore
+/devTools/            export-ignore
+/tests/               export-ignore
 /.editorconfig        export-ignore
 /.gitattributes       export-ignore
-/.github              export-ignore
 /.gitignore           export-ignore
 /.php_cs.dist         export-ignore
 /.travis.yml          export-ignore
+/appveyor.yml         export-ignore
+/box.json.dist        export-ignore
+/codecov.yml          export-ignore
 /infection.json.dist  export-ignore
+/Makefile             export-ignore
 /phpunit.xml.dist     export-ignore
-/tests                export-ignore
\ No newline at end of file
+/setup_environment.sh export-ignore
```

gmponos: I think i applied the changes.

.gitattributes
gmponos: Yea I know this plugins and features of PHPStorm, but I think it without a new line at master and I kept it this way.

.gitattributes
```diff
@@ -4,5 +4,21 @@
...
\ No newline at end of file
+# Directories and files we do not want to distribute
+/.build/              export-ignore
+/.ci/                 export-ignore
+/.github/             export-ignore
+/devTools/            export-ignore
+/tests/               export-ignore
+/.editorconfig        export-ignore
+/.gitattributes       export-ignore
+/.github              export-ignore
```
localheinz: This is probably not needed, can we remove it? We already have a rule above for `/.github/`.

gmponos: 😃 you are correct.

done.

borNfreee:
Thank you @gmponos

PR #538, opened 2018-11-01
Enhancement: Fail with a better assertion message

localheinz:
This PR

- [x] fails with a better assertion message

💁‍♂️ Noticed this when working on the unwrap mutators. Maybe even extract a method?


borNfreee:
Thank you @localheinz

localheinz:
Thank you, @borNfreee and @sidz!

PR #541, opened 2018-11-03
Fix: Keep mutators in mutator profiles sorted by name

localheinz:
This PR

- [x] asserts that mutators in mutator profiles are sorted by name
- [x] sorts mutators in mutator profiles by name

💁‍♂️ Unless the order is relevant, it's probably good to keep them sorted by name.


borNfreee:
Thank you @localheinz

localheinz:
Thank you, @borNfreee!

PR #546, opened 2018-11-04
Fix: Use plural for method name as it returns a generator

localheinz:
This PR

- [x] renames a method returning a generator

💁‍♂️ I think it makes sense as the generator potentially yields more than one index.


borNfreee:
Thank you @localheinz

localheinz:

PR #552, opened 2018-11-06
Permit for occasional skipped E2E test due to Composer's failure to install dependencies

sanmai:
Packagist is a free service and is not 100% reliable as one may guess, but since we run multiple same tests at once in different environments, if we occasionally skip a test for reasons we can't control, it should do no harm. [E.g. in case of this kind of failures.](https://ci.appveyor.com/project/borNfreee/infection/builds/20071579/job/benm8m46h7d8mbwb#L215)

- [x] Also raise the process timeout up to 5 minutes.

### DIFF
--- a/tests/Console/E2ETest.php
+++ b/tests/Console/E2ETest.php
@@ -44,6 +44,7 @@ use PHPUnit\Framework\TestCase;
 use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Console\Output\BufferedOutput;
 use Symfony\Component\Finder\Finder;
+use Symfony\Component\Process\Exception\ProcessTimedOutException;
 use Symfony\Component\Process\Process;
 
 /**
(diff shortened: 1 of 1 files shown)


tests/Console/E2ETest.php
```diff
@@ -184,8 +188,23 @@ private function installComposerDeps(): void
             // Install deps only if there's none
             $this->assertNotEmpty(getenv('PATH') ?: getenv('Path'), 'E2E tests need a system composer installed, but it could not be found without a PATH set');
 
-            $process = new Process(sprintf('%s %s', (new ComposerExecutableFinder())->find(), 'install'));
-            $process->mustRun();
+            try {
+                $process = new Process(sprintf('%s %s', (new ComposerExecutableFinder())->find(), 'install'));
```
sidz: Passing a command as string is deprecated since symfony 4.2
https://github.com/symfony/process/blob/master/Process.php#L146

could you please change this behaviour on our side?

sanmai: I'll make another PR with all such instances fixed.

borNfreee:
Thank you @sanmai

PR #553, opened 2018-11-06
Enhancement: Unwrap all available arguments to array_map()

localheinz:
This PR

- [x] unwraps all additional arguments to `array_map()`

Follows #513.


localheinz:

PR #569, opened 2018-11-14
Add file to DebugFileLogger

gmponos:
This PR:

- [x] Bug

At the moment debug log prints something like this:

```
Total: 1737
Killed mutants:
===============


Mutator: PublicVisibility
Line 33

Mutator: PublicVisibility
Line 25
```

As you can understand it does not make any sense without the file name,

### DIFF
--- a/src/Logger/DebugFileLogger.php
+++ b/src/Logger/DebugFileLogger.php
@@ -82,7 +82,8 @@ final class DebugFileLogger extends FileLogger
         foreach ($processes as $mutantProcess) {
             $logParts[] = '';
             $logParts[] = 'Mutator: ' . $mutantProcess->getMutator()::getName();
-            $logParts[] = 'Line ' . $mutantProcess->getOriginalStartingLine();
+            $logParts[] = 'File: ' . $mutantProcess->getOriginalFilePath();
+            $logParts[] = 'Line: ' . $mutantProcess->getOriginalStartingLine();
         }
 
         return implode("\n", $logParts) . "\n";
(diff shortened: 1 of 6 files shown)


borNfreee:
I'm curious do we need a `debug` logger at all in this case. After this change, it will be like this:

```bash
Mutator: PublicVisibility
File: /path/to/infection/src/Command/BaseCommand.php
Line 40
```

And this is almost what we already have in a `text` logger, only without a diff:

```bash
Escaped mutants:
================

1) /path/to/infection/src/Command/BaseCommand.php:40    [M] PublicVisibility

--- Original
+++ New
@@ @@
-    public function getContainer() : Container
+    protected function getContainer() : Container
```

Is the only reason of creating `debug` mutator was to use it in tests @BackEndTea (exactly to skip file paths)? If so, I would mark this logger as internal and remove from the documentation. 

People will use `text` logger. If you @gmponos don't need a diff in your logs, we can introduce a new log verbosity value instead: https://infection.github.io/guide/command-line-options.html#log-verbosity

gmponos:
> If you @gmponos don't need a diff in your logs, we can introduce a new log verbosity value instead

This PR was not created or driven by necessity. What I mean is that at the moment I am also discovering this library. I was experimenting with it and I had in my config to print all available logs. When I saw this particular kind of log I though that most probably there is a bug and I tried to contribute back with fixing it.

To tell you the truth TextLogger is my best choice including the diff.. so the solution you propose about marking it as internal I believe to be the best. Want me to revert and proceed?

gmponos:
Closing as the logger is already marked as internal.. opened related PR for docs.

PR #571, opened 2018-11-15
Fix: Debugger is called Xdebug

localheinz:
This PR

- [x] fixes a small spelling issue

💁‍♂️ For reference, https://xdebug.org:

> This release only updates **Xdebug** to support PHP 7.3. There are a few known issues that still need to be addressed, such as a problem with Code Coverage and multiple catches. These will be addressed in a future beta or Release Candidate.


borNfreee:
Thank you @localheinz

localheinz:
Thank you, @borNfreee, @sanmai, and @sidz!

PR #574, opened 2018-11-18
Remove Mockery as dev dependency

sidz:
This PR removes Mockery from the dev dependencies and replaces on equivalent calls from phpunit

### DIFF
--- a/tests/Process/Runner/InitialTestsRunnerTest.php
+++ b/tests/Process/Runner/InitialTestsRunnerTest.php
@@ -59,9 +59,8 @@ final class InitialTestsRunnerTest extends TestCase
     {
         $this->eventDispatcher = $this->createMock(EventDispatcherInterface::class);
 
+        /** @var MockObject|Process $process */
         $process = $this->createMock(Process::class);
-        $process->expects($this->never())
-            ->method('stop');
 
         $process->expects($this->once())
             ->method('run')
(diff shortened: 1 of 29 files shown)


tests/Process/Runner/InitialTestsRunnerTest.php
```diff
@@ -41,32 +41,51 @@
...
-        $process = Mockery::mock(Process::class);
-        $process->shouldReceive('run');
+        $this->eventDispatcher = $this->createMock(EventDispatcherInterface::class);
+
+        /** @var MockObject|Process $process */
+        $process = $this->createMock(Process::class);
+
+        $process->expects($this->once())
+            ->method('run')
+            ->will($this->returnCallback(function ($type) use ($process): void {
```
sidz: If anybody has idea how to do it better fill free to let me know :)

tests/Process/Runner/InitialTestsRunnerTest.php
```diff
@@ -41,32 +41,53 @@
...
     {
-        $process = Mockery::mock(Process::class);
-        $process->shouldReceive('run');
+        $this->eventDispatcher = $this->createMock(EventDispatcherInterface::class);
+
+        /** @var MockObject|Process $process */
+        $process = $this->createMock(Process::class);
+
+        $process->expects($this->once())
+            ->method('run')
```
borNfreee: as discussed:

```php
$process->expects($this->once())
            ->method('run')
            ->with($this->callback(function ($type): bool {
                $type(Process::OUT);

                return true;
            }));
```

borNfreee:
Thank you @sidz

PR #575, opened 2018-11-19
Run Infection only for changed files in PRs

borNfreee:
During working on [the first how-to article](https://infection.github.io/guide/how-to.html#How-to-run-Infection-only-for-changed-files) in our docs, I realized that we can run Infection only for the changed files in this repository as well.

This PR brings at least the following benefits:

1. Firstly, all the new code will have the minimum MSI of `90%` (we can change this number up to `100%`). While increasing MSI for the whole project is [quite tricky](https://github.com/infection/infection/issues/344), we can easily control MSI for the changed files only. 

    When PR will contain only new files, the author will must write the proper tests without any exceptions.

    When PR will contain updated file, the author will have to update current tests in order to achieve `90%`. But, in comparison to current situation, the author of PR will have to do it only for the limited set of files that they has changed.

2. Secondly, this PR significantly reduces the build time on Travis. Let's compare the build from this branch with the recently executed build from #574:

    | Build  | Real Build Time  | Machine Build Time  |
    |---|---|---|
    |  [This branch](https://travis-ci.org/infection/infection/builds/457120502) | 20 min 56 sec  | 1 hr 11 min 57 sec  |
    |  [remove-mockery-as-dev-dependency](https://travis-ci.org/infection/infection/builds/457087662?utm_source=github_status&utm_medium=notification) | 53 min 37 sec  | 2 hrs 55 min 50 sec  |

    So its **~2.5x** faster. Each jobs takes 5-7 minutes.

3. Killing the Mutants becomes much easier just because the number of them:

    ```bash
    .......                                              (7 / 7)
    7 mutations were generated:
           7 mutants were killed
           0 errors were encountered
           0 time outs were encountered
    ```
    
    Only 7 Mutants for the changed files in this PR. This will definitely help contributors killing Mutants more efficiently without reading the log file with hundreds (or thousands in our case) of mutations.

### DIFF
--- a/.ci/travis-functions.sh
+++ b/.ci/travis-functions.sh
@@ -19,3 +19,17 @@ function xdebug-enable() {
         phpenv config-add /tmp/xdebug.ini
     fi
 }
+
+function get-infection-pr-flags() {
+    if [[ "${TRAVIS_PULL_REQUEST}" == "false" ]]; then
+        INFECTION_PR_FLAGS="";
+    else
+        git remote set-branches --add origin $TRAVIS_BRANCH;
+        git fetch;
+
+        CHANGED_FILES=$(git diff origin/$TRAVIS_BRANCH --diff-filter=AM --name-only | grep src/ | paste -sd "," -);
+        INFECTION_PR_FLAGS="--filter=${CHANGED_FILES} --ignore-msi-with-no-mutations --only-covered --min-msi=90";
+    fi
+
+    echo $INFECTION_PR_FLAGS;
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ addons:
       - expect
 
 before_install:
-  - source .ci/xdebug.sh
+  - source .ci/travis-functions.sh
   - xdebug-disable
 
 jobs:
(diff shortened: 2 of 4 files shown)


.ci/travis-functions.sh
```diff
@@ -18,4 +18,18 @@ function xdebug-enable() {
...
         phpenv config-add /tmp/xdebug.ini
     fi
+}
+
+function get-infection-pr-flags() {
+    if [[ "${TRAVIS_PULL_REQUEST}" == "false" ]]; then
+        INFECTION_PR_FLAGS="";
+    else
+        git remote set-branches --add origin $TRAVIS_BRANCH;
+        git fetch;
```
BackEndTea: Do we need to fetch everything? We might aswell only fetch the branches we need here

borNfreee:
```bash
git clone --depth=50 --branch=remove https://github.com/infection/infection.git infection

git branch
# output 
* remove

git remote set-branches --add origin master
git fetch;

# From https://github.com/infection/infection
# * [new branch]      master     -> origin/master

git branch
# output
* remove
```

it doesn't fetch everything as I can see

sanmai: `git fetch` has no `--depth=` here

sanmai: Please never mind, `git fetch` should respect shallow clone

BackEndTea: My bad, i wrongly assumed `git fetch` would fetch all branches

Issue #564, opened 2018-11-10
all mutators are escaped


achrafSallemi:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.9.3
| Test Framework version | 6.5.13
| PHP version | 7.0
| Platform    |  MacOS
| Github Repo | -

<details>
 <summary>infection.json.dist</summary>
 
 ```
{
    "timeout": 10,
    "bootstrap": "./bootstrap.php",
    "tmpDir": "tests/logs/infection",
    "testFramework": "phpunit",
    "source": {
        "directories": [
            "app"
        ]
    },
    "logs": {
        "text": "tests/logs/infection/infection-log.txt",
        "summary": "tests/logs/infection/summary.log",
        "debug": "tests/logs/infection/debug.log",
        "perMutator": "tests/logs/infection/per-mutator.md"
    },
    "phpUnit": {
        "configDir": "."
    }
}
 ```
</details>

<details>
 <summary>phpunit.xml</summary>
 
 ```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.5/phpunit.xsd"
        backupGlobals="false"
        backupStaticAttributes="false"
        bootstrap="./bootstrap.php"
        addUncoveredFilesFromWhitelist="true"
        convertErrorsToExceptions="true"
        convertNoticesToExceptions="true"
        convertWarningsToExceptions="true"
        colors="true"
        strict="true"
        verbose="true">
    <logging>
        <log type="junit" target="./tests/logs/phpunit/junit.xml"/>
        <log type="coverage-clover" target="./tests/logs/phpunit/coverage-clover.xml"/>
        <log type="coverage-php" target="./tests/logs/phpunit//unittest.cov"/>
        <log type="coverage-text" target="php://stdout" showUncoveredFiles="false"/>
        <log type="coverage-html" target="./tests/logs/phpunit/coverage-html/" charset="UTF-8"
             highlight="false" lowUpperBound="35" highLowerBound="70"/>
    </logging>
    <filter>
        <whitelist processUncoveredFilesFromWhitelist="true">
            <directory suffix=".php">app/</directory>
        </whitelist>
    </filter>
    <testsuites>
        <testsuite name="all">
            <directory>tests/</directory>
        </testsuite>
    </testsuites>
</phpunit>

 ```
</details>

[...]

Issue #576, opened 2018-11-20
Infection gets stuck in the first execution after configuration file is created


y-uti:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.11.2 and also master branch
| Test Framework version | PHPUnit 7.4.4
| PHP version | 7.2.12
| Platform    | CentOS 7.5
| Github Repo | -

At the first execution of Infection, it creates infection.json.dist and then it runs testcases. In this time, the configuration is not loaded properly and the execution goes wrong.

<details>
<summary>Output with issue</summary>

In the following output, Infection is executed with phpdbg and breakpoint is set on https://github.com/infection/infection/blob/ab723af4078d0c7cee48c143eb1d037d6fb3ee4e/src/Command/InfectionCommand.php#L207. Here configuration should be loaded properly IMO.

At the first execution (I don't have infection.json.dist)
```
$ phpdbg vendor/bin/infection
[Welcome to phpdbg, the interactive PHP debugger, v0.5.0]
To get help using phpdbg type "help" and press enter
[Please report bugs to <http://bugs.php.net/report.php>]
[Successful compilation of /tmp/sandbox/vendor/infection/infection/bin/infection]
prompt> break vendor/infection/infection/src/Command/InfectionCommand.php:207
[Breakpoint #0 added at /tmp/sandbox/vendor/infection/infection/src/Command/InfectionCommand.php:207]
prompt> run
You are running Infection with phpdbg enabled.
     ____      ____          __  _
    /  _/___  / __/__  _____/ /_(_)___  ____
    / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
  _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
 /___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/



  Welcome to the Infection config generator



We did not find a configuration file. The following questions will help us to generate it for you.


Which source directories do you want to include (comma separated)? [src]:
  [0] .
  [1] src
  [2] src.bak
  [3] tests
  [4] tests.bak
  [5] vendor
 >

There can be situations when you want to exclude some folders from generating mutants.
You can use glob pattern (*Bundle/**/*/Tests) for them or just regular dir path.
It should be relative to the source directory.
You should not mutate test suite files.
Press <return> to stop/skip adding dirs.

Any directories to exclude from within your source directories?:

Single test suite timeout in seconds [10]:

Where do you want to store the text log file? [infection.log]:

[...]

PR #579, opened 2018-11-25
Add a test to check Infection works with PSR-0 compliant autoloader

borNfreee:
Related to https://github.com/infection/infection/issues/564

### DIFF
--- a/tests/Fixtures/e2e/PSR_0_Autoloader/src/PSR_0_Autoloader/SourceClass.php
+++ b/tests/Fixtures/e2e/PSR_0_Autoloader/src/PSR_0_Autoloader/SourceClass.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace PSR_0_Autoloader;
+
+class SourceClass
+{
+    public function hello(): string
+    {
+        return 'hello';
+    }
+}
--- a/tests/Fixtures/e2e/PSR_0_Autoloader/README.md
+++ b/tests/Fixtures/e2e/PSR_0_Autoloader/README.md
@@ -1,0 +1,5 @@
+# PSR-0 test
+
+* Related to https://github.com/infection/infection/issues/564
+
+This test checks that Infection correctly works with PSR-0 compliant autoloader
--- a/tests/Fixtures/e2e/PSR_0_Autoloader/composer.json
+++ b/tests/Fixtures/e2e/PSR_0_Autoloader/composer.json
@@ -1,0 +1,15 @@
+{
+    "require-dev": {
+        "phpunit/phpunit": "^6.5"
+    },
+    "autoload": {
+        "psr-0": {
+            "PSR_0_Autoloader": "src/"
+        }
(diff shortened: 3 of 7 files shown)


tests/Console/E2ETest.php
```diff
@@ -243,6 +243,19 @@ private function installComposerDeps(): void
...
+            foreach ($paths as $path) {
+                if (strpos($path, $vendorDir) !== false) {
+                    // Skip known dependency from autoloading
+                    continue 2;
+                }
+            }
+
+            $loader->set($namespace, $paths);
+        }
+
```
sanmai: Is this for things inside `/usr/share/php/` and so on?

tests/Fixtures/e2e/PSR_0_Autoloader/src/PSR_0_Autoloader/SourceClass.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace PSR_0_Autoloader;
+
+class SourceClass
```
sanmai: That's just as same as the current PSR-4. I think if test best to test a really old-style `Namespace_Class_Name` style project. 

For example, I could not run Infection on [pear/OLE](https://github.com/pear/OLE). It gives:
```
Processing source code files: 4/7
In CodeCoverageClassIgnoreVisitor.php line 63:
                                 
  Class OLE_Test does not exist  
```

On subsequent, for some very interesting reason:
```
In JsonValidationException.php line 51:
                                                                                            
  "OLE/infection.json.dist" does not match the expected JSON schema:  
   - source.excludes : Object value found, but an array is required                         
```
The last error possibly related to these configuration steps:
```
Any directories to exclude from within your source directories?: tests
Any directories to exclude from within your source directories?: coverage
Any directories to exclude from within your source directories?: build
Any directories to exclude from within your source directories?: vendor
Any directories to exclude from within your source directories?: 
```
infection.json has this:
```
 "excludes": {
     "0": "vendor",
     "1": "tests",
     "3": "coverage",
     "4": "build"
 }
```

tests/Console/E2ETest.php
borNfreee: not sure I understood the question. The code above is for registering PSR-0 namespaces that are located inside `'vendor/composer/autoload_namespaces.php'`, because without this change the code in `E2ETest.php` loads only namespaces registered from `autoload_psr4.php`

sanmai: Never mind, that's just me before a morning coffee.

tests/Fixtures/e2e/PSR_0_Autoloader/src/PSR_0_Autoloader/SourceClass.php
sanmai: Actually, that's not an unknown issue with "Class OLE_Test does not exist", just #576

PR #582, opened 2018-12-05
Fix: Remove cacheResultFile attribute from phpunit.xml

### LINKED ISSUE #572: Running infection with PHPUnit result cache in subdirectories fails
| Question    | Answer
| Infection version | `0.11.1`
| Test Framework version | `phpunit/phpunit:7.4.3`
| PHP version | `7.1.22`
| Platform    | macOs
| Github Repo | n/a |

Related to #487.

Running `infection` with a `phpunit` configuration where result caches are stored in a subdirectory (which is created before running the tests) fails, as neither `phpunit` nor `infection` create the subdirectory when running mutation tests.

Not sure if this is something that should be handled on `phpunit`'s side, or something `infection` should handle.

I think, though, that the easiest would be if this were handled by `phpunit`.

<details>
 <summary>test/Unit/phpunit.xml</summary>
 
 ```xml
<phpunit
    cacheResult="true"
    cacheResultFile="../../.phpunit/unit-result.cache"
>
</phpunit>
 ```
</details>

<details>
 
 ```

         Infection runs the test suite in a RANDOM order. Make sure your tests do not have hidden dependencies.

         You can add these attributes to `phpunit.xml` to check it: <phpunit executionOrder="random"
         resolveDependencies="true" ...

         If you don't want to let Infection run tests in a random order, set the `executionOrder` to some value, for
         example <phpunit executionOrder="default"

         '/Users/am/Sites/localheinz/composer-json-normalizer/vendor/phpunit/phpunit/phpunit' '--configuration'
         '/var/folders/yf/j36qt8n126sf70t0rr_mms500000gn/T/infection/phpunitConfiguration.initial.infection.xml'

         PHPUnit reported an exit code of 143.

         STDOUT:

         PHPUnit 7.4.3 by Sebastian Bergmann and contributors.

         Runtime:       PHP 7.1.22 with Xdebug 2.6.1
         Random seed:   1542452711
         Configuration:
         /private/var/folders/yf/j36qt8n126sf70t0rr_mms500000gn/T/infection/phpunitConfiguration.initial.infection.xml

         ...............................................................  63 / 346 ( 18%)
         ............................................................... 126 / 346 ( 36%)
         ............................................................... 189 / 346 ( 54%)
         ............................................................... 252 / 346 ( 72%)
         ............................................................... 315 / 346 ( 91%)
         ...............................                                 346 / 346 (100%)

         STDERR:

         PHP Warning:
         file_put_contents(/private/var/folders/yf/j36qt8n126sf70t0rr_mms500000gn/T/infection/../../.phpunit/unit-result
         .cache): failed to open stream: No such file or directory in
         /Users/am/Sites/localheinz/composer-json-normalizer/vendor/phpunit/phpunit/src/Util/TestResultCache.php on line
         82
         PHP Stack trace:


 ```
</details>

localheinz:
This PR:

- [x] removes the `cacheResultFile` attribute from `phpunit.xml`

Fixes #572.

### DIFF
--- a/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
+++ b/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
@@ -107,6 +107,7 @@ class InitialConfigBuilder implements ConfigBuilder
         $this->xmlConfigurationHelper->replaceWithAbsolutePaths($xPath);
         $this->xmlConfigurationHelper->setStopOnFailure($xPath);
         $this->xmlConfigurationHelper->deactivateColours($xPath);
+        $this->xmlConfigurationHelper->removeCacheResultFile($dom, $xPath);
         $this->xmlConfigurationHelper->removeExistingLoggers($dom, $xPath);
         $this->xmlConfigurationHelper->removeExistingPrinters($dom, $xPath);
 
--- a/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
+++ b/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
@@ -87,6 +87,7 @@ class MutationConfigBuilder extends ConfigBuilder
         $this->xmlConfigurationHelper->replaceWithAbsolutePaths($xPath);
         $this->xmlConfigurationHelper->setStopOnFailure($xPath);
         $this->xmlConfigurationHelper->deactivateColours($xPath);
+        $this->xmlConfigurationHelper->removeCacheResultFile($dom, $xPath);
         $this->xmlConfigurationHelper->removeExistingLoggers($dom, $xPath);
         $this->xmlConfigurationHelper->removeExistingPrinters($dom, $xPath);
         $this->xmlConfigurationHelper->removeDefaultTestSuite($dom, $xPath);
--- a/src/TestFramework/PhpUnit/Config/XmlConfigurationHelper.php
+++ b/src/TestFramework/PhpUnit/Config/XmlConfigurationHelper.php
@@ -73,6 +73,15 @@ final class XmlConfigurationHelper
         }
     }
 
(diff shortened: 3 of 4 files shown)


src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
```diff
@@ -210,6 +211,11 @@ private function addRandomTestsOrderAttributesIfNotSet(string $version, \DOMXPat
         }
     }
 
+    private function removeCacheResultFileAttribute(\DOMXPath $xPath): void
+    {
```
sanmai: Shouldn't this be part of XmlConfigurationHelper?

localheinz: @sanmai 

Thank you - give me a moment!

borNfreee:
Initially, [you removed](https://github.com/infection/infection/pull/582#discussion_r239008696) `cacheResultFile` attribute from the "initial test suite" config file in `InitialConfigBuilder`, which is correct (according to the failed build https://travis-ci.com/localheinz/phpstan-rules/jobs/162844743#L695-L802).

But now it's inside `MutationConfigBuilder` - this file builds configuration files for each Mutant, not for initial test run.

I think it should be in both `InitialConfigBuilder` and in `MutationConfigBuilder`.

Does it make sense? :)

localheinz:

borNfreee:
Releases [0.11.3](https://github.com/infection/infection/releases/tag/0.11.3)

localheinz:
@borNfreee @sanmai 

Just realized that I now end up with a dirty index because the default `cacheResult` file setting is used now. Should we disable result caching entirely, perhaps?

borNfreee:
you mean to set `cacheResult="false"`? Sounds good to me

localheinz:
On it!

PR #585, opened 2018-12-08
Fix: Set cacheResult attribute to false in phpunit.xml

localheinz:
This PR:

- [x] sets the `cacheResult` attribute to `false` in `phpunit.xml` to disable result caching entirely

Follows https://github.com/infection/infection/pull/582.


src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
```diff
@@ -107,6 +107,7 @@ public function build(string $version): string
         $this->xmlConfigurationHelper->replaceWithAbsolutePaths($xPath);
         $this->xmlConfigurationHelper->setStopOnFailure($xPath);
         $this->xmlConfigurationHelper->deactivateColours($xPath);
+        $this->xmlConfigurationHelper->deactivateResultCaching($xPath);
         $this->xmlConfigurationHelper->removeCacheResultFile($dom, $xPath);
```
localheinz: The question is whether we still need to remove the `cacheResultFile` attribute, then.

What do you think?

borNfreee:
> 🤔

All _new_ files must pass this check, but for the old ones, I think we can ignore it (it's unrelated to your PR). Of course, increasing the MSI is a good thing though.

Probably we need target `0.11` branch, what do you think?

borNfreee:
Released 0.11.4

PR #586, opened 2018-12-08
Fix: Use assertXmlStringEqualsXmlString() instead of assertSame()

localheinz:
This PR

- [x] uses `assertXmlStringEqualsXmlString()` instead of `assertSame()`
- [x] formats XML strings in tests so they are easier to maintain


localheinz:
Thank you, @borNfreee!

PR #587, opened 2018-12-08
Enhancement: Extract methods for removing attributes and setting attribute values

localheinz:
This PR

* [x] extracts methods in `XmlConfigurationHelper` for removing attributes and setting attribute values


localheinz:
Thank you, @borNfreee!

PR #589, opened 2018-12-09
Add a few more tests to mutators

BackEndTea:
This PR:

- [x] Adds a few more tests for mutators

### DIFF
--- a/tests/Fixtures/Autoloaded/PublicVisibility/pv-protected-parent.php
+++ b/tests/Fixtures/Autoloaded/PublicVisibility/pv-protected-parent.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace ProtectedParent;
+
+abstract class SameAbstract
+{
+    protected abstract function foo();
+}
+class Child extends SameAbstract
+{
+    public function foo()
+    {
+    }
+}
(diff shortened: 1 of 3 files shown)


tests/Fixtures/Autoloaded/PublicVisibility/pv-protected-parent.php
```diff
@@ -0,0 +1,14 @@
...
+abstract class SameAbstract
+{
+    protected abstract function foo();
+}
+class Child extends SameAbstract
+{
+    public function foo()
+    {
+    }
+}
```
BackEndTea: Note to self: missing new line at end of file

tests/Mutator/Number/DecrementIntegerTest.php
```diff
@@ -303,6 +303,39 @@ public function provideMutationCases(): array
...
+$foo = 0;
+PHP
+                ,
+                <<<'PHP'
+<?php
+
+$foo = -1;
+PHP
+            ],
+            'It does not decrement an assignment of 1' => [
```
borNfreee: could you please remind why don't we decrement `1`?

BackEndTea: Because of the `OneZeroInteger` mutator, which does exactly that.

borNfreee: ah, thanks, probably worths adding a comment before `if (!$node instanceof Node\Scalar\LNumber || $node->value === 1) {`

borNfreee:
Thank you @BackEndTea

Issue #578, opened 2018-11-23
boolean mutators and array_search, in_array, is_subclass_of, is_a


kalessil:
| Question    | Answer
| ------------| ---------------
| Infection version | latest
| Test Framework version | PHPUnit latest
| PHP version | 7.2
| Platform    | Ubuntu
| Github Repo | -

The booleans mutator (true => false, false => true) should skip some of PHP core function:
- array_search
- in_array
- is_subclass_of
- is_a

The optional parameters there are intended for hardening workflows, and writing tests for those cases are normally is overkill.

borNfreee:
Thanks @kalessil. I agree with this proposal. Let's implement it for the next release ;)

Also, this is related to https://github.com/infection/infection/issues/206

kalessil:
Super, thank you @borNfreee !

kalessil:
Super, thank you @borNfreee !

borNfreee:
No problem ;) Will be released soon, I hope this week. 

Thanks for pushing it forward

PR #599, opened 2019-01-05
Disabling mutating "true" -> "false" in TrueValue mutator for in_array/array_search

borNfreee:
With this PR, `TrueValue` mutator will no longer mutate `true` to `false` for `in_array` and `array_search` functions. This can be explicitly enabled in settings.

```json
{
  "timeout": 25,
  "source": {
    "directories": ["src"]
  },
  "mutators": {
    "TrueValue": {
      "settings": {
        "in_array": true, // default "false"
        "array_search": true // default "false"
      }
    }
  }
}
```

Also, add json schema validation for `TrueValue` mutator in `infection.json`.

- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/105

Fixes https://github.com/infection/infection/issues/578
Fixes https://github.com/infection/infection/issues/206

### DIFF
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -90,7 +90,33 @@
         },
         "mutators": {
             "type": "object",
-            "description": "Contains the settings for different mutations and profiles"
+            "description": "Contains the settings for different mutations and profiles",
+            "properties": {
+                "TrueValue": {
+                    "type": {
+                        "anyOf": [
+                            {
+                                "type": "boolean"
+                            },
+                            {
+                                "type": "object",
+                                "additionalProperties": false,
+                                "properties": {
+                                    "ignore": {"type": "array", "items": {"type": "string"}},
+                                    "settings": {
+                                        "type": "object",
+                                        "additionalProperties": false,
+                                        "properties": {
+                                            "in_array": {"type": "boolean"},
+                                            "array_search": {"type": "boolean"}
+                                        }
+                                    }
+                                }
+                            }
+                        ]
+                    }
+                }
+            }
         },
         "testFramework": {
             "type": "string",
--- a/src/Mutator/Arithmetic/Assignment.php
+++ b/src/Mutator/Arithmetic/Assignment.php
@@ -54,7 +54,7 @@ final class Assignment extends Mutator
         return new Node\Expr\Assign($node->var, $node->expr, $node->getAttributes());
     }
 
-    protected function mutatesNode(Node $node): bool
+    protected function mutatesNode(Node $node, array $mutatorSettings): bool
     {
         return $node instanceof Node\Expr\AssignOp;
     }
--- a/src/Mutator/Arithmetic/Assignment.php
+++ b/src/Mutator/Arithmetic/Assignment.php
@@ -54,7 +54,7 @@ final class Assignment extends Mutator
         return new Node\Expr\Assign($node->var, $node->expr, $node->getAttributes());
     }
 
-    protected function mutatesNode(Node $node, array $mutatorSettings): bool
+    protected function mutatesNode(Node $node): bool
     {
         return $node instanceof Node\Expr\AssignOp;
     }
--- a/src/Mutator/Boolean/TrueValue.php
+++ b/src/Mutator/Boolean/TrueValue.php
@@ -36,6 +36,7 @@ declare(strict_types=1);
 namespace Infection\Mutator\Boolean;
 
 use Infection\Mutator\Util\Mutator;
+use Infection\Visitor\ParentConnectorVisitor;
 use PhpParser\Node;
 
 /**
@@ -43,6 +44,11 @@ use PhpParser\Node;
  */
 final class TrueValue extends Mutator
 {
+    private const DEFAULT_SETTINGS = [
+        'in_array' => false,
+        'array_search' => false,
+    ];
+
     /**
      * Replaces "true" with "false"
      *
@@ -54,12 +60,29 @@ final class TrueValue extends Mutator
         return new Node\Expr\ConstFetch(new Node\Name('false'));
     }
 
-    protected function mutatesNode(Node $node): bool
+    protected function mutatesNode(Node $node, array $mutatorSettings): bool
     {
         if (!($node instanceof Node\Expr\ConstFetch)) {
             return false;
         }
 
-        return $node->name->toLowerString() === 'true';
+        if ($node->name->toLowerString() !== 'true') {
+            return false;
+        }
+
+        $parentNode = $node->getAttribute(ParentConnectorVisitor::PARENT_KEY);
+        $grandParentNode = $parentNode !== null ? $parentNode->getAttribute(ParentConnectorVisitor::PARENT_KEY) : null;
+
+        if ($grandParentNode instanceof Node\Expr\FuncCall) {
+            $resultSettings = array_merge(self::DEFAULT_SETTINGS, $mutatorSettings);
+
+            $functionName = $grandParentNode->name->toLowerString();
+
+            if (array_key_exists($functionName, $resultSettings) && $resultSettings[$functionName] === false) {
+                return false;
+            }
+        }
+
+        return true;
     }
 }
--- a/src/Mutator/Boolean/TrueValue.php
+++ b/src/Mutator/Boolean/TrueValue.php
@@ -77,7 +77,7 @@ final class TrueValue extends Mutator
             return true;
         }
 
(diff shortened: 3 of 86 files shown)


src/Mutator/Arithmetic/Assignment.php
```diff
@@ -54,7 +54,7 @@ public function mutate(Node $node)
         return new Node\Expr\Assign($node->var, $node->expr, $node->getAttributes());
     }
 
-    protected function mutatesNode(Node $node): bool
+    protected function mutatesNode(Node $node, array $mutatorSettings): bool
```
BackEndTea: What if we make the mutatorSettings par of the `MutatorConfig` class, and let it expose a `getSettings()` method? This means we don't have to pass mutatorSettings to every node that doesn't need it.

src/Mutator/Boolean/TrueValue.php
```diff
@@ -54,12 +60,29 @@ public function mutate(Node $node)
         return new Node\Expr\ConstFetch(new Node\Name('false'));
     }
 
-    protected function mutatesNode(Node $node): bool
+    protected function mutatesNode(Node $node, array $mutatorSettings): bool
```
BackEndTea: What if we move the 'settings check' to a separate method, like `isAllowedBySettings(Node $node): bool`.

Then the abstract Mutator class simly has a return true there, and if a mutator has settings, then it can override that method.

resources/schema.json
```diff
@@ -90,7 +90,33 @@
         },
         "mutators": {
             "type": "object",
-            "description": "Contains the settings for different mutations and profiles"
+            "description": "Contains the settings for different mutations and profiles",
```
BackEndTea: I like how php-cs-fixer does their settings validation. There every fixer class has some kind of a settings generation method that defines what settings there are for that specific fixer. 

Though that may be too much for now

src/Mutator/Boolean/TrueValue.php
```diff
@@ -54,12 +60,29 @@ public function mutate(Node $node)
...
+            return false;
+        }
+
+        $parentNode = $node->getAttribute(ParentConnectorVisitor::PARENT_KEY);
+        $grandParentNode = $parentNode !== null ? $parentNode->getAttribute(ParentConnectorVisitor::PARENT_KEY) : null;
+
+        if ($grandParentNode instanceof Node\Expr\FuncCall) {
+            $resultSettings = array_merge(self::DEFAULT_SETTINGS, $mutatorSettings);
+
+            $functionName = $grandParentNode->name->toLowerString();
```
BackEndTea: This will crash if a function with a variable name is used, e.g

```php
<?php

$a = 'foo';
$a(true);

'function_name'(true) || ('function_name')(true);
```

So we should add an `instanceof Node\Name` check before that part, and add tests with these cases.

src/Mutator/Boolean/TrueValue.php
```diff
@@ -54,12 +60,29 @@ public function mutate(Node $node)
...
 
-        return $node->name->toLowerString() === 'true';
+        if ($node->name->toLowerString() !== 'true') {
+            return false;
+        }
+
+        $parentNode = $node->getAttribute(ParentConnectorVisitor::PARENT_KEY);
+        $grandParentNode = $parentNode !== null ? $parentNode->getAttribute(ParentConnectorVisitor::PARENT_KEY) : null;
+
+        if ($grandParentNode instanceof Node\Expr\FuncCall) {
```
BackEndTea: We can reduce indentation by inverting this if, and returing  true much earlier e.g.

```php
if (! $grandParentNode instanceof Node\Expr\FuncCall) {
    return true;
}
// Rest of the code ...
```

tests/Mutator/Boolean/TrueValueTest.php
```diff
@@ -90,6 +91,133 @@ public function provideMutationCases(): array
...
+
+\array_search($a, $b, false);
+PHP
+            ,
+            [
+                'settings' => ['array_search' => true],
+            ],
+        ];
+
+        yield 'It does not mutate when used in "\array_search" function and explicitly disabled' => [
```
BackEndTea: How about adding a test case where array_search is wrongly capitalized.

tests/Mutator/AbstractMutatorTestCase.php
```diff
@@ -66,7 +66,7 @@ protected function setUp(): void
         $this->mutator = $this->getMutator();
     }
 
-    public function doTest(string $inputCode, $expectedCode = null): void
+    public function doTest(string $inputCode, $expectedCode = null, $settings = []): void
```
BackEndTea: What about 
```suggestion
    public function doTest(string $inputCode, ?string $expectedCode = null, array $settings = []): void
```

src/Mutator/Boolean/TrueValue.php
```diff
@@ -36,13 +36,19 @@
...
 use PhpParser\Node;
 
 /**
  * @internal
  */
 final class TrueValue extends Mutator
 {
+    private const DEFAULT_SETTINGS = [
+        'in_array' => false,
+        'array_search' => false,
```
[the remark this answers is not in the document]
borNfreee: done

tests/Json/JsonFileTest.php
```diff
@@ -113,4 +113,153 @@ public function test_it_throws_schema_validation_exception(): void
...
+    }
+
+    public function validTrueValueProvider(): \Generator
+    {
+        yield 'Boolean value' => [
+            <<<JSON
+{
+    "timeout": 25,
+    "source": {"directories": ["src"]},
+	"mutators": {
```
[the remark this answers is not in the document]
borNfreee: fixed

src/Mutator/Arithmetic/Assignment.php
borNfreee: done

src/Mutator/Boolean/TrueValue.php
borNfreee: I would keep the number of the methods in Mutator class as few as possible. Also, I think `isAllowedBySettings()` is very coupled with `mutatesNode()` and almost always will need to do the same things in the beginning (like checking if it's a `ConstFetch` in `TrueValue` mutator), so we will end up with code duplication

resources/schema.json
borNfreee: With json schema validation, we get the feedback very fast, even without running initial test suite and/or generating mutations.

We can improve (or add an additional layer of validation) later if that will be needed, but I'm not sure there is a clear reason for doing it now.

src/Mutator/Boolean/TrueValue.php
borNfreee: thanks, done

src/Mutator/Boolean/TrueValue.php
borNfreee: done

tests/Mutator/Boolean/TrueValueTest.php
borNfreee: done

tests/Mutator/AbstractMutatorTestCase.php
borNfreee: `?string $expectedCode` - we can't do it, since we use strings and arrays for 1 and multiple mutations cases correspondingly 

`array $settings` - done

src/Mutator/Boolean/TrueValue.php
```diff
@@ -60,6 +66,21 @@ protected function mutatesNode(Node $node): bool
...
+        }
+
+        $parentNode = $node->getAttribute(ParentConnectorVisitor::PARENT_KEY);
+        $grandParentNode = $parentNode !== null ? $parentNode->getAttribute(ParentConnectorVisitor::PARENT_KEY) : null;
+
+        if (!$grandParentNode instanceof Node\Expr\FuncCall || !$grandParentNode->name instanceof Node\Name) {
+            return true;
+        }
+
+        $resultSettings = array_merge(self::DEFAULT_SETTINGS, $this->config->getMutatorSettings());
```
BackEndTea: What about keeping the `$config` private, and instead exposing a protected `getSettings` method in the abstract Mutator class. Then we could do this here.
```diff
- $this->config->getMutatorSettings()
+ $this->getSettings()
```

borNfreee: sounds good, will do it

borNfreee:
@BackEndTea could you please review again?

borNfreee:
thank you @BackEndTea @sanmai

PR #604, opened 2019-01-11
Fix: Run php-cs-fixer on GitHub Actions

### LINKED ISSUE #1100: PrettyCI is shutting down
@localheinz I'm pretty sure you will be happy! :D

_Original email:_

<details>

PrettyCI is shutting down

PrettyCI started in 2017 as a side project. Initially aiming to solve my needs, I soon opened it to open source projects, and later paid customers.

Given the introduction of free alternatives like GitHub Actions, and after years of maintenance for next to no revenue, I am sorry to say I will be shutting it down. PrettyCI has been useful, but it isn't a commercial success. It is time to move on to new adventures.
How to replace PrettyCI?

You are encouraged to migrate to GitHub Actions. GitHub Actions is free and easy to setup, which makes it a good alternative to PrettyCI.

To help you, I have prepared a guide with examples: Check out the migration guide here.
Is there a transition period?

Yes, builds will continue to run for a while. You projects will not break.

Open source projects will keep building until December 2020.

Paid customers will be switched to a free plan that includes their private projects. To be clear, private projects will continue to build as before, until December 2020.
What will change from now until December 2020?

It will not be possible to add new projects to PrettyCI. PrettyCI will stop receiving updates. Remember this is a transition period, projects should be moved to GitHub Actions.
Thank you for your support

Thank you to everyone who used PrettyCI. Thank you to those who were patient when it needed fixing, who were helpful when it needed improvements. And finally, a big thank you to those who trusted the project to become a paid subscriber.

If you have any question, you can reach out to me at matthieu at mnapoli dot fr.

</details>

I think we need to move it to GitHub Actions. Help is very welcome.

localheinz:
This PR

* [x] runs `php-cs-fixer` on GitHub Actions

Fixes #1100.

💁‍♂ For reference, see 

- https://github.com/localheinz/infection/pull/1
- https://github.com/localheinz/infection/pull/1/checks

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,5 +8,8 @@ insert_final_newline=true
 indent_style=space
 indent_size=4
 
+[*.yml]
+indent_size=2
+
 [Makefile]
 indent_style=tab
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,0 +1,38 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
+
+name: "Continuous Integration"
+
+on:
+  - "pull_request"
+  - "push"
+
+jobs:
+  coding-standards:
+    name: "Coding Standards"
+
+    runs-on: "ubuntu-latest"
+
+    env:
+      PHP_CS_FIXER_BINARY: ".tools/php-cs-fixer"
+      PHP_CS_FIXER_CACHE_DIRECTORY: "build/cache/php-cs-fixer"
+      PHP_CS_FIXER_URL: "https://cs.sensiolabs.org/download/php-cs-fixer-v2.phar"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+
+      - name: "Download friendsofphp/php-cs-fixer to ${{ env.PHP_CS_FIXER_BINARY }}"
+        run: "wget ${{ env.PHP_CS_FIXER_URL }} --output-document=${{ env.PHP_CS_FIXER_BINARY }}"
+
+      - name: "Create cache directory for friendsofphp/php-cs-fixer"
+        run: "mkdir -p ${{ env.PHP_CS_FIXER_CACHE_DIRECTORY }}"
+
+      - name: "Cache cache directory for friendsofphp/php-cs-fixer"
+        uses: "actions/cache@v1"
+        with:
+          path: "${{ env.PHP_CS_FIXER_CACHE_DIRECTORY }}"
+          key: "php-7.2-php-cs-fixer"
+          restore-keys: "php-7.2-php-cs-fixer"
+
+      - name: "Run friendsofphp/php-cs-fixer"
+        run: "php7.2 ${{ env.PHP_CS_FIXER_BINARY }} fix --cache-file=${{ env.PHP_CS_FIXER_CACHE_DIRECTORY }}/.php_cs.cache --diff --diff-format=udiff --dry-run --verbose"
--- a/.prettyci.composer.json
+++ b/.prettyci.composer.json
@@ -1,5 +1,0 @@
-{
-    "require-dev": {
-        "friendsofphp/php-cs-fixer": "^2.16"
-    }
-}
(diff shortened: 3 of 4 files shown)


localheinz:
Failure on Travis CI appears to be unrelated.

localheinz:
### Running on Travis CI

Running `php-cs-fixer` on Travis CI for this build increases the build time by 9 seconds:

screen shot 2019-01-11 at 19 42 42

For reference, see https://travis-ci.org/infection/infection/jobs/478404589#L570-L576. 

💡 Note that we are currently always downloading `php-cs-fixer-v2.phar`, even though we are currently not using it on Travis CI.

### Running on PrettyCI

Running on PrettyCI takes *less than 5 seconds*.

screen shot 2019-01-11 at 19 43 29

For reference, see https://github.com/infection/infection/pull/604/checks?check_run_id=48977222.

### Question

Do we prefer to use PrettyCI to save a few seconds and give up control over 

1) which PHP version is used 
2) which version of `friendsofphp/php-cs-fixer` is used

or prefer to run it ourselves?

localheinz:
Actually, no - running in PrettyCI is even slower as it is not using the cache.

mnapoli:
Thanks for the link to this issue on Twitter.

FYI if you want you can control the version of php-cs-fixer on PrettyCI by requiring it in `.prettyci.composer.json` (see https://prettyci.com/documentation/#using-a-custom-standard).

Regarding the build execution time you are missing something important if you only look at the tool's execution time. Let me re-post here a screenshot of https://prettyci.com

capture d ecran 2019-01-13 a 18 58 39

Here is a fair comparison:

- Travis CI: wait for a free container (sometimes a few minutes) + wait for *the whole job* to run (init + php-cs-fixer + phpstan) which took 1min14s in the one you linked (the github commit status is only updated at the end of all of that)
- PrettyCI: (no need to wait for a free container as it runs on lambda) wait for the lambda to boot (300ms) + wait for php-cs-fixer to run (15 seconds)

PrettyCI is much faster in any case. And it frees up containers on Travis (on open source plans it's pretty common to have queued jobs when you run too many things on Travis). If you want a proof just push a commit to github and check up how long it takes for commit statuses to update, it's pretty obvious.

Anyway I don't really care which tool you want to use, I just want to clarify misunderstandings.

sanmai:
PrettyCI is better because it is integrated with GitHub, and you don't have to dig through build logs only to find that there's a styling issue.

localheinz:
@mnapoli @sanmai 

As pointed in the description of the PR, if we disregard the question of speed (which, let's be honest, doesn't matter much), this is mostly about the advantages and disadvantages of using PrettyCI vs. running `friendsof/php-cs-fixer` on Travis in regard to having control over versions of

* PHP where `php-cs-fixer` is run
* `php-cs-fixer`

.editorconfig
```diff
@@ -8,5 +8,8 @@ insert_final_newline=true
 indent_style=space
 indent_size=4
 
+[*.yml]
+indent_size=2
```
theofidry: Can we keep 4? It's really unsettling to have a vertical spacing as wide as the indentation :(

theofidry:
Do I understand correctly that it is downloading PHP-CS-Fixer each time?

localheinz:
@theofidry 

> Do I understand correctly that it is downloading PHP-CS-Fixer each time?

That is correct - I'm mimicking the behaviour as observed in `Makefile`. That's probably not what I would do, but this is the way it is done here. Takes less than 1 second in GitHub Actions.

.editorconfig
localheinz: Personally, I would always use 2 spaces for indentation with YAML files as 4 spaces makes it rather awkward when working with definitions of objects in lists.

I can change to 4 spaces, though, if that is a requirement.

Note that we currently have mixed indentation - sometimes 2, sometimes 4 spaces.

theofidry:
Except the Makefile does cache the result (based on the Makefile timestamp - not perfect). I guess it's fine, I'm more concerned about hitting the PHP-CS-Fixer URL each time to be honest but I guess they can handle the load

.editorconfig
theofidry: Guess so... Could we uniformize it somehow? might be more proper to do it in another PR - whatever the choice is

localheinz:
Thank you, @maks-rafalko and @theofidry!

.editorconfig
sanmai: Two spaces work best for YML because lists call for two. 
```
list:
  - foo: 10
    bar: 20
  - foo: 11
    bar: 21
```

With four spaces you'll need to manually pad them:

```
list:
    - foo: 10
      bar: 20
    - foo: 11
      bar: 21
```

Issue #596, opened 2019-01-02
infection own tests are failing on windows


siad007:
| Question    | Answer
| ------------| ---------------
| Infection version | dev-master (`infection.phar --version`)
| Test Framework version | PHPUnit 6.5.13
| PHP version | 7.1.8 (`php -v`)
| Platform    | Windows / cygwin or cmd.exe
| Github Repo | -

Executed with cygwin
```
There was 1 failure:

1)
         Infection\Tests\Console\E2ETest::test_it_runs_an_e2e_test_with_success
         with data set "Code_Coverage_Ignore"
         (Symfony\Component\Finder\SplFileInfo Object
         (...))
C:\Users\xxx\GITRepos\infection\tests\Fixtures\e2e\Code_Covera
         ge_Ignore/expected-output.txt is not same as infection.log (if that is
         OK, run GOLDEN=1 vendor/bin/phpunit)
Failed asserting that two strings
         are equal.
--- Expected
+++ Actual
@@ @@
-'Total: 1\r\n
-Killed:
         0\r\n
-Errored: 0\r\n
-Escaped: 0\r\n
-Timed Out: 0\r\n
+'Total:
         1\n
+Killed: 0\n
+Errored: 0\n
+Escaped: 0\n
+Timed Out:
         0\n

```
When executed with `cmd.exe`
```
There was 1 failure:

1) Infection\Tests\Config\ValueProvider\TestFrameworkConfigPathProviderTest::test_it_asks_question_if_no_config_is_found_in_current_dir
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'C:\Users\said\GITRepos\infection\tests\Fixtures\Files\phpunit'
+''

C:\Users\said\GITRepos\infection\tests\Config\ValueProvider\TestFrameworkConfigPathProviderTest.php:119

--
```

sanmai:
Are you sure you haven't edited that file?

I see only normal line endings:
```
$ cat -ev tests/Fixtures/e2e/Code_Coverage_Ignore/expected-output.txt 
Total: 1$
Killed: 0$
Errored: 0$
Escaped: 0$
Timed Out: 0$
```
Compare with this example where `\r` is visible:

```
$ echo -e 'abc\ndef\r\n' | cat -ev
abc$
def^M$
$
```

sanmai:
Oh right. The test wants to see `\r\n` where actually there are only `\n`s. Workaround is easy, but I'm not sure that's going to be the right solution.

[...]

PR #607, opened 2019-01-12
Fixed E2ETest::test_it_runs_an_e2e_test_with_success() for win

siad007:
Fixed E2ETest::test_it_runs_an_e2e_test_with_success() under windows.

Related to #596

### DIFF
--- a/tests/Logger/PerMutatorLoggerTest.php
+++ b/tests/Logger/PerMutatorLoggerTest.php
@@ -59,11 +59,11 @@ final class PerMutatorLoggerTest extends TestCase
             ->method('dumpFile')
             ->with(
                 sys_get_temp_dir() . '/fake-file.md',
-                "# Effects per Mutator\n" .
-                "\n" .
-                "| Mutator | Mutations | Killed | Escaped | Errors | Timed Out | MSI | Covered MSI |\n" .
-                "| ------- | --------- | ------ | ------- |------- | --------- | --- | ----------- |\n" .
-                "| For_ | 15 | 10 | 0 | 0 | 0 | 66| 100|\n" .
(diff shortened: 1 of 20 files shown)


tests/Logger/TextFileLoggerTest.php
```diff
@@ -276,7 +277,12 @@ public function test_it_logs_correctly_with_mutations_and_debug_verbosity_and_no
 
 Diff Diff
 
-TXT
+TXT;
+
+        $content = str_replace("\n", PHP_EOL, $content);
```
sanmai: I seems like other places about would also fancy this approach with `str_replace`. It is way more readable than `'foo'.PHP_EOL.'bar'`.

tests/Logger/PerMutatorLoggerTest.php
```diff
@@ -59,11 +59,11 @@ public function test_it_correctly_build_log_lines(): void
             ->method('dumpFile')
             ->with(
                 sys_get_temp_dir() . '/fake-file.md',
-                "# Effects per Mutator\n" .
-                "\n" .
-                "| Mutator | Mutations | Killed | Escaped | Errors | Timed Out | MSI | Covered MSI |\n" .
-                "| ------- | --------- | ------ | ------- |------- | --------- | --- | ----------- |\n" .
-                "| For_ | 15 | 10 | 0 | 0 | 0 | 66| 100|\n" .
+                '# Effects per Mutator' . PHP_EOL .
```
borNfreee: could you please use `str_replace` here as well?

also, PHPCSFixer is failing

borNfreee:
Thank you @siad007

Issue #606, opened 2019-01-11
Schema does not include initialTestsPhpOptions


nesl247:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.11.4
| Test Framework version | PHPUnit 7.5.1
| PHP version | 7.2.13
| Platform    | Docker
| Github Repo | -

The schema in https://github.com/infection/infection/blob/64c474fcd2017663383a08edf05e123bf31988dc/resources/schema.json does not include `initialTestsPhpOptions` even though https://infection.github.io/guide/usage.html shows it in an example config.

<details>
 <summary>Output with issue</summary>
 
 ```
 In JsonValidationException.php line 51:

                                                                               

  "/application/infection.json.dist" does not match the expected JSON schema:  

   - The property initialTestsPhpOptions is not defined and the definition do  

  es not allow additional properties              
 ```
</details>

sidz:
as well as `testFrameworkOptions`

sidz:
Fixed. and release in 0.11.5

PR #608, opened 2019-01-12
Add missed properties initialTestsPhpOptions and testFrameworkOptions to the json validation scema

sidz:
Fixes https://github.com/infection/infection/issues/606


borNfreee:
Thank you @sidz

PR #618, opened 2019-01-28
Add `--no-progress` option to differentiate between progress bars output

### LINKED ISSUE #613: Provide compact output for CI environments
The amount of text generated by the initial processes (running tests, generating mutations) is quite big, specially when running it on medium to large projects. This is basically due to the progress bars that are being used and that, for CI environments without proper ANSI support, is quite irrelevant.

**Example:**
image

It would be nice if could have a flag to omit the progress bars of these listeners:

* `InitialTestsConsoleLoggerSubscriber`
* `MutantCreatingConsoleLoggerSubscriber`
* `MutationGeneratingConsoleLoggerSubscriber`

Another option is to use something like the `DotFormatter` instead of the progress bars.

borNfreee:
This PR:

- [x] Adds new option `--ci-friendly` to differentiate between progress bars output. The name is not related to progress bars only, it will be used for any CI-related things.
- [x] Covered by tests
- [x] Doc updated

Fixes #613 

----

From https://github.com/infection/infection/issues/613

> Another option is to use something like the `DotFormatter` instead of the progress bars.

Current implementation saves the number of files/mutations generated in strings, which gives more useful information for developer (as well as with debugging for maintainers) rather than dots.

Before:

ci-2

[After](https://travis-ci.org/infection/infection/jobs/485497282#L1379-L1390):

ci-2

❗️ We can also implement Travis autodetection to enable new loggers automatically even if `--ci-friendly` is not passed by developer. WDYT?

---
This change is [Reviewable](https://reviewable.io/reviews/infection/infection/618)


sanmai:
What do you think about renaming this option into `--no-progress`? This way it will be obvious exactly what it does, and it is a common standard option for other tools. E.g. composer has it.

borNfreee:
* Renamed to `--no-progress` (however, I still think `--ci-friendly` may be useful if we are going to extend its meaning not for progress bars only
* Added CI autodetection

borNfreee:
Thanks @sanmai for the review

sanmai:
No problem, please ask me anytime if there's a need for a second set of eyes.

PR #622, opened 2019-02-05
Strange phpdbg failures on Travis

borNfreee:
This PR shows, how adding three (almost empty) files lead to `phpdbg` fails on Travis.

Build: https://travis-ci.org/infection/infection/builds/489218026
Error: https://travis-ci.org/infection/infection/jobs/489218030#L3136-L3142

```bash
The process "'/home/travis/.phpenv/versions/7.1.11/bin/phpdbg' '-qrr'
         '/hom                                                                  
                                                                                
         e/travis/build/infection/infection/tests/Fixtures/e2e/Custom_tmp_dir/ve
         ndor                                                                   
           /phpunit/phpunit/phpunit' '--version'" exceeded the timeout of 60    
         seconds.   
```

I started investigating this issue because of these strange fails in my latest PR: https://github.com/infection/infection/pull/618

I've been trying [many](https://github.com/infection/infection/commits/ci-friendly-output-debug) [different](https://github.com/infection/infection/commits/test-phpdbg-failures) things but with no luck.

What is interesting, when I connect to the failed build via Travis Debug feature, running command by command from our `.travis.yml` file, the issue is not reproduced - build is always successfully finished.

To be honest, I have no more ideas what to do.. Spent too much time on it.

This is a critical issue since it blocks any new PR that adds new files, including https://github.com/infection/infection/pull/618.

Advice would be much appreciated.

PR #626, opened 2019-02-12
Raise the number of open files limit to work around PHPDBG failures

sanmai:
Should fix #622 for time being. If fix at all.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: php
 
-sudo: false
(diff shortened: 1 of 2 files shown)


sanmai:
Limit on the number of open files [is already quite high](https://travis-ci.org/infection/infection/jobs/491943704#L499).
```
open files                      (-n) 30000
```
So I guess this will be it.

PR #627, opened 2019-02-12
Upgrade PHPUnit from ^6.5 to ^7.5 and all dependencies, including root differ

borNfreee:
This PR:

- [x] Upgrades `PHPUnit` from `^6.5` to `^7.5`
- [x] Upgrades `sebastian/diff` to `^3.0` (see composer.lock). This returns back diffs with a context lines (lines before and after a modified one 🎉)
- [x] Run tests in random order by default

#### Why not PHPUnit 8.0?

It [requires](https://github.com/sebastianbergmann/phpunit/blob/d64e502503bcfab5d46c0b74ffd41f4986831dc7/composer.json#L24) PHP ^7.2 while we support 7.1 yet.

#### Next steps

* integrate `pcov`
* use `--dump-xdebug-filter` (see https://thephp.cc/news/2019/01/faster-code-coverage)

---

Had to upgrade dependencies of one of the e2e test because of fatal error:

```bash
PHP Fatal error:  Declaration of PHPUnit\Framework\Constraint\ArrayHasKey::matches($other) must be compatible with PHPUnit\Framework\Constraint\Constraint::matches($other): bool in tests/Fixtures/e2e/Save_PHPUnit_Bootstrap_File/vendor/phpunit/phpunit/src/Framework/Constraint/ArrayHasKey.php on line 83
```

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -4,6 +4,7 @@
          bootstrap="./vendor/autoload.php"
          colors="true"
          verbose="true"
+         executionOrder="random"
 >
     <testsuites>
         <testsuite name="Application Test Suite">
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -3,8 +3,8 @@
          xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/phpunit.xsd"
          bootstrap="./vendor/autoload.php"
          colors="true"
-         verbose="true"
          executionOrder="random"
+         verbose="true"
 >
     <testsuites>
         <testsuite name="Application Test Suite">
--- a/tests/TestFramework/PhpUnit/Config/XmlConfigurationHelperTest.php
+++ b/tests/TestFramework/PhpUnit/Config/XmlConfigurationHelperTest.php
@@ -535,7 +535,7 @@ XML
 
         $xmlHelper = new XmlConfigurationHelper($this->getPathReplacer(), '');
 
-        $this->assertTrue(true, $xmlHelper->validate($dom, $xPath));
+        $this->assertTrue($xmlHelper->validate($dom, $xPath));
     }
 
     public function test_it_removes_default_test_suite(): void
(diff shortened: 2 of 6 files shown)


tests/TestFramework/PhpUnit/Config/XmlConfigurationHelperTest.php
```diff
@@ -535,7 +535,7 @@ public function test_it_passes_validation_by_xsd(string $xsdSchema): void
 
         $xmlHelper = new XmlConfigurationHelper($this->getPathReplacer(), '');
 
-        $this->assertTrue(true, $xmlHelper->validate($dom, $xPath));
```
BackEndTea: :thinking:

phpunit.xml.dist
```diff
@@ -4,6 +4,7 @@
          bootstrap="./vendor/autoload.php"
          colors="true"
          verbose="true"
+         executionOrder="random"
```
localheinz: What do you think about keeping attributes sorted by name?

borNfreee: thanks, done

borNfreee:
> Looks like `--min-msi` needs to be adjusted?

that's almost impossible in this case, because Infection is run with empty `--filter` for this PR (https://travis-ci.org/infection/infection/jobs/492324430#L1048)

This is because no PHP source files were modified from `src/`.

It means that to achieve 90%, I need to do it for the whole project ;) Not feasible indeed.

The goal of this target was to target 90% for the _changed files_, it just does not work properly with empty filter.

borNfreee:
misread your comment. Ok - the target MSI for *changed files* is ok to be 90% (original `--min-msi=67` in our `.travis.yml`). 

But it shouldn't be 90% when the `--filter` is empty. This is the issue, but I guess out of the scope of this PR

borNfreee:
Created https://github.com/infection/infection/issues/631 for MSI issue.

Thank you.

PR #630, opened 2019-02-12
Fix/anonymous class

### LINKED ISSUE #616: Error when parsing method that creates anonymous class
| Question    | Answer
| Infection version | 0.11.5 (`infection.phar --version`)
| Test Framework version | PHPUnit 7.5.2
| PHP version | 7.1.27 (`php -v`)
| Platform    | Linux
| Github Repo | -

Infection reports an error if the following code occurs in source folder:
```php
namespace My\Namespace;

class Bug
{

    public function createAnonymousClass()
    {
        new class
        {
            public function foo()
            {
            }
        };
    }
}
```
The error is:
```
In CodeCoverageMethodIgnoreVisitor.php line 56:

  Method foo does not exist
```
Seems that code seeks method in host class instead of anonymous one, because if I add `private function foo() {}` to `Bug` class, the error dissapears.
I don't know either bug is in PHP-Parser or in Infection, but it affects your software anyway.

BackEndTea:
This PR:

- [x] Fixes #616 

This is fixed in two ways:

First of all, it no longer uses reflection for the doc comments, since those are available from the parser. Secondly, to stop this from coming back in the future, it now keeps a stack of the current class, which does make mutators using the `REFLECTION_CLASS_KEY` responsible for checking if it is a valid value. 

The current uses in the public and protected mutators already do this.

### DIFF
--- a/src/Visitor/CodeCoverageClassIgnoreVisitor.php
+++ b/src/Visitor/CodeCoverageClassIgnoreVisitor.php
@@ -36,7 +36,6 @@ declare(strict_types=1);
 namespace Infection\Visitor;
 
 use PhpParser\Node;
-use PhpParser\Node\Name;
 use PhpParser\Node\Stmt;
 use PhpParser\NodeTraverser;
 use PhpParser\NodeVisitorAbstract;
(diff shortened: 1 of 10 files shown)


BackEndTea:
> It would be really nice to see the test case from the issue in subject.

No problem, i should have some time this weekend to work on it

borNfreee:
Thank you @BackEndTea

BackEndTea:

jakzal:
Any chance to get this fix released? Thanks for all your hard work!

PR #633, opened 2019-02-16
Enhancement: Implement UnwrapArrayUintersectUassoc mutator

localheinz:
This PR

- [x] Adds new feature `UnwrapArrayUintersectUassoc` mutator
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/111

Related to #514.

💁‍♂️ Follows the example in https://github.com/infection/infection/pull/513.

PR #628, opened 2019-02-12
Enhancement: Implement UnwrapArrayUintersectAssoc mutator

localheinz:
This PR

- [x] Adds new feature `UnwrapArrayUintersectAssoc` mutator
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/110

Related to #514.

💁‍♂️ Follows the example in https://github.com/infection/infection/pull/513.

PR #636, opened 2019-02-16
Enhancement: Yield more mutations

localheinz:
This PR

* [x] yields more mutants from `UnwrapArray*` mutators

💁‍♂️ I'm unsure where to draw the line, although I remember that we briefly discussed this in previous PRs. Maybe worth taking a look?

### DIFF
--- a/src/Mutator/Unwrap/UnwrapArrayDiff.php
+++ b/src/Mutator/Unwrap/UnwrapArrayDiff.php
@@ -49,6 +49,6 @@ final class UnwrapArrayDiff extends AbstractUnwrapMutator
 
     protected function getParameterIndexes(Node $node): \Generator
     {
-        yield 0;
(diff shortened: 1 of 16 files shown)


borNfreee:
For the latest 2 PRs we changed 1 to N mutations because they were `array_*intersect*` functions:

* https://github.com/infection/infection/pull/628
* https://github.com/infection/infection/pull/633

which is totally makes sense

borNfreee:
Let's make a decision about this one, to not keep it open without any activity.

I'm 👎 on addition multiple mutations be default (current state of PR) since it produces too many mutants that will likely be killed.

I'm 👍 on adding it by demand (enabling explicitly in `settings`).

I'm also 👍 to just close it and keep the number of mutations balanced.

what do you think?

PR #639, opened 2019-02-17
Reduce ammount of mutators generated by changing numbers

BackEndTea:
This PR:

- [x] Reduce ammount of mutators generated by changing numbers
- [x] Covered by tests

For infection this only reduces the amount of mutators by 5. But for applications which do a lot of comparison against numbers, this will greatly reduce the amount of duplicate mutators. Since changing the number by one is basically the same as changing the comparison operator.


BackEndTea:
Thank you @borNfreee

PR #640, opened 2019-02-17
Add a few more count esque functions to not decrement against

BackEndTea:
They dont need decrementing for the same reason as count functions.

### DIFF
--- a/src/Mutator/Number/DecrementInteger.php
+++ b/src/Mutator/Number/DecrementInteger.php
@@ -44,7 +44,7 @@ use PhpParser\Node;
  */
 final class DecrementInteger extends Mutator
 {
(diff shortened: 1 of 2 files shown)


src/Mutator/Number/DecrementInteger.php
```diff
@@ -44,7 +44,7 @@
  */
 final class DecrementInteger extends Mutator
 {
-    private const COUNT_NAMES = ['count', 'sizeof'];
+    private const COUNT_NAMES = ['count', 'grapheme_strlen', 'iconv_strlen', 'mb_strlen', 'sizeof', 'strlen'];
```
localheinz: What do you think about starting to wrap here sooner or later?

BackEndTea:
Thank you @borNfreee, @sanmai  & @localheinz

PR #641, opened 2019-02-17
Mutator: AssignCoalesce. Upgrade PHPParser to 4.2.1

borNfreee:
This PR:

- [x] Adds new mutator `AssignCoalesce`
- [x] Ugrades PHPParser to [4.2.1](https://github.com/nikic/PHP-Parser/releases/tag/v4.2.1)
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/117

See https://wiki.php.net/rfc/null_coalesce_equal_operator

Produces the following mutation:

```diff
- $this->request->data['comments']['user_id'] ??= 'value';
+ $this->request->data['comments']['user_id'] = 'value';
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         "ext-libxml": "*",
         "composer/xdebug-handler": "^1.3",
         "justinrainbow/json-schema": "^5.2",
-        "nikic/php-parser": "^4.2",
+        "nikic/php-parser": "^4.2.1",
         "ocramius/package-versions": "^1.2",
         "padraic/phar-updater": "^1.0.4",
         "pimple/pimple": "^3.2",
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,7 @@
         "ext-libxml": "*",
         "composer/xdebug-handler": "^1.3",
         "justinrainbow/json-schema": "^5.2",
-        "nikic/php-parser": "^4.1",
+        "nikic/php-parser": "^4.2",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -43,7 +43,7 @@
         "ext-libxml": "*",
         "composer/xdebug-handler": "^1.3",
         "justinrainbow/json-schema": "^5.2",
-        "nikic/php-parser": "^4.1",
+        "nikic/php-parser": "^4.2",
```
localheinz: What do you think about making it explicit that we require at least `nikic/php-parser:^4.2.1`?

borNfreee: done

tests/Mutator/Operator/AssignCoalesceTest.php
```diff
@@ -0,0 +1,97 @@
...
+{
+    /**
+     * @dataProvider provideMutationCases
+     */
+    public function test_mutator($input, $expected = null): void
+    {
+        $this->doTest($input, $expected);
+    }
+
+    public function provideMutationCases(): \Generator
```
BackEndTea: What about adding a test case for not mutating of non assignment coalesce

e.g.
`$a = $b['value'] ?? $c;`

borNfreee: thanks, added

borNfreee:
Thank you @sanmai, @localheinz and @BackEndTea

Issue #644, opened 2019-02-18
Add a new unwrap mutator: ucwords


drupol:
Related to #514

Issue #514, opened 2018-10-19
New @unwrap mutators ideas


maks-rafalko:
After merging https://github.com/infection/infection/pull/513, it's very easy to create new mutators that unwrap parameter from function. Let's keep the list of desired mutators and discuss them here.

- [x] `array_chunk()` https://github.com/infection/infection/pull/558
- [x] `array_column()` https://github.com/infection/infection/pull/590
- [x] `array_combine` https://github.com/infection/infection/pull/550
- [x] `array_flip()` https://github.com/infection/infection/pull/529
- [x] `array_diff($a, $b);` -> `$a;` https://github.com/infection/infection/pull/544
- [x] `array_diff_assoc` https://github.com/infection/infection/pull/562
- [x] `array_diff_uassoc` https://github.com/infection/infection/pull/568
- [x] `array_diff_key` https://github.com/infection/infection/pull/563
- [x] `array_diff_ukey` https://github.com/infection/infection/pull/583
- [x] `array_pad` https://github.com/infection/infection/pull/680
- [x] `array_reverse()` https://github.com/infection/infection/pull/527
- [x]  `array_replace()` https://github.com/infection/infection/pull/536
- [x]  `array_reduce($array, $func, $initial)` -> `$initial` https://github.com/infection/infection/pull/533
- [x] `array_merge($a, $b, $c);` -> `$a;`, `$b;`, `$c` https://github.com/infection/infection/pull/542
- [x] `array_merge_recursive($a, $b, $c);` -> `$a;`, `$b;`, `$c` https://github.com/infection/infection/pull/594
- [x] `array_replace_recursive()` https://github.com/infection/infection/pull/545
- [x] `array_slice()` https://github.com/infection/infection/pull/598
- [x] `array_splice()` https://github.com/infection/infection/pull/605
- [x] `array_intersect($a, $b, $c);` -> `$a;`, `$b;`, `$c` https://github.com/infection/infection/pull/543
- [x] `array_intersect_assoc()` https://github.com/infection/infection/pull/679 
- [x] `array_intersect_key($a, $b, $c);` -> `$a;`, `$b;`, `$c` 
https://github.com/infection/infection/pull/584
- [x] `array_intersect_ukey($a, $b, $c);` -> `$a;`, `$b;`, `$c` 
https://github.com/infection/infection/pull/593
- [x] `array_intersect_uassoc()` https://github.com/infection/infection/pull/591
- [x] `array_keys` https://github.com/infection/infection/pull/555
[...]

PR #645, opened 2019-02-18
Issue #644: Add the new unwrap mutator ucwords.

drupol:
This PR:

* [x] Adds new unwrap mutator ucwords
* [x] Covered by tests
* [x] Doc PR: https://github.com/infection/site/pull/119

Fixes: #644 

Related: #514

### DIFF
--- a/src/Mutator/Unwrap/UnwrapUcWords.php
+++ b/src/Mutator/Unwrap/UnwrapUcWords.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017-2019, Maks Rafalko
+ * All rights reserved.
+ *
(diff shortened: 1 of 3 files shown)


drupol:
Sorry it's just an habit at work and other project. Basically it helps me finding a proper name for the branch :-)

localheinz:
@drupol 

What do you think about adding a test case that shows that the mutator works correctly when the `$delimiter` parameter is present?

For reference, see http://php.net/manual/en/function.ucwords.php.

drupol:
> @drupol
> 
> What do you think about adding a test case that shows that the mutator works correctly when the `$delimiter` parameter is present?
> 
> For reference, see http://php.net/manual/en/function.ucwords.php.

Ok will update the PR.

borNfreee:
Thank you @drupol

PR #650, opened 2019-03-03
WIP: Adding Ability To Load External Plugins

Fenikkusu:
This PR:

- [ ] Adds new feature ...
- [ ] Covered by tests

-- WORK IN PROGRESS --

Posting here for reference.

This is just a preliminary build. More to come. The general idea is to give the ability to load additional packages/'plugins' into infection when it runs. While additional logic is needed first, it gives, for example, the ability to split the testing frameworks into their own packages or splitting the mutations into packages, etc.


BackEndTea:
I like the idea of allowing users to create their own plugins, however there are some issues that come along with this

**Phar**
The phar we provide is scoped, so class names won't match with the non scoped version. This means we either have to forgo types, and hope the user pass in the right classes, or not scope a bunch of classes. (Or maybe a combination of those 2). 

If we choose not to scope certain classes, those can only be classes that are Infections own classes, as otherwise we may still run into issues with conflicting class declarations. Then however, users can't use any classes in their plugins that aren't infections own. So it wouldn't be possible to usethe Input or Output interfaces from symfony for example.

I'm not sure what the *correct* way of solving this would be. We could ignore plugins for the phar, but i'd like infection to provide the same experience, whether its installed through composer or as a phar.

**Exposing classes**
I don't think exposing our container to the users is a good idea. Doing so would make every class in there part of a Public API. I do think we need to expose an API of some sort, but we need to make A decision on exactly what classes are available to these plugins, and review those. 

We aren't in a `1.0` version yet, so we can be a bit more relaxed on BC breaks, but i'd prefer to start off with a very limited amount of classes that can be used in these plugins, and then add more as needed. This means we also make it easier for those who wish to write and maintain plugins, as they don't have to rewrite them every time we make a small change in the code base.

Fenikkusu:
FYI: I've not had the chance to go back to this recently, but the main reason I was doing this was to separate the test framework from the main package. In this way, for example, the codeception test framework logic would live in a separate project.

sanmai:
Isn't this idea like the proposal from #669?

theofidry:
Is the idea worth pursuing still? I'm not convinced that attempting to move the test frameworks in a separate package brings anything, plus it is making things more complicated for the scoped PHAR.

This also would only make sense if we expose a decent API to extend, which is far from being the case for now at least.

maks-rafalko:
> I'm not convinced that attempting to move the test frameworks in a separate package brings anything, plus it is making things more complicated for the scoped PHAR.

It definitely worth it. Please give me a couple of week to finish my RFC, I will cover all of these and this PR as well

maks-rafalko:
See https://github.com/infection/infection/issues/922

Fenikkusu:
Assuming This Isn't Needed Any More. Closing

PR #577, opened 2018-11-20
Use new config data when Infection has been executed in a first time (without .json or json.dist)

sidz:
This PR fixes https://github.com/infection/infection/issues/576

Issue #631, opened 2019-02-13
Target MSI on Travis with empty --filter


maks-rafalko:
Target MSI for _changed files_ should be `90%` on Travs.
Target MSI for the whle project is `67%` in `.travis.yml`.

When PR is created, and `--filter` is empty because no files are modified in `src/**`, target MSI should not be `90%` as it fails the builds, see https://github.com/infection/infection/pull/627.

PR #651, opened 2019-03-04
Use new config data when Infection has been executed at the first time

### LINKED ISSUE #576: Infection gets stuck in the first execution after configuration file is created

sanmai:
For one, `InfectionConfig` should be using known default values, but not access the configuration file that is non-existing at the time of configuration. For another, `JsonFile` should not continue as usual even if it is given a non-existing file.

Fixes #576

Kudos to @sidz for his effort in #577 (E2E test there is not exactly correct).

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -362,10 +362,9 @@ final class InfectionCommand extends BaseCommand
             $locator->locateOneOf(InfectionConfig::POSSIBLE_CONFIG_FILE_NAMES);
         } catch (\Exception $e) {
             $configureCommand = $this->getApplication()->find('configure');
-            $config = $this->getContainer()->get('infection.config');
 
             $args = [
-                '--test-framework' => $this->input->getOption('test-framework') ?: $config->getTestFramework(),
+                '--test-framework' => $this->input->getOption('test-framework') ?: TestFrameworkTypes::PHPUNIT,
             ];
 
             $newInput = new ArrayInput($args);
--- a/src/Config/InfectionConfig.php
+++ b/src/Config/InfectionConfig.php
@@ -35,6 +35,7 @@ declare(strict_types=1);
 
 namespace Infection\Config;
 
+use Infection\TestFramework\TestFrameworkTypes;
 use Symfony\Component\Filesystem\Filesystem;
 
 /**
@@ -152,7 +153,7 @@ class InfectionConfig
 
     public function getTestFramework(): string
     {
-        return $this->config->testFramework ?? 'phpunit';
+        return $this->config->testFramework ?? TestFrameworkTypes::PHPUNIT;
     }
 
     public function getInitialTestsPhpOptions(): string
--- a/src/Json/JsonFile.php
+++ b/src/Json/JsonFile.php
@@ -71,6 +71,10 @@ final class JsonFile
 
     private function parse(): void
     {
+        if (!is_file($this->path)) {
+            throw ParseException::invalidJson($this->path, 'file not found');
+        }
+
         $data = json_decode((string) file_get_contents($this->path));
 
         if (null === $data && JSON_ERROR_NONE !== json_last_error()) {
--- a/tests/Fixtures/e2e/Initial_Configuration/.gitignore
+++ b/tests/Fixtures/e2e/Initial_Configuration/.gitignore
@@ -1,0 +1,1 @@
+/infection.json*
--- a/tests/Fixtures/e2e/Initial_Configuration/Initial_Configuration/SourceClass.php
+++ b/tests/Fixtures/e2e/Initial_Configuration/Initial_Configuration/SourceClass.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Initial_Configuration;
+
+class SourceClass
+{
+    public function hello(): string
+    {
+        return 'hello';
+    }
+}
--- a/tests/Fixtures/e2e/Initial_Configuration/composer.json
+++ b/tests/Fixtures/e2e/Initial_Configuration/composer.json
@@ -1,0 +1,15 @@
+{
+    "require-dev": {
+        "phpunit/phpunit": "^6.5"
+    },
+    "autoload": {
+        "psr-4": {
+            "Initial_Configuration\\": "Initial_Configuration/"
+        }
+    },
+    "autoload-dev": {
+        "psr-0": {
+            "Initial_Configuration": "tests/"
+        }
+    }
+}
--- a/tests/Fixtures/e2e/Initial_Configuration/do_configure.expect
+++ b/tests/Fixtures/e2e/Initial_Configuration/do_configure.expect
@@ -1,0 +1,28 @@
+#!/usr/bin/env expect
+set timeout 20
+eval spawn $env(INFECTION)
+
+proc configure { input value } {
+    expect $input {
+        send $value
+    } timeout {
+        send_user "Test failed\n"
+        exit 1
+    }
+}
+
+configure "directories do you want to include" "1\r"
+configure "Any directories to exclude from" "\r"
+configure "timeout in seconds" "\r"
+configure "text log file?" "\r"
+
+expect {
+    "does not exist" {
+        close
+        exit 1
+    }
+    "Please note that some mutants will inevitably be harmless"
+}
+
+send_user "\nTest succeeded!\n"
+exit 0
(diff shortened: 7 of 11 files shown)


src/Console/InfectionContainer.php
```diff
@@ -232,12 +232,19 @@ public function __construct(array $values = [])
...
             $facade = new ConfigCreatorFacade($this['locator'], $this['filesystem']);
 
             return $facade->createConfig($input->getOption('configuration'));
         };
 
+        /*
+         * Configuration step needs a distinct instance of a configuration as it may be inexistent.
+         * Therefore we can avoid using this stale and invalid configuration later on.
+         */
+        $this['infection.config.initial'] = $infectionConfigMaker;
```
sidz: I'm not sure that this is a good idea to have more than one config class.

sanmai: What makes you concerned?

Another option is to re-initialize this `infection.config` maker, which is a painful job with lots of moving parts.

sanmai:
Builds are failing because of #631. Otherwise all seems fine.

src/Command/InfectionCommand.php
```diff
@@ -362,7 +362,7 @@ private function runConfigurationCommand(Locator $locator): void
             $locator->locateOneOf(InfectionConfig::POSSIBLE_CONFIG_FILE_NAMES);
         } catch (\Exception $e) {
             $configureCommand = $this->getApplication()->find('configure');
-            $config = $this->getContainer()->get('infection.config');
+            $config = $this->getContainer()->get('infection.config.initial');
```
borNfreee: Seems like the only one reason why `$config` variable is used here is to get default test framework name (`$config->getTestFramework()`) when `infection.json` **does not exist**.

And since config does not exist, it always defaults to `phpunit`.

https://github.com/infection/infection/blob/5b02e8e3c77b10732e69e7aad005946d06a431ad/src/Config/InfectionConfig.php#L155

`$this->config->testFramework` is empty when `infection.json` does not exist. Because of

https://github.com/infection/infection/blob/5b02e8e3c77b10732e69e7aad005946d06a431ad/src/Config/ConfigCreatorFacade.php#L80-L85

If we replace `$config->getTestFramework()` with `TestFrameworkTypes::PHPUNIT`, this will be the same PR as https://github.com/infection/infection/pull/577

I think we don't need `$config` instance here at all.

sanmai: I've seen this done in #577, but decided to keep the current behavior. For one because if configuration values are coming from one source, then you don't have to walk around all code changing them. Change them in one place and be done. For another we may have an idea to guess the value in the future. Then, again, everything will be as ready for this as it is. 

I think some of these reasons are behind why this was done like that in the first place.

Do you still think we should stick to PHPUnit defaulting with a plain constant?

borNfreee: Apologize for the late reply.

What you say is perfectly valid. The only issue with this particular line of code is that it was my mistake (read "bug") that I've added it here. Let me explain:

This is a commit that adds this config here: https://github.com/infection/infection/commit/2859730b9ffa5678e3a90226cae0d1647f0ff8f6

From git commit message:

> * Use default test framework if neither CLI nor config has the value for it

So, all I wanted to do, is to use **default** test framework when it's impossible to get it neither  from CLI option nor from the config. 

Why it was a mistake? Because `$config->getTestFramework()` is intended return value from `infection.json`, not a default TF. It just has a fallback.

What I think would be the right solution:

instead of calling instance level method `->getTestFramework()`, I would introduce a new static method `::getDefaultTestFramework()` and call it here statically.

1. It would resolve the issue with config initialization, because of a statically called method
2. It would make code more expressive because what we want to do here is really get default TF

sanmai: OK, I got what you're thinking. This makes sense.

sanmai: I went with `TestFrameworkTypes::PHPUNIT` for now.

(We can make an easy switch to `InfectionConfig::getDefaultTestFramework()` when it'll do anything in particular other than just returning `TestFrameworkTypes::PHPUNIT`.)

borNfreee:
Thank you @sanmai and @sidz

PR #653, opened 2019-03-08
Run cs-fixer

BackEndTea:
`array_key_exists` got some optimizations in 7.4(?) so it
is now included in the settings for the `native_function_invocation` fixer


BackEndTea:

PR #656, opened 2019-03-08
Stop traversal of interfaces and abstract methods

BackEndTea:
This PR:

- [x] Stops traversal of interfaces and abstract methods
- [x] Covered by tests

This means we no longer have to check for that in the mutation collecting visitor


borNfreee:
Thank you @BackEndTea

BackEndTea:
Thabk you @borNfreee

PR #657, opened 2019-03-08
Use php-parser build in function for blacklist

BackEndTea:
This PR:

- [x] simplifies blacklist logic for public method mutation


BackEndTea:
Thank you @borNfreee

PR #660, opened 2019-03-09
Preventing Issue Where Extra Commas Caused Infection To Ignore Filter

Fenikkusu:
This PR:

- [x] Adds new feature ...
- [x] Covered by tests

While working on another PR, I came across an issue where if there extra commas (IE: Trailing) in the filter option, it would completely negate the filter and run all files. 

    infection --filter=src/Service/Mailer.php,,src/Entity/Foobar.php

This resolves that issue by filtering out any empty elements in the array.

### DIFF
--- a/src/Finder/SourceFilesFinder.php
+++ b/src/Finder/SourceFilesFinder.php
@@ -80,7 +80,7 @@ final class SourceFilesFinder extends Finder
             return $this;
         }
 
-        $filters = explode(',', $filter);
+        $filters = array_filter(explode(',', $filter));
 
         foreach ($filters as $filter) {
             $this->filters[] = $filter;
--- a/tests/Finder/SourceFilesFinderTest.php
+++ b/tests/Finder/SourceFilesFinderTest.php
@@ -109,6 +109,32 @@ final class SourceFilesFinderTest extends TestCase
         }
     }
 
+    /**
+     * IE: --filter=1,,2,3,
+     */
+    public function test_extra_commas_in_filters(): void
+    {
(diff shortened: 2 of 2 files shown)


src/Finder/SourceFilesFinder.php
```diff
@@ -80,7 +80,7 @@ public function getSourceFiles(string $filter = ''): Finder
             return $this;
         }
 
-        $filters = explode(',', $filter);
+        $filters = array_filter(explode(',', $filter));
```
sanmai: Hopefully no one will ever need to run Infection over a file named `"0"`. What do you think?

Fenikkusu: Considering their PHP Install would have to be very specific to do that, I don't think it'll be a problem.

Fenikkusu:
Adding the below test will kill the mutant, but I'm not sure you care enough in this instance. Should I add it in too?

```php
    /**
     * Hacky way of preventing mutant, but gets the job done quickly.
     *
     * @throws \ReflectionException
     */
    public function test_it_calls_parent_construct(): void
    {
        $testSubject = new SourceFilesFinder(['tests/Fixtures/Files/Finder'], []);
        $reflectionProperty = new \ReflectionProperty(Finder::class, 'ignore');
        $reflectionProperty->setAccessible(true);

        $this->assertNotEquals(0, $reflectionProperty->getValue($testSubject));
    }
```

BackEndTea:
I dislike testing for internal state of an object. So i'd go without that test-case

Fenikkusu:
I can respect that. As it stands, I decided to take another route ;p. Hope to have a MR tonight or tomorrow.

borNfreee:
Thank you @Fenikkusu

PR #661, opened 2019-03-09
Adjusting Makefile To Allow Running Similar To Travis

Fenikkusu:
This PR:

- [x] Adds new feature ...
- [ ] Covered by tests

While working on another MR, I was getting errors in travis and couldn't figure out the issues. Once I realized it was because of extra parameters used to run in travis, I wanted to replicate them locally so I wouldn't have to fix/push/fix/push etc, as well as get an output of the issues.

This PR adds additional options to the Makefile to allow simple access to travis mode. By running `TEST_TRAVIS_OPTIONS=true make test` (Or another other run command), make will automatically append additional options that are pretty much the same as travis would run. I also added in the PHP_OPTIONS flag to pass options to php when running. For example, I kept getting a few memory limit issues, so I can now run `PHP_OPTIONS="-d memory_limit=-1" make test` (Yes, I know about the infection config option as well, but this came in hand for other items too).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ PHPSTAN_URL="https://github.com/phpstan/phpstan/releases/download/0.10.3/phpstan
 TEST_TRAVIS_OPTIONS ?= false
 ifeq ($(TEST_TRAVIS_OPTIONS),false)
 else
-	FILTER_OPTIONS := --filter=$(shell git whatchanged HEAD^! --name-only | grep \.php | tr '\n' ',')
+	FILTER_OPTIONS := --filter=$(shell git diff master --name-only | grep \.php | tr '\n' ',')
 	INFECTION_TRAVIS_OPTIONS := --ignore-msi-with-no-mutations --only-covered --min-msi=90 --threads=4 --min-covered-msi=75 --log-verbosity=none $(FILTER_OPTIONS)
 endif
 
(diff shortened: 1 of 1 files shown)


Fenikkusu:
Tests Failing Due To #631

Makefile
```diff
@@ -9,6 +9,16 @@ BOX_URL="https://github.com/humbug/box/releases/download/3.1.0/box.phar"
 PHP-CS-FIXER_URL="https://cs.sensiolabs.org/download/php-cs-fixer-v2.phar"
 PHPSTAN_URL="https://github.com/phpstan/phpstan/releases/download/0.10.3/phpstan.phar"
 
+TEST_TRAVIS_OPTIONS ?= false
+ifeq ($(TEST_TRAVIS_OPTIONS),false)
```
BackEndTea: Wouldn't it be better to do some if not eq to false, instead of an empty branch and then an else statement?
or even  `ifeq ($(TEST_TRAVIS_OPTIONS),true)` ?

theofidry:
Is this still relevant with the recent changes we did regarding the CI configuration and Makefile improvements?

Fenikkusu:
Honestly, I haven't been able to get back to this in quite a while so I'm not sure. I'll just close it for now.

PR #662, opened 2019-03-09
Update some dependencies to their latest version for the phar

BackEndTea:
This PR:

- [ ] Updates dependencies to their latest version

The Phar uses the lockfile, so its a good idea to keep those up to date


borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee

PR #663, opened 2019-03-10
Adding Line Numbers To Mutator Ignore List

Fenikkusu:
This PR:

- [x] Adds new feature ...
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/121

This started off as adding a 'falsePositives' feature to the config, but then I saw the ignore stuff in the json schema. Good to know. The only shortcoming I see is that there wasn't the ability to specify line numbers specifically. IE: Ignore this call in this function on line 123 but process call in the same function on line 456. This PR alters the logic to allow for line numbers in the search string as well.

IE:
```json
    "mutators": {
        "MethodCallRemoval": {
            "ignore": [
                "Infection\\Finder\\SourceFilesFinder::__construct::63"
            ]
        }
    }
```

Note: The config change to infection.json.dist will fix the testing in PR #660.

### DIFF
--- a/infection.json.dist
+++ b/infection.json.dist
@@ -10,5 +10,12 @@
         "badge": {
             "branch": "master"
         }
+    },
+    "mutators": {
+        "MethodCallRemoval": {
+            "ignore": [
+                "Infection\\Finder\\SourceFilesFinder::__construct::63"
+            ]
+        }
     }
 }
--- a/src/Mutator/Util/Mutator.php
+++ b/src/Mutator/Util/Mutator.php
@@ -67,7 +67,11 @@ abstract class Mutator
             return true;
         }
 
-        return !$this->config->isIgnored($reflectionClass->getName(), $node->getAttribute(ReflectionVisitor::FUNCTION_NAME, ''));
+        return !$this->config->isIgnored(
+            $reflectionClass->getName(),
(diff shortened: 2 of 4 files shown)


Fenikkusu:
The way I see it, this is more for false positives than anything. While it is true that they could change frequently, the idea here is to mark items that likely don't.

infection.json.dist
```diff
@@ -10,5 +10,12 @@
         "badge": {
             "branch": "master"
         }
+    },
+    "mutators": {
+        "MethodCallRemoval": {
+            "ignore": [
+                "Infection\\Finder\\SourceFilesFinder::__construct::63"
```
sanmai: This is for parent's `__construct`. It does some things, but they're quite hard to test, and that is kind of out of scope of our project.

Fenikkusu: That's the reason I put it in there. Granted, I do have to disagree to some point. It is quite easy to test. I had already commented the test code in #660. The problem is that you have to test using reflection, something most people tend to frown upon.

sanmai: Yeah, true. What do you think about adding this rationale to the PR description? It is not immediately obvious, that why I had to comment.

borNfreee:
Thank you @Fenikkusu

PR #667, opened 2019-03-15
Preliminary support for PCOV

### LINKED ISSUE #665: Add support for pcov
No.

Add support for https://github.com/krakjoe/pcov.

Thank you :)

sanmai:
This PR:

- [x] Add support for running Infection with PCOV (this is as easy as 1-2-3)

Fixes #665

I can verify that Infection can indeed recognize PCOV and do not fail with an error. Looks like this:

```
tests/Fixtures/e2e/PCOV_PHPUnit8$ php ../../../../bin/infection
     ____      ____          __  _

Running initial test suite...
... and so on
```

Getting this new changes tested on CI is a bit tricky, because even though PCOV is kind of exclusive with xdebug, they both can be loaded at the same time, and it is not like PHPUnit is willing to report exactly what it is using for coverage. Extra care should be taken to be sure that PCOV does its thing, like by disabling and/or not installing xdebug in the first place. In Debian this is done by:

    sudo phpdismod -v ALL xdebug
    sudo phpenmod -v ALL pcov

PCOV is supported only under PHPUnit 8, which only works under PHP 7.2 and later, therefore if we wish to test Infection under PHP 7.1, we can't upgrade to PHPUnit 8 yet. There's [an option of pcov/clobber](https://gist.github.com/krakjoe/8272c747ed247fedbed1c25a2831f09f), but that's kind of gross in my opinion. I'd rather have a single E2E test for PCOV for time being, than to resort to pcov/clobber with no clear benefit (and it is not our job to test PCOV).

### DIFF
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -188,6 +188,8 @@ ASCII;
             $this->consoleOutput->logRunningWithDebugger(\PHP_SAPI);
         } elseif (\extension_loaded('xdebug')) {
             $this->consoleOutput->logRunningWithDebugger('Xdebug');
+        } elseif (\extension_loaded('pcov')) {
+            $this->consoleOutput->logRunningWithDebugger('PCOV');
         }
     }
 }
--- a/src/Process/Coverage/CoverageRequirementChecker.php
+++ b/src/Process/Coverage/CoverageRequirementChecker.php
@@ -63,6 +63,7 @@ final class CoverageRequirementChecker
         return $this->skipCoverage
             || \PHP_SAPI === 'phpdbg'
             || \extension_loaded('xdebug')
+            || \extension_loaded('pcov')
             || XdebugHandler::getSkippedVersion()
             || $this->isXdebugIncludedInInitialTestPhpOptions();
     }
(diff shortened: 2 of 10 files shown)


src/Console/Application.php
```diff
@@ -188,6 +188,8 @@ private function logRunningWithDebugger(): void
             $this->consoleOutput->logRunningWithDebugger(\PHP_SAPI);
         } elseif (\extension_loaded('xdebug')) {
             $this->consoleOutput->logRunningWithDebugger('Xdebug');
+        } elseif (\extension_loaded('pcov')) {
+            $this->consoleOutput->logRunningWithDebugger('PCOV');
```
sanmai: I'm not entirely sure if we need this, but this may be a good idea in sake of verbosity.

sanmai:
> Does it need some more work (in progress label)?

Yes, there isn't a test that would work properly on CI.

sanmai:
There's an E2E test one can run, but it is not used on CI yet. Hopefully we can add it later.

borNfreee:
Thank you @sanmai

PR #671, opened 2019-03-22
Enhancement/simplify mutations collector

BackEndTea:
This PR:

- [x] Extracts functionality out of the `MutationsCollectorVisitor`
- [x] Allows Mutators to be decorated
- [x] Covered by tests

The main goal of this PR was to extract a bit more functionality out of the `MutationsCollectorVisitor`. The decoration of the mutators is simply a byproduct of that.

I'm not too happy about the `FilePathVisitor`, as its output is really only needed if a mutator is 'broken'. Otherwise all it does is eat up space. So another way to get the file path to the exception for error reporting would be preferred.

I tried to keep the commits as atomic as possible, so it may be easier to reviewer per commit, due to the size of the `Rename Mutator to BaseMutator` commit

### DIFF
--- a/src/Exception/InvalidMutatorException.php
+++ b/src/Exception/InvalidMutatorException.php
@@ -48,7 +48,7 @@ final class InvalidMutatorException extends \Exception
             sprintf(
                 'Encountered an error with the "%s" mutator in the "%s" file. ' .
                 'This is most likely a bug in Infection, so please report this in our issue tracker.',
-                $mutator::getName(),
+                $mutator->getName(),
                 $filePath
             ),
             0,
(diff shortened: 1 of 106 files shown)


src/Exception/InvalidMutatorException.php
```diff
@@ -48,7 +48,7 @@ public static function create(string $filePath, Mutator $mutator, \Throwable $pr
             sprintf(
                 'Encountered an error with the "%s" mutator in the "%s" file. ' .
                 'This is most likely a bug in Infection, so please report this in our issue tracker.',
-                $mutator::getName(),
+                $mutator->getName(),
```
borNfreee: What's the goal of this change?

BackEndTea: With that method being static, it becomes a lot harder (impossible?) to make a proper decorator for mutators.

BackEndTea:
I'm gonna close this for now, as it is producing some rather hard to debug issues, and i want to attempt this simplification in a smaller way.

PR #673, opened 2019-03-22
Do not mutate * to / and vice versa if one of the operands is numeric ±1.0

borNfreee:
Do not mutation `*` to `/` and vice versa if one of the operands is `1` or `-1` or `1.0` or `-1.0` to avoid equivalent mutations.

Fixes the issue mentioned by Badoo in their [talk about Mutation Testing](https://www.youtube.com/watch?v=EPU47QqtMiU):

```diff
- $a = $b * 1;
+ $a = $b / 1;
```

^ this is an equivalent mutant that can not be killed.


borNfreee:
Thank you @sanmai

PR #676, opened 2019-03-26
Get vendor autoloader location from composer config

### LINKED ISSUE #675: Set autoloaderInWorkingDirectory from composer config vendor-dir
My composer.json for some reason contains config/vendor-dir, thats not matched described paths in https://github.com/infection/infection/blob/master/bin/infection

As I know there is a way to get vendor dir from composer itself - `composer config vendor-dir`

Vendor dir can be configured by user with any path, I don't think there is a way for do this without calling composer get vendor dir

heahoh:
This PR:

- [x] Adds new feature - vendor dir fetched from composer config (if that accessible)
- [x] Covered by tests

Fixes #675


bin/infection
```diff
@@ -6,9 +6,12 @@
  * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
  */
 $autoloaderInWorkingDirectory = getcwd() . '/vendor/autoload.php';
+$vendorDir = exec('composer config vendor-dir');
```
sanmai: So is it like “composer config...” will be executed every time someone wants to run Infection?

tests/Fixtures/e2e/Composer_Config_Custom_Vendor_Dir/run_tests.bash
```diff
@@ -0,0 +1,17 @@
...
+
+if [ "$PHPDBG" = "1" ]
+then
+    phpdbg -qrr $INFECTION
+else
+    php $INFECTION
+fi
+
+diff -w expected-output.txt infection.log
+diff -w expected-file.txt infection-file.txt
```
sanmai: Is it not created by anyone?

bin/infection
heahoh: Missed this, moved to elseif branch

tests/Fixtures/e2e/Composer_Config_Custom_Vendor_Dir/run_tests.bash
heahoh: No, removed

heahoh:
uuf as I see there is no way to write test - vendor autoloader paths hardcoded inside e2etest runner

tests/Fixtures/e2e/Composer_Config_Custom_Vendor_Dir/run_tests.bash
sanmai: This was a fine script, but `expected-file.txt` and `infection-file.txt` were out of band.

sanmai:
E2ETest runner will gladly ignore any test that has `run_tests.bash` in it, just like your had.

bin/infection
```diff
@@ -9,6 +9,10 @@ $autoloaderInWorkingDirectory = getcwd() . '/vendor/autoload.php';
 
 if (is_file($autoloaderInWorkingDirectory)) {
     require_once $autoloaderInWorkingDirectory;
+} elseif ($vendorDir = exec('composer config vendor-dir')) {
+    if (is_file(getcwd() . '/' . $vendorDir . '/autoload.php')) {
+        require_once getcwd() . '/' . $vendorDir . '/autoload.php';
+    }
```
sanmai: Below this line there is the place where we try to guess the path to `autoload.php`. If you do ask composer for the vendor-dir here, then that guessing below is almost never going to be used.

So I think for now best would be to use this new measure as the last effort measure, when everything else fails. Otherwise put, this better be right before that `if (!isset($file)) { fwrite(STDERR, ...`.

Also, it does not make sense to run `composer config` if there's no `composer.json` present.

heahoh: While I'm running test __DIR__ constant (used in $files) and getcwd() (which used in $autoloaderInWorkingDirectory) is not equal. First one install classmap from infection/composer.json (that included infection sources) and the last one - from composer.json that located in my e2e test directory and contains only test source and phpunit. Maybe you can help me - I don't fully understand what's goes on production enviroment

maks-rafalko:
I'm revising issues from [0.14.0 milestone](https://github.com/infection/infection/milestone/25), guys what else needs to be done here?

@sanmai have you completed review? (I see you assigned it to yourself)

@heahoh will you be able to complete this PR or should we take over it?

sanmai:
> @sanmai have you completed review? (I see you requested a review from yourself)

I've tagged it for the future review. All previously requested changes are still in.

heahoh:
@maks-rafalko I don't fully understand what should I do to complete this pr.

bin/infection
```diff
@@ -38,6 +38,12 @@ $autoloaderInWorkingDirectory = getcwd() . '/vendor/autoload.php';
 
 if (is_file($autoloaderInWorkingDirectory)) {
     require_once $autoloaderInWorkingDirectory;
+} else {
+    $vendorDir = exec('composer config vendor-dir');
```
theofidry: relying on an `exec()` and `getcwd()` is quite tricky here... At the very least I would use it as a last restart after we try to load the other files. Plus as @sanmai noted you need to gracefully fail when it fails.

theofidry:
I'm not entirely convinced the proposed PR is the way to go and it's quite old as well. Instead I would rather recommend to use the scoped PHAR here instead of installing Infection as a dev dependency;

WDYT?

maks-rafalko:
Closing as very old ad stale

Issue #658, opened 2019-03-08
Family bc*-functions mutators (bcmath support)


majkel89:
Similarly to https://github.com/infection/infection/issues/654 there should be bcmath support

Example

```diff
- $c = bcadd($a, b);
+ $c = $a + $b;
```

I think this mutator should be enabled on demand and not by default.

There should be one mutator for all bcmath functions

borNfreee:
A big 👍 for this mutator.

> I think this mutator should be enabled on demand and not by default.

Could you please provide arguments? 

From my point of view, if developers' of a particular project have `ext-bcmath` installed, and intentionally use slow `bc-*` functions - they know what are they doing.

And Infection should force them write a test to prove `bc-*` function is used with a proper reason.

We probably should skip mutating if extension is not loaded, just for better performance:

```
// early return
protected function mutatesNode(Node $node): bool 
{
    if (!extension_loaded('bcmath')) {
        return false;
    }
}
```

or even don't add this mutator to mutator list if the extension is not loaded, to prevent its execution for *each* traversed node.

sanmai:
>> I think this mutator should be enabled on demand and not by default.

> Could you please provide arguments?

For one because bcmath functions work with strings of numbers, and first return value check will have these mutators busted. [Example:](https://3v4l.org/aKqim)
```php
var_dump('1' + '2'); \\ int(3)
var_dump(bcadd('1', '2')); \\ string(1) "3"
```
That assuming strict types are enabled.

borNfreee:
We can cast to `string` then.

```diff
- $c = bcadd($a, b);
+ $c = (string) ($a + $b);
```

sanmai:
This may work, but only before we get into the float territory. Bcmath functions [do not exactly work with floats](https://3v4l.org/IBHsc).
```
var_dump((string) ('11111111111111111111' + '2222222222222222222'));
// string(19) "1.3333333333333E+19"
var_dump(bcadd((string) ('11111111111111111111' + '2222222222222222222'), '1'));
// string(1) "1" - bummer!
var_dump(bcadd(bcadd('11111111111111111111', '2222222222222222222'), '1'));
// string(20) "13333333333333333334"
```

borNfreee:
> Otherwise put, just a single test that actually uses very large numbers will catch our mutations.

But this is the goal of this mutator - to force developers to write such a test (with big numbers or with floats), isn't it?

What I mean:

We have a method that uses `bcadd`:

```php
[...]

PR #664, opened 2019-03-10
Mutator/mbstring

majkel89:
This PR:

- [x] Adds mbstring extension mutator (MBString)
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/122

Relates to issue https://github.com/infection/infection/issues/654

Converts function calls like `mb_chr(74, 'utf-8')`,  `mb_strlen('test', 'utf-8')` to `chr('test')`, `strlen('test')`.

Supported functions:

**Pay attention to the function mapping when doing code review.**

 - `mb_chr` -> `chr`
 - `mb_ord` -> `ord`
 - `mb_parse_str` -> `parse_str`
 - `mb_send_mail` -> `mail`
 - `mb_strcut` -> `substr`
 - `mb_stripos` -> `stripos`
 - `mb_stristr` -> `stristr`
 - `mb_strlen` -> `strlen`
 - `mb_strpos` -> `strpos`
 - `mb_strrchr` -> `strrchr`
 - `mb_strripos` -> `strripos`
 - `mb_strrpos` -> `strrpos`
 - `mb_strstr` -> `strstr`
 - `mb_strtolower` -> `strtolower`
 - `mb_strtoupper` -> `strtoupper`
 - `mb_substr_count` -> `substr_count`
 - `mb_substr` -> `substr`
 - `mb_convert_case` -> `strtolower`, `strtoupper`, `ucwords` depending on mode argument

New profile `@extensions` created to hold this mutator. It is on by default.

**Configuration**

```json
    "mutators": {
        "MBString": {
            "settings": {
                "mb_strlen": false,
                "mb_ord": false,
            }
        }
    },
```

PR #678, opened 2019-03-30
New mutator for bcmath

majkel89:
This PR:

- [x] Adds BCMath extension mutator (MBString)
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/125

Fixes https://github.com/infection/infection/issues/658

Supported functions:
 - `bcadd($a, $b, $mod)` -> `(string) ($a + $b)`
 - `bcdiv($a, $b, $mod)` -> `(string) ($a / $b)`
 - `bcmod($a, $b, $mod)` -> `(string) ($a % $b)`
 - `bcmul($a, $b, $mod)` -> `(string) ($a * $b)`
 - `bcpow($a, $b, $mod)` -> `(string) ($a ** $b)`
 - `bcsub($a, $b, $mod)` -> `(string) ($a - $b)`
 - `bcsqrt($a, $mod)` -> `(string) sqrt($a)`
 - `bcpowmod($a, $b, $c, $mod)` -> `(string) (pow($a, $b) % $c)`
 - `bccomp($a, $b, $mod)` -> `$a <=> $b`

New profile @extensions created to hold this mutator. It is on by default.

**Configuration**

```
"mutators": {
  "BCMath": {
    "settings": {
      "bcpowmod": false,
      "bcsqrt": false,
    }
  }
},
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,8 @@
         "symfony/process": "3.4.2"
     },
     "require-dev": {
-        "phpunit/phpunit": "^7.5"
+        "phpunit/phpunit": "^7.5",
+        "ext-bcmath": "*"
     },
     "config": {
         "platform": {
--- a/composer.json
+++ b/composer.json
@@ -60,8 +60,7 @@
         "symfony/process": "3.4.2"
     },
     "require-dev": {
-        "phpunit/phpunit": "^7.5",
-        "ext-bcmath": "*"
+        "phpunit/phpunit": "^7.5"
     },
     "config": {
         "platform": {
--- a/devTools/Dockerfile-php71-xdebug
+++ b/devTools/Dockerfile-php71-xdebug
@@ -1,6 +1,7 @@
 FROM php:7.1-cli
 
-RUN pecl install xdebug && docker-php-ext-enable xdebug
+RUN pecl install xdebug && docker-php-ext-enable xdebug && \
+    docker-php-ext-install -j$(nproc) bcmath
 RUN apt-get update && apt-get install -y --no-install-recommends \
     expect git zip \
     && rm -rf /var/lib/apt/lists/*
--- a/devTools/Dockerfile-php71-xdebug
+++ b/devTools/Dockerfile-php71-xdebug
@@ -1,7 +1,6 @@
 FROM php:7.1-cli
 
-RUN pecl install xdebug && docker-php-ext-enable xdebug && \
-    docker-php-ext-install -j$(nproc) bcmath
+RUN pecl install xdebug && docker-php-ext-enable xdebug
 RUN apt-get update && apt-get install -y --no-install-recommends \
     expect git zip \
     && rm -rf /var/lib/apt/lists/*
--- a/src/Mutator/Extensions/BCMath.php
+++ b/src/Mutator/Extensions/BCMath.php
@@ -72,9 +72,7 @@ final class BCMath extends Mutator
             return false;
         }
 
-        $functionName = $node->name->toLowerString();
-
-        return isset($this->converters[$functionName]) && \function_exists($functionName);
+        return isset($this->converters[$node->name->toLowerString()]);
     }
 
     private function setupConverters(array $functionsMap): void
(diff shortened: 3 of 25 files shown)


src/Mutator/Extensions/BCMath.php
```diff
@@ -0,0 +1,163 @@
...
+            )),
+            'bcpowmod' => $this->mapCheckingMinArgs(3, $this->mapCastToString(
+                $this->mapPowerModulo()
+            )),
+            'bccomp' => $this->mapCheckingMinArgs(2,
+                $this->mapBinaryOperator(Node\Expr\BinaryOp\Spaceship::class)
+            ),
+        ];
+
+        $functionsToRemove = \array_filter($functionsMap, static function ($isOn) {
```
borNfreee: please add a typehint for `$isOn`

src/Mutator/Util/MutatorProfile.php
```diff
@@ -232,6 +232,10 @@ final class MutatorProfile
         Mutator\Unwrap\UnwrapUcWords::class,
     ];
 
+    public const EXTENSIONS = [
+        Mutator\Extensions\BCMath::class,
+    ];
+
     public const DEFAULT = [
```
borNfreee: Shouldn't `@extensions` be added here to the default profile? As it is done for `mbstring` extension?

https://github.com/infection/infection/pull/664/files#diff-071045e7645d747e46b60edd948a3537R254

composer.json
```diff
@@ -60,7 +60,8 @@
         "symfony/process": "3.4.2"
     },
     "require-dev": {
-        "phpunit/phpunit": "^7.5"
+        "phpunit/phpunit": "^7.5",
+        "ext-bcmath": "*"
```
borNfreee: Infection itself should not depend on `ext-bcmath`. We don't use any of those functions.

devTools/Dockerfile-php71-xdebug
```diff
@@ -1,6 +1,7 @@
 FROM php:7.1-cli
 
-RUN pecl install xdebug && docker-php-ext-enable xdebug
+RUN pecl install xdebug && docker-php-ext-enable xdebug && \
+    docker-php-ext-install -j$(nproc) bcmath
```
borNfreee: As in the previous comment, from my point of view - Infection does not depend on `bcmath`, thus it's not needed.

`\function_exists($functionName);` in this condition already covers all we need:

```php
public function mutatesNode(...)
{
    ...
    return isset($this->converters[$functionName]) && \function_exists($functionName);
}
```

composer.json
sanmai: IMO `function_exists` should not be here, for one, because there's no actual need for it, and for another, Infection can be run under PHP without bcmath, where tested code can have it loaded dynamically. Best if we make no assumtions about missing or present modules: if there's a function in code, and we can mutate it, we should.

src/Mutator/Extensions/BCMath.php
majkel89: fixed - also for MBString

src/Mutator/Util/MutatorProfile.php
majkel89: fixed by merging with master

composer.json
majkel89: That was leftover from testing - removed

devTools/Dockerfile-php71-xdebug
majkel89: change reverted

src/Mutator/Extensions/BCMath.php
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        return isset($this->converters[$node->name->toLowerString()]);
+    }
+
+    private function setupConverters(array $functionsMap): void
+    {
+        $converters = [
+            'bcadd' => $this->mapCheckingMinArgs(2, $this->mapCastToString(
+                $this->mapBinaryOperator(Node\Expr\BinaryOp\Plus::class)
```
sanmai: Seems all right, but I feel somewhat uneasy about these nested mappers. Let me check with an IDE later if there's a way to improve the situation here.

(What if the were called differently? Like `makeBinaryOperatorMapper`, as that's these functions do: they don't map things by themselves, but instead return, or make, a mapping function.)

majkel89: fixed also for MBString.

PR #685, opened 2019-05-01
Add test cases for all mutators

BackEndTea:
This PR:

- [x] Adds new tests to make sure mutators can handle more 'exotic' code

All mutators will now be used to parse files in the 'CodeSample' folder. Wheter or not they mutate
this code doesn't matter, the test here is to make sure that this code does not crash the mutator.

One issue found by this test has been fixed, namely the `DecrementInteger` mutator could crash when 0 was compared to a function with a variable name.

This means we can remove all the 'It does not break on ...' test cases that we now have for every mutator that in some way deals with functions/methods.

### DIFF
--- a/tests/Fixtures/CodeSamples/ReturnTypes.php
+++ b/tests/Fixtures/CodeSamples/ReturnTypes.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Infection\CodeSamples;
+
+class ReturnTypes
+{
+    public function withReturnType(): int
+    {
+        return 3;
+    }
+
+    public function nullableReturnType(): ?int
+    {
+        return null;
+    }
+
+    public function withoutReturnType()
+    {
+        return 3;
+    }
+
+    public function selfReturn(): self
+    {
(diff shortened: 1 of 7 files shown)


tests/Fixtures/CodeSamples/ReturnTypes.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        return null;
+    }
+
+    public function withoutReturnType()
+    {
+        return 3;
+    }
+
+    public function selfReturn(): self
```
borNfreee: https://3v4l.org/2tQLa

it's possible to have `parent` as a return typehint, as well as a parameters typehint

BackEndTea: Cool, didn't know that was an option in PHP, ill add it gere

borNfreee:
Thank you @BackEndTea

BackEndTea:
Thank you @borNfreee

PR #549, opened 2018-11-05
Fix variable names in mutators

BackEndTea:
This PR:

- [x] Makes sure that variable function names don't break unwrap mutators
- [x] Covered by tests

CC @localheinz, once this gets merged could you update your open PRs to include this test case as well?

PR #687, opened 2019-05-01
Chore/phpstan upgrade

BackEndTea:
This PR:

- [x] Upgrades PHPStan to 0.11.5
- [x] Fixes new errors found by PHPStan
- [x] Removes the two rules about the `Node` class missing properties or methods

Blocked by #685 Since that fixes issues surrounding names on nodes for decrement integer

That last one is the biggest change here. In most cases it means updating the doc comment to accurately reflect what is actually going into the method. 
In some other cases it means assigning a property to a value, and then telling PHPStan what it is through `/** @var */` notation.

The big upside of no longer ignoring these issues is that we now are notified of errors with `$node->name` during development. Like the ones found in:
* #685  
* #384
* #549

What do you think, should we stop ignoring these errors, and have a bit more code, like in this PR? 
Or should we ignore the errors like previously?
Or should we add rules for specific files, stating for exactly which files these rules can be ignored?

### DIFF
--- a/devTools/phpstan-src.neon
+++ b/devTools/phpstan-src.neon
@@ -1,7 +1,5 @@
 parameters:
     ignoreErrors:
-        - '#Access to an undefined property PhpParser\\Node(.*?)#'
-        - '#Call to an undefined method PhpParser\\Node(.*?)#'
         - '#expects .*, array<string>\|bool\|string\|null given#'
         - '#cast array<string>\|bool\|string\|null to#'
         - '#Return type \(void\) of method Infection\\Visitor\\.*::.*\(\) should be compatible with return type .* of method PhpParser\\Node.*::.*\(\)#'
--- a/devTools/phpstan-src.neon
+++ b/devTools/phpstan-src.neon
@@ -4,5 +4,5 @@ parameters:
         - '#cast array<string>\|bool\|string\|null to#'
         -
             path: '%currentWorkingDirectory%/src/Process/Runner/Parallel/ParallelProcessRunner.php'
-            message: '#Left side of [\||\&]{2} is always true#'
+            message: '#Left side of [\&]{2} is always true#'
 
--- a/devTools/phpstan-src.neon
+++ b/devTools/phpstan-src.neon
@@ -4,6 +4,8 @@ parameters:
         - '#Call to an undefined method PhpParser\\Node(.*?)#'
         - '#expects .*, array<string>\|bool\|string\|null given#'
         - '#cast array<string>\|bool\|string\|null to#'
-        - '#Offset array<string>\|bool\|string\|null does not#'
-        - '#Possibly invalid array key type array<string>\|bool\|string\|null#'
-        - '#Access to undefined constant PhpParser\\NodeTraverser::DONT_TRAVERSE_CURRENT_AND_CHILDREN#'
+        - '#Return type \(void\) of method Infection\\Visitor\\.*::.*\(\) should be compatible with return type .* of method PhpParser\\Node.*::.*\(\)#'
+        -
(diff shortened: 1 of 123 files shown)


devTools/phpstan-src.neon
```diff
@@ -1,9 +1,9 @@
...
-        - '#Access to an undefined property PhpParser\\Node(.*?)#'
-        - '#Call to an undefined method PhpParser\\Node(.*?)#'
         - '#expects .*, array<string>\|bool\|string\|null given#'
         - '#cast array<string>\|bool\|string\|null to#'
-        - '#Offset array<string>\|bool\|string\|null does not#'
-        - '#Possibly invalid array key type array<string>\|bool\|string\|null#'
-        - '#Access to undefined constant PhpParser\\NodeTraverser::DONT_TRAVERSE_CURRENT_AND_CHILDREN#'
+        - '#Return type \(void\) of method Infection\\Visitor\\.*::.*\(\) should be compatible with return type .* of method PhpParser\\Node.*::.*\(\)#'
+        -
+            path: '%currentWorkingDirectory%/src/Process/Runner/Parallel/ParallelProcessRunner.php'
```
borNfreee: What is the line of this error? Trying to understand the reason..

BackEndTea: The left side of the || and the && are always true according to the inspector.
From what i can tell that is actually not the case, because it is running in paralel, meaning that things can change between the first and second check.

borNfreee:
> Blocked by #685 Since that fixes issues surrounding names on nodes for decrement integer

#685 has been merged

devTools/phpstan-src.neon
```diff
@@ -1,9 +1,9 @@
 parameters:
     ignoreErrors:
-        - '#Access to an undefined property PhpParser\\Node(.*?)#'
-        - '#Call to an undefined method PhpParser\\Node(.*?)#'
         - '#expects .*, array<string>\|bool\|string\|null given#'
         - '#cast array<string>\|bool\|string\|null to#'
-        - '#Offset array<string>\|bool\|string\|null does not#'
-        - '#Possibly invalid array key type array<string>\|bool\|string\|null#'
-        - '#Access to undefined constant PhpParser\\NodeTraverser::DONT_TRAVERSE_CURRENT_AND_CHILDREN#'
+        - '#Return type \(void\) of method Infection\\Visitor\\.*::.*\(\) should be compatible with return type .* of method PhpParser\\Node.*::.*\(\)#'
```
borNfreee: I guess this is because we have `void` in our visitors, e.g. in `\Infection\Visitor\MutationsCollectorVisitor`

```php
public function leaveNode(Node $node): void
```

while it should be `null|int|Node|Node[]`

Is it better to fix the issue rather than ignore it?

BackEndTea: I have fixed this issue, by adding the correct return types, and returning `null` explicitly from some methods.

devTools/phpstan-src.neon
sanmai: Here's what PHPStan tells us:
```
 ------ --------------------------------- 
  Line   ParallelProcessRunner.php        
 ------ --------------------------------- 
  92     Left side of && is always true.  
  122    Left side of || is always true.  
 ------ --------------------------------- 
```
On line 92 we truly have a bogus situation, because if we remove that supposedly always-true check, `startProcess()` will try to unshift a process from an empty array, and fail. I could not understand why PHPStan decided it is so, so this could be a bug in PHPStan.

On the contrary, situation on the line 122 is different. A job is always either inside `$processesQueue` or `$currentProcesses`, so it may really make sense to remove that check from the left side of `||`. As far as I can tell E2E tests are okay with this change.

BackEndTea:
I fixed the issue in the `ParallelProcessRunner`, and updated the ignore rule accordingly

Issue #694, opened 2019-05-28
Allow percent numbers with decimals in --min-msi and --min-covered-msi


arudolph-cxt:
**Is your feature request related to a problem? Please describe.**
We calculate the MSI and mutation coverage manually from the summary log and store the values with 2 decimals. When we add those values to the ci command-line script like so: `... --min-msi=15.77` it fails:
`[ERROR] The minimum required MSI percentage should be 15.77%, but actual is 15%. Improve your tests!`
Although the actual percentage should be compared with 15.77424023154848% to be exact and therefore pass.

**Describe the solution you'd like**
The comparison should use float values instead of rounded-down integers to decide if the ci fails.

**Describe alternatives you've considered**
Rounding down the value to an int, with the drawback of having much less accuracy on bigger projects.

arudolph-cxt:
According to the code, the flooring seems to be deliberate:
https://github.com/infection/infection/blob/7894cf5f6030c3c499004d57f7a29db52aeb2af8/src/Mutant/MetricsCalculator.php#L161
Is there a reason for this? Since the values seem to be available as floats, this looks as simple as removing the floor call and ensuring the values are calculated as float and not auto-casted into integer values.
Unless i'm missing something and there was a good reason for this?

This commit seems to be the last change https://github.com/infection/infection/commit/b2a7a2d50d2be36c963abd88298ddef67ee0adb9 and i understand why the change from `round` to `floor` was done, but why there was actual rounding in the first place seems confusing to me.

borNfreee:
Personally, I don't remember any reasons for rounding.

arudolph-cxt:
I'm currently working on one and as expected, two tests fail as they compare against rounded values.

My current approach would be to just use float values instead, a more nuanced approach would be to allow control over the accuracy/decimal amount maybe with a parameter.
I think I can provide a PR shortly.

PR #695, opened 2019-05-29
Improve min ci score accuracy with the option to pass float values as well

arudolph-cxt:
This PR:

- [X] Covered by tests

Fixes https://github.com/infection/infection/issues/694

- Internally, values are compared as floats instead of floored integers.
- if min-msi or min-coverage-msi is set as integer, the comparison worked as before
- if min-msi or min-coverage-msi is set as a float, the comparison takes the higher accuracy into account, instead of comparing both as floored integer as before
- if the score is below min-*msi, the error message displays the values with their expected decimal accuracies, depending on the input. For example with `--min-msi=10.55` the error message displays the actual percentage up to two decimal points as well.
- in the console output summary, the metrics display the full float value. (Previously: `15%`, New: `15.774240231548%`)

### DIFF
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -100,6 +100,24 @@ ASCII;
             $output->setVerbosity(OutputInterface::VERBOSITY_QUIET);
         }
 
+        if ($input->hasOption('min-msi')) {
+            $minMsi = \trim($input->getOption('min-msi'));
+            $dotPos = \strpos($minMsi, '.');
+
+            if ($dotPos !== false) {
+                $this->consoleOutput->setMsiDecimalAccuracy(\strlen($minMsi) - $dotPos + 1);
+            }
+        }
+
+        if ($input->hasOption('min-covered-msi')) {
+            $minCoveredMsi = \trim($input->getOption('min-covered-msi'));
+            $dotPos = \strpos($minCoveredMsi, '.');
+
+            if ($dotPos !== false) {
+                $this->consoleOutput->setCoverageDecimalAccuracy(\strlen($minCoveredMsi) - $dotPos + 1);
+            }
+        }
+
         $this->logRunningWithDebugger();
 
         if (!$this->isAutoExitEnabled()) {
--- a/src/Console/ConsoleOutput.php
+++ b/src/Console/ConsoleOutput.php
@@ -60,6 +60,16 @@ final class ConsoleOutput
      */
     private $io;
 
+    /**
+     * @var int
+     */
+    private $msiDecimalAccuracy = 0;
+
+    /**
+     * @var int
+     */
+    private $coverageDecimalAccuracy = 0;
+
     public function __construct(SymfonyStyle $io)
     {
         $this->io = $io;
@@ -112,17 +122,21 @@ final class ConsoleOutput
             throw MsiCalculationException::create('min-msi');
         }
 
-        $this->io->error(
+        $errorMessage = $type === TestRunConstraintChecker::MSI_FAILURE ?
             sprintf(
                 self::CI_FLAG_ERROR,
-                ($type === TestRunConstraintChecker::MSI_FAILURE ? 'MSI' : 'Covered Code MSI'),
-                $minMsi,
-                ($type === TestRunConstraintChecker::MSI_FAILURE ?
-                    $metricsCalculator->getMutationScoreIndicator() :
-                    $metricsCalculator->getCoveredCodeMutationScoreIndicator()
-                )
-            )
-        );
+                'MSI',
+                number_format($minMsi, $this->getMsiDecimalAccuracy()),
+                number_format($metricsCalculator->getMutationScoreIndicator(), $this->getMsiDecimalAccuracy())
+            ) :
+            sprintf(
+                self::CI_FLAG_ERROR,
+                'Covered Code MSI',
+                number_format($minMsi, $this->getCoverageDecimalAccuracy()),
+                number_format($metricsCalculator->getCoveredCodeMutationScoreIndicator(), $this->getCoverageDecimalAccuracy())
+            );
+
+        $this->io->error($errorMessage);
     }
 
     public function logMissedDebuggerOrCoverageOption(): void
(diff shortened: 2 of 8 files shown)


arudolph-cxt:
it's debatable up to which point the decimal accuracy is interesting, especially in the metrics summary.
Personally, i think it is fine to display the raw value like so:
```
Metrics:
         Mutation Score Indicator (MSI): 15.774240231548%
         Mutation Code Coverage: 34.587554269175%
         Covered Code MSI: 45.606694560669%
```
but i can understand if someone would say, that only up to the xth decimal is all one will ever need.
One thing to keep in mind is, that the bigger the codebase is, the more accuracy is needed.

sanmai:
> Mutation Score Indicator (MSI): 15.774240231548%

Not a big fan of this, tbh. But this makes sense if you're comparing raw digits. Say, if someone specifies minimal MSI as 15.7742%, it will help them if they could see that the actual MSI is 15.7741, but not just 15.77. And as you're saving requested accuracy, you can show the actual MSI with that accuracy. What do you think?

arudolph-cxt:
@sanmai showing the same accuracy in the console output as the input would be a compromise. I'll have to see how easy it is to pass the digit accuracy to whatever is printing the console summary, but I don't think it would be too much trouble.

As I'd like to put the full accuracy _somewhere_, I have another idea: maybe the logs/summary file would be a good place to put the MSI, MSI-coverage and code-covered-msi raw values into.
Or was it a deliberate decision to not include them there? I'm speaking about the file that contains output like this:
```
Total: 1392
Killed: 213
Errored: 0
Escaped: 260
Timed Out: 6
Not Covered: 913
```

Concerning the tests: Any suggestions what to improve?
The travis build seems to fail because it detects mutations in code I dont recognize and don't seem to have touched. Advice would be appreciated.

sanmai:
> As I'd like to put the full accuracy _somewhere_, I have another idea: maybe the logs/summary file would be a good place to put the MSI, MSI-coverage and code-covered-msi raw values into.

As I were reading this, I though maybe, just maybe, we'll be better off if we ditch that autodetection, and add a configuration option instead? Then showing the full accuracy (or any desired accuracy) will be a matter of changing a configuration variable.

arudolph-cxt:
hm, a configuration value that only controls how many digits to display? Not sure if that's needed.
Altough i think it would make sense if the configurable accuracy not only controls how accurate to display a value, but also how accurate the comparison would be?

so if i set the accuracy to two, `15.11 >= 15.11` would pass, but with the accuracy set to three and comparing it to a previous value like `15.110 >= 15.119` would fail.
I'd be okay with that.

sanmai:
How about if it'll work on both counts?

arudolph-cxt:
Ok, so to review the suggested changes:

- Remove auto-detection of decimal accuracy from both console arguments
- Add a config value (like `outputDecimalAccuracy`) to set the general decimal accuracy by number of decimals
- Make the three % values in Console output summary and in the error message use this accuracy
- Compare all values as actual full float values to determine ci pass/fail
- Print the full numbers into the summary log file
- Adjust tests that take the summary file output into account
- Improve tests

borNfreee:
> * Remove auto-detection of decimal accuracy from both console arguments
> * Add a config value (like `outputDecimalAccuracy`) to set the general decimal accuracy by number of decimals
> * Make the three % values in Console output summary and in the error message use this accuracy
> * Compare all values as actual full float values to determine ci pass/fail
> * Print the full numbers into the summary log file
> * Adjust tests that take the changes in the summary file output into account
> * Improve tests

sounds good to me, what do you think @infection/core ?

```bash
Metrics:
         Mutation Score Indicator (MSI): 15.774240231548%
         Mutation Code Coverage: 34.587554269175%
         Covered Code MSI: 45.606694560669%
```

Looks weird. At least, I would vote for leaving 2 digits as a _default_ accuracy to be displayed.

BackEndTea:
We could compare the values to an arbirary precision, and in the error message, display the full percentage when informing users that they don't have the required msi. For the Metrics part of the output, i'd say cap it off at 2 decimals.

sanmai:
> For the Metrics part of the output, i'd say cap it off at 2 decimals.

Why not show it with the configured accuracy, as proposed?

arudolph-cxt:
>Best if they would be compared with the configured accuracy, floored down to it. Exactly the same rationale why we're flooring these values now.

One case i can think of where this could be a bit awkward is this:
Imagine you run the command with `--min-msi=15.119999` and a configured decimal accuracy of 2, but the actual msi score is `15.111111%`. The ci would pass due to the rounding down of both values to `15.11`. (hence me not really liking the rounding in general)

At least then a warning should point out that the actual score is below the requested min-msi, but due to the accuracy setting was made to pass. Altough i have a hard time thinking why anyone would want that behaviour in a real scenario. Maybe @sanmai could get into a bit more detail why reducing accuracy and provoking these cases is desired?

Some more background on why this concerns me: We pull the `--min-msi` and `--min-covered-msi` from the gitlab cache and never "see" or enter these values ourselves a lot, so if a higher accuracy is pulled from the cache, but due to a config value it gets rounded down, that would seem a bit odd for me. Especially since we're dealing with ci tools here that should exhibit more accurate and predictable behavior in my opinion.

sanmai:
OK, you have your point. How about if we round only the actual MSI score, while leave the required minimal MSI exactly as it is?

arudolph-cxt:
>On second thought, what kind of harm there could be if we compare the raw values?.. If we do not round them for comparison at all, everything should be all right. Am I missing anything here?

That was my initial plan in this PR as well. The only concern was how these values should be displayed in the console output, since many (predictably) didn't like seeing a score value outputted as `15.774240231548%`
While I support the idea of rounding values in the console output, there needs to be a way to let the user see why it failed. if the error message then displays "`15.11%` is not enough, min value is `15.11%`" it would do nothing else than confuse.

BackEndTea:
> While I support the idea of rounding values in the console output, there needs to be a way to let the user see why it failed. if the error message then displays "15.11% is not enough, min value is 15.11%" it would do nothing else than confuse.

I think we can do this by having the 'normal' display be the rounded down metrics, and for the error message, we can go with as much decimals as possible. 

I'm somewhat against configuring this, but that may just be because i personally don't have a use case for it.

maks-rafalko:
Closing in favor of https://github.com/infection/infection/pull/701

PR #700, opened 2019-06-12
Bump requirements up to PHP 7.2

sanmai:
Taking PHP 7.2 as a baseline we'll be able to fully utilize features PHPUnit 8 brings on the table, most notable [reportedly much faster code coverage reporting](https://twitter.com/slamzoe/status/1138432239455354880) with PCOV.

(Makefile probably need updating.)

(I guess the switch to PHPUnit 8 + pcov driver has to come in a later PR.)


BackEndTea:
👍 To anything that reduces build time on travis/ appveyor 😅 

Seems like a sensible change, as long as we communicate it properly to the end users.

sanmai:
Yeah, builds are failing because of #631.

PR #701, opened 2019-06-12
Do not round down values in MetricsCalculator

### LINKED ISSUE #694: Allow percent numbers with decimals in --min-msi and --min-covered-msi

sanmai:
But round them down for display purposes in the relevant logger.

Provided that Infection is being run like so:

    vendor/bin/infection -j2 --min-msi=71.43

It'll report the whole unrounded metric in the case of failure:

> The minimum required MSI percentage should be 71.43%, but actual is 71.428571428571%. Improve your tests!

If there isn't a failure, then there's no change in output.

This change is backward compatible because if `floor(71.42) < 72`, then `71.42 < 72` also true.

Fixes #694

Related to #695

(Showing values with the same precision is a bit tricky, considering that [PHP will gladly switch to using scientific notation](https://stackoverflow.com/questions/2430084/php-get-number-of-decimal-digits#comment45763898_2430144), and then all checks like simple `strpos($float, '.')` will naturally go out of window. This feature will require extensive testing, and not in a way related to solving the practical problem from #694, so I suggest we keep showing the value as it is, for now.)


sanmai:
Builds are failing because MSI is lower than 90 for the files changed here, but I'm not sure it is exactly the best idea to work on improving MSI in this bug-fixing PR.

maks-rafalko:
anyone wants to make the second review or fine to merge?

maks-rafalko:
Thank you @sanmai

PR #702, opened 2019-06-13
Update PHP version badge to 7.2.9+

maks-rafalko:
Continue our internal discussion, let's see what else needs to be configured to unlock Merge button for Members.

PR #706, opened 2019-06-17
Makefile improvements

theofidry:
This PR focus on the `Makefile` and attempts to:

- Makes the discovery of the available tasks easier
- Make the `Makefile` easier to read/maintain (from a maintainer PoV)

The exhaustive list of changes:

- Change the default command to point to a friendly help message listing the available commands instead of blindly running "everything" (it was untrue since a couple of commands where not run by the `all` target, so a few commands where more meant for a regular dev workflow but hidden not easily discoverable). Here is the current output:

Screen Shot 2019-06-17 at 22 07 45

- Remove `/devTools/.gitignore` in favour of ignoring the content in the main `.gitignore` (it was the only case where we used a sub `.gitignore` instead of one at the root); Besides being more consistent, it is IMO easier to see what is ignored or not by having to check only the root file instead of checking left and right for a `.gitignore` file
- Move `build/bin/infection.phar` to `build/infection.phar`: not a critical change however might be reverted...
- Make use of the full options instead of short options when possible in the `Makefile` as it's easier to read and requires less familiarity with the diverse tools (the shortcuts are less explicit)
- Move the various tools downloaded PHARs under `.tools` making it cacheable

There is a few TODOs introduced in this PR but I think they should be handled separately


.gitignore
```diff
@@ -1,13 +1,18 @@
 /.idea
+/.composer/
```
theofidry: Not sure why this directory popped up... Might be a Composer mismanipulation from my part or the Symfony flex plugin. In any case I don't think it really hurts to have it here

Makefile
```diff
@@ -1,137 +1,205 @@
-PHP_CS_FIXER_FUTURE_MODE=1
-PHPSTAN=./phpstan.phar
-PHP-CS-FIXER=./php-cs-fixer-v2.phar
-PHPUNIT=vendor/bin/phpunit
-INFECTION=build/bin/infection.phar
+.DEFAULT_GOAL := help
```
theofidry: The new Makefile structure is as follows:

```Makefile
.DEFAULT_GOAL := help

.PHONY: help
help: ...

# Variables

...

# Commands

...

# Rules
```

PHONY targets are also declared as such right above the target: I believe we have already quite a few targets which makes it hard to follow hence having the PHONY target right above the declaration is both less error prone and easier to check IMO. IIRC it was a suggestion from @sanmai and I must say I never had to look back at it ever since I've adopted that practice

Makefile
```diff
@@ -1,137 +1,205 @@
...
+
 
-# URLs to download all tools
-BOX_URL="https://github.com/humbug/box/releases/download/3.1.0/box.phar"
-PHP-CS-FIXER_URL="https://cs.sensiolabs.org/download/php-cs-fixer-v2.phar"
+#
+# Variables
+#---------------------------------------------------------------------------
+BOX=./.tools/box
+BOX_URL="https://github.com/humbug/box/releases/download/3.8.0/box.phar"
```
theofidry: upgrade Box to 3.8 on the way... It was a really minor change so I don't think it pollutes that PR too much :)

Makefile
```diff
@@ -1,137 +1,205 @@
...
+compile:	## Bundles Infection into a PHAR
+compile: $(INFECTION)
 
-./phpstan.phar:
-	wget $(PHPSTAN_URL)
-	chmod a+x ./phpstan.phar
+.PHONY: cs-fix
+cs-fix:	  	## Runs PHP-CS-Fixer
+cs-fix: $(PHP_CS_FIXER)
+	# TODO: disable verbose mode?
```
theofidry: Should the verbose mode be disabled for PHP-CS-Fixer? What's the point of having it if it's run only locally? IMO if you want extra verbosity you can run the command manually instead

Makefile
```diff
@@ -1,137 +1,205 @@
-PHP_CS_FIXER_FUTURE_MODE=1
-PHPSTAN=./phpstan.phar
-PHP-CS-FIXER=./php-cs-fixer-v2.phar
-PHPUNIT=vendor/bin/phpunit
-INFECTION=build/bin/infection.phar
+.DEFAULT_GOAL := help
+
+.PHONY: help
+# TODO: add a check to ensure the result is always the expected one
```
theofidry: For this kind of TODO I think it would be best to add a PHPUnit test. I did that for example with PHP-Scoper and it works wonders, see [MakefileE2ETest.php](https://github.com/humbug/php-scoper/blob/188ff49/tests/MakefileE2ETest.php)

Makefile
```diff
@@ -1,137 +1,205 @@
...
+.PHONY: compile
+compile:	## Bundles Infection into a PHAR
+compile: $(INFECTION)
 
-./phpstan.phar:
-	wget $(PHPSTAN_URL)
-	chmod a+x ./phpstan.phar
+.PHONY: cs-fix
+cs-fix:	  	## Runs PHP-CS-Fixer
+cs-fix: $(PHP_CS_FIXER)
```
theofidry: I quite didn't get when you are expected to run `cs-fix`. I get the use of `cs-check` for the CI but is there a case where you prefer `cs-check` over `cs-fix` locally?

Makefile
```diff
@@ -1,137 +1,205 @@
...
-	$(DOCKER_RUN_73) $(PHPUNIT)
+.PHONY: analyze-ci
+analyze-ci:	## Runs static analyzers and various other checks
+analyze-ci: phpstan validate
 
-.PHONY: test-infection-phpdbg test-infection-phpdbg-72 test-infection-phpdbg-73
-#infection with phpdbg
-test-infection-phpdbg: test-infection-phpdbg-72 test-infection-phpdbg-73
+.PHONY: auto-review
+auto-review:	## TODO: find a suitable description
```
theofidry: TODO: I quite didn't get what this command was about neither when it's supposed to be run

Makefile
```diff
@@ -1,137 +1,205 @@
...
+.PHONY: validate
+validate:	## Checks that the composer.json file is valid
+validate:
+	composer validate --strict
 
-test-infection-phpdbg-73: build-xdebug-73
-	$(DOCKER_RUN_73) phpdbg -qrr bin/infection --threads=4
+.PHONY: test
+test:		## Runs all the tests
+# TODO: add a test to ensure we are not missing targets here
```
theofidry: Noteworthy TODO

Makefile
```diff
@@ -1,137 +1,205 @@
...
+test: test-unit test-e2e-phpdbg test-e2e-xdebug test-infection-phpdbg test-infection-xdebug
+
+.PHONY: test-unit
+test-unit:	## Runs the unit tests
```
theofidry: A suggestion here. As of now you will see (in the make help command) the:

```
test-unit: ...
test-unit-72: ...
test-unit-73: ...
```

same for all others. What do you think instead of "hidding" (you just need to remove the command line) the `7*` ones? As a result you would only show:

```
test-unit: ...
test-e2e: ...
test-infection: ...
```

The commands don't disappear, but they are less highlighted because more of a detail that you will figure out when running the above commands if necessary. I.e. what you care about is `test-unit`, if `test-unit-72` in peculiar fails and you want to run only this one, you will discover that command when the failure occur and will be free to run this more specific command if desired

Makefile
```diff
@@ -1,137 +1,205 @@
...
+test-infection-phpdbg: test-infection-phpdbg-72 test-infection-phpdbg-73
 
-.PHONY: test-e2e-xdebug test-e2e-xdebug-72 test-e2e-xdebug-73
-#e2e tests with xdebug
-test-e2e-xdebug: test-e2e-xdebug-72 test-e2e-xdebug-73
+.PHONY: test-infection-phpdbg-72
+test-infection-phpdbg-72:	## Runs Infection with PHPDBG against itself in PHP 7.2
+test-infection-phpdbg-72: $(DOCKER_RUN_72_IMAGE)
+	# TODO: calculate the number of thread here instead of using the arbitrary number 4
+	# TODO: check other occurrences
```
theofidry: noteworthy TODO

Makefile
```diff
@@ -1,137 +1,205 @@
...
-.PHONY: analyze cs-fix cs-check phpstan validate auto-review
-analyze: cs-check phpstan validate
+#
+# Rules from files (non-phony targets)
+#---------------------------------------------------------------------------
 
-# PHP-CS-Fixer is checked by PrettyCI
-.PHONY: analyze-ci
-analyze-ci: phpstan validate
+# TODO: some tools such as Box & PHP-CS-Fixer could be handled in a better way to ensure we are using the last version.
```
theofidry: Noteworthy TODO. This as of now is _already_ an issue at the very least locally. Indeed if you run:

```
make compile
```

it will download Box, but if you update the Box URL, the existing one won't be invalidated... This is however a problem that should be tackled in a dedicated PR IMO

Makefile
```diff
@@ -1,137 +1,205 @@
...
-	# All tests finished without errors
+.PHONY: cs-check
+cs-check:	## Runs PHP-CS-Fixer in dry mode
+cs-check: $(PHP_CS_FIXER)
+	# TODO: disable verbose mode?
+	$(PHP_CS_FIXER) fix -v --cache-file=$(PHP_CS_FIXER_CACHE) --dry-run --stop-on-violation
 
-.PHONY: test-unit test-unit-72 test-unit-73
-#php unit tests
-test-unit: test-unit-72 test-unit-73
```
theofidry: I see 7.2 and 7.3 but what about 7.1?

Makefile
BackEndTea: I have thought about this issue before, for a future PR: how about moving all phars into a folder (maybe `/build`?), and having a `make clean` command which deletes all files in the folder.

Makefile
BackEndTea: We dropped PHP 7.1 support for `0.14` in #702

Makefile
BackEndTea: Originally the auto-review tests weren't part of the normal phpunit run, but that got changed.

I'm thinking we can safely remove this make command. 

And on another note, i think we may be able to replace some of the auto-review tests with PHPStan rules.

theofidry: :+1: will remove it then

Makefile
theofidry: I used `.tools` in this PR but could be `build` as well it's not an issue. I liked to use `.tools` since it makes it cacheable...

Maybe an alternative would be combine the PHAR with a cache key with the cache key dependent on the corresponding URL: if the URL changes so does the cache key hence the PHAR needs to be downloaded again

Makefile
theofidry: 👌

Makefile
BackEndTea: `.tools` works as well, i just use `build` in most of my personal projects, and didn't notice `the `.tools` in this PR 😅 

> Maybe an alternative would be combine the PHAR with a cache key with the cache key dependent on the corresponding URL: if the URL changes so does the cache key hence the PHAR needs to be downloaded again

That would be a solution i think.

theofidry: For another PR though as I really didn't want to think of the details tonight :)

.gitignore
sanmai: Shouldn't it've been in `$HOME`?

.travis.yml
```diff
@@ -12,6 +12,7 @@ cache:
   directories:
     - $HOME/.composer/cache
     - build/cache
+    - .tools
```
sanmai: What if we ever want to update the tools?

Makefile
sanmai: It depends on the situation with files with syntax errors. I had some problem with PHP-CS-Fixer silently ignoring such files in the default non-verbose mode, but this may be the thing of past.

(I prefer to run PHP-CS-Fixer in the verbose mode at all times because it may go great lengths looking over things do not need fixing if you happen to, say, make a very unfortunate symlink somewhere else. At least you will see it being stuck. It is not like Infection needs any of the symlinks, just saying.)

Makefile
sanmai: More threads do not necessarily make things substantially better. My workstation has tens of threads, and I experimented to see if there's any benefit in running Infection over Infection using all available cores, and found out that anything more than 8 resulted in only marginal improvements, in my case.

Makefile
sanmai: I think there's a way to make dynamic rules (like, e.g. with `$(PHP_CS_FIXER)` rule below), so *theoretically* all these rules like `test-unit-72` and `test-unit-73` can be compacted into one. But, well, this will make things less explicit, and, not necessarily easier to understand, so I didn't think into this approach too far.

.gitignore
theofidry: yeah... I have no idea why really

Makefile
theofidry: ok

Makefile
theofidry: Yeah, I'm fine with having it like that and try not too be too smart about since it will make the Makefile less accessible for not much.

The point above though was more about the displayed command when running `make` rather than the actual targets.

Makefile
theofidry: So you think it should stay at 4 instead of taking advantage of all the detected cores?

.travis.yml
theofidry: Discussed in [this thread](https://github.com/infection/infection/pull/706#discussion_r294505032). It might be wiser to disable the caching of that directory meanwhile though so that the problem doesn't propagate to the CI yet

Makefile
sanmai: I think four is fine, unless you want to look into this. Also it is a question how many processes Infection uses for a single mutant. I recall it being two for some valid reason, but can't look into the code right now. If so you'd have to divide number of your cores by two, and there's IPC which also can be a limiting factor.

.travis.yml
BackEndTea: IIRC the cache doesn't persist between PR's. 
And maybe we can make the cache 'depend' on the Makefile? like appveyor allows you to purge the cache if a specific file changes

sanmai: We can actually pull this off.

theofidry: I think if we employee the cache mechanism used in the other thread we are good, but Travis allows you to clear the cache if desired so.

The PR cache is not triggered until merged. Once merged, the following PRs will benefit from it (like for the Composer cache)

Makefile
theofidry: I think this could be simplified to use `cs` and fix the CS when expected: there is already a dedicated CS check on a PR basis I don't think it provides anything to check it locally as well...

thoughts @infection/core?

Makefile
theofidry: Let's leave it at that for now then...

theofidry:
New output:

Screen Shot 2019-06-18 at 21 58 39

theofidry:
@infection/core thoughts? Or you don't fancy much the changes? Note that I actually didn't change any target in the end except adding the help one and making the other defaults, precisely to avoid changing too many things for everyone :)

.gitignore
sanmai:
```suggestion
```

Makefile
sanmai: I'd love to use `cs` instead, because it just happens that's how this rule usually named elsewhere.

maks-rafalko:
Thank you @theofidry

Issue #705, opened 2019-06-17
Infection fails with "Allowed memory size ... exhausted" for a project with a really big coverage


maks-rafalko:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.13.2
| Test Framework version | PHPUnit
| PHP version | 7.2
| Github Repo | https://github.com/vimeo/psalm 

Extracted from https://twitter.com/sorsoup/status/1140021198765658118

> Infection seems to be crashing when I try to run it to check the test suite of Psalm - output at (link: https://circleci.com/gh/bdsl/psalm/232) circleci.com/gh/bdsl/psalm/…
> 
> Don't know if I'm doing anything wrong.

Command line:

```bash
php -d memory_limit=12G ~/sites/infection/bin/infection --coverage=build/phpunit --only-covered --threads=4
```

Output:

```bash
     ____      ____          __  _
    /  _/___  / __/__  _____/ /_(_)___  ____
    / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
  _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
 /___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/

Running initial test suite...

PHPUnit version: 7.5.12

    1 [============================] < 1 sec

Generate mutants...

Processing source code files:   0/543

PHP Fatal error:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 20480 bytes) in /infection/src/TestFramework/PhpUnit/Coverage/CoverageXmlParser.php on line 170
```

So it fails with when the coverage data is being accumulated in the following lines of the code (depending on how much `memory_limit` is set

* on [`src/TestFramework/PhpUnit/Coverage/CoverageXmlParser.php:170`](https://github.com/infection/infection/blob/b9ebc49f2b5fd923b848a2ea035f90577d02bc5c/src/TestFramework/PhpUnit/Coverage/CoverageXmlParser.php#L170)
* on [`src/TestFramework/Coverage/CodeCoverageData.php:220`](https://github.com/infection/infection/blob/b9ebc49f2b5fd923b848a2ea035f90577d02bc5c/src/TestFramework/Coverage/CodeCoverageData.php#L220)

[...]

PR #708, opened 2019-06-18
Reduce memory usage for Coverage Data

maks-rafalko:
This PR:

- [x] Reduces memory usage when Coverage Data is being collected
- [x] Covered by tests

Partially fixes issues from https://github.com/infection/infection/issues/705. At least I now able to run Infection for Psalm locally with `memory_limit=12G` :D


sanmai:
PHP is very liberal with copying a whole array if you happen to change a single value, or if you wish to pass it around from function to function.

One way is to use references, but that's very troublesome and requires a great care. Another way is to wrap a big array with an object, and access this object's property instead.

src/TestFramework/Coverage/CodeCoverageData.php
```diff
@@ -200,23 +200,21 @@ private function getCoverage(): array
...
+            $this->addTestExecutionInfo($coverage);
 
             $this->coverage = $coverage;
         }
 
         return $this->coverage;
     }
 
-    private function addTestExecutionInfo(array $coverage): array
+    private function addTestExecutionInfo(array &$coverage): void
```
theofidry: Might be good to add a comment here to explain why we passed it by reference

maks-rafalko:
> Another way is to wrap a big array with an object, and access this object's property instead.

done in https://github.com/infection/infection/pull/710

Thank you @theofidry & @sanmai

PR #710, opened 2019-06-20
Dramatically reduce memory usage by using classes instead of object-like arrays

maks-rafalko:
All proofs are here: https://steemit.com/php/@crell/php-use-associative-arrays-basically-never

* Objects of DTO classes takes less memory in PHP and as performant as arrays (or even better).
* Moreover, instead of `array` typehints we now have a real classes or typed collections, which is a big win IMO.
* No more passed variable by reference (`&$coverage`)

This PR follows https://github.com/infection/infection/pull/708 as @sanmai requested, I just extracted it to another PR for easier reviews.

Partially fixes issue https://github.com/infection/infection/issues/705

---

#### What is the memory benefit?

1. Originally, I wasn't able to run Infection for Psalm with `memory_limit=12G`
2. with #708, I was able to run Infection for Psalm with `memory_limit=12G`
3. With this PR, I'm able to run Infection for Psalm with `memory_limit=5G`

So it is `7Gb` or `58%` less memory usage for Psalm case.

### DIFF
--- a/src/TestFramework/Coverage/CoverageFileData.php
+++ b/src/TestFramework/Coverage/CoverageFileData.php
@@ -38,31 +38,25 @@ namespace Infection\TestFramework\Coverage;
 /**
  * @internal
  */
-final class CachedTestFileDataProvider implements TestFileDataProvider
+final class CoverageFileData
 {
     /**
-     * @var TestFileDataProvider
+     * @var array<int, array<int, CoverageLineData>>
      */
-    private $testFileDataProvider;
+    public $byLine = [];
 
     /**
-     * @var array<string, TestFileTimeData>
+     * @var CoverageMethodData[]
      */
-    private $testFileInfoCache = [];
+    public $byMethod = [];
 
-    public function __construct(TestFileDataProvider $testFileDataProvider)
+    public static function from(array $byLine, array $byMethod): self
     {
-        $this->testFileDataProvider = $testFileDataProvider;
-    }
+        $self = new self();
 
-    public function getTestFileInfo(string $fullyQualifiedClassName): TestFileTimeData
-    {
-        if (\array_key_exists($fullyQualifiedClassName, $this->testFileInfoCache)) {
-            return $this->testFileInfoCache[$fullyQualifiedClassName];
-        }
+        $self->byLine = $byLine;
+        $self->byMethod = $byMethod;
 
-        return $this->testFileInfoCache[$fullyQualifiedClassName] = $this->testFileDataProvider->getTestFileInfo(
-            $fullyQualifiedClassName
-        );
+        return $self;
     }
 }
--- a/src/TestFramework/Coverage/CoverageFileData.php
+++ b/src/TestFramework/Coverage/CoverageFileData.php
@@ -50,13 +50,13 @@ final class CoverageFileData
      */
     public $byMethod = [];
 
-    public static function from(array $byLine, array $byMethod): self
+    /**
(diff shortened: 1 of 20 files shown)


src/TestFramework/Coverage/CoverageLineData.php
```diff
@@ -0,0 +1,77 @@
...
+    public static function withTestMethod(string $testMethod): self
+    {
+        $self = new self();
+
+        $self->testMethod = $testMethod;
+
+        return $self;
+    }
+
+    public static function with(string $testMethod, string $testFilePath, float $time): self
```
BackEndTea: From what i can see this method isn't used outside of the tests. Meaning we can probably remove this, as well as the properties that it sets.

maks-rafalko: properties are used here https://github.com/infection/infection/pull/710/files#diff-d0cc0b68f2e397d7080d2f7c58834ce3R236, we can't remove them.

Test method still used in tests only though

bdsl:
I had a go at using this in CircleCI to test Psalm. Installed from commit ec775c9 via composer. Unfortunately I'm still getting the same sort of failures.

Strangely it took less time to fail when I increased the memory limit, but I suspect that may just be random variation:

https://circleci.com/gh/bdsl/psalm/494
https://circleci.com/gh/bdsl/psalm/495

theofidry:
I would be interesting to run a blackfire profile to see the memory difference. That said, since the difference is still easy to see even without profiling I'd say it's more out of curiosity than a necessity

theofidry:
@bdsl a good profiling could help out, but it's also very likely that the coverage data is just horribly big. We can work on making it better, but ultimately you do need more resources or alternatively leverage incremental mutations, i.e. filter out the unchanged files, instead of running Infection against the whole codebase all the time

gmponos:
Does this improve speed as well?

sanmai:
It may or may not, but this is not the place where Infection spend most of the time. So that the overall difference in run times excepted to negligible.

src/TestFramework/Coverage/CoverageFileData.php
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Infection\TestFramework\Coverage;
+
+/**
+ * @internal
+ */
+final class CoverageFileData
```
sidz: Do you mind to define default constructor as private as we use named constructor here?

p.s. the same for all DTO's

src/TestFramework/Coverage/CoverageMethodData.php
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Infection\TestFramework\Coverage;
+
+/**
+ * @internal
+ */
+final class CoverageMethodData
```
sidz: Is it ok that we may initialize an empty object through the default constructor?

suggestions if no:
- define default constructor as private
- use default constructor instead of named one.

src/TestFramework/Coverage/TestFileTimeData.php
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Infection\TestFramework\Coverage;
+
+/**
+ * @internal
+ */
+final class TestFileTimeData
```
sidz: the same rules as for `CoverageMethodData`

maks-rafalko:
@sidz done

theofidry:
Cool one :)

PR #713, opened 2019-06-23
Upgrade to PHPUnit 8.2.3

maks-rafalko:
Since we now require PHP 7.2+, it's time to use PHPUnit 8.

Just a first step - upgrading test framework. 

Next step: replace XDebug with pcov (but not in this PR).


maks-rafalko:
Thank you @sanmai & @sidz

Issue #580, opened 2018-11-26
Infection config generator fails to handle multiple ignored directories


sanmai:
| Question    | Answer
| ------------| ---------------
| Infection version | dccbf0181 (current master)
| Test Framework version | PHPUnit 5.7.27
| PHP version | PHP 7.2.9-1
| Platform    | Linux
| Github Repo | https://github.com/pear/OLE

When excluding multiple directories, Infection generates a broken configuration which fails to pass Infection's own checks.

```
Any directories to exclude from within your source directories?: build
Any directories to exclude from within your source directories?: vendor
Any directories to exclude from within your source directories?: tests
Any directories to exclude from within your source directories?: coverage
Any directories to exclude from within your source directories?: 
```

Gets you:
```
"excludes": {
    "0": "vendor",
    "1": "tests",
    "2": "build",
    "5": "coverage"
}
```

#576 and #577 are related.

PR #714, opened 2019-06-24
Fix Infection config builder: make sure it always creates an array for exluded dirs but not object

maks-rafalko:
Fixes https://github.com/infection/infection/issues/580

`array_unique([['a', 'a', 'b'])` returns `[0 => 'a', 2 => 'b']` which in its turn breaks `infection.json` because now we have an object

```json
{
    "excludeDirs": {
        "0": "a",
        "2": "b",
    }
}
```

instead of array

```json
{
    "excludeDirs": [
        "a",
        "b",
    ]
}
```

### DIFF
--- a/src/Config/ValueProvider/ExcludeDirsProvider.php
+++ b/src/Config/ValueProvider/ExcludeDirsProvider.php
@@ -120,7 +120,7 @@ final class ExcludeDirsProvider
             }
         }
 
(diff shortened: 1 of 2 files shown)


maks-rafalko:
> Maybe this could target 0.13? as it is bugfix and not a new feature

retargeted

Thank you @BackEndTea, @sidz and @localheinz

PR #718, opened 2019-06-26
 Replace service names by their FQCN

theofidry:
As discussed in the organisation chat, here's a PR to switch from `dot.lowercase.service.name` to the class FQCN when possible. Whilst it makes the service naming a bit inconsistent (not all services have a class, e.g. `'coverage.path'`), it makes the services easier to discover and to browser. For example:

```php
$container->get('foo'); // Have to look at how `'foo'` is registered to the container

// Whereas

$container->get(Foo::class); // You know the returned value is a Foo instance and you can check the class definition right away
```

Bonus: I added a phpstorm metadata file which should help out in most cases


theofidry:
Arg, works fine locally, what is going on :(

.gitignore
```diff
@@ -1,18 +1,19 @@
 /.idea
 /.composer/
-/vendor
+/vendor/
```
localheinz: 👍 

What do you think about also sorting entries here

* by type (directories first, then files) first
* alphabetically second

?

localheinz:
Not directly related to this PR, but maybe it makes even sense to write the container by hand?

See for example https://matthiasnoback.nl/2019/03/hand-written-service-containers/.

.gitignore
theofidry: 👍 I'll do that in a separate PR though

theofidry:
@localheinz I think that's a good idea if the rest of the team is up for it. I would argue however that it's for another PR :)

sanmai:
> but maybe it makes even sense to write the container by hand?

It it'll be like this:
```php
public function getInfectionConfig(): InfectionConfig
{
    return $container->get(InfectionConfig::class);
}
```

I'm all onboard.

theofidry:
I'm not sure what is going wrong here... But let's progressively lean towards @localheinz solution. If someone wants to pick up that PR feel free to do so. For my part I would prefer to focus on moving forward :)

PR #719, opened 2019-06-26
Simplify the CS and Analyze Makefile commands

theofidry:
This PR:

- Removes `make cs-fix` and `make cs-check` in favour of `make cs`
- Removes `make analyze-ci` in favour of the already existent `make analyze`

This should make the commands a bit easier to use (less commands to use/understand) and also reduce the friction of environment difference between local dev and CI which is always a source of pain.

I took the opportunity to slightly change the outdated contributing docs as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,18 +8,27 @@
 [![Slack channel: #infection on the Symfony slack](https://img.shields.io/badge/slack-%23infection-green.svg?style=flat-square)](https://symfony.com/slack-invite)
 
 
-Infection - Mutation Testing framework
-=========
+# Infection - Mutation Testing framework
 
-Please read documentation here: http://infection.github.io
+Please read documentation here: [infection.github.io][doc]
 
-Twitter: http://twitter.com/infection_php
+Twitter: [@infection_php][twitter]
+
 
 ### Contributing
 
-Infection is an open source project that welcomes pull requests and issues from anyone.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -8,18 +8,27 @@
...
+
 
 ### Credits
 
-This project is highly inspired from Pádraic Brady ([@padraic](https://github.com/padraic))'s [Humbug library](https://github.com/humbug/humbug). Humbug has since then been discountinued in favour of this project.
+This project is highly inspired from Pádraic Brady ([@padraic][padraic])'s [Humbug library][humbug]. 
+Humbug has since then been discontinued in favour of this project.
+
+
+[doc]: http://infection.github.io
```
maks-rafalko: what is the benefit of using such links?

theofidry: - makes the links re-usable
- easier to read the markdown (non rendered) content as the `[doc]` or other tends to be more readable than the link
- it makes it easier to check all the links of the file if you group them all at the bottom like this

So it's arguably not the best usage of it here, I did it more out of habit than necessity

maks-rafalko: Why I'm asking is because it's possible now to have dead links, isn't it? Which are declared at the bottom but not used.

I'm fine with the current syntax ;)

theofidry: Yes... There is a couple of things I would like to add to Markdown files:

- that all links are used
- that all links are valid

However this is quite hard to achieve and since we have so little Markdown files I don't think it's worth it to spend too much time on it.

PR #720, opened 2019-06-26
Update the license

theofidry:
As per the discussion with @sanmai:

- Update the license your to the start year `2017` to avoid to have to update the end year all the time
- Update manually the license header in `bin/infection` and `bin/infection-debug` as PHP-CS-Fixer seemed unable to handle those two files for some reasons
- Apply the CS fixer

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,18 +1,29 @@
-<?php
+<?php declare(strict_types=1);
 
-$header = trim('This code is licensed under the BSD 3-Clause License.'.substr(file_get_contents('LICENSE'), strlen('BSD 3-Clause License')));
+use PhpCsFixer\Config;
+use PhpCsFixer\Finder;
 
+$header = trim(
+    sprintf(
+        'This code is licensed under the BSD 3-Clause License.%s',
+        substr(
+            file_get_contents('LICENSE'),
+            strlen('BSD 3-Clause License')
+        )
+    )
+);
 
-$finder = \PhpCsFixer\Finder::create()
+$finder = Finder::create()
     ->in('src')
     ->in('tests')
-    ->exclude('Fixtures')
     ->in('bin')
+    ->exclude('Fixtures')
     ->name('*php')
     ->name('infection')
+    ->name('infection-debug')
 ;
 
-return \PhpCsFixer\Config::create()
+return Config::create()
     ->setRiskyAllowed(true)
     ->setRules([
         '@Symfony' => true,
@@ -46,7 +57,7 @@ return \PhpCsFixer\Config::create()
             'commentType' => 'PHPDoc',
             'header' => $header,
             'location' => 'after_open',
-            'separate' => 'bottom',
+            'separate' => 'both',
         ],
         'native_function_invocation' => [
             'include' => ['@compiler_optimized']
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,5 +1,4 @@
 <?php
-
 /**
  * This code is licensed under the BSD 3-Clause License.
  *
@@ -100,7 +99,7 @@ return Config::create()
             'commentType' => 'PHPDoc',
             'header' => $header,
             'location' => 'after_open',
-            'separate' => 'both',
+            'separate' => 'bottom',
         ],
         'native_function_invocation' => [
             'include' => ['@compiler_optimized'],
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,5 +1,37 @@
 <?php
 
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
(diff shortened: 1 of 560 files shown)


sanmai:
> Fatal error: strict_types declaration must be the very first statement in the script in /home/travis/build/infection/infection/bin/infection on line 36

theofidry:
@sanmai I'm really confused :( the code looks correct to me and works locally

.php_cs.dist
```diff
@@ -1,18 +1,29 @@
-<?php
+<?php declare(strict_types=1);
```
sidz: could we please be consistent with the place where we add `declare(strict_types=1);`? Usually it's the third line from top.

.php_cs.dist
```diff
@@ -1,18 +1,31 @@
...
+    sprintf(
+        'This code is licensed under the BSD 3-Clause License.%s',
+        substr(
+            file_get_contents('LICENSE'),
+            strlen('BSD 3-Clause License')
+        )
+    )
+);
+
+$finder = Finder::create()
```
localheinz: What do you think about 

```diff
diff --git a/.php_cs.dist b/.php_cs.dist
index 436c2027..21e60be3 100644
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -16,11 +16,19 @@ $header = trim(
 );

 $finder = Finder::create()
-    ->in('src')
-    ->in('tests')
-    ->in('bin')
-    ->exclude('Fixtures')
+    ->ignoreDotFiles(false)
+    ->in(__DIR__)
+    ->exclude([
+        '.ci',
+        '.composer',
+        '.github',
+        'build',
+        'devTools',
+        'resources',
+        'tests/Fixtures',
+    ])
     ->name('*php')
+    ->name('.php_cs.dist')
     ->name('infection')
     ->name('infection-debug')
```

?

Then this file here will be automatically fixed as well!

theofidry: Done :)

.php_cs.dist
```diff
@@ -46,13 +100,13 @@ return \PhpCsFixer\Config::create()
             'commentType' => 'PHPDoc',
             'header' => $header,
             'location' => 'after_open',
-            'separate' => 'bottom',
+            'separate' => 'both',
```
sanmai: I wonder if this extra line above the header comment is really necessary.

theofidry: It's not necessary per se but I think it's nicer & more consistent: for pretty much every block we have a blank line on both sides:

```php
<php

/**
 * ...
 */

declare(strict_types=1);

namespace Acme;
```

Instead of:

```php
<php
/**
 * ...
 */

declare(strict_types=1);

namespace Acme;
```

Although tbh my personal preference would be:

```php
<php declare(strict_types=1);

/*
 * ...
 */

namespace Acme;
```

sanmai: The thing is that without this extra line header is very tidy:

Screenshot from 2019-06-27 16-51-58

> `<php declare(strict_types=1);`
This totally breaks header comment folding in Eclipse.

theofidry: I find it as tidy tbh :) Also worth mentioning it's actually PHP-CS-Fixer default. But anyway since it's such a source of debate let's revert that.

>This totally breaks header comment folding in Eclipse.

 I'm not surprised: I found Eclipse to be really lacking and lagging behind when it comes to PHP support (worth mentioning this declaration is valid even in 5.x)... Which is why I switched over to PhpStorm which is still at a very reasonable price for the time I spent on it. It is nonetheless a bug you may want to report

theofidry:
Still not sure what exactly is causing the PR to fail

.php_cs.dist
```diff
@@ -1,18 +1,71 @@
...
+);
+
+$finder = Finder::create()
+    ->in(__DIR__)
+    ->exclude([
+        '.ci',
+        '.composer',
+        '.github',
+        '.tools',
+        'build',
```
sanmai: This exclude option operates on path basis. For example, if we happen to have another directory named `build` anywhere in the code, it will also be ignored by the fixer. It is not like this should stop us, but that's the thing that may cause serious head scratching. I do not know if there's a workaround or anything better.

theofidry: Yeah... I got bitten by that quite a bit within Box which is relying a lot on `Finder` :( I think it should still be ok here at least for now

theofidry:
>Fatal error: strict_types declaration must be the very first statement in the script in /home/travis/build/infection/infection/bin/infection on line 36

I still really don't get what this is about and works fine locally...

sanmai:
```bash
$ chmod +x ./tests/Fixtures/e2e/Empty_Path/bootstrap
$ ./tests/Fixtures/e2e/Empty_Path/bootstrap
PHP Fatal error:  strict_types declaration must be the very first statement in the script in bin/infection on line 10
PHP Stack trace:
PHP   1. {main}() tests/Fixtures/e2e/Empty_Path/bootstrap:0
```

Is it like we have to add `declare(strict_types=1);` to that file too?

theofidry:
Arg I see, it's because with have `include 'path/to/infection';` in which case having a shebang line is an issue with the declare strict type

sanmai:
This `bootstrap` should work:
```
#!/bin/bash
cd $(dirname "$0")
PHP=$(command -v php)

export PATH=
$PHP ../../../../bin/infection
```

theofidry:
Thanks! Let's see how that goes

sanmai:
Well, it turns out we may not need that bootstrap file at all. 
Here's updated `tests/Fixtures/e2e/Empty_Path/run_tests.bash`:

```bash
#!/usr/bin/env bash

set -eo pipefail

readonly INFECTION=../../../../${1}

if [ "$PHPDBG" = "1" ]
then
    PHPDBG=$(which phpdbg)
    PATH= $PHPDBG -qrr $INFECTION
else
    PHP=$(which php)
    PATH= $PHP $INFECTION
fi

diff expected-output.txt infection.log
```

Can you tell if E2E tests are passing locally with this file?

theofidry:
@sanmai I allowed edit for you: I cannot check it atm and the end-to-end tests are... weird locally. Last time I tried it literally took over 30min and on-going

theofidry:
@sanmai I'm at a loss...

```bash
#!/usr/bin/env bash
# tests/Fixtures/e2e/Empty_Path/run_tests.bash

set -e pipefail

cd $(dirname "$0");

readonly INFECTION="./../../../../bin/infection";
readonly PHP=$(command -v php);

export PATH=;

if [ "$PHPDBG" = "1" ]
then
    $(which phpdbg) -qrr ${INFECTION};
else
    ${PHP} ${INFECTION};
fi

cat expected-output.txt
cat infection.log

diff expected-output.txt infection.log

```
$ cd tests/Fixtures/e2e/Empty_Path
$ bash -x ./run_tests.bash; echo $?
+ set -e pipefail
++ dirname ./run_tests.bash
+ cd .
+ readonly INFECTION=./../../../../bin/infection
+ INFECTION=./../../../../bin/infection
++ command -v php
+ readonly PHP=/path/to/bin/php
+ PHP=/path/to/bin/php
+ export PATH=
+ PATH=
+ '[' '' = 1 ']'
+ /path/to/php ./../../../../bin/infection
You are running Infection with Xdebug enabled.
     ____      ____          __  _

Running initial test suite...

PHPUnit version: 6.5.14

    8 [============================]  1 secProcessing source code files: 0/1

Generate mutants...

Processing source code files: 1/1
Creating mutated files and processes: 1/1

.                                                    (1 / 1)

1 mutations were generated:
       1 mutants were killed
       0 errors were encountered
       0 time outs were encountered

Metrics:
         Mutation Score Indicator (MSI): 100%
         Mutation Code Coverage: 100%
         Covered Code MSI: 100%

Time: 0s. Memory: 10.00MB
+ cat expected-output.txt
./run_tests.bash: line 19: cat: No such file or directory
1
```

Just what the heck is going on: how can the `expected-output.txt` not exist when it clearly does?!

sanmai:
> cat: No such file or directory
We might be testing an impossible proposition here, because there isn't a real system that does not have `$PATH` defined in any way. Without `PATH` nothing works, just like we observing. Even simple bash scripts cannot work, because, say, `cat` is a command located at `/bin/cat`, and without `PATH` the shell just cannot find it.

How do you think if we step back from this problem, revert what is already done with `e2e/Empty_Path`, and also keep `bin/infection` strict-types free for time being?

Alternatively, we can nuke that e2e test altogether, but still it may be better to do it in a separate PR while adding a different test for the problem from #295.

theofidry:
>I'm mildly surprised to find that it was me who filed #295. Well, now I feel real bad that we have to get on with this. As I started all this in #295, it should be probably me to deal with e2e/Empty_Path.

Haha I guess that happens :) In what context did you have `PATH` not set? I never heard of that case

PR #723, opened 2019-06-27
Add tests for the Makefile

### LINKED ISSUE #716: Add tests for Makefile
As per:

- https://github.com/infection/infection/pull/706/files#r294502276
- https://github.com/infection/infection/pull/706/files#r294503142

Some targets in the Makefile requires an exhaustive list of targets. As the project grows it's very easy to forget one of those.

Example of such test: https://github.com/humbug/php-scoper/blob/188ff496423e90984db90c9be43d87df080bbc9f/tests/MakefileE2ETest.php

theofidry:
Closes #716.

The test should be self explanatory. What is worth mentioning however is that unlike in Box or PHP-Scoper where I went for an extremely simple regex approach, here I introduced a parser to be able to get some additional checks. It may be a bit overkill in itself, but I had in mind extracting it in a separate package to re-use it in Box & PHP-Scoper. So the Parser class itself and its test may be considered temporary :)


.gitignore
```diff
@@ -16,3 +16,6 @@
 /build/bin/infection.phar.pubkey
 .phpunit.result.cache
 /devTools/Docker*.json
+/tests/Fixtures/e2e/*/infection/
+/tests/Fixtures/e2e/*/coverage/
+/tests/Fixtures/e2e/Configure/infection.json.dist
```
theofidry: I'm not sure why this was missing: did I run a wrong command or something?

tests/Config/ValueProvider/AbstractBaseProviderTest.php
```diff
@@ -70,7 +70,7 @@ protected function createStreamableInputInterfaceMock($stream = null, $interacti
     {
         $mock = $this->createMock(StreamableInputInterface::class);
         $mock->method('isInteractive')
-            ->will($this->returnValue($interactive));
+            ->willReturn($interactive);
```
theofidry: I did not touch this so it's most likely the CS fixer...

tests/Makefile/MakefileTest.php
```diff
@@ -0,0 +1,303 @@
...
+use PHPUnit\Framework\TestCase;
+use function shell_exec;
+use function sprintf;
+use function strpos;
+use function substr_count;
+
+/**
+ * @coversNothing
+ *
+ * @group infra
```
theofidry: I added (back IIRC?) a test group `infra`. For now it doesn't make any difference for the tests but I believe it makes sense to have a test group that is more about organisation & conventions than actual source code test

tests/Makefile/MakefileTest.php
```diff
@@ -0,0 +1,303 @@
...
+[33mtest-infection:[0m  Runs Infection against itself
+
+EOF;
+
+        $this->assertSame($expectedOutput, $output);
+    }
+
+    public function test_the_makefile_can_be_parsed(): void
+    {
+        (new Parser())->parse(file_get_contents(self::MAKEFILE_PATH));
```
theofidry: Since the parser is stateless unlike what I originally thought it was gonna be, I think `Parser::parse()` can be made static

tests/Makefile/MakefileTest.php
```diff
@@ -0,0 +1,303 @@
...
+    protected function tearDown(): void
+    {
+        chdir(self::$originalCwd);
+    }
+
+    public function test_the_default_goal_is_the_help_command(): void
+    {
+        $output = shell_exec(sprintf('make -f %s', self::MAKEFILE_PATH));
+
+        $expectedOutput = <<<'EOF'
```
theofidry: I did not find an easy way to have one test with the ANSI characters like here and another one without... If someone has an idea I'm all ears. Otherwise we can just keep carry one it's not a big deal

Makefile
```diff
@@ -66,8 +63,7 @@ validate:
 
 .PHONY: test
 test:		## Runs all the tests
-# TODO: add a test to ensure we are not missing targets here
-test: test-unit test- test-infection
```
theofidry: Not sure when `test-e2e` has been changed to `test-` but in any case, at least now the MakefileTest should detect this kind of error :)

tests/Fixtures/e2e/Code_Coverage_Ignore/infection/coverage-xml/IgnoreClass.php.xml
```diff
@@ -1,83 +0,0 @@
-<?xml version="1.0"?>
```
theofidry: Looks like those files have been introduced by mistake in #720. In any case the updated `.gitignore` should deal with this and I double checked that I removed all of them

tests/AutoReview/MutatorTest.php
```diff
@@ -44,6 +44,8 @@
  *
  * @coversNothing
  *
+ * @group infra
```
theofidry: I introduced it to all test cases that are under `AutoReview` to exclude those tests to be run on AppVeyor. What do you think of `@group auto-review` instead and adding a test to ensure all the test cases under `AutoReview` have this group?

Makefile
sanmai: Looks like here https://github.com/infection/infection/commit/8b5fe199e222eff7a0fd5960815c91f825b0489f#diff-b67911656ef5d18c4ae36cb6741b7965L70

theofidry: So myself... which further proves that I am the main danger to myself and that I cannot code without testing everything

sanmai: Also this might be the reason e2e tests took so much time.

tests/AutoReview/Makefile/MakefileTest.php
```diff
@@ -0,0 +1,305 @@
...
+     * {@inheritdoc}
+     */
+    protected function tearDown(): void
+    {
+        chdir(self::$originalCwd);
+    }
+
+    public function test_the_default_goal_is_the_help_command(): void
+    {
+        $output = shell_exec(sprintf('make -f %s', self::MAKEFILE_PATH));
```
sanmai: I imagine that for some reason the default goal is as before, but build anything in in all possible combination, which also includes running the very this test. This can go on for a long time. How about `make --dry-run` to make sure the above does not happen?

theofidry: I'm not sure to understand you: the make default goal is `help` since #706. Running `--dry-run` right now would output:

```
echo "\033[33mUsage:\033[0m\n  make TARGET\n\n\033[32m#\n# Commands\n#---------------------------------------------------------------------------\033[0m\n"
fgrep -h "##"  Makefile | fgrep -v fgrep | sed -e 's/\\$//' | sed -e 's/##//' | awk 'BEGIN {FS = ":"}; {printf "\033[33m%s:\033[0m%s\n", $1, $2}'
```

Instead of testing the actual output of the help command (which is the purpose of this test)

Makefile
theofidry: Most likely :D

tests/AutoReview/Makefile/MakefileTest.php
sanmai: You can test that it mentions `Usage:`, and then proceed with the current test, or stop right there.

What will be procedure to update this fixture? Will it be better to store it in a file and implement [an automatic updater](https://github.com/infection/infection/blob/master/tests/Console/E2ETest.php#L178)?

theofidry: The idea is to really test the output which includes the alignment (although unfortunately is not easy to test). So I was thinking more of: if the Makefile output changes, then you try the output see if everything is alright and when it is you update the expected string. I don't mind storing it in a separate file, but I'm a bit anxious of people just doing `make > expected-file` which is not desired since you should manually check the output for the colors & alignment

theofidry: If you are concerned about the potential time the command takes, how about adding a very low timeout there? That way if the makefile starts doing something unexpected it will fail fast

theofidry: Couldn't find an easy way to do this :(

sanmai: timeout -k 5 make -f ...

tests/AutoReview/Makefile/MakefileTest.php
```diff
@@ -0,0 +1,305 @@
...
```
sanmai:
```suggestion
        $output = shell_exec(sprintf('timeout -k 5 make -f %s', self::MAKEFILE_PATH));
```
That will send KILL signal after 5 seconds.

tests/AutoReview/Makefile/MakefileTest.php
theofidry: >I'm more concerned about running make from within make running the same Makefile. It may cause failures that will be very hard to reason about.

 Yeah that's a valid concern. But I think the timeout + the fact that the `help` command doesn't really execute anything is good enough. With https://github.com/infection/infection/pull/729 I'm also isolating from where those tests are being run which should make it easier to spot if there is an issue

theofidry:
Rebased; Is there anyone on Linux that could help out to debug the `MakefileTest::test_the_default_goal_is_the_help_command()` command?

tests/AutoReview/Makefile/MakefileTest.php
```diff
@@ -0,0 +1,309 @@
...
+    {
+        self::$originalCwd = getcwd();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function tearDown(): void
+    {
+        chdir(self::$originalCwd);
```
sanmai: I do not see if you're using chdir in any other place in the test. Am I missing something?

tests/AutoReview/Makefile/MakefileTest.php
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    public function test_the_default_goal_is_the_help_command(): void
+    {
+        $output = shell_exec(sprintf(
+            '%s make -f %s 2>&1',
```
sanmai: Builds don't like `-f`, but that's probably should work fine without it.

You can also test if timeout command is available with `command -v timeout`

tests/AutoReview/Makefile/MakefileTest.php
theofidry: No you're right, it was before I made use of `make -f`

tests/AutoReview/Makefile/MakefileTest.php
theofidry: >You can also test if timeout command is available with command -v timeout

 Done.

>Builds don't like -f, but that's probably should work fine without it. Or you'll have to strip the first line.

Not sure what you mean by the builds don't like `-f`, but I think I prefer to use that command over changing of directory to make sure the Makefile used is the right one...

theofidry:
I think it's finally ready :)

tests/AutoReview/Makefile/MakefileTest.php
sanmai: I've tested on a Linux box I have, and it has no those extra lines...
```
$ make -v
GNU Make 4.1
$ make -f ../Makefile  2>&1
It works!
$ cd /tmp/
/tmp$ make -f ~/Makefile  2>&1
It works!
```
So *maybe* straight stripping the first line is no the best workaround.

sanmai: It appears `make -s -f ...` will not output any of these lines.

theofidry: Trying it out

sanmai:
```diff
 test-e2e: 	  Runs the end-to-end tests
-test-infection:   Runs Infection against itself'
+test-infection:   Runs Infection against itself
+'
```
PR isn't merged yet, and we already have these annoying problems. Really this should have been saved in a file, with [an update mechanism like in this E2ETest.](https://github.com/infection/infection/blob/master/tests/Console/E2ETest.php#L178) Then we can expect that the output will be compared verbatim at all times. Still, it's OK to keep it as it is now if you insist. Current approach has a better visibility.

If you're worry about losing colors during future updates, it may worth running `make help` as one of the scripts on Travis. IIRC it'll happily display all the colors.

PR #726, opened 2019-06-27
Add composer flags

theofidry:
Add various composer flags to make the composer install/update command faster and more digest in the CI


theofidry:
@sanmai can you confirm me that the AppVeyor error is just a random one? Looks weird to me that it would fail on `composer install` but not `composer update`

sanmai:
tbh I have no idea what is that going on AppVeyor. Puzzled just like you are.

theofidry:
Seems recurrent... @infection/core does anyone has an idea?

PR #729, opened 2019-06-28
Add autoReview group

### LINKED ISSUE #727: Ensure all tests under AutoReview belong to the `auto-review` group
See https://github.com/infection/infection/pull/723#discussion_r298497357 for more details

theofidry:
- Create a new PHPUnit configuration with a test suite for `AutoReview` only
- Add an entry to the `Makefile` to easily run the auto-review test suite
- Add the auto-review test suite to `Travis#Static Code Analysis`: could be put in a dedicated build but I think it makes sense to have a build with code quality related stuff

Closes #727


theofidry:
Reworked the PR for a _much_ simpler approach than the `@group autoReview` approach. Bonus: the tests should also be slightly faster because longer part of the default test suite (which adds a bit + the coverage - phpdbg or xdebug, overhead).

BackEndTea:
I don't think we should remove the autoreview tests from the default test suite. The fact that it is ran automatically when running `vendor/bin/phpunit` means that new contributor will know what is going on, and what to fix.

theofidry:
The contributor should run `make test` not `vendor/bin/phpunit`, precisely because we can't encapsulate all of our tests with just `vendor/bin/phpunit` :)

theofidry:
You can check the previous WIP commit for reference: it's an order of magnitude more complex (although not overwhelming) and the DX is less nice as well when it comes to add an auto-review test.

Also one advantage of this is:

- It makes the default test suite faster
- We are sure there is no risk of code coverage being generated for an auto-review test hence no possible mutant for it either (which would only slow things down for no reasons)

PR #733, opened 2019-06-30
Move coverage data to the mutation

BackEndTea:
This PR has 2 goals

* Reduce the amount of times the same information about coverage is calculated.
* Simplify the API of the `CodeCoverageData` class.

This is done by fetching all the tests a mutation has when it is created. Whether or not it is tested is then determined by checking if the amount of tests is not 0.

Previously, the coverage information was calculated 3 times per mutation.

First it was checked if it had tests, this was then stored in the mutation, along with extra information to later determine what tests covered this mutation.

Then it was calculated twice per line, first to see if it has test for that line, and then to figure out what tests there are for that line. This is now done in one step, during mutation creation.

Because we do this step earlier, we don't have to pass along a bunch of data, simplifying the following classes:

* The `MutantCreator`
* The `Mutant`, and its interface
* The `Mutation`, and its interface
* The `MutationTestingRunner`
* The `CodeCoverageData`
* The `CoverageMethodData`
* The `MutationsCollectorVisitor`

### DIFF
--- a/src/Mutation.php
+++ b/src/Mutation.php
@@ -69,16 +69,6 @@ final class Mutation implements MutationInterface
     private $mutatedNodeClass;
 
     /**
-     * @var bool
-     */
-    private $isOnFunctionSignature;
-
-    /**
-     * @var bool
-     */
-    private $isCoveredByTest;
-
-    /**
      * @var string
      */
     private $hash;
@@ -94,9 +84,9 @@ final class Mutation implements MutationInterface
     private $mutationByMutatorIndex;
 
     /**
-     * @var array|int[]
+     * @var array
      */
-    private $lineRange;
+    private $tests;
 
     public function __construct(
         string $originalFilePath,
@@ -104,22 +94,18 @@ final class Mutation implements MutationInterface
         Mutator $mutator,
         array $attributes,
         string $mutatedNodeClass,
-        bool $isOnFunctionSignature,
-        bool $isCoveredByTest,
         $mutatedNode,
         int $mutationByMutatorIndex,
-        array $lineRange
+        array $tests
     ) {
         $this->originalFilePath = $originalFilePath;
         $this->originalFileAst = $originalFileAst;
         $this->mutator = $mutator;
         $this->attributes = $attributes;
         $this->mutatedNodeClass = $mutatedNodeClass;
-        $this->isOnFunctionSignature = $isOnFunctionSignature;
-        $this->isCoveredByTest = $isCoveredByTest;
         $this->mutatedNode = $mutatedNode;
         $this->mutationByMutatorIndex = $mutationByMutatorIndex;
-        $this->lineRange = $lineRange;
+        $this->tests = $tests;
     }
 
     public function getMutator(): Mutator
@@ -173,14 +159,14 @@ final class Mutation implements MutationInterface
         return $this->originalFileAst;
     }
 
-    public function isOnFunctionSignature(): bool
+    public function getAllTests(): array
     {
-        return $this->isOnFunctionSignature;
+        return $this->tests;
     }
 
     public function isCoveredByTest(): bool
     {
-        return $this->isCoveredByTest;
+        return \count($this->getAllTests()) !== 0;
     }
 
     /**
@@ -189,10 +175,5 @@ final class Mutation implements MutationInterface
     public function getMutatedNode()
     {
         return $this->mutatedNode;
-    }
-
-    public function getLineRange(): array
-    {
-        return $this->lineRange;
     }
 }
--- a/src/Mutation.php
+++ b/src/Mutation.php
@@ -36,6 +36,7 @@ declare(strict_types=1);
 namespace Infection;
 
 use Infection\Mutator\Util\Mutator;
+use Infection\TestFramework\Coverage\CoverageLineData;
 use PhpParser\Node;
 
 /**
@@ -84,7 +85,7 @@ final class Mutation implements MutationInterface
     private $mutationByMutatorIndex;
 
     /**
-     * @var array
+     * @var CoverageLineData[]
      */
     private $tests;
 
@@ -159,6 +160,9 @@ final class Mutation implements MutationInterface
         return $this->originalFileAst;
     }
 
+    /**
+     * @return CoverageLineData[]
+     */
     public function getAllTests(): array
     {
         return $this->tests;
--- a/src/TestFramework/Coverage/CodeCoverageData.php
+++ b/src/TestFramework/Coverage/CodeCoverageData.php
@@ -98,7 +98,22 @@ class CodeCoverageData
         return \count($coveredLineTestMethods) > 0;
     }
 
-    public function hasTestsOnLine(string $filePath, int $line): bool
+    /**
+     * @return CoverageLineData[]
+     */
+    public function getAllTestsForMutation(
+        string $filePath,
+        array $lineRange,
+        bool $isOnFunctionSignature
+    ): array {
+        if ($isOnFunctionSignature) {
+            return iterator_to_array($this->getTestsForFunctionSignature($filePath, $lineRange), false);
+        }
+
+        return iterator_to_array($this->getTestsForLineRange($filePath, $lineRange), false);
+    }
+
+    private function hasTestsOnLine(string $filePath, int $line): bool
     {
         $coverageData = $this->getCoverage();
 
@@ -113,7 +128,7 @@ class CodeCoverageData
         return !empty($coverageData[$filePath]->byLine[$line]);
     }
 
-    public function hasExecutedMethodOnLine(string $filePath, int $line): bool
+    private function hasExecutedMethodOnLine(string $filePath, int $line): bool
     {
         $coverage = $this->getCoverage();
 
@@ -128,21 +143,6 @@ class CodeCoverageData
         }
 
         return false;
-    }
-
-    /**
-     * @return CoverageLineData[]
-     */
-    public function getAllTestsForMutation(
-        string $filePath,
-        array $lineRange,
-        bool $isOnFunctionSignature
-    ): array {
-        if ($isOnFunctionSignature) {
-            return iterator_to_array($this->getTestsForFunctionSignature($filePath, $lineRange), false);
-        }
-
-        return iterator_to_array($this->getTestsForLineRange($filePath, $lineRange), false);
     }
 
     private function getTestsForFunctionSignature(string $filePath, array $lineRange): \Generator
--- a/src/TestFramework/Coverage/CodeCoverageData.php
+++ b/src/TestFramework/Coverage/CodeCoverageData.php
@@ -41,7 +41,7 @@ use function Safe\file_get_contents;
 /**
  * @internal
  */
-class CodeCoverageData
+final class CodeCoverageData implements CodeCoverageDataInterface
 {
     public const PHP_UNIT_COVERAGE_DIR = 'coverage-xml';
     public const PHP_SPEC_COVERAGE_DIR = 'phpspec-coverage-xml';
--- a/src/TestFramework/Coverage/CodeCoverageData.php
+++ b/src/TestFramework/Coverage/CodeCoverageData.php
@@ -99,6 +99,8 @@ final class CodeCoverageData implements CodeCoverageDataInterface
     }
 
     /**
+     * @param int[] $lineRange
+     *
      * @return CoverageLineData[]
      */
     public function getAllTestsForMutation(
@@ -113,6 +115,11 @@ final class CodeCoverageData implements CodeCoverageDataInterface
         return iterator_to_array($this->getTestsForLineRange($filePath, $lineRange), false);
     }
 
+    /**
+     * @param int[] $lineRange
+     *
+     * @return \Generator<CoverageLineData>
+     */
     private function getTestsForFunctionSignature(string $filePath, array $lineRange): \Generator
     {
         foreach ($lineRange as $line) {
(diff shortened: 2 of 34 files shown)


tests/Mutant/Generator/MutationsGeneratorTest.php
```diff
@@ -107,50 +91,6 @@ public function test_it_can_skip_not_covered_on_file_level(): void
         $this->assertCount(0, $mutations);
     }
 
-    public function test_it_can_skip_not_covered_on_file_line_level(): void
```
BackEndTea: These 2 tests are failing, and i can't really figure out how they are working.

The problem with most of the tests here is that they are mocking dependencies of dependencies, which makes it difficult to figure out what is going on.

It is also testing behavior of child classes, instead of things it does itself.

tests/MutationTest.php
```diff
@@ -74,34 +72,6 @@ public function test_it_correctly_generates_hash(): void
         $this->assertSame('2930c05082a35248987760a81b9f9a08', $mutation->getHash());
     }
 
-    public function test_it_correctly_sets_is_on_function_signature(): void
```
BackEndTea: This is no longer done, so the test is no longer needed

tests/MutationTest.php
```diff
@@ -121,42 +91,11 @@ public function test_it_correctly_sets_original_file_ast(): void
...
-            true,
             new Node\Scalar\LNumber(1),
             0,
             [1, 2, 3]
         );
 
         $this->assertSame($fileAst, $mutation->getOriginalFileAst());
     }
-
-    public function test_it_correctly_sets_line_range(): void
```
BackEndTea: This is no longer done, so the test is no longer needed

tests/Mutant/Generator/MutationsGeneratorTest.php
```diff
@@ -61,15 +64,7 @@ final class MutationsGeneratorTest extends TestCase
 {
     public function test_it_collects_plus_mutation(): void
     {
-        $codeCoverageDataMock = $this->createMock(CodeCoverageData::class);
```
maks-rafalko: Could you please replace (not remove) it with asserting `getAllTestsForMutation()` calls of `$codeCoverageDataMock`?

same below

theofidry:
@BackEndTea ping us when you think it's ready for review!

src/TestFramework/Coverage/CodeCoverageData.php
```diff
@@ -98,73 +97,32 @@ static function ($testMethods) {
...
-        }
-
-        if (!isset($coverageData[$filePath]->byLine[$line])) {
-            return false;
-        }
-
-        return !empty($coverageData[$filePath]->byLine[$line]);
-    }
-
-    public function hasExecutedMethodOnLine(string $filePath, int $line): bool
```
maks-rafalko: 1. Why don't we need it anymore? 
2. where are values `executed`, `coverage` used from `CoverageMethodData` now? If they are not needed, we shouldn't save them in that DTO class

maks-rafalko:
> @BackEndTea ping us when you think it's ready for review!

* Github has [Draft Pull Requests](https://github.blog/2019-02-14-introducing-draft-pull-requests/)
* We have `in progress` / `Needs Review` labels if that helps

src/TestFramework/Coverage/CodeCoverageData.php
BackEndTea: > Why don't we need it anymore?

This function determines if the method is tested or not, if it is tested, we do (almost) the same thing again to determine what test methods test it. Otherwise we skip that.

The method that determines the tests has almost duplicate logic, so instead of checking it first, we simply let that method resolve it.

Where if the method has no tests, it simply returns an empty array (or generator).

src/TestFramework/Coverage/CoverageMethodData.php
```diff
@@ -50,21 +50,9 @@ final class CoverageMethodData
...
-     */
-    public $executed;
-
-    /**
-     * @var int
-     */
-    public $coverage;
-
-    public function __construct(int $startLine, int $endLine, int $executed, int $coverage)
+    public function __construct(int $startLine, int $endLine)
```
BackEndTea: This method no longer contains information about the coverage data of the method, just about where it is located in the file, so it probably needs renaming. I can't find a good name for it atm.

BackEndTea:
@infection/core This is ready for review

BackEndTea:
> Also, I would like to avoid `array`-only typehints as much as possible, adding PHPDoc with a concrete type (like `CoverageLineData[]` instead of `array`). I'm pretty sure I saw CS fixer or PHPStan rule to control it, but can't find it. Is anyone aware of it?

I don't know if our cs-fixer has this, but i know `PHPCodeSniffer` has it, as it is part of the doctrine coding standard

src/Mutation.php
```diff
@@ -173,14 +160,17 @@ public function getOriginalFileAst(): array
...
+    public function getAllTests(): array
     {
-        return $this->isOnFunctionSignature;
+        return $this->tests;
     }
 
     public function isCoveredByTest(): bool
     {
-        return $this->isCoveredByTest;
+        return \count($this->getAllTests()) !== 0;
```
theofidry: let's witch to a `use` statement :) I know it's better to fix it with a CS fixer, but I think we can already change it for newly introduced/changed code

src/TestFramework/Coverage/CodeCoverageData.php
```diff
@@ -98,73 +97,44 @@ static function ($testMethods) {
...
-
-        return iterator_to_array($this->getTestsForMutation($mutation), false);
+        return iterator_to_array($this->getTestsForLineRange($filePath, $lineRange), false);
     }
 
-    private function getTestsForMutationOnFunctionSignature(MutationInterface $mutation): \Generator
+    /**
+     * @param int[] $lineRange
+     *
+     * @return \Generator<CoverageLineData>
```
theofidry: `Generator` should be imported here

src/TestFramework/Coverage/CodeCoverageDataInterface.php
```diff
@@ -0,0 +1,49 @@
...
+ * @internal
+ */
+interface CodeCoverageDataInterface
+{
+    public function hasTests(string $filePath): bool;
+
+    /**
+     * @return CoverageLineData[]
+     */
+    public function getAllTestsForMutation(string $filePath, array $lineRange, bool $isOnFunctionSignature): array;
```
theofidry: what is `$lineRange`?

src/TestFramework/Coverage/CoverageMethodData.php
theofidry: `LineCodeCoverage`?

src/Visitor/MutationsCollectorVisitor.php
```diff
@@ -183,8 +169,12 @@ private function getOuterMostArrayNode(Node $node): Node
     /**
      * @return array|int[]
      */
-    private function getNodeRange(Node $node): array
+    private function getNodeRange(Node $node, bool $isOnFunctionSignature): array
```
theofidry: Side note (it doesn't need to be addressed now): I think `bool $onFunctionSignature` is better as the `is` or `has` prefix are more suited for their corresponding getters

src/TestFramework/Coverage/CodeCoverageData.php
```diff
@@ -35,13 +35,12 @@
...
 namespace Infection\TestFramework\Coverage;
 
-use Infection\MutationInterface;
 use Infection\TestFramework\PhpUnit\Coverage\CoverageXmlParser;
 
 /**
  * @internal
  */
-class CodeCoverageData
+final class CodeCoverageData implements CodeCoverageDataInterface
```
theofidry: it's not really clear for me:

- why switching to an interface + class when there is only a single implementation?
- I quite dislike `CodeCoverageData` as the name, and this IMO is the issue with naming things with the `Interface` prefix.

If you have `FooInterface`, it's easy to name your implementation `Foo` which is not helpful as it doesn't explain what `Foo` is about at all. On the other hand, if you have `Foo` as the interface, besides being more clear as a consumer (because as a consumer you don't care if it's an interface or not), it also force you to choose a better suited name for its implementations.

In this peculiar case, I think `LineCodeCoverage` or `XmlLineCodeCoverage` would be a better suited name?

src/TestFramework/Coverage/CodeCoverageData.php
```diff
@@ -98,73 +97,44 @@ static function ($testMethods) {
         return \count($coveredLineTestMethods) > 0;
     }
 
-    public function hasTestsOnLine(string $filePath, int $line): bool
```
BackEndTea: As a side note, thanks to running infection on itself i discovered that these two methods weren't useful, as any mutation that made them return `true` was never caught, and i couldn't produce tests to catch them.

src/TestFramework/Coverage/CodeCoverageDataInterface.php
BackEndTea: `$lineRange` is the range (of lines) that this node is on. This is especially important for arrays where coverage of them can differ greatly, and we need to figure out exactly all the tests that touch that array.

But it probably isn't that obvious of a name, so i'm open to suggestions. 

As an example

```php
$a = [
    1,
    2,
    3,
];

var_dump(
$a
);
```

In this case the `lineRange` of the node for `$a` would be `[1,2,3,4,5]`. (And due to how we calculate array line ranges, this is the case for every element in that array).

Now for the `var_dump` it would be `[7,8,9]`, but the for the `$a` in there it would be `[8]`

src/TestFramework/Coverage/CodeCoverageData.php
BackEndTea: `LineCodeCoverage` doesn't fit the purpose of this class in my opinion, as the class is also used to check if a file has any coverage at all.

However, for the interface it may be called `CodeCoverageData`, while the implementation is `XMLCodeCoverageData`, as it is a representatation of the xml output from phpunit.

src/TestFramework/Coverage/CodeCoverageDataInterface.php
theofidry: Maybe this could be moved to a dedicated value object with that explanation as a comment?

src/TestFramework/Coverage/CodeCoverageData.php
theofidry: >LineCodeCoverage doesn't fit the purpose of this class in my opinion, as the class is also used to check if a file has any coverage at all.

I don't really see the problem here: it uses an XML based line code coverage report to provide all this information. Also from an API PoV it makes sense that if you can provide the tests for a given mutation, you can tell if there is any test for the given file at all

theofidry:
@BackEndTea any idea when this will be ready?

BackEndTea:
I think this is ready @infection/core

maks-rafalko:
could you please resolve conflicts?

phpunit.xml.dist
```diff
@@ -30,6 +30,7 @@
     <groups>
         <exclude>
             <group>large</group>
+            <group>e2e</group>
```
theofidry: This change looks a bit confusing - I would prefer it being introduced in a dedicated PR where we can talk more of the different PHPUnit groups

src/TestFramework/Coverage/NodeLineRangeData.php
```diff
@@ -0,0 +1,58 @@
...
+
+declare(strict_types=1);
+
+namespace Infection\TestFramework\Coverage;
+
+use Webmozart\Assert\Assert;
+
+/**
+ * This class represents the lines a given node may take.
+ * For example, an array whose declaration takes multiple lines.
```
theofidry: Suggestion:

```php
/** 
  * This class represents the line range for a given node. Indeed some statements can count as one line but be 
  * declared over multiple lines, e.g.:
  *
  * ```
  * $x = [
  *  'a',
  *  'b',
  * ];
  * ```
  *
  * @internal
  */
final class NodeLineRangeData
{
    // ...
} 
```

tests/TestFramework/Coverage/NodeLineRangeDataTest.php
```diff
@@ -0,0 +1,66 @@
...
+use Infection\TestFramework\Coverage\NodeLineRangeData;
+use InvalidArgumentException;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @internal
+ */
+final class NodeLineRangeDataTest extends TestCase
+{
+    public function test_it_errors_if_incorrect_range_is_created(): void
```
theofidry: `test_it_cannot_have_an_incorrect_range`?

tests/TestFramework/Coverage/NodeLineRangeDataTest.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        $this->expectException(InvalidArgumentException::class);
```
theofidry: I kinda dislike `expectException()` since it "hides" which statement can throw, but I guess it's fine in that case

tests/TestFramework/Coverage/NodeLineRangeDataTest.php
```diff
@@ -0,0 +1,66 @@
...
+        new NodeLineRangeData(20, 10);
+    }
+
+    public function test_it_generates_a_range_for_a_one_line_node(): void
```
theofidry: I would group the two other cases into 

```php
test_it_generates_a_range_for_a_line(int $startRange, int $endRange, array $expectedRange): void
```

with a data provider

phpunit.xml.dist
BackEndTea: This is probably due to me turnign it off quickly for local dev to check a few things faster, but i forgot to remove that

BackEndTea:
Resolved all issues now.

Travis failures are due to min MSI

theofidry:
Nice one!

BackEndTea:
Thanks for the review and the merge everyone

Issue #690, opened 2019-05-06
RFC: leave only one element in the non empty returned array


maks-rafalko:
For the following method

```php
public function getSomeCollection(): array
{
    // ...
    
    return $collection;
}
```

we can return a mutated collection with only one element instead of N.

```diff
- return $collection;
+ return \count($collection) > 1 ? [array_keys($collection)[0] => array_values($collection)[0]] : $collection;
```

#### Why `[array_keys($c)[0] => array_values($c)[0]]`?

Because array can be indexed by string keys, by IDs, so not just by indexes `0..N`.

With such code

```php
$collection = [33 => $userObject1, 44 => $userObject2];
```

`[array_keys($collection)[0] => array_values($collection)[0]]` will be evaluated to `[33 => $userObject1];`

and 

```php
$collection = ['abc'=> 1, 'def' => 2];
```
will be evaluated to `['abc'=> 1];`

What do you think?

BackEndTea:
I think this could be a decent mutator, but it may be too complex of a mutation for it to make sense to users.

If i see the following diff, then i don't really know what is being mutated. So far, all diffs make it pretty clear what the 'new' behaviour of the code is.
```diff
- return $collection;
+ return \count($collection) > 1 ? [array_keys($collection)[0] => array_values($collection)[0]] : $collection;
```

So personally i'd say no to this mutator.

localheinz:
How about using `array_slice()` then?

borNfreee:
`array_slice()` with `preserve_keys=true` should do the job, thank you @localheinz 👍 

```php
array_slice(['abc'=> 1, 'def' => 2], 0, 1, true) // ['abc' => 1]

array_slice([33 => $userObject1, 44 => $userObject2], 0, 1, true) // [33 => $userObject1]
```

So the diff will be:

```diff
- return $collection;
+ return \count($collection) > 1 ? array_slice($collection, 0, 1, true) : $collection;
```

PR #649, opened 2019-03-03
#597 Array item removal mutator

majkel89:
This PR:

- [x] Adds new mutator - ArrayItemRemoval
- [x] Covered by tests
- [x] Doc PR: `https://github.com/infection/site/pull/120`

Fixes #597

PR #735, opened 2019-07-01
Leave only one element in the non empty returned array

maks-rafalko:
This PR:

- [x] Adds new Mutator described in https://github.com/infection/infection/issues/690
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/131

Closes https://github.com/infection/infection/issues/690

### DIFF
--- a/src/Mutator/ReturnValue/ArrayOneItem.php
+++ b/src/Mutator/ReturnValue/ArrayOneItem.php
@@ -1,0 +1,116 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Mutator\ReturnValue;
+
+use Infection\Mutator\Util\Mutator;
+use Infection\Visitor\ReflectionVisitor;
+use PhpParser\Node;
+
+/**
+ * @internal
+ */
+final class ArrayOneItem extends Mutator
+{
+    /**
+     * Leaves only one item in the returned array
+     *
+     * Replaces "return $collection;" with "return count($collection) > 1 ? array_slice($collection, 0, 1, true) : $collection;"
+     *
+     * @param Node&Node\Stmt\Return_ $node
+     *
+     * @return Node\Stmt\Return_
+     */
+    public function mutate(Node $node)
+    {
+        /** @var Node\Expr\Variable $expression */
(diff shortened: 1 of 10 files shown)


tests/Fixtures/Autoloaded/ArrayOneItem/does-not-mutate-raw-array.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace ArrayOneItem_RawArray;
+
+class Test
+{
+    public function getCollection(): array
+    {
+        return [];
```
BackEndTea: What is the reason for not mutating a raw array?

src/Mutator/ReturnValue/ArrayOneItem.php
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        $returnType = $functionScope->getReturnType();
+
+        if ($returnType instanceof Node\Identifier) {
+            $returnType = $returnType->name;
+        }
+
+        // no return value specified
+        if (null === $returnType) {
```
BackEndTea: I think we can safely ommit this check here, since `\is_string($returnType)` already catches that

BackEndTea:
I was thinking maybe we could use `array_key_first`, but because `array_value_first` didn't get implemented, that isn't as useful.

tests/Fixtures/Autoloaded/ArrayOneItem/does-not-mutate-raw-array.php
maks-rafalko: Well, there are several reasons:

1. First of all, when you return a raw array (not a variable) from the method/function, you are likely using object-like arrays, not collections. For example

    ```php
    public function getDefaults(): array
    {
        return [
            'a' => 0,
            'b' => 1,
            'c' => 2,
        ];
    }
    ```

    Mutating such an array will produce the following diff, which is very unfriendly:

    ```diff
    public function getDefaults(): array
    {
    -        return [
    -            'a' => 0,
    -            'b' => 1,
    -            'c' => 2,
    -        ];
    +        return count([
    +            'a' => 0,
    +            'b' => 1,
    +            'c' => 2,
    +        ]) > 1 ? array_splice([
    +            'a' => 0,
    +            'b' => 1,
    +            'c' => 2,
    +        ], 0, 1, true) : [
    +            'a' => 0,
    +            'b' => 1,
    +            'c' => 2,
    +        ];
    }
    ```

    This is hard to read and understand.

2. For object-like arrays, we already have the following mutator:

    ```diff
    public function getDefaults(): array
    {
        return [
    -       'a' => 0,
    +       'a' > 0,
            'b' => 1,
            'c' => 2,
        ];
    }
    ```

    Thus, if you are killing such mutant by test, you will 99% kill the `ArrayOneItem` mutant since you are checking the keys of the array. Here we will have a "duplication".

BackEndTea: Makes sense, forgot about that last mutator.

maks-rafalko:
Thank you @sanmai & @BackEndTea

tests/Mutator/ReturnValue/ArrayOneItemTest.php
```diff
@@ -0,0 +1,102 @@
...
+<?php
+
+namespace ArrayOneItem_NotNullableArray;
+
+class Test
+{
+    public function getCollection() : array
+    {
+        $collection = [1, 2, 3];
+        return count($collection) > 1 ? array_slice($collection, 0, 1, true) : $collection;
```
morozov: The fact that the mutation produces a conditional statement opens up the possibility of false positives. At the code execution time, depending on the size of the collection, the mutation may or may not effectively happened but it will also be counted as such.

Some collections have one element by design. For instance, all sorts of JSON-based API responses that have one top-level element:
```js
{
   "response": { /* ... */}
}
```

The code that consumes this API and the corresponding tests don't expect any other elements in the collection, therefore the mutation (that effectively doesn't change anything) passes.

maks-rafalko: Agree. And this is where the user can disable mutator for particular line, method or completely for the project.

Do you have any suggestions here?

This mutator is useful for the opposite scenario, where we are working with e.g. Doctrine repositories and expect > 1 elements in the returned array.

morozov: > This mutator is useful for the opposite scenario, where we are working with e.g. Doctrine repositories and expect > 1 elements in the returned array.

Do you mean that they expect more than one item but only check the first one? Shouldn't the `ArrayItemRemoval` mutation flag such tests?

> Do you have any suggestions here?

Judging by the documentation, this is the only mutation that generates a conditional, one branch of which is the same as the original code. I suggest to avoid that and make sure the code is mutated unconditionally (however it's still not clear how it's different from `ArrayItemRemoval`).

maks-rafalko: > Shouldn't the `ArrayItemRemoval` mutation flag such tests?

`ArrayItemRemoval` mutates the raw arrays, not variables. When you literally have `['a' => 1, 2, 3, 'd' => 4]` for example.

This mutator, `ArrayOneItem`, mutates variables. That's the difference.

morozov: Indeed, even item removal cannot be done unconditionally. You cannot remove it from an empty array. Still, throwing an exception on an unsuccessful mutation should help avoid false positives.

maks-rafalko: > even item removal cannot be done unconditionally. You cannot remove it from an empty array

we [don't mutate ](https://github.com/infection/infection/pull/649/files#diff-871cc5d111cdcdea40d147449608faeeR55-R107)empty array with `ArrayItermRemoval1 mutator :) So there are no conditions.

> Still, throwing an exception on an unsuccessful mutation should help avoid false positives.

Do you have false-positives with `ArrayOneItem` mutator because of conditions? What is the real source code?

I'm open to change it to throw an exception, just need more information about real cases and then we can create issue and discuss farther / implement.

morozov: > we [don't mutate ](https://github.com/infection/infection/pull/649/files#diff-871cc5d111cdcdea40d147449608faeeR55-R107)empty array with `ArrayItermRemoval` mutator :) So there are no conditions.

That's the thing. You can see if the array literal is empty at mutation time and avoid false positives but you cannot do the same with a variable.

> Do you have false-positives with `ArrayOneItem` mutator because of conditions? What is the real source code?

Yes. See https://github.com/morozov/kih/pull/13.

BackEndTea: I have put this into an issue so we can track this a bit better: https://github.com/infection/infection/issues/803

Issue #615, opened 2019-01-23
Multiple extra test framework options escape in the wrong way


sidz:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.11.5
| PHP version | 7.2.14
| Platform    | Ubuntu/Windows/MacOS


The issue initially reported by **dkarlovi** in Slack:

Infection throws an exception for case when user adds more then one test framework option in the `---test-framework-options`. The issue is caused by not enough extra options escaping.

_command:_
`phpdbg -qrr /tools/infection --test-framework=phpunit --test-framework-options="--verbose --group unit"`

_throws:_
```
[ERROR] Project tests must be in a passing state before running Infection.                                            
                                                                                                                       
        Check the executed command to identify the problem: '/usr/local/bin/phpdbg' '-qrr'                            
        '/project/vendor/phpunit/phpunit/phpunit' '--configuration'                                                    
        '/tmp/infection/phpunitConfiguration.initial.infection.xml' '--verbose --group unit'                          
                                                                                                                       
        PHPUnit reported an exit code of 1.                                                                            
                                                                                                                       
        Refer to the PHPUnit's output below:                                                                          
                                                                                                                       
        STDOUT:                                                                                                        
                                                                                                                       
        PHPUnit 7.0.0 by Sebastian Bergmann and contributors.                                                          
                                                                                                                       
        unrecognized option --verbose --group unit
```


https://github.com/infection/infection/blob/c1fa0751e8e57c72cd767c86302a4b24cfa1ce2a/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php#L50

maks-rafalko:
@sidz did you have an idea how to fix it? I remember we discussed something related to all this stuff, but not exaclty

sidz:
@maks-rafalko  yup. Just need to explode `--verbose --group unit` with an empty space delimiter at the top layer of Infection. Then pass this array to the instead of string:

https://github.com/infection/infection/blob/c1fa0751e8e57c72cd767c86302a4b24cfa1ce2a/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php#L45-L52 

That's it. Symfony Process will do all the things inside it.

p.s. Hope I will find some time on this week.

maks-rafalko:
Released https://github.com/infection/infection/releases/tag/0.13.6

sanmai:
[...]

PR #736, opened 2019-07-05
Multiple extra test framework options escape in the wrong way

sidz:
This PR:
Fixes https://github.com/infection/infection/issues/615

p.s. Probably need to write an e2e test to avoid such issues in the future.


maks-rafalko:
@sanmai I've added tests to ensure it works as expected

Thank you @sidz

PR #743, opened 2019-07-12
Bump the versions used where appropriate

sanmai:
This should not affect a .phar build in any way as it uses these versions anyways.


maks-rafalko:
Will our users, who use Infection as a composer dependency, love this change? I doubt :(

If I'm correctly understand it, Infection will only be possible to be installed for those who use Symfony 4.3 (as an example), but not 4.2.

sanmai:
That's understood. Lets define the minimal supported version for all packages we depend on, shall we?

I don't see this as a major drawback since .phar is the recommended way to use Infection, and there it does not matter if you use Symfony 4.3 or 4.2.

maks-rafalko:
https://twitter.com/infection_php/status/954603440394719232?s=20

63% of developers prefer install Infection as a comoser dependency, *unfortunately*.

Also, according to github api, latest PHAR (`0.13.4`) downloaded 95 times (`"download_count": 95,`) while there are [207 downloads](https://packagist.org/packages/infection/infection/stats#0.13.4) per day (!) on packagist.

What I'm trying to say is that (again, unfortunately) composer dependency is a majority of our installations.

> Lets define the minimal supported version for all packages we depend on, shall we?

Sounds ok. But what issues we want to to solve in the first place?
Probably, we can check what old package versions does not work with PHP `7.2.9` and simply drop them from our `composer.json`, because we Infection requires this version at minimum

sanmai:
The issue at hand is the issue with DEPS="LOW" and PHPUnit you had #741. But this issue was fixed for good by #744. So naturally we don't have any solid issues with older packages, but we don't know if we're missing anything. Say, a newest version has a feature X, but we don't use it, not even aware of it, because we're permitting some older versions. So a part of the problem boils down to the things we don't know.

So far almost every package we depend on works fine. So we don't have a solid reason to this forced upgrade from the faults alone. This is the second part of the problem.

Third and the most important part is compatibility. Say, we're promising to support `compatbsebastian/diff` with versions 1.4, 2.0, 3.0. But in reality we can't ever test Infection with all these versions because [PHPUnit 8 requires ^3.0.2](https://packagist.org/packages/phpunit/phpunit#8.2.4). Otherwise put, we're making a false promise and we don't even know if Infection works with `compatbsebastian/diff` 1.4 or not. 

So I think next step will be finding the lowest version of every package we can have installed on CI, and then removal of everything we can't install and test with. That's what I'm going to do.

BackEndTea:
Even PHPUnit 7.5 requires 3.0 for the diff package: https://github.com/sebastianbergmann/phpunit/blob/7cf077177f6ae0152fd219d14958c05062bfec4d/composer.json#L40

So i'm fine with dropping support for lower versions of that.

BackEndTea:
I do think its in our best interest to support installation with LTS versions of symfony, especially since composer is the most used installation method, by far.

I'm however totally fine with bumping the minimum (patch) versions for those versions.
This means we can also clean up the `conflict` part of our composer.json.

Personally i'd say bump minimum 4.* version to 4.2, and bump the 3.4 branch to `3.4.26`. Thats the `.26` is because that is the latest security fix.

We could probably bump those a bit for every minor release

theofidry:
A rebase is needed, but LGTM otherwise

composer.json
```diff
@@ -37,27 +37,24 @@
...
+        "pimple/pimple": "^3.2.3",
+        "sebastian/diff": "^3.0.2",
+        "symfony/console": "^4.3.2",
+        "symfony/filesystem": "^4.3.2",
+        "symfony/finder": "^4.3.2",
+        "symfony/process": "^4.3.2",
+        "symfony/yaml": "^4.3.2",
+        "webmozart/assert": "^1.4"
     },
+    "conflict": {},
```
theofidry: why is the conflict key necessary here?

maks-rafalko:
With the current state of the PR, users will only be able to install Infection with Symfony 4.3, but not with the earlier versions. Is it really what we want?

sanmai:
> Personally i'd say bump minimum 4.* version to 4.2, and bump the 3.4 branch to 3.4.26. Thats the .26 is because that is the latest security fix.
I've bumped the 3.4 branch to 3.4.26. Shall we bump 4.x to 4.2.10, @maks-rafalko?

composer.json
```diff
@@ -51,24 +51,23 @@
...
+        "symfony/process": "^3.4.29 || ^4.0",
+        "symfony/yaml": "^3.4.29 || ^4.0",
         "thecodingmachine/safe": "^0.1.16",
         "webmozart/assert": "^1.3",
         "webmozart/path-util": "^2.3"
     },
     "conflict": {
-        "symfony/console": "=3.4.16 || =4.1.5",
-        "symfony/process": "3.4.2"
+        "symfony/console": "=4.1.5"
```
theofidry: why is that one necessary?

BackEndTea: 4.1.5 has an issue that throwing an exception still causes you to exit with a 0 code.

sanmai: That's #523

maks-rafalko:
> Shall we bump 4.x to 4.2.10

@sanmai 

In my opinion, it's better to have `conflicts: =4.1.5` and allow `^4.0` as it's currently done in this PR rather that dropping `conflicts` and update to `^4.2`.

We can afford it. But not every team can afford updating the whole project with the minor version of the framework.

On the other side, people can use old Infection versions if they stuck with the Symfony < 4.2 *and* don't want to use PHAR installation. I don't know...

Let's see how it goes. If it will be too expensive to maintain `^4.0` (which I doubt), we can bump it to `^4.2.x` at any time later.

composer.json
```diff
@@ -42,7 +42,7 @@
         }
     ],
     "require": {
-        "php": "^7.2",
+        "php": "^7.2.9",
```
morozov: I'm curious where this specific constraint comes from. Does it contain some bug fix that Infection relies on?

I maintain a package which requires `"php": "^7.2"` where I cannot upgrade Infection. While PHP 7.2 is still supported by its maintainers, I'm curious if I should start requiring `"php": "^7.2.9"` as well or not upgrade Infection until the PHP 7.2's EOL.

sanmai: How about that's the version Infection was tested to work with. If we'd advertise that we support everything down to PHP 7.2.0, that will mean that we got to be testing at all these versions. But we don't, so here we go. Otherwise put, this requirement is primarily a reflection of an actual deployment situation.

(Honestly I think it is about time to bump that requirement up to 7.3. Even Debian stable already has it.)

sanmai: The problem with you package is that you require that [platform must be PHP 7.2.0 exactly](https://github.com/morozov/kih/blob/master/composer.json#L13). Most of the time you don't need to declare a platform, Infection being one of the special cases where this is useful. E.g. if people install your package from composer, then you don't ever need to declare the platform, but if you're building a Phar that is supposed to work with a specific version of PHP, but you also intend to test that Phar under other versions of PHP like 7.3 or 7.4, then it is imperative to have this constrain because it makes sure every Phar you build even under PHP 7.3 is compatible with PHP 7.2.

Bottom line: remove your platform requirement and sleep well.

morozov:
@sanmai has it been tested with all versions between `7.2.10` and `7.2.21`? It should have been according to the argument above.

> (Honestly I think it is about time to bump that requirement up to 7.3. Even Debian stable already has it.)

It is but it's irrelevant to the question. By dropping PHP 7.2 you may rely on the features available in the newer release. By dropping PHP <= 7.2.8 you're not getting anything. The "actual deployment situation" does not reflect the actual dependency constraint `^7.2.9` which also includes `7.4` and `dev-master`.

composer.json
morozov: @sanmai has it been tested with all versions between `7.2.10` and `7.2.21`? It should have been according to the argument above.

> (Honestly I think it is about time to bump that requirement up to 7.3. Even Debian stable already has it.)

It is but it's irrelevant to the question. By dropping PHP 7.2 you may rely on the features available in the newer release. By dropping PHP <= 7.2.8 you're not getting anything. The "actual deployment situation" does not reflect the actual dependency constraint `^7.2.9` which also includes `7.4` and `dev-master`.

sanmai: Well, no. I think I took the wrong turn with that my argument, because your problem is that you say you package is supposed to run under exactly PHP 7.2.0 by means of the platform directive. But is this true? What is the actual version that you use? I doubt it is exactly PHP 7.2.0.

sanmai: 7.2.8 was released like, what, summer 2018, over a year ago. 7.2.9 was also released over a year ago. I don't see how this could be the actual problem here.

morozov: > The problem with you package […]

Not being able to upgrade Infection is not my problem. I know _why_ I cannot upgrade it and was asking about the specific version constraint in Infection. This platform configuration exists in order to not let Composer install the dependencies which would break the `^7.2` requirement. E.g. if run `composer update` on PHP 7.3, it will install `ocramius/package-versions 1.5` which will require PHP 7.3 which I'd like to avoid.

> […] sleep well

As a package maintainer, it's not your concern. Please watch your tone.

morozov: > If you feel like we must support installing Infection on PHP 7.2.0 exactly, you can certainly open a PR with that suggestion.

No. I was asking why exactly `^7.2.9`. It's up to the package maintainers to decide which versions to support.

sanmai: I recall this might be the version of the PHP 7.2 branch installed at that time. Makes sense?

morozov: Installed where?

sanmai: It was on my development machine I used then. I'm sorry I do not exactly have time now for this or any interrogation, so let's not continue with it. You're totally free to open a PR with this change, and if there's a consensus, it might even get merged.

morozov: I'm asking these questions to understand the reasoning behind this change while you're being too emotional. You're free to not answer them if it feels like an interrogation.

As I understand them so far, the reasons are:

1. PHP 7.2.8 and the older versions are too old.
2. The non-latest PHP versions have security issues, however, some of the non-latest PHP versions are still supported.
3. PHP 7.2.9 was installed on your development machine at some point in time.

It all makes sense. Thank you for the explanation.

PR #744, opened 2019-07-12
Allow installation only with the most recent versions for dev dependencies

sanmai:
Just like #743, but for dev dependencies only.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,24 +37,27 @@
         }
     ],
     "require": {
-        "php": "^7.2.9",
+        "php": "^7.2",
         "ext-dom": "*",
         "ext-json": "*",
         "ext-libxml": "*",
         "composer/xdebug-handler": "^1.3.3",
-        "justinrainbow/json-schema": "^5.2.8",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -59,8 +59,8 @@
         "symfony/process": "3.4.2"
     },
     "require-dev": {
-        "helmich/phpunit-json-assert": "^2.1 || ^3.0",
-        "phpunit/phpunit": "^8"
+        "helmich/phpunit-json-assert": "^3.0",
```
sanmai: [3.x is the only version that supports PHPUnit 8](https://github.com/martin-helmich/phpunit-json-assert#compatibility); we can't use 2.x in any case

composer.json
```diff
@@ -59,8 +59,8 @@
...
+        "phpunit/phpunit": "^8.2.4"
```
localheinz: 👍 

I understand developers have different opinions on this, but personally, I always use full version constraints (i.e., `major.minor.patch`). In my opinion, this documents awareness of features and bug fixes well.

PR #745, opened 2019-07-13
Use codingmachine/safe

BackEndTea:
This PR:

- [x] Adds [`thecodingmachine/safe`](https://github.com/thecodingmachine/safe) as a dependency
- [x] Fixes `assert` calls and (mostly) useless casts.

I can't fix everything at this moment, for example `preg_replace` has different return types bases on its parameters, and PHPStan can't understand that from the safe function. So i've opted to keep those as is right now.


BackEndTea:

maks-rafalko:
Understood. Thank you @BackEndTea

PR #747, opened 2019-07-16
Refactoring of the Locator

theofidry:
Working on the configuration I found out I need some tweaks in the `Locator`. However after check it, I think there is a few changes that could be done upstream before that.

Convention changes:

- Propose to drop the `Interface` prefix, which forces the default implementation to have a more explicit name. For example here instead of having `LocatorInterface` and `Locator`, we have `Locator` and `RootsFileOrDirectoryLocator` (and in the future probably a `RootsFileLocator` coming up).
- In the same spirit propose to change `LocatorException` for names without the `Exception` suffix, transforming `throw LocatorException::fileOrDirectoryDoesNotExist(...)` for `throw FileNotFound::createForFile(...);`
- Add a `integration` PHPUnit group:  I guess in the future it will be useful to be able to distinguish between pure unit tests and tests which have any IO/DB/network interaction.

Code changes:

- It felt a bit weird to have `Locator` in `Finder` with all the finder related things so I moved it in a dedicated namespace. That said I get they are kinda related so I don't mind so much keeping them under `Finder`
- Add a few more tests
- Make the error messages more explicit
- Make the locator more forgiving in the case of directories by allowing them to end with a trailing slash, e.g. `Locator::locate('dir/')`
- Add guards and phpdoc where appropriate


.gitignore
```diff
@@ -1,20 +1,20 @@
-/build/
-!/build/cache/.gitkeep
+!/.tools/.gitkeep
```
theofidry: It looks like the order is platform dependent, maybe it would be wise to run that command in a docker container to avoid this issue

composer.json
```diff
@@ -58,7 +58,8 @@
         "symfony/process": "^3.4 || ^4.0",
         "symfony/yaml": "^3.4 || ^4.0",
         "thecodingmachine/safe": "^0.1.16",
-        "webmozart/assert": "^1.3"
+        "webmozart/assert": "^1.3",
+        "webmozart/path-util": "^2.3"
```
theofidry: I'm not super found of depending on it, but I think it's ok for now. Long term plan would be https://github.com/symfony/symfony/pull/30969 if it lands

src/Command/InfectionCommand.php
```diff
@@ -325,7 +326,7 @@ private function includeUserBootstrap(InfectionConfig $config): void
 
         if ($bootstrap) {
             if (!file_exists($bootstrap)) {
-                throw LocatorException::fileOrDirectoryDoesNotExist($bootstrap);
+                throw FileNotFound::createForFile($bootstrap);
```
theofidry: I did not realise that exception was used elsewhere this way, that said in this specific case it feels a `LogicException` (instead of a `RuntimeException` which is what `FileNotFound` is) is more appropriate since it is a configuration error that requires immediate fix from the user

src/Locator/RootsFileOrDirectoryLocator.php
```diff
@@ -33,66 +33,89 @@
...
-    public function locate(string $name): string
+    /**
+     * {@inheritdoc}
+     */
+    public function locate(string $fileFileName): string
     {
-        if ($this->filesystem->isAbsolutePath($name)) {
-            if ($this->filesystem->exists($name)) {
-                return realpath($name);
+        $fileFileName = Path::canonicalize($fileFileName);
```
maks-rafalko: 1. What does `Path::canonicalize()` fix? Why do we need it?

2. If it's really needed, could we please not modify original parameter but created a local var instead? (common good practise)

tests/Config/ConfigCreatorFacadeTest.php
```diff
@@ -55,7 +55,7 @@ final class ConfigCreatorFacadeTest extends TestCase
     private $creatorFacade;
 
     /**
-     * @var LocatorInterface|MockObject
+     * @var \Infection\Locator\Locator|MockObject
```
maks-rafalko: Just `Locator|MockObject` ?

src/Locator/RootsFileOrDirectoryLocator.php
```diff
@@ -33,66 +33,89 @@
...
+
+        $this->roots = $roots;
         $this->filesystem = $filesystem;
     }
 
-    public function locate(string $name): string
+    /**
+     * {@inheritdoc}
+     */
+    public function locate(string $fileFileName): string
```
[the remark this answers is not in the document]
theofidry: Definitely, mishandled a phpStorm refactor command

src/Locator/RootsFileOrDirectoryLocator.php
theofidry: 1. It is mostly to take care of the trailing `/`. It would be trivial to not depend on `Path` here on linux machines, but I'm less confident about handling it properly on Windows (I don't know if there is any special case I'm not aware of to handle).

2. Noted

tests/Config/ConfigCreatorFacadeTest.php
theofidry: Should actually be `&`, but should be imported yes

.gitignore
```diff
@@ -1,20 +1,20 @@
...
 !/build/.gitkeep
+!/build/cache/.gitkeep
```
sanmai: Not a big fan of a single project-wide .gitignore tbh.

.gitignore
sanmai: Or locale dependent. `LC_ALL=C sort` may help.

.gitignore
theofidry: By experience with Symfony, it is way more confusing for people in general to have gitignore files left and right.

.gitignore
theofidry: Will try in a separate PR and remove the changes made to the gitignore here

.gitignore
sanmai: True, having `.gitignore` left and right isn't right too, pun intended.

sanmai: Anyway, it is already as it is, and this is not about it here.

tests/Locator/RootsFileOrDirectoryLocatorTest.php
```diff
@@ -0,0 +1,437 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Tests\Locator;
+
+use Generator;
+use Infection\Locator\FileNotFound;
+use Infection\Locator\RootsFileOrDirectoryLocator;
+use function Infection\Tests\normalizePath as p;
```
localheinz: What do you think about avoiding an alias here?

tests/Locator/RootsFileOrDirectoryLocatorTest.php
```diff
@@ -0,0 +1,437 @@
...
+        array $roots,
+        string $file,
+        string $expectedErrorMessage
+    ): void {
+        $locator = new \Infection\Locator\RootsFileOrDirectoryLocator($roots, $this->filesystem);
+
+        try {
+            $locator->locate($file);
+
+            $this->fail();
```
localheinz: What do you think about moving this statement out of the `try`-block, and instead using a `return` statement in the `catch` block after successful inspection of the expected exception?

src/Locator/RootsFileOrDirectoryLocator.php
```diff
@@ -33,66 +33,89 @@
...
+    private function innerLocateOneOf(array $fileNames): ?string
+    {
+        if ([] === $fileNames) {
+            return null;
         }
 
         try {
-            return $this->locate($fileNames[0]);
-        } catch (\Exception $e) {
+            return $this->locate(current($fileNames));
```
localheinz: Personally, I’m not a friend of using a `return` statement in a `try`-block (as I prefer to reduce the scope of it to statements that can actually throw an exception, and `return` isn’t one of it) what do you think about introducing a temporary variable and returning it after the `try`-and-`catch` block?

src/Locator/FileNotFound.php
```diff
@@ -33,25 +33,37 @@
...
  * @internal
  */
-final class LocatorException extends \RuntimeException
+final class FileNotFound extends RuntimeException
 {
-    public static function fileOrDirectoryDoesNotExist(string $name): self
+    /**
+     * @param string[] $roots
+     */
+    public static function createForFile(string $file, array $roots): self
```
localheinz: Perhaps `fromFilename()` is a better choice, along with `$fileName` as parameter name?

src/Locator/FileNotFound.php
```diff
@@ -63,8 +75,27 @@ public static function multipleFilesDoNotExist(string $path, array $files): self
         );
     }
 
-    public static function filesNotFound(): self
+    /**
+     * @param string[] $files
+     * @param string[] $roots
+     */
+    public static function createForFiles(array $files, array $roots): self
```
localheinz: Perhaps `fromFilenames()` is a better choice, along with `$fileNames` as parameter name?

src/Locator/FileNotFound.php
```diff
@@ -63,8 +75,27 @@ public static function multipleFilesDoNotExist(string $path, array $files): self
...
     {
-        return new self('Files are not found');
+        Assert::allString($files);
+        Assert::allString($roots);
+
+        if ([] === $files) {
+            $message = 'Could not locate any files (no file provided).';
```
localheinz: What do you think about using an early return here, so we can avoid the `else` branch, and perhaps make it more obvious that the latter is the happy path?

src/Locator/Locator.php
```diff
@@ -45,7 +45,7 @@ interface Locator
      *
      * @throws FileNotFound
      */
-    public function locate(string $fileName): string;
+    public function locate(string $name): string;
```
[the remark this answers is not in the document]
theofidry: I thought i did, fixed

tests/Locator/RootsFileOrDirectoryLocatorTest.php
theofidry: I personally find it clearer like that IMO. The flow is harder to follow if you rely on a return statement here and as error pron

src/Locator/RootsFileOrDirectoryLocator.php
sanmai: One more variable, one more thing keep track of. That's really taxing. Better without.

theofidry: Like so?

```php
try {
    $path = $this->locate(current($fileNames));
} catch (FileNotFound $exception) {
    // Continue
}

if (isset($path)) {
    return $path;
}

array_shift($fileNames);

return $this->innerLocateOneOf($fileNames);
```

Not any clearer/readable IMO

.gitignore
sanmai: Think you can merge with master.

src/Locator/FileNotFound.php
theofidry: I like it

theofidry: except it's `$file` or maybe `$filePath`  since in this peculiar case it's the path of the file not its name

src/Locator/FileNotFound.php
theofidry: We can actually completely get rid of the if statement with a ternary and the `$message` variable which is a single usage variable

src/Locator/RootsFileOrDirectoryLocator.php
```diff
@@ -33,66 +33,89 @@
...
+        return $file;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    private function innerLocateOneOf(array $fileNames): ?string
+    {
+        if ([] === $fileNames) {
+            return null;
```
localheinz: Perhaps an empty `string` would be a better option than using `null`?

src/Locator/RootsFileOrDirectoryLocator.php
localheinz: More like

```php
try {
    $path = $this->locate(current($fileNames));
} catch (FileNotFound $exception) {
    // ...
}

return $path;
```

src/Locator/RootsFileOrDirectoryLocator.php
theofidry: I personally find using an empty string over `null` more error prone, but I don't feel strongly about it either so if others feel that way too I'm fine with changing it

tests/Locator/RootsFileOrDirectoryLocatorTest.php
theofidry: fine by me, I would also make it snake case as well (as a function), but since it's a project convention that is easy to change I would change it everywhere in a dedicated PR

src/Locator/RootsFileOrDirectoryLocator.php
localheinz: At least we could get rid of the nullable return type declaration.

tests/Locator/RootsFileOrDirectoryLocatorTest.php
localheinz: For the reader it’s not clear though which of these statements might produce an exception.

src/Locator/RootsFileOrDirectoryLocator.php
theofidry: It is on purpose though.

In the public method `locateOneOf()`, it is "get the result or throw" whereas here it's the internal version which is "try to _find_ a result", i.e. not having a result is a happy path (consequence of the recursion)

src/Locator/RootsFileOrDirectoryLocator.php
theofidry: But then you are returning in the `catch` block... why is returning in the `try` problematic but not in the `catch`?

tests/Locator/RootsFileOrDirectoryLocatorTest.php
theofidry: I don't think it's causing any confusion because you're not supposed to be able to catch PHPUnit's exceptions that way, neither people are necessary aware that `fail()` actually just throw a specific exception. To me it reads more clearly as "I've reached that line but that shouldn't have happened".

That being said I called `$this->fail()` out of habit but I guess I can try to be more friendly here like I did in other tests and put a more helpful error message

src/Locator/RootsFileOrDirectoryLocator.php
```diff
@@ -33,66 +33,89 @@
...
+    /**
+     * {@inheritdoc}
+     */
     public function locateOneOf(array $fileNames): string
     {
-        if (!$fileNames) {
-            throw LocatorException::filesNotFound();
+        $file = $this->innerLocateOneOf($fileNames);
+
+        if (null === $file) {
```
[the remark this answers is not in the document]
maks-rafalko: hm.. we even have a rule to prevent this https://github.com/infection/infection/blob/master/.php_cs.dist#L121

but PrettyCI is green :|

src/Locator/RootsFileOrDirectoryLocator.php
```diff
@@ -33,66 +33,89 @@
...
                 return realpath($file);
             }
         }
 
-        throw LocatorException::filesOrDirectoriesDoNotExist($name, $this->paths);
+        throw FileNotFound::createForFile($fileFileName, $this->roots);
     }
 
+    /**
+     * {@inheritdoc}
```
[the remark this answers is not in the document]
theofidry: I would be for doing a separate PR removing all of them then

src/Locator/RootsFileOrDirectoryLocator.php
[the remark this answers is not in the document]
theofidry: I can fix it manually but it would be better that the CS-Fixer could take care of it...

src/Locator/RootsFileOrDirectoryLocator.php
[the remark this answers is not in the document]
sanmai: Please reconsider. Some static analysers use this tag to infer the proper typing. IIRC this was PHPStan or Psalm. Without `{@inheritdoc}` you will get a very annoying head-scratching issues from them, or even make them useless.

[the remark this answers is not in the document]
theofidry: To be continued in https://github.com/infection/infection/issues/749

src/Locator/RootsFileOrDirectoryLocator.php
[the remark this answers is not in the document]
BackEndTea: > hm.. we even have a rule to prevent this 

I think that just turns off the rule, as it is part of the symfony profile.

theofidry:
Ok now PHP-CS-Fixer is being ridiculous, if you have:

```php
use function defined;

defined(...);
```

It will add the backslash... I'm really looking forward the time where we will get rid of the rule

src/Locator/RootsFileOrDirectoryLocator.php
localheinz: It’s not problematic to use a `return` statement in a `try`-block, it’s only a personal convention to reduce the scope of a `try`-block to statements that can actually throw an exception, that’s all.

theofidry: Ah I see. It does make sense, but I don't think it really helps out here :)

theofidry:
The Travis failure is, if I understand correctly, due to some files being touched but (e.g. InfectionCommand) but their tests untouched (hence decreasing the overall MSI). In which case I can't do much about it.

So another round of review or approval this time 🙏 :)

tests/Locator/FileNotFoundTest.php
```diff
@@ -0,0 +1,177 @@
...
+        $this->assertSame(0, $exception->getCode());
+        $this->assertNull($exception->getPrevious());
+    }
+
+    public function test_multiple_files_do_not_exist(): void
+    {
+        $exception = FileNotFound::multipleFilesDoNotExist('foo/bar/', ['file1', 'file2']);
+
+        $this->assertSame(
+            'The path foo/bar/ does not contain any of the requested files: file1, file2',
```
maks-rafalko: What about 

```diff
- The path foo/bar/ does not contain any of the requested files: file1, file2
+ The path "foo/bar/" does not contain any of the requested files: "file1", "file2"
```

as it is done for `fromFiles()` method

theofidry:
Thanks for the reviews!

PR #748, opened 2019-07-17
Specify the local for the .gitignore file sorting

theofidry:
Since I'm on OSX, could someone try on Linux to see if the order changes still?


Makefile
```diff
@@ -44,7 +44,7 @@ compile: $(INFECTION)
 cs:	  	 ## Runs PHP-CS-Fixer
 cs: $(PHP_CS_FIXER)
 	$(PHP_CS_FIXER) fix -v --cache-file=$(PHP_CS_FIXER_CACHE)
-	sort -u .gitignore -o .gitignore
+	LC_ALL=C sort -u .gitignore -o .gitignore
```
localheinz: Nice, great idea to automatically sort!

PR #750, opened 2019-07-23
Rework the configuration

theofidry:
The goal of this PR is to split the configuration in two: `SchemaConfiguration` and `Configuration`
  - The `SchemaConfiguration` class should model closer to the `schema.json` file. I.e. this class aims to translate the structure of `schema.json` in PHP.
  - The `Configuration` class should be a higher level class which is meant to be consumed by our app

In practice it means we can easily create the `SchemaConfiguration` from the configuration file found and then a `Configuration` object from the `SchemaConfiguration` and user inputs (options from the CLI). This translate to using `Configuration` directly in Infection, without having to worry about where the parameter comes from or if it needs normalization etc. In other words, much easier to consume.

I've limited to possible refactoring to the bare minimum in order to reduce the number of changes.

To illustrate what could change is, `Configuration` could for example provide the list of source files to mutate instead of the "raw" configuration dirs + excludes: in other words an already "interpreted" config easier to consume for our app. I believe this could drastically simplify how we currently consume the configuration.

Another benefit of this approach is debugability. For example the `Configuration` could be "dumpable" which means when running Infection in debug mode, this dump of the config, which would provide deeper insights of how we interpreted the user configuration, could be provided. An example of this is [Box](https://github.com/humbug/box). It's [Configuration](https://github.com/humbug/box/blob/master/src/Configuration/Configuration.php) is [exportable](https://github.com/humbug/box/blob/master/src/Configuration/Configuration.php#L554) which allows one to provide it when opening an issue. An example of the output can be found [here](https://github.com/humbug/box/blob/master/tests/Configuration/ConfigurationTest.php#L2928-L2993).


theofidry:
I was more thinking of a POPO generator based on the `schema.json`, but the projects found in the wild, e.g. https://github.com/swaggest/php-code-builder or https://github.com/RoundingWellOS/schematic are not good enough.

One way like another: it's fine by me, I just don't want to dive in there at least yet tbh. But it could definitely be a huge improvement

sanmai:
For example, as far as I am aware, JMS Serializer cannot tackle this type of arrays:
```
[
  "foo": true,
  "bar": {"baz": 1},
]
```
If we can somehow group different values in different arrays of settings, we can totally switch to the JMS Serializer.

theofidry:
Also worth mentioning I'm not too happy about some parts of the configuration either. But I think it's more important to consolidate the existing, including adapting the API and the way it is consumed for it to have a clearer layer, before starting another approach to refactor it.

maks-rafalko:
> For example, as far as I am aware, JMS Serializer cannot tackle this type of arrays:
> 
> ```
> [
>   "foo": true,
>   "bar": {"baz": 1},
> ]
> ```

it's not a valid JSON

sanmai:
> it's not a valid JSON

Right. Let me check back what was the problem.

sanmai:
The problem with JMS Serializer is that we allow both object and booleans for hash values. E.g. for a mutator this is valid:

```
{
  "foo": true
}
```
As is this:
```
{
  "foo": {"bar": 5}
}
```
Stock JMS Serializer can only understand one specific type for a key.

But it seems like there could be a workaround [with a custom handler](https://stackoverflow.com/a/49722227/93540).

theofidry:
@infection/core hey team it's been a while.

To be honest I'm having a bit of trouble to see what was missing with this PR. So I've fixed some stuff and pushed the latest state hoping this could get a first review.

I've edited the description in order to clarify what this PR is about.

tests/Configuration/ConfigurationFileLoaderTest.php
```diff
@@ -0,0 +1,111 @@
...
+    private $configFactoryProphecy;
+
+    /**
+     * @var ConfigurationFileLoader
+     */
+    private $loader;
+
+    protected function setUp(): void
+    {
+        $this->schemaValidatorProphecy = $this->prophesize(SchemaValidator::class);
```
BackEndTea: I don't think we use `prophesize` in any other tests, is there a reason for not using `createMock`?

tests/Configuration/RawConfiguration/RawConfigurationTest.php
```diff
@@ -0,0 +1,179 @@
...
+        $this->assertSame(
+            $expectedException->getMessage(),
+            $exception->getMessage()
+        );
+        $this->assertSame(
+            $expectedException->getCode(),
+            $exception->getCode()
+        );
+
+        if (null === $expectedException->getPrevious()) {
```
BackEndTea: I'm really not a fan of `if `statemetns inside of tests, to assert differnt things. I think its better to split up the tests in to multiple tests.

tests/Configuration/RawConfiguration/RawConfigurationTest.php
```diff
@@ -0,0 +1,179 @@
...
+    public function test_it_cannot_retrieve_or_decode_invalid_contents(
+        string $path,
+        Exception $expectedException
+    ): void {
+        $config = new RawConfiguration($path);
+
+        try {
+            $config->getDecodedContents();
+
+            $this->fail('Expected the config contents to be invalid.');
```
BackEndTea: Why aren't we using `exceptException` here?

src/Configuration/RawConfiguration/RawConfiguration.php
```diff
@@ -33,78 +33,77 @@
...
-        return $infectionConfig;
-    }
+        if (!is_readable($this->path)) {
+            throw InvalidFile::createForFileNotReadable($this);
+        }
 
-    private function locateConfigFile(?string $customConfigPath): string
-    {
-        $configPaths = [];
+        $contents = @file_get_contents($this->path);
```
BackEndTea: Silencing errors probably isn't a good idea. Why not use the `Safe\file_get_contentes` function? Then we can remov all the earlier checks, and let that handle it for us.

src/Configuration/Schema/InvalidSchema.php
```diff
@@ -0,0 +1,67 @@
...
+ * @internal
+ */
+final class InvalidSchema extends UnexpectedValueException
+{
+    /**
+     * @param string[]
+     */
+    public static function create(RawConfiguration $config, array $errors): self
+    {
+        Assert::allString($errors);
```
BackEndTea: I'm not if i'm a fan of these asserts. I'd say the validation for this is done in our CI, as tyring to apss an array of not strings would casue PHPStan to error.

resources/schema.json
```diff
@@ -1,15 +1,40 @@
 {
     "$schema": "https://json-schema.org/draft-07/schema#",
+    "definitions": {
+        "mutator_profile": {
+            "type": "boolean"
```
BackEndTea: The mutator profile can take the same properties as a mutator. e.g. the ignore properties.

src/Configuration/Entry/Mutator/GenericMutator.php
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getIgnore(): array
+    {
+        return $this->ignore;
+    }
+
+    public function getSettings(): BCMathSettings
```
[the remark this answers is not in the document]
theofidry: Indeed, and was unused

tests/Configuration/ConfigurationFileLoaderTest.php
theofidry: Out of habit. Never really used Mock extensively and tend to use Prophecy instead. I guess I'll have to convert it here though

tests/Configuration/RawConfiguration/RawConfigurationTest.php
theofidry: I tend to agree when it's about a case where it's valid in one block and an exception in another. However here this if block is to do extra assertions on the exception thrown. I think splitting it in two test methods will increase make it a bit more confusing

tests/Configuration/RawConfiguration/RawConfigurationTest.php
theofidry: I really tend to avoid `exceptException` overall since it really makes it confusing what block is supposed to throw. It also makes harder to have assertions against the thrown exception like we have here

src/Configuration/RawConfiguration/RawConfiguration.php
theofidry: I think switching to `Safe\file_get_contents` is a good idea.

Removing the earlier checks however highly depends on how explicit the error messages are, will check.

src/Configuration/Schema/InvalidSchema.php
theofidry: It is IMO how an assertion library should be used. PHPStan will help those undesired scenario to happen, but won't provide any run-time check. The whole point (IMO) of those assert libraries is to be the last line of defence against those scenarios that should not happen and fail faster, with a helpful error message/stack trace if this scenario was ever to happen

resources/schema.json
theofidry: Yes I learned that later... I think I'm missing a rebase actually

resources/schema.json
```diff
@@ -3,13 +3,13 @@
     "type": "object",
```
theofidry: I would say the changes done here could be done in another PR

src/Command/InfectionCommand.php
```diff
@@ -255,7 +257,7 @@ protected function initialize(InputInterface $input, OutputInterface $output): v
         }
 
         $this->consoleOutput = $this->getApplication()->getConsoleOutput();
-        $this->skipCoverage = \strlen(trim($input->getOption('coverage'))) > 0;
+        $this->skipCoverage = '' !== trim((string) $input->getOption('coverage'));
```
theofidry: I changed that since I find it easier to read, but maybe you don't agree on it or it's not the convention in the project, you tell me

src/Command/InfectionCommand.php
```diff
@@ -159,8 +162,7 @@ protected function configure(): void
                 'coverage',
                 null,
                 InputOption::VALUE_REQUIRED,
-                'Path to existing coverage (`xml` and `junit` reports are required)',
-                ''
```
theofidry: There is absolutely no benefit to have the default as an empty string so I removed it

src/Command/InfectionCommand.php
```diff
@@ -368,58 +371,65 @@ private function checkMetrics(): bool
...
 
-        if ($mutators === '') {
-            $mutators = null;
+        if (null !== $minCoveredMsi && $minCoveredMsi !== (string) (float) $minCoveredMsi) {
+            throw new InvalidArgumentException(sprintf('Expected min-covered-msi to be a float. Got "%s"', $minCoveredMsi));
         }
 
+        $mutators = trim((string) $input->getOption('mutators'));
+
         $this->container = $this->getApplication()->getContainer()->withDynamicParameters(
```
theofidry: There is actually really not much changes here. The main differences are:

- The `Configuration` still returns `null` for a couple of properties: this gives greater flexibility for the future to take advantage of the `null` value to say infer a configuration (e.g. pick a default or guess a default based on other values). The only downside was that I had to add a couple of casting in some places for BC, but that's it so I think it's ok
- I simplified some of the expressions (no semantic change) e.g.:

```php
$x = $input->hasOption('foo') ? trim((string) $input->getOption('foo')) : null;

if ('' === $x) {
  $x = null;
}

// use $x
```

into:

```php
$x = trim((string) $input->getOption('foo'));

// when using $x
'' === $x ? null : $x;
```

And:

```
(bool) $input->getOption('foo')
```

I removed the casting where unnecessary

src/Command/InfectionCommand.php
```diff
@@ -368,58 +371,65 @@ private function checkMetrics(): bool
...
-    private function includeUserBootstrap(InfectionConfig $config): void
+    private function includeUserBootstrap(Configuration $config): void
     {
         $bootstrap = $config->getBootstrap();
 
-        if ($bootstrap) {
-            if (!file_exists($bootstrap)) {
-                throw FileNotFound::fromFileName($bootstrap, [__DIR__]);
-            }
+        if (null === $bootstrap) {
```
theofidry: The file existence is asserted upstream: IMO it should be done at the configuration level as part of the configuration validation rather than when using its values

src/Command/InfectionCommand.php
```diff
@@ -435,8 +445,10 @@ private function getCodeCoverageData(string $testFrameworkKey): LineCodeCoverage
 
     private function getTestFrameworkExtraOptions(string $testFrameworkKey): TestFrameworkExtraOptions
     {
-        $extraOptions = $this->input->getOption('test-framework-options')
-            ?? $this->container['infection.config']->getTestFrameworkOptions();
+        /** @var Configuration $config */
+        $config = $this->container[Configuration::class];
```
theofidry: Pretty much everywhere we are using the values from the config directly, no longer from the input (when the values _can_ come from the config)

src/Config/InfectionConfig.php
```diff
@@ -40,6 +40,8 @@
 
 /**
  * @internal
+ *
+ * @deprecated Should use {@see \Infection\Configuration\Configuration} instead
```
theofidry: Should be able to remove this class in this PR

src/Configuration/Configuration.php
```diff
@@ -0,0 +1,243 @@
...
+        bool $onlyCovered,
+        string $formatter,
+        bool $noProgress,
+        bool $ignoreMsiWithNoMutations,
+        ?float $minMsi,
+        bool $showMutations,
+        ?float $minCoveredMsi,
+        ?string $stringMutators
+    ) {
+        Assert::nullOrGreaterThanEq($timeout, 1);
```
theofidry: Those are "harsh" errors: they will end up in a straight up failure. I think however that if we want to avoid them, this should be done downstream either by validating the configuration file (which we already do, so it would highlight a missing validation rule from the `schema.json`, or the input)

src/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,109 @@
...
+
+use Infection\Configuration\Schema\SchemaConfiguration;
+use function sprintf;
+use function sys_get_temp_dir;
+use Webmozart\PathUtil\Path;
+
+/**
+ * @final
+ */
+class ConfigurationFactory
```
theofidry: This is where the magic happens. So we have:

- `$schema` which should be a 1:1 with the `infection.json.dist`
- The values coming from the input

In this class we can:

- Decide which value to take
- Infer values base on various parameters, e.g. if nothing is provided we can take some values from the Composer files
- Set default values

All of this is for the future however. For now I've sticked to the simplest way to keep BC and changes minimal

src/Configuration/Schema/SchemaConfiguration.php
```diff
@@ -0,0 +1,148 @@
...
+        ?string $tmpDir,
+        PhpUnit $phpUnit,
+        array $mutators,
+        ?string $testFramework,
+        ?string $bootstrap,
+        ?string $initialTestsPhpOptions,
+        ?string $testFrameworkOptions
+    ) {
+        Assert::nullOrGreaterThanEq($timeout, 1);
+        Assert::nullOrOneOf($testFramework, self::TEST_FRAMEWORKS);
```
theofidry: same comment as before: those guards are the last resort to prevent the system to run with invalid data. This however should be avoided, here for example by validation rules with the `schema.json`

src/Configuration/Schema/SchemaConfigurationFactory.php
```diff
@@ -0,0 +1,137 @@
...
+use Infection\Configuration\Entry\Source;
+use stdClass;
+use function trim;
+
+/**
+ * @final
+ */
+class SchemaConfigurationFactory
+{
+    public function create(string $path, stdClass $rawConfig): SchemaConfiguration
```
theofidry: This is purely parsing the raw config data into the `SchemaConfiguration`. I hope in the future we could replace this part by an automated tool, but for now I prefer to keep it simple with this manual parsing

src/Console/InfectionContainer.php
```diff
@@ -71,7 +79,7 @@
 use PhpParser\PrettyPrinter\Standard;
 use Pimple\Container;
 use SebastianBergmann\Diff\Differ as BaseDiffer;
-use Symfony\Component\Console\Exception\InvalidArgumentException;
+use function sprintf;
```
theofidry: Maybe we could invest in a tool to replace all the native calls by the `Safe\*` ones?

src/Console/InfectionContainer.php
```diff
@@ -221,103 +336,57 @@ public function withDynamicParameters(
...
-        string $initialTestsPhpOptions,
+        ?string $existingCoveragePath,
+        ?string $initialTestsPhpOptions,
         bool $ignoreMsiWithNoMutations,
-        float $minMsi,
-        float $minCoveredMsi
+        ?float $minMsi,
+        ?float $minCoveredMsi,
+        ?string $testFramework,
+        ?string $testFrameworkOptions
```
theofidry: One big change is happening here:

- dynamic parameter is about the schema file so we configure it here
- the other dynamic parameters are now about the "domain" config (`Configuration`) which we can set here
- all the other services depending of those dynamic parameters no longer do: they depend on the domain config which has the desired and ready to consume values. There is no longer a dependency on this external input or trying to sort out which value to take (from the input or the config) - which drastically simplify things here

tests/e2e/SymfonyFlex/.env
```diff
@@ -0,0 +1,14 @@
+# In all environments, the following files are loaded if they exist,
```
theofidry: For some reasons this SymfonyFlex thing keeps poping up on my machine... I would like to get it fixed in a separate PR

tests/phpunit/TestFramework/PhpUnit/Config/XmlConfigurationHelperTest.php
```diff
@@ -486,9 +489,12 @@ public function schemaProvider(): \Generator
 
     /**
      * @dataProvider schemaProvider
+     *
+     * @group integration
      */
     public function test_it_passes_validation_by_xsd(string $xsdSchema): void
     {
+        $this->markTestSkipped();
```
theofidry: to be reverted

tests/phpunit/TestFramework/PhpUnit/Config/XmlConfigurationHelperTest.php
```diff
@@ -486,9 +489,12 @@ public function schemaProvider(): \Generator
 
     /**
      * @dataProvider schemaProvider
+     *
+     * @group integration
```
theofidry: to be added in a separate PR; apply same comment as above in the phpdoc (dependency on network)

tests/phpunit/TestFramework/PhpUnit/Config/XmlConfigurationHelperTest.php
```diff
@@ -452,9 +452,12 @@ public function test_it_validates_xml_by_root_node(): void
 
     /**
      * @dataProvider schemaProvider
+     *
+     * @group integration Might require an external connection to download the XSD
      */
     public function test_it_validates_xml_by_xsd(string $xsdSchema): void
     {
+        $this->markTestSkipped();
```
theofidry: to be reverted

tests/phpunit/TestFramework/PhpUnit/Config/XmlConfigurationHelperTest.php
```diff
@@ -452,9 +452,12 @@ public function test_it_validates_xml_by_root_node(): void
 
     /**
      * @dataProvider schemaProvider
+     *
+     * @group integration Might require an external connection to download the XSD
```
theofidry: to be done in a separate PR

src/Command/InfectionCommand.php
BackEndTea: 👍 for that change

tests/phpunit/AutoReview/ProjectCode/ProjectCodeProvider.php
```diff
@@ -99,6 +117,21 @@ final class ProjectCodeProvider
         TestFrameworkTypes::class,
         MutationsCollectorVisitor::class,
         ParentConnectorVisitor::class,
+        ArrayItemRemoval::class,
```
BackEndTea: I think some of these no longer exist.

BackEndTea: I'd also really prefer we do not add to this list

src/Command/InfectionCommand.php
```diff
@@ -368,58 +371,65 @@ private function checkMetrics(): bool
 
     private function initContainer(InputInterface $input): void
     {
-        /** @var string|null $configFile */
-        $configFile = $input->hasOption('configuration')
-            ? trim((string) $input->getOption('configuration'))
-            : null
-        ;
+        // Currently the configuration is mandatory hence there is no way to
```
BackEndTea: We can probably move this comment to the doc block

src/Console/InfectionContainer.php
```diff
@@ -209,6 +229,101 @@ public static function create(): self
...
+                    $container[SchemaConfigurationFactory::class]
+                );
+            },
+            SchemaValidator::class => static function (): SchemaValidator {
+                return new SchemaValidator();
+            },
+            SchemaConfigurationFactory::class => static function (): SchemaConfigurationFactory {
+                return new SchemaConfigurationFactory();
+            },
+            ConfigurationFactory::class => static function (): ConfigurationFactory {
```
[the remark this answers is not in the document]
theofidry: For now one or the other is the same thing as the service is stateless and without dependencies. So it's a matter of preference for the time being

tests/phpunit/AutoReview/ProjectCode/ProjectCodeProvider.php
theofidry: >I think some of these no longer exist.

Indeed

>I'd also really prefer we do not add to this list

It's mostly about classes which are effectively tested by `SchemaConfigurationFactory`, it's just that you don't always have a 1:1 class unit test mapping

src/Command/InfectionCommand.php
theofidry: Why? It's only about the config file

src/Command/InfectionCommand.php
theofidry: Actually kept the check so if to be changed, it should be done in a separate PR

src/Configuration/RawConfiguration/RawConfiguration.php
```diff
@@ -33,78 +33,77 @@
...
+use function Safe\file_get_contents;
+use Seld\JsonLint\JsonParser;
+use Seld\JsonLint\ParsingException;
+use stdClass;
 
 /**
  * @internal
  */
-final class ConfigCreatorFacade
+final class RawConfiguration
```
theofidry: Hm maybe `ConfigurationFile` would be a better name

src/Config/InfectionConfig.php
theofidry: Thinking about it again: the PR is big enough as it is. It can be done later in a separate PR

theofidry:
@infection/core PR pretty much done. There is I guess 1 or 2 details to take care of, but it's ready for review

src/TestFramework/Factory.php
```diff
@@ -142,11 +145,6 @@ public function create(string $adapterName, bool $skipCoverage): AbstractTestFra
             );
         }
 
-        throw new \InvalidArgumentException(
-            sprintf(
-                'Invalid name of test framework. Available names are: %s',
-                implode(', ', [TestFrameworkTypes::PHPUNIT, TestFrameworkTypes::PHPSPEC])
-            )
-        );
+        throw new \InvalidArgumentException(sprintf('Invalid name of test framework "%s". Available names are: %s', $adapterName, implode(', ', [TestFrameworkTypes::PHPUNIT, TestFrameworkTypes::PHPSPEC])));
```
maks-rafalko: I guess you ran `make cs` before the fix was merged. Let's revert it - it's not readable now.

src/Configuration/Configuration.php
```diff
@@ -0,0 +1,243 @@
...
+        'progress',
+    ];
+
+    private $timeout;
+    private $source;
+    private $logs;
+    private $logVerbosity;
+    private $tmpDir;
+    private $phpUnit;
+    private $mutators;
```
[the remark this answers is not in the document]
theofidry: It's not even `MutatorInterface[]` at this point yet... It's gonna be once it's parsed with the string mutators but not before. It is planned parse it earlier and then have VOs here though

src/Configuration/Schema/SchemaConfigurationFileLoader.php
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Configuration\Schema;
+
+use Infection\Configuration\RawConfiguration\RawConfiguration;
+
+/**
+ * @final
```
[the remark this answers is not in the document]
theofidry: Yep

src/TestFramework/Factory.php
theofidry: arg indeed

src/Configuration/Configuration.php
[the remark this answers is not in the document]
maks-rafalko: anyway, type hint would be useful anywhere where `array` is used

src/Command/InfectionCommand.php
```diff
@@ -368,58 +371,65 @@ private function checkMetrics(): bool
...
+        $coverage = trim((string) $input->getOption('coverage'));
+        $testFramework = trim((string) $this->input->getOption('test-framework'));
+        $testFrameworkOptions = trim((string) $this->input->getOption('test-framework-options'));
+        $initialTestsPhpOptions = trim((string) $input->getOption('initial-tests-php-options'));
+
+        $minMsi = $input->getOption('min-msi');
 
-        if ($configFile === '') {
-            $configFile = null;
+        if (null !== $minMsi && $minMsi !== (string) (float) $minMsi) {
```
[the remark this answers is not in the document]
theofidry: Indeed it should

theofidry:
Alright PR ready. I hope all comments have been addressed. The CI is failing due to `InfectionCommand` being touched resulting in a decrease of the MSI

theofidry:
@infection/core I know it's a big PR but please 🙏There is a lot coming after and it's only the beginning 😄

src/Configuration/Configuration.php
```diff
@@ -0,0 +1,243 @@
...
+ */
+class Configuration
+{
+    private const LOG_VERBOSITY = [
+        'all',
+        'none',
+        'default',
+    ];
+
+    private const TEST_FRAMEWORKS = [
```
sidz: what about using `TestFrameworkTypes::ALL` or at least do:
`private const TEST_FRAMEWORKS = TestFrameworkTypes::ALL`
because we need to keep an eye at least 2 placeses when we're adding support for a new test framework.

src/Configuration/Configuration.php
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    public function getMutators(): array
+    {
+        return $this->mutators;
+    }
+
+    public function getTestFramework(): ?string
+    {
+        return $this->testFramework;
```
sidz:
```
public function getTestFramework(): string
{
    return $this->testFramework ?: TestFrameworkTypes::PHPUNIT;
}
```

^talking about this a above.

src/Configuration/ConfigurationFactory.php
```diff
@@ -0,0 +1,110 @@
...
+        if (null === $phpUnitConfigDir) {
+            $schema->getPhpUnit()->setConfigDir($configDir);
+        } elseif (!Path::isAbsolute($phpUnitConfigDir)) {
+            $schema->getPhpUnit()->setConfigDir(sprintf(
+                '%s/%s', $configDir, $phpUnitConfigDir
+            ));
+        }
+
+        return new Configuration(
+            $schema->getTimeout() ?? 10,
```
sidz: I'm voting for moving scalar to the constant with the relevant name

src/Configuration/Schema/SchemaConfiguration.php
```diff
@@ -0,0 +1,148 @@
...
+use Infection\Configuration\Entry\PhpUnit;
+use Infection\Configuration\Entry\Source;
+use Webmozart\Assert\Assert;
+
+/**
+ * @internal
+ */
+final class SchemaConfiguration
+{
+    private const TEST_FRAMEWORKS = [
```
sidz: `private const TEST_FRAMEWORKS = TestFrameworkTypes::ALL` 

p.s. third place

src/Command/InfectionCommand.php
```diff
@@ -266,12 +270,12 @@ private function startUp(): AbstractTestFrameworkAdapter
...
 
-        /** @var InfectionConfig $config */
-        $config = $this->container['infection.config'];
+        /** @var Configuration $config */
+        $config = $this->container[Configuration::class];
 
         $this->includeUserBootstrap($config);
 
-        $this->testFrameworkKey = trim((string) $this->input->getOption('test-framework') ?: $config->getTestFramework());
+        $this->testFrameworkKey = $config->getTestFramework() ?? TestFrameworkTypes::PHPUNIT;
```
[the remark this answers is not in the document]
theofidry: Hm indeed it could be moved to the factory

src/Command/InfectionCommand.php
```diff
@@ -286,17 +290,17 @@ private function startUp(): AbstractTestFrameworkAdapter
...
         $processBuilder = new InitialTestRunProcessBuilder($adapter);
 
         $initialTestsRunner = new InitialTestsRunner($processBuilder, $this->eventDispatcher);
-        $initialTestsPhpOptions = trim((string) $this->input->getOption('initial-tests-php-options') ?: $config->getInitialTestsPhpOptions());
+
         $initialTestSuitProcess = $initialTestsRunner->run(
             $this->testFrameworkOptions->getForInitialProcess(),
             $this->skipCoverage,
-            explode(' ', $initialTestsPhpOptions)
+            explode(' ', (string) $config->getInitialTestsPhpOptions())
```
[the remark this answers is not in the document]
theofidry: It's the kind of "optimization" I want to follow up afterwards. You might think that it's quite easy in this case, but turns out there is about a dozen of such cases if you start to check them. So I had to draw a line somewhere...

src/Configuration/Configuration.php
theofidry: 👍

.travis.yml
```diff
@@ -4,7 +4,7 @@ sudo: false
 
 env:
   global:
-    - INFECTION_FLAGS='--threads=4 --min-msi=67 --min-covered-msi=75 --coverage=coverage --log-verbosity=none'
+    - INFECTION_FLAGS='--threads=4 --min-msi=66 --min-covered-msi=75 --coverage=coverage --log-verbosity=none'
```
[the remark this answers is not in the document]
theofidry: It didn't actually fixed my issue so gonna revert that change

theofidry:
Thanks for the reviews!

PR #751, opened 2019-07-23
Replace usage of parameters from the container with the configuration

theofidry:
As per the description, in order to:

- reduce the number of parameters registered to the container
- avoid unnecessary redundancy


theofidry:
The failure is... well we will have to do with it I think

PR #752, opened 2019-07-24
Move the registration of the built-in services in a factory function

theofidry:
This is the first step of the container refactoring, what I would like to do next is:

- Replace the service names by their FQCN
- Expose the services by typehinted getters instead of the PSR11 container interface
- Make the container readonly - I think by decorating the pimple container

Note that **no service definition has been changed in this PR**.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -111,7 +111,7 @@ return Config::create()
         'phpdoc_align' => false,
         'phpdoc_summary' => false,
         'php_unit_set_up_tear_down_visibility' => true,
-        'php_unit_strict' => false,
+        'php_unit_strict' => true,
         'php_unit_test_annotation' => [
             'style' => 'prefix',
             'case' => 'snake',
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -111,7 +111,7 @@ return Config::create()
         'phpdoc_align' => false,
         'phpdoc_summary' => false,
         'php_unit_set_up_tear_down_visibility' => true,
-        'php_unit_strict' => true,
+        'php_unit_strict' => false,
         'php_unit_test_annotation' => [
             'style' => 'prefix',
             'case' => 'snake',
--- a/src/Console/InfectionContainer.php
+++ b/src/Console/InfectionContainer.php
@@ -80,164 +80,137 @@ use Symfony\Component\Filesystem\Filesystem;
  */
 final class InfectionContainer extends Container
 {
-    public function __construct(array $values = [])
+    public static function create(): self
     {
-        parent::__construct($values);
-
-        $this['project.dir'] = getcwd();
-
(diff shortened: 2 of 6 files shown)


src/Console/InfectionContainer.php
```diff
@@ -80,143 +80,116 @@
...
+                return new Filesystem();
+            },
+            'tmp.dir.creator' => function (): TmpDirectoryCreator {
+                return new TmpDirectoryCreator($this['filesystem']);
+            },
+            'tmp.dir' => function (): string {
+                return $this['tmp.dir.creator']->createAndGet($this->getInfectionConfig()->getTmpDir());
+            },
+            'coverage.dir.phpunit' => function () {
+                return sprintf('%s/%s', $this['coverage.path'], CodeCoverageData::PHP_UNIT_COVERAGE_DIR);
```
maks-rafalko: PHPStan does not happy with using `$this` in a static context

theofidry: And it is right, damn I wanted to do it afterwards; well I'll have to change them now:

```php
static function (Container $container) {
    return sprintf('%s/%s', $container['coverage.path'], CodeCoverageData::PHP_UNIT_COVERAGE_DIR);
}
```

.php_cs.dist
```diff
@@ -111,7 +111,7 @@ return Config::create()
         'phpdoc_align' => false,
         'phpdoc_summary' => false,
         'php_unit_set_up_tear_down_visibility' => true,
-        'php_unit_strict' => true,
+        'php_unit_strict' => false,
```
maks-rafalko: I would avoid it. There is always a possibility to rewrite tests to use `assertSame` instead of `assertEquals`. Do we need to decrease the strict level just because of 1 assertion from thousand of tests?

theofidry: >There is always a possibility to rewrite tests to use assertSame instead of assertEquals

How do you want to compare two objects by their values then? (in a sane way)

theofidry:
The failure is due to the MSI check

.php_cs.dist
maks-rafalko: [Here](https://github.com/infection/infection/pull/752/files#diff-949f564bd26d117d5334b0c8ff652cccR79)?

There are many ways to do it (including converting objects to arrays, checking objects by json schema, checking only some of the fields, etc. etc. etc.). 

For the simplicity for this particular example, I would say doing this will be enough:

```php
$this->assertTrue($container['regular service']->regular, 'Error message...');
```

theofidry:
Urg what's up with the Codecov report?

maks-rafalko:
> Urg what's up with the Codecov report?

hm.. no idea. CodeCov's comments are turned off https://github.com/infection/infection/blob/db77ba6a21cdaca78fbe09996aae63a010621911/codecov.yml#L1

PR #753, opened 2019-07-24
Rework ProjectCodeTest

theofidry:
As per an internal discussion, since the test code is not public from a library consumer PoV, it is unnecessary to mark it as `@internal`.

Before actually removing the `@internal` tag (which will be done in a separate PR for clarity), I took the opportunity to refactor a bit the `ProjectCodeTest`:

- I extracted the providers into a dedicated class in order to ease its testing, which makes `ProjectCodeTest` slimmer and less confusing (as the provider tests no longer need to be there)
- I introduced a convention `provideX(): Generator` for a regular generator and `xProvider()` for a PHPUnit provider
- a new utility function `generator_to_phpunit_data_provider` has been introduced to facilitate the transformation of a regular provider to a PHPUnit data provider
- Remove the dependency on statically declared variables for the providers
- All the code has been reviewed to:
  - Import the functions when necessary including switching to `Safe\*` functions when appropriate
  - Review all the error messages


tests/AutoReview/ProjectCode/ProjectCodeDataProvider.php
```diff
@@ -0,0 +1,86 @@
...
+    public function sourceClassesToCheckForPublicPropertiesProvider(): Generator
+    {
+        yield from generator_to_phpunit_data_provider(
+            ProjectCodeProvider::provideSourceClassesToCheckForPublicProperties()
+        );
+    }
+
+    // "testClassesProvider" would be more correct but PHPUnit will then detect this method as a
+    // test instead of a test provider.
+    public function classesTestProvider(): Generator
```
localheinz: Why not use `provide` as a prefix and then remove the prefix from the invoked methods?

That is, `classesTestProvider()` becomes `provideTestClasses()` (or even `provideTestClassNames()`), and `provideTestClasses()` becomes `testClasses()` (or even `testClassNames()`). 

What do you think?

theofidry: I feel like this convention, `testClasses()`, will very easily introduce name conflicts and the name not obvious from a regular accessor. It also makes sense to have an "action name" for an "action".

That being said, I don't feel strongly about it either so I would be fine to change it if the rest of the team feels that way too

tests/AutoReview/ProjectCode/ProjectCodeDataProvider.php
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Tests\AutoReview\ProjectCode;
+
+use Generator;
+use function Infection\Tests\generator_to_phpunit_data_provider;
+
+trait ProjectCodeDataProvider
```
localheinz: This doesn’t need to be a trait - it is possible to use `static` methods on classes as data providers.

theofidry: But it is used in two places:

- `ProjectCodeDataProviderTest` which ensures those providers are all valid
- `ProjectCodeTest` in which their values are being used

Or you mean you can do something like:

```php
/**
 * @dataProvider SomeClass::fooProvider
 */
```

?

localheinz: The latter! 👍

theofidry: Oh ok, well then I'll remove the trait in that case :)

localheinz: Apologies for the brevity, I’m on vacation and on mobile phone only. 🌞

theofidry: no worries, enjoy your vacations! :)

theofidry: Applied your change, but not sure it is more readable tbh since the data provider requires the FQCN

keradus:
I like how `AutoReview` idea is growing ;)

tests/AutoReview/ProjectCode/ProjectCodeProvider.php
```diff
@@ -0,0 +1,292 @@
...
+use function iterator_to_array;
+use ReflectionClass;
+use const SORT_STRING;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\SplFileInfo;
+
+final class ProjectCodeProvider
+{
+    /**
+     * List of classes which should not have a matching test.
```
BackEndTea: Looking at the old class i notice that this never had a propper comment.

These classes are not *intended* to not have tests. Its simply that this is somewhat 'legacy' and they aren't tested *yet*. And i think this comment should reflect that.

tests/AutoReview/ProjectCode/ProjectCodeProvider.php
```diff
@@ -0,0 +1,292 @@
...
+
+    /**
+     * This array contains all classes that can be extended by our users.
+     */
+    public const EXTENSION_POINTS = [
+        Mutator::class,
+        OutputFormatter::class,
+    ];
+
+    private static $sourceClasses;
```
BackEndTea: It would be nice for these 2 to have a doc comment, for type hinting etc.

tests/AutoReview/ProjectCode/DocBlockParserTest.php
```diff
@@ -0,0 +1,117 @@
...
+            <<<'PHP'
+/** Inlined doc-block */
+PHP
+            ,
+            'Inlined doc-block',
+        ];
+
+        yield [
+            <<<'PHP'
+// Comment
```
BackEndTea: Will these even be picked up by the parser?

This may be a weird result, but its not a use case the class needs to support anyways, i'd say remove the test here as well

theofidry: >Will these even be picked up by the parser?

Well the parser itself don't pick anything actually, it's the reflection. But to answer the question, it doesn't.

It's more to document the case and pin a behaviour than anything: i.e. later we can change the behaviour in this case and we know how it behaved before.

tests/AutoReview/ProjectCode/ProjectCodeProvider.php
theofidry: I'm open to suggestion on a wording, I would not however block the PR on that.

BackEndTea: If we somewhat mimic the `NON_FINAL_EXTENSION_CLASSES` a bit here we can do the following:
```
   /**
     * This array contains all classes that don't have tests yet, due to legacy
     * reasons. This list should never be added to, only removed from.
     */
```

tests/AutoReview/ProjectCode/ProjectCodeProvider.php
theofidry: I updated the type, as to the doc comment I think it's unwise to do so since there usage are fairly internal and documenting how a property is used is... always gonna out of sync with the underlying code resulting in a false hence potentially confusing doc. The code usage speaks for itself IMO, and it's not a case that requires to go to great length to document

PR #755, opened 2019-07-24
Add Symfony PHPUnitBridge

theofidry:
This allows to:

- Fail hard if our code is using (direct) deprecated code
- Add a coverage listener which should improve how the coverage is measured

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -67,7 +67,8 @@
     },
     "require-dev": {
         "helmich/phpunit-json-assert": "^3.0",
-        "phpunit/phpunit": "^8.2.4"
+        "phpunit/phpunit": "^8.2.4",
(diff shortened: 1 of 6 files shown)


theofidry:
What is the purpose of the `test failed` label? The CI state is now available in the PR list so it looks redundant and a bit annoying since needs to be handled manually

maks-rafalko:
> What is the purpose of the test failed label?

Very often, CI state is red because of MSI which we are not trying to increase for existing files. And I don't want to open Travis and see why the buids are failed, because of MSI or because of other issues (tests, etc.).

That's why it's easier to mark it as Tests Failed unless PR author fixes all the issues.

I've added it when someone I merged broken code to the master branch because I was too lazy to open Travis to see what is the reason of red status for the N-th time.

PR #756, opened 2019-07-25
Add a test for NoCodeCoverageException

BackEndTea:
This PR:
- [x] Adds a test for the `NoCodeCoverageException` class and reworks it to use a named constructor.

### DIFF
--- a/tests/TestFramework/PhpSpec/Config/NoCodeCoverageExceptionTest.php
+++ b/tests/TestFramework/PhpSpec/Config/NoCodeCoverageExceptionTest.php
@@ -33,19 +33,23 @@
 
 declare(strict_types=1);
 
-namespace Infection\TestFramework\PhpSpec\Config;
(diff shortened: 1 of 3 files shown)


tests/TestFramework/PhpSpec/Config/NoCodeCoverageExceptionTest.php
```diff
@@ -0,0 +1,55 @@
...
+ * @internal
+ */
+final class NoCodeCoverageExceptionTest extends TestCase
+{
+    public function test_from_test_framework(): void
+    {
+        $exception = NoCodeCoverageException::fromTestFramework('Foo');
+
+        $this->assertSame(
+            "No code coverage Extension detected for Foo. \nWithout code coverage, running Infection is not useful.",
```
theofidry: It's a case where heredoc makes more sense IMO

PR #757, opened 2019-07-25
Fix typo

theofidry:


tests/AutoReview/ProjectCode/ProjectCodeProviderTest.php
```diff
@@ -118,7 +118,7 @@ class_exists($className, true)
             || interface_exists($className, true)
             || trait_exists($className, true),
             sprintf(
-                'The "%s" class was picked up by the test files finder, but it not a class,'
+                'The "%s" class was picked up by the test files finder, but it is not a class,'
```
localheinz: Not sure, but how would the first occurrence of _class_ make sense here?

PR #759, opened 2019-07-25
Remove the service factories dependency on the input object

theofidry:


theofidry:
PR failing due to low MSI score, ready otherwise @infection/core

src/Command/InfectionCommand.php
```diff
@@ -203,26 +209,24 @@ protected function configure(): void
...
-        if (!$container->get('coverage.checker')->hasDebuggerOrCoverageOption()) {
+        if (!$this->container['coverage.checker']->hasDebuggerOrCoverageOption()) {
             $this->consoleOutput->logMissedDebuggerOrCoverageOption();
 
             return 1;
         }
 
         /** @var InfectionConfig $config */
-        $config = $container->get('infection.config');
+        $config = $this->container['infection.config'];
```
maks-rafalko: What is the benefit of this change? It's a step back from [Continainer PSR](https://www.php-fig.org/psr/psr-11/) compatibility.

Here and in other places below.

src/Command/InfectionCommand.php
```diff
@@ -86,6 +87,11 @@ final class InfectionCommand extends BaseCommand
      */
     private $skipCoverage;
 
+    /**
+     * @var InfectionContainer|null
```
maks-rafalko: Can it be really `null`?

src/Command/ConfigureCommand.php
```diff
@@ -116,7 +116,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $excludeDirsProvider = new ExcludeDirsProvider(
             $consoleHelper,
             $questionHelper,
-            $this->getContainer()->get('filesystem')
+            $this->getApplication()->getContainer()['filesystem']
```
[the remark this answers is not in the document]
theofidry: Since it's a temporary step I wouldn't bother with that tbh

src/Command/InfectionCommand.php
theofidry: Yes since it is set in `initialize()`

src/Command/InfectionCommand.php
```diff
@@ -202,26 +208,24 @@ protected function configure(): void
...
-        if (!$container->get('coverage.checker')->hasDebuggerOrCoverageOption()) {
+        if (!$this->container['coverage.checker']->hasDebuggerOrCoverageOption()) {
             $this->consoleOutput->logMissedDebuggerOrCoverageOption();
 
             return 1;
         }
 
         /** @var InfectionConfig $config */
-        $config = $container->get('infection.config');
+        $config = $this->container['infection.config'];
```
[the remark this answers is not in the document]
theofidry: >If we no longer use that container, then we can also drop the dependency

That's the plan although I don't know if it can be done in that PR yet;

>I personally prefer to use the PSR container, but i'm okay with dropping that.

Well from the internal discussions we are moving towards typehinted getters, so PSR11 doesn't provide much there; We are still using a typehint-less way here simply because it would be too big of a change

src/Command/InfectionCommand.php
theofidry: Ultimately the goal would be typehinted getters from an immutable container as per [this comment](https://github.com/infection/infection/pull/718#issuecomment-506549634).

theofidry:
To re-iterate in case this was not clear: the goal (and this PR is only a step since it would be too big of a change) is to:

- Have an immutable container
- Services registered by their typehint instead of arbitrary strings when appropriate 
- Have typehinted getters for real "public" services (we use a lot of services mostly to create other services), i.e.:

```php
class Container {
  function getFoo(): Foo;
}
```

Which avoids any phpStorm metadata tempering and is more SA friendly

src/Command/InfectionCommand.php
maks-rafalko: Still don't get how [Container Interface](https://github.com/php-fig/container/blob/014d250daebff39eba15ba990eeb2a140798e77c/src/ContainerInterface.php) conflicts with immutable container, when it's only has `get` and `has` methods.

So, for the referenced comment, we can do:

```php
public function getInfectionConfig(): InfectionConfig
{
    return $this->container->get(InfectionConfig::class);
}
```

while with this PR it is:

```php
public function getInfectionConfig(): InfectionConfig
{
    return $this->container[InfectionConfig::class];
}
```

src/Command/InfectionCommand.php
maks-rafalko: Ah, ok, then this is not because it really *can* be null in our application, but because of how symfony commands are being initialized.

In fact, Infection *can not* work without container. It's impossible.

And [this check](https://github.com/infection/infection/pull/759/files#diff-a99db3f318d0547a301f216cb03b51a7R346) is redundant:

```php
$this->container !== null && $this->container->offsetExists($testFileDataProviderServiceId)
```

because container must be initialized.

What I suggest here is to remove `|null`, remove `$this->container &&` conditions and just add `Assert::notNull($this->container)` in the beginning of the `execute()` method if it's required for static analyzers.

Each time I see `Type|null`, I can't use it without checking beforehand, and IDE complains about possible "null pointer exception".

The right way would be rewriting the process how `InfectionCommand` is initialized and pass container in `__construct()`, but this is out of the scope of this PR.

maks-rafalko:
Why do we need to drop PSR container interface? We can still have typehinted methods with it as I [commented here](https://github.com/infection/infection/pull/759#discussion_r307979327).

The fact that we can use another PSR compatible container in the future is not unreal. (e.g. to have compiled container, tags or something else). In the end, it's just 2 methods: `has` / `get`.

As for me, our container can be a child of `ContainerInterface`, or decorator.

src/Command/InfectionCommand.php
theofidry: >Still don't get how Container Interface conflicts with immutable container, when it only has get and has methods.

It doesn't, but it doesn't have typehints.

I'm confused by your example however.

What I did here is change the container we are using in the application. I.e. instead of consuming a generic PSR11 container in `InfectionCommand` and potential other commands, we use ours `InfectionContainer`, because instead of having:

```php
/** @var Foo $foo */
$foo = $container->get(Foo::class);
// Not typhint safe since there is no guarantee `$foo` is an 
// instance of `Foo` or that the docblock above is correct
```

We will have:

```php
$foo = $container->getFoo(); // return type `Foo`, so type safe
```

As of now, `InfectionContainer` _extends_ `Pimple\Container`. Since the later is mutable, I would like to move away from it. This however is for later and I'm not 100% on the specifics yet

src/Command/InfectionCommand.php
theofidry: >The right way would be rewriting the process how InfectionCommand is initialized and pass container in __construct(), but this is out of the scope of this PR.

Yes that would be the goal.

theofidry:
@maks-rafalko hope https://github.com/infection/infection/pull/759#discussion_r307980754 clarifies it.

>As for me, our container can be a child of ContainerInterface (as it is now), or decorator.

Correction: right now our container _extends_ `PimpleContainer` and we are using a PSR-11 adapter to consume a PSR-11 container in our app.

I however don't really get the point of using a PSR-11 container if we want a typesafe container though.

maks-rafalko:
> I however don't really get the point of using a PSR-11 container if we want a typesafe container though.

To make container implementation interchangeble. It does not conflict with "typesafe container", neither with immutability.

If our goal is a set of typehinted public methods with autocomplete support:

```php
class InfectionContainer
{
    public function __construct(PsrContainerInterface $container)
    {
        $this->psrContainer = $container;
    }

    public function getFoo(): Foo
    {
        return $this->psrContainer->getFoo();
    }
}
```

So, 

* we will use our `InfectionContainer` type everywhere as a type declaration.
* we can change internal implementation from Pimple (through adapter) to any other one with compile/tag support (just as an example), even our own.

`InfectionContainer` is just a read-only facade that wraps *any* PSR implementation.

--

Given the above, I don't understand why we have such update in this PR:

```diff
/** @var InfectionConfig $config */
- $config = $container->get('infection.config');
+ $config = $this->container['infection.config'];
```

It should be directly

```diff
/** @var InfectionConfig $config */
- $config = $container->get('infection.config');
+ $config = $this->container->getInfectionConfig();
```

then.

The name of this PR says only about `Input` in the factories. Does it make sense to separate these 2 completely independent things with separate PRs?

Btw, still not answered:

> Could you please explain why it is bad when factories depend on the InputInterface and what is the final goal of of this PR?

theofidry:
>To make container implementation interchangeble. It does not conflict with "typesafe container", neither with immutability.

It doesn't with immutability, it however does with:

- typesafe container, since if you use the PSR-11 one, you will rely on the typeless `get()`
- changeable container, e.g. here we have the factory `withInput()` - it requires access to `InfectionContainer`, not the PSR-11 one

So in both cases what we want to consume is the `InfectionContainer`, not the PSR11 one.

>we can change internal implementation from Pimple (through adapter) to any other one with compile/tag support (just as an example), even our own.

Yes we can have that. This change will however occur in `InfectionContainer`, internal to our container, not `InfectionCommand` where we consume the container. This however is outside the scope of this PR.

>Given the above, I don't understand why we have such update in this PR:

This is because this PR is not about introducing the typehinted getters _yet_, but removing the dependency of the factory from `$input`, which required some extra tweaking making the scope of the PR a bit more blurry, namely interact with the 

>Does it make sense to separate these 2 completely independent things with separate PRs?

Attempt in #761.

>Could you please explain why it is bad when factories depend on the InputInterface and what is the final goal of of this PR?

IMO there is several issues in that:

- You are making the container depend on IO for configuring its services
- The IO dependency introduced is based on arguments/options that may or may not be defined depending of the calling context (e.g. a new command doesn't have any of the options, resulting in services that are registered but would throw an error if attempted to be used)
- The IO input is a bad one to use: consuming it is awkward and requires a normalization step pretty much all the time

Note that for now I did `withInput()` and do those kind of normalizations in it. I'm however wondering if it would not be better to get it out so that the Container doesn't depend on `InputInterface` at all

src/Command/InfectionCommand.php
```diff
@@ -324,6 +328,61 @@ protected function initialize(InputInterface $input, OutputInterface $output): v
         $this->eventDispatcher = $this->container['dispatcher'];
     }
 
+    private function initContainer(InputInterface $input): void
```
theofidry: Disclaimer: I'm not a big fan of this code, but IMO it makes more sense that way. Indeed those parameters are defined here in `InfectionCommand`. So normalizing and interpreting those values should be the responsibility of the command, not the container.

If you are worried this makes the command needlessly bloated: don't worry, I got your back. In the future we want to make the command slimmer and, ideally, reach the point where it's only about interpreting the arguments & options and triggering the appropriate service with the right value. When this stage is reached, the command should look way slimmer than it is now. So fear not, this "bloated" state is only temporary.

src/Command/InfectionCommand.php
```diff
@@ -324,6 +328,61 @@ protected function initialize(InputInterface $input, OutputInterface $output): v
...
+            ? trim((string) $input->getOption('coverage'))
+            : ''
+        ;
+        /** @var string $initialTestsPhpOptions */
+        $initialTestsPhpOptions = trim((string) $input->getOption('initial-tests-php-options') ?: '');
+        /** @var bool $ignoreMsiWithNoMutations */
+        $ignoreMsiWithNoMutations = $input->getOption('ignore-msi-with-no-mutations');
+        $minMsi = (float) $input->getOption('min-msi');
+        $minCoveredMsi = (float) $input->getOption('min-covered-msi');
+        $mutators = $input->hasOption('mutators')
```
theofidry: As of now normalizing this input is kinda verbose. I think it would make sense to have a normalizer of some sorts around the `InputArgument` and `InputOption`, but I prefer to check that out later rather than here

src/Console/InfectionContainer.php
```diff
@@ -213,24 +212,33 @@ public static function create(): self
         ]);
     }
 
-    public function buildDynamicDependencies(InputInterface $input): void
-    {
-        $this['infection.config'] = static function (self $container) use ($input): InfectionConfig {
+    public function withDynamicParameters(
+        ?string $configFile,
```
theofidry: This is a lot of parameters... which may be why it was so conveniently hidden behind `Input`. However I think that:

- It is a hideous hack: `Input` is a horrible API to work with in general (as most UI adapters), you need to figure out the origin (argument or option), if the argument/option has a value, the type of the value, normalizing the value... It doesn't make sense to leak this API here in the container IMO
- It hides the complexity which in this case should show a deeper problem: we shouldn't have that many dynamic services. That said for now it is my opinion, I plan to look into it whether or not we can reduce that number, but it may be that I'm just overly optimistic about it as well

src/Console/InfectionContainer.php
```diff
@@ -213,24 +212,33 @@ public static function create(): self
...
+        bool $onlyCovered,
+        string $formatter,
+        bool $noProgress,
+        string $existingCoveragePath,
+        string $initialTestsPhpOptions,
+        bool $ignoreMsiWithNoMutations,
+        float $minMsi,
+        float $minCoveredMsi
+    ): self {
+        $clone = clone $this;
```
maks-rafalko: Is it kind of immutability for container or any other purpose?

src/Process/Builder/SubscriberBuilder.php
```diff
@@ -67,57 +66,29 @@
  */
 final class SubscriberBuilder
 {
-    /**
```
maks-rafalko: I would move this decision to a separate discussion / issue since we didn't agree on this topic (AFAIK).

Even with constructor type inference support from PHPStan and now PHPStorm we should agree on this topic and, probably, if it makes sense, fix it once and for all by 1 commit (or add a rule to always have typehinst, depending on a decision).

src/Console/InfectionContainer.php
theofidry: just for immutability

src/Process/Builder/SubscriberBuilder.php
theofidry: you mean just for the `@var` or something else?

maks-rafalko: yes, sorry, just for the property PHPDoc typehints:

```php
/**
 * @var MetricsCalculator
 */
```

theofidry:
Thanks for the reviews!

PR #760, opened 2019-07-27
Remove internal tags on test code

theofidry:


theofidry:
Given the diff is it could not be staled too long it would be awesome 🙏

PR #761, opened 2019-07-28
Consume directly the InfectionContainer instead of a generic PSR-11

theofidry:
After @maks-rafalko comments, here is an attempt to try to reduce the scope of #759.

The idea here is that in our application, namely the console commands `InfectionCommand`  and `ConfigureCommand` (for now), we now consume `InfectionContainer` instead of a generic PSR11 one. This will later allow us to:

- Use typehinted getters
- Use more specifics functions e.g. `withInput()` introduced in #759


theofidry:
Thanks for review

PR #766, opened 2019-08-03
PHPUnit-Bridge 4.3.3 and earlier are broken with PHPUnit 8.2.5+.

sanmai:
- PHPUnit-Bridge 4.3.3 and earlier versions are broken with PHPUnit 8.2.5+. Seems like that the reason of [current build failures](https://travis-ci.org/infection/infection/builds/565992632). Related issue: https://github.com/symfony/symfony/issues/32879 (4.3.4 is not released yet)
- PHPUnit 8+ started to change APIs rapidly, therefore versions of the same branch may be interface-incompatible. We should start E2E tests in an isolated environment.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,7 +66,7 @@
     },
     "require-dev": {
         "helmich/phpunit-json-assert": "^3.0",
-        "phpunit/phpunit": "^8.2.5",
+        "phpunit/phpunit": "^8.2.5 <8.3",
         "symfony/phpunit-bridge": "^4.3.2"
     },
     "config": {
--- a/composer.json
+++ b/composer.json
@@ -66,8 +66,8 @@
     },
     "require-dev": {
         "helmich/phpunit-json-assert": "^3.0",
-        "phpunit/phpunit": "^8.2.5 <8.3",
-        "symfony/phpunit-bridge": "^4.3.2"
+        "phpunit/phpunit": "^8.2.4 <8.2.5",
+        "symfony/phpunit-bridge": "^4.3.3"
     },
     "config": {
         "platform": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "073062ffb79e33a639656800e969c6fe",
+    "content-hash": "851ecd834be4b56fc76b77f375c4e89b",
     "packages": [
         {
             "name": "composer/xdebug-handler",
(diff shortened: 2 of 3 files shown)


sanmai:
```
There was 1 error:

1) Infection\Tests\Config\ValueProvider\TestFrameworkConfigPathProviderTest::test_it_asks_question_if_no_config_is_found_in_current_dir
TypeError: Argument 1 passed to PHPUnit\Framework\MockObject\Builder\InvocationMocker::will() must implement interface PHPUnit\Framework\MockObject\Stub, instance of PHPUnit\Framework\MockObject\Stub\ConsecutiveCalls given, called in /home/travis/build/infection/infection/tests/Config/ValueProvider/TestFrameworkConfigPathProviderTest.php on line 103
```

sanmai:
Tests are still failing, will have another look at this.

sanmai:
Awesome, [now I'm getting](https://travis-ci.org/infection/infection/jobs/571703517#L781):

>  "This build of PHPDBG does not support code coverage..."

sanmai:
I don't think I can fix that error with missing coverage support from PHPDBG. Other than that, we're ready to go.

maks-rafalko:
Thanks for looking into failing tests @sanmai.

Could you please give a few words explaining what is going on here ;) I'm wasn't be able to understand all the commits.

By the way, the last commit fails with the suspicious error:

```

         '/home/travis/.phpenv/versions/7.3.8/bin/phpdbg' '-qrr'                

         '/home/travis/.phpenv/versions/7.3.8/bin/phpunit' '--configuration'    

         '/home/travis/build/infection/infection/tests/Fixtures/e2e/SymfonyBridg

         e/./infection/phpunitConfiguration.initial.infection.xml'
```

Look at the `'/home/travis/.phpenv/versions/7.3.8/bin/phpunit' '--configuration'` - seems like this is Travis' phpunit, not the phpunit from the vendor folder

sanmai:
> seems like this is Travis' phpunit, not the phpunit from the vendor folder

That's interesting. My first guess this has something to do with symfony/phpunit-bridge.

sanmai:
I've updated the description.

sanmai:
PHPUnit-Bridge 4.3.4 still not released. Shall we wait for the release, or shall we merge this as it is?

maks-rafalko:
I'm ok with merging it now. 

One question:

> We should start E2E tests in an isolated environment.

could you please explain the purpose of adding `@runInSeparateProcess` in this PR? how and what does it solve?

sanmai:
For example, [here you can see an interface changed significantly](https://github.com/sebastianbergmann/phpunit/compare/8.2.5...8.3.0#diff-39597e2cb39fd137c7cf564b608d8295L18). If we happen to use an older version of PHPUnit and import that interface, a class that implements that interface in a different version of PHPUnit may not have the methods warranted by the interface, hence causing a failure. But if we run the tests in a separate environment, we would isolate half of the problem so that classes or interfaces imported from older versions of PHPUnit won't break things in the more current version we're using as the main version.

As an alternative we can update all PHPUnit versions in every E2E test at once, but this may not be feasible nor it is obvious. `@runInSeparateProcess` seems to me more like a viable solution in the long run, set and forget.

PR #767, opened 2019-08-06
Rework infection command

BackEndTea:
This PR:

- [x] Reworks part of the infection command
- [x] Covered by tests

The goal of this PR is to make the infection command slightly more manageable, as it is arguably the worst part of the code base atm. This doesn't completely fix this, but it will make it easier to refactor in the future.

This is done by doing two things.

First of all, the really large `execute` method now only has 5 statements.
The logic is split into 4 main chunks:

* `startUp()`
* `initialTest()`
* `infectedTest()`
* `check()`

Secondly, instead of logging a specific error message, and then returning 1, an exception is thrown, which is almost the same, but slightly easier to follow.

Some other details:
* The ProcesBuilder has been split into two classes, one for the Mutated test runs, and one for the Initial test run.
* The ConsoleOutput became a lot smaller as those messages are now handled by exceptions


src/Command/InfectionCommand.php
```diff
@@ -210,52 +219,99 @@ protected function configure(): void
     }
 
     protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $adapter = $this->startUp();
```
BackEndTea: I'm not 100% happy with these method names, but it was the best i could come up with right now, so i'm open to suggestions.

src/Command/InfectionCommand.php
```diff
@@ -210,52 +219,99 @@ protected function configure(): void
...
+        return 0;
+    }
+
+    /**
+     * Run configuration command if config does not exist
+     *
+     *
+     * @throws InfectionException
+     */
+    protected function initialize(InputInterface $input, OutputInterface $output): void
```
BackEndTea: Nothing changed for this method, but git doesn't fully understand all the movement in this file

src/Process/Builder/MutatedProcessBuilder.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    private $timeout;
+
+    public function __construct(AbstractTestFrameworkAdapter $testFrameworkAdapter, int $timeout)
+    {
+        $this->testFrameworkAdapter = $testFrameworkAdapter;
+        $this->timeout = $timeout;
+    }
+
+    public function getProcessForMutant(MutantInterface $mutant, string $testFrameworkExtraOptions = ''): MutantProcess
```
BackEndTea: This was copy pasted without change from the original.

Any changes to this method are out of scope for the PR.

BackEndTea:
90% of the changes are just moving code to a new location.

BackEndTea:
Travis failures are due to min MSI, the Appveyor build, i'm not so sure about.

src/Command/InfectionCommand.php
maks-rafalko: While I like the simplification of the main `execute()` method, I would suggest to name these methods as clear as possible, to understand from their names what they really do.

A while ago, I created such image with the mutation process to explain to other people how it really works. Probably it will give us ideas how to name these methods

infection-flow

For example, my suggestions:

* `startUp()` - not so descriptive, but at the moment I don't know a better option
* `initialTest()` - I would rename it to `runInitialTestSuite()` or similar. The verb `run` is a key here (IMO)
* `infectedTest()` - let's not use different words. We already have `mutated` like for the process `MutatedProcess`. It can be renamed to `runMutationTesting()`, indicating there is a whole process (set of tests), not just one test
* `check()` - is too common. Probably `checkMetrics()`?

src/Command/InfectionCommand.php
```diff
@@ -210,52 +219,99 @@ protected function configure(): void
...
+        $this->initialTest($adapter);
+        $this->infectedTest($adapter);
+        $this->check();
+
+        return 0;
+    }
+
+    /**
+     * Run configuration command if config does not exist
+     *
```
maks-rafalko: while we are here, could you please remove extra line?

src/Process/Builder/InitialProcessBuilder.php
```diff
@@ -36,30 +36,22 @@
...
-use Infection\Mutant\MutantInterface;
-use Infection\Process\MutantProcess;
 use Infection\TestFramework\AbstractTestFrameworkAdapter;
 use Symfony\Component\Process\Process;
 
 /**
  * @internal
  */
-class ProcessBuilder
+class InitialProcessBuilder
```
maks-rafalko: To be consistent with the `runInitialTestSuite()` method, what do you think about renaming it to `InitialTestsProcessBuilder`, or `InitialTestRunProcessBuilder` to be consisten with this class'  method name?

`InitialProcess` is too common IMO

src/Process/Builder/MutatedProcessBuilder.php
```diff
@@ -0,0 +1,78 @@
...
+
+use Infection\Mutant\MutantInterface;
+use Infection\Process\MutantProcess;
+use Infection\TestFramework\AbstractTestFrameworkAdapter;
+use Symfony\Component\Process\Process;
+
+/**
+ * @internal
+ */
+final class MutatedProcessBuilder
```
maks-rafalko: What about `MutantProcessBuilder` as we return `MutantProcess`?

theofidry:
@BackEndTea got a bit of a hectic schedule right now, but will def review that PR ASAP

maks-rafalko:
Removed and recreated Infection project on AppVayor, still same error (not related to this PR as far as I can understand)

src/Process/Builder/MutatedProcessBuilder.php
theofidry: `createForMutant()` is explicit enough IMO, or `createProcessForMutant()` if you want to to be more explicit... But I would use `create` over `get` in any case as the later is usually meant as a property accessor, not a factory

src/Process/Runner/InitialTestDidNotPass.php
```diff
@@ -0,0 +1,79 @@
...
+namespace Infection\Process\Runner;
+
+use Exception;
+use Infection\TestFramework\AbstractTestFrameworkAdapter;
+use Symfony\Component\Process\Process;
+
+/**
+ * @internal
+ */
+final class InitialTestDidNotPass extends Exception
```
theofidry: `InitialTestsDidNotPass`?

src/Process/Runner/MutatedTestDidNotPass.php
```diff
@@ -0,0 +1,67 @@
...
+namespace Infection\Process\Runner;
+
+use Exception;
+use Infection\Mutant\Exception\MsiCalculationException;
+use Infection\Mutant\MetricsCalculator;
+
+/**
+ * @internal
+ */
+final class MutatedTestDidNotPass extends Exception
```
theofidry: `Failed` instead of `DidNotPass`?

src/TestFramework/Coverage/CoverageDoesNotExistException.php
```diff
@@ -35,14 +35,18 @@
...
 
-use Infection\Console\ConsoleOutput;
 use Infection\Console\Exception\InfectionException;
 
 /**
  * @internal
  */
 final class CoverageDoesNotExistException extends InfectionException
 {
+    private const INFECTION_USAGE_SUGGESTION = '- Enable xdebug and run infection again' . "\n" .
```
theofidry: what about using an HEREDOC block here instead?

tests/Process/Builder/InitialProcessBuilderTest.php
```diff
@@ -0,0 +1,59 @@
...
+
+namespace Infection\Tests\Process\Builder;
+
+use Infection\Process\Builder\InitialProcessBuilder;
+use Infection\TestFramework\AbstractTestFrameworkAdapter;
+use PHPUnit\Framework\TestCase;
+
+final class InitialProcessBuilderTest extends TestCase
+{
+    public function test_getProcessForInitialTestRun_has_no_timeout(): void
```
theofidry: `test_it_creates_a_process_with_no_timeout`?

BackEndTea:
I've updated names according to feedback

src/Command/InfectionCommand.php
```diff
@@ -397,10 +425,15 @@ private function getCodeCoverageData(string $testFrameworkKey): LineCodeCoverage
 
     private function getTestFrameworkExtraOptions(string $testFrameworkKey): TestFrameworkExtraOptions
```
theofidry: I'm really not found of that one tbh. IMO that option should be parsed once at the beginning and re-injected elsewhere instead of relying on static state

src/Process/Builder/InitialTestRunProcessBuilder.php
```diff
@@ -36,36 +36,28 @@
...
     {
         $this->testFrameworkAdapter = $testFrameworkAdapter;
-        $this->timeout = $timeout;
     }
 
     /**
      * Creates process with enabled debugger as test framework is going to use in the code coverage.
      */
-    public function getProcessForInitialTestRun(
+    public function createProcessForInitialTestRun(
```
theofidry: `createProcess()`? the name is a bit redundant with the class name...:

```php
$initialTestRunProcessBuilder->createProcessForInitialTestRun();

// IMO this is clear enough:
$initialTestRunProcessBuilder->createProcess();
```

Not feeling strongly about it though

BackEndTea:
@infection/core This is ready for another review, i've adressed the issues that were raised. The failures on travis/appveyor look like they are related to #774

maks-rafalko:
1. appveyor build fails https://ci.appveyor.com/project/theofidry/infection/builds/27012310/job/97yci1c0o9syj5fy#L144
2. also, I'm not sure we need to throw an exception when we get MSI below some expected threshold.

Compare

* old result https://travis-ci.org/infection/infection/jobs/577748495#L850
new

* and new result (this PR) https://travis-ci.org/infection/infection/jobs/577735788#L1754
old

With the new implementation, it really looks like uncaught exception, and I believe with `-v(vv)` mode it will print the whole stack trace right before the red message, which is not what we really want.

IMO, exceptions and error messages are different things.

src/Console/ConsoleOutput.php
```diff
@@ -64,6 +68,25 @@ public function logUnknownVerbosityOption(string $default): void
...
+    public function logBadMsiErrorMessage(MetricsCalculator $metricsCalculator, float $minMsi, string $type): void
+    {
+        if (!$minMsi) {
+            throw MsiCalculationException::create('min-msi');
+        }
+
+        $this->io->error(
+            sprintf(
+                self::CI_FLAG_ERROR,
+                ($type === TestRunConstraintChecker::MSI_FAILURE ? 'MSI' : 'Covered Code MSI'),
```
maks-rafalko: looks like not needed parentheses? and on line 82

maks-rafalko:
I would like to merge it before releasing 0.14.0 because it will be difficult to support `0.14` branch without these changes, too much differences (difficult to backport bugs, etc.)

Will you be able to sort it out @BackEndTea? If not, I will look into it

BackEndTea:
I think the only blocker right now is the new lines, i'll fix that before the end of the day.

If anything else comes out of the appveyor tests, then i probably won't be able to fix it for a while. So feel free to take over after that.

BackEndTea:
@maks-rafalko The appveyor issues have been resolved, travis failures are due to min MSI

BackEndTea:
Current appveyor failure is due to not being able to connect to codecov.

maks-rafalko:
Great job @BackEndTea 

Thank you!

BackEndTea:

PR #768, opened 2019-08-09
Extract the CoverageParser interface from the XML version

BackEndTea:
This PR:

- [x] Extracts the `CoverageParser` interface from its implementation.


maks-rafalko:
Why do we need interface with the only one implementation?

theofidry:
ping @BackEndTea

theofidry:
Like @maks-rafalko, I don't feel this change is necessary (for now at least) unless there is an immediate plan to have another parser implementation

PR #769, opened 2019-08-09
Dynamically build the full mutator list

### LINKED ISSUE #420: The MutatorProfile class needs rework
Currently creating a new mutator requires a lot of boilerplate. It needs to be added tot he `MutatorProfile` in three different ways. 

This process should be simplified. We should take a look at how php-cs-fixer does things.

BackEndTea:
This PR:

- [x] Dynamically builds the full mutator list.

This means that instead of updating the list of all possible mutators, we instead create it during run time.

Fixes #420

### DIFF
--- a/src/Mutator/Util/MutatorProfile.php
+++ b/src/Mutator/Util/MutatorProfile.php
@@ -36,6 +36,8 @@ declare(strict_types=1);
 namespace Infection\Mutator\Util;
 
 use Infection\Mutator;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\SplFileInfo;
 
 /**
  * @internal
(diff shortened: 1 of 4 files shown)


src/Mutator/Util/MutatorProfile.php
```diff
@@ -259,159 +261,42 @@ final class MutatorProfile
...
-        'UnwrapTrim' => Mutator\Unwrap\UnwrapTrim::class,
-        'UnwrapUcFirst' => Mutator\Unwrap\UnwrapUcFirst::class,
-        'UnwrapUcWords' => Mutator\Unwrap\UnwrapUcWords::class,
-
-        // Extensions
-        'BCMath' => Mutator\Extensions\BCMath::class,
-        'MBString' => Mutator\Extensions\MBString::class,
-    ];
+    /**
+     * @var array<string,string>|null
```
maks-rafalko: space after comma?

tests/Mutator/Util/MutatorProfileTest.php
```diff
@@ -64,10 +64,9 @@ public function test_all_mutators_are_part_of_the_full_mutators_list(): void
 
             $this->assertArrayHasKey(
                 $class,
-                MutatorProfile::FULL_MUTATOR_LIST,
+                MutatorProfile::getFullMutatorList(),
                 sprintf(
-                    'The mutator "%s" located in "%s" has not been added to the FULL_MUTATOR_LIST in the MutatorProfile class. ' .
-                    'Please add it to ensure it can be used.',
+                    'The mutator "%s" located in "%s" has not been "getFullMutatorList" method in the MutatorProfile class. ' .
```
maks-rafalko: some word is missed in this sentence

theofidry:
I'm not a big fan to introduce an IO dependency there for that tbh... How about we auto-generate that list instead? That way we keep the list as is, without an IO dependency, but get rid of the boilertemplate.

sanmai:
> There was 1 error:
> 1) Infection\Tests\Mutator\Util\MutatorProfileTest::test_all_mutators_are_part_of_the_full_mutators_list
> sprintf(): Too few arguments
> tests/Mutator/Util/MutatorProfileTest.php:71

sanmai:
On one hand this is cool because you can take one thing off your mind. On other hand this makes impossible to check for the full list of mutators from the code. This is not as cool. So there must be some greater additional rationale for this change.

(I don't think the milliseconds the IO takes make any weather.)

theofidry:
IMO relying on a generated class would do the work:

- It is harmless and fast to always update, hence there is not risk of it getting out of sync
- It does not introduce an IO dependency on something that has no reason to depend on IO
- You don't generate something dynamically at every runtime (and IO bound dependencies are often trickier due to the nature of IO)
- You still get access to the list of all mutators

>(I don't think the milliseconds the IO takes make any weather.)

For the record: this is always assuming the IO is working fine, which is a big assumption. Typical example I had a hidden IO dependency on a class in a project, it went unnoticed because was not impacting performance in any way. A colleague on Docker (on linux!) noticed it because it was going from 1s to me to run the tests to 2min on his.

Anyway what I'm trying to say: the less you have to deal with IO, network & co the better, and I'm not convinced having a dependency here is providing any value.

sanmai:
> Anyway what I'm trying to say: the less you have to deal with IO, network & co the better, and I'm not convinced having a dependency here is providing any value.

Infection would still need to read these files to mutate the code, so they'll come from the disk cache anyways this time or the next time. There's no additional IO involved.

theofidry:
It’s not because the app is guaranteed to have an IO dependency that it is
why to spread this dependency to all of the app components. And I don’t
feel this is necessary here

Issue #503, opened 2018-10-13
Travis failures on PHPDBG and PHP 7.2


borNfreee:
This is a weird issue that I don't think I can understand myself.
From the Travis log, it appears it started failing after https://github.com/infection/infection/commit/4f364b814a5187966a71c168fb685a035ef8cf21, but I tried to restart successful builds before this commit and all of them now failing as well.

#### Example of failing build

https://travis-ci.org/infection/infection/jobs/440835337#L1044-L1062

#### What causes the failure

e2e test `Phpunit_Bat_Wrapper`

#### Why?

Ok, that's interesting. I started debugging by running `make test-e2e-phpdbg-72` - but this works without any errors. Then I ssh'd to the Travis build (using [Travis debug mode](https://docs.travis-ci.com/user/running-build-in-debug-mode/)) and found the following:

Initial Run command line:

`/home/travis/.phpenv/versions/7.2.11/bin/phpdbg -qrr /home/travis/.phpenv/versions/7.2.11/bin/phpunit --configuration /home/travis/build/infection/infection/tests/Fixtures/e2e/Phpunit_Bat_Wrapper/./infection/phpunitConfiguration.initial.infection.xml --stop-on-failure`

Note which PHPUnit Infection tries to run, it is: `/home/travis/.phpenv/versions/7.2.11/bin/phpunit `. What the heck?

When I ran this command, I got:

```
notice
[PHP Notice:  Undefined index: _SERVER in /home/travis/.phpenv/versions/7.2.11/bin/phpunit on line 18]
```

Version:

```
/home/travis/.phpenv/versions/7.2.11/bin/phpunit --version
PHPUnit 7.4.0 by Sebastian Bergmann and contributors.
```

What I understand is that Infection runs incorrect PHPUnit. Instead of running `phpunit.bat` wrapper (this is actually why this e2e test exists), it is running Travis' global PHPUnit.

Questions that I don't know answers for:

* how did it work before?
* how can Infection find `./phpunit.bat` at all?

sanmai:
The thing is TestFrameworkFinder is looking for a `phpunit` executable, and it is customary *not* to have it installed globally, that test passed fine before. 

https://github.com/infection/infection/blob/ef45b8e6f6eec00821b385ba670a4ca95c2a9d20/src/Finder/TestFrameworkFinder.php#L110-L125

Now it appears that Travis CI decided to add a globally installed `phpunit` binary (not `.phar`) and the test started to find that globally installed phpunit binary first.

[...]

PR #773, opened 2019-08-26
Remove Travis' phpunit binaries since they conflicts with vendor's phpunit

maks-rafalko:
The goal of this PR is to have green builds again. I'm fed up fighting with Travis' phpunit (see [this](https://github.com/infection/infection/issues/503), or [this](https://twitter.com/maks_rafalko/status/1082742792475107328?s=20)), so let's remove built in binaries once and for all.

By the way, this is exactly what guys from Travis suggest: https://twitter.com/travisci/status/1083116158205784065

Originally, found and reported here: https://github.com/infection/infection/pull/766#issuecomment-522059768

Fixes the build.

However, there is still one failed job for PHP 7.3 and `pcov`: https://travis-ci.org/infection/infection/jobs/577016266#L973

It tries to install PHPUnit 7.4.5 while `pcov` works with 8+ only. But this is for another free evening...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,7 @@ jobs:
       install:
         - if [ "$DEPS" == "LOW" ]; then composer update --prefer-lowest $COMPOSER_FLAGS; else composer install $COMPOSER_FLAGS; fi
       before_script:
+          - rm /home/travis/.phpenv/versions/$TRAVIS_PHP_VERSION/bin/phpunit
           - |
               if [[ "$DRIVER" == "pcov" ]]; then
                 echo > $HOME/.phpenv/versions/$TRAVIS_PHP_VERSION/etc/conf.d/xdebug.ini
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,7 @@ jobs:
       install:
         - if [ "$DEPS" == "LOW" ]; then composer update --prefer-lowest $COMPOSER_FLAGS; else composer install $COMPOSER_FLAGS; fi
       before_script:
+          - rm /home/travis/.phpenv/shims/phpunit || true
           - rm /home/travis/.phpenv/versions/$TRAVIS_PHP_VERSION/bin/phpunit
           - |
               if [[ "$DRIVER" == "pcov" ]]; then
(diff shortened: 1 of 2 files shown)


BackEndTea:
Is this an issue that can only happen to our builds, or can this also occur for projects using infection on travis?

Because if this can occur for others, then we should fix this in our code base to prevent that from happening.

maks-rafalko:
> Is this an issue that can only happen to our builds, or can this also occur for projects using infection on travis?

Well, let's see how it works now:

1. We are prependind PATH with vendor dir https://github.com/infection/infection/blob/e80f7b076018c6787ccab48818a84ff0d943f5bb/src/Finder/TestFrameworkFinder.php#L75
2. Then, we are adding names of the executable candidates (like `phpunit.bat`, `phpunit`, `phpunit.phar`, `simple-phpunit.bat`, `simple-phpunit`, `simple-phpunit.phar`)
3. Then, we run `ExecutableFinder::find(...)` for each of the candidates

What the `ExecutableFinder::find(...)` does is it tries to find the executable by provided name

* firstin the available folders in `PATH`, and 
* then in `$extraDirs`, where `$extraDirs` in our case are `[%cwd%, %cwd%/bin]`

#### Why `SymfonyBridge` e2e test fails?

Because our `$candidates` array contains the following items: [`phpunit.bat`, `phpunit`, `phpunit.phar`, `simple-phpunit.bat`, `simple-phpunit`, `simple-phpunit.phar`], and since `phpunit` is the earliest, it is found in the PATH earlier than `simple-phpunit` in `vendor/bin` folder.

So, in theory, if the project uses only symfony bridge with `simple-phpunit` executable, Infection can find global Travis' `phpunit` earlier.

You can suggest to add `simple-phpunit` earlier, before `phpunit` to the `$candidates`. But look at this [commit](https://github.com/infection/infection/commit/9431ba288631401433cd0f4cb2e5edcc98cd5765) - I tried it half a year ago, and something went wrong so I had to revert it.

Since we don't have any complaints about it, and since I don't want to waste time again and again, I won't do this re-order one more time, unless someone report an issue.

Also, the very simple fix for such (theoretic) cases would be just to set `customPath` for phpunit in `infection.json` and this will fix any potential issues for projects who uses Infection and Travis.

maks-rafalko:
Finally, all builds are green!

Thank you @BackEndTea and @sanmai

PR #775, opened 2019-08-26
Fix appveyor build

maks-rafalko:
AppVeyor is no longer lists PHP versions below `7.3.8`:

```bash
choco search php --exact --all-versions -r
php|7.3.8
```

It led to the following command fail when we pass `7.2`:

```
appveyor-retry cinst --params '""/InstallDir:C:\tools\php""' --ignore-checksums -y php --version $env:php_ver_target | sort { [version]($_ -split '\|' | select -last 1) } -Descending | Select-Object -first 1) -replace '[php|]','')
```

This command is used to install the latest version of for example `7.2.*` when we pass `7.2`, but as you can see `choco search` no longer works as it was expected for this command.

Simple solution is to just install `7.2.20` at this moment.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,10 +6,10 @@ clone_folder: c:\projects\workspace
 environment:
   matrix:
   - dependencies: highest
-    php_ver_target: 7.2
+    php_ver_target: 7.3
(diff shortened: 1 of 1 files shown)


maks-rafalko:
Thank you @BackEndTea & @sanmai

PR #779, opened 2019-09-09
Fix pre-push hook after Makefile updates

maks-rafalko:
`make test-unit` no longer runs autoreview tests, so they are separated


maks-rafalko:
Thank you @sanmai

PR #780, opened 2019-09-10
Move e2e tests to the correct place.

maks-rafalko:
Several reasons:

1. e2e tests were located in `tests/Fixtures` which does not make sense
2. PHPUnit scans *ALL* files inside `tests` folder in order to find tests and it takes up to 1 minute on my machine because of huge amount of `vendor` folders inside `tests/Fixtures/e2e/*`. I have reported a bug https://github.com/sebastianbergmann/php-file-iterator/issues/61
3. PHPStan *reads all those files* even they are excluded, it just do not show errors for them. As a bonus, `make analyze` now works much faster

This is how the structure looks like:

```bash
tests
├── add_new_e2e (file)
├── e2e_tests (file)
├── e2e
└── phpunit
```

After this change, tests start almost immediately (`vendor/bin/phpunit`)


maks-rafalko:
Thank you @BackEndTea

PR #781, opened 2019-09-10
Use xdebug-filter to reduce the time needed to collect coverage

maks-rafalko:
Everything is explained here https://thephp.cc/news/2019/01/faster-code-coverage

Time before: [6.04 minutes](https://travis-ci.org/infection/infection/jobs/583147910#L397)
Time after: [4.05 minutes](https://travis-ci.org/infection/infection/jobs/583148631#L407)

for the same job that runs `phpunit` and generates coverage.

Which is a huge win IMO

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@
 /tests/Fixtures/e2e/SymfonyFlex/bin/.phpunit/
 /vendor/
 /phpunit.xml
+/xdebug-filter.php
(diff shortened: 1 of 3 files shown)


maks-rafalko:
> Maybe we could tell our users about this as well, in order for them to speed up infection runs.

Well, I've added a tweet from Infection account about this feature and blog post: https://twitter.com/infection_php/status/1171519807167926272

I will also add this info to the blog post about 0.14.0 and, probably, to the documentation.

Thank you @BackEndTea & @localheinz

sanmai:
I'm sorry that I'm late to the party, but you don't have to generate `xdebug-filter.php` every time. It is tied to our configuration so it can be generated one, committed to the repository, and left alone.

maks-rafalko:
The only one reason I didn't do it is to not forget to update it after `phpunit.xml` is modified. But let's see..

Opened a PR https://github.com/infection/infection/pull/783

Issue #732, opened 2019-06-29
The profile keys are missing in the schema.json


theofidry:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.13.3
| Test Framework version | -
| PHP version | -
| Platform    | -
| Github Repo | -

We are defining the allowed schema for the `infection.json.dist` in the `resources/schema.json` file. I however fail to see _any_ profile keys listed [in the doc](https://infection.github.io/guide/profiles.html) in that schema file.

Is that expected?

maks-rafalko:
https://github.com/infection/infection/pull/524

> Also I didn't add validation for the mutators profiles and particular classes. For case like:

But I think we can (and should) add it, at least I see no reason why not.

theofidry:
I see thanks, will take a look at it. My main focus for the next weeks
should be the configuration :)

maks-rafalko:
Implemented

PR #782, opened 2019-09-10
Add missed profile and mutator keys to the validation schema.json

sidz:
This PR adds missed profile and mutator keys to the validation schema.json

Requested here: https://github.com/infection/infection/issues/732
Missed here: https://github.com/infection/infection/pull/524


resources/schema.json
```diff
@@ -92,6 +92,994 @@
             "type": "object",
             "description": "Contains the settings for different mutations and profiles",
             "properties": {
+                "@arithmetic": {
+                    "anyOf": [
+                        {
+                            "type": "boolean"
+                        },
+                        {
+                            "type": "object",
```
theofidry: oh, this is valid for profiles as well?

theofidry:
@sidz might be worth taking the `schema.json` from https://github.com/infection/infection/pull/750, there is a little trick there to remove the redundancy of declaring a schema over and over

sidz:
@theofidry didn't find about which trick are you talking. Could you please point what where it is?

theofidry:
Sorry I actually didn't push that part, done now: https://github.com/infection/infection/pull/750/files#diff-29ca588982da1e8d8324a8f6cf78ce34R3

sidz:
@theofidry cheers. applied

resources/schema.json
```diff
@@ -281,5 +448,39 @@
...
+            "anyOf": [
+                {
+                    "type": "boolean"
+                },
+                {
+                    "$ref": "#/definitions/ignore-option"
+                }
+            ]
+        },
+        "mutator": {
```
maks-rafalko: seems like unused and exactly the same as `mutator_profile`

sanmai:
I wonder if we should have a test for this list, to check that all mutators are indeed present.

maks-rafalko:
Thank you @sidz

PR #784, opened 2019-09-13
[RFC] Mutator: Spread operator in Array Expression - leave only the first element

maks-rafalko:
RFC: https://wiki.php.net/rfc/spread_operator_for_array
PHP 7.4+

- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/134

This mutator does the following mutations:

```diff
- $a = [...[1, 2, 3], 4];
+ $a = [[1, 2, 3][0], 4];

- $a = [...$collection, 4];
+ $a = [$collection[0], 4];

- $a = [...getCollection(), 4];
+ $a = [getCollection()[0], 4];
```

The only one issue I think will be with this mutator, is that it can work with any `\Traversable`, including `\Generator`, and the following mutation will be self-killed:

```diff
function arrGen() {
    for($i = 11; $i < 15; $i++) {
        yield $i;
    }
}

- $arr = [...arrGen()];
+ $arr = [arrGen()[0]];
```

Produces an error:

```php
Fatal error: Uncaught Error: Cannot use object of type Generator as array in ...
```

#### What can we do?

1. The first and the best solution is to exactly know what is the return type hint of `arrGen()` function and knowing that, do either `arrGen()[0]` for array or `iterator_to_array(arrGen())[0]` for iterator. However, unfortunately, Infection is not able to guess types at the moment (as for example PHPStan/Psalm do it)

2. The second option is to always do such mutation

```diff
- $arr = [...arrGen()];
+ $_traversable = arrGen();
+ $arr = [is_array($_traversable) ? $_traversable[0] : iterator_to_array($_traversable))[0];
```

3. Do not mutate variables and functions and methods, only raw arrays (the very first diff above)

Thoughts?

### DIFF
--- a/tests/phpunit/Mutator/Operator/SpreadTest.php
+++ b/tests/phpunit/Mutator/Operator/SpreadTest.php
@@ -94,21 +94,6 @@ PHP
             ,
         ];
 
-        yield 'Spread for a method call' => [
-            <<<'PHP'
-<?php
-
-$a = [...$object->getCollection(), 4];
-PHP
-            ,
-            <<<'PHP'
-<?php
-
-$a = [is_array($object->getCollection()) ? $object->getCollection()[0] : iterator_to_array($object->getCollection())[0], 4];
-PHP
-            ,
-        ];
-
         yield 'It does not mutate argument unpacking' => [
             <<<'PHP'
 <?php
--- a/tests/phpunit/Mutator/Operator/SpreadTest.php
+++ b/tests/phpunit/Mutator/Operator/SpreadTest.php
@@ -94,6 +94,21 @@ PHP
             ,
         ];
 
+        yield 'Spread for a method call' => [
+            <<<'PHP'
+<?php
+
+$a = [...$object->getCollection(), 4];
+PHP
+            ,
+            <<<'PHP'
+<?php
+
+$a = [is_array($object->getCollection()) ? $object->getCollection()[0] : iterator_to_array($object->getCollection())[0], 4];
+PHP
+            ,
+        ];
+
         yield 'It does not mutate argument unpacking' => [
             <<<'PHP'
 <?php
--- a/tests/phpunit/Mutator/Operator/SpreadTest.php
+++ b/tests/phpunit/Mutator/Operator/SpreadTest.php
@@ -45,7 +45,7 @@ final class SpreadTest extends AbstractMutatorTestCase
     public function test_mutator($input, $expected = null): void
     {
         if (version_compare(PHP_VERSION, '7.4.0') < 0) {
-            $this->markTestSkipped('Spread operator in Array Expression requires PHP 7.4+');
+//            $this->markTestSkipped('Spread operator in Array Expression requires PHP 7.4+');
         }
 
         $this->doTest($input, $expected);
@@ -63,7 +63,7 @@ PHP
             <<<'PHP'
 <?php
 
-$a = [[1, 2, 3][0], 4];
+$a = [array_slice([1, 2, 3], 0, 1)[0], 4];
 PHP
             ,
         ];
@@ -78,7 +78,7 @@ PHP
             <<<'PHP'
 <?php
 
-$a = [is_array($collection) ? reset($collection) : reset(iterator_to_array($collection)), 4];
+$a = [array_slice(is_array($collection) ? $collection : iterator_to_array($collection), 0, 1)[0], 4];
 PHP
             ,
         ];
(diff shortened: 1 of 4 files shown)


sanmai:
Great idea! I'm all in for the variant with an `is_array` check. It shall be the most useful.

maks-rafalko:
@sanmai no problem, implemented. The new mutations are:

```diff
- $a = [...[1, 2, 3], 4];
+ $a = [[1, 2, 3][0], 4];

- $a = [...$collection, 4];
+ $a = [is_array($collection) ? $collection[0] : iterator_to_array($collection)[0], 4];

- $a = [...getCollection(), 4];
+ $a = [is_array(getCollection()) ? getCollection()[0] : iterator_to_array(getCollection())[0], 4];

- $a = [...$object->getCollection(), 4];
+ $a = [is_array($object->getCollection()) ? $object->getCollection()[0] : iterator_to_array($object->getCollection())[0], 4];
```

maks-rafalko:
Seems like it's not possible to use method calls with spread operators: https://3v4l.org/51ssQ

Updated PR.

sanmai:
I wish `iterator_to_array` would take arrays, returning arrays as they were! Oh well...

It may be better to use `reset()` because [PHP will happily unroll non-zero indexed arrays.](https://3v4l.org/PcIs5)

```php
$array = [1 => 'bar','baz'];
$example = [...$array, 'foo'];
```

maks-rafalko:
makes sense, updated the code

maks-rafalko:
Unfortunately, `reset(arrGen())` [produces a notice](https://3v4l.org/pCEg0):

```bash
Notice: Only variables should be passed by reference
```

I see only one solution - introduce a variable:

```diff
- $arr = [...arrGen()];
+ $_traversable = arrGen();
+ $arr = [is_array($_traversable) ? reset($_traversable) : reset(iterator_to_array($_traversable)));
```

tests/phpunit/Mutator/Operator/SpreadTest.php
```diff
@@ -45,7 +45,7 @@ final class SpreadTest extends AbstractMutatorTestCase
     public function test_mutator($input, $expected = null): void
     {
         if (version_compare(PHP_VERSION, '7.4.0') < 0) {
-            $this->markTestSkipped('Spread operator in Array Expression requires PHP 7.4+');
+//            $this->markTestSkipped('Spread operator in Array Expression requires PHP 7.4+');
```
sanmai: I wonder if we can use [PHPUnit's annotations](https://phpunit.readthedocs.io/en/8.3/incomplete-and-skipped-tests.html#incomplete-and-skipped-tests-requires-tables-api) for this check:

```
@requires PHP >= 7.4
```

tests/phpunit/Mutator/Operator/SpreadTest.php
```diff
@@ -40,14 +40,11 @@
 final class SpreadTest extends AbstractMutatorTestCase
 {
     /**
+     * @requires PHP >= 5.3
```
sanmai: 7.4 I guess

tests/phpunit/Mutator/Operator/SpreadTest.php
maks-rafalko: yes, it works, thanks

tests/phpunit/Mutator/Operator/SpreadTest.php
maks-rafalko: my bad, updated

maks-rafalko:
Updated to use `array_slice` as discussed above. For integer keys, `array_slice` returns a new array with reset indexes, started from `0` and we safely using `array_slice(...)[0]`.

If the array has string keys, it's not a problem, because [spread operator can not work with string keys](https://3v4l.org/6b47A).

sanmai:
Yes, otherwise put you can't use a spread operator with such an array in the first place.

Do you think this can work?
```
$a = [[...$collection][0], 4];
```

maks-rafalko:
> Do you think this can work?
> 
> ```
> $a = [[...$collection][0], 4];
> ```

Not sure I got the question. If this is the original code before mutation, then the current state of PR works nice with it:

```diff
$collection = [5, 6, 7];

- $a = [[...$collection][0], 4];
+ $a = [[array_slice(is_array($collection) ? $collection : iterator_to_array($collection), 0, 1][0], 4];
```

which results to

```php
[[5][0], 4]

// and then 

[5, 4]
```

sanmai:
I'm more about this mutation:
```diff
- $a = [...$collection, 4];
+ $a = [[...$collection][0], 4];

- $a = [...getCollection(), 4];
+ $a = [[...getCollection()][0], 4];
```

If it works, we should be able to cut on `array_slice()` and so forth.

maks-rafalko:
👏 awesome

Updated the code and added more tests for `[...new ArrayIterator(['a', 'b', 'c'])]` case from RFC

maks-rafalko:
Appveyor [fail](https://ci.appveyor.com/project/borNfreee/infection/builds/27441899/job/518n0b0xe5n0920l#L145) is unrelated, see https://github.com/infection/infection/pull/786

maks-rafalko:
Thank you @sanmai for review and ideas

PR #790, opened 2019-09-17
Fix: Run 'make cs'

localheinz:
This PR

- [x] runs `make cs`


localheinz:
Thank you, @maks-rafalko and @sanmai!

PR #792, opened 2019-09-18
Refactor XmlConfigurationHelper

sanmai:
- [x]  Methods need no DOMDocument, \DOMXPath is enough
- [x]  Refactor tests

### DIFF
--- a/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
+++ b/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
@@ -101,7 +101,7 @@ class InitialConfigBuilder implements ConfigBuilder
 
         $xPath = new \DOMXPath($dom);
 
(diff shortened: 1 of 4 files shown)


src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
```diff
@@ -141,9 +141,9 @@ private function addJUnitLogger(\DOMDocument $dom, \DOMXPath $xPath): void
 
     private function addCodeCoverageLogger(\DOMDocument $dom, \DOMXPath $xPath): void
```
BackEndTea: Can we remove the $dom paramenter for this mehtod as wel?

sanmai: I'll make a PR for that because this is another review anyway.

PR #796, opened 2019-09-22
Get the first (root) directory because PHPUnit <8 and PHPUnit 8+ works differently

maks-rafalko:
PHPUnit <8: `<directory name="/absolute/path/to/root">`
PHPUnit 8+: `<directory name="/">`

Because of this behaviour, Infection `0.14.0` does not work for my old projects.


maks-rafalko:
Merging and tagging 0.14.0 as I can't sleep with broken release

PR #809, opened 2019-10-18
Added version number to ASCII banner

### LINKED ISSUE #808: Infection should emit its version when run
As per PHPUnit, it is very useful to see the version number in the output. Instead, we get a giant ASCII banner containing no useful information whatsoever. Nevertheless, this ticket is not a crusade against ASCII art, but if we are going to consume five lines where one would do, at least put something useful on one of them.

Bilge:
Closes #808.


maks-rafalko:
Can I suggest moving it one line below?

It looks prettier IMO, see

```
     ____      ____          __  _
 /___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/  0.14.0
 
      ____      ____          __  _
   _/ // / / / __/  __/ /__/ /_/ / /_/ / / / / 0.14.0
```

definitely need another pair of eyes @infection/core

theofidry:
I would prefer to stick to the Composer way there tbh:

```
   ______
  / ____/___  ____ ___  ____  ____  ________  _____
 / /   / __ \/ __ `__ \/ __ \/ __ \/ ___/ _ \/ ___/
/ /___/ /_/ / / / / / / /_/ / /_/ (__  )  __/ /
\____/\____/_/ /_/ /_/ .___/\____/____/\___/_/
                    /_/
Composer version 1.8.6 2019-06-11 15:03:05
```

Did the same in Box btw in case you are looking for an easier way to copy it: https://github.com/humbug/box/blob/master/src/Console/Application.php

BackEndTea:
I think i'd prefer it belows the asii, like @theofidry suggests

Bilge:
OK, but do you just want the version number bare or do you want some preamble like "Infection version x.y.z"?

Bilge:
Now looks something like

```
         ____      ____          __  _

Infection - PHP Mutation Testing Framework dev-master@2c915a4aa2c2b96a9ca2282254aebe8f7af29489
```

src/Console/Application.php
```diff
@@ -76,15 +74,15 @@ final class Application extends BaseApplication
      */
     private $consoleOutput;
 
-    public function __construct(InfectionContainer $container, string $name = self::NAME, string $version = self::VERSION)
+    public function __construct(InfectionContainer $container)
     {
-        parent::__construct($name, $version);
+        parent::__construct(self::NAME, Versions::getVersion('infection/infection'));
```
theofidry: to be checked in `Versions::getVersion()` works fine within the PHAR. If it does, you can also remove https://github.com/infection/infection/blob/master/box.json.dist#L4

src/Console/Application.php
```diff
@@ -53,18 +53,16 @@
...
     private const INFECTION_PREFIX = 'INFECTION';
 
-    private const LOGO = <<<'ASCII'
-     ____      ____          __  _
-    /  _/___  / __/__  _____/ /_(_)___  ____
-    / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
-  _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
- /___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/
-
-ASCII;
```
maks-rafalko: I would suggest to revert adding 4 space instead of 1, it looks strange https://travis-ci.org/infection/infection/jobs/599743124#L812

src/Console/Application.php
maks-rafalko: it works https://travis-ci.org/infection/infection/jobs/599743124#L812

src/Console/Application.php
```diff
@@ -119,24 +117,13 @@ public function run(InputInterface $input = null, OutputInterface $output = null
         return parent::run($input, $output);
     }
 
-    public function doRun(InputInterface $input, OutputInterface $output)
+    public function doRun(InputInterface $input, OutputInterface $output): int
     {
-        $output->writeln(self::LOGO);
+        $output->writeln([self::LOGO, $this->getLongVersion(), "\n"]);
```
maks-rafalko: why do we need `"\n"` here? it adds too much space before the next non-empty line: https://travis-ci.org/infection/infection/jobs/599743124#L810-L812

I would say it's not needed or just `''`

src/Console/Application.php
maks-rafalko: can we make it aligned the same way as composer?

composer_align

src/Console/Application.php
Bilge: It does in this case, but that's due to what comes after emitting a bunch of blank lines, rather than the fault of this logic. When I ran Infection locally, the order of output was not consistent: it alternated between two variants, and depending on exactly what was output, the blank line was either needed to separate the logo from the following section or it wasn't because the subsequent section emitted a bunch of blank lines as its preamble. Suffice it to say, I don't know what governs the output order or why it's not always consistent, but this problem may be better addressed elsewhere.

Bilge: Specifically, the prompt asking *"Which source directories do you want to include (comma separated)?:"* would sometimes appear before any other output. In any case, I've removed the line break as requested.

maks-rafalko: this is very strange :| Feel free to open a separate issue if this is consistently reproduced

src/Console/Application.php
```diff
@@ -119,24 +117,13 @@ public function run(InputInterface $input = null, OutputInterface $output = null
...
-    public function doRun(InputInterface $input, OutputInterface $output)
+    public function doRun(InputInterface $input, OutputInterface $output): int
     {
-        $output->writeln(self::LOGO);
+        $output->writeln([self::LOGO, $this->getLongVersion()]);
 
         return parent::doRun($input, $output);
     }
 
-    public function getLongVersion()
```
maks-rafalko: Why did we remove it?

I guess now it returns `Infection - PHP Mutation Testing Framework dev-master@a1eb3df1012a2c86c0c32de9503e03140b1b7e9d` instead of `Infection - PHP Mutation Testing Framework dev-master`?

Bilge: It was removed because it wasn't used.

Bilge:
What *should* happen when calling `-V`? Should it still output the ASCII banner or just the version number?

Bilge:
I changed the version command (`--version`/`-V`) to no longer show the ASCII banner. It is still shown for all other commands. My reasoning is that whether people wish to visually parse the version with their eyes or machine parse the version programmatically, they don't want to deal with the ASCII spam in either case. However, it still emits debug warning lines, which may also interfere with naive parsing.

image

I would suggest not emitting the debug warning with the version (`-V`) command or outputting it on STDERR rather than STDOUT to make it easier to filter, but probably either solution is out of scope for this particular PR.

maks-rafalko:
Sorry for the delay with review.

I'm quite happy with the result. Just 1 question: do we need that long hash in a version? Please compare

```bash
# before
Infection - PHP Mutation Testing Framework dev-version

# after
Infection - PHP Mutation Testing Framework dev-version@36f7fe04eca492c12ed5af23a4049c26214447f4
```

--

I agree, there is no point to display ascii logo with `-V | --version`

Bilge:
I presume the long hash is only displayed when the version isn't tagged. Most people would use a release version and thus never see it, and for those not using a tag it is probably useful to see the hash.

theofidry:
I would display the hash in any case, but only showing the 8 first characters is enough, no need for more.

Bilge:
I think this should be merged to 0.13 and 0.14 branches also, so people using older versions can actually see that they are.

maks-rafalko:
We don't support `0.13` anymore. If you want this to be in the patch `0.14.x` release, please retarget this PR to `0.14` branch.

Bilge:
I want it in both.

theofidry:
We just can not afford to maintain multiple branches, it is one of the
reasons why Infection is still not in stable release as well. So if you
want it on an older version I believe you should fork the project to merge
your change there. But I admit I don’t really see the value of keeping to
use an old version at this point

Bilge:
If you support and thus test on older versions of PHP, you will end up with
an older version of infection, because infection has already dropped
support for them.

sanmai:
It makes all kinds of sense to run Infection only under the most recent version of PHP, primarily because of the most recent language features being fully supported, and because it is super easy with Docker.

As for the older versions of PHP, not only they're not very worthy for a CI tool like Infection, but also supporting unlimited number of PHP versions requires unlimited resources, which is unaffordable for any volunteer-maintained project like Infection. Or PHPUnit, for that matter. We already have important PRs in the works for weeks and months, as that's where ours and yours invaluable time and effort will make the most impact.

Bilge:
Nobody is asking anyone to support "unlimited versions". Notwithstanding there are not unlimited versions to begin with, and thus the whole argument is predicated on nonsense, can we just merge this and then I'll backport it to 0.14 in a separate PR.

theofidry:
But what you are requesting is to allow backports to previous versions which means multiple branch supports and management. That it may not look like much work to you is one thing, but really nobody in the team is willing to put extra efforts into this. We are already overwhelmed by the sheer amount of work to do (~50 issues and ~15 PR opened on average at all time).

A good example of this is PHP-CS-Fixer: its support policy is impressive, but the result is a huge slowdown in activity despite a lot of enthusiastic contributions. While it may be annoying for this peculiar case, we ought to draw a line somewhere

maks-rafalko:
Thank you @Bilge.

sanmai:
Regarding a deleted comment: Infection is not a testing tool. If you need to verify that your program is correct as it can be, running tests under an earlier version of PHP should be enough. On the other hand, using Infection to infer *quality* of your tests under all deployed versions of PHP is a waste of resources since the only the most recent versions of Infection will give you the most accurate result. Earlier versions do not support all the new mutations, and therefore by definition are less accurate. 

Running a most recent version of Infection just once in the CI pipeline is enough to tell you everything you need to know about certain quality of your tests.

Bilge:
@sanmai Thanks, that makes sense. Although it requires more effort to limit to only the latest build, it's probably worth doing. 

Not sure any comments were deleted?

Issue #810, opened 2019-10-28
Notify user if msi is over specified min-msi


LeoVie:
**Is your feature request related to a problem? Please describe.**
I'm using infection as a composer script inside a ci pipeline.
I'm kind of hard coding the minimum msi via the `--min-msi` option.
Everytime I run infection, I have to check if the actual msi has raised over the min-msi. If so, I can raise the min-msi, too.

**Describe the solution you'd like**
It would be nice, if infection would check the min-msi and compare it with the actual msi.
If the msi is like 2% over the min-msi, it should print out a notice like "Your msi has raised. You should run infection with a higher min-msi next time." or something.

maks-rafalko:
Nice suggestion, thank you! Seems like easy to implement, feel free to contribute ;)

LeoVie:
Cool, I'll take care of it. :)

LeoVie:
Maybe we could even add an opt-in configuration option which lets infection fail, if the actual msi is much greater than the min-msi. This would force users to keep their min-msi updated.
What do you think about that?

maks-rafalko:
Do you mean a new option (`--something`) that is disabled by default? And what is "much greater"?

Looks like your build will always fail, either by actual MSI below the `--min-msi` or above ;)

BackEndTea:
> Maybe we could even add an opt-in configuration option which lets infection fail, if the actual msi is much greater than the min-msi. This would force users to keep their min-msi updated.
What do you think about that?

The notification about being able to bump your min-msi is fine. I think this is a bit overkill.

sanmai:
This and any other similar types of behavior can be emulated with a simple bash script provided Infection writes the actual MSI metrics to a file, which then can be cached at the CI side, and then parsed to use the latest metrics when running Infection.

No other extra command line option is needed here, but an additional logging format, which is much more straightforward to implement.

LeoVie:
I'd not compare by equality, but the user would define a tolerance range of 2 percent points for example.
The build would fail then, if `msi > (min-msi + 2 percent points)`.
The configuration option could be something like `"fail-if-msi-x-percent-points-over-min-msi": 2` (or a smoother name :))

This would cause infection to fail, if it gets run with a min-msi of 70%, but the actual msi is over 72%.

Through this, no manual comparing of the min-msi and the actual msi is needed, anymore. The user could be sure, that the min-msi is always more or less up to date.

Do you get my point? Or is this use case to specific?

sanmai:
[...]

PR #811, opened 2019-10-29
Add notice to console output if actual msi is higher than required msi

LeoVie:
Fixes https://github.com/infection/infection/issues/810


LeoVie:
I just added some unit tests for the newly added methods. I don't really get, why Travis fails now for some configurations.

BackEndTea:
Tests fail because the infection command got changed. Unfortunately we dont'have good (enough) tests for that in place, so the minimum MSI isn't hit when that class is changed.

This is a known issue though, and not considered blocking for this PR

tests/phpunit/Process/Runner/TestRunConstraintCheckerTest.php
```diff
@@ -202,4 +202,85 @@ public function test_run_passes_on_exactly_covered_min_msi(): void
 
         $this->assertTrue($constraintChecker->hasTestRunPassedConstraints());
     }
+
+    /** @dataProvider is_msi_over_minimum_msi_provider */
```
BackEndTea: We generally use multi line comments for dataproviders, even if they are the only comment.

tests/phpunit/Process/Runner/TestRunConstraintCheckerTest.php
```diff
@@ -202,4 +202,85 @@ public function test_run_passes_on_exactly_covered_min_msi(): void
...
+    public function test_is_msi_over_minimum_msi(?float $actualMsi, ?float $actualCoveredMsi, float $minMsi, float $minCoveredMsi, string $type, bool $expected): void
+    {
+        $metrics = $this->createMock(MetricsCalculator::class);
+
+        if ($actualMsi !== null) {
```
BackEndTea: I'm not a fan of `if` statements within a tets, maybe we can split up the test cases?

tests/phpunit/Process/Runner/TestRunConstraintCheckerTest.php
```diff
@@ -202,4 +202,85 @@ public function test_run_passes_on_exactly_covered_min_msi(): void
...
+            $metrics,
+            false,
+            $minMsi,
+            $minCoveredMsi
+        );
+
+        $this->assertSame($expected, $constraintChecker->isMsiOverMinimumMsi($type));
+    }
+
+    public function is_msi_over_minimum_msi_provider(): array
```
BackEndTea: We generally use a Generator for data providers.

Data provider also aren't snake cased, but use normal camel case

src/Command/InfectionCommand.php
```diff
@@ -360,6 +360,16 @@ private function checkMetrics(): bool
             return false;
         }
 
+        foreach ([TestRunConstraintChecker::MSI_FAILURE, ''] as $type) {
```
BackEndTea: I'd prefer the second input to be `COVERED_MSI_FAILURE ` instead of an empty string.

src/Console/ConsoleOutput.php
```diff
@@ -87,6 +88,28 @@ public function logBadMsiErrorMessage(MetricsCalculator $metricsCalculator, floa
         );
     }
 
+    public function logMinMsiCanGetIncreasedNotice(MetricsCalculator $metricsCalculator, float $minMsi, string $type): void
+    {
+        $typeString = ($type === TestRunConstraintChecker::MSI_FAILURE ? 'MSI' : 'Covered Code MSI');
```
BackEndTea: Probably cleaner to check if it is 

MSI or covered MSI, and throw an error otherwise

src/Console/ConsoleOutput.php
```diff
@@ -87,6 +88,28 @@ public function logBadMsiErrorMessage(MetricsCalculator $metricsCalculator, floa
...
+
+        $msi = ($type === TestRunConstraintChecker::MSI_FAILURE ?
```
BackEndTea: I think these parenthesis are not needed

tests/phpunit/Process/Runner/TestRunConstraintCheckerTest.php
LeoVie: Done

src/Command/InfectionCommand.php
LeoVie: You're right. I guess, I saw this in another place in the infection code. Maybe we should change this everywhere it occures.

src/Console/ConsoleOutput.php
LeoVie: Done. I also decided to get rid of the double check `$type === TestRunConstraintChecker::MSI_FAILURE`

LeoVie:
Hm, now AppVeyor fails, too. This should also not beeing caused by this pr.
Seems like something is wrong with the composer executable for the end to end tests.

LeoVie:
Ah, I see my fault. Will fix it soon.

LeoVie:
This works now.

theofidry:
Replaced by #877

LeoVie:
Oh, nice. Thank you.

PR #812, opened 2019-10-30
Ensure only unit tests are run with `make test-unit`

theofidry:
We currently have 3 PHPUnit groups: `default`, `integration` and `e2e` (which could be seen as a
sub-part of integration tests but we can't have any form of taxonomy with PHPUnit groups).

All those 3 groups were actually run when running `make test-unit` which is confusing. The following
changes ensure the `e2e` and `integration` PHPUnit groups are run only with `make test-e2e`.


theofidry:
AppVeyor failure is unrelated

Issue #774, opened 2019-08-26
Fix build for PHP 7.4 before November, 28


maks-rafalko:
Fix Travis build for PHP 7.4 before the stable release

/remind @infection/core on November, 1

maks-rafalko:
Well, currently [Infection fails](https://travis-ci.org/infection/infection/jobs/604424085#L348) on `7.4snapshot` with:

```
PHPUnit\Framework\Exception: Notice: stream_get_contents(): read of 8192 bytes failed with errno=9 Bad file descriptor in Standard input code on line 66
```

@sanmai I remember this is something you have investigated in https://github.com/sebastianbergmann/phpunit/issues/3772 and related tickets. Do you think it is related to mentioned issue?

To be honest, I didn't try anything yet (like running PHP on Docker with 7.4 RC etc.), perhaps your input would be a good start ;)

sanmai:
That issue was about PHPDBG misbehaving with `strict_types` enabled. And there's [a still-unaddressed but report](https://bugs.php.net/bug.php?id=78424).

I also remember reading that PHP 7.4 became more strict about file descriptors. Otherwise put, if this is a plain old bug, it might be that we had this bug before but it did not manifest itself in any way.

sanmai:
I can't even start testing on 7.4:
```
$ make test-unit-74
./devTools/flock devTools/*php74*.json docker run --tty --rm --volume "$PWD":/opt/infection --workdir /opt/infection infection_php74 vendor/bin/phpunit --group default
PHPUnit 8.4.2 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.4.0RC5 with Xdebug 2.8.0
Configuration: /opt/infection/phpunit.xml.dist
Random seed:   1572915764

Testing Infection Test Suite

Fatal error: Uncaught ReflectionException: Property PHPUnit\Util\Test::$annotationCache does not exist in /opt/infection/vendor/symfony/phpunit-bridge/Legacy/CoverageListenerTrait.php on line 73
```

sanmai:
And... This is a know issue: https://github.com/symfony/symfony/issues/34231

sanmai:
After #818 I can say that Infection passes 100% of the tests under PHP 7.4 locally, so the issue we're having with Travis builds might be related to either their build of PHP, or just that it is outdated.

sanmai:
[...]

PR #818, opened 2019-11-07
Fix Docker builds for PHP 7.4

sanmai:
This PR:

- [x] Fixes Makefile so that touch won't create empty files where a previous command failed without a correct exit code
- [x] Raises the open file limit for box to function correctly
- [x] Raises the PHP memory limit for the programs running inside Docker VMs

Overall this command should finish running with success, albeit after a long time:

```
make test-unit-74 test-e2e-phpdbg-74 test-infection-phpdbg-74 test-infection-xdebug-74
```

Related #774


maks-rafalko:
Thank you @sanmai

PR #820, opened 2019-11-09
Update PHP-CS-Fixer configuration, apply rules (for version 2.16.0)

sanmai:
This PR:

- [x] PHP-CS-Fixer 2.16 comes with `single_line_throw` enabled for our default profiles, which breaks the carefully crafted multi-line exception statements. Disable that rule.
- [x] Apply changes from `phpdoc_trim_consecutive_blank_line_separation`

PrettyCI uses PHP CS Fixer 2.14, but we use 2.16, which is a bummer. Is there a switch somewhere to let the CS Fixer ignore rules it does not understand yet?..

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -120,6 +120,7 @@ return Config::create()
         'static_lambda' => true,
         'strict_comparison' => true,
         'yoda_style' => false,
+        'single_line_throw' => false,
     ])
     ->setFinder($finder)
 ;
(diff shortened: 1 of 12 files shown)


maks-rafalko:
We can set fixer's version [using `.prettyci.composer.json`](https://prettyci.com/documentation#using-a-custom-standard)

sanmai:
Does not work. It still uses 2.14.0.

maks-rafalko:
Thank you @sanmai

PR #821, opened 2019-11-11
Allow text logs to be written to a PHP stream

duncan3dc:
This PR:

- [x] Adds new feature ...
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/145

When running in a CI context it's useful to display the log file contents, to avoid the time wasted running it again locally to access the details of the escaped mutants.

The simple option of just dumping `infection.log` to the console after the tests complete is not that simple, because most CI will stop once the infection command exits with non-zero status.

So to avoid complex and error prone scripts being duplicated in every project's CI, this test ensures that the text log can be configured to be written to a PHP stream.

### DIFF
--- a/src/Logger/FileLogger.php
+++ b/src/Logger/FileLogger.php
@@ -101,8 +101,17 @@ abstract class FileLogger implements MutationTestingResultsLogger
 
     public function log(): void
     {
+        $content = implode(PHP_EOL, $this->getLogLines());
+
+        // If the output should be written to a stream then just write it directly
+        if (0 === strpos($this->logFilePath, 'php://')) {
+            file_put_contents($this->logFilePath, $content);
+
+            return;
+        }
+
         try {
-            $this->fs->dumpFile($this->logFilePath, implode(PHP_EOL, $this->getLogLines()));
+            $this->fs->dumpFile($this->logFilePath, $content);
         } catch (IOException $e) {
             $this->output->writeln(sprintf('<error>%s</error>', $e->getMessage()));
         }
--- a/src/Logger/FileLogger.php
+++ b/src/Logger/FileLogger.php
@@ -105,7 +105,12 @@ abstract class FileLogger implements MutationTestingResultsLogger
 
         // If the output should be written to a stream then just write it directly
         if (0 === strpos($this->logFilePath, 'php://')) {
-            file_put_contents($this->logFilePath, $content);
+            if (\in_array($this->logFilePath, ['php://stdout', 'php://stderr'], true)) {
+                file_put_contents($this->logFilePath, $content);
+            } else {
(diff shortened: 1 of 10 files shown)


tests/e2e/Output_Stream/.gitignore
```diff
@@ -0,0 +1 @@
+/stderr.log
```
maks-rafalko: Can we use `infection.log` as ignored globally in `tests/e2e/.gitignore`?

tests/e2e/Output_Stream/infection.json.dist
```diff
@@ -0,0 +1,12 @@
+{
+    "timeout": 10,
+    "source": {
+        "directories": [
+            "src"
+        ]
+    },
+    "logs": {
+        "text": "php://stderr"
```
maks-rafalko: Please, use `debug` logger as in other tests. It was specifically designed for e2e tests, and has more convenient output for tests.

src/Logger/FileLogger.php
```diff
@@ -101,8 +101,17 @@ public function __construct(
 
     public function log(): void
     {
+        $content = implode(PHP_EOL, $this->getLogLines());
+
+        // If the output should be written to a stream then just write it directly
+        if (0 === strpos($this->logFilePath, 'php://')) {
```
maks-rafalko: I would restrict it to `stderr` and `stdout` only, because it doesn't make sense to use e.g. `php://memory` and [other available wrappers](https://www.php.net/manual/en/wrappers.php.php)

I would also suggest to unit test this restriction. We can not throw an exception here, but log with `error` as it's done on the [116 line below](https://github.com/infection/infection/pull/821/files#diff-db63c624738b8c5979eeb77e9fc84314R116)

tests/e2e/Output_Stream/.gitignore
duncan3dc: I wasn't to use a non-standard filename for this test, to ensure that we weren't getting `infection.log` written by something else (other than the output redirect). Just to make the test a little more robust against regressions.

Happy to change this to `infection.log`, delete the others, and add a more general `.gitignore` if you prefer?

maks-rafalko: I don't think "something else" can write that file, since it's being removed right before the infection run, and no other loggers are configured. So I would remove it to make it a little bit more simple and consistent

src/Logger/FileLogger.php
duncan3dc: I've re-factored that now

duncan3dc:
@maks-rafalko I've opened a docs PR and linked it above. I took a look at the tests but the style is unfamiliar so I'm going to ignore those if that's cool

maks-rafalko:
Thank you, @duncan3dc.

PR #824, opened 2019-11-12
Move tests from XmlConfigurationHelperTest which depends on the network to the integration PHPUnit group

theofidry:


maks-rafalko:
Seems like an easy change, but I'm starting loose a track. What we are using `@integration` group for?

Currently, there is only one test class marked with this group: `\Infection\Tests\Locator\RootsFileOrDirectoryLocatorTest`.

What are the criteria to mark test as `@integration`?

theofidry:
I would say anything that has an IO, DB or network dependency, ie likely to
fail if to be parallelised or because a network is down or something.

I’m not sure we are distinguishing it yet in the way we run our tests though

maks-rafalko:
Thank you @theofidry

PR #825, opened 2019-11-12
Fix CS

theofidry:
So I'm not sure why but I get those changes locally.

I'm not fond of having everything on a single line like suggested here, so if someone has an idea on what action to take to this fix it would be appreciated


maks-rafalko:
I guess we disabled this rule in https://github.com/infection/infection/pull/820 (still not merged)

theofidry:
Oh I see, I'll be closing this one in favour of #820 then

PR #837, opened 2019-11-21
According to PHP 7.4 changelog, `stream_set_option()` should always return false

maks-rafalko:
[The changelog says](https://www.php.net/manual/en/migration74.incompatible.php#migration74.incompatible.core.stream-wrappers) that `stream_set_option() ` must always return false, [where our implementation does not](https://github.com/infection/infection/blob/8b5fe199e222eff7a0fd5960815c91f825b0489f/src/StreamWrapper/IncludeInterceptor.php#L286).

_Originally posted by @sanmai in https://github.com/infection/infection/issues/774#issuecomment-554380486_


maks-rafalko:
I'm not sure as well.. Let's merge for now.

Thank you @sanmai

PR #839, opened 2019-11-22
Use constant for test framework instead of hardcoded values

maks-rafalko:
Extracted from another PR, requested in https://github.com/infection/infection/pull/800#discussion_r348343749


maks-rafalko:
Thank you @sanmai, @BackEndTea & @localheinz.

PR #840, opened 2019-11-22
Introduce TestFrameworkAdapter interface

maks-rafalko:
Extracted from https://github.com/infection/infection/pull/800 to reduce the size of the PR.

This is needed because `CodeceptionAdapter` does not extend our `AbstractTestFramework` abstract class and works differently.

More improvements in https://github.com/infection/infection/pull/800

### DIFF
--- a/src/TestFramework/TestFrameworkAdapter.php
+++ b/src/TestFramework/TestFrameworkAdapter.php
@@ -33,46 +33,38 @@
 
 declare(strict_types=1);
 
-namespace Infection\Process\Builder;
+namespace Infection\TestFramework;
 
 use Infection\Mutant\MutantInterface;
-use Infection\Process\MutantProcess;
-use Infection\TestFramework\TestFrameworkAdapter;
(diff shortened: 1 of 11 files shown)


src/TestFramework/TestFrameworkAdapter.php
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * @return string[]
+     */
+    public function getMutantCommandLine(string $configPath, string $extraOptions): array;
+
+    public function getVersion(): string;
+
+    public function getInitialTestsFailRecommendations(string $commandLine): string;
+
+    public function buildInitialConfigFile();
```
theofidry: should those (the next one as well) have a `: void` return type?

maks-rafalko: yes, thanks. But they will be removed in Codeception's PR so I will merge as is for now

maks-rafalko:
Thank you @sanmai, @BackEndTea & @theofidry.

PR #841, opened 2019-11-23
Move the parsing & generation of mutators

theofidry:
Mutators from the schema config are now generated in the configuration factory and the string mutators passed as an input from the command as well. This allows to have the parsed mutators directly exposed by the `Configuration` class.

### DIFF
--- a/devTools/phpstan-tests.neon
+++ b/devTools/phpstan-tests.neon
@@ -2,6 +2,5 @@ parameters:
     excludes_analyse:
         - %currentWorkingDirectory%/tests/e2e/*
         - %currentWorkingDirectory%/tests/phpunit/Fixtures/*
-        - %currentWorkingDirectory%/tests/phpunit/Fixtures/*
     ignoreErrors:
         - '#Return type \(void\) of method .*[Visitor|Traverser].* should be compatible with return type .* of method PhpParser\\Node.*#'
--- a/devTools/phpstan-tests.neon
+++ b/devTools/phpstan-tests.neon
@@ -2,5 +2,6 @@ parameters:
     excludes_analyse:
         - %currentWorkingDirectory%/tests/e2e/*
         - %currentWorkingDirectory%/tests/phpunit/Fixtures/*
+        - %currentWorkingDirectory%/tests/phpunit/Fixtures/*
     ignoreErrors:
         - '#Return type \(void\) of method .*[Visitor|Traverser].* should be compatible with return type .* of method PhpParser\\Node.*#'
--- a/src/Configuration/ConfigurationFactory.php
+++ b/src/Configuration/ConfigurationFactory.php
@@ -36,6 +36,8 @@ declare(strict_types=1);
 namespace Infection\Configuration;
 
 use Infection\Configuration\Schema\SchemaConfiguration;
+use Infection\Mutator\Util\MutatorParser;
+use Infection\Mutator\Util\MutatorsGenerator;
 use function sprintf;
 use function sys_get_temp_dir;
 use Webmozart\PathUtil\Path;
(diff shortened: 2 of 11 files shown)


tests/phpunit/Configuration/ConfigurationFactoryTest.php
```diff
@@ -793,7 +844,89 @@ private static function createValueForInitialTestsFrameworkOptions(
...
+            null,
+            false,
             null,
         ];
     }
+
+    /**
+     * @return array<string, Mutator>
+     */
+    private static function getDefaultMutators(): array
```
theofidry: not found of this but could find a better way for now. Indeed here we just want the default mutators listed and since it is a bit expansive to produce, I'm storing it in a static class. However right now providers are evaluated before anything else so this cannot be moved to `setUpBeforeClass()`. I can however unset it afterwards

tests/phpunit/Configuration/ConfigurationAssertions.php
```diff
@@ -94,7 +93,7 @@ private function assertConfigurationStateIs(
             $expectedPhpUnit->getConfigDir(),
             $expectedPhpUnit->getCustomPath()
         );
-        $this->assertSame($expectedMutators, $configuration->getMutators());
+        $this->assertEqualsWithDelta($expectedMutators, $configuration->getMutators(), 10.);
```
theofidry: Not cool with that but the only alternative is allowing `assertEquals()`. `assertSame()` cannot work here since the mutators are different.

Another alternative is to have a dedicated assertion for those, but it's a lot of work just for that at this point so I did not pursue this idea

src/Configuration/ConfigurationFactory.php
```diff
@@ -90,7 +92,10 @@ public function create(
             $logVerbosity,
             $tmpDir,
             $schema->getPhpUnit(),
-            $schema->getMutators(),
+            (new MutatorParser(
+                $mutators,
+                (new MutatorsGenerator($schema->getMutators()))->generate()
+            ))->getMutators(),
```
maks-rafalko: As I can see, we call `new MutatorsGenerator($schema->getMutators()))->generate()` twice for some reason.

Here, and inside [`MutatorParser::php`](https://github.com/infection/infection/blob/5dffa826a07b0f2ee6f6843aea621346f5194c05/src/Mutator/Util/MutatorParser.php#L80)

It's not an issue of this PR of course, but do you think this is something we should look at? 

From what I understand, we can first merge `$schema->getMutators()` and values from `--mutators`, and only then pass them to `(new MutatorsGenerator($mergedMutators))->generate()`

theofidry: That is a good point. I think I'm gonna focus on this mutators generation very soon; I would say right now it's not a performance issue big enough to be a blocker and that's how we were doing it before already as well.

So while I think we should go through, it's definitely something I/we should be aware of when touching that part again

maks-rafalko:
Please fix the phpstan errors https://travis-ci.org/infection/infection/jobs/616012390#L542

src/Configuration/ConfigurationFactory.php
maks-rafalko: not a blocker at all, extracted to https://github.com/infection/infection/issues/844

theofidry:
Failures are related to the MSI score check

devTools/phpstan-tests.neon
```diff
@@ -2,5 +2,6 @@ parameters:
     excludes_analyse:
         - %currentWorkingDirectory%/tests/e2e/*
         - %currentWorkingDirectory%/tests/phpunit/Fixtures/*
+        - %currentWorkingDirectory%/tests/phpunit/Fixtures/*
```
maks-rafalko: This line is a duplicate of line 4

PR #842, opened 2019-11-23
Enable Symfony 5 components

edorosh:
This PR:

- [x] Adds support of new Symfony 5 Components.  #SymfonyHackday
- Not covered by tests. Let's see what CI shows

There 2 classes using deprecated (and removed in Symfony5) feature like [`Process::inheritEnvironmentVariables()`](https://github.com/symfony/process/blob/master/CHANGELOG.md)

* Infection\Process\Builder\MutantProcessBuilder
* Infection\Process\Builder\InitialTestRunProcessBuilder

That could be tricky to get rid of and support earlier versions of `symfony/process`. Here is the [source code for 4.4](https://github.com/symfony/process/blob/4.4/Process.php#L1222)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,11 +53,11 @@
         "pimple/pimple": "^3.2",
         "sebastian/diff": "^3.0.2",
         "seld/jsonlint": "^1.7",
-        "symfony/console": "^3.4.29 || ^4.0",
-        "symfony/filesystem": "^3.4.29 || ^4.0",
-        "symfony/finder": "^3.4.29 || ^4.0",
-        "symfony/process": "^3.4.29 || ^4.0",
-        "symfony/yaml": "^3.4.29 || ^4.0",
+        "symfony/console": "^3.4.29 || ^4.0 || ^5.0",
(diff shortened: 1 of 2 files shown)


maks-rafalko:
Thanks for creating this PR @edorosh, nice to meet you here LOL ;)

> There 2 classes using deprecated (and removed in Symfony5) feature like Process::inheritEnvironmentVariables()

Just opened a PR to fix it https://github.com/infection/infection/pull/843

maks-rafalko:
> That could be tricky to get rid of and support earlier versions of symfony/process

We are still supporting symfony `3.4.x` (LTS) versions of components, and to avoid BC breaks, can't remove this method call, so I just wrapped it with a check of version.

edorosh:
@maks-rafalko thanks! I was trying to find any behavioural way to figure out if you need to call that deprecated method. I mean smth like `isEnvInheritable`. But it seems Symfony didn't implement any marker to check the interface.

maks-rafalko:
Thank you @edorosh

PR #845, opened 2019-11-24
Run e2e tests with `bin/infection` by default if parameter is skipped

maks-rafalko:
2 issues are fixed here:

* there is no `build/bin/infection.phar` (incorrect path), it is now `build/infection.phar`
* when you run multiple times e2e tests locally, it is much more convenient to just run `tests/e2e_tests` and be sure that `bin/infection` (the latest changed code) is executed rather than compiled PHAR with the old code

This patch does not affect Travis, since we always pass arguments:

https://github.com/infection/infection/blob/d146b35f0d98d0b95b28f70dc9ffdc92e1fadd61/.travis.yml#L65-L69

### DIFF
--- a/tests/e2e_tests
+++ b/tests/e2e_tests
@@ -25,7 +25,7 @@ do
     then
         output="$(bash run_tests.bash)"
     else
-        output="$(bash ../standard_script.bash ${1:-build/bin/infection.phar})"
+        output="$(bash ../standard_script.bash ${1:-bin/infection})"
     fi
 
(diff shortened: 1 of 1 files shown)


theofidry:
It seems one of the e2e tests is failing. It doesn't appear to be related to the PR, but would be interesting to get it fixed as well separately

maks-rafalko:
Restarted and everything is ok.

Thank you @sanmai and @theofidry

Issue #836, opened 2019-11-20
TypeError not detected as failing test


Bilge:
| Question    | Answer
| ------------| ---------------
| Infection version | x.y.z (`infection.phar --version`)
| Test Framework version | PHPUnit x.y.z
| PHP version | 7.2.x
| Platform    | Ubuntu/MacOS
| Github Repo | -

We have observed a false positive when a mutation causes a `TypeError`. The most obvious case is when removing a cast in a file that declares `strict_types=1`. More subtly, it can also occur when a call to a parent constructor is removed as part of the method removal mutation, because a subsequent call to a getter method will now return `null` instead of the type that should have been passed in to the parent's constructor. In all cases where `TypeError` is thrown, no matter how it comes about (does not depend on a particular mutation), Infection claims the mutant was not caught. However, making exactly the same change manually (according to the diff) and rerunning tests causes our test suite to fail, so this seems to be clearly a false positive.

Unfortunately I cannot provide clear steps to reproduce this issue. We tried creating a contrived test case but failed. In practice, the test is a Symfony functional test that makes a subrequest by booting the kernel and emulating an HTTP request (but it doesn't actually despatch a real request via a web server). Ultimately this shouldn't matter, but again, I cannot reproduce it with a simple contrived case and I cannot share the code because it's owned by a company. So this is just a note in case anyone has experienced something similar and might be able to intuit where the problem lies.

sanmai:
If a code style assumes always adding a constructor, even if an empty one, and calling it from siblings, then it really looks like a false positive. Yet I don't see what we can do here other than by offering to disable the mutator in principle, or on the very line.

Bilge:
Closed via #846.

PR #846, opened 2019-11-24
Make interceptor resilient to file not found warnings

nico-incubiq:
This PR:

- [x] Covered by tests

Fixes https://github.com/infection/infection/issues/836

Warnings are converted to Errors while running tests under PHPUnit. But due to the fact the interceptor wrapper disables itself before calling fopen, and re-enables itself just after, if an error is thrown in between, the interceptor is never re-enabled.

This is actually the issue we experienced with @Bilge. Because we use Symfony, and one of the cache components attempts to read files in `./var/cache/test/pools` to determine hits and misses as part of the Kernel initialisation. Which means that very early on the interceptor gets disabled, and then mutations are never really applied to the code, so all mutants *appear* to be escaping.

--I've only added the try/finally in the `stream_open` function because I didn't have any example for the other functions, but tell me if you'd rather do the same for all functions on principle.--

I have added a try/finally to all functions disabling and re-enabling the interceptor around a system call, and added specific tests for all of them.

### DIFF
--- a/src/StreamWrapper/IncludeInterceptor.php
+++ b/src/StreamWrapper/IncludeInterceptor.php
@@ -110,7 +110,11 @@ final class IncludeInterceptor
                 }
             }
 
-            $this->fp = fopen($path, $mode, (bool) $options, $this->context);
+            if (isset($this->context)) {
+                $this->fp = fopen($path, $mode, (bool) $options, $this->context);
+            } else {
+                $this->fp = fopen($path, $mode, (bool) $options);
+            }
         } finally {
             self::enable();
         }
@@ -132,7 +136,11 @@ final class IncludeInterceptor
         self::disable();
 
         try {
-            $this->fp = opendir($path, $this->context);
+            if (isset($this->context)) {
+                $this->fp = opendir($path, $this->context);
+            } else {
+                $this->fp = opendir($path);
+            }
         } finally {
             self::enable();
         }
@@ -162,10 +170,16 @@ final class IncludeInterceptor
         $isRecursive = (bool) ($options & STREAM_MKDIR_RECURSIVE);
 
         try {
-            return mkdir($path, $mode, $isRecursive, $this->context);
+            if (isset($this->context)) {
+                $return = mkdir($path, $mode, $isRecursive, $this->context);
+            } else {
+                $return = mkdir($path, $mode, $isRecursive);
+            }
         } finally {
             self::enable();
         }
+
+        return $return;
     }
 
     public function rename($path_from, $path_to)
@@ -173,10 +187,16 @@ final class IncludeInterceptor
         self::disable();
 
         try {
-            return rename($path_from, $path_to, $this->context);
+            if (isset($this->context)) {
+                $return = rename($path_from, $path_to, $this->context);
+            } else {
+                $return = rename($path_from, $path_to);
+            }
         } finally {
             self::enable();
         }
+
+        return $return;
     }
 
     public function rmdir($path)
@@ -184,10 +204,16 @@ final class IncludeInterceptor
         self::disable();
 
         try {
-            return rmdir($path, $this->context);
+            if (isset($this->context)) {
+                $return = rmdir($path, $this->context);
+            } else {
+                $return = rmdir($path);
+            }
         } finally {
             self::enable();
         }
+
+        return $return;
     }
 
     public function stream_cast()
@@ -324,10 +350,16 @@ final class IncludeInterceptor
         self::disable();
 
         try {
-            return unlink($path, $this->context);
+            if (isset($this->context)) {
+                $return = unlink($path, $this->context);
+            } else {
+                $return = unlink($path);
+            }
         } finally {
             self::enable();
         }
+
+        return $return;
     }
 
     public function url_stat($path)
--- a/src/StreamWrapper/IncludeInterceptor.php
+++ b/src/StreamWrapper/IncludeInterceptor.php
@@ -135,12 +135,15 @@ final class IncludeInterceptor
     {
         self::disable();
 
-        if (isset($this->context)) {
-            $this->fp = opendir($path, $this->context);
-        } else {
-            $this->fp = opendir($path);
+        try {
+            if (isset($this->context)) {
+                $this->fp = opendir($path, $this->context);
+            } else {
+                $this->fp = opendir($path);
+            }
+        } finally {
+            self::enable();
         }
-        self::enable();
 
         return $this->fp !== false;
     }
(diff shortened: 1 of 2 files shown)


src/StreamWrapper/IncludeInterceptor.php
```diff
@@ -82,33 +87,43 @@ public static function enable(): void
...
+        self::$isEnabled = true;
     }
 
     public static function disable(): void
     {
         stream_wrapper_restore('file');
+        self::$isEnabled = false;
+    }
+
+    public static function isEnabled(): bool
```
BackEndTea: I'm not really a fan of adding a property, and a getter for it, just so we can check the internal workings in a test.

tests/phpunit/StreamWrapper/IncludeInterceptorTest.php
```diff
@@ -223,4 +224,20 @@ public function test_passthrough_dir_methods_pass(): void
...
     }
+
+    public function test_it_re_enables_interceptor_after_file_not_found(): void
+    {
+        IncludeInterceptor::intercept(self::$files[1], self::$files[2]);
+        IncludeInterceptor::enable();
+
+        try {
+            fopen('file-does-not-exist.txt', 'r');
+        } catch (Warning $e) {
```
BackEndTea: Would `$this->expectException(Warning::class)` do the trick here as well?

(I think phpunit 8.4 introduced a `expectWarning`, but i'm not sure what phpunit version we run atm)

nico-incubiq:
No because the test doesn't test an exception is thrown, it tests that the
interceptor is still enabled after an exception occurs

-- 

Nicolas Séverin

Senior Software Developer

World First, UK


MSc in Computer Sciences and Networks | 2014

Grenoble-INP ESISAR, France

MBA | 2015

IAE de Grenoble, France


Tel : +44 (0) 7756 734336
Mail : niiko3oo@gmail.com

Le dim. 24 nov. 2019 à 12:21, Gert de Pagter <notifications@github.com> a
écrit :

src/StreamWrapper/IncludeInterceptor.php
```diff
@@ -82,33 +87,43 @@ public static function enable(): void
...
-        }
 
-        if (isset($this->context)) {
-            $this->fp = fopen($path, $mode, (bool) $options, $this->context);
-        } else {
-            $this->fp = fopen($path, $mode, (bool) $options);
+            if (isset($this->context)) {
+                $this->fp = fopen($path, $mode, (bool) $options, $this->context);
+            } else {
+                $this->fp = fopen($path, $mode, (bool) $options);
```
Bilge: Not that it is in any way the point of this PR, but why can't this if/else branch be condensed down to just the following?

```php
$this->fp = fopen($path, $mode, (bool) $options, $this->context ?? null);
```

nico-incubiq: it could even be `fopen($path, $mode, (bool) $options, $this->context)` all the time, as `$this->context` defaults to null

src/StreamWrapper/IncludeInterceptor.php
nico-incubiq: i removed the property, i now just try to fetch a file and confirm it still intercept it. cheers

BackEndTea:
It seems passing in null explicitly, instead of passing in nothing has a different result. Could you revert that change?

src/StreamWrapper/IncludeInterceptor.php
nico-incubiq: @BackEndTea I hope that's fine by you, I applied this suggestion throughout the file as it makes the code a lot clearer to understand.

nico-incubiq:
Done

BackEndTea:
Test failures are due to the interceptor not being 'cleaned up'. That means that after every test we would also need to do a `::disable()`. 

I would also be okay with having this in the `tearDown` method, and having any errors it creates silenced. The same way as it is in the `tearDownAfterClass` method.

nico-incubiq:
Okay i've done that, interestingly it wasn't failing locally for me. I've noticed one of the tests using `chmod` was not passing on Windows, so I fixed that too.

tests/phpunit/StreamWrapper/IncludeInterceptorTest.php
```diff
@@ -223,4 +230,188 @@ public function test_passthrough_dir_methods_pass(): void
...
+            return;
+        }
+
+        $this->fail('Badly set up test, exception was not thrown');
+    }
+
+    public function test_it_re_enables_interceptor_after_file_not_found_with_stream_metadata(): void
+    {
+        if ('\\' === \DIRECTORY_SEPARATOR) {
+            $this->markTestSkipped('This test cannot be run on Windows.');
```
Bilge: If this is correct, the more canonical way to express this would be with the `@requires` annotation, e.g. `@requires osfamily Windows`. However, I suspect it is probably not correct, as most file operations are valid on Windows/NTFS, too. Not really sure what "metadata" is supposed to be testing, though.

nico-incubiq: i've copied that bit of code someplace else in the codebase. happy to ear what @BackEndTea  thinks of this, sure

nico-incubiq: `stream_metadata` is called by other system functions https://www.php.net/manual/en/streamwrapper.stream-metadata.php. but it seems to be implemented differently on Windows and doesn't emit warnings or errors in invalid situations

maks-rafalko: Seems like [the recommended way](https://phpunit.readthedocs.io/en/8.4/incomplete-and-skipped-tests.html#skipping-tests-using-requires) is using `@requires OSFAMILY Windows` as @Bilge said. Let's use it

nico-incubiq:
i'd need to do a rebase to do that, there's quite a lot of commits in my branch which aren't in 0.14 as it was based on master

tests/phpunit/StreamWrapper/IncludeInterceptorTest.php
theofidry: Might be done in a separate PR to fix it elsewhere then

theofidry: Done in https://github.com/infection/infection/pull/856

maks-rafalko:
Thank you @nico-incubiq for this contribution!

maks-rafalko:
Released: [0.14.3](https://github.com/infection/infection/releases/tag/0.14.3)

PR #847, opened 2019-11-24
Make test case classes non-final

theofidry:
I do not think it brings any benefit to mark those classes as internal. Strict reviews are enough to ensure one is not abusing of tests as extension points.

You could argue this might not be a big deal as you could set up your PHP templates in phpStorm to always include final, or that typing final is not a big deal. I however argue that this default:

- requires an extra step from the "default": you need an extra action, typing a keyword or an IDE configuration
- might confuse new contributors as it is not a common practice meaning they will likely have the (auto-review) tests failing either locally or on the CI requiring an extra action again

In other words: yes it's not that big of a deal but I feel it's a small thing that unnecessarily increase the friction when you don't have the right set-up, reflex or are new to the project

### DIFF
--- a/tests/phpunit/AutoReview/BuildConfigYmlTest.php
+++ b/tests/phpunit/AutoReview/BuildConfigYmlTest.php
@@ -42,7 +42,7 @@ use Symfony\Component\Yaml\Yaml;
 /**
  * @coversNothing
  */
-final class BuildConfigYmlTest extends TestCase
+class BuildConfigYmlTest extends TestCase
 {
     /**
      * @dataProvider providesYamlFilesForTesting
--- a/tests/phpunit/AutoReview/Makefile/MakefileTest.php
+++ b/tests/phpunit/AutoReview/Makefile/MakefileTest.php
@@ -49,7 +49,7 @@ use function substr_count;
 /**
  * @coversNothing
  */
-final class MakefileTest extends TestCase
+class MakefileTest extends TestCase
 {
     private const MAKEFILE_PATH = __DIR__ . '/../../../../Makefile';
 
--- a/tests/phpunit/AutoReview/Makefile/Parser.php
+++ b/tests/phpunit/AutoReview/Makefile/Parser.php
@@ -48,7 +48,7 @@ use function strpos;
 use function substr;
 use function trim;
 
-final class Parser
+class Parser
 {
     /**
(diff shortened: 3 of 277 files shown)


maks-rafalko:
As long as autoreview tests will ignore my default `final` keyword for tests, I'm ok with this change.

So, autoreview tests

* it should not require `final` for tests
* should not fail when there is a `final` keyword in test

theofidry:
I would prefer to have it in a consistent way but I don’t think it’s
important enough or at least I don’t feel strongly enough about it to force
it on you. So did it that way

Bilge:
I don't think the arguments are strong enough to change the style. `final` is correct and any new contributor is going to be able to handle comprehension of said keyword. This is not a beginner's tool, and modifying it, even less so.

theofidry:
I'm not saying it's incorrect, I'm saying it however brings unnecessary friction for no value since tests are strictly internal.

BackEndTea:
I'd prefer to have the tests as `final`. As it signals the intent. But then again i'd prefer if that was default behaviour, and that we would have to explicitly allow classes to be extended.

I'm against this change, but i can live with it.

theofidry:
Let's forget about it then

sanmai:
I'm aware I am late to party, but excuse me for not being able write this out before. Having a `final` keyword everywhere is a great deal because this assumes responsibility for removing this keyword if an extension of a class is needed, which in turn is a questionable thing for non-abstract classes. And this small change will be readily seen in a diff. Without `final` everywhere you always have to look up the parent class if it was made with extending in mind, which is not only usually not true, but a major pain. When you know you have a rule to have final or abstract classes only, you're saved the hassle looking things up: if something extends something without `final` in the diff, then it is definitely allowed.

theofidry:
Sure, but this is for tests which is rarely if ever an extension point. I don't think we had that much issues in that regard and I was just more about the new contributors confusion

sanmai:
How do you think if we'd add a CS-like rule to add a final keyword for everything that isn't tagged as abstract or `@internal`? Will that help new contributors?

theofidry:
if we can find an easy way then yes because it could remove the need to check it (at least for tests) in `ProjectCodeTest` and more part as a regular CS. But it looks quite hard to achieve tbh

localheinz:
How about using the `final_internal_class` fixer?

PR #848, opened 2019-11-24
Merge quality check steps together

theofidry:
I would like to simplify the "quality check" process locally and instead of having multiple commands to run:

- `make cs`
- `make analyze` which also offers the choice of the two subsets
  - `make phpstan`
  - `make validate`
- `make test-autoreview`

Instead I'm proposing `make autoreview` which does:

- `phpstan`
- `validate`
- `test-autoreview`

In other words, `make analyze`, `make cs` is kept as a separate step for now (maybe worth integrating there? I couldn't really make up my mind here). The others are **NOT** removed, simply hidden from the default help list.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,26 +49,23 @@ cs: $(PHP_CS_FIXER)
 	LC_ALL=C sort -u .gitignore -o .gitignore
 
 .PHONY: phpstan
-phpstan:  	 ## Runs PHPStan
 phpstan: vendor $(PHPSTAN)
 	$(PHPSTAN) analyse src --level=max --configuration ./devTools/phpstan-src.neon --no-interaction --no-progress
 	$(PHPSTAN) analyse tests/phpunit --level=4 --configuration ./devTools/phpstan-tests.neon --no-interaction --no-progress
 
-.PHONY: analyze
-analyze:	 ## Runs Static analyzers and various other checks
-analyze: phpstan validate
-
 .PHONY: validate
-validate:	 ## Checks that the composer.json file is valid
 validate:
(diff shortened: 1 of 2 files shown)


maks-rafalko:
It appears that `make autoreview` and `make analyze` are now partly duplicated, which does not make sense (at least for now). Do we need them both?

Also, I would leave the name `analyze` rather than `autoreview` because many projects use the same name, so it makes sense to continue using it here. Wdyt?

theofidry:
`analyze` has actually been removed here, but you can still call `phpstan` or `validate`. They are simply not exposed in the help but those targets still exists.

>Also, I would leave the name analyze rather than autoreview because many projects use this name, so it makes sense to continue using it here. Wdyt?

I personally find `autoreview` more explicit than `analyze` but if you prefer it I can change it. Out of curiosity what other projects use `analyze`?

maks-rafalko:
> Out of curiosity what other projects use `analyze`?

At the moment of writing, I was 100% sure that at least PHPStan / PHP-CS-Fixer use that, but it appears they not, can't find anything about it 🤦‍♂ 

What I see right now - I've been using `analyze` for the last 3 years at the company I'm working for, this is what affected my mind I guess. Btw, I constantly mix up `make cs` with `cs:fix` for the same reason - this is how that command is named on our projects.

TL;DR if you guys think `autoreview` is better - ok, but if you don't mind, I would leave it as `analyze` just to not think about it after switching from the job's projects.

theofidry:
I know the trouble of switching :) I'm always mixing up `master` and `develop` for my rebase & merges haha

PR #850, opened 2019-11-25
Pass the computed tmp-dir to the Configuration

theofidry:
Instead of having the configuration with the base tmp-dir, we directly pass it the final tmp-dir which removes the need to configure the final tmp-dir as a service.


tests/phpunit/Helpers.php
```diff
@@ -56,3 +64,46 @@ function generator_to_phpunit_data_provider(iterable $source): Generator
...
+
+/**
+ * Creates a temporary directory.
+ *
+ * @param string $namespace the directory path in the system's temporary directory
+ * @param string $className the name of the test class
+ *
+ * @return string The path to the created directory
+ *
+ * @TODO: extract the FS utils from Box if they are not going to be merged to Symfony
```
theofidry: I'll work on this TODO in the coming month, TLDR we'll have a richer file system class

phpunit.xml.dist
```diff
@@ -4,7 +4,6 @@
          bootstrap="./vendor/autoload.php"
          colors="true"
          executionOrder="random"
-         stopOnFailure="true"
```
maks-rafalko: revert? (done in a separate PR)

src/Utils/TmpDirectoryCreator.php
```diff
@@ -59,14 +65,25 @@ public function __construct(Filesystem $fileSystem)
...
-    public function createAndGet(string $tempDir): string
+    public function createAndGet(string $tmpDir): string
     {
-        if (!$this->path) {
-            $path = sprintf('%s/%s', $tempDir, self::BASE_DIR_NAME);
+        Assert::true(
+            Path::isAbsolute($tmpDir),
+            sprintf(
+                'Expected the temporary directory passed to be an '
+                . 'absolute path. Got "%s"',
```
maks-rafalko: do we really need concatenating here?

tests/phpunit/Configuration/ConfigurationFactoryTest.php
```diff
@@ -66,6 +74,19 @@ public static function tearDownAfterClass(): void
...
 
+    /**
+     * {@inheritdoc}
+     */
+    protected function setUp(): void
+    {
+        $this->configFactory = new ConfigurationFactory(
+            new TmpDirectoryCreator(
+                $this->createMock(Filesystem::class
+                )
```
maks-rafalko: should be on the line above

tests/phpunit/Configuration/ConfigurationFactoryTest.php
```diff
@@ -66,6 +74,19 @@ public static function tearDownAfterClass(): void
         self::$mutators = null;
     }
 
+    /**
+     * {@inheritdoc}
```
maks-rafalko: is it needed for test's `setUp()`?

tests/phpunit/Helpers.php
```diff
@@ -56,3 +64,46 @@ function generator_to_phpunit_data_provider(iterable $source): Generator
...
+    $basePath = $systemTempDir . '/' . $namespace . '/' . $shortClass;
+
+    $result = false;
+    $attempts = 0;
+
+    do {
+        $tmpDir = normalizePath($basePath . random_int(10000, 99999));
+
+        try {
+            (new Filesystem())->mkdir($tmpDir, 0777);
```
maks-rafalko: shouldn't the object instantiation be outside the loop?

tests/phpunit/Utils/TmpDirectoryCreatorIntegrationTest.php
```diff
@@ -0,0 +1,95 @@
...
+
+    protected function setUp(): void
+    {
+        $this->fileSystem = new Filesystem();
+
+        $this->tmpDirCreator = new TmpDirectoryCreator($this->fileSystem);
+
+        // Cleans up whatever was there before. Indeed upon failure PHPUnit fails to trigger the `tearDown()` method
+        // and as a result some temporary files may still remain.
+        $this->fileSystem->remove(normalizePath(realpath(sys_get_temp_dir())) . '/infection-test');
```
maks-rafalko: could you please clarify, why do we need to remove this folder if later we are using `$this->tmp` which has different real path?

src/Utils/TmpDirectoryCreator.php
theofidry: hm not sure if I did 80 out of habit, will check

tests/phpunit/Configuration/ConfigurationFactoryTest.php
theofidry: it's not needed, but I put it by habit as an inherited method. I would like us to address this convention though; opening up a ticket for it

theofidry: #860

PR #854, opened 2019-11-25
Remove min-msi check

### LINKED ISSUE #849: Move MSI score check elsewhere
At the moment we enforce a min MSI score check in all our builds which causes two problems:

- The MSI score is not necessarily calculated correctly since it depends on the changed files so a trivial change in `InfectionCommand` will very likely result in a red CI although it's not the scope of the PR to fix the whole command
- The MSI score might be inconsistent from a build to another depending of the coverage tool

I feel really gets in our way as it prevents us to view in a efficient manner what PR is really failing or not, resulting in the extra steps of checking the CI output, adding a "tests failing" label which is a flimsy thing since a label is never up to date with the content of the PR.

So instead I would suggest to have a dedicated build allowed to fail or even not enforce the MSI check for now: if we are worried about regression we can always check the PR in question before a merge

theofidry:
Closes #849

I couldn't find an easy way to restore an MSI check somewhere, if someone is willing to take up on that it would be great but meanwhile I neither find the MSI check useful or required for this PR.


maks-rafalko:
I would like to hear @infection/core input here and don't hurry up with this one. Personally, having 1 build with `--min-msi` check (with `allow_failure: true`) is better IMO than completely removing it.

Probably someone has another ideas? What about checking MSI only for the new (added) files?

```diff
- CHANGED_FILES=$(git diff origin/$TRAVIS_BRANCH --diff-filter=AM --name-only | grep src/ | paste -sd "," -);
+ CHANGED_FILES=$(git diff origin/$TRAVIS_BRANCH --diff-filter=A --name-only | grep src/ | paste -sd "," -);
```

maks-rafalko:
Well, I'm still suggesting to not remove min-msi check, but try checking only *added* files.

Thus,

* new files will be checked and no failure will be allowed (literally at all)
* existing files (like `InfectionCommand`) won't fail builds just because they were touched
* we will still at least not decrease our MSI during the time (which is not the case with removing a check)

theofidry:
@maks-rafalko sorry I included you again :) I like the suggestion, I just wished to hear the opinion of someone else as well

BackEndTea:
👍 For the suggetion by Maks, we should be strict(er) on new files, but currently we get too many build failures due hard to test/ poorly tested classes like the infection command.

theofidry:
Implemented @maks-rafalko suggestion

maks-rafalko:
`make: *** No rule to make target 'analyze'.`

https://travis-ci.org/infection/infection/jobs/618272720#L536

PR #856, opened 2019-11-26
Replace Windows platform checks by @requires tags

theofidry:
As per [@maks-rafalko comment](https://github.com/infection/infection/pull/846#discussion_r350400541), let's make use of the more conventional way

### DIFF
--- a/tests/phpunit/AutoReview/ProjectCode/ProjectCodeTest.php
+++ b/tests/phpunit/AutoReview/ProjectCode/ProjectCodeTest.php
@@ -55,12 +55,11 @@ use function Safe\sprintf;
  */
 final class ProjectCodeTest extends TestCase
 {
+    /**
+     * @requires OSFAMILY Windows Cannot check if the file is executable on Windows
+     */
     public function test_infection_bin_is_executable(): void
     {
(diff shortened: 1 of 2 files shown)


tests/phpunit/AutoReview/ProjectCode/ProjectCodeTest.php
```diff
@@ -55,12 +55,11 @@
  */
 final class ProjectCodeTest extends TestCase
 {
+    /**
+     * @requires OSFAMILY Windows Cannot check if the file is executable on Windows
```
Bilge: Don't know if it's safe to comment on the annotation line. It's probably not future-proof at the very least. Might be better to comment on preceding lines instead.

theofidry: Unless PHPUnit goes from a "non-standard" annotation system, this should work fine, at least it's confirm to the PHP annotation PSR-5 spec for what it's worth.

If it breaks in the future, we can safely update as well it's not a critical part; I personally prefer to have it on the same line as it makes it clear it belongs to this specific mention. That said I don't feel strongly about it either so if others disagree I'm happy to adjust the code

PR #865, opened 2019-11-27
Tweak CS

theofidry:
I've included all the changes to illustrate the new rules; please report anything you feel shouldn't be introduced; I'll do a reset hard & apply the new rules once we are cool with the CS


theofidry:
@infection/core will need your participation on this one :)

.php_cs.dist
```diff
@@ -121,6 +121,8 @@ return Config::create()
         'strict_comparison' => true,
         'yoda_style' => false,
         'single_line_throw' => false,
+        'compact_nullable_typehint' => true,
+        'final_static_access' => true,
```
localheinz: What do you think about keeping rules sorted by name?

.php_cs.dist
```diff
@@ -121,6 +120,33 @@ return Config::create()
         'strict_comparison' => true,
         'yoda_style' => false,
         'single_line_throw' => false,
+        'compact_nullable_typehint' => true,
```
localheinz: What do you think about keeping rules sorted by name?

.php_cs.dist
```diff
@@ -121,6 +120,33 @@ return Config::create()
...
+        'final_static_access' => true,
+        'fully_qualified_strict_types' => true,
+        'global_namespace_import' => [
```
localheinz: Do we really want to enable this fixer?

.php_cs.dist
```diff
@@ -102,9 +102,8 @@ return Config::create()
             'location' => 'after_open',
             'separate' => 'bottom',
         ],
-        'native_function_invocation' => [
-            'include' => ['@compiler_optimized'],
-        ],
+        'native_function_invocation' => false,
```
localheinz: Just as an idea: in my experience working with legacy projects I have found it a lot easier to enable fixers one-by-one:

- push a commit enabling a fixer, usually breaking the build
- push a commit applying the fixer

This has the following advantages:

- pull requests are easier to review
- discussions are only related to enabling a single fixer, not a bunch of them

What do you think?

theofidry: Of course it's easier, but it requires me to open 10 PRs each conflicting :)

.php_cs.dist
theofidry: For me yes on all levels:

- Importing a symbol is a good practice, instead of relying on a dark fallback mechanism
- It is a performance improvement
- It is more readable than using leading backslashes occasionally (as we are using only for the opcached optimized functions)
- It makes the symbol resolution easier, cf. PHP-Scoper & static analysis tools
- It is in the folded import statements: it doesn't hurt readability at least not more than the huge license header

.php_cs.dist
theofidry: :+1

.php_cs.dist
theofidry: Also note that I kinda capture them in each commit. Some commits may include multiple ones but it means it had no effect

src/Mutator/Unwrap/AbstractUnwrapMutator.php
```diff
@@ -35,6 +35,8 @@
 
 namespace Infection\Mutator\Unwrap;
 
+use function array_key_exists;
+use const Generator;
```
maks-rafalko: https://travis-ci.org/infection/infection/jobs/617500222#L358

> PHP Fatal error:  Declaration of Infection\Mutator\Unwrap\UnwrapArrayChangeKeyCase::getParameterIndexes(PhpParser\Node\Expr\FuncCall $node): Generator must be compatible with Infection\Mutator\Unwrap\AbstractUnwrapMutator::getParameterIndexes(PhpParser\Node\Expr\FuncCall $node): Infection\Mutator\Unwrap\Generator in /home/travis/build/infection/infection/src/Mutator/Unwrap/UnwrapArrayChangeKeyCase.php on line 55

Is it a bug in PHP-CS-Fixer?

src/Config/Guesser/SourceDirGuesser.php
```diff
@@ -35,22 +35,29 @@
 
 namespace Infection\Config\Guesser;
 
+use const DIRECTORY_SEPARATOR;
+use function in_array;
+use function is_array;
```
maks-rafalko: is it a bug of PHP-CS-Fixer? (duplicate use statement)

src/Config/ValueProvider/TestFrameworkConfigPathProvider.php
```diff
@@ -77,8 +80,8 @@ public function get(InputInterface $input, OutputInterface $output, array $dirsI
         try {
             $this->testFrameworkConfigLocator->locate($testFramework);
 
-            return null;
```
maks-rafalko: is it a correct change? I think `?string` is expected

src/Visitor/CodeCoverageClassIgnoreVisitor.php
```diff
@@ -48,19 +48,17 @@ final class CodeCoverageClassIgnoreVisitor extends NodeVisitorAbstract
     public function enterNode(Node $node)
     {
         if (!$node instanceof Stmt\ClassLike) {
-            return null;
```
maks-rafalko: it should be `null|int|Node` from interface's PHPDoc. Looks like incorrect change. Do we need to disable this rule?

.php_cs.dist
```diff
@@ -73,7 +71,9 @@ return Config::create()
         '@Symfony:risky' => true,
         '@PHP71Migration' => true,
         '@PHP71Migration:risky' => true,
+        '@PHP73Migration' => true,
```
BackEndTea: Our minimum php version is 7.2 right now i think, will this cause any issues?

theofidry: Indeed :/

src/Visitor/CodeCoverageClassIgnoreVisitor.php
theofidry: This was caused by `simplified_null_return` but I don't like the result either so reverted

src/Config/ValueProvider/TestFrameworkConfigPathProvider.php
theofidry: https://github.com/infection/infection/pull/865#discussion_r351839497

src/Mutator/Unwrap/AbstractUnwrapMutator.php
theofidry: Yes, but that we can manually fix not an issue

theofidry:
Updated & Fixed

src/Mutator/Unwrap/AbstractUnwrapMutator.php
theofidry: I mean correct it manually and at the next run it won't cause any issue; the issue there was likely a wrong guess to convert `Generator` to an import for some reasons

theofidry:
Updated and ran again. Please let's be quick about it to avoid tons of conflict! 😄

BackEndTea:
I don't think the current import style is PSR-12, which sorts alphabetically, but keeps functions, consts and classes separated. I'd prefer that style. But importing functions is already a 👍 from my side. Keeping it as is isn't a blocker for me.

theofidry:
>I don't think the current import style is PSR-12, which sorts alphabetically, but keeps functions, consts and classes separated. I'd prefer that style. But importing functions is already a 👍 from my side. Keeping it as is isn't a blocker for me.

I'm indifferent to the import statement sorting, so if someone wants to change it later I'm cool with that; it doesn't appear there is a lot of options here though unless I missed something

PR #866, opened 2019-11-27
Remove usage of @inheritdoc

### LINKED ISSUE #860: Remove or add @inheritdoc blocks
The codebase is a bit inconsistent regarding this so should we:

- add `{@inheritDoc}` where missing?
- remove `{@inehritDoc}`

In both cases there is an additional action to take. For the first one, we need to ensure it is `@inehritdoc` and not `@inheritDoc` as the former is not well understood by neither phpStorm & PHPDocumentor.

For the second one we should make sure they cannot be added (by an automatic CS fix or something)

theofidry:
Closes #860

I could find a way to check for the `@inheritdoc` tags but honestly the solution was way too slow & expansive. Since there is very little usage, I just removed it and I believe we can keep it under control with the PR review.

I'm also taking this opportunity to introduce an [ADR file](https://medium.com/better-programming/here-is-a-simple-yet-powerful-tool-to-record-your-architectural-decisions-5fb31367a7da) (thanks @Ocramius for the recommendation) which is a useful technique to document some decisions taken.

### DIFF
--- a/adr/0001-inheritdoc.md
+++ b/adr/0001-inheritdoc.md
@@ -1,0 +1,29 @@
+# Inheritdoc usage [(#860)](https://github.com/infection/infection/issues/860)
+
+### Decision
+
+Do not use `@inheritdoc` tags or any of its variants.
+
+
+###Status
+
+Accepted
+
+
+### Context
+
+Using `@inheritdoc` was done inconsistently across the codebase so the decision of whether we use it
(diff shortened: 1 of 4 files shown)


theofidry:
Did not, I just followed a random template that looked good enough for not at least. I think we can also adjust the template/format depending of our need without having to retrofit all the existing ones unless there is a real value.

PR #867, opened 2019-11-28
Do not allow using Infection with Codeception < 3.1.1

maks-rafalko:
This one and ready to go.

When Test Framework adapters will be extracted to own composer packages, this will be just in `infection/codeception-adapter`'s `composer.json`:

```json
{
    "conflicts": {
        "codeception/codeception": "<3.1.1"
    }
}
```

Now, we can't do it in the root's `composer.json`, because one can use Infection with e.g. PHPSpec while having Codeception installed as well.


maks-rafalko:
Thank you @sidz

PR #869, opened 2019-11-29
Introduce a FileSystemTestCase to simplify some I/O dependent test cases

theofidry:
I wanted to get rid of the `TmpDirectoryCreator` class but turns out it's used a lot in our tests, so here is an attempt at addressing the issue.

I think I'll create a filesystem repository soon with some elements from Box since it doesn't look like https://github.com/symfony/symfony/pull/31965 and https://github.com/symfony/symfony/pull/30969 will get merged

### DIFF
--- a/tests/phpunit/FileSystem/FileSystemTestCase.php
+++ b/tests/phpunit/FileSystem/FileSystemTestCase.php
@@ -1,0 +1,126 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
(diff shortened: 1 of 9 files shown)


tests/phpunit/FileSystem/FileSystemTestCase.php
```diff
@@ -0,0 +1,126 @@
...
+        $this->cwd = getcwd();
+        $this->tmp = make_tmp_dir('infection-test', self::class);
+
+        chdir($this->tmp);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function tearDown(): void
```
BackEndTea: I may be wrong here, but i think `tearDown` is not called after a failed test. Meaning we also need the clean up in a `tearDownAfterClass`

theofidry: I don't know if `tearDownAfterClass()` is called after a failing test either... But `teardown()` is definitely not. It should be done here still though in case you require a clean environment in-between each test method. Also the workspace used is in a deterministic base tmp dir hence, since we remove that base dir at the beginning, we always limit the number of unclean artefact

theofidry: Also note that FS is never transactional, so there is no safe way to guarantee a perfect environment with no artefact leftover. That's why the strategy here is more to limit where those artefacts can be and clean them up whenever we can (see the `setUp()`)

BackEndTea: It should indeed be done here, i'm suggesting to duplicate it in a `tearDownAfterClass`.

theofidry: Done

theofidry:
@infection/core any opinion regarding this one?

PR #870, opened 2019-11-29
CoverageDoesNotExistException - fix message

kubawerlos:
Since PCOV is allowed let's the message - when no coverage tool is used - reflects that.


maks-rafalko:
Thank you @kubawerlos

PR #871, opened 2019-11-29
Update .gitattributes

kubawerlos:
- `/.*` will make all files starting with `.` excluded witch is consistent with the convention that those are hidden ones.
- add `adr` directory
- fix directory name `.build` to `build`
- add file `phpunit_autoreview.xml`
- add file `setup_environment.sh`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,19 +5,16 @@
 *.jpg binary
 
 # Directories and files we do not want to distribute
-/.build/              export-ignore
-/.ci/                 export-ignore
-/.github/             export-ignore
-/devTools/            export-ignore
-/tests/               export-ignore
-/.editorconfig        export-ignore
-/.gitattributes       export-ignore
-/.gitignore           export-ignore
-/.php_cs.dist         export-ignore
-/.travis.yml          export-ignore
-/appveyor.yml         export-ignore
-/box.json.dist        export-ignore
-/codecov.yml          export-ignore
-/infection.json.dist  export-ignore
-/Makefile             export-ignore
-/phpunit.xml.dist     export-ignore
+/.*                     export-ignore
+/adr/                   export-ignore
+/build/                 export-ignore
+/devTools/              export-ignore
+/tests/                 export-ignore
+/appveyor.yml           export-ignore
+/box.json.dist          export-ignore
+/codecov.yml            export-ignore
+/infection.json.dist    export-ignore
+/Makefile               export-ignore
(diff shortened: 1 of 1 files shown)


sanmai:
This is more that just `.anything` to `.*` change?

kubawerlos:
@sanmai yes, I've updated the first post.

.gitattributes
```diff
@@ -5,19 +5,16 @@
...
-/phpunit.xml.dist     export-ignore
+/.*                     export-ignore
+/adr/                   export-ignore
+/build/                 export-ignore
+/devTools/              export-ignore
+/tests/                 export-ignore
+/appveyor.yml           export-ignore
+/box.json.dist          export-ignore
+/codecov.yml            export-ignore
+/infection.json.dist    export-ignore
```
theofidry: I would skip those configuration files: no one wants to keep track of them and they are literally just a few KB... It won't impact the resulting artefact in any way

kubawerlos: No one wants to have them in the project even more.

We can add a test to ensure only allowed paths are in the artefact.

theofidry: >No one wants to have them in the project even more.

It feels to me more an additional burden/pressure putted in the maintainers for absolutely no gain other than a certain sense of purity that you will never achieve.

Removing tests and bigger non-prod directories is acceptable because there is usually not many of them and they don't change often, so having someone updating a `.gitattributes` to get rid of them is ok. But going further and trying to trim the git artefact from absolutely all unnecessary files is an unnecessary strain for absolutely no value (you will gain maybe 100KB max from it which is nothing)

You could add a test, but it doesn't remove the problem: it means further friction when contributing/maintaining because you may need to update that test, it means having an additional test/code for it which is to be maintained as well... So again, for 100KB it's just not worth it IMO

theofidry:
@infection/core any opinion on this one? I'm obviously quite biased so I would like someone else to settle this one so that we can move forward by either accepting the PR as is or updating it to get it merged

maks-rafalko:
Personally, I'm not for and not against. Too little issue to spent time here ;) So I'm ok with merging as is and I'm ok with not adding particular files to `.gitattributes`.

But let's see how another big projects do it:

* PHP-Parser https://github.com/nikic/PHP-Parser/pull/622/files (ignoring folders and several files)
* Symfony https://github.com/symfony/symfony/pull/33579/files (ignoring folders and particular files)

theofidry:
Then let's done with it; thanks @kubawerlos

PR #872, opened 2019-11-30
Remove the MutationInterface

theofidry:
Since there is only one implementation for now, I don't think it's worth it to have an interface. I would like at least as a first step consolidate a bit more the base (review the fields, dependencies and API of the mutator/mutation/mutant) before really working on an interface with a public API for them


maks-rafalko:
I'm not sure we need this change. Some time ago I started experimenting with extracting Test Framework Adapters to separate packages, as well as creating abstract and stable `infection/contracts` package (as a result of applying package design patterns). 

And we have the following dependency sub-graph:

`TestFrameworkAdapterInterface` -> `MutantInterface` -> `MutationInterface`

So, we  will definitely need this interface, even though it may be changed.

theofidry:
Having an interface is a requirement for splitting it, but until we are not splitting it I don't think it's a good idea because that's exactly how you end up with:

- An interface which has undesirable dependencies
- An interface which has too much or too little
- An API which although provides an interface, end up requiring a base abstract class which really defeat the purpose of having an interface

I'm not against opening up the API, but I think it's way too early for that. We still have yet to introduce key changes in Infection such as the Tracing abstraction, categorising mutations, metrics features or normalizing the API based on the existing tools such as mbj/mutant, Stryker or Pitest. And I don't think this is gonna be finished before at least another year (which is actually not that much in OSS)

theofidry:
Edit: also we don't have a mutator interface yet...

sanmai:
I like the way we had it with a closed class and an interface, very different from what you typically see in legacy/proprietary projects where everything extends god knows what over what in what combinations. It is fine when you get paid to work with these things, but it's like a plague still. 

(I might be just spoiled by Go, where you can't make a step without using some sort of interface. But that's the way you really feel at ease with the code.)

theofidry:
I'm for keeping it final: I don't want a god class either. I'm suggesting to have a real public interface for it once we need it

maks-rafalko:
> It would be possible to publish an `infection/contracts` or equivalent before that, but it's really gonna add a big overhead for this work so it wouldn't be wise to do so too early

Are you saying we can't create `infection/contracts` before implementing all those features ("...Tracing abstraction, categorising mutations, metrics features or normalizing the API based on the existing tools such as mbj/mutant, Stryker or Pitest...")? 

If so, I completely disagree. We will never create good interfaces from the first try. And we don't to. `infection/contracts` does *not* mean that people will depend on it right from the start and we won't give any guarantees unless `1.0.0` about these interfaces.

From my POV, `infection/contracts` (at least interfaces I mentioned above) are just to extract some part of the code (test framework adapters) out of `infection/infection` and nothing more. These are not "extension points" for everyone right after publishing the package. The whole abstract package can be `@internal` for some period.

I guess we just thinking about different purposes: extension points _vs_ splitting the core.

---

Regarding this PR, for me it looks like this: we 100% sure that we need `MutationInterface`, but we remove it, changing ~20 files. Tomorrow (or year later, does not matter), we will revert this commit, requiring reviewer attention etc.

Don't get me wrong, I'm not against it, just can't understand (or want understand) why are we doing it.

theofidry:
I have hard time understanding the benefits of `infection/contracts` - be it in terms of extension point or code splitting, on the short term when all the areas it will touch are _very_ likely to not say certain to undergo massive changes in a very near future. I hardly see a case where it's not going to cause a massive overhead work-wise while not providing great benefits for consumers.

To clarify: I'm not against the idea at all, I'm against it before introducing those key abstractions because they bring crucial architectural changes. They also by design (because I'm not making them up out of no where, they are actually already well rounded existing designs) brings the essential elements we want to use for exposing a public API as an extension point.

>Regarding this PR, for me it looks like this: we are 100% sure that we need MutationInterface, but we remove it, changing ~20 files. Tomorrow (or year later, does not matter), we will revert this commit, requiring reviewer attention etc.

How are you 100% sure of this? Because I actually have doubts we want to offer an extension point there hence I question the need of introducing an abstraction here. As it stands, the `Mutation` is:

- The piece of code representing the change to apply
- Expose its origin (the Mutator - which is not an interface)
- Expose PHP-Parser implementation details (I am not saying it is wrong but it is tightly coupled to it as Infection as a whole is)
- Expose a mixture of test and coverage dependency

I think if we want to expose something as an extension point in the future, it is rather going to be `Mutator` (not necessarily as it is right now) rather than `Mutation`.

src/Mutation/Mutation.php
```diff
@@ -42,9 +42,11 @@
 use PhpParser\Node;
 
 /**
+ * @final
+ *
  * @internal
  */
-final class Mutation implements MutationInterface
+class Mutation
```
sanmai: I tend to think that having an interface and a closed class just by its own virtue has a merit, even if don't plan to export that interface. I might be spoiled by the fruit of Go, where you can't make a step without an interface of some sorts, but that's the better way because you naturally avoid all the problems of inheritance.

theofidry: I mean the class would be final (`@final` only for now though) so inheritance is definitely out of question here.

The question is more that `Mutation` is more an object for the _engine_ to consume: it is not where you want to allow an extension point because that would mean allowing the engine to be extended (I don't mean by inheritance but in terms of features/behaviour) in some ways which is a very different idea

sanmai: I see that `MutationInterface` is tagged `@internal` so it isn't nor wasn't an extension point per se.

maks-rafalko:
After [we decoupled](https://github.com/infection/infection/pull/928) `TestFrameworkInterface` from `MutantInterface`, which in its turn depended on `MutationInterface` (being removed here), we are safe to replace `MutationInterface` with a concrete class for now.

PR #873, opened 2019-11-30
Tweak MutatorProfile

theofidry:
Rework `MutatorProfile` to attempt to have more accurate names and more performant/easier to read tests.

- Rename `MutatorProfile` to `ProfileList` which is slightly more accurate but still fails to capture the `ALL_MUTATORS` one; that said I think the name is description enough without being overly too long
- Move up `MutatorProfile` by a namespace: I don't think it belongs to `Util` (which ultimately I want to get rid of)
- Rename `MUTATOR_PROFILE_LIST` to `ALL_PROFILES`
- Rename `ARITHMETIC` and other profile constants to `ARITHMETIC_PROFILE`
- Rename `FULL_MUTATOR_LIST` to `ALL_MUTATORS`
- Make `ProfileList ` non instantiable making it closer to an actual enum
- Refactor `ProfileListTest` (previously `MutatorProfileTest`) to make use of an external provider, `ProfileListProvider` the same way we do for `ProjectCodeTest` with `ProjectCodeTestProvider`
- Consolidate the tests and rework the failure messages to be more in line with the regular PHPUnit failure messages

### DIFF
--- a/tests/phpunit/Mutator/Util/MutatorsGeneratorTest.php
+++ b/tests/phpunit/Mutator/Util/MutatorsGeneratorTest.php
@@ -58,7 +58,7 @@ final class MutatorsGeneratorTest extends TestCase
     public static function setUpBeforeClass(): void
     {
         foreach (ProfileList::DEFAULT_PROFILE as $profileName) {
-            self::$countDefaultMutators += count(\Infection\Mutator\ProfileList::ALL_PROFILES[$profileName]);
+            self::$countDefaultMutators += count(ProfileList::ALL_PROFILES[$profileName]);
         }
     }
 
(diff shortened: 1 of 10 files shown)


tests/phpunit/Mutator/Util/MutatorsGeneratorTest.php
```diff
@@ -98,7 +98,7 @@ public function test_profiles_can_be_ignored(): void
         ]);
         $mutators = $mutatorGenerator->generate();
 
-        $this->assertCount(self::$countDefaultMutators - count(MutatorProfile::BOOLEAN), $mutators);
+        $this->assertCount(self::$countDefaultMutators - count(\Infection\Mutator\ProfileList::BOOLEAN_PROFILE), $mutators);
```
theofidry: Hm not sure why this is happening occasionally but I'm changing this test in my next PR anyway

theofidry:
@infection/core?

theofidry:
@infection/core if I could get a fast review from someone... There is actually not much change here and it's blocking for already 4 pending PRs...

PR #874, opened 2019-12-01
Refactor MutatorsGenerator

theofidry:
- Rename `MutatorsGenerator` to `MutatorFactory`
- Move `MutatorFactory` up by a namespace (moving it out of `Util`)
- Make `MutatorFactory` stateless, allowing it to be injected as a service
- No longer handle defaults in `MutatorFactory`: if no profile/mutator is passed, `@default => true` is given from `ConfigurationFactory` instead of `MutatorFactory`
- No longer throw an `InvalidConfigException` and instead throw an `InvalidArgumentException`: those errors can only happen by bypassing the `schema.json` validation hence is going to be a programmer error instead of a legit runtime error
- Add additional tests to capture the current behaviour of `MutatorFactory` regarding the order of the profiles and mutators passed


tests/phpunit/Mutator/MutatorFactoryTest.php
```diff
@@ -0,0 +1,343 @@
...
+use InvalidArgumentException;
+use PhpParser\Node;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use ReflectionClass;
+use function Safe\sprintf;
+
+final class MutatorFactoryTest extends TestCase
+{
+    private static $countDefaultMutators = 0;
```
theofidry: can be safely removed

theofidry:
@sanmai will be easier to see after #873 since it's based on it

theofidry:
Rebased, cleaned up, ready for review

src/Mutator/MutatorFactory.php
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /**
+     * @param array<string, string>|bool           $settings
+     * @param array<string, array<string, string>> $mutators
+     */
+    private static function registerFromProfile(
+        string $profile,
+        $settings,
+        array &$mutators
```
BackEndTea: I'm not really a fan of passing the `$mutators` array around by reference.

But i'm not 100% sure what another solution would be.

theofidry:
Also it’s a private method so it’s more as a kind of optimisation and
clarity than necessity. I could create a new instance and return it but I
don’t think the code would be much cleaner and it would certainly be less
efficient

tests/phpunit/Mutator/MutatorFactoryTest.php
```diff
@@ -0,0 +1,334 @@
...
+    public function test_it_can_creates_the_mutators_with_empty_settings_for_a_given_profile(): void
+    {
+        $mutators = $this->mutatorFactory->create(['@boolean' => []]);
+
+        $this->assertSameMutatorsByClass(ProfileList::BOOLEAN_PROFILE, $mutators);
+    }
+
+    public function test_it_can_creates_the_profile_mutators_with_the_given_settings(): void
+    {
+        // TODO: refactor this test to check the mutator configuration directly instead of relying
```
maks-rafalko: Is there any ticket for it? just to not leave it forever.

Same for other TODOs

src/Mutator/MutatorFactory.php
maks-rafalko: Perhaps, we can use `\ArrayObject` here without `&`? See https://3v4l.org/UBaZI

theofidry: Also note in case this is not clear: this array is strictly internal to the class, it is a temporary artefact. It is not what we provide as a result of `MutatorFactory::create()`

tests/phpunit/Mutator/MutatorFactoryTest.php
theofidry: It's the next PR already created waiting for a rebase :)

PR #875, opened 2019-12-01
Rework the MutatorParser

### LINKED ISSUE #844: Duplicated execution of MutatorsGenerator
As I can see, we call `new MutatorsGenerator($schema->getMutators()))->generate()` twice for some reason.

Here, and inside [`MutatorParser::php`](https://github.com/infection/infection/blob/5dffa826a07b0f2ee6f6843aea621346f5194c05/src/Mutator/Util/MutatorParser.php#L80)

From what I understand, we can first merge `$schema->getMutators()` and values from `--mutators`, and only then pass them to `(new MutatorsGenerator($mergedMutators))->generate()`

_Originally posted in https://github.com/infection/infection/pull/841#discussion_r349878447_

theofidry:
Depends on #874

- Move `MutatorParser` up by a namespace
- Move the existing config handling (i.e. picking between the parsed input & the schema config) to the `ConfigurationFactory`; allowing to trim down the `MutatorParser` to parsing the input only
- Consolidate the tests and better handle some edge cases
- Force `mutatorsInput` to be a `string` instead of `string|null` allowing to simplify the handling of that input in `InfectionCommand`

Closes #844


src/Configuration/ConfigurationFactory.php
```diff
@@ -130,4 +134,15 @@ public function create(
             $minCoveredMsi
         );
     }
+
+    private function retrieveMutators(array $schemaMutators, string $mutatorsInput): array
```
maks-rafalko: could you please specify array types?

theofidry: it's `mixed[]`... hence `array`. It's our "messy" type coming from the configuration file hence if it was typescript:

```typescript
type Ignore = {
  ignore?: string[];
}

type Settings = {
  settings:? {
    [key: string]?: string;
  }
}

type MutatorConfig = boolean | (Ignore & Settings);
```

maks-rafalko: Then we can have exactly the same DTO:

```php
final class MutatorConfig {
    public array<int, string>|null $ignore;

    public array<string, string>|null $settings
}
```

and then

```diff
- private function retrieveMutators(array $schemaMutators, string $mutatorsInput): array
+ /**
+ * @param bool|MutatorConfig $schemaMutators
+ * ...
+ */
+ private function retrieveMutators(array $schemaMutators, string $mutatorsInput): array
```

Would it work? This is definitely out of the scope of this PR, but if it would work, we will implement it as a separate ticket to remove all these not clear `array`s

theofidry: It's not exactly the same though but I think we can improve it indeed. Opened https://github.com/infection/infection/issues/882 for it

PR #877, opened 2019-12-04
Add notice to console output if actual msi is higher than required msi

### LINKED ISSUE #810: Notify user if msi is over specified min-msi

theofidry:
Replacement for #811

Closes #810


theofidry:
Thank you @LeoVie

PR #878, opened 2019-12-04
Replace `TmpDirCreator` by `TmpDirProvider `

theofidry:
With #869 merged, `TmpDirCreator` is now only used for our "production" code, i.e. is no longer used for the tests.

As a result, I'm taking this opportunity to change it: indeed I think the class (and its testing) is overly complex because besides providing a temporary path, it attempts to create it. This also has the unpleasant side effect of potentially creating a temporary directory whenever a `ConfigurationFactory` is used to create a configuration.

In other words: this PR simplifies `TmpDirCreator` by making it responsible only to provide a temporary directory, making the creation of `Configuration` via `ConfigurationFactory` side-effect free.


src/Configuration/ConfigurationFactory.php
```diff
@@ -103,7 +103,7 @@ public function create(
             // calculation" and the "tmp dir create" in order to not have
```
maks-rafalko: is this comment still needed?

src/Command/InfectionCommand.php
```diff
@@ -285,6 +286,11 @@ private function startUp(): TestFrameworkAdapter
 
         $this->includeUserBootstrap($config);
 
+        /** @var Filesystem $fileSystem */
+        $fileSystem = $this->container['filesystem'];
+
+        $fileSystem->mkdir($config->getTmpDir());
```
maks-rafalko: do we need `, 0777` as a second parameter (as it was in the `TmpDirCreator`)?

theofidry: we can put it but it's the default so not a necessity

PR #881, opened 2019-12-07
Move the filter input as a dynamic container parameter

theofidry:
This PR:

- Moves the `filter` parameter of `MutationsGenerator` in the constructor instead of the `generate()` method
- Moves the passed `filter` value from the input to the `Configuration` instead, allowing:
  - `MutationsGenerator` to depend on `Configuration` only instead of `Configuration` and `InputInterface` (to be instantiated, so it's not a dependency at the class level itself but about how we use it)
  - Move `filter` into `InfectionContainer::withDynamicParameters()` with the other parameters which makes it more consistent with the rest of the code

If you pay attention you might notice that we then have all the necessary parameters to collect the source files right in the constructor of `MutatorsGenerator`. The ultimate goal here is to move the collection of the source files in `ConfigurationFactory`. I kept the PR at this because attempting to go further involves a lot more changes. It was tough but I found a way to split this work into 3-4 PRs this one being the first one.

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -336,18 +336,19 @@ final class InfectionCommand extends BaseCommand
         $processBuilder = new MutantProcessBuilder($adapter, $this->versionParser, $config->getProcessTimeout());
 
         $codeCoverageData = $this->getCodeCoverageData($this->testFrameworkKey, $adapter);
+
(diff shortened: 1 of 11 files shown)


theofidry:
@infection/core I have about 10 PRs following this one, if we could manage to review them this week and next one that would be great because I'm not sure about my availabilities after until mid-January :)

PR #885, opened 2019-12-10
Use stable thecodingmachine/safe

snapshotpl:


theofidry:
why not pushing for the beta only instead of allowing `^0.1` as well?

sidz:
do we really would like to use library in the beta/alpha version stage? @theofidry @maks-rafalko

maks-rafalko:
> do we really would like to use library in the beta/alpha version stage?

Well, although we are safe since we are using `composer.lock` file, I would still avoid using beta versions as a direct dependencies in Infection.

As I understand, @snapshotpl wants to be able to install Infection to his project, where `1.0.0.@beta` is already used. 

What if I suggest you using Infection PHAR distribution instead and forget about conflicted dependencies ;) ?

snapshotpl:
@maks-rafalko that's great idea, however it's not simple to install and maintain infection's phar like [phpstan](https://github.com/phpstan/phpstan) or [rector](https://github.com/rectorphp/rector-prefixed)

sidz:
What is difference between phar for phpstan and infection?

ср, 11 дек. 2019 г., 10:26 Witold Wasiczko <notifications@github.com>:

snapshotpl:
I can download phpstan as phar via composer

theofidry:
>I would still avoid using beta versions as a direct dependencies in Infection.

In this specific case we use the dev version (0.1) which is not any safer than a beta

maks-rafalko:
What is the benefit to use PHAR through composer rather than just download it or use https://phar.io/ to manage PHARs?

snapshotpl:
I don't need to change nothing in my project. Like phpstan did in 0.12. After composer install I just get phar. In your case in need to change my local and production deployment process.

theofidry:
@snapshotpl any chance to do a `composer update`? Would love to get this one merged :)

snapshotpl:
@theofidry  `thecodingmachine/safe` is now in stable version :)

theofidry:
Thank you @snapshotpl

PR #888, opened 2019-12-13
Move the collection of source files in the ConfigurationFactory

theofidry:


tests/phpunit/Configuration/ConfigurationFactoryTest.php
```diff
@@ -73,10 +76,26 @@ public static function tearDownAfterClass(): void
 
     protected function setUp(): void
     {
+        /** @var SourceFileCollector&ObjectProphecy $sourceFilesCollectorProphecy */
+        $sourceFilesCollectorProphecy = $this->prophesize(SourceFileCollector::class);
```
theofidry: I used prophecy here because I had no idea how to do it with `Mock`. If someone has the alternative I'll take it, otherwise the following works...

tests/phpunit/FileSystem/SourceFileCollectorTest.php
```diff
@@ -103,17 +99,6 @@ public function sourceFilesProvider(): Generator
             ],
         ];
 
-        yield 'one directory, empty filter, no excludes' => [
```
theofidry: removed because it was a duplicate

PR #889, opened 2019-12-13
Remove the deprecated InfectionConfig class

theofidry:


src/Config/ValueProvider/TimeoutProvider.php
```diff
@@ -1,84 +0,0 @@
-<?php
```
theofidry: This class is no longer necessary: we pass the default timeout value (10s) in `ConfigurationFactory` and this setting was never required in `schema.json`.

sanmai:
There's an E2E test failing for the `Configure` operation, do you think it might be related?

theofidry:
Indeed it was related :)

theofidry:
@maks-rafalko is there a way to configure the cov service? I'm not quite sure why it was failing in this PR but it looked wrong

maks-rafalko:
what do you mean by configuring? the CodeCov config is located here:

https://github.com/infection/infection/blob/master/codecov.yml

PR #890, opened 2019-12-14
Rename Mutant\MutationsGenerator into Mutation\MutationGenerator

theofidry:
I think the `Mutation` semantic is more appropriate here:

- We generate `Mutation[]` not `Mutant[]`
- Mutants would be the subject to which the mutation has been applied, i.e. the mutated code.

I do realise having a `Mutation` namespace for only one class looks more being of a semantic nitpick than a necessity. I'm however deciding to push for it for two reasons:

- the `Mutation` namespace is going to contain more stuff in the immediate next PR. I'm not doing it in this one to ease the review
- I would like to introduce an official nomenclature at some point, but meanwhile I think this one is a good one as a base: https://github.com/mbj/mutant/blob/master/docs/nomenclature.md


src/Mutation/MutationGenerator.php
```diff
@@ -118,9 +118,7 @@ public function generate(bool $onlyCovered, array $extraNodeVisitors = []): arra
         $this->eventDispatcher->dispatch(new MutationGeneratingStarted(count($this->sourceFiles)));
 
         foreach ($this->sourceFiles as $file) {
-            if (!$onlyCovered || $this->hasTests($file)) {
```
theofidry: I moved this check in the `getMutationsFromFile()` which will be more in line with what's coming next

src/Command/InfectionCommand.php
```diff
@@ -341,7 +341,7 @@ private function runMutationTesting(TestFrameworkAdapter $adapter): void
         /** @var FileParser $parser */
         $parser = $this->container[FileParser::class];
 
-        $mutationsGenerator = new MutationsGenerator(
+        $mutationsGenerator = new MutationGenerator(
```
maks-rafalko: Should we rename a variable as well?

PR #892, opened 2019-12-15
Introduce File Parser

theofidry:
This adds an adapter `FileParser` to wrap around the nikic/PHP-Parser one. The adapter is very thin because there is not much to do, but good to do nonetheless IMO. Indeed it provides:

- An easier way to test _our_ PHP-Parser integration (the adapter test are an integration one - we test against how we configured the PHP-Parser in the container)
- Allows to adapt the usage to handle:
  - parsing a `SplFileInfo` instead of `string`
  - throwing one of our own exception instead of an arbitrary one
  - properly return `Node[]` instead of `Stmt[]|null`

### DIFF
--- a/src/Console/InfectionContainer.php
+++ b/src/Console/InfectionContainer.php
@@ -54,6 +54,7 @@ use Infection\Locator\RootsFileLocator;
 use Infection\Locator\RootsFileOrDirectoryLocator;
 use Infection\Mutant\MetricsCalculator;
 use Infection\Mutant\MutantCreator;
+use Infection\Mutation\FileParser;
 use Infection\Mutator\MutatorFactory;
 use Infection\Mutator\MutatorParser;
 use Infection\Performance\Limiter\MemoryLimiter;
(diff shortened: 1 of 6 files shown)


src/Console/InfectionContainer.php
```diff
@@ -204,15 +205,24 @@ public static function create(): self
...
-                return (new ParserFactory())->create(ParserFactory::PREFER_PHP7, $container['lexer']);
+            Parser::class => static function (self $container): Parser {
+                /** @var Lexer $lexer */
+                $lexer = $container[Lexer::class];
+
+                return (new ParserFactory())->create(ParserFactory::PREFER_PHP7, $lexer);
+            },
+            FileParser::class => static function (self $container): FileParser {
+                /** @var Parser $phpParser */
+                $phpParser = $container[Parser::class];
```
BackEndTea: I wouldn't mind using the values directly, instead of assigning a temporary variable, but thats not blocking for me

src/Console/InfectionContainer.php
```diff
@@ -204,15 +205,24 @@ public static function create(): self
             VersionParser::class => static function (): VersionParser {
                 return new VersionParser();
             },
-            'lexer' => static function (): Lexer {
+            Lexer::class => static function (): Lexer {
```
BackEndTea: 👍 For these changes

src/Console/InfectionContainer.php
theofidry: It's ultimately what I would like to do but I think it's better to do so once we introduce getters in the container for the IDE to be able to infer the type. Otherwise right now if you do "Find by usage" the IDE will not find it and SA have trouble with it as well

PR #891, opened 2019-12-14
Trim MutationGenerator

theofidry:
(not ready for review)

Gist of the PR split some elements of `MutationGenerator` in multiple smaller bits.

PR #893, opened 2019-12-15
Move the instantiation of the traverser from MutatorsGenerator to a new NodeTraverserFactory class

theofidry:


src/Mutant/Generator/MutationsGenerator.php
```diff
@@ -137,31 +135,21 @@ private function getMutationsFromFile(SplFileInfo $file, bool $onlyCovered, arra
...
-        $traverser->addVisitor($mutationsCollectorVisitor, 10);
-
-        foreach ($extraNodeVisitors as $priority => $visitor) {
-            $traverser->addVisitor($visitor, $priority);
-        }
-
         $traverser->traverse($initialStatements);
 
-        return $mutationsCollectorVisitor->getMutations();
+        return $traverser->getMutationsCollectorVisitor()->getMutations();
```
BackEndTea: What if we cut out the `getMutationsCollectorVisitor` here, and  have a `getMutations` method on the traverser itself?

theofidry:
I don’t like it either; I managed to get something better while on the
plane. I’ll push later in the night when I get home

theofidry:
Pushed

src/Mutation/PriorityNodeTraverser.php
```diff
@@ -33,7 +33,7 @@
 
 declare(strict_types=1);
 
-namespace Infection\Traverser;
+namespace Infection\Mutation;
```
maks-rafalko: What's the reason of this change? 

Currently, this namespace contains `FileParser` and `UnparsableFile` classes. Adding Traverser there makes it even more "too common". At lease I don't understand what we are going to add to this namespace.

src/Mutation/PriorityNodeTraverser.php
```diff
@@ -52,4 +52,12 @@ public function addVisitor(NodeVisitor $visitor, int $priority = 1): void
 
         krsort($this->visitors);
     }
+
+    /**
+     * @return NodeVisitor[]
+     */
+    public function getVisitors(): array
```
maks-rafalko: IMO, adding `getter` just for the test is not the best idea. There is an existing class that tests order of visitors without getter

theofidry: There is not really any way to test the factory otherwise. It's fine by me, but a bit of a shame as well and I would argue it is more of a design issue of the inherited interface to not be able to have any runtime introspection of the registered visitors.

The existing class you are referring to is the traverser itself. It does test that if you register visitors in a specific priority order they are called in the expected order. It is however not what I wished to test here in the traverser factory: I wanted to test that the created traverser instance was configured the expected way.

theofidry: So we either need to choose between having this method for test purposes or having the factory untested

src/Mutation/PriorityNodeTraverser.php
theofidry: The reason of this specific change is that having a `Traverser` namespace here does not make any sense to me at least right now.

There is a lot of debate on how to organise things, but unless you have a specific adapter in your applications, I feel the "keep things that change together close to each other" a good rule of thumb. So in `Infection\Mutation` it's primarily `MutationGenerator`. The parse, traverser, traverser factory are merely sub elements of it.

The main alternative I can think of that makes sense as well is having "PhpParser adapter", e.g. like [here in PHP-Scoper](https://github.com/humbug/php-scoper/tree/master/src/PhpParser). I think however that it won't suit Infection as well as I would really prefer to keep the mutators, which directly depend on PhpParser, under the `Mutator` directory.

src/Mutation/PriorityNodeTraverser.php
maks-rafalko: > of a design issue of the inherited interface to not be able to have any runtime introspection of the registered visitors

I'm not sure it is a design issue if nobody uses this internal thing outside of the Traverser class.

> So we either need to choose between having this method for test purposes or having the factory untested

No, let's test it. But adding a *public* method only for tests is not better than testing private variable with reflection inside PHPUnit test.

Since this class is internal, we probably can live with it.

theofidry: I'm confused, how do you want to do it?

The factory is about creating a traverser instance and configure visitors to it. If you cannot inspect the visitors then you cannot check it.

Event the traverser test itself won't work here since we configure a bunch of default visitors, so you cannot test the call the same it it was done there.

maks-rafalko: > I'm confused, how do you want to do it?

let's leave the current implementation. I'm ok with it.

> how do you want to do it

For me, adding a public method only for test is worse than for example testing internal state of the object by reflection. Because we are creating a method that nobody uses (and should not). 

Here we have an array, which is not returned by reference but by the value. If it would be e.g. an object, this starts to be unsafe, because the caller of `getVisitors()` [can influence](https://3v4l.org/Ml4QR) to the code - to the internal state of the object

theofidry: Why not accepting the PR changes but removing the factory test and the traverser `getVisitors()` then? The intent behind is to trim `MutatorsGenerator` which is doing too much (IMO - at least the way it is it's hard to introduce the tracer abstraction). See https://github.com/infection/infection/pull/891/files#diff-befced5be7cb2779621eec604f8caf9c for the kind of thing I would prefer the mutator to be like, although it's not its end form yet either.

The traverser needs to be instantiated and configured in any case: it can either be hidden in the `MutationGenerator` making it a bit more bloated or it can be moved to a dedicated factory to try to trim down the former. But in both cases (unless `getVisitors()` is kept), it will be untested

maks-rafalko: > Why not accepting the PR changes but removing the factory test and the traverser getVisitors() then? 

Sorry, I didn't get it. Let's move forward.

theofidry: Another possibility would be to change `NodeTraverserFactory` into `NodeTraverserConfigurator` instead:

```php
final class NodeTraverserConfigurator
{
    /**
     * @param NodeVisitor[] $extraVisitors
     */
    public function configure(PriorityNodeTraverser $traverser, array $extraVisitors): void
    {
        $traverser->addVisitor(new NotMutableIgnoreVisitor(), 50);
        $traverser->addVisitor(new ParentConnectorVisitor(), 40);
        $traverser->addVisitor(new FullyQualifiedClassNameVisitor(), 30);
        $traverser->addVisitor(new ReflectionVisitor(), 20);

        foreach ($extraVisitors as $priority => $visitor) {
            $traverser->addVisitor($visitor, $priority);
        }
    }
}
```

This can be done in another PR though. WDYT?

src/Mutation/PriorityNodeTraverser.php
theofidry: One additional note @maks-rafalko because I feel I failed to express that. I don't think my current changes will be the "final" way either. I however prefer this structure over a separation of things by type (e.g. putting every model under `Model`, every exception under `Exception`) because IMO it's makes it easier to see what's closely related to each other.

I still have quite _a lot_ of refactoring planned for the next months; after those I think it would be interesting to have a review as a team about the organisation to see if what we are not satisfied with and if we could improve it in some way.

maks-rafalko: no problem, thanks for the information

PR #894, opened 2019-12-15
Simplify `MutationTestingResultsLoggerSubscriber`

BackEndTea:
This PR:

- [x] Simplifies the `MutationTestingResultsLoggerSubscriber` and introduces a LoggerFactory
- [x] Covered by tests

The `MutationTestingResultsLoggerSubscriber` was responsible for both creation of the loggers, and calling log on every single one of them. We extract the creation logic into a `LoggerFactory`. This simplifies both the subscriber and its tests by a lot.


src/Console/InfectionContainer.php
```diff
@@ -333,7 +332,8 @@ public static function create(): self
                     $config->getTmpDir(),
                     $container['timer'],
                     $container['time.formatter'],
-                    $container['memory.formatter']
+                    $container['memory.formatter'],
+                    $container[LoggerFactory::class]
```
theofidry: As mentioned somewhere else, I would do the effort to put it in a temporary variable. I think this effort is really necessary for now otherwise IDE inspections, which are by no means mandatory but helps a lot, won't work in cases like those.

src/Console/InfectionContainer.php
```diff
@@ -342,6 +342,18 @@ public static function create(): self
             SourceFileCollector::class => static function (): SourceFileCollector {
                 return new SourceFileCollector();
             },
+            LoggerFactory::class => static function (self $container): LoggerFactory {
+                /** @var Configuration $config */
+                $config = $container[Configuration::class];
+
+                return new LoggerFactory(
+                    $container['metrics'],
+                    $container['filesystem'],
```
theofidry: same there for those two

src/Logger/LoggerFactory.php
```diff
@@ -0,0 +1,188 @@
...
+    private $isDebugMode;
+
+    /**
+     * @var bool
+     */
+    private $isOnlyCoveredMode;
+
+    public function __construct(
+        MetricsCalculator $metricsCalculator,
+        Filesystem $fs,
```
theofidry: I would call it `$fileSystem` (and same for the variable) :) I really tend to read it as "fuck sake" 😄

src/Logger/LoggerFactory.php
```diff
@@ -0,0 +1,188 @@
...
+        string $logVerbosity,
+        bool $isDebugMode,
```
theofidry: bool values should not be prefixed by `is`: this should be reserved for their accessors

src/Logger/LoggerFactory.php
```diff
@@ -0,0 +1,188 @@
...
+        $this->fs = $fs;
+        $this->logVerbosity = $logVerbosity;
+        $this->isDebugMode = $isDebugMode;
+        $this->isOnlyCoveredMode = $isOnlyCoveredMode;
+    }
+
+    /**
+     * @return MutationTestingResultsLogger[]
+     */
+    public function createLoggersFromLogEntries(Logs $logs, OutputInterface $output): array
```
theofidry: `createFromLogEntries()`? I find `Loggers` a bit redundant here given we are in the `LoggerFactory`

src/Logger/LoggerFactory.php
```diff
@@ -0,0 +1,188 @@
...
+    {
+        $isDebugVerbosity = $this->logVerbosity === LogVerbosity::DEBUG;
+        $badge = $logs->getBadge();
+        $debug = $logs->getDebugLogFilePath();
+        $perMutator = $logs->getPerMutatorFilePath();
+        $summary = $logs->getSummaryLogFilePath();
+        $text = $logs->getTextLogFilePath();
+
+        /** @var array<MutationTestingResultsLogger> $loggers */
```
theofidry: `MutationTestingResultsLogger[]` is simpler in this case

tests/phpunit/Logger/LoggerFactoryTest.php
```diff
@@ -0,0 +1,184 @@
...
+
+        $loggers = $factory->createLoggersFromLogEntries(
+            new Logs(
+                '/a/file',
+                '/a/file',
+                '/a/file',
+                '/a/file',
+                null
+            ),
+            $this->createMock(OutputInterface::class)
```
theofidry: since you use a similar mock in every test method I would move it to `setUp()`

tests/phpunit/Logger/LoggerFactoryTest.php
```diff
@@ -0,0 +1,184 @@
...
+        );
+
+        $this->assertCount(0, $loggers);
+    }
+
+    /**
+     * @dataProvider provideLogTypesAndClasses
```
theofidry: I would recommend `logTypesAndClassesProvider` rather: `*Provider` for a PHPUnit data provider and `provide*` for a regular generator. See for example `ProjectCodeProvider` where both are used

tests/phpunit/Logger/LoggerFactoryTest.php
```diff
@@ -0,0 +1,184 @@
...
+     */
+    public function test_it_creates_logger_for_log_type(Logs $logs, string $expectedLogClass): void
```
theofidry: `test_it_creates_a_logger_for_log_type`?

tests/phpunit/Logger/LoggerFactoryTest.php
```diff
@@ -0,0 +1,184 @@
...
+            $logs,
+            $this->createMock(OutputInterface::class)
+        );
+        $this->assertCount(1, $loggers);
+        $this->assertInstanceOf($expectedLogClass, current($loggers));
+    }
+
+    public function provideLogTypesAndClasses(): Generator
+    {
+        yield [
```
theofidry: I'm a bit confused a the order: wouldn't it make more sense to have a case for each argument in order? i.e.

- text
- summary
- debug
- per mutator

instead of the current order.

Also feels like it's note quite clear why we are missing the following cases:

- no logs
- the last log (badge)
- all of them

tests/phpunit/Logger/LoggerFactoryTest.php
```diff
@@ -0,0 +1,184 @@
...
+            SummaryFileLogger::class,
+        ];
+
+        yield [
+            new Logs(null, null, 'debug_file', null, null),
+            DebugFileLogger::class,
+        ];
+    }
+
+    public function test_it_creates_a_bade_logger(): void
```
theofidry: `test_it_creates_a_badge_logger` or `test_it_creates_a_badge_logger_on_normal_verbosity` since it appears to be important here

theofidry: Why is this one not included in `test_it_creates_logger_for_log_type` instead?

tests/phpunit/Logger/LoggerFactoryTest.php
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * @dataProvider provideLogTypesAndClasses
+     */
+    public function test_it_creates_logger_for_log_type(Logs $logs, string $expectedLogClass): void
+    {
+        $factory = new LoggerFactory(
+            new MetricsCalculator(),
+            new Filesystem(),
+            LogVerbosity::NORMAL,
```
theofidry: if the log verbosity may play a role (which is what `test_it_creates_a_bade_logger_on_low_verbosity` hints at), then it may be worth adding it as an input

tests/phpunit/Logger/LoggerFactoryTest.php
BackEndTea: I find `provide*` to be better for a method name, as its an action, where `*Provider` is more suited for a class name.

theofidry:
I agree with you and it was my first choice as well. But you sometimes have
both a generator and a data provider (which is a more specific generator).

It’s not a big deal though, but I would have preferred the naming to Ben
consistent

tests/phpunit/Logger/LoggerFactoryTest.php
BackEndTea: I don't like doing so for mocks, it makes it harder to track IMO. Once we can use PHPUnit 8.4+ then we can use it as a stub explicitly. In that case it should be fine to do so.

tests/phpunit/Logger/LoggerFactoryTest.php
theofidry: I agree with you and it was my first choice as well. But you sometimes have
both a generator and a data provider (which is a more specific generator).

It’s not a big deal though, but I would have preferred the naming to Ben
consistent

src/Logger/LoggerFactory.php
```diff
@@ -0,0 +1,188 @@
...
+     * @var bool
+     */
+    private $onlyCoveredMode;
+
+    public function __construct(
+        MetricsCalculator $metricsCalculator,
+        Filesystem $filesystem,
+        string $logVerbosity,
+        bool $debugMode,
+        bool $onlyCoveredMode
```
theofidry: `$onlyCoveredMode` or `$onlyCoveredCode`?

tests/phpunit/Logger/LoggerFactoryTest.php
```diff
@@ -0,0 +1,199 @@
...
+use Infection\Logger\SummaryFileLogger;
+use Infection\Logger\TextFileLogger;
+use Infection\Mutant\MetricsCalculator;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Filesystem\Filesystem;
+
+final class LoggerFactoryTest extends TestCase
+{
+    public function test_it_creates_no_logger_for_low_verbosity(): void
```
theofidry: `test_it_does_not_create_any_logger_for_no_verbosity_level`?

src/Logger/LoggerFactory.php
BackEndTea: I think we use `onlyCoveredCode` in other places, so i guess we should do so here as well.

theofidry:
Closed in favour of #908

PR #895, opened 2019-12-15
Tweak the PHP-Parser ParseException

theofidry:
Upon failure, the `FileParser` would throw a `ParserException` exception. There is several things however that I would like to change here:

- The current `ParserException` does not give a friendly error message if the file real path could not be retrieved. I do not thing it is a
  problem in practice since as far as I'm aware we don't handle symlinks, but it does not cost much to fix that either...
- Rename `ParserException` for a more expressive form `UnparseableFile`
- Move the old `ParserException` next to the parser rather than a generic `Exception` directory

### DIFF
--- a/src/Mutation/FileParser.php
+++ b/src/Mutation/FileParser.php
@@ -35,7 +35,6 @@ declare(strict_types=1);
 
 namespace Infection\Mutation;
 
-use Infection\Mutant\Exception\ParserException;
 use PhpParser\Node;
 use PhpParser\Parser;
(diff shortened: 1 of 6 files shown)


src/Mutation/FileParser.php
```diff
@@ -54,16 +53,21 @@ public function __construct(Parser $parser)
     }
 
     /**
-     * @throws ParserException
+     *@throws UnparsableFile
```
sanmai:
```suggestion
     * @throws UnparsableFile
```

PR #896, opened 2019-12-15
Rework the traverser implementation

theofidry:
The goal of this PR is to try to make the traverser we use slightly more robust. I identified to sources of problems with the current implementation:

- We are extending a class instead of relying on the interface, which means we are more likely to be impacted by a change of the parent class
- The parent class API is not adapted to our usage (prioritizing visitors or removing them)

I think there is two good ways to approach the problem:

1. Make use an adapter (the choice of this PR)
2. Use the standard implementation and move the priority related concerns to the traverser factory instead

Regarding 1., since it's the implementation chosen here I will let the code speak for itself.

Regarding 2., I was thinking something along the lines of:

```
final class NodeTraverserFactory
{
    /**
     * @param NodeVisitor[] $extraVisitors
     */
    public function create(array $extraVisitors): PrioritizedVisitorsNodeTraverser
    {
        $traverser = new NodeTraverser();

        $extraVisitors[50] = new NotMutableIgnoreVisitor();
        $extraVisitors[40] = new ParentConnectorVisitor();
        $extraVisitors[30] = new FullyQualifiedClassNameVisitor();
        $extraVisitors[20] = new ReflectionVisitor();

        krsort($extraVisitors, SORT_NUMERIC);

        foreach ($extraVisitors as $visitor) {
            $traverser->addVisitor($visitor);
        }
        
        return $traverser;
    }
}
```

I think however this approach is more fragile since the traverser remains mutable and more error prone (you could add the same visitors again for example it would not fail like it currently [prior and in this PR still] does).


src/Mutation/NodeTraverserFactory.php
```diff
@@ -49,17 +50,17 @@ final class NodeTraverserFactory
     /**
      * @param NodeVisitor[] $extraVisitors
      */
-    public function create(array $extraVisitors): PriorityNodeTraverser
+    public function create(array $extraVisitors): PrioritizedVisitorsNodeTraverser
```
theofidry: The choice here is to make `PrioritizedVisitorsNodeTraverser` a full-pledge adapter and no longer depend on `NodeTraverserInterface` when consuming the traverser. It is not a final decision but there is no need to be more generic for now and facilitates a few things, since we can be sure about the traverser downstream when using it.

src/Mutation/PrioritizedVisitorsNodeTraverser.php
```diff
@@ -0,0 +1,122 @@
...
+    private $visitors = [];
+
+    private $traverser;
+
+    public function __construct(NodeTraverserInterface $decoratedTraverser)
+    {
+        $this->traverser = $decoratedTraverser;
+    }
+
+    public function addPrioritizedVisitor(NodeVisitor $visitor, int $priority): void
```
theofidry: So basically here we ditch the original interface API for ours, which allows to simplify things by:

- requiring the priority
- not supporting undesired cases (e.g. removing a visitor) which would only bloat our implementation and require us to test this scenario which we would actually not use in practice

tests/phpunit/Fixtures/Mutation/FakeNodeTraverser.php
```diff
@@ -0,0 +1,40 @@
...
+declare(strict_types=1);
+
+namespace Infection\Tests\Fixtures\Mutation;
+
+use LogicException;
+use PhpParser\Node;
+use PhpParser\NodeTraverserInterface;
+use PhpParser\NodeVisitor;
+
+final class FakeNodeTraverser implements NodeTraverserInterface
```
theofidry: those `Fake*` classes allow to simplify some tests by not requiring to declare a mock with no expected calls. So it saves a bit of boiler template (which makes tests harder to read) and also speeds up things a bit (mocks are slower)

tests/phpunit/Mutation/PrioritizedVisitorsNodeTraverserTest.php
```diff
@@ -0,0 +1,247 @@
...
+        $this->traverser = new PrioritizedVisitorsNodeTraverser(new FakeNodeTraverser());
+    }
+
+    public function test_it_cannot_add_visitors_without_a_priority(): void
+    {
+        try {
+            $this->traverser->addVisitor(new FakeVisitor());
+
+            $this->fail('Expected an exception to be thrown.');
+        } catch (DomainException $exception) {
```
BackEndTea: How about using expectException, and expectExceptionMessage here?

src/Mutation/PrioritizedVisitorsNodeTraverser.php
```diff
@@ -0,0 +1,122 @@
...
+
+    public function removeVisitor(NodeVisitor $visitor): void
+    {
+        throw new DomainException('Removing a visitor is not supported.');
+    }
+
+    /**
+     * @return array<int,NodeVisitor>
+     */
+    public function getVisitors(): array
```
BackEndTea: Is this method only used in tests?

theofidry: Yes. See the discussion [here](https://github.com/infection/infection/pull/893#discussion_r358302539). I'm not too happy about it but I didn't find a satisfying solution other than removing the tests for the factory...

tests/phpunit/Mutation/PrioritizedVisitorsNodeTraverserTest.php
theofidry: I really don't like the expect exception because it hides what exact statement is expected to fail.

In other cases where you want to check the status code or previous, the PHPUnit API for it makes it more awkward than anything IMO. I also recall Sebastian wanting to removing at some point but it would just be too much of a BC break for PHPUnit

theofidry:
@infection/core ?

src/Mutation/PrioritizedVisitorsNodeTraverser.php
```diff
@@ -0,0 +1,123 @@
...
+     */
+    public function traverse(array $nodes): array
+    {
+        // This could on theory be optimized by doing it only once. This would however require to
+        // keep track of whether or not visitors have been already sorted and reset that state
+        // when adding a visitor.
+        // Since in practice this method is used only once this would be within the real of
+        // nano-optimization hence not worth pursuing since would introduce more state and make
+        // the tests of this class unnecessarily more complex.
+        krsort($this->visitors, SORT_NUMERIC);
```
maks-rafalko: if this is really a "nano-optimization" as you say, would it make sense to move it back to `addPrioritizedVisitor()` and don't have this sorting in 2 places with these long comments? :)

Personally, I think we are ok to call `krsort` 5-10 times while mutation testing can take hours

src/Mutation/PrioritizedVisitorsNodeTraverser.php
```diff
@@ -0,0 +1,123 @@
...
+use const SORT_NUMERIC;
+
+/**
+ * @internal
+ * @final
+ */
+class PrioritizedVisitorsNodeTraverser implements NodeTraverserInterface
+{
+    /**
+     * @var array<int,NodeVisitor>
```
maks-rafalko:
```suggestion
     * @var array<int, NodeVisitor>
```

sometimes it's `array<int, NodeVisitor>`, sometimes `array<int,NodeVisitor>`.  Please add a space after comma

tests/phpunit/Mutation/PrioritizedVisitorsNodeTraverserTest.php
```diff
@@ -0,0 +1,247 @@
...
+        $priority2 = 10;
+
+        $expectedVisitors = [
+            $priority2 => $visitor2,
+            $priority1 => $visitor1,
+            $priority0 => $visitor0,
+        ];
+
+        // With the given fixtures above and while keeping the same visitor priority pair, then
+        // for each possible permutations of the possible order of those pairs should yiel an
```
maks-rafalko:
```suggestion
        // for each possible permutations of the possible order of those pairs should yield an
```

tests/phpunit/Fixtures/PhpParser/RecordedVisitor.php
```diff
@@ -0,0 +1,22 @@
...
+
+use PhpParser\NodeVisitorAbstract;
+
+final class RecordedVisitor extends NodeVisitorAbstract
+{
+    private $records;
+
+    public function __construct(array &$records)
+    {
+        $this->records = &$records;
```
maks-rafalko: This one create a recursion in the tree https://3v4l.org/J751R

I can't right now come up with an idea how it can influence us. Memory leaks in tests? 

Is there a way to do it in another way?

src/Mutation/PrioritizedVisitorsNodeTraverser.php
theofidry: To be honest it should be `NodeVisitor[]` in those case, bit I agree I prefer with a space

tests/phpunit/Fixtures/PhpParser/RecordedVisitor.php
theofidry: I'm going to check for a potential memory leak. That said note that this is only for testing purposes (to check that the visitors are called in the right order) and that this was already existing code (I moved it from a private method of the test case into a dedicated fixture class)

src/Mutation/PrioritizedVisitorsNodeTraverser.php
theofidry: I guess yeah

src/Mutation/PrioritizedVisitorsNodeTraverser.php
theofidry: Hm I stand corrected, `NodeVisitor[]` implies it's a list which is not the case since it may have holes

tests/phpunit/Fixtures/PhpParser/RecordedVisitor.php
theofidry: Couldn't find any evidence of a memory leak due to this so I think it's ok. It's not like I like this, but it's what was there and I didn't find a way to get around it either

theofidry:
@maks-rafalko rebased & addressed your comments

PR #898, opened 2019-12-17
Extract the file mutation generator into its own class

theofidry:
Continue to trim down `MutationGenerator` by extracting the generation of the mutation generations into its own class. There is 3 direct benefits of this:

- `FileMutationGenerator` is simpler (as a class), hence easier to test
- `FileMutationGenerator` can be registered as a service (helps to move towards having the `MutationGenerator` as one)
- Simplify drastically `MutationGenerator`


tests/phpunit/Mutation/FileMutationGeneratorTest.php
```diff
@@ -0,0 +1,351 @@
...
+use InvalidArgumentException;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use function Safe\sprintf;
+use Symfony\Component\Finder\SplFileInfo;
+
+final class FileMutationGeneratorTest extends TestCase
+{
+    private const FIXTURES_DIR = __DIR__ . '/../Fixtures/Files';
+    /**
```
maks-rafalko: please add empty line here

theofidry: meh indeed, good catch

PR #899, opened 2019-12-18
Move code coverage base path to the configuration

theofidry:


src/Configuration/ConfigurationFactory.php
```diff
@@ -150,6 +154,23 @@ public function create(
...
+
+        if ($existingCoveragePath === null) {
+            return $tmpDir;
+        }
+
+        if (Path::isAbsolute($existingCoveragePath)) {
+            return $existingCoveragePath;
+        }
+
+        return sprintf('%s/%s', getcwd(), $existingCoveragePath);
```
theofidry: Not fond of the `getcwd()` here, should it be made relative to the config file? This would be a change, but IMO would make more sense

src/Configuration/Configuration.php
```diff
@@ -210,9 +210,9 @@ public function getTestFrameworkOptions(): ?string
         return $this->testFrameworkOptions;
     }
 
-    public function getExistingCoveragePath(): ?string
```
maks-rafalko: I would love to retain this name as it is clear that we are working with the path to coverage provided by user, hence existing.

Probably `getExistingCoverageBasePath()`?

`getCoverageBasePath()` is too common and easy to mix up with another `retrieveCoverageBasePath()` method.

src/Configuration/ConfigurationFactory.php
maks-rafalko: sounds good to me, as it would be in line with other things that are related to the config file

src/Configuration/Configuration.php
theofidry: 👍

PR #902, opened 2019-12-18
Remove unnecessary phpdoc

theofidry:
⚠️  _Controversial PR_ ⚠️

IMO, phpdoc in PHP is here only to complement the typesystem, e.g. to annotate `Node[]` instead of `array` or provide more explanation regarding an element, e.g. about the effect of one parameter.

As a result, in the following case:

```php
class Foo {
  private $x;

  public function __construct(X $) {
    $this->x = $x;
  }

  public function getX(): X {
    return $this->x;
  }
}
``` 

I find excessive to add:

```php
/**
 * @var X
 */
private $x
```

On the same token, if you have:

```php
class Foo {
  private $x = 0;
}
```

I find unnecessary to add `@var int` unless you want to expand the type to put `@var int|false` for example.

To me there is three reasons for this change:

- It is unneeded (an SA tool or IDE will understand it) and does not improve readability
- Only increase redundancy, i.e. make the code more noisy
- Provides no guarantee meaning if you change the typehint in the constructor or the `@param` in the constructor, you need to rely on the SA to catch that the `@var` has not been updated

In light of this, I'm proposing the following PR where I remove the `@var` _only in the two scenarios listed above_.

### DIFF
--- a/src/Mutation.php
+++ b/src/Mutation.php
@@ -73,8 +73,8 @@ final class Mutation implements MutationInterface
     private $tests;
 
     /**
-     * @param Node[]   $originalFileAst
-     * @param (string|int|float)[]   $attributes
+     * @param Node[] $originalFileAst
+     * @param array<string|int|float> $attributes
      * @param Node|Node[] $mutatedNode
      */
     public function __construct(
--- a/src/Mutation.php
+++ b/src/Mutation.php
@@ -46,10 +46,11 @@ use PhpParser\Node;
  */
 final class Mutation implements MutationInterface
 {
-    /**
-     * @var Mutator
-     */
+    private $originalFilePath;
     private $mutator;
+    private $mutatedNodeClass;
+    private $mutatedNode;
+    private $mutationByMutatorIndex;
 
     /**
      * @var array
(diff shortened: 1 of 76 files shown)


sanmai:
Last time I checked Psalm was very reliant on these phpdocs. It could be very annoying to re-add these comments if we ever want to use Psalm. 

(FWIW it helped me to find bugs nothing else noticed.)

theofidry:
If we want to a tool like Rector can introduce them back without any
problem. I’ve also stuck to those two simple cases for that reason

BackEndTea:
How does tooling deal with me writing to a variable?

e.g.

```php
class Config
{
    privage $dir;

    public function __construct(string $dir)
    {
        $this->dir = $dir;
    }
    public function doThing(): void
    {
        $this->dir =null;
    }
}
```

will the `$this->dir= null` result in an error?

src/Mutation.php
```diff
@@ -116,6 +102,9 @@ public function getMutator(): Mutator
         return $this->mutator;
     }
 
+    /**
+     * @return (string|int|float)[]   $attributes
```
BackEndTea: I'd prefer this to be `array<string|int|float>`

theofidry:
Both phpStan and phpStorm deals with it. Not 100% sure about phpStorm
because it does on mine but I have the PHP inspection plugin as well so
might be from it

theofidry:
No problem

theofidry:
Since it's a kind of controversial PR and I don't want us to waste too much time on this, I would like to propose a simple vote. Just add a 👍 or 👎 on this comment and in say 2 days time go with the result.

Just as a reminder, the choice is between:

- Removing unnecessary `@var` declarations [in the simple case described here](https://github.com/infection/infection/pull/902#issue-354646524)
- Consistently and systematically add the `@var` declarations to the properties

Also worth mentioning that regardless of our choice it won't hurt any migration towards PHP 7.4 (tools like Record _will_ be able to automatically add the types to the properties)

PR #906, opened 2019-12-18
Move code coverage instantation from InfectionCommand to a Factory

theofidry:
See https://github.com/infection/infection/projects/1#card-30761480

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -68,6 +68,7 @@ use Infection\TestFramework\Coverage\CachedTestFileDataProvider;
 use Infection\TestFramework\Coverage\CoverageDoesNotExistException;
 use Infection\TestFramework\Coverage\LineCodeCoverage;
 use Infection\TestFramework\Coverage\XMLLineCodeCoverage;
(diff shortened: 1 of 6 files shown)


src/Command/InfectionCommand.php
```diff
@@ -355,14 +356,15 @@ private function runMutationTesting(TestFrameworkAdapter $adapter): void
 
         $processBuilder = new MutantProcessBuilder($adapter, $this->versionParser, $config->getProcessTimeout());
 
-        $codeCoverageData = $this->getCodeCoverageData($this->testFrameworkKey, $adapter);
```
maks-rafalko: [This method definition](https://github.com/infection/infection/blob/131d75c9e71c3aa690c05c8452ece5c19c22330e/src/Command/InfectionCommand.php#L491-L504) should be removed since it's unsed after this change

maks-rafalko:
Build [is still broken](https://travis-ci.org/infection/infection/jobs/626956104#L444)

src/Command/InfectionCommand.php
theofidry: 👍

PR #907, opened 2019-12-19
Remove the need to register `@final` classes to `ProjectCodeProvider::NON_FINAL_EXTENSION_CLASSES`

theofidry:


tests/phpunit/AutoReview/PhpDoc/PHPDocParser.php
```diff
@@ -0,0 +1,67 @@
...
+use const SORT_STRING;
+
+final class PHPDocParser
+{
+    /**
+     * Parses the given PHP doc and returns the list of tags found.
+     *
+     * @return string[]
+     */
+    public function parse(string $phpDoc): array
```
theofidry: this could be replaced by a library such as phpdoc. However given how simple our use case is, I would rather go with this in-house implementation which is 10x times faster and lighter

tests/phpunit/AutoReview/PhpDoc/PHPDocParserTest.php
```diff
@@ -0,0 +1,136 @@
...
+     * @dataProvider phpDocProvider
+     */
+    public function test_it_can_parse_phpdoc(string $phpDoc, array $expected): void
+    {
+        $actual = (new PHPDocParser())->parse($phpDoc);
+
+        $this->assertSame($expected, $actual);
+    }
+
+    public function phpDocProvider(): Generator
```
theofidry: I added a bunch of tests more to capture the current behaviour than actual need. We mostly only have `@final` and `@internal` tags to parse...

maks-rafalko:
Build [is failing](https://travis-ci.org/infection/infection/jobs/627118171#L390)

PR #909, opened 2019-12-19
Migrate to thecodingmachine/safe

theofidry:


src/Console/InfectionContainer.php
```diff
@@ -219,7 +220,7 @@ public static function create(): self
                 return new VersionParser();
             },
             Lexer::class => static function (): Lexer {
-                return new Lexer\Emulative([
+                return new Emulative([
```
theofidry: @ TomasVotruba is it expected that it does this change although it was not requested?

theofidry:
@TomasVotruba I've tried to run Rector as recommended [here](https://github.com/thecodingmachine/safe#automated-refactoring) but I can see a couple of issues:

- The `use` statements are not applied correctly. It uses `use` only even for functions e.g. `use Safe\sprintf` instead of `use function Safe\sprintf`
- It adds `use` statement for classes which were not fully imported

Regarding the first one it's clearly a bug, but is it from https://github.com/thecodingmachine/safe/blob/master/rector-migrate.yml or from Rector?

Regarding the second one is it an expected behaviour? I would have preferred that Rector does not alter this kind of things it makes refactoring harder (of course a dedicated set could be done for it).

src/Console/InfectionContainer.php
TomasVotruba: See https://github.com/rectorphp/rector#import-use-statements

theofidry:
>How many cases is it? If I should give it a high priority or you can fix manually?

It's all the import cases of https://github.com/thecodingmachine/safe. I would just like to know if you have an idea of whether it's from Rector itself or most likely https://github.com/thecodingmachine/safe/blob/master/rector-migrate.yml

TomasVotruba:
Config looks ok, it's a function name = Rector bug

theofidry:
I'll be closing this PR for now as it looks like more work needs to be done upstream. I'll try to have a look at that Rector bug

PR #910, opened 2019-12-20
Move skipCoverage to the config

theofidry:


src/Configuration/ConfigurationFactory.php
```diff
@@ -201,18 +207,18 @@ private static function retrieveExistingCoverageBasePath(
         ?string $existingCoveragePath,
         string $configDir,
         string $tmpDir
-    ): string {
```
BackEndTea: Does phpstan understand annotations like: ` @return array{0:bool,1:string}`?

theofidry: At least it doesn't seem to complain about it 👍

src/Configuration/ConfigurationFactory.php
```diff
@@ -113,6 +113,12 @@ public function create(
 
         $testFramework = $testFramework ?? $schema->getTestFramework() ?? TestFrameworkTypes::PHPUNIT;
 
+        [$skipCoverage, $existingCoverageBasePath] = self::retrieveExistingCoverageBasePath(
```
maks-rafalko: IMO, now the method name conflicts with return values. It doesn't return path, but some flag as well.

What about setting it outside the method as:

```php
$skipCoverage = $existingCoverageBasePath !== null;
```

theofidry: 👍

PR #914, opened 2019-12-20
Move more services from the InfectionCommand to the InfectionContainer

theofidry:
Depends on #913


theofidry:
@maks-rafalko if you have some time this is an easy one

PR #916, opened 2019-12-20
Make command steps dependencies more explicit (part1: services)

theofidry:
Depends on #915

The goal here is to make more explicit the dependencies of each steps. Since some dependencies, especially the input/output might be a bit more tricky, I'm doing the first part that touches the services only.


theofidry:
Closed by:

- #911
- #912 
- #913
- #914
- #915
- #932

PR #817, opened 2019-11-05
Add PHP 7.4 to Travis CI

sanmai:
- Add a testing stage for PHP 7.4 + xdebug
- Require build passing for 7.4

This PR:
- [x] Hopefully does something about #774

PR #917, opened 2019-12-20
Tweak the Travis build

theofidry:


.ci/safe-bash-scripts.md
```diff
@@ -0,0 +1,219 @@
+# Safe bash script configuration instructions
```
theofidry: Note sure if you want to keep it or not. I find it useful to have but I'm also fine with omitting it

.ci/travis-functions.sh
```diff
@@ -5,38 +5,87 @@
 # - https://docs.travis-ci.com/user/languages/php#Disabling-preinstalled-PHP-extensions
 # - https://docs.travis-ci.com/user/languages/php#Custom-PHP-configuration
 
-config="/home/travis/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini"
+set -Eeuo pipefail
```
theofidry: Purely aesthetic changes, the behaviour is exactly the same

.travis.yml
```diff
@@ -4,7 +4,7 @@ sudo: false
 
 env:
   global:
-    - INFECTION_FLAGS='--threads=4 --min-msi=67 --min-covered-msi=75 --coverage=coverage --log-verbosity=none'
+    - INFECTION_FLAGS='--threads=$(nproc) --min-msi=67 --min-covered-msi=75 --coverage=coverage --only-covered --log-verbosity=none'
```
theofidry: I think this will take better advantage of the Travis config

theofidry: Any reason why we don't have the `--only-covered` flag by default?

.travis.yml
```diff
@@ -4,7 +4,7 @@ sudo: false
 
 env:
   global:
-    - INFECTION_FLAGS='--threads=4 --min-msi=67 --min-covered-msi=75 --coverage=coverage --log-verbosity=none'
+    - INFECTION_FLAGS='--threads=$(nproc) --min-msi=67 --min-covered-msi=75 --coverage=coverage --log-verbosity=none'
```
theofidry: I think it's better to adjust to Travis CI here instead of hardcoding the value

.travis.yml
```diff
@@ -35,30 +35,28 @@ jobs:
       install:
         - composer install $COMPOSER_FLAGS
       script:
-        - make autoreview --keep-going
```
theofidry: As far as I can tell the `--keep-going` is not doing anything

.travis.yml
```diff
@@ -67,13 +65,16 @@ jobs:
...
-          INFECTION_PR_FLAGS=$(get-infection-pr-flags);
-          echo "INFECTION_PR_FLAGS=$INFECTION_PR_FLAGS";
+          INFECTION_PR_FLAGS=$(get_infection_pr_flags);
+
+          echo "Flags used: ${INFECTION_FLAGS} ${INFECTION_PR_FLAGS};";
 
           if [[ "$DRIVER" != "phpdbg" ]]; then
             bin/infection $INFECTION_FLAGS $INFECTION_PR_FLAGS;
+          elif [[ "$DRIVER" == "none" ]]; then
+            exit 0
```
theofidry: I don't think there is any reason to run infection without phpdbg for the driver "none": it already run all the tests and there is a dedicated build for the phpdbg driver

.travis.yml
```diff
@@ -106,7 +107,7 @@ jobs:
 
     -
       <<: *STANDARD_TEST_JOB
-      php: 7.4snapshot
+      php: 7.4
```
theofidry: no need for a snapshot anymore

.travis.yml
```diff
@@ -35,30 +35,28 @@ jobs:
...
-                git clone --single-branch --branch=v1.0.6 --depth=1 https://github.com/krakjoe/pcov
-                cd pcov
-                phpize
-                ./configure
-                make clean install
-                echo "extension=pcov.so" > $HOME/.phpenv/versions/$TRAVIS_PHP_VERSION/etc/conf.d/pcov.ini
-                cd $TRAVIS_BUILD_DIR
-              fi
+            if [[ "$DRIVER" == "pcov" ]]; then
+              pecl install pcov;
```
theofidry: simplified the pcov install process

.travis.yml
```diff
@@ -67,37 +65,39 @@ jobs:
...
           else
             phpdbg -qrr bin/infection $INFECTION_FLAGS $INFECTION_PR_FLAGS;
           fi
 
       after_success:
         - bash <(curl -s https://codecov.io/bash)
 
-    -
-      <<: *STANDARD_TEST_JOB
+    - <<: *STANDARD_TEST_JOB
```
theofidry: I've reordered the jobs from fastest to slowest

.travis.yml
theofidry: Note to myself: should be place _above_ the previous condition, otherwise will never be executed

.travis.yml
sanmai: Yes. If there's a single failure, you may want to know if it is the only failure. Because the first failure may not point to the root of the problem.

.travis.yml
BackEndTea: I think we can also add the other 'steps' for 7.4.

e.g. with deps and drivers

theofidry:
As far as I know Infection is not working with PHP 7.4 yet so I would
prefer to wait on @sanmai PR for that

.travis.yml
sanmai: There's #817 to do something like that, but there's also some difficulties which prevent builds from succeeding. I can imagine this PR isn't about that.

.travis.yml
sanmai: It's a kind of a good idea, but really the throughput does not scale linearly (it was saturating very quick last time I checked) so we may be better off with just 4 processes, letting the CI system use the remaining resources.

PR #921, opened 2019-12-21
Cosmetic visitor changes

theofidry:


src/Visitor/ReflectionVisitor.php
```diff
@@ -160,27 +160,14 @@ private function isInsideFunction(Node $node): bool
...
-        }
-
-        if ($this->isFunctionLikeNode($parent)) {
-            return true;
-        }
-
-        return $this->isInsideFunction($parent);
+        return $parent->getAttribute(self::IS_INSIDE_FUNCTION_KEY)
+            || $this->isFunctionLikeNode($parent)
+            || $this->isInsideFunction($parent)
```
sanmai: This can be troublesome MSI-wise.

src/Visitor/ReflectionVisitor.php
```diff
@@ -160,27 +160,14 @@ private function isInsideFunction(Node $node): bool
...
-        if ($node instanceof Node\Stmt\ClassMethod) {
-            return true;
-        }
-
-        if ($node instanceof Node\Expr\Closure) {
-            return true;
-        }
-
-        return false;
+        return $node instanceof Node\Stmt\ClassMethod || $node instanceof Node\Expr\Closure;
```
sanmai: This also add a bunch of mutations you'll have to test against.

theofidry:
@sanmai I'm not sure to understand your worries: the two conditions changed are strictly identical semantically (it's a pure CS change). I can't easily compare before & after MSI-wise but we're still 4.5% the aimed target so I don't think there was any regression here

sanmai:
There wasn't a regression yet, but this change opens us up for them. Say, before you had a test for line 163, and a test for line 167. And then you wouldn't even need Infection to tell you that you're covering all relevant code paths. With your proposal it won't be no longer possible with a coverage report _alone_. You would have to use a tool like Infection to see if you're covering all code paths. Otherwise put, this change is for the worse, not the best.

theofidry:
Is it a real problem though? In practice tools like pcov should help out to get a more accurate result, but overall I'm not sure you should adjust your code more than that to please a specific coverage report tool. Plus in our case we are always using Infection, so it's not like this stylistic change has any impact on us does it?

theofidry:
I mean otherwise I feel we should then have stronger checks against any `||` and `&&` used in expression, as your argument stands for any of them not the case just above

sanmai:
As I repeatedly try to convey, this is not a stylistic/CS change. It takes a second to open a coverage report, where it takes minutes to run Infection. And while for the coverage report you'll need to see it just once to ensure you're going the right way, in case of Infection you may not nail it from the first try.

See, it is the difference between _knowing_ and _guessing_. Before this change you could know you're doing all right. After this change you could only speculated that you're _probably_ doing it right. And all at the cost of, what, seven lines of code? Price well paid, I think.

theofidry:
I've reverted those changes as you requested as I don't feel strongly about it. But IMO if this matters for you, we should push further than that for example:

- banning ternary operators
- banning some logic operators in some expressions

The rational is: personally I don't really want to spend time thinking over if my stylistic choice of code (I can't call it any differently when the code path & behaviour is the same - maybe there is a more appropriate word for it but I don't know it) pleases or not various code coverage tools if there is no easy way to check against them in the IDE or other...

sanmai:
Thank you for the revert. I don't have a solid rule for all operators, and sometimes it may be better for readability sake to write it in a single line, but as disjunction goes, it is typically beneficial to split them the way it's done here.

src/Visitor/MutationsCollectorVisitor.php
```diff
@@ -59,24 +59,24 @@ final class MutationsCollectorVisitor extends NodeVisitorAbstract
...
     private $filePath;
-    private $fileAst;
+    private $fileNodes;
     private $codeCoverageData;
     private $onlyCovered;
 
     /**
      * @param Mutator[] $mutators
-     * @param  Node[] $fileAst
+     * @param Node[]    $fileNodes
```
maks-rafalko: since when we started doing it? I would like to avoid it as adding one more parameter with a longer type will lead to indenting **all* the rows to the right

theofidry: I think it's my IDE, looks like it's not something enforced by the CS fixer yet

PR #924, opened 2019-12-23
Tweak AbstractMutatorTestCase

theofidry:
- Rename `AbstractMutatorTestCase` to `BaseMutatorTestCase`
- Make protected methods finals
- Move some stateless services such as the parser and lexer as static properties

### DIFF
--- a/tests/phpunit/Mutator/AbstractMutatorTestCase.php
+++ b/tests/phpunit/Mutator/AbstractMutatorTestCase.php
@@ -61,16 +61,18 @@ use function Safe\sprintf;
 use function str_replace;
 use function substr;
 
-abstract class BaseMutatorTestCase extends TestCase
+abstract class AbstractMutatorTestCase extends TestCase
 {
     /**
      * @var Mutator
      */
     protected $mutator;
+
     /**
      * @var Parser|null
      */
     private static $parser;
+
     /**
      * @var PrettyPrinterAbstract|null
      */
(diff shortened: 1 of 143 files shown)


tests/phpunit/Mutator/AbstractMutatorTestCase.php
```diff
@@ -135,46 +148,69 @@ protected function mutate(string $code, array $settings = []): array
...
-            $mutants[] = $prettyPrinter->prettyPrintFile($mutatedStatements);
-
-            $traverser->removeVisitor($mutatorVisitor);
+            $mutants[] = $this->getPrinter()->prettyPrintFile($mutatedStatements);
         }
 
         return $mutants;
     }
 
+    private function getParser(): Parser
```
sanmai: `private static`?

theofidry:
>I'm not sure about renaming Abstract*** to *** and amend so many files. I mean - it's ok in this case, but a) was it really needed? b) we have many other Abstract*** classes - are we going to rename them as well?

I admit it's more a stylistic choice than anything :) I can make arguments about not having suffxies in general such as `*Exception`, `*Interface`, `*Trait`, but the difference between `Abstract*` and `*Base` is IMO expressiveness, but admittedly in this case it's very subjective.

That said looking at it again, it might be better to take a decision on it and rename it everywhere instead of on a case per case basis like I'm doing

theofidry:
@maks-rafalko I'm reverting the renaming change: I'll open a dedicated PR for us to take a team decision on this and apply it on the whole codebase.

It appears there might be an issue related to making the properties static, I'll investigate it

PR #925, opened 2019-12-23
[Mutator] `Equal`

theofidry:
This PR:

- Renames `Equal` to `EqualOrIdentical` ⚠️ this is a BC break if no further work is done
- Cosmetic changes

Depends on #924


theofidry:
Closed in favour of #937

PR #926, opened 2019-12-23
[Mutator] `Assignment `

theofidry:
Depends on #924


theofidry:
Closed in favour of #937

Issue #922, opened 2019-12-22
Extract Test Framework Adapters to separate composer packages


maks-rafalko:
Work is based on the following resources

* [Principles of OOD](http://butunclebob.com/ArticleS.UncleBob.PrinciplesOfOod), Robert Martin
* [Principles of Package Design](https://www.apress.com/us/book/9781484241189), Matthias Noback

This is the RFC with detailed description of Package Design Principles applied to Infection, that show the benefits of splitting `infection/infection` to separate packages. The main topic here is Test Framework Adapters, but splitting Infection is not limited to them (see another [RFC for extracting Mutators](https://github.com/infection/infection/issues/669)).

For me, there are no doubts we will split it. The question is how deep we will go.

# Principles about package cohesion

## Common Reuse Principle

_Classes that are used together are packaged together_

We clearly violate this priciple. People who use only `PHPSpec` adapter should not download `Codeception`/`PHPUnit` adapters.

At the moment, `infection/infection` contains classes that are not all used by everyone who uses Infection package in their project, so it violates the _Common reuse principle_. It also contains classes (the same classes actually) that are not closed against the same kinds of changes, so the package violates the _Common closure principle_. More on it later.

### Dependencies of Test Framework

Another thing that shows we violate this package principle is that `infection/infection` [requires](https://github.com/infection/infection/blob/7adc1fa1fdab6f33734249787082f094e00f16a9/composer.json#L60) `symfony/yaml`. The key thing is that Infection has nothing to do with YAML. It's a dependency of `Codeception` and `PHPSpec` adapters, but not Infection itself. Each time you use only `PHPUnit` adapter (and don't use `Codeception`/`PHPSpec`), you have

* the code you don't really need, that is not executed
* one more reason to be affected by additional dependency
* one more possible conflicting dependency when install Infection via Composer; and so on

### Cohesion

When we extract each Test Framework Adapter to a separate package, they will be coherent: all the classes a separate package contains are about the same thing. When user needs `PHPUnit` adapter, all the classes from `infection/phpunit-adapter` will be used together. And *no classes* will be used and downloaded from other adapters.

## The Common closure principle

* _The classes in a package should be closed together against the same kinds of changes. A change that affects a package affects all the classes in that package._
* _The primary justification for this principle is that we want a change to be limited to the smallest number of packages possible._

[...]

PR #928, opened 2019-12-25
Get rid of MutantInterface in TestFrameworkAdapter interface

maks-rafalko:
First step to extract Test Framework Adapters out of the core. See https://github.com/infection/infection/issues/922

This is needed to decouple `TestFrameworkAdapter` from the chain
of several interfaces and classes

Before:
`TestFrameworkAdapter -> MutantInterface -> MutationInterface -> Mutator concrete class -> MutatorConfig`

After:
`TestFrameworkAdapter` independent interface

* It allows to make only 1 interface public - `TestFrameworkAdapter` instead of making public `TestFrameworkAdapter`, `MutantInterface`, `MutationInterface`, `Mutator`
* Related to https://github.com/infection/infection/pull/872. Without this PR, #872 can not be merged, because in this case we will have to make public a concrete `Mutation` class, which is not what we want

### DIFF
--- a/src/TestFramework/AbstractTestFrameworkAdapter.php
+++ b/src/TestFramework/AbstractTestFrameworkAdapter.php
@@ -35,9 +35,9 @@ declare(strict_types=1);
 
 namespace Infection\TestFramework;
 
-use Infection\Mutant\MutantInterface;
 use Infection\TestFramework\Config\InitialConfigBuilder;
 use Infection\TestFramework\Config\MutationConfigBuilder;
+use Infection\TestFramework\Coverage\CoverageLineData;
 use Infection\Utils\VersionParser;
 use InvalidArgumentException;
 use Symfony\Component\Process\Process;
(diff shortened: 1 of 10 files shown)


src/TestFramework/AbstractTestFrameworkAdapter.php
```diff
@@ -99,11 +99,28 @@ public function getInitialTestRunCommandLine(
...
      *
+     * @param CoverageLineData[] $coverageTests
+     *
      * @return string[]
      */
-    public function getMutantCommandLine(MutantInterface $mutant, string $extraOptions): array
-    {
-        return $this->getCommandLine($this->buildMutationConfigFile($mutant), $extraOptions, [], false);
+    public function getMutantCommandLine(
+        array $coverageTests,
```
sanmai: Passing arrays around might need be worked around later.

maks-rafalko: Could you please elaborate on that? this is an array of `CoverageLineData`, not just dumb native arrays

sanmai: Yes, yet it can be copied on write still. It is not as bad as an array of arrays. In a perfect world we never need to pass arrays as arguments, but you can do only as much. So that I'm not to say there's anything to be done right now, this also could be overengineering and/or premature optimization.

maks-rafalko: Got it, thanks

maks-rafalko:
Thank you @sanmai

PR #931, opened 2019-12-25
Mutator tests 💅

theofidry:
This PR:

- Rename the test method `test_mutator` to `test_it_can_mutate` which is a bit more conventional
- Use the default value `[]` instead of `null`: it should be possible to remove the `null` value completely later on without too much work
- Rename the data provider `provideMutationCases` to `mutationsProvider` (a bit more conventional)
- Move from array data providers to generators

### DIFF
--- a/tests/phpunit/Mutator/Unwrap/UnwrapArrayChangeKeyCaseTest.php
+++ b/tests/phpunit/Mutator/Unwrap/UnwrapArrayChangeKeyCaseTest.php
@@ -41,14 +41,16 @@ use Infection\Tests\Mutator\AbstractMutatorTestCase;
 final class UnwrapArrayChangeKeyCaseTest extends AbstractMutatorTestCase
 {
     /**
-     * @dataProvider provideMutationCases
+     * @dataProvider mutationsProvider
+     *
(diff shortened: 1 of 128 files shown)


tests/phpunit/Mutator/Unwrap/UnwrapArrayChangeKeyCaseTest.php
```diff
@@ -41,14 +41,16 @@
 final class UnwrapArrayChangeKeyCaseTest extends AbstractMutatorTestCase
 {
     /**
-     * @dataProvider provideMutationCases
+     * @dataProvider mutationsProvider
+     *
+     * @param string|string[] $expected
      */
-    public function test_mutator($input, $expected = null): void
+    public function test_it_can_mutate(string $input, $expected = []): void
```
maks-rafalko: What about changing `null` to `[]` inside `\Infection\Tests\Mutator\AbstractMutatorTestCase::doTest`?

theofidry: Done

PR #932, opened 2019-12-25
Introduce the Engine

theofidry:
Move the core elements of the `InfectionCommand` into its dedicated `Engine` class. The ultimate goal is to make the `InfectionCommand` a very thin IO adapter orchestrating the engine.

I am not too happy yet about the state & dependencies of both the command and engine, but like for the previous PRs: a small step after the other.


theofidry:
Failure is expected

theofidry:
@maks-rafalko this one should be an easy/quick one. It's just moving code around

src/Command/InfectionCommand.php
```diff
@@ -205,17 +201,65 @@ protected function configure(): void
...
+            $mutationGenerator,
+            $mutationTestingRunner,
+            $constraintChecker,
+            $this->consoleOutput,
+            $metricsCalculator
+        );
+
+        $result = $engine->execute((int) $this->input->getOption('threads'));
+
+        return (int) !$result;
```
BackEndTea: I'd prefer to make this slightly more explicit and use a ternary operator, e.g.:
```suggestion
        return $result ? 0 : 1;
```

theofidry: I'm not convinced tbh: I don't find using a ternary operator clearer plus it also involves a loose comparison which is not any better than a casting. Everything for a more verbose expression as well. But I'm making it to make you happy since it's not something I feel that strongly either

Issue #930, opened 2019-12-25
[RFC] Mono-repo or not


theofidry:
As per #922, with the discussion of splitting the test framework bridges, let's make a dedicated RFC about this topic.

Before getting started, it is worth mentioning that the _current_ approach is multi-repo. Indeed we have:

- [infection](https://github.com/infection/infection)
- [infection.github.io](https://github.com/infection/infection.github.io): a readonly "dump" of the site
- [infection/site](https://github.com/infection/site): the actual code of the documentation
- [infection/tests-checker](https://github.com/infection/tests-checker): bot for `infection/infection`
- [homebrew-infection](https://github.com/infection/homebrew-infection): the Homebrew tap

The advantages of the multi-repo is that it's extremely easy to set up: you don't need any additional tooling.

Now what about mono-repositories?

Pros:

- Will make easier any cross-repo changes:
  - if we change the test-framework interface, we can update all the bridges in a compatible way in one go
  - we can more easily have a way to generate parts of the doc or at least keep some elements in sync
  - you can update the code and the doc at the same time in one go
 - You make easier the release process: you can focus on only tagging `infection/infection` and the rest should be tagged at the same time. It avoids to have a lot of different releases for each individual packages the way amphp or Zend has

Cons:

- Requires extra tooling

Now @maks-rafalko also mentioned [this article](https://medium.com/@mattklein123/monorepos-please-dont-e9a279be011b). A few elements regarding the raised points there:

>Monorepo proponents will argue that when all code is present within a single repository, the likelihood of code duplication is small, and it’s more likely that different teams will collaborate together on shared infrastructure.
>Here is the ugly truth about even a medium size monorepo (and this will be a recurring theme throughout this section): it quickly becomes unreasonable for a single developer to have the entire repository on their machine, or to search through it using tools like grep. Thus, any monorepo that hopes to scale must provide two things

I don't know what "medium" size this person is talking about, but I'm aware of several mono-repos with 100-500kLoC and multiple languages and it's neither unreasonable or hard. The only question is how much you are ready to invest in the tooling - and even if you don't go mono-repo, as soon as you have different stacks & techs past a certain size you will need to seriously invest in it anyway.

>Theoretical benefit 2: Single build / no dependency management [...] (proceeds to explain this is a lie)

I won't argue against that: but it's not why we are suggesting a mono-repo either. It is certain though that past a certain size, you want to invest in some tooling to run only the necessary tests. That is however true regardless of the mono-repo or multi-repo choice.

[...]

PR #933, opened 2019-12-25
Extract Abstract TestFrameworkAdapter & Codeception to separate packages

maks-rafalko:
This PR:

* Removes `TestFrameworkAdapter` from the core package
* Uses `TestFrameworkAdapter` from a [separate package](https://github.com/infection/abstract-testframework-adapter) as it was explained with many details in https://github.com/infection/infection/issues/922

tf-adapter

(!) Note:

For now, it's a separate repository under `infection` organization. [After we decide](https://github.com/infection/infection/issues/930) what to use - monorepo or multi repos - we will merge it into monorepo or leave as is.

### Next Steps

- ~Register `infection/abstract-testframework-adapter` package in Packagist once we are agree with this PR~
- Extract PHPUnit, Codeception, PHPSpec Test Framework Adapters to their own packages. At this step, `infection/infection` will still require other packages in `composer.json`
- Write installer to automatically install needed Test Framework Adapter. At this step, `infection/infection`'s composer.json will not require any adapter by default anymore. Installation will be done by the Installer


maks-rafalko:
Hmm.. for now, I have no idea what's going on with the build. The log [is truncated](https://travis-ci.org/infection/infection/jobs/629464600#L577) and everything is ok locally.

theofidry:
@maks-rafalko does it make sense to separate the package as is? I mean isn't there more work to do regarding each adapters before doing the split?

For example right now all the coverage data is put in the framework adapters which is far from ideal since I would like to introduce a "Tracer" which is pretty much what the interface `LineCodeCoverage` is but offering more implementations (not necessarily framework specific)

maks-rafalko:
I'm not following the question tbh. 

No matter what needs to be done in each Test Framework, `infection/abstract-testframework-adapter` will be used in the same way - as shown on the image. Because it's an abstraction (`TestFrameworkAdapter`). Other concrete classes can change at any time, and that's ok.

I can't predict all the changes anyone want to introduce in the future, and I have no idea what the Tracer is.

If we will have any changes with coverage in the future - we will make them, I don't see any issues here. Moreover, in this [Tracer draft PR](https://github.com/infection/infection/pull/927), no Test Framework Adapters are touched.

theofidry:
If you require a cross change, e.g. a change in the adapter interface, you will need to change X packages not 1. But this is the discussion of #930 which is why I would like us to solve it first before jumping with PR which is basically going for the polyrepo approach described in #930

>I can't predict all the changes anyone wants to introduce in the future, and I have no idea what the Tracer is.
>If we will have any changes with coverage in the future - we will make them, I don't see any issues here. Moreover, in this Tracer draft PR, no Test Framework Adapters are touched.

The idea of the tracer is the existing `LineCoverageInterface`, i.e. the interface allowing you to trace (hence the name) the tests for the given class/piece of code. I would like to push this in order to no longer require code coverage to make Infection work, which is a mandatory piece for the support of some external test frameworks

theofidry:
@maks-rafalko for the code you putted the other repos is there any change other than the namespace? I have a few changes starting to pile up on the test frameworks in order to be able to invert the dependency of `LineCoverageInterface` (so that the framework adapters depends on it instead of vice-versa). 

Should I keep pilling them up and push them once we decided on #930 or should I push them here and depending on our decision here, we'll update the underlying package accordingly? (IMO it's easier to fix it _before_ #930)

maks-rafalko:
> for the code you putted the other repos is there any change other than the namespace

there is only 1 repository I've created so far - [`infection/abstract-testframework-adapter`](https://github.com/infection/abstract-testframework-adapter). So I didn't extract any of the adapters yet.

and it has only 2 files:

* `TestFrameworkAdapter` interface and
* `CoverageLineData` DTO (which can be removed and replaced with `array` untill we are done with refactoring, but then moved back under same or other name)


> `LineCoverageInterface` (so that the framework adapters depends on it instead of vice-versa)

I'm not entirely see the picture, but `infection/***-adapter` should not depend on `infection/infection`. I've described this with details & pictures in ["The Stable Dependencies Principle (SDP)"](https://github.com/infection/infection/issues/922) section. So, probably this interface (only abstraction) should be in `infection/abstract-test-framework-adapter`? But again, I don't see the whole picture yet.

maks-rafalko:
Tests are green now.

> So, probably this interface (only abstraction) should be in infection/abstract-test-framework-adapter? 

@theofidry I can confirm it, this is exactly what needs to be done to decople Test Framework Adapters from `infection/infection`.

I started extracting Codeception Adapter out of Infection, and these are current dependencies of `CodeceptionAdapter` class:

* `use Infection\TestFramework\Coverage\CoverageLineData;`
* `use Infection\TestFramework\CommandLineBuilder;`
* `use Infection\TestFramework\Coverage\JUnitTestCaseSorter;`
* `use Infection\TestFramework\MemoryUsageAware;`
* `use Infection\TestFramework\TestFrameworkAdapter;`
* ...

^ all these classes must be extracted as interfaces to `infection/abstract-testframework-adapter`. This is basically what I'm going to do next.

maks-rafalko:
please review - quite an easy change, shouldn't take much time

theofidry:
As expressed previously: I'm 👍 for the change but not up for the timing. 

As it currently stands out, even if this is affecting only a few interfaces in this PR, it would effectively head for the polyrepo approach of #930 and hence require to edit not one repo but multiple ones. If we were to choose the mono-repo approach and this change incorporate or happen when we have the necessary tooling to handle those repos this would not be a problem.

However, just alone based on the fact that it's a part I definitely plan to rework quite a bit (it's at least on the same level as the `Configuration` - based on a gut feeling so I could be wrong) and based on [how this refactoring has been going so far]((https://github.com/infection/infection/pulls/theofidry), I am _really_ not keen to have to edit 2 repositories (even if it's just a couple of interfaces), let alone more.

But again, that's me and my opinion. The rest of the team is free to disagree

maks-rafalko:
Well, blocking the following development just because you assume you are going to update 2 interfaces is too much TBH.

Anyway, I believe even if there will be a change in the abstraction, this will be done 1 time during your refactoring (I don't see a reason why it should be updated more than once) - and I'm pretty sure you be able to do it ;)

I will then update Codeception & PHPSpec adapters myself, and PHPUnit stays inside the core for the time being.

theofidry:
Well it's going to be more than 2 interfaces and then there is the other framework adapters + I'm still unsure how to handle the multiple test-framework support to be honest. So while I see the benefit of your proposal and whole-heartedly agree with it, I just don't think it's worth _rushing_ for either: it can wait a few months it won't hurt us in any way.

So that's _my_ opinion. But as expressed before and it still stands, if the rest of the team feels differently I'll happily compromise. And since I think it's an important step with a well thought RFC, I feel it also deserves more than just my opinion to approve/disapprove of this as well.

maks-rafalko:
No longer want to spend my time rebasing and resolving conflicts. 

Since the only concern here is mutlirepo (which is not a concern for me at all) - please spent _your_ time investigating this topic and answer to all the questions from #930.

PR #934, opened 2019-12-26
Move coverage dir constants to Test Framework Adapters

maks-rafalko:
* Move coverage dir constants to Test Framework Adapters
* Move `COVERAGE_INDEX_FILE_NAME` const from concrete class (`XMLLineCodeCoverage`) to its interface (abstraction - `LineCodeCoverage`)

`XMLLineCodeCoverage` (and its interface) should not be aware of what and how many Test Framework Adapters we have. That's why each Test Framework should be responsible for own code coverage dir.

This is a preparation for extracting test framework adapters to separate packages - se we decouple Test Framework Adapter from another classes.

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -60,7 +60,7 @@ use Infection\Process\Runner\InitialTestsRunner;
 use Infection\Process\Runner\MutationTestingRunner;
 use Infection\Process\Runner\TestRunConstraintChecker;
 use Infection\TestFramework\Coverage\CoverageDoesNotExistException;
-use Infection\TestFramework\Coverage\XMLLineCodeCoverage;
+use Infection\TestFramework\Coverage\LineCodeCoverage;
 use Infection\TestFramework\HasExtraNodeVisitors;
 use Infection\TestFramework\TestFrameworkAdapter;
 use Infection\TestFramework\TestFrameworkTypes;
@@ -458,7 +458,7 @@ final class InfectionCommand extends BaseCommand
 
         $coverageDir = $config->getCoveragePath();
 
-        $coverageIndexFilePath = $coverageDir . '/' . XMLLineCodeCoverage::COVERAGE_INDEX_FILE_NAME;
+        $coverageIndexFilePath = $coverageDir . '/' . LineCodeCoverage::COVERAGE_INDEX_FILE_NAME;
 
         $processInfo = sprintf(
             '%sCommand line: %s%sProcess Output: %s',
--- a/src/Configuration/ConfigurationFactory.php
+++ b/src/Configuration/ConfigurationFactory.php
@@ -43,7 +43,6 @@ use Infection\FileSystem\SourceFileCollector;
 use Infection\FileSystem\TmpDirProvider;
 use Infection\Mutator\MutatorFactory;
 use Infection\Mutator\MutatorParser;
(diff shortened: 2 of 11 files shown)


src/Configuration/ConfigurationFactory.php
```diff
@@ -65,9 +64,9 @@ class ConfigurationFactory
     private const DEFAULT_TIMEOUT = 10;
 
     private const TEST_FRAMEWORK_COVERAGE_DIRECTORY = [
-        TestFrameworkTypes::PHPUNIT => XMLLineCodeCoverage::PHP_UNIT_COVERAGE_DIR,
-        TestFrameworkTypes::PHPSPEC => XMLLineCodeCoverage::PHP_SPEC_COVERAGE_DIR,
-        TestFrameworkTypes::CODECEPTION => XMLLineCodeCoverage::CODECEPTION_COVERAGE_DIR,
+        TestFrameworkTypes::PHPUNIT => 'coverage-xml',
```
maks-rafalko: We can't depend here on for example `PhpUnitAdapter::COVERAGE_DIR` because it will be in the separate package.

Later, we can refactor this code to return something like `$adapter::COVERAGE_DIR` - so return a dynamic variable value

src/Command/InfectionCommand.php
```diff
@@ -458,7 +458,7 @@ private function assertCodeCoverageExists(Process $initialTestsProcess, string $
 
         $coverageDir = $config->getCoveragePath();
 
-        $coverageIndexFilePath = $coverageDir . '/' . XMLLineCodeCoverage::COVERAGE_INDEX_FILE_NAME;
+        $coverageIndexFilePath = $coverageDir . '/' . LineCodeCoverage::COVERAGE_INDEX_FILE_NAME;
```
theofidry: Isn't this file PHPUnit specific?

theofidry:
Correct me if I'm wrong but I feel there is a coupling issue in the current coverage with the test framework adapters.

`LineCodeCoverage` should be generic,  but each test framework may have a different coverage tool or format right? So it would make sense that each test framework adapter has to provide its own coverage implementation too no?

src/Command/InfectionCommand.php
maks-rafalko: No, `index.xml` is for all adapters - this is basically index file of XML report. All 3 adapters generate XML coverage reports in their own folders, but all reports has `index.xml`

theofidry:
So whatever the test framework (so far) the generated XML and Junit report
are the same formats?

maks-rafalko:
> `LineCodeCoverage` should be generic, but each test framework may have a different coverage tool or format right? So it would make sense that each test framework adapter has to provide its own coverage implementation too no?

Fortunately, PHPUnit, PhpSpec, Codeception - all 3 adapters use the very same XML format. PhpSpec and Codeception under the hood use PHPUnit's code that generates XML files.

So we have 1 implementation of `LineCodeCoverage` interface - `XMLLineCodeCoverage` class. I really hope that other frameworks (e.g. Behat) will not invent the wheel and use the same PHPUnit's XML report generator.

maks-rafalko:
> So whatever the test framework (so far) the generated XML and Junit report are the same formats?

Exactly, just answered here https://github.com/infection/infection/pull/934#issuecomment-569445196

maks-rafalko:
Thank you @theofidry

theofidry:
@maks-rafalko thanks for the clarifications 👌

PR #935, opened 2019-12-29
Fix "Mutation score" badge

kubawerlos:
This PR:

- [x] Updates the URL of "Mutation score" badge as the old one stopped working - guys from Stryker did the same change: https://github.com/stryker-mutator/stryker/blob/master/README.md
- [x] Removes trailing whitespaces in README.md


theofidry:
Thank you @kubawerlos!

PR #936, opened 2019-12-29
Extracts the mutation generation from `MutationsCollectorVisitor`

theofidry:
This allows to simplify `MutationsCollectorVisitor` and let it focus only on collecting the mutations at each traverse.

This allowed to highlight and test the statefulness of `MutationsCollectorVisitor`. I took the opportunity to ensure its state is reset before each traverse otherwise the result of the previous traverse would pollute the second one (although it is unlikely to cause an issue in practice).

### DIFF
--- a/src/Mutator/NodeMutationGenerator.php
+++ b/src/Mutator/NodeMutationGenerator.php
@@ -1,0 +1,172 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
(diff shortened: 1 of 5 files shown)


src/Mutator/NodeMutationGenerator.php
```diff
@@ -0,0 +1,172 @@
...
+                return $this->generateForMutator($node, $mutator, $mutations);
+            },
+            []
+        );
+    }
+
+    /**
+     * @return Mutation[]
+     */
+    public function generateForMutator(Node $node, Mutator $mutator, array $mutations): array
```
maks-rafalko: this one can be private

tests/phpunit/Visitor/MutationsCollectorVisitorTest.php
```diff
@@ -0,0 +1,124 @@
...
+            [$visitor]
+        );
+
+        $this->assertSame([$mutation, $mutation], $visitor->getMutations());
+    }
+
+    public function test_it_resets_its_state_between_two_traverse(): void
+    {
+        /** @var ObjectProphecy|NodeMutationGenerator $nodeMutationGeneratorProphecy */
+        $nodeMutationGeneratorProphecy = $this->prophesize(NodeMutationGenerator::class);
```
maks-rafalko: as we are trying to not use Prophecy, is it possible to use `createMock` here?

theofidry: I couldn't manage :( That said I'm not that familiar with PHPUnit mocks so I might just not have found the way to do it.

There is another similar case where I used Prophecy as well, so if someone finds a solution it will be worth double checking the rest of the codebase (it should be simple cases nonetheless)

theofidry:
@maks-rafalko as far as I can tell, the MSI failures are on the `NodeMutationGenerator` which is untested because this code was untested before and I'm still not clear on how and what to test there

PR #937, opened 2019-12-30
Revisit the existing mutators

theofidry:
Left to do:

- [ ] Review all the mutators
- [ ] Ensure there is no hard BC break


src/Mutator/Arithmetic/EqualOrIdenticalAssignment.php
```diff
@@ -35,16 +35,36 @@
...
 use Infection\Mutator\Util\Mutator;
 use PhpParser\Node;
 
 /**
  * @internal
  */
-final class AssignmentEqual extends Mutator
+final class EqualOrIdenticalAssignment extends Mutator
 {
+    public static function getDefinition(): ?Definition
```
theofidry: @infection/core what do you think of this idea of "definition"?

The thought process here is that I would like to:

- review the mutators (see if they make sense, see what is missing, of check if there is any cosmetic change to do)
- make sure there is a meaningful and friendly explanation of what the mutation operator do
- classify the mutator
- provide a "remedy": a solution or guideline on how to kill the mutant

In order to do so I'm proposing to provide a way for the mutators to declare what they are about (via `Definition` and `Classification`) and have a quality check test (in the autoreview `MutatorTest`) to warn about the mutators which don't have a definition.

The classification chosen so far are the "classic" ones taken from other mutation testing tools, the terminology being mostly pushed by mbj from mutant (ruby mutation testing tool). It obviously can be expanded later on or we can add more classification means, but I think the above is a good first base.

Since this push for more self-documented mutators, this also opens the door to a way to generate the mutators doc from the source code rather than requiring us to manually keep things in sync.

theofidry: Oopsy looks like I completely messed up my merges so the above description is not on the right mutator. Still showcase well the example though.

If you agree on the idea I'll submit a dedicated PR with just the introduction of the definition thing and in another PR (this one for example) add it to the mutators

resources/schema.json
```diff
@@ -115,14 +115,14 @@
...
                 "Assignment": { "$ref": "#/definitions/default-mutator-config" },
-                "AssignmentEqual": { "$ref": "#/definitions/default-mutator-config" },
                 "BitwiseAnd": { "$ref": "#/definitions/default-mutator-config" },
                 "BitwiseNot": { "$ref": "#/definitions/default-mutator-config" },
                 "BitwiseOr": { "$ref": "#/definitions/default-mutator-config" },
                 "BitwiseXor": { "$ref": "#/definitions/default-mutator-config" },
                 "Decrement": { "$ref": "#/definitions/default-mutator-config" },
                 "DivEqual": { "$ref": "#/definitions/default-mutator-config" },
                 "Division": { "$ref": "#/definitions/default-mutator-config" },
+                "EqualOrIdenticalAssignment": { "$ref": "#/definitions/default-mutator-config" },
```
sanmai: TBH I'm not a big fan of this change because there is an usability issue: imagine you need to disable this mutation, or tell about it someone. `AssignmentEqual` is so much easier to type that you're probably just type it, but `EqualOrIdenticalAssignment` is more likely to get copy-n-pasted, or typed with a typo, or else.

theofidry: it's not just a renaming, the mutator itself changed here so `AssignmentEqual` would now be incorrect. Maybe it can be two distinct mutators, I'm not sure yet...

maks-rafalko:
If needed, this should be opened agains the curren master. Closing as stale

PR #938, opened 2019-12-30
Check for trailing whitespaces

kubawerlos:
This PR:

- [x] Adds script `devTools/check_trailing_whitespaces.sh`
- [x] Fixes the issues found by added script

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,10 @@ FLOCK=./devTools/flock
 compile:	 ## Bundles Infection into a PHAR
 compile: $(INFECTION)
 
+.PHONY: check_trailing_whitespaces
+check_trailing_whitespaces:
+	./devTools/check_trailing_whitespaces.sh
+
 .PHONY: cs
 cs:	  	 ## Runs PHP-CS-Fixer
 cs: $(PHP_CS_FIXER)
@@ -63,7 +67,7 @@ validate:
 
 .PHONY: autoreview
 autoreview: 	 ## Runs various checks (static analysis & AutoReview test suite)
-autoreview: phpstan validate test-autoreview
+autoreview: check_trailing_whitespaces phpstan validate test-autoreview
 
 .PHONY: test
 test:		 ## Runs all the tests
--- a/devTools/check_trailing_whitespaces.sh
+++ b/devTools/check_trailing_whitespaces.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -eu
+
+files_with_trailing_whitespaces=$(
+    find . \
+        -type f \
+        -not -path "./.git/*" \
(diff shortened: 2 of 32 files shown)


devTools/check_trailing_whitespaces.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -eu
```
theofidry: I would put `set -Eeuox pipefail` just for good measure

devTools/check_trailing_whitespaces.sh
```diff
@@ -0,0 +1,21 @@
...
+        -type f \
+        -not -path "./.git/*" \
+        -not -path "./vendor/*" \
+        -exec grep -EIHn "\\s$" {} \;
+)
+
+if [[ "$files_with_trailing_whitespaces" ]]
+then
+    printf '\033[97;41mTrailing whitespaces detected:\033[0m\n'
+    e=$(printf '\033')
```
theofidry: I'm not sure to understand why the `e` here: what does it represent or why is it necessary?

tests/e2e/Codeception_Basic/tests/_support/_generated/FunctionalTesterActions.php
```diff
@@ -12,5 +12,5 @@ trait FunctionalTesterActions
      */
     abstract protected function getScenario();
 
-    
+
```
theofidry: I would get rid of that line, there is two blank lines after `getScenario();` right now. I'm surprised this is not being caught by the CS fixer

Makefile
```diff
@@ -63,7 +67,7 @@ validate:
 
 .PHONY: autoreview
 autoreview: 	 ## Runs various checks (static analysis & AutoReview test suite)
-autoreview: phpstan validate test-autoreview
+autoreview: check_trailing_whitespaces phpstan validate test-autoreview
```
theofidry: since it does a CS change as opposed to a quality control per se, I think it is more appropriate to do it in `cs` than `autoreview`

tests/e2e/Codeception_Basic/tests/_support/_generated/FunctionalTesterActions.php
BackEndTea: Its part of the e2e fixtures, which are ignored by it ( i think)

BackEndTea:
Is this a copy paste from the php-cs-fixer repo? If so it may be appropriate to credit them.

devTools/check_trailing_whitespaces.sh
sanmai: Used in the next line it seems, as a shortcut.

theofidry:
Closed in favour of #949

Issue #929, opened 2019-12-25
Support Github Actions for Mutation Badge


drupol:
Hello,

The current [BadgeLogger.php](https://github.com/infection/infection/blob/master/src/Logger/BadgeLogger.php) only supports Travis CI for the moment.

How about supporting multiple CI system like:
* Github Actions
* CircleCI
* Drone

I think creating new loggers would be sufficient, what do you think ?

maks-rafalko:
> I think creating new loggers would be sufficient

I would not create additional `**Logger` files, but instead add an abstraction with the common interface here to reuse all existing logic in one `BadgeLogger.php` 

From what I see, for any of the CI platform, we need to know from it

* whether it is PR build or not (we don't want to update MSI on PRs)
* repository slug
* branch name

So we can create an interface and its 2 implementations:

```php
interface CiPlatform {
    public function isPullRequest(): bool;

    public function getRepositorySlug(): string;

    public function getBranchName(): string;
}

class TravisCI implements CiPlatform
{
    public function isPullRequest(): bool
    {
        return getenv('TRAVIS_PULL_REQUEST') === 'true';
    }

    public function getRepositorySlug(): string
    {
        return getenv('TRAVIS_REPO_SLUG');
    }

    public function getBranchName(): string
    {
        return getenv('TRAVIS_BRANCH');
    }
}

class GithubActions implement CiPlatform 
{
    // ...
}
```

Then, `BadheLogger` can be updated to use this interface instead of working with `getenv()`

This will allow adding other CI platform (e.g. CircleCI) by just adding new implementation of the interface (ideally - just one file).

drupol:
I pushed a first commit (WIP), now I need to know how you plan to get the CI autodetection working so I can work on it.

rask:
Status of this? Just set up Github Actions and was met with a blank badge. Did not know that Infection has hardcoded Travis' ENV vars.

maks-rafalko:
Not implemented yet. Subscribe to https://github.com/infection/infection/pull/939

Some base work is almost merged here https://github.com/infection/infection/pull/1047, but additional programming will be required.

rask:
I see, fill follow that. Thanks!

rask:
I see the CI detection feature was implemented, so is this resolved now or is there something hanging which needs to be done still?

localheinz:
@rask 

I guess we only need a new release now!

🤓

maks-rafalko:
[...]

PR #939, opened 2020-01-01
Issue #929: CI Platform abstraction.

drupol:
Preliminary work.

This PR:

- [ ] Enable Github Actions CI
- [ ] Covered by tests

Fixes https://github.com/infection/infection/issues/929


maks-rafalko:
> now I need to know how you plan to get the CI autodetection working so I can work on it.

Could you please elaborate on that? Not sure what do you mean. 

In order to understand what CI platform to use, we need some kind of factory that would return any of the `Platform` implementation. 

```php
# factory pseudocode

if (getenv('TRAVIS') === 'true') {
    return new TravisPlatform(...);
}

if (getenv('GITHUB_ACTION') === 'true') {
    return new GithubPlatform(...);
}

if (getenv('CIRCLECI') === 'true') {
    return new GithubPlatform(...);
}

return null; // we are on the unsupported platform
```

* Travis https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
* Github https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-environment-variables#default-environment-variables
* CircleCI https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables

BackEndTea:
What about a 'chain' implementation that takes an array of possible CI classes, and returns the first one that is valid.

src/Logger/CI/Github.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public function getRepositorySlug(): string
+    {
+        return getenv('GITHUB_REPOSITORY');
+    }
+
+    public function getBranchName(): string
+    {
+        return getenv('GITHUB_REF');
```
hpatoio: Quoting https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-environment-variables

> The branch or tag ref that triggered the workflow. For example, refs/heads/feature-branch-1. If neither a branch or tag is available for the event type, the variable will not exist.

So this should be something like

```suggestion
        return substr(getenv('GITHUB_REF'), 11, strlen(getenv('GITHUB_REF')));  
```

localheinz:
For reference, also see https://github.com/codeclimate/php-test-reporter/blob/master/src/TestReporter/Entity/CiInfo.php.

Lctrs:
There is also https://github.com/OndraM/ci-detector. It has support for branch name, but it will require a PR for detecting if we're in a PR and to get the repository slug.

Lctrs:
PR to get repository name is here : https://github.com/OndraM/ci-detector/pull/42

Lctrs:
Version 3.2.0 of OndraM/ci-detector is out with everything needed to implement this :)

localheinz:
Also see https://github.com/OndraM/ci-detector/pull/55.

maks-rafalko:
Closed in favor of #1135

Thank you!

PR #941, opened 2020-01-02
Exclude --configuration from PhpUnit options

duncan3dc:
This PR:

- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/153

If you run `infection --test-framework-options="--configuration=custom.xml"` then no mutations are detected as the generated configuration is overridden and regular autoloading is used.


duncan3dc:
That's correct, it does **not** solve that issue

maks-rafalko:
Thank you @duncan3dc

PR #943, opened 2020-01-02
Introduce Mutator definitions

theofidry:
This is a simple yet key PR in order to better assess the efficiency, documentation and quality of the mutators. This PR introduces:

- A mutator taxonomy via `OperationCategory` which provides the basic groups of mutators. More of them exists and may be introduced later, but I think this should be enough for now based on a quick assessment via #937.
- A better place to document a mutator by introducing the `Definition` object and `Mutator::getDefinition()`.
- Definitions are encouraged but *not* necessary

The goal of the `Definition` object is to:

- Provide a better mutator explanation. It allows for more readable content (IMO) than in the phpdoc and kinda forces more the hand to write a proper explanation (you either don't write a definition for the mutator or you do one with a decent description. You can technically still write BS or an empty string but it is kinda more highlighted so easier to spot in PRs)
- Classify the mutator ⚠️ IMO this is a _very_ important step. It allows to have a first pass of review of the mutators and spot some incoherences. For example a classification I did not put because should _not_ be done is `semantic addition`. An example is transforming `===` to `==`. Those mutators only lead to false-positives and redundancy.
- Provides guidelines on how to test it: I've not been very strict on that in #937 because most of them can be explained purely by their classification alone. It however still provides useful in more subtle, controversial or experimental mutators

Last but not least, this allows to:

- Encourage a first review & inspection of the current mutators
- Opens the door to potentially generate the mutator doc from the mutators directly

### DIFF
--- a/src/Mutator/Definition.php
+++ b/src/Mutator/Definition.php
@@ -37,6 +37,9 @@ namespace Infection\Mutator;
 
 use Webmozart\Assert\Assert;
 
+/**
+ * @internal
+ */
 final class Definition
 {
     private $description;
@@ -52,7 +55,7 @@ final class Definition
         string $category,
         ?string $remedies
     ) {
-        Assert::oneOf($category, OperationCategory::ALL);
+        Assert::oneOf($category, MutatorCategory::ALL);
 
         $this->description = $description;
         $this->category = $category;
--- a/src/Mutator/Definition.php
+++ b/src/Mutator/Definition.php
@@ -43,6 +43,10 @@ final class Definition
     private $category;
     private $remedies;
 
+    /**
+     * @param string      $description Explanation on what the mutator is about
+     * @param string|null $remedies Guidelines or recommendations on how to kill the generated mutations
+     */
     public function __construct(
         string $description,
         string $category,
--- a/src/Mutator/Definition.php
+++ b/src/Mutator/Definition.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Mutator;
+
+use Webmozart\Assert\Assert;
+
+final class Definition
+{
+    private $description;
+    private $category;
+    private $remedies;
+
+    public function __construct(
+        string $description,
+        string $category,
+        ?string $remedies
+    ) {
+        Assert::oneOf($category, OperationCategory::ALL);
+
+        $this->description = $description;
+        $this->category = $category;
+        $this->remedies = $remedies;
+    }
+
+    public function getDescription(): string
+    {
+        return $this->description;
+    }
+
+    public function getCategory(): string
+    {
+        return $this->category;
+    }
+
+    public function getRemedies(): ?string
+    {
+        return $this->remedies;
+    }
+}
--- a/src/Mutator/OperationCategory.php
+++ b/src/Mutator/OperationCategory.php
@@ -1,0 +1,99 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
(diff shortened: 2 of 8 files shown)


src/Mutator/OperationCategory.php
```diff
@@ -0,0 +1,99 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Mutator;
+
+final class OperationCategory
```
theofidry: More explanations regarding each category is available here

tests/phpunit/Mutator/OperationCategoryTest.php
```diff
@@ -0,0 +1,152 @@
...
+    {
+        $classReflection = new ReflectionClass(OperationCategory::class);
+
+        $this->assertFalse(
+            $classReflection->isInstantiable(),
+            sprintf('Did not expect "%s" to be instantiable', OperationCategory::class)
+        );
+    }
+
+    public function test_it_lists_all_its_exposed_constants(): void
```
theofidry: this test might appear as a bit overkill here regarding the assertions. I was a bit lazy and took some elements of a private codebase where we have a simple convention for "enums". Maybe this can be drastically simplified (although it is a bit overkill, it does have the merit of being more readable IMO) and could be re-used for other pseudo-enums we have in the codebase

src/Mutator/OperationCategory.php
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Neutral mutations aim at discovering integration errors. Example:
+     *
+     * ```
+     * $a = $a + $b;
+     *
+     * // to
+     *
+     * $a += $b;
```
maks-rafalko: isn't it an equivalent mutant? I don't remember we have such ones

src/Mutator/OperationCategory.php
maks-rafalko: Is it a replacement or an addition for `ProfileList`? What is the difference?

It's not immediately clear why do we have 2 categorization now.

src/Mutator/Util/Mutator.php
```diff
@@ -48,6 +49,11 @@ public function __construct(MutatorConfig $config)
         $this->config = $config;
     }
 
+    public static function getDefinition(): ?Definition
```
maks-rafalko: Note that this class (or extracted interface) will be an extension point. Thus, users will have to implement this method, which I don't really think they should.

Definitions (in any way) are definitely useful for Infection itself, but not for the custom mutators (Imagine you are creating a custom mutator Infection should use in your project).

Probably, this method should be on the higher level, in another base class that Infection core's mutator would extend.

I just see when the developers are not clear understand what is the `?Definition` is and will just return `null`. And even if they understand, there is no point to return any Definition from a custom mutator (from the first look).

src/Mutator/Definition.php
```diff
@@ -0,0 +1,72 @@
...
+final class Definition
+{
+    private $description;
+    private $category;
+    private $remedies;
+
+    public function __construct(
+        string $description,
+        string $category,
+        ?string $remedies
```
maks-rafalko: From the tests: `'This text is for providing guidelines on how to kill the mutant.'`

It would be more clear to either rename this parameter or add this comment to a property.

src/Mutator/OperationCategory.php
theofidry: It is. Hence if it fails we know there is an issue (e.g. the tests not passing initially or the JVM is corrupted). We don't have any use for it right now as the two main usages of such mutants are:

- making sure everything is in order, but it's more relevant when you have state kept between two mutations, e.g. if we were to use monkey patching to apply & revert a mutation or when you have long-living processes like in Python, Ruby or Java
- trying to assess if we can augment the information know, e.g. to deduce some types and be able to create a new class of mutator thanks to that

In other words: it's not relevant for us right now so can be removed

src/Mutator/OperationCategory.php
theofidry: Let's take the example of the two following mutators:

```
$a + $b -> $b - $a
$a - $b -> $a + $b 
```

You can say the following for both:

- they are an orthogonal replacement
- they are mutators operating on an arithmetic operation
- they are dual

`ProfileList` is a different categorization which is orthogonal to `OperationCategory`: it helps to indicate on what kind of elements the mutator operates on, but there is no overlap whatsoever between `OperationCategory` and `ProfileList`.

Another aspect is that `OperationCategory` is purely internal whereas `ProfileList` helps to expose profiles which can be consumed by the user.

In other words: it's a different kind of categorization which is internal and serves more for double checking the mutators (e.g. if you see a semantic addition you know it's wrong) and providing basic remedies (for most of them you can infer it from its type)

src/Mutator/Util/Mutator.php
theofidry: If we provide the base abstract class as an extension point, then the problem is taken care of. If it's an interface, we can not include this from the interface but include it in our mutators still.

So I don't think it will a problem, and since it's not entirely clear yet how we are opening up this bit I wouldn't worry too much about it.

src/Mutator/OperationCategory.php
maks-rafalko: now it's more clear, thanks

* Are these "OperationCategory" and "remedies" some kind of well known definitions? If so, probably we can leave a couple of links so anyone can dig into the terminology
* If they are our definitions - does it makes sense to add everything you described to our docs or ADR?

Why is it needed: because in a week I will struggle to identify what `OperationCategory::*` really means why is it different / orthogonal with the `ProfileList` ;) WDYT?

src/Mutator/Util/Mutator.php
maks-rafalko: This 100% will be a problem, but let's think about it later.

src/Mutator/OperationCategory.php
maks-rafalko: If it's not related to us - let's remove it. As it will definitely confuse people as it confused me during reviewing

src/Mutator/OperationCategory.php
theofidry: >Are these "OperationCategory" and "remedies" some kind of well known definitions?

Not at the moment. I'm keeping track of those terms on a separate file to lay down a nomenclature that I wish to submit on https://github.com/theofidry/awesome-mutation-testing (instead of like initially thought just infection) which would be peered review by the other mutation testing library authors.

The only well-agreed on terms here for now are those categories `semantic reduction`, `orthogonal replacement`, `neutral` and `semantic addition`. 

I'm not too satisfied with `OperationCategory`, could also be `MutatorCategory` instead... And `remedies` is one term that popped up as an idea of "fixing a mutant". So those two terms can be changed to something more fitting if you have any suggestion.

>Why is it needed: because in a week I (end even worse - someone who didn't see your explanation) will struggle to identify what OperationCategory::* really means why is it different / orthogonal with the ProfileList ;) WDYT?

I'm assuming you are talking of the category alone here as I hope the PR description highlights better the goal of `Definition`. IMO the category is important for internal quality and documentation in a number of ways:

- the category forces you, when you add or a review an undocumented mutator, to put some thoughts into what that mutator is doing. That process is necessary and make easier to spot smelly mutators (e.g. the ones which are a semantic addition or even the neutral ones which in our case should not be necessary yet)
- infer properties from the mutators for future improvements. 
  - For orthogonal replacement ones in peculiar where you can have the concept of `duality` and have auto-review checks to make sure when you declare a dual, that the other dual is declared
   - You can suggested "default" remedies based on the category, cf. the explanation in each category which I hope gives some hints on that

Last but not least, another goal is to improve the mutators taxonomy overall to make it as complete as possible in order to be able to better communicate and compare our mutators with other tools or review their efficiency.

src/Mutator/Definition.php
theofidry: Worth mentioning that there is little mutators where I started to document it. The reason being that for most of them, you can provide "default" guidelines based on the mutator category. For example:

```
Semantic reductions exposes unused semantics. If the semantics are unneeded, they should be removed. Otherwise they should be tested.
```

So I only documented it when I deemed worthy of it more as an internal documentation for now. It's definitely something that will need some adjustments (e.g. providing a default guideline instead of nothing) when we decide to expose it to the consumer via the doc for example. In other words: I laid down the foundation for it and attempt to better document the tricky cases I identified, but it will definitely require more work before exposing it to the users

sanmai:
> An example is transforming === to ==. Those mutators only lead to false-positives and redundancy.

This is not true for this example. When you have code that uses both `===` and `==` here and there due to backward-compatibility reasons or whatnot (say, your code is expected to work with strings-as-numbers), this and other mutator can easily tell you if your tests are doing good or not.

theofidry:
@sanmai I'm not convinced this would be a good thing. You are basically asking to require tests that can guarantee that "not more than specified" is implemented, which either doesn't make sense IMO. I would be interested in a counter-example, but AFAICS this would only lead to a tons of false-positives (which is one of the reason why it was disabled in the default right?)

src/Mutator/OperationCategory.php
theofidry: Renamed `OperationCategory` to `MutatorCategory` as it makes more sense for now I think. I hope we can improve the naming later on as we develop the taxonomy

sanmai:
> would only lead to a tons of false-positives (which is one of the reason why it was disabled in the default right?)

I recall it wasn't making false-positives, but rather mutations that nobody's willing to write tests around. Say, if you have a CS rule saying that `==` is not allowed, it does not make sense to write a test for a change from `===` to `==` because this change is never going to happen, and even if it does, there are strict types... So it's just an annoying mutation for most users not excluding myself. Yet annoying does not mean 100% useless.

theofidry:
I would honestly be interested in a non-false positive case out of curiosity.

But to not derail the conversation too much: they have at the very least been found to be horribly inefficient and given how expansive mutation testing is, cost-prohibitive. At the very least this is the reason why most mutation testing tools avoid such mutants

src/Mutator/OperationCategory.php
maks-rafalko: > I'm keeping track of those terms on a separate file to lay down a nomenclature that I wish to submit on https://github.com/theofidry/awesome-mutation-testing

IMO we need to refer to this file, otherwise nobody will understand what does it all mean

theofidry: I'll do so as soon as this file is public! I'm not publishing it yet because for now it's a bunch of notes and terms...

sanmai:
> I would honestly be interested in a non-false positive case out of curiosity.

They're truly hard to come by. If you have strong feelings against having a `===` to `==` mutator in a non-default profile, you may open a PR to remove it, and we'll have hopefully short and fruitful discussion about it.

PR #947, opened 2020-01-04
Introduce IgnoreMutator and a new Mutator interface

theofidry:
The goal here is to remove `Infection\Mutator\Util\Mutator`. The reason being are that I'm not fond of its API and I do not thing an abstract class is needed here and much less appropriate as an extension point if we ever expose it as such.

The proposal is:

- Move the `Mutator::shouldMutate()` stuff which is mostly about the ignore config in a dedicated class `IgnoreMutator`, relying on decoration instead of inheritance
- Introduce a simpler `Mutator` interface

Since I did not want to introduce too many changes at once, I did keep this class as a BC layer and just marked some elements as `@deprecated` instead.


theofidry:
Codecov you're drunk

src/Mutator/IgnoreMutator.php
```diff
@@ -0,0 +1,103 @@
...
+    private $config;
+    private $mutator;
+
+    public function __construct(MutatorConfig $config, Mutator $mutator)
+    {
+        $this->config = $config;
+        $this->mutator = $mutator;
+    }
+
+    public function shouldMutate(Node $node): bool
```
maks-rafalko: Now we have:

* `shouldMutate()`
* `canMutate()`
* `mutatesNode()`

Any reason why this method can't be named `canMutate`? Related to the previous comment that decorator should implemented same interface as decorated object

src/Mutator/IgnoreMutator.php
```diff
@@ -0,0 +1,103 @@
...
+    public function mutate(Node $node)
+    {
+        return $this->mutator->mutate($node);
+    }
+
+    // TODO: this function is necessary for now mostly because Mutations requires Mutator which is
+    //  expected to be the underlying mutator for now.
+    //  It might be possible to remove this getter if it turns out that Mutation may not require
+    //  the full Mutator object but just some elements of it.
+    public function getMutator(): Mutator
```
maks-rafalko: this method won't be needed when the `IgnoreMutator` implements `Mutator` interface

src/Mutator/MutatorFactory.php
```diff
@@ -166,16 +165,19 @@ private static function registerFromClass(
...
     {
         $mutators = [];
 
         foreach ($mutatorNames as $mutatorClass => $settings) {
+            $mutatorConfig = new MutatorConfig($settings);
+
             /** @var Mutator $mutator */
-            $mutator = new $mutatorClass(new MutatorConfig($settings));
-            $mutators[$mutator::getName()] = $mutator;
+            $mutator = new $mutatorClass($mutatorConfig);
```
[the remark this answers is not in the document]
theofidry: I admit it's still a point of confusion to me to be honest. To clear a few things up:

- Right now `schema.json` and `ProfileList` do not allow custom mutators
- The config API is a bit awkward: you are passing a raw object hence requiring to normalize & validate the config (although for now we managed to validate most of it if not all in `schema.json`) in the mutator which IMO is quite late in the process to find out there is a configuration error

Setting those potential custom mutator aside, I was thinking of being able to normalize this config upstream at the `ConfigurationFactory` level (not necessarily in that class but in that process, i.e. the `Configuration` object would have its mutators with a normalized config already). The few mutators (we have 2 or 3 of them) which require a special config to be passed in the constructor still can be handled in `MutatorFactory`.

Now what about custom mutator? I think this should be discussed more, but maybe an idea is to switch from a JSON config file to a PHP one? This way you could have something along the lines of:

```php
<?php

require __DIR__.'/path/to/custom-mutator/autoloader.php;

use Infection\Mutator\MutatorDefinition;
use Infection\Mutator\Mutator;
use Acme\MyMutatorConfig;
use Acme\MyMutator;

return [
     'mutators' => [
         new MutatorDefinition(
             'MyMutator', 
             MyMutator::class, 
             static function (array $settings): Mutator {
                 return new MyMutator(new MyMutatorConfig($settings));
             }
         )
     ],
];
```

We then have all the information needed and that factory provided by the user can be injected & used in the `MutatorFactory`.

It's just an idea, I think we need to think this through. But since this is involving a lot of non-trivial elements (how much of our API are we sharing? how? what about the autoloading? What about when we scope the PHAR? what about name conflicts? If we disable "no more elements" in the schema.json how to make sure we don't miss any new mutator we add?) I didn't want to think to much of it yet.

src/Mutator/IgnoreMutator.php
```diff
@@ -0,0 +1,103 @@
...
+ * of checking if he should mutate before attempting to mutate.
+ *
+ * A better alternative would be to allow to blindly mutate and do this "ignore + should mutate"
+ * check internally. We however do not do so because before actually mutating, there is a few
+ * expansive steps (e.g. retrieving the tests methods). Hence the currently chosen pattern allows
+ * better performance optimization in our case.
+ *
+ * @internal
+ */
+final class IgnoreMutator
```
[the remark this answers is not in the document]
theofidry: That was my initial thought to be honest but I reverted it because this was changing too much our API all over the place. For example:

- `$mutator::getName()`
- `get_class($mutator)`

So for now `IgnoreMutator` is an internal and closed _adapter_.

After changing it that way, I thought it was ok the way it is: it allows to have this tiny wrapper around your mutator which handles the annoying shouldMutator & ignore stuff and leave your mutator only worry about if they can mutate a specific node or not.

That being said, we can still look at a way to make `IgnoreMutator` a `Mutator`, but it would be wiser to do it in a separate PR IMO.

src/Mutator/IgnoreMutator.php
theofidry: Previously we had:

- `shouldMutate()`
- `mutatesNode()`

The new `canMutate()` is more meant to replace `mutatesNode()` than being an addition. I did not outright replace it and only deprecated it in `Util\Mutator` instead to preserve a certain BC layer to not have to change too much in one go

theofidry:
@maks-rafalko any doubt regarding those changes? If you wish I can do another PR based on this one where I can attempt to make `IgnoreMutator` a `Mutator`, i.e. a proper mutator decorator as opposed to an adapter

src/Mutator/IgnoreMutator.php
maks-rafalko: If we are going to remove deprecated code in the following PRs, I'm ok with that. So, basically, I would like to avoid 3 methods where only 1 is needed and seems like BC layer is needed only "to not have to change too much in one go"

maks-rafalko:
> I can do another PR based on this one where I can attempt to make `IgnoreMutator` a `Mutator`, i.e. a proper mutator decorator as opposed to an adapter

that would be great, thanks

PR #948, opened 2020-01-05
Rework AllMutatorTest

theofidry:
- Cosmetic changes regarding the `AllMutatorTest` class
- Rename `AllMutatorTest` into `MutatorRobustnessTest` to:
  - better convey what the test is about (testing the robustness of the mutators)
  - avoid potential confusion with `MutatorTest`
- Take advantage of the `NodeTraverserFactory` when we need a traverser with the "Infection native" visitors
- Make use of Infection "native" PHP parser in `AllMutatorTest`
- Add missing `@coversNothing` annotation to `AllMutatorTest`


tests/phpunit/Mutator/MutatorRobustnessTest.php
```diff
@@ -35,88 +35,120 @@
...
+                $fileInfo->getFilename(),
+                $fileInfo->getContents(),
+            ];
+        }
+
+        ksort($files, SORT_STRING);
+
+        self::$files = array_values($files);
+
+        yield from self::$files;
```
BackEndTea: Won't this now yield all files twice?

theofidry:
Hm I need to double check that, as it would apply elsewhere as well

PR #950, opened 2020-01-05
Tweak JUnitTestFileDataProvider

theofidry:
Cosmetic changes, not much to say about it


src/TestFramework/Coverage/JUnitTestFileDataProvider.php
```diff
@@ -36,38 +36,55 @@
...
 {
-    private $jUnitFilePath;
+    private $jUnitPath;
 
     /**
      * @var DOMXPath|null
      */
     private $xPath;
 
-    public function __construct(string $jUnitFilePath)
```
maks-rafalko: I would like to preserve this name as it immediately clear that the path is to file, not directory of files (as for XML coverage)

PR #951, opened 2020-01-05
Tweak CoverageXmlParser

theofidry:
Rename `CoverageXmlParser` into `IndexXmlCoverageParser` which I believe better reflect the intent. Also tweak the code for better safety and the tests for readability.


src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
```diff
@@ -0,0 +1,293 @@
...
+
+        $linesNode = self::safeQuery($xPath, '/phpunit/file/totals/lines')[0];
+
+        $percentage = $linesNode->getAttribute('percent');
+
+        if ($percentage === '') {
+            $data[$sourceFilePath] = new CoverageFileData();
+
+            return;
+        }
```
theofidry: we do not have any tests for this case. Do you remember a situation where this can happen @maks-rafalko?

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
```diff
@@ -0,0 +1,293 @@
...
+                '',
+                $relativeCoverageFilePath
+            );
+        }
+
+        $path = $projectSource . '/' . ltrim($relativeFilePath, '/') . '/' . $fileName;
+        $realPath = native_realpath($path);
+
+        if ($realPath === false) {
+            throw CoverageDoesNotExistException::forFileAtPath($fileName, $path);
```
theofidry: how can this scenario happen in practice? It's not covered by the tests either

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
```diff
@@ -0,0 +1,293 @@
...
+    {
+        // PHPUnit >= 6
+        $sourceNodes = self::safeQuery($xPath, '//project/@source');
+
+        if ($sourceNodes->length > 0) {
+            return $sourceNodes[0]->nodeValue;
+        }
+
+        // PHPUnit < 6
+        return self::safeQuery($xPath, '//project/@name')[0]->nodeValue;
```
theofidry: Can't see when this happens either even with PHPUnit 5... Could you point it out so we could cover it?

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
```diff
@@ -0,0 +1,306 @@
...
+            // The relative path is not present for old versions of PHPUnit. As a result we parse
+            // the relative path from the source file path and the XML coverage file
+            $relativeFilePath = str_replace(
+                sprintf('%s.xml', $fileName),
+                '',
+                $relativeCoverageFilePath
+            );
+        }
+
+        $path = implode(
```
theofidry: here `trim($relativeFilePath, '/')` could result in an empty string hence the path displayed by `CoverageDoesNotExistException::forFileAtPath()` could contain a`//`. It's not a problem for `realpath()` which handles it, but could be confusing for the exception message

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
maks-rafalko: See https://github.com/infection/infection/commit/1acf6866c4846cabea4fce91ee7165e1bf6022cb#diff-e64a6836ac47da4eaa945539e07f8ec7L78

Why is it being compared to empty string now?

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
maks-rafalko: I don't know, probably it's just for PHPStan?

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
maks-rafalko: You can reproduce it with this repo https://github.com/maks-rafalko/tactician-domain-events

Just checked it

```xml
<?xml version="1.0"?>
<phpunit xmlns="http://schema.phpunit.de/coverage/1.0">
  <project name="/Users/maksrafalko/sites/tactician-domain-events/src">
```

theofidry: 👍 Adding a test for it then

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
theofidry: Actually can happen if the coverage is out of sync with the source code

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
theofidry: Hm looks like a mistake, but looking at the coverage it appears that the result is the same as you would end up in `$coveredLineNodes->length === 0` and return there. So it effectively void another query. Is that the goal?

theofidry:
@maks-rafalko addressed a few of your comments as well as increased the overall coverage.

I'm not fully satisfied with this as I don't feel like the tests really provide the cover they should: we have what appears to be some half-made fixtures and it's hard to tell what elements are really considered there and which ones are ignored.

We also have a few peculiar cases: where there is no percentage, when it is null, when there is a method or a trait which I feel may seem quite artificial in the current test and leaves the doubt that it's really working on a real report.

Maybe a path of improvement would be to create a real micro project with real code and different scenarios of coverage, being able to dump the coverage with old & new PHPUnit versions (so that we can check BC when PHPUnit is bumped for example) and check the results.

If you fancy the idea, then maybe we can open an issue about it as a potential improvement

maks-rafalko:
> Maybe a path of improvement would be to create a real micro project with real code and different scenarios of coverage

Do you mean some kind of e2e tests for all those cases? Sounds good.

theofidry:
Should be fixed now! I opened https://github.com/infection/infection/issues/955 as well since I'm more convinced than ever we are missing a critical part in our testing. I'm now 100% sure how to achieve this, hence the issue opened to attempt to tackle it later

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
```diff
@@ -154,6 +155,13 @@ private function processCoverageFile(
 
         $percentage = $linesNode->getAttribute('percent');
 
+        if (substr($percentage, -1) === '%') {
+            // In PHPUnit <6 the percentage value would tak ethe form "0.00%" in _some_ cases.
```
maks-rafalko: `tak ethe` -> `take the`

PR #952, opened 2020-01-05
Tweak CachedTestFileDataProvider

theofidry:
- Rename `CachedTestFileDataProvider` to `MemoizedTestFileDataProvider` which better communicates on the cache strategy employed. Indeed`Cached` could imply using a cache store which usually can be busted or purged.
- Make `XMLLineCodeCoverageFactory` depend on the interface `TestFileDataProvider` instead of a specific implementation


maks-rafalko:
What about `Memoized***` instead of `Memoizer***` which is IMO better suites this case

theofidry:
👍

PR #958, opened 2020-01-08
Move PHPUnit coverage implementation in a dedicated directory

theofidry:
The goal is to move what's strictly related to PHPUnit XML coverage implementation in its own directory to make easier to distinguish what's belong to a specific implementation and what belongs to the core interface itself.

The goal behind it is to pave the way for #927 where we will have more than one implementation of `LineCodeCoverage`


src/Console/InfectionContainer.php
```diff
@@ -75,10 +75,10 @@
 use Infection\Process\Runner\TestRunConstraintChecker;
 use Infection\TestFramework\CommandLineBuilder;
 use Infection\TestFramework\Config\TestFrameworkConfigLocator;
-use Infection\TestFramework\Coverage\JUnitTestFileDataProvider;
-use Infection\TestFramework\Coverage\MemoizedTestFileDataProvider;
-use Infection\TestFramework\Coverage\TestFileDataProvider;
-use Infection\TestFramework\Coverage\XMLLineCodeCoverageFactory;
+use Infection\TestFramework\Coverage\PhpUnit\JUnitTestFileDataProvider;
```
maks-rafalko: It is not PHPUnit specific. Actually, any test framework that can generate JUnit report can use it. Now, in Infection, these are

* Codeception
* PHPUnit

Actually, all these 4 files are *not* PHPUnit specifc.

src/TestFramework/Coverage/PhpUnit/JUnitTestCaseSorter.php
```diff
@@ -33,10 +33,11 @@
 
 declare(strict_types=1);
 
-namespace Infection\TestFramework\Coverage;
+namespace Infection\TestFramework\Coverage\PhpUnit;
```
maks-rafalko: It's not PHPUnit specific. It's used for example by Codeception adapter

src/Console/InfectionContainer.php
theofidry: Didn't you say those frameworks were leveraging PHPUnit (I assumed some sub-libraries) to generate those coverage reports?

maks-rafalko: Exactly, then I didn't get the purpose of this PR tbh.

> to make easier to distinguish what's belong to a specific implementation and what belongs to the core interface itself

what is 'core interface itself'? Isn't it confusing that Codeception adapter will use some class from PHPunit namespace?

Let's take `JUnitTestCaseSorter` as an example. It's just sorter and has nothing to do with PHPUnit test framework.

theofidry: >Isn't it confusing that Codeception adapter will use some class from PHPunit namespace

Not any more confusing than Codeception is relying on PHPUnit (at least partially) to generate the coverage.

>Let's take JUnitTestCaseSorter as an example. It's just sorter and has nothing to do with PHPUnit test framework.

But it is used only for this PHPUnit-generated Junit - If we had other coverage to support that generated JUnit files as well without PHPUnit then it could be worth extracting it.

<hr/>

Right now we have:

```
TestFramework/
    Coverage/
        Coverage interfaces
        Coverage implementations
```

Since I plan to add more implementations of `LineCodeCoverage`, I would like to re-organise it to make it easier to distinguish the "core" (i.e. the interfaces) and the implementations. For example:

```
TestFramework/
    Coverage/
        Coverage interfaces
        PhpUnit/
        PhpSpec <- one based on naming convention enforced by the phpspec framework
```

theofidry: Btw this is also why I'm proposing `TestFramework\Coverage\PhpUnit` not `TestFramework\PhpUnit\Coverage`: I don't want it to be out if you split the test framework adapters in different packages, it should be accessible via the abstract

maks-rafalko: We are mixing up definitions.

Let's start from the scratch. We have 

* XML coverage report
* JUnit coverage report

these reports are generated by a `sebastianbergmann/php-code-coverage` package and we by mistake named it "PHPUnit's coverage report" - this is incorrect. We don't place our finders under `Symfony` namespace because they use Symfony's code, right?

These _Report Formats_ are (or should be) test framework agnostic.

In fact, XML report is generated by PHPSpec, Codeception & PHPUnit.

That's why I'm not so like the `PHPUnit` word in the namespace. Perhapse it should be just XML / JUnit.

What other `LineCodeCoverage` implementations are you going to add? When we will know what they are, probably it will be easier to understand the naming.

Btw, `XMLLineCodeCoverage` (implementation of `LineCodeCoverage` is named correctly (without PHPunit words).

> Btw this is also why I'm proposing `TestFramework\Coverage\PhpUnit` not `TestFramework\PhpUnit\Coverage`

That's fine, but can we replace `PHPUnit` with `Xml` here?

Btw, I've already extracted Codeception Test Framework, and here how the *already working* extracted package looks like:

https://github.com/maks-rafalko/infection-codeception-adapter/tree/master/src

Note that I had to extract [`Coverage/JUnitTestCaseSorter.php`](https://github.com/maks-rafalko/infection-codeception-adapter/blob/master/src/Coverage/JUnitTestCaseSorter.php), otherwise this class should be moved to a separate package. But this is ok to have a small duplication at this point.

theofidry:
If it’s the name that’s bothering you we can try to find a more suitable
name yes. My understanding though that both formats were defined by PHPUnit
(sub package or not) although I suspect the JUnit one should be framework
and language agnostic.

As examples of other coverage: PhpSpec based on a naming convention, a
generic naming convention one and one leveraging PHPUnit @covers
annotation. None of them are meant as a direct replacements, but more as
alternatives say for example to be able to run Infection without the
initial test run

maks-rafalko:
> If it’s the name that’s bothering

Exactly, because it's a bit confusing.

> My understanding though that both formats were defined by PHPUnit

PHPUnit here is just a vendor name in Packagist, but since this report (XML report) is used by other frameworks, I would really like to name it in a framework agnostic way, like XML.

PR #959, opened 2020-01-08
Fix trailing slashes checker to ignore cache files and logs

theofidry:


devTools/check_trailing_whitespaces.sh
```diff
@@ -26,6 +26,8 @@ set -Eeuo pipefail
 files_with_trailing_whitespaces=$(
     find . \
         -type f \
+        -not -path "./*.cache" \
+        -not -path "./*.log" \
```
sanmai: Could there be such files deeper in the hierarchy? If so, then this should work better:
```
-not -name \*.cache -not -name \*.log
```

PR #961, opened 2020-01-08
Deprecate mutators that are a semantic addition

theofidry:
The general consensus is that killing a mutant that is a semantic addition would require a test which tests against unspecified additions. Although it is not excluded to have some value in very specific cases, it is very hard to find reliable case where those mutants performs well and result in the vast majority in false positives.

I'll be more than happy to discuss on a case per case basis for those to prove otherwise, but based on the original promise, I think the stance should be their existence and efficiency needs to be demonstrated.

Also note that for now those mutators are simply marked as deprecated: concretely it does not change _anything_ yet. I would however work in the future on a system that would allow us to communicate to the users that a given mutator is deprecated and may be removed in future versions

### DIFF
--- a/src/Mutator/Boolean/IdenticalEqual.php
+++ b/src/Mutator/Boolean/IdenticalEqual.php
@@ -40,6 +40,7 @@ use PhpParser\Node;
 
 /**
  * @internal
+ *
  * @deprecated This mutator is a semantic addition
  */
 final class IdenticalEqual extends Mutator
--- a/src/Mutator/Boolean/IdenticalEqual.php
+++ b/src/Mutator/Boolean/IdenticalEqual.php
@@ -40,6 +40,7 @@ use PhpParser\Node;
 
 /**
  * @internal
+ * @deprecated This mutator is a semantic addition
  */
 final class IdenticalEqual extends Mutator
 {
--- a/src/Mutator/ConditionalBoundary/GreaterThan.php
+++ b/src/Mutator/ConditionalBoundary/GreaterThan.php
@@ -40,6 +40,7 @@ use PhpParser\Node;
 
 /**
  * @internal
+ *
  * @deprecated This mutator is a semantic addition
  */
 final class GreaterThan extends Mutator
--- a/src/Mutator/ConditionalBoundary/GreaterThan.php
+++ b/src/Mutator/ConditionalBoundary/GreaterThan.php
@@ -40,6 +40,7 @@ use PhpParser\Node;
 
 /**
  * @internal
+ * @deprecated This mutator is a semantic addition
  */
 final class GreaterThan extends Mutator
 {
--- a/src/Mutator/Boolean/NotIdenticalNotEqual.php
+++ b/src/Mutator/Boolean/NotIdenticalNotEqual.php
@@ -40,6 +40,7 @@ use PhpParser\Node;
 
 /**
  * @internal
+ *
  * @deprecated This mutator is a semantic addition
  */
 final class NotIdenticalNotEqual extends Mutator
--- a/src/Mutator/Boolean/NotIdenticalNotEqual.php
+++ b/src/Mutator/Boolean/NotIdenticalNotEqual.php
@@ -40,6 +40,7 @@ use PhpParser\Node;
 
 /**
  * @internal
+ * @deprecated This mutator is a semantic addition
  */
 final class NotIdenticalNotEqual extends Mutator
 {
--- a/src/Mutator/ConditionalBoundary/LessThan.php
+++ b/src/Mutator/ConditionalBoundary/LessThan.php
@@ -40,6 +40,7 @@ use PhpParser\Node;
 
 /**
  * @internal
+ *
  * @deprecated This mutator is a semantic addition
  */
 final class LessThan extends Mutator
--- a/src/Mutator/ConditionalBoundary/LessThan.php
+++ b/src/Mutator/ConditionalBoundary/LessThan.php
@@ -40,6 +40,7 @@ use PhpParser\Node;
 
 /**
  * @internal
+ * @deprecated This mutator is a semantic addition
  */
 final class LessThan extends Mutator
 {


src/Mutator/ConditionalBoundary/GreaterThan.php
```diff
@@ -40,6 +40,8 @@
 
 /**
  * @internal
+ *
+ * @deprecated This mutator is a semantic addition
  */
 final class GreaterThan extends Mutator
```
sanmai: Ow, these are very annoying to test against.

src/Mutator/Boolean/IdenticalEqual.php
```diff
@@ -40,6 +40,8 @@
 
 /**
  * @internal
+ *
+ * @deprecated This mutator is a semantic addition
  */
 final class IdenticalEqual extends Mutator
```
sanmai: No question about deprecating these two.

src/Mutator/ConditionalBoundary/GreaterThan.php
theofidry: `a >= b` is a shorthand of `a > b || a == b` hence makes sense to be mutated in `a > b`, `a == b`. But going from `a > b` to `a >= b` is "adding a column to the truth table", hence hard to test and not going to the right direction IMO.

Also note as it happened to me: it's not because a mutator helped once that it's a good one. It may have helped to report a case and you took a closer look and found a problem and addressed it. But you need to consider the two:

- was the "took a closer look and found a problem and addressed it" strictly about that mutation?
- would that problem been spotted by a different mutator?

But to re-iterate: I'm basing those deprecations on the basis they are semantic additions hence will lead to more false-positives than anything. I don't claim this is absolute, but I would prefer if we can find solid cases where they find anything to justify their existence if we want to keep them

maks-rafalko: > But going from `a > b` to `a >= b` is "adding a column to the truth table", hence hard to test and not going to the right direction IMO.

I didn't get your explanation. Any real example where this mutator leads to "more false-positives"? Why is it "hard to test"?

theofidry: Actually I have to take that back: I don't think it's harder to test against here, but it creates redundant mutants.

So trying to formalize it a bit, the problems reported by semantic additions are:

- Expecting to have tests against unspecified semantics which is hard to do
- Create redundant mutants

So as you requested, here's an example for both cases.

`=== -> ==`. This is a semantic addition that will be hard and in some cases just impossible to tests. For example:

```php
function foo(string $a, string $b): bool {
  return $a === $b; // mutating to `==` cannot be tested here
}
```

Another example:

```php
/**
 * @parm string[] $strings
 */
function foo(array $strings): bool {
  return in_array('bar', $strings, true);
  // mutating `true` to `false` here makes the operation looser. And unless you are "violating"
  // the interface to pass non-string values (e.g. stringeable objects), you cannot test this
  // change either
}
```

Those two are examples of false positives just straight up impossible to solve.

Now back to our specific case of `> -> >=`. This one alone is ok. Because although it's a semantic addition, you are not expanding types but simplify shifting the value. For example if you had `$a > 10` your truth table was `[11;+∞[` and once mutated into `$a >= 10` it's now `[10;+∞[`. So in practice this one is not harder to test: you just need a test against that magic limit `10`. **However** it will result in a redundant mutation with `literal -> literal - 1` which would transform `$a > 10` into `$a > 9`.

Most research papers focus on finding and removing redundant mutants, but from what I saw from mbj/mutant and talking with the maintainer, the most efficient method there seem to be to avoid semantic additions in the first place.

Note that if you go with that logic, you also remove a great deal of possible mutants. If the original postulate is that a mutation testing just introduce a random change in the program, then you have virtually an infinite amount of possible mutations. If you go with the logic of the mutation testing tool should find semantically dead compositions (i.e. code not run or tested), hence _remove_ complexity (instead of adding), then you a finite number of them and they are easier to resonate about.

maks-rafalko: Thanks for the explanation. First to examples (`===` and `in_array`) are ok. But I'm still not convinced about deprecating `> -> >=`.

> **However** it will result in a redundant mutation with `literal -> literal - 1` which would transform `$a > 10` into `$a > 9`.

You are almost right. But our `DecrementInteger` is quite smart - it does not decrement integers in such comparison expressions:

https://github.com/infection/infection/blob/6d9bf6b81144c2cf50fc1d341c02ab9287b15f38/src/Mutator/Number/DecrementInteger.php#L123-L133

We don't have redundant mutations here. So, `> -> >=` is a very useful mutator IMO, and I would say this deprecation should be reverted.

theofidry:
Won’t it still create a redundant mutation with > -> <=?

maks-rafalko:
> Won’t it still create a redundant mutation with > -> <=?

Didn't get it. `GreaterThan` mutator only mutates `>` to `>=`

theofidry:
Yes but we have another mutator changing > to its counterpart <=. In this
case combined with > to >=, you have redundant mutations

maks-rafalko:
Imagine you have the following array and function

```php
$ints = [1, 2, -1, -2]

function filterSqrMoreThanOne(array $ints): array {
    return array_filter($ints, fn($x) => pow($x, 2) > 1);
}

filterSqrMoreThanOne($ints); // [2, -2]
```

And then you write quite a stupid tests (that we would like to improve applying Mutation Testing):

```php
$ints = [1, 2, -1, -2]
$filtered = filterSqrMoreThanOne($ints); // [2, -2]

assertTrue(count($filtered) > 0); // pass
```

First mutation (`GreaterThan`):

```diff
function filterSqrMoreThanOne(array $ints): array {
-    return array_filter($ints, fn($x) => pow($x, 2) > 1);
+    return array_filter($ints, fn($x) => pow($x, 2) >= 1);
}
```

Tests still pass (Mutant is escaped), as this function returns `[1, 2, -1, -2]` now. We are improving the tests by changing assertion:

```diff
- assertTrue(count($filtered) > 0);
+ assertTrue(count($filtered) === 2);
```

This one kills this Mutant.

Second mutation (`GreaterThanNegotiation`):

```diff
function filterSqrMoreThanOne(array $ints): array {
-    return array_filter($ints, fn($x) => pow($x, 2) > 1);
+    return array_filter($ints, fn($x) => pow($x, 2) <= 1);
}
```

**Tests still pass (!!!)**. Why? Because now function returns `[1, -1]` and this is exactly 2 elements.

So, while our tests kills `GreaterThan` it does not kill `GreaterThanNegotiation`.

Then we write more clever tests:

```diff
- assertTrue(count($filtered) === 2);
+ assertTrue($filtered === [2, -2]);
```

which kills both mutants (generated by both mutators). So, these 2 mutators are not redundant as it seems from the first look.

theofidry:
Hm indeed, without the literal change (here `a > 1 -> a > 0; a > 2`) it does make sense as otherwise we're missing the 1 value shifting change

PR #964, opened 2020-01-08
Move creation of Test Framework Adapters to their own factory classes

maks-rafalko:
### Why?

Each Test Framework will be extracted to its own package. And to be able to create all of them in a similar manner (even without any knowledge what the adapter is being created), we need some common Factory interface.

Here it is:

```php
interface TestFrameworkAdapterFactory
{
        string $testFrameworkExecutable,
        string $tmpDir,
        string $testFrameworkConfigPath,
        string $testFrameworkConfigDir,
        string $jUnitFilePath,
        string $projectDir,
        array $sourceDirectories,
        bool $skipCoverage
    ): TestFrameworkAdapter;
}
```

So, instead of 1 class to be responsible for creating all possible adapters, now each adapter's factory knows how to create particular adapter.

### Future goal

Then, when adapters are extracted, and we will implement autodiscovery of adapters, this code will look like

```php
# pseudocode

foreach($autodiscoveredAdapterFactories as $adapterFactory) {
    if ($adapterName === $adapterFactory::getAdapterName()) {
        return $adapterFactory::create(
            (new TestFrameworkFinder($adapterFactory::getAdapterName()))->find(),
            $this-tmpDir,
            ...
        );
    }
}
```

Thus, we can register any adapter from any package, and this package will be responsible for instantiating an adapter object.

### DIFF
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -443,7 +443,8 @@
             "description": "Sets the framework to use for testing. Defaults to phpunit.",
             "anyOf": [
                 "phpunit",
-                "phpspec"
+                "phpspec",
+                "codeception"
             ]
         },
         "bootstrap": {
--- a/devTools/phpstan-src.neon
+++ b/devTools/phpstan-src.neon
@@ -11,5 +11,5 @@ parameters:
         - '#Method Infection\Configuration\Schema\SchemaConfigurationFile::getDecodedContents() should return stdClass but returns stdClass|null.#'
 
         -
-            path: '%currentWorkingDirectory%/src/TestFramework/Factory.php'
+            path: '%currentWorkingDirectory%/src/TestFramework/Codeception/Adapter/CodeceptionAdapterFactory.php'
             message: '#Access to constant VERSION on an unknown class Codeception\\Codecept#'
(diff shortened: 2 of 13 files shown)


theofidry:
Looking at it, I'm 👍 for having a factory by adapter, but I like less the common interface: it implies each adapter has the same capabilities which, although so far is true, is unlikely to remain true (e.g. if we support a generic test framework which might not have any coverage support).

Do you have any idea how you want to auto-discover those factories?

maks-rafalko:
> Do you have any idea how you want to auto-discover those factories?

Yes, you can read about my initial thoufht in "Autodiscovering of Packages" in https://github.com/infection/infection/issues/922

> it implies each adapter has the same capabilities which, although so far is true, is unlikely to remain true (e.g. if we support a generic test framework which might not have any coverage support)

Well, when we introduce a generic test framework, we will revise this logic. For now, we have 3 adapters and all of them more or less works in the same way.

When I will dig to autodiscovering, I will better see what we can do with it. Now it's only assumptions and ideas.

resources/schema.json
```diff
@@ -443,7 +443,8 @@
             "description": "Sets the framework to use for testing. Defaults to phpunit.",
             "anyOf": [
                 "phpunit",
-                "phpspec"
+                "phpspec",
+                "codeception"
```
theofidry: IMO this change warrants an update in the `SchemaConfigurationFactoryTest`

maks-rafalko:
Thank you @theofidry

PR #965, opened 2020-01-08
Tweak the Mutation class

theofidry:
This PR:

- Exposes only the `Mutator` name instead of the full instance in order to reduce the (mutable) state exposed (besides not being used for anything else that getting the name)
- Slightly optimize the hash creation
- Validate the mutator name
- Narrow `$mutatedNode` to `Node` instead of `Node|Node[]` (no other implications in practice)
- Tweak the Lexer configured in the container to make sure it includes the attributes used by `Mutation`
- Only select the necessary attributes and ensure they all exists

Note that the hash calculation is slightly changed since we pick the mutator name instead of class but it work as well.

### DIFF
--- a/src/Logger/PerMutatorLogger.php
+++ b/src/Logger/PerMutatorLogger.php
@@ -79,7 +79,7 @@ final class PerMutatorLogger extends FileLogger
         $processPerMutator = [];
 
         foreach ($processes as $process) {
-            $mutatorName = $process->getMutatorClass()::getName();
+            $mutatorName = $process->getMutatorName();
             $processPerMutator[$mutatorName][] = $process;
         }
 
--- a/src/Logger/PerMutatorLogger.php
+++ b/src/Logger/PerMutatorLogger.php
@@ -79,7 +79,7 @@ final class PerMutatorLogger extends FileLogger
         $processPerMutator = [];
 
         foreach ($processes as $process) {
-            $mutatorName = $process->getMutator()::getName();
+            $mutatorName = $process->getMutatorClass()::getName();
             $processPerMutator[$mutatorName][] = $process;
         }
 
--- a/src/Logger/TextFileLogger.php
+++ b/src/Logger/TextFileLogger.php
@@ -104,7 +104,7 @@ final class TextFileLogger extends FileLogger
             $index + 1,
             $mutantProcess->getOriginalFilePath(),
             $mutantProcess->getOriginalStartingLine(),
-            $mutantProcess->getMutatorClass()::getName()
+            $mutantProcess->getMutatorName()
         );
     }
 }
--- a/src/Logger/TextFileLogger.php
+++ b/src/Logger/TextFileLogger.php
@@ -104,7 +104,7 @@ final class TextFileLogger extends FileLogger
             $index + 1,
             $mutantProcess->getOriginalFilePath(),
             $mutantProcess->getOriginalStartingLine(),
-            $mutantProcess->getMutator()::getName()
+            $mutantProcess->getMutatorClass()::getName()
         );
     }
 }
--- a/src/Mutation.php
+++ b/src/Mutation.php
@@ -36,8 +36,6 @@ declare(strict_types=1);
 namespace Infection;
 
 use function count;
-use function get_class;
-use Infection\Mutator\Mutator;
 use Infection\TestFramework\Coverage\CoverageLineData;
 use PhpParser\Node;
 
@@ -48,7 +46,7 @@ use PhpParser\Node;
 class Mutation
 {
     private $originalFilePath;
-    private $mutator;
+    private $mutatorClass;
     private $mutatedNodeClass;
     private $mutatedNode;
     private $mutationByMutatorIndex;
@@ -81,7 +79,7 @@ class Mutation
     public function __construct(
         string $originalFilePath,
         array $originalFileAst,
-        Mutator $mutator,
+        string $mutatorClass,
         array $attributes,
         string $mutatedNodeClass,
         $mutatedNode,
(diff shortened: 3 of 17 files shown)


src/Logger/PerMutatorLogger.php
```diff
@@ -79,7 +79,7 @@ private function setUpPerCalculatorMutator(): void
         $processPerMutator = [];
 
         foreach ($processes as $process) {
-            $mutatorName = $process->getMutator()::getName();
+            $mutatorName = $process->getMutatorClass()::getName();
```
maks-rafalko: Does it work since mutator class is a string?

src/Logger/TextFileLogger.php
```diff
@@ -104,7 +104,7 @@ private function getMutatorFirstLine(int $index, MutantProcessInterface $mutantP
             $index + 1,
             $mutantProcess->getOriginalFilePath(),
             $mutantProcess->getOriginalStartingLine(),
-            $mutantProcess->getMutator()::getName()
+            $mutantProcess->getMutatorClass()::getName()
```
sanmai:
```suggestion
            $mutantProcess->getMutatorClass()
```

src/Logger/PerMutatorLogger.php
theofidry: Yep

src/Logger/TextFileLogger.php
theofidry: Hm it's not the same thing: `::getName()` will return `Plus` whereas `getMutatorClass()` will be the FQCN

src/Mutation.php
```diff
@@ -90,17 +88,17 @@ public function __construct(
...
         $this->mutatedNode = $mutatedNode;
         $this->mutationByMutatorIndex = $mutationByMutatorIndex;
         $this->tests = $tests;
     }
 
-    public function getMutator(): Mutator
+    public function getMutatorClass(): string
     {
-        return $this->mutator;
+        return $this->mutatorClass;
```
sanmai:
```suggestion
        return $this->mutator::getName();
```

src/Logger/TextFileLogger.php
sanmai: Yes, getMutatorClass needs to do that call apprently. And should it be renamed to getMutatorName?

src/Process/Listener/MutationTestingConsoleLoggerSubscriber.php
```diff
@@ -142,7 +142,7 @@ private function showMutations(array $processes, string $headlinePrefix): void
                     $index + 1,
                     $mutation->getOriginalFilePath(),
                     (int) $mutation->getAttributes()['startLine'],
-                    $mutation->getMutator()::getName()
+                    $mutation->getMutatorClass()::getName()
```
sanmai:
```suggestion
                    $mutation->getMutatorClass()
```

theofidry:
Added a few more tweaks, updated the PR description

maks-rafalko:
master now fails with the following error:

```
PHP Fatal error:  Uncaught Error: Class 'MethodCallRemoval' not found in /Users/maksrafalko/sites/infection/src/Process/Listener/MutationTestingConsoleLoggerSubscriber.php:145
Stack trace:
#0 /Users/maksrafalko/sites/infection/src/Process/Listener/MutationTestingConsoleLoggerSubscriber.php(111): Infection\Process\Listener\MutationTestingConsoleLoggerSubscriber->showMutations(Array, 'Escaped')
#1 /Users/maksrafalko/sites/infection/src/EventDispatcher/EventDispatcher.php(55): Infection\Process\Listener\MutationTestingConsoleLoggerSubscriber->onMutationTestingFinished(Object(Infection\Events\MutationTestingFinished))
#2 /Users/maksrafalko/sites/infection/src/Process/Runner/MutationTestingRunner.php(101): Infection\EventDispatcher\EventDispatcher->dispatch(Object(Infection\Events\MutationTestingFinished))
#3 /Users/maksrafalko/sites/infection/src/Engine.php(149): Infection\Process\Runner\MutationTestingRunner->run(Array, 1, '')
#4 /Users/maksrafalko/sites/infection/src/Engine.php(115): Infection\Engine->runMutationAnalysis(1)
#5 / in /Users/maksrafalko/sites/infection/src/Process/Listener/MutationTestingConsoleLoggerSubscriber.php on line 145
```

when `--show-mutations` is used and there is escaped mutants

theofidry:
@maks-rafalko hasn't it been fixed with https://github.com/infection/infection/pull/971?

maks-rafalko:
just checked with latest master

theofidry:
Ok will look, that must be something else not covered :/

theofidry:
I couldn't reproduce locally, do you have a specific config or running something else than `bin/infection`?

maks-rafalko:
As simple as that:

```
bin/infection --threads=$(nproc) --filter=src/TestFramework/Codeception/Adapter/CodeceptionAdapter.php --only-covered --show-mutations
```

But I recommend to test with a smaller project since initial tests run for Infection takes ~7min on my VM.

Basically, it can ve reproduced with any project with `--show-mutations` and escaped mutants

theofidry:
@maks-rafalko didn't work... what the heck

theofidry:
I also find it surprising because `--show-mutations` is used in the CI right?

maks-rafalko:
> @maks-rafalko didn't work... what the heck

do you mean you can't reproduce it?

> I also find it surprising because `--show-mutations` is used in the CI right?

It does not work on Travis as well:

https://travis-ci.org/infection/infection/jobs/634925288#L773

PR #967, opened 2020-01-09
Tweak the differs

theofidry:
- Replaces the differ service IDs by their FQCN
- Mark the differs as `@final`
- Micro-optimize the code by importing the used symbols
- Add more tests for the `Differ` class
- Remove no longer necessary code


theofidry:
Hm looks like I made AppVeyor fail with another PR, will have a look at it ASAP

PR #968, opened 2020-01-09
Remove the MutantInterface

theofidry:
The `MutantInterface` has only one implementation and since the mutant is an object created internally by infection, it is unlikely to be a part opened up. And even if we do open up that part, then it is likely going to be more appropriate to design the desired interface from the need at that point rather than the existing one.


src/Mutant/Mutant.php
```diff
@@ -54,7 +55,7 @@ public function __construct(string $mutatedFilePath, Mutation $mutation, string
         $this->diff = $diff;
     }
 
-    public function getMutatedFilePath(): string
+    public function getMutantFilePath(): string
     {
         return $this->mutatedFilePath;
```
maks-rafalko: let's rename a property as well to make it consistent

theofidry: 👍

PR #969, opened 2020-01-09
Mark test cases with `@group integration` when I/O operations are found

theofidry:
Some test cases executes (or are very likely) I/O operations (read and/or write). Those test cases should be marked as `integration` in order to distinguish them from "pure" test cases as they are more likely to be:

- slower
- flaky if ever run in parallel or sub-process

This does not change much overall, just at _what_ time those tests are executed.

Since it can be quite hard/tedious to detect such cases manually, especially when retro-inspecting the code, I added a test case - which I don't claim to be full-proof, which allows to detect such test cases. And as the consequences of a mistake are minimal, it is a not a big deal if we happen to have a few false-positives.

### DIFF
--- a/tests/phpunit/AutoReview/IntegrationGroup/IntegrationGroupProvider.php
+++ b/tests/phpunit/AutoReview/IntegrationGroup/IntegrationGroupProvider.php
@@ -1,0 +1,191 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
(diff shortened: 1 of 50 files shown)


tests/phpunit/AutoReview/IntegrationGroup/IntegrationGroupProvider.php
```diff
@@ -0,0 +1,191 @@
...
+            Assert::isInstanceOf($parentTestCaseClassReflection, ReflectionClass::class);
+        }
+
+        //
+        // Check the source class code
+        //
+        $classFileName = $classReflection->getFileName();
+        $classCode = file_get_contents($classFileName);
+
+        // Case where the test case itself use I/O functions
```
maks-rafalko: seems like a copy-pasted wrong comment, as here we are checking source class' code, not the code of the test

tests/phpunit/AutoReview/IntegrationGroup/IntegrationGroupProvider.php
```diff
@@ -0,0 +1,191 @@
...
+                return [$testCaseClass, $parentClassFileName];
+            }
+
+            $parentClassReflection = $parentClassReflection->getParentClass();
+        }
+
+        return null;
+    }
+
+    private static function codeContainsIoFunctions(string $code): bool
```
maks-rafalko: what about `Filesystem::**` checks? If that's hard to implement, let's ignore that for now ;)

theofidry:
Arg, went to fast for the merge, I'm a bit too trigger merge happy today

theofidry:
>Could you please give an idea how are you using @group integration right now, when we don't have parallel tests?

For now, at least by default in the `Makefile`, they are used to distinguish the `unit` tests (PHPUnit group `default`) and `e2e` (which is no longer really e2e...) (PHPUnit group `integration` and `e2e`).

So in practice for now it just means they will be run _after_

PR #971, opened 2020-01-09
Revert change regarding Mutation#mutatedNode

theofidry:
`Mutation#mutatedNode` is indeed a `Node|Node[]`. I was confused by the original where it was and was not captured by the tests. So this PR reverts this change and ensure this is better tested

### DIFF
--- a/devTools/check_trailing_whitespaces.sh
+++ b/devTools/check_trailing_whitespaces.sh
@@ -23,6 +23,7 @@
 
 set -Eeuo pipefail
 
+# TODO: some tests could leverage the new StringNormalizerTest to avoid having to be ignored here
 files_with_trailing_whitespaces=$(
     find . \
         -type f \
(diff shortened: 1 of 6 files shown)


devTools/check_trailing_whitespaces.sh
```diff
@@ -23,6 +23,7 @@
 
 set -Eeuo pipefail
 
+# TODO: some tests could leverage the new StringNormalizerTest to avoid having to be ignored here
```
theofidry: Will check that next

src/Mutation.php
```diff
@@ -109,6 +110,19 @@ public function __construct(
         $this->coveredByTests = count($tests) > 0;
     }
 
+    public function getOriginalFilePath(): string
```
theofidry: minor re-ordering to match the constructor parameters order

PR #975, opened 2020-01-09
Move `Mutation` under the `Infection\Mutation` namespace

theofidry:


src/Mutator/NodeMutationGenerator.php
```diff
@@ -82,7 +82,7 @@ public function __construct(
     }
 
     /**
-     * @return Mutation[]
+     * @return \Infection\Mutation\Mutation[]
```
maks-rafalko: why is it needed? I think should be reverted

src/Visitor/MutationsCollectorVisitor.php
```diff
@@ -46,7 +45,7 @@
 final class MutationsCollectorVisitor extends NodeVisitorAbstract
 {
     /**
-     * @var Mutation[]
+     * @var \Infection\Mutation\Mutation[]
```
maks-rafalko: same

src/Visitor/MutationsCollectorVisitor.php
```diff
@@ -72,7 +71,7 @@ public function leaveNode(Node $node): ?Node
     }
 
     /**
-     * @return Mutation[]
+     * @return \Infection\Mutation\Mutation[]
```
maks-rafalko: same

src/Mutator/NodeMutationGenerator.php
theofidry: Happens occasionally when doing a refactor to change the class namespace with phpStorm

PR #976, opened 2020-01-10
Introduce MutatedNode

theofidry:
This PR introduces a tiny wrapper for `Node|Node[]` representing a mutated node. I do not believe this wrapper is that useful in terms of typings, but it has the merit of having a self-explanatory interface which might help with the understanding of the code. Indeed you now have:

```
Mutation::getMutatedNode(): MutatedNode()
```

instead of passing around `Node`.

### DIFF
--- a/src/EvilNode.php
+++ b/src/EvilNode.php
@@ -33,47 +33,43 @@
 
 declare(strict_types=1);
 
-namespace Infection\Visitor;
+namespace Infection;
 
-use function array_key_exists;
-use function get_class;
-use Infection\Mutation;
+use function is_array;
 use PhpParser\Node;
-use PhpParser\NodeVisitorAbstract;
+use Webmozart\Assert\Assert;
 
-/**
- * @internal
- */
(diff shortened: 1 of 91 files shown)


sanmai:
The aspect of returning a generator was discussed in great detail in #450, and I recall the consensus was, for some reason, not to return a generator. I might even been a proponent of returning a generator, but the consensus was as such.

theofidry:
@sanmai I think you should redirect your comment in the other PR: I removed the mutator change from this one

sanmai:
Sure, haven't seen since starting to writing that comment

maks-rafalko:
Build is failed https://travis-ci.org/infection/infection/jobs/635164978#L513

maks-rafalko:
Theo, issue with Fatal Error and `--show-mutations` can be reproduced with this build as well https://travis-ci.org/infection/infection/jobs/635199304#L1027

theofidry:
Just saw it: checking it

PR #977, opened 2020-01-10
Simplify the Mutator::mutate() contract by forcing to return a Generator

theofidry:
The return type of `Mutator::mutate()` is `Node|Node[]|Generator|array` which is kinda annoying to work with and a bit confusing as well.

This PR proposes to force `Generator<Node|Node[]>` instead.


sanmai:
The aspect of returning a generator was discussed in great detail in #450, and I recall the consensus was, for some reason, not to return a generator.

theofidry:
As mentioned by @sanmai this has already been discussed in #450 among other changes.

I would like however to re-visit the idea now as a dedicated PR, properly highlighting the changes strictly related to that choice so that we are better able to assess if it's worth it or not.

IMO this makes it a lot simpler. However I cannot help to note [@sanmai 's comment](https://github.com/infection/infection/pull/450#discussion_r216214625). Note that for now there is no failure hinting that this is a problem for us right now, but I would like to pay a closer look at it in order to see if it really doesn't apply to use yet or if we are missing a test somewhere

theofidry:
@sanmai regarding your example, it is a bit confusing but I think the really confusion here is not from the generator itself. Looking more closely at the code it actually make total sense, what is a lot more confusing is `iterator_to_array()` which you might forget by default retains the _keys_. Not retaining them lead to a less confusing scenario in your example: https://3v4l.org/6B4Nl

src/Mutator/NodeMutationGenerator.php
```diff
@@ -130,11 +129,7 @@ private function generateForMutator(Node $node, IgnoreMutator $mutator, array $m
             return $mutations;
         }
 
-        $mutatedResult = $mutator->mutate($node);
-
-        $mutatedNodes = $mutatedResult instanceof Generator ? $mutatedResult : [$mutatedResult];
-
-        foreach ($mutatedNodes as $mutationByMutatorIndex => $mutatedNode) {
+        foreach ($mutator->mutate($node) as $mutationByMutatorIndex => $mutatedNode) {
```
sanmai: This is certainly simpler.

Here is another small catch, which is absent with arrays: generators can have multiple values for the same key.

sanmai:
> what is a lot more confusing is `iterator_to_array()` which you might forget by default retains the _keys_. 

Just like the `foreach` we're using. And we're passing the keys, which we can't reason about, down the line.

src/Mutator/NodeMutationGenerator.php
theofidry: Caught that after reviewing a bit more in details your comment. I added an `iterator_to_array($x, false)` with a comment about it. I would like to cover it in a test but unfortunately `NodeMutationGenerator` is completely untested at the moment

src/Mutator/NodeMutationGenerator.php
```diff
@@ -129,7 +130,11 @@ private function generateForMutator(Node $node, IgnoreMutator $mutator, array $m
             return $mutations;
         }
 
-        foreach ($mutator->mutate($node) as $mutationByMutatorIndex => $mutatedNode) {
+        // It is important to not rely on the keys here. It might otherwise result in some elements
+        // being overridden, see https://3v4l.org/JLN73
+        $mutatedNodes = iterator_to_array($mutator->mutate($node), false);
+
+        foreach ($mutatedNodes as $mutationByMutatorIndex => $mutatedNode) {
```
sanmai: So it is correct that you propose we basically ignore all keys from mutators? I don't think this is a major problem, but it can be non-obvious.

sanmai: Alternatively we can increment `$mutationByMutatorIndex` as we go.

theofidry:
@sanmai correct, corrected it.

Now that this issue is sorted, I think this PR is clearer. The issue described is solved although untested it has at least a note and matters in only two places. I actually don't like that it's more than one but it's what it is...

Also note that we can decide to go the other way as well and force to return an array instead. But I think that one is a bit more tricky. 

In any case, my wish would go towards simplifying that interface: being array or generator, one but not both.

src/Mutator/NodeMutationGenerator.php
theofidry: Yes, I find it a bit tricky in this case as it may really be non-obvious, but truth be told it's something you always need to be wary of with _any_ generator...

tests/phpunit/Fixtures/SimpleMutationsCollectorVisitor.php
```diff
@@ -41,9 +42,9 @@ public function leaveNode(Node $node)
             return;
         }
 
-        $mutatedResult = $this->mutator->mutate($node);
-
-        $mutatedNodes = $mutatedResult instanceof Generator ? $mutatedResult : [$mutatedResult];
+        // It is important to not rely on the keys here. It might otherwise result in some elements
+        // being overridden, see https://3v4l.org/JLN73
+        $mutatedNodes = iterator_to_array($this->mutator->mutate($node), false);
```
sanmai: In this case `foreach ($generator as $val)` will keep all values regardless of keys, so you don't have to convert it to an array.

sanmai:
> Also note that we can decide to go the other way as well and force to return an array instead.

No, generators are certainly better.

tests/phpunit/Fixtures/SimpleMutationsCollectorVisitor.php
theofidry: Yes but we use the index: so we either do that or have a `$i = 0` that we increment. In both scenarios I think it's non-obvious at first glance and warrants a comment. I would love a test as well, but that's out of scope for now

sanmai: We don't here.

sanmai: Something like below should work just fine:
```
 foreach ($this->mutator->mutate($node) as $mutatedNode) {
```

src/Mutator/Arithmetic/Assignment.php
```diff
@@ -59,13 +60,11 @@ public static function getDefinition(): ?Definition
     }
 
     /**
-     * @param Node&Node\Expr\AssignOp $node
-     *
-     * @return Node\Expr\Assign
```
maks-rafalko: Why not to preserve all return type hints wrapped with `Generator`? This will help PHPStan to control more mistakes: 

https://phpstan.org/r/531725c8-1222-4e75-ac55-964614565919

theofidry: I was under the impression that we don't really care about being more specific than `Generator<Node|Node[]>` here

maks-rafalko: Why?

While we now use `$mutator->mutate($node);` only in one place (`NodeMutationGenerator`), it does not mean specific types will be useless in the future, especially when we extract mutators to a separate package.

In fact, they are useful even now:

```php
/**
 * @param Node $node
 * @return Generator<Node\Scalar\DNumber>
 */
public function mutate(Node $node): Generator
{
    {
        if ($node->value === 0.0) {
            if ($node->value === 0.0) {
                yield new Node\Scalar\DNumber(1.0);

                return;
            }
        }

        yield new Node\Scalar\DNumber(0.0);
    }
}
```

so, if in this mutator someone make a mistake, returning `LNumber` instead of `DNumber`, that would be reported by analyzer.

Anyway, replacing `Generator<type>` with `Generator` is like a decreasing type coverage, similar to replacing `int[]` with `array`.

theofidry: Fixed... that was tedious

PR #978, opened 2020-01-10
Fix CI

theofidry:


theofidry:
It requires to add a test in `MutationTestingConsoleLoggerSubscriberTest` since this one is not properly covered. I'll try to have a look later at this class

PR #981, opened 2020-01-10
Simplify `IgnoreMutator`

theofidry:
- Rename `IgnoreMutator` `shouldMutate()` to `canMutate()` to be more in line with the `Mutator` interface
- Only expose the decorated mutator name instead of the instance itself

### DIFF
--- a/src/Mutator/IgnoreMutator.php
+++ b/src/Mutator/IgnoreMutator.php
@@ -92,12 +92,8 @@ final class IgnoreMutator
         return $this->mutator->mutate($node);
     }
 
-    // TODO: this function is necessary for now mostly because Mutations requires Mutator which is
-    //  expected to be the underlying mutator for now.
-    //  It might be possible to remove this getter if it turns out that Mutation may not require
-    //  the full Mutator object but just some elements of it.
-    public function getMutator(): Mutator
+    public function getMutatorName(): string
     {
-        return $this->mutator;
+        return $this->mutator::getName();
     }
 }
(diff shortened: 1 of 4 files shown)


maks-rafalko:
From slack:

> regarding the IgnoreMutator : I cannot find a way to make it implement Mutator
the blocker is Mutator::getName()  which is static: I cannot find a way to return the decorated mutator name in any way.

So am I right that we can't have a decorator when the decorated class has static public methods? Isn't it a marker that something is wrong with our hierarchy or method types?

I also can't find a way on how to decorate static method of decorated class, but I have a gut feeling that we are doing something wrong here ;)

theofidry:
You are correct, at leat with the current design. Maybe another way to do
it is to not make getName static, and if still necessary we can use
__callstatic for it: it’s also a PHP limitation not necessarily a design
issue

maks-rafalko:
> Maybe another way to do it is to not make getName static

I thought so, but it used in a static context in many places and it looks like in a convenient way

> it’s also a PHP limitation not necessarily a design issue

Out of curiosity, do you think in other languages it's possbile? (to decorate a class with static public methods)

theofidry:
>Out of curiosity, do you think in other languages it's possbile? (to decorate a class with static public methods)

Note sure, it's been a while since I've did some Java and pretty sure you cannot declare static methods in TypeScript either...

>I thought so, but it used in a static context in many places and it looks like in a convenient way

I'll do another attempt in another PR

PR #982, opened 2020-01-11
Remove the deprecated `Infection\Mutator\Util\Mutator` class

theofidry:
>what is the final goal of removing abstract class and moving its methods to trait?

Now this is more about what's left about the current `Util\Mutator`. We got the following choice:

1. Have `Util\Mutator` and rename it to maybe `BaseMutator` instead where it just contains `getName()`
2. Have a trait `GetMutatorName`


src/Mutator/DefaultMutatorSettings.php
```diff
@@ -33,25 +33,18 @@
...
+use function end;
+use function explode;
 
 /**
- * @deprecated Should not be needed anymore. More work needs to be done but ultimately this class
- *             should go away.
+ * @internal
  */
-abstract class Mutator implements MutatorInterface
+trait DefaultMutatorSettings
```
maks-rafalko: 1. Why was it deprecated in the first place?
2. How is the trait better that the abstract class?
3. what is the final goal of removing abstract class and moving its methods to trait?
4. `// TODO: only pass the mutator config if necessary` - how would you do that? by checking mutators?

theofidry: >Why was it deprecated in the first place?

- It was the extension point for `Mutator`, but I'm not fan of having an abstract class as the base and instead the `Mutator` interface is more fitting
- I really want to get rid of `Mutator\Util` where a random bunch of things was added because nothing better could be found at that time
- It requires `__construct(MutatorConfig $config)` which is no longer required for every mutator but only a handful of them (5 in total IIRC) and the goal is to have better config objects which avoid a lot of validation handling in the mutator (to move it more upstream)

>what is the final goal of removing abstract class and moving its methods to trait?

Now this is more about what's left about the current `Util\Mutator`. We got the following choice:

1. Have `Util\Mutator` and rename it to maybe `BaseMutator` instead where it just contains `getName()`
2. Have a trait `GetMutatorName` or something like that (which is what we can rename `DefaultMutatorSettings` into once I remove the `getDefinition()` from there)

I tend to prefer 2., but it's purely a question of taste in this scenario.

>// TODO: only pass the mutator config if necessary - how would you do that? by checking mutators?

Handle it in `MutatorConfig` indeed. I'm still not super satisfied with it, but I think it's ok based on what we have right now.

src/Mutator/MutatorFactory.php
```diff
@@ -174,6 +174,7 @@ private static function createFromNames(array $mutatorNames): array
         foreach ($mutatorNames as $mutatorClass => $settings) {
             $mutatorConfig = new MutatorConfig($settings);
 
+            // TODO: only pass the mutator config if necessary
```
BackEndTea: Prehaps we should have an interface for configurable mutators (`ConfigurableMutator`)? And do some kind of side injection.

```php

interface ConfigurableMutator
{
    public function configure(MutatorConfig $config): void;
}

finalclass TrueValue implements ConfigurableMutator, Mutator
{
    private $settings = [];
    
    public function configure(MutatorConfig $config): void
    {
        $this->settings = $config->getMutatorSettings()
    }
}
```

theofidry: Could do: I'm still hesitating on how to do so because I would like more specific VOs than `MutatorConfig`, to remove the settings parsing (and validation) from the mutators more upstream

theofidry:
Ping @infection/core

PR #985, opened 2020-01-12
Consolidate ProfileList tests

theofidry:


tests/phpunit/Mutator/ProfileListProvider.php
```diff
@@ -70,8 +70,8 @@ private function __construct()
 
     public static function mutatorNameAndClassProvider(): Generator
     {
-        foreach (ProfileList::ALL_MUTATORS as $name => $class) {
-            yield [$name, $class];
+        foreach (self::implementedMutatorProvider() as [$filePath, $className, $shortClassName]) {
```
theofidry: This is more reliable as the source of truth for the list of mutators are the ones _found_ not the ones registered in the `ProfileList` which we are testing...

tests/phpunit/Mutator/ProfileListProvider.php
```diff
@@ -103,7 +103,7 @@ public static function implementedMutatorProvider(): Generator
                 continue;
             }
 
-            if (!self::extendsMutator($mutatorReflection)) {
+            if (!$mutatorReflection->implementsInterface(Mutator::class)) {
```
theofidry: although this does not change anything in practice for now, this is still more correct as if a new mutator does not extend the util one, it would not be collected

Issue #956, opened 2020-01-07
Create a PHAR build for composer


1ed:
Hello!

As the preferred installation is the phar version maybe a composer installable phar build would be beneficial. Other projects like phpstan https://github.com/phpstan/phpstan and psalm https://github.com/psalm/phar do that as well nowadays.

maks-rafalko:
From https://github.com/psalm/phar

> This allows you to install Psalm without worrying about composer conflicts.

Can someone explain me *why* using a PHAR through composer is better than just using prefixed standalone PHAR or using [Phive](https://phar.io/).

After reading [the RFC of PHPStan](https://github.com/phpstan/phpstan/issues/2074) I'm still miss the point.

Let's accumulate the pros and cons of this feature request.

@1ed in your particular case, what are the advantages of composer based prefixed PHAR? Can't you just use the existing PHAR distribution? Phive? Any inconveniences?

theofidry:
>Can someone explain me why using a PHAR through composer is better than just using prefixed standalone PHAR or using Phive.

Personally: I find annoying to have to use yet another tool to manage my PHARs. I also don't remember if you can save PHARs on a project basis either

sanmai:
Pros:
- Uses composer to install and manage a standard .phar without a need to install it independently, or use anything else to manage it. (PHIVE what?... Don't even need to learn it!)
- No need to manage and install any of the Infection's dependencies. Phar it is. No conflicts, no problems.
- Flawless and easy upgrades. Just run the usual `composer update` command, and there you have a new version.

Cons:
- Need to update another package with every release of the main package. Can be automated with a daily cron job, but still - another package.

1ed:
As others said before, for me the main benefit is convince too. No need to install and maintain and learn just another tool and configuration file just to manage dependencies, which composer already does.

The con would be to more work on the maintainer side (but it can be automatic). I can understand that you do not want to support just another release medium. Another con that phars are not first class citizens in composer so no support on that side, but maybe it can change over time. As Ocramius said in the phpstan RFC fully scoped phars should work well. Infection is fully scoped, right?

maks-rafalko:
> Infection is fully scoped, right?

Right

martinssipenko:
Maybe we can revive the conversation about this topic? @maks-rafalko what's your take on this?

PR #986, opened 2020-01-12
Fix usage of SymfonyProcess 5.x

theofidry:
Since the SymfonyProcess component has no dependencies and requires PHP 7.2 which we do as well, we can bump that dependency and get rid of some unnecessary code

### DIFF
--- a/src/Process/Builder/MutantProcessBuilder.php
+++ b/src/Process/Builder/MutantProcessBuilder.php
@@ -70,7 +70,7 @@ final class MutantProcessBuilder
             )
         );
 
-        $process->setTimeout($this->timeout);
+        $process->setTimeout((float) $this->timeout);
 
         $symfonyProcessVersion = $this->versionParser->parse(Versions::getVersion('symfony/process'));
 


BackEndTea:
👎 Unless we implement something like #956, and don't have to worry about dependency conflicts.

theofidry:
? I was just saying this was a safe bump: it won’t  impact any of our
dependencies whatsoever since its only requirement is PHP ^7.2

BackEndTea:
This forces any users who want to install infection through composer to use version 5.0 of this dependency. I'd rather have our users to be able to stay on a version 4.* if they want to do iso.

maks-rafalko:
Agree with @BackEndTea, it's not too hard to maintain removed code and I also see no reason of removing possibility of using Infection for majority of the users (who are on <5.0)

theofidry:
I'd prefer that we would be more aggressive with our dependencies since we have the scoped PHAR, #956 being only a nice to have improvement over it...

Anyway let's see about that another time: I did it because we are checking the version at _every_ call, e.g. for _each mutation_ when creating a mutant. Will change that

theofidry:
Turns out we are making it more complicated than necessary. So changed this PR:

- `setTimeout()` now forces `float|null` (in 5.x) hence we have to cast our values: done
- `inheritEnvironmentVariables()` is unnecessary, the method is deprecated in 4.4 but it's necessary to be called to _disallow_ it, i.e. do `inheritEnvironmentVariables(false)`. So we don't need to check the versions to know if we can call this method, we don't need to call it in the first place

sanmai:
This is a welcome change, but I see from composer.json that we're letting people use Infection with Symfony 3.x - 4.x. Can we tell that these versions won't need this workaround?

sanmai:
I'd limit us to using only the current and/or supported branches of Symfony, and TBH I don't see a huge problem here because there are older versions of Infection (take last released for example), and they work just as fine. So it is a matter of reaching consensus.

sanmai:
Well, that's a bummer. There's nothing else done here, so this could be closed just as well.

theofidry:
>Well, that's a bummer. There's nothing else done here, so this could be closed just as well.

There is :) I didn't do this PR for the sake of purity but for 2 issues:

- Our processes are not consistent from one part of the code to another and neither from a Symfony version to another. For example in 3.4 one will inherit environment variables and another will not (although they both will in 4.x and 5.x)
- In some places where `inheritEnvironmentVariables()` is necessary, we parse the process version _each time_. I don't think it's a super big overhead, but it should not be necessary at all

Plus this PR highlights a critical lack of testing regarding our Symfony versions: it's cool to want to maintain 3 different major Symfony versions, but we need to test this (and that shows the cost of supporting that many!).

Proof: our current code just breaks with SymfonyProcess 5.x (this PR fix it)

theofidry:
>Inheriting env variable by default was added in 4.0.0 version only, and we still need to call this method for 3.4.* versions.

I totally missed that thanks, I tried another approach to fix it

sanmai:
Do we really have to check for versions? Can we just call `inheritEnvironmentVariables()` if it exists, and call it a day?

theofidry:
>Do we really have to check for versions? Can we just call inheritEnvironmentVariables() if it exists, and call it a day?

I was more referring to the fact that our code completely breaks with Symfony 5.0 despite allowing it in `composer.json` (timeout was passed as `int` despite being typehinted as `?float`)

sanmai:
> our code completely breaks with Symfony 5.0 ... int passed despite being typehinted as ?float

No question about that.

sanmai:
Do you mean that they changed it between versions? Is this a thing that also needs us to act differently for each version?

theofidry:
3.3.0: deprecated not inheriting environment variables
4.0.0: environment variables will always be inherited
4.4.0: deprecated Process::inheritEnvironmentVariables(): env variables are always inherited.
5.0.0: removed Process::inheritEnvironmentVariables()

It's not a big deal per se, just a bit annoying when you support 3 major versions like us :)

theofidry:
CI is failing due to low MSI: that's ok it's not a piece of code we want to keep forever and battle-test

theofidry:
ping @infection/core

src/Process/ProcessFactory.php
```diff
@@ -0,0 +1,120 @@
...
+use Infection\Console\Util\PhpProcess;
+use Infection\Utils\VersionParser;
+use PackageVersions\Versions;
+use Symfony\Component\Process\Process;
+use function version_compare;
+
+/**
+ * @internal
+ *
+ * This factory is only used for dealing with the SymfonyProcess 4.4 deprecation. It can be safely
```
maks-rafalko: With current implementation, after we upgrade to `^4.0 || ^5.0` dependencies as we 

1. will have to literally do a revert of this commit 
2. and remove version comparison

On the other side, leaving the code as is (in `master`), we will only need to 

1. remove version comparison

So isn't it a complication?

theofidry: It is a complication for a performance improvement. But the question is whether we plan to bump those dependencies soon or not

sanmai: I'd rather see us upgrading to `^4.0 || ^5.0` tbh

maks-rafalko: > It is a complication for a performance improvement

Here is what we have running that code 1 time instead of N (number of Mutants)

```php
$start = microtime(true);

for ($i = 0; $i < 1; $i++) {
    $symfonyProcessVersion = $versionParser->parse(Versions::getVersion('symfony/process'));

    version_compare($symfonyProcessVersion, '4.4.0', '<');
}

var_dump(microtime(true) - $start);
die;
```

Result: `0.0033740997314453s` for 2000 calls

1 call is `8.9883804321289E-5s`.

We can neglect these multiple calls to version parsers&comparers and not introduce this complexity.

> I'd rather see us upgrading to ^4.0 || ^5.0 tbh

Well, I'm not strongly against, but is this version comparison a really reason to bump dependencies? IMO doesn't worth it.

PR #988, opened 2020-01-12
Tweak VersionParser

theofidry:
- Make `VersionParser` final: the few places where it was mocked it was an unspecified stub hence safe to replace by a plain instance
- Change `[0-9]` for `\d` in the regex when possible (purely aesthetic change unless the regex engine has an optimization of some kinds that I'm not aware of)
- Change the error message to make it more friendly
- Add a few more tests


theofidry:
ping @infection/core

PR #989, opened 2020-01-13
Make `Mutator::getName()` non-static

theofidry:
When attempting to make `IgnoreMutator` a `Mutator` instead of an adapter, it turned out to not be possible due to `getName()` which is static, hence `IgnoreMutator` cannot provide the decorated instance name.

There is really no (easy at least) way to get around the problem. So as a last ditch attempt, out of curiosity, I tried to make `getName()` non-static. Turns out we are only relying on the static access _in the tests_. As a result, I think a utility `MutatorName` introduced here is good enough for keeping it simple in our tests whilst making the design simpler.

### DIFF
--- a/src/Mutator/IgnoreMutator.php
+++ b/src/Mutator/IgnoreMutator.php
@@ -92,8 +92,8 @@ final class IgnoreMutator
         return $this->mutator->mutate($node);
     }
 
-    public function getMutatorName(): string
+    public function getName(): string
     {
-        return $this->mutator::getName();
+        return $this->mutator->getName();
     }
 }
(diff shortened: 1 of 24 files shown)


theofidry:
>Is it the only change needed to be able to make IgnoreMutator a Mutator?

Yep, I have the draft PR already for that change so can guarantee it

>What about getDefifnition()?

Since it's not actually used yet I did not worry about it. Plus unlike `getName()` it is not used at all in "prod" (in the source code), only for test/documentation purposes. If necessary though, I'm sure we can do a similar change

PR #990, opened 2020-01-13
Make `IgnoreMutator` a `Mutator`

theofidry:


src/Mutator/IgnoreMutator.php
```diff
@@ -64,6 +66,17 @@ public function __construct(MutatorConfig $config, Mutator $mutator)
         $this->mutator = $mutator;
     }
 
+    public static function getDefinition(): ?Definition
```
maks-rafalko: I'm definitely don't like this as it's a violation of LSP. If its' a temp solution - we can leave with it, but this should be fixed eventually as this looks very strange IMO

theofidry: I'm leaving it that way because the usage it's not clear. If we really need it, we can always make it non-static or even just return `null` here. I added that guard just to make sure we don't accidentally call it meanwhile

PR #991, opened 2020-01-14
Extract the mutant process creation out of the runner in its own factory

theofidry:
I'm not 100% this split is a good idea tbh, since we now can't easily test/guarantee the order of the events... But maybe that's something that should be tested at a higher level anyway.

In any case, the split makes it a bit easier to test so the runner is now tested.

This PR also introduces back the `EventDispatcher` interface since we are implementing another one here. The previous implementation has been renamed `SyncEventDispatcher`: not sure it's the best name so I'm open to propositions


theofidry:
ping @infection/core

src/Process/Runner/MutantProcessFactory.php
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * @param Mutation[] $mutations
+     *
+     * @return MutantProcess[]
+     */
+    public function create(array $mutations, string $testFrameworkExtraOptions): array
+    {
+        $this->eventDispatcher->dispatch(new MutantsCreatingStarted(count($mutations)));
```
BackEndTea: I wouldn't expect a factory to dispatch events about it creating its contents.
So i'm not 100% sure if this requires a factory, or if it should be named otherwise

theofidry: It would be weird to have `MutantCreated` dispatched here and `MutantsCreating{Started,Finished}` elsewhere...

I'm open to suggestions for a more appropriate name... I picked `Factory` because at the end of the day it is a factory, I don't think it's weird that a factory dispatches an event per se.

src/Config/ValueProvider/SourceDirsProvider.php
```diff
@@ -36,6 +36,7 @@
 namespace Infection\Config\ValueProvider;
 
 use function count;
+use function implode;
```
BackEndTea: Does the cs fixer not pick these up?

theofidry: nope :/

src/Event/EventDispatcher/EventDispatcher.php
```diff
@@ -0,0 +1,48 @@
...
+declare(strict_types=1);
+
+namespace Infection\Event\EventDispatcher;
+
+use Infection\Event\Subscriber\EventSubscriber;
+
+/**
+ * @internal
+ */
+interface EventDispatcher
```
maks-rafalko: Why is it needed again? We removed it in one of the PRs :)

theofidry: because we introduce another implementation in this one

PR #992, opened 2020-01-14
Move to in-house container

theofidry:
I tried different ways but in the end, I think just using our own `Container` class with type-hinted getters is the simplest way. Judge by yourselves with the code diff


theofidry:
Failure is related to lower MSI (we don't test much of the container so that's understandable...)

sidz:
remember that around 1.5year ago I suggest to do this :)

anyway I'm 👍 with removing pimple out of our composer.json

maks-rafalko:
> Judge by yourselves with the code diff

as I understand, we could add all those getters for `InfectionContainer` as well. Out of curiosity, what's wrong with Pimple container?

theofidry:
>as I understand, we could add all those getters to InfectionContainer as well. Out of curiosity, what's wrong with Pimple container?

It turned out that with all the getters it really felt like we were more accommodating to be able to use pimple than using pimple because we needed it. You can see the code itself to fully replace pimple is barely 10 LoC.

Also another advantage is that you can no longer register a service at runtime as easily: so you are kind of forced to really use the container public API to instantiate and configure it. Not a necessity by any means, but another bonus

maks-rafalko:
> You can see the code itself to fully replace pimple is barely 10 LoC.

Let's be honest, those 10 LoC is not a replacement of Pimple :)

IMO, we should also add to our implementation "frozen" services feature.

https://github.com/silexphp/Pimple/blob/21e3e0dc89183c96fad7f12e09e1d345558e9a56/src/Pimple/Container.php#L81-L83

So once the service is used, there should be no way of overriding it (hence "frozen").

theofidry:
>Let's be honest, those 10 LoC is not a replacement of Pimple :)

Replacement of what we are using of `Pimple`*

>So once the service is used, there should be no way of overriding it (hence "frozen").

I don't think there is a need for it since there is only 2 places you can register services:

- `Container::create()` (where you don't have any control it's the default services being registered)
- `Container::withDynamicDependencies()` (same, you pass some parameters but you don't have control over the registered services)

The only way you could do it it calling `__construct` but... I don't think we should care about it tbh

PR #993, opened 2020-01-15
Wrap reflection with our own reflection class

BackEndTea:
This PR:

- [x] Wraps reflection with our own class
- [x] Covered by tests

This will make some things a bit easier: It allows us to make a switch to better reflection, removing our need to autoload the users code.

It also means we could possibly update the canMutate function to take this information as a param. 
e.g.

```php
public function canMutate(Node $node, InfectionReflectionClass $reflection): bool;
```

Once again ive chosen terrible names, and as im writing this i realize `InfectionReflectionClass` should probably be something like `ClassReflection`.


src/Reflection/CoreInfectionReflectionClass.php
```diff
@@ -0,0 +1,72 @@
...
+use ReflectionException;
+
+/**
+ * @internal
+ */
+final class CoreInfectionReflectionClass implements InfectionReflectionClass
+{
+    private $reflectionClass;
+
+    public function __construct(ReflectionClass $reflectionClass)
```
theofidry: Maybe you can add a static factory method to avoid to have to instantiate a `ReflectionClass` (as a consumer)?

src/Reflection/CoreInfectionReflectionClass.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        $this->reflectionClass = $reflectionClass;
+    }
+
+    public function hasParentOfVisibility(string $methodName, Visibility $visibility): bool
```
theofidry: `hasParentMethodWithVisibility`?

src/Reflection/InfectionReflectionClass.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Reflection;
+
+/**
+ * @internal
+ */
+interface InfectionReflectionClass
```
theofidry: I'm really not convinced by having an interface here

src/Reflection/Visibility.php
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Reflection;
+
+/**
+ * @internal
+ */
+final class Visibility
```
theofidry: For me this class smells the "I want a real enum in PHP" even at the cost of working around the language too much. IMO just having a `public const ALL = [self::PUBLIC, self::PROTECTED]` instead of L46-L74 would be enough. You can add an `Assert::oneOf($visibility, Visibility::ALL)` statement if you wish.

tests/phpunit/Reflection/CoreInfectionReflectionClassTest.php
```diff
@@ -0,0 +1,119 @@
...
+    public function test_it_wraps_the_name(): void
+    {
+        $reflection = new CoreInfectionReflectionClass(new ReflectionClass($this));
+        $this->assertSame($reflection->getName(), get_class($this));
+    }
+
+    /**
+     * @dataProvider provideParentMethodCases
+     */
+    public function test_it_knows_if_a_function_is_inherited(CoreInfectionReflectionClass $reflection, string $method, Visibility $visibility, bool $hasParent): void
```
theofidry: this like looks a bit long, maybe some line returns here for the parameters?

tests/phpunit/Reflection/CoreInfectionReflectionClassTest.php
```diff
@@ -0,0 +1,119 @@
...
+use Generator;
+use Infection\Reflection\CoreInfectionReflectionClass;
+use Infection\Reflection\Visibility;
+use Infection\Visitor\CloneVisitor;
+use PHPUnit\Framework\TestCase;
+use ReflectionClass;
+
+final class CoreInfectionReflectionClassTest extends TestCase
+{
+    public function test_it_wraps_the_name(): void
```
theofidry: `it_exposes_the_class_name`

src/Reflection/Visibility.php
BackEndTea: Its probably a bit overkill, but now this class has an API that can't be worngly used.

BackEndTea: True, but now this class can't be used wrong, but i agree its a bit overkill

src/Reflection/InfectionReflectionClass.php
BackEndTea: I have updated the PR, and there are now multiple implementation of this interface.

tests/phpunit/Mutator/AbstractMutatorTestCase.php
```diff
@@ -145,7 +145,7 @@ protected function setUp(): void
         $mutants = [];
 
         foreach ($mutations as $mutation) {
-            $mutatorVisitor = new SimpleMutatorVisitor($mutation);
+            $mutatorVisitor = new MutatorVisitor($mutation);
```
BackEndTea: I ran into some issues with the `SimpleMutatorVisitor`.
Mainly that it calls 'canMutate' on every node again, which isn't something the 'normal' implementation does, so i've opted to replace it with the normal variant

src/Mutation/NodeTraverserFactory.php
```diff
@@ -56,6 +56,10 @@ public function create(array $extraVisitors): PrioritizedVisitorsNodeTraverser
         $traverser = new PrioritizedVisitorsNodeTraverser(new NodeTraverser());
 
         $traverser->addPrioritizedVisitor(new NotMutableIgnoreVisitor(), 50);
+        $traverser->addPrioritizedVisitor(new NodeVisitor\NameResolver(null, [
```
theofidry: missing an import statement: we didn't use partial imports for the visitors in other places here

src/Mutator/IgnoreMutator.php
```diff
@@ -82,10 +82,10 @@ public function canMutate(Node $node): bool
             return false;
         }
 
-        /** @var ReflectionClass|false $reflectionClass */
-        $reflectionClass = $node->getAttribute(ReflectionVisitor::REFLECTION_CLASS_KEY, false);
+        /** @var ClassReflection|null $reflectionClass */
+        $reflectionClass = $node->getAttribute(ReflectionVisitor::REFLECTION_CLASS_KEY);
 
-        if ($reflectionClass === false) {
+        if (!$reflectionClass instanceof ClassReflection) {
```
theofidry: `!($reflectionClass instanceof ClassReflection)`? I always find the `!` horribly confusing in this scenario

src/Reflection/AnonymousClassReflection.php
```diff
@@ -0,0 +1,118 @@
...
+
+        return self::hasMethodRecursively($parent, $methodName, $visibility);
+    }
+
+    private static function hasMethod(
+        ReflectionClass $reflectionClass,
+        string $methodName,
+        Visibility $visibility
+    ): bool {
+        if ($reflectionClass->hasMethod($methodName)) {
```
theofidry: I would invert the expression to reduce the indent here

src/Reflection/CoreClassReflection.php
```diff
@@ -0,0 +1,77 @@
...
+final class CoreClassReflection implements ClassReflection
+{
+    private $reflectionClass;
+
+    private function __construct(ReflectionClass $reflectionClass)
+    {
+        $this->reflectionClass = $reflectionClass;
+    }
+
+    public static function fromClassName(string $className): self
```
theofidry: I would say the static methods go before the non-static ones... But maybe it's something we should fix via the CS-Fixer rather

src/Reflection/Visibility.php
theofidry: I think it's the kind of decision that fits better in a dedicated PR with an ADR entry tbh: it's not critical to the PR and should be relatively easy to enforce everywhere in one go

On a side note: if you wanted a close real enum, it would be preferable to implement the singleton pattern there to ease the comparison of two instances

tests/phpunit/Fixtures/SimpleMutation.php
```diff
@@ -4,10 +4,12 @@
 
 namespace Infection\Tests\Fixtures;
 
+use Infection\MutatedNode;
+use Infection\Mutation\Mutation;
 use Infection\Mutator\Mutator;
 use PhpParser\Node;
 
-class SimpleMutation
+class SimpleMutation extends Mutation
```
theofidry: Might be worth looking in another PR if `SimpleMutation` is still necessary? I feel like a test factory for `Mutator` might be more appropriate than this class... if necessary at all

tests/phpunit/Visitor/ReflectionVisitorTest.php
```diff
@@ -316,6 +338,10 @@ private function parseAndTraverse(string $code, ?NodeVisitor $nodeVisitor = null
         $this->traverse(
             $nodes,
             [
+                new NodeVisitor\NameResolver(null, [
```
theofidry: missing import statement?

tests/phpunit/Visitor/ReflectionVisitorTest.php
```diff
@@ -316,6 +339,10 @@ private function parseAndTraverse(string $code, ?NodeVisitor $nodeVisitor = null
         $this->traverse(
             $nodes,
             [
+                new NameResolver(null, [
```
theofidry: Instead of having it repeated over, since it requires some special settings, what about having a static factory for it?

src/Reflection/Visibility.php
theofidry: Still don't like it, but I'll leave it to the others to whether they are fine with it or would prefer the issue to be tackled in a dedicated PR

src/Visitor/ReflectionVisitor.php
```diff
@@ -184,4 +182,26 @@ private function isFunctionLikeNode(Node $node): bool
...
+    private function getClassReflectionForNode(Node\Stmt\ClassLike $node): ClassReflection
+    {
+        /** @var Node\Name|null $fqn */
+        $fqn = $node->getAttribute(FullyQualifiedClassNameVisitor::FQN_KEY);
+
+        if ($fqn instanceof Node\Name) {
+            return CoreClassReflection::fromClassName($fqn->toString());
+        }
+
+        assert($node instanceof Node\Stmt\Class_);
```
theofidry: I'm not sure to understand this assert neither how we get from `ClassLike` to `Class_` here... Is it based on the existence of the name?

BackEndTea: If the classlike has no name then it is an anonymous class. This is the best way i could think of to make that clear to our tooling etc.

PR #995, opened 2020-01-17
An empty phpfile should have no AST

BackEndTea:
This PR fixes the fileparser test, so that it also works on php 7.4


theofidry:
I'm not sure to understand: how was that working before?

BackEndTea:
There was no new line after the `<php`, which meant it was seen as inline HTML, rather than the php opening tag, which was fixed in php 7.4

theofidry:
👍

PR #996, opened 2020-01-17
This should be ran when the normal mutation runs

BackEndTea:
This PR changes the mutator tests so the mutation happends at the right time.

### DIFF
--- a/tests/phpunit/Mutator/AbstractMutatorTestCase.php
+++ b/tests/phpunit/Mutator/AbstractMutatorTestCase.php
@@ -190,7 +190,7 @@ abstract class AbstractMutatorTestCase extends TestCase
         );
 
         (new NodeTraverserFactory())
-            ->create([100 => $mutationsCollectorVisitor])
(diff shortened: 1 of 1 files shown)


maks-rafalko:
What was the issue that this change fixes? Is it possible to cover it by test?

BackEndTea:
It shouldn't be an issue, because this visitor uses leaveNode, while a lot of others use enterNode.

I ran into this while working on the reflection wrappers, so i decided to make a seperate PR for it.

theofidry:
>I ran into this while working on the reflection wrappers, so i decided to make a seperate PR for it.

I mean the question here is why that change? We just see a `100` -> `10` but why is that?

BackEndTea:
the 100 to 10 is the change of priority for the muation collector that is used by the tests. (The simple mutation collector).

In the current situation this visitor was actually the first to ran. The only reason this wasn't an issue, is that this visitor uses leaveNode, and the other ones use enterNode, causing them to still be executed earlier

PR #997, opened 2020-01-17
Move event related things under the same namespace

theofidry:
This PR proposes to group the event dispatchers, subscribers, events and listeners under the `Event` namespace.

In general, I believe, when you have categories of things, you either group them by boundaries or by type. For example:

```
# Grouping by boundary
src/
    User/
        User.php
        UserId.php
        UserRepository.php
```

```
# Grouping by type
src/
    ValueObject/
        UserId.php
    Model/
        User.php
    Repository/
        UserRepository.php
```

Those two approaches are not necessarily exclusive, for example you can apply one into other or vice-versa.

In any case, I feel the current event-related organisation is a bit too spread without any real value, so the proposal here is to group them under a common namespace and then keep their separation by category


theofidry:
@maks-rafalko pushed something but not 100% that I understood what you meant

theofidry:
>I also suggested to rename EventDispatcher folder to just Dispatcher since it's already in Event dir (some king of duplicate wording), but it's up to you

Coming up next!

PR #998, opened 2020-01-17
Rename respectively `EventDispatcher` and `EventDispatcherInterface` into `SimpleEventDispatcher` and `EventDispatcher`

theofidry:


theofidry:
You're right let's remove it for now

PR #1000, opened 2020-01-17
Rename `Timer` to `Stopwatch`

theofidry:
- Rename `Timer` to `Stopwatch`
- Leverage the already installed PHPUnit Bridge to mock the time-sensitive functions to be able to better test `Timer`


src/Performance/Listener/PerformanceLoggerSubscriber.php
```diff
@@ -54,7 +54,7 @@ final class PerformanceLoggerSubscriber implements EventSubscriber
     private $memoryFormatter;
 
     public function __construct(
-        Timer $timer,
+        Stopwatch $timer,
```
maks-rafalko: What do you think about being consistent here: `Stopwatch` class and `$stopwatch` parameters, variables, properties?

Here and in other classes below.

(similar to what Symfony does https://symfony.com/doc/current/components/stopwatch.html#sections)

theofidry: totally, should be corrected

PR #1001, opened 2020-01-17
Consolidate TimeFormatter tests

theofidry:


tests/phpunit/Performance/Time/TimeFormatterTest.php
```diff
@@ -100,23 +106,13 @@ public function minutesProvider()
...
-    }
-
-    public function hoursProvider()
+    private static function hoursProvider(): Generator
     {
         yield [3600, '1h'];
 
         yield [3600.99, '1h'];
 
-        yield [3601, '1h 0m 1s'];
```
maks-rafalko: how does it still work? this is a breaking change but I don't see that the source code is changed

theofidry: I was as confused and its simply becasue the provider was not used...

maks-rafalko: heh, interesting ;)

PR #1002, opened 2020-01-18
Improve MemoryFormatter

theofidry:
Depends on #1001

### DIFF
--- a/tests/phpunit/Performance/Memory/MemoryFormatterTest.php
+++ b/tests/phpunit/Performance/Memory/MemoryFormatterTest.php
@@ -37,6 +37,7 @@ namespace Infection\Tests\Performance\Memory;
 
 use Generator;
 use Infection\Performance\Memory\MemoryFormatter;
+use InvalidArgumentException;
 use PHPUnit\Framework\TestCase;
 
 final class MemoryFormatterTest extends TestCase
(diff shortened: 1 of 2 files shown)


tests/phpunit/Performance/Memory/MemoryFormatterTest.php
```diff
@@ -53,19 +54,33 @@ protected function setUp(): void
...
         $timeString = $this->memoryFormatter->toHumanReadableString($bytes);
 
         $this->assertSame($expectedString, $timeString);
     }
 
-    public function bytesProvider()
+    public function bytesProvider(): Generator
     {
-        yield [1048576, '1.00MB'];
+        yield [-10, '-10.00B'];
```
maks-rafalko: Is it real example? why would we need it? I'm not sure [memory_get_usage()](https://www.php.net/manual/ru/function.memory-get-usage.php) can return negative numbers

theofidry: not a real need but there is also not a need to guard against it... so at least it captures the behaviour

theofidry: Actually we do adapt the code so instead we might just guard against this case. Updated

PR #1005, opened 2020-01-19
Introduce specific configuration objects for the mutators

theofidry:
Introduces more specific configuration objects for the mutators to ease the consumption of the mutator settings, effectively removing the need for `Util\MutatorConfig`.

The creation of the config objects as well as further configuration in the mutators (e.g. `MBString` that creates converters based on the config) is all done whilst creating the `Configuration` object. As a result: if there is any error due to a misconfiguration, this will happen early in the process


src/Mutator/Boolean/TrueValueConfig.php
```diff
@@ -0,0 +1,81 @@
...
+
+    private $allowedFunctions = [];
+
+    /**
+     * @param array<string, bool> $settings
+     */
+    public function __construct(array $settings)
+    {
+        foreach ($settings as $functionName => $enabled) {
+            Assert::boolean(
```
theofidry: I added tests for the asserts here in this PR. I don't think we should usually do this, but since there is a special error message and it will be potentially be helpful even for us, not the users (they shouldn't hit any of those), I added the tests to make sure the error message is the one we expect

src/Mutator/MutatorFactory.php
```diff
@@ -70,14 +77,36 @@ public function create(array $resolvedMutators): array
...
-            /** @var string[] $settings */
+            /** @var mixed[] $settings */
             $settings = $config['settings'] ?? [];
             /** @var string[] $ignored */
             $ignored = $config['ignore'] ?? [];
 
-            // TODO: only pass the mutator config if necessary
-            /** @var Mutator $mutator */
-            $mutator = new $mutatorClass(new MutatorConfig($settings));
+            switch ($mutatorClass) {
```
theofidry: I think we can find a more elegant way to do this, but I prefer to keep it simple for this PR

tests/phpunit/Mutator/AbstractMutatorTestCase.php
```diff
@@ -76,8 +75,14 @@ abstract class AbstractMutatorTestCase extends TestCase
...
 
+    /**
+     * @var MutatorFactory|null
+     */
+    private static $mutatorFactory;
+
     protected function setUp(): void
     {
+        // TODO: refactor this bit...
         $this->mutator = $this->createMutator();
```
theofidry: I need to double check this: it is weird that we instantiate the mutator twice in this test-case... And once with the default settings

tests/phpunit/Mutator/AbstractMutatorTestCase.php
```diff
@@ -126,10 +131,12 @@ protected function setUp(): void
 
     final protected function createMutator(array $settings = []): Mutator
     {
-        $class = get_class($this);
-        $mutator = substr(str_replace('\Tests', '', $class), 0, -4);
+        $mutatorClassName = SourceTestClassNameScheme::getSourceClassName(get_class($this));
 
-        return new $mutator(new MutatorConfig($settings));
+        // TODO: this is a bit ridicule...
+        return self::getMutatorFactory()->create([
```
theofidry: As per the TODO: I think this is really bad in terms of DX (for us). It's ok in the source codebase because of the way it is used, but maybe here in the test codebase we can create a helper to make it a bit nicer...

src/Mutator/Extensions/BCMath.php
```diff
@@ -106,46 +107,71 @@ public function canMutate(Node $node): bool
...
-                $this->makeBinaryOperatorMapper(Node\Expr\BinaryOp\Spaceship::class)
-            ),
-        ];
-
-        $functionsToRemove = array_filter($functionsMap, static function (bool $isOn): bool {
-            return !$isOn;
-        });
-
-        $this->converters = array_diff_key($converters, $functionsToRemove);
+        return array_intersect_key(
```
theofidry: hm that CS change is the fixer I think... well at least it doesn't hit the 140 limit now I guess

src/Mutator/Extensions/BCMath.php
```diff
@@ -106,46 +107,71 @@ public function canMutate(Node $node): bool
...
+                'bccomp' => self::makeCheckingMinArgsMapper(
+                    2,
+                    self::makeBinaryOperatorMapper(Node\Expr\BinaryOp\Spaceship::class)
+                ),
+            ],
+            array_fill_keys($functionsMap, null)
+        );
     }
 
-    private function makeCheckingMinArgsMapper(int $minimumArgsCount, callable $converter)
```
theofidry: `callable` is a far more tricky type to handle than `Closure` hence if we can, I think `Closure` is better

theofidry:
ping @infection/core: if you have concerns about some aspects of the PR please say it so

theofidry:
Indeed since we refactored the mutator factory and call it in the `ConfigurationFactory`, any config error would be thrown at that level, before anything runs. So there was nothing to do there! Sorry for the confusion I forgot to remove it (I wanted to double check there was not anything lazily done in the mutators as well).

Regarding the other critics, I thought a bit more about it and reflected about the changes of this PR and I would note the following:

- We have little configurable mutators
- None of the configs are incredibly complex (they all have been simplified a lot by moving out the ignore part in `IgnoreMutator`
- The current setup is simple: one class, on constrained constructor, convention to parsed the config at instantiation time

What I don't like about it:

- We hardcode the configurable mutators
- We do the parsing + validation in the mutators which means:
  - Adding noise to the mutator (although as you pointed out, we do only need one class)
  - Probably add more noise in the future: I would like to register "recommendations" and "warnings" when interpreting the config inputs to create the `Configuration` object [like I do in Box](https://github.com/humbug/box/pull/283/files#diff-966304436a0107a79fa1f8b381b2d86bR2915)
- Rely on the mutator doing the parsing of the config in the constructor: if it does it later, i.e. lazily, the config error will appear much later in the process

Now how does this PR solves any of this?

- Make the configurable mutators "discoverable" (by checking the implemented interface instead of hardcoded list)
- Move the parsing + validation of the config out in a dedicated class which is passed to the mutator constructor hence we are guaranteed that this process is done when creating the mutators in `ConfigurationFactory` and not lazily

That said, in light of the initial notes this current approach might be a bit "heavy" since we have so few configurable mutators and makes it a bit more verbose to implement a configurable mutator as well.

So I think we "just" have to make a choice here: accept the additional noise of parsing + validation of the config in the mutators themselves, which allows to keep things simple, or move them out to keep the mutators simpler.

I don't think the discoverability point is an issue in either case: I'm confident we can solve it efficiently in both cases.

theofidry:
Hm correction: the discoverability part is actually implemented in #1012 right now.

src/Mutator/Extensions/MBString.php
```diff
@@ -56,13 +57,16 @@ final class MBString implements Mutator
...
     private $converters;
 
-    public function __construct(MutatorConfig $config)
+    public function __construct(MBStringConfig $config)
     {
-        $settings = $config->getMutatorSettings();
+        $this->converters = self::createConverters($config->getAllowedFunctions());
 
-        $this->setupConverters($settings);
+        $x = '';
```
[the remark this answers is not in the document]
theofidry: Leftover of a debug statement... 😅

theofidry:
ping @infection/core

PR #1006, opened 2020-01-20
Introduce `MutatorResolver`

theofidry:
The current `MutatorFactory` is doing a bit too much IMO. Indeed it:

- _resolves_ the mutators, i.e. working out from the list of mutator names/class names and profiles with settings, the final list of mutator class names with their settings
- instantiate the mutators from that resolved list

Since so far the instantiation part was relatively trivial, it was ok. But as I'm introducing #1004 and #1005, the instantiate is going to get a bit more interesting. As a result, it makes more sense to split up the factory now with `MutatorResolver` which does the resolving part

### DIFF
--- a/src/Configuration/ConfigurationFactory.php
+++ b/src/Configuration/ConfigurationFactory.php
@@ -36,13 +36,16 @@ declare(strict_types=1);
 namespace Infection\Configuration;
 
 use function array_fill_keys;
+use function count;
 use function dirname;
 use Infection\Configuration\Entry\PhpUnit;
 use Infection\Configuration\Schema\SchemaConfiguration;
 use Infection\FileSystem\SourceFileCollector;
 use Infection\FileSystem\TmpDirProvider;
+use Infection\Mutator\Mutator;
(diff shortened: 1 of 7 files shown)


src/Configuration/ConfigurationFactory.php
```diff
@@ -189,6 +186,28 @@ private function retrievePhpUnit(SchemaConfiguration $schema, string $configDir)
...
+     */
+    private function retrieveMutators(array $schemaMutators, string $mutatorsInput): array
+    {
+        if (count($schemaMutators) === 0) {
+            $schemaMutators = ['@default' => true];
+        }
+
+        $parsedMutatorsInput = $this->mutatorParser->parse($mutatorsInput);
+
+        if ([] === $parsedMutatorsInput) {
```
maks-rafalko: what do you think about `[] === $parsedMutatorsInput` as we don't use yoda style here? Line 194 above have another style ;)

theofidry: I'll look into the CS-Fixer... I hate to have to think about that one

PR #1008, opened 2020-01-20
Use method_exists instead of versions

### LINKED ISSUE #1007: Don't use PackageVersions to examine symfony/process version
| Question    | Answer
| Infection version | 0.15.0
| Test Framework version | PHPUnit 8.5.0 (doesn't matter)
| PHP version | 7.4
| Platform    | any 
| Github Repo | https://github.com/scaytrase/package-version-fallback-issue

Related travis build

https://travis-ci.org/scaytrase/package-version-fallback-issue/builds/639467674

Related code
https://github.com/infection/infection/commit/a6598582ac1c4eb9b83ff0acdb829ed1d0b76b61

This commit makes infection fail for several reasons:

1. symfony/symfony:^5 is used instead of symfony/process:^5
This causes `Version` class to fallback to it's own composer.lock which contains version 4 of symfony/process. Infection thinks that there is needed method but it fails to call it, since the installed version comes from `symfony/symfony:^5`

2. Versions are not generated if `--no-scripts` used. This also makes infection think, that it uses version 4 of PackageVersions' lock file. Moreover in this case the infection itself thinks that it runs versions 0.13.4, which is pinned in PackageVersions' lock file

Please examine related repo and travis build for more info.

Suggested fix: use `method_exists` instead (I'll supply a PR)

scaytrase:
This PR:

Fixes #1007

Replace usage of `VersionParser` with `method_exists` since its behavior is not consistent with different dependency options (for example. transitive dependencies)

### DIFF
--- a/src/Process/Builder/InitialTestRunProcessBuilder.php
+++ b/src/Process/Builder/InitialTestRunProcessBuilder.php
@@ -37,8 +37,7 @@ namespace Infection\Process\Builder;
 
 use Infection\Console\Util\PhpProcess;
 use Infection\TestFramework\TestFrameworkAdapter;
-use Infection\Utils\VersionParser;
-use PackageVersions\Versions;
+use function method_exists;
 use Symfony\Component\Process\Process;
 
 /**
@@ -47,12 +46,10 @@ use Symfony\Component\Process\Process;
 class InitialTestRunProcessBuilder
 {
     private $testFrameworkAdapter;
-    private $versionParser;
 
-    public function __construct(TestFrameworkAdapter $testFrameworkAdapter, VersionParser $versionParser)
+    public function __construct(TestFrameworkAdapter $testFrameworkAdapter)
     {
         $this->testFrameworkAdapter = $testFrameworkAdapter;
-        $this->versionParser = $versionParser;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/Process/Builder/InitialTestRunProcessBuilder.php
```diff
@@ -77,9 +73,7 @@ public function createProcess(
 
         $process->setTimeout(null); // ignore the default timeout of 60 seconds
 
-        $symfonyProcessVersion = $this->versionParser->parse(Versions::getVersion('symfony/process'));
-
-        if (version_compare($symfonyProcessVersion, '4.4.0', '<')) {
+        if (method_exists($process, 'inheritEnvironmentVariables')) {
```
theofidry: This function should be imported (`use function method_exists;`)

scaytrase: fixed

theofidry:
@scaytrase needs CS fixing now :)

maks-rafalko:
Thank you @scaytrase for the reported issue and this fix.

Question: shouldn't it emit a deprecation now when for example Symfony 4.4.1 is installed? 

https://github.com/symfony/process/blob/v4.4.1/Process.php#L1222

We are ok in Infection itself as we have a lock file (`4.3.6` is locked), but what if the target project uses Symfony 4.4.1? Not sure this is a problem though ...



> @maks-rafalko in light of this PR, I believe the `VersionParser` might be moved under the `TestFramework` namespace as it is used only for the test frameworks now.

yep, thank you

sanmai:
It should, and it'll be a bummer if we happen to collect deprecation warnings ourselves.

PR #1012, opened 2020-01-21
Introduce ConfigurableMutator

theofidry:
Depends on #1005 

This PR introduces an abstraction to allow a bit more of flexibility for introducing a configurable mutator:

- A configurable mutator should implement `ConfigurableMutator` which requires to declare the `getConfigClassName()` method
- The `GetConfigClassName` which can be used default the config to being `\...\MutatorClassNameConfig`
- A `MutatorConfig` interface which each config needs to implement and force how a config should be instantiated
- More mutator related autoreview tests to ensure all the above is done correctly


theofidry:
ping @infection/core

src/Mutator/MutatorFactory.php
```diff
@@ -82,33 +76,16 @@ public function create(array $resolvedMutators): array
...
-                    /** @var Mutator $mutator */
-                    $mutator = new $mutatorClass();
+                $mutator = new $mutatorClassName(new $configClassName($settings));
+            } else {
+                $mutator = new $mutatorClassName();
             }
+            /* @var Mutator $mutator */
 
-            $mutators[$mutator->getName()] = new IgnoreMutator(
+            $mutators[(string) $mutator->getName()] = new IgnoreMutator(
```
maks-rafalko: `public function getName(): string;`

why is it needed?

src/Mutator/MutatorFactory.php
```diff
@@ -82,33 +76,16 @@ public function create(array $resolvedMutators): array
             /** @var string[] $ignored */
             $ignored = $config['ignore'] ?? [];
 
-            switch ($mutatorClass) {
-                case BCMath::class:
-                    $mutator = new BCMath(new BCMathConfig($settings));
+            if (in_array(ConfigurableMutator::class, class_implements($mutatorClassName), true)) {
```
BackEndTea: Why not use an instance of check here?

BackEndTea: If we use this interface it should also declare the constructor.

theofidry: >Why not use an instance of check here?

Could probably do the job, dunno might have mixed things up with the test provider 😅 

>If we use this interface it should also declare the constructor.

Why?

src/Mutator/MutatorFactory.php
theofidry: I think it was an SA error, will double check

src/Mutator/MutatorFactory.php
BackEndTea: Because we rely on the fact that the constructor takes The config class as a param.

theofidry: This won't work https://3v4l.org/407tp

You could keep `MutatorConfig` but then we would loose all the benefits of having a specific config... Also as of now this is guaranteed by the tests which uses on the factory

Another alternative is too keep `__construct(array $settings)` in `Mutator` and just instantiate the config directly from the mutator...

src/Mutator/MutatorFactory.php
theofidry:
```
Method Infection\Mutator\MutatorFactory::create() should return array<string, Infection\Mutator\Mutator> but returns array<int|string,
         Infection\Mutator\IgnoreMutator>.
```

It is unclear to me why...

src/Mutator/MutatorFactory.php
theofidry: >Why not use an instance of check here?

Ha obviously: because we don't have the mutator instance yet, just its class name. So `instanceof` won't work there

PR #1016, opened 2020-01-21
Move the Command namespace under Console

theofidry:


maks-rafalko:
I'm personally -1 on this one, don't see a reason for this change. Also, I get used to see `src/Command` in Symfony applications so it's more convenient to see `src/Command` here

theofidry:
The reason for this change is that it’s a bit confusing to have both
Console and Command as top level namespaces when they effectively belong to
the same component.

And the commands are not the most interesting part of the application
either (IMO)

sanmai:
Yet that's where you would start exploring an application to learn how it works.

PR #1017, opened 2020-01-21
Group Mutant, Mutation and Mutator under the Mutagen namespace

theofidry:
I've been trying to find a suitable name to group those 3 namespaces together for a while. I'm not 100% set on `Mutagen`, but it looked like a nice fit. I've also seen `mutagenesis` but this is more the process

>Mutagen
>an agent, such as radiation or a chemical substance, which causes genetic mutation.

>Mutagenesis /mjuːtəˈdʒɛnɪsɪs/ is a process by which the genetic information of an organism is changed, resulting in a mutation.

But maybe those 3 namespaces don't need to be grouped together... Let's hear what you think of it :)


maks-rafalko:
> maybe those 3 namespaces don't need to be grouped together.

This is what I have in my mind. There is no need to group these 3 namespaces. I'm find it more easier to follow the code with split ones, and all of them are from Mutation Testing definitions.

PR #1019, opened 2020-01-24
Use `infection/extension-installer` to automatically register Test Framework Adapters

maks-rafalko:
This will allow to not add any new code to `infection/infection` when the new Test Framework Adapter is created.

It will be automatically discovered and used thanks to `infection/extensions-installer`.

See a detailed documentation about how it works internally: [`infection/extension-installer`](https://github.com/infection/extension-installer#infection---extensions-installer)

Basically, this is an idea taken from `phpstan/extension-installer` and Psalm:

* https://github.com/phpstan/extension-installer
* https://psalm.dev/docs/running_psalm/plugins/authoring_plugins/#authoring-composer-based-plugins

## How it works?

Each Test Framework Adapter (a separate composer package) should use a special type - `infection-extension`

https://github.com/infection/codeception-adapter/blob/6ba9944830a510d10bad9f95e2a1e04d91b9e2a5/composer.json#L2-L3

and declare a class that is responsible to adapter's creation

https://github.com/infection/codeception-adapter/blob/6ba9944830a510d10bad9f95e2a1e04d91b9e2a5/composer.json#L39-L42

`composer require infection/codeception-adapter`:

zsh 2020-01-24 11-05-10

Then, as soon as such a package is installed, `infection/extension-installer` scans a packages (after `composer update` and `composer require` commands) and creates a config inside `vendor` folders.

Then, Infection during its work, checks whether this config contains any of the autodiscovered packages, and if so - uses them.

As simple as that!

## What does it mean for us?

It means that we can add additional Test Framemowrk Adapters and they will be installed / registered automatically after `composer require infection/x-adapter` execution.

## Next steps

Next steps (in different PRs)

* remove a `infection/codeception-adapter` dependency from `infection/infection`'s composer.json file and execute `composer require infection/codeception-adapter` *automatically* during the Configuration Phase
* Add an ability to enable / disable plugins in `infection.json`

### DIFF
--- a/src/TestFramework/Factory.php
+++ b/src/TestFramework/Factory.php
@@ -37,14 +37,15 @@ namespace Infection\TestFramework;
 
 use function implode;
 use Infection\AbstractTestFramework\TestFrameworkAdapter;
+use Infection\AbstractTestFramework\TestFrameworkAdapterFactory;
 use Infection\Configuration\Configuration;
 use Infection\FileSystem\Finder\TestFrameworkFinder;
-use Infection\TestFramework\Codeception\CodeceptionAdapterFactory;
 use Infection\TestFramework\Config\TestFrameworkConfigLocatorInterface;
 use Infection\TestFramework\PhpSpec\Adapter\PhpSpecAdapterFactory;
 use Infection\TestFramework\PhpUnit\Adapter\PhpUnitAdapterFactory;
 use InvalidArgumentException;
 use function Safe\sprintf;
+use Webmozart\Assert\Assert;
 
 /**
  * @internal
@@ -109,19 +110,33 @@ final class Factory
             );
         }
 
-        if ($adapterName === TestFrameworkTypes::CODECEPTION) {
-            $codeceptionConfigPath = $this->configLocator->locate(TestFrameworkTypes::CODECEPTION);
+        $extensionsClassExists = class_exists('\Infection\ExtensionInstaller\GeneratedExtensionsConfig');
 
-            return CodeceptionAdapterFactory::create(
-                $this->testFrameworkFinder->find('codecept'),
-                $this->tmpDir,
-                $codeceptionConfigPath,
-                null,
-                $this->jUnitFilePath,
-                $this->projectDir,
-                $this->infectionConfig->getSourceDirectories(),
-                $skipCoverage
-            );
+        if ($extensionsClassExists) {
+            $installedExtensions = \Infection\ExtensionInstaller\GeneratedExtensionsConfig::EXTENSIONS;
+
+            foreach ($installedExtensions as $packageName => $installedExtension) {
+                $factory = $installedExtension['extra']['class'];
+
+                Assert::classExists($factory);
+
+                if (!is_a($factory, TestFrameworkAdapterFactory::class, true)) {
+                    continue;
+                }
+
+                if ($adapterName === $factory::getAdapterName()) {
+                    return $factory::create(
+                        $this->testFrameworkFinder->find($factory::getExecutableName())
+                        $this->tmpDir,
+                        $this->configLocator->locate($factory::getAdapterName()),
+                        null,
+                        $this->jUnitFilePath,
+                        $this->projectDir,
+                        $this->infectionConfig->getSourceDirectories(),
+                        $skipCoverage
+                    );
+                }
+            }
         }
 
         throw new InvalidArgumentException(sprintf(
--- a/src/TestFramework/Factory.php
+++ b/src/TestFramework/Factory.php
@@ -39,6 +39,7 @@ use function implode;
 use Infection\AbstractTestFramework\TestFrameworkAdapter;
 use Infection\AbstractTestFramework\TestFrameworkAdapterFactory;
 use Infection\Configuration\Configuration;
+use Infection\ExtensionInstaller\GeneratedExtensionsConfig;
 use Infection\FileSystem\Finder\TestFrameworkFinder;
 use Infection\TestFramework\Config\TestFrameworkConfigLocatorInterface;
 use Infection\TestFramework\PhpSpec\Adapter\PhpSpecAdapterFactory;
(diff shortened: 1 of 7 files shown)


src/TestFramework/Factory.php
```diff
@@ -105,19 +106,33 @@ public function create(string $adapterName, bool $skipCoverage): TestFrameworkAd
             );
         }
 
-        if ($adapterName === TestFrameworkTypes::CODECEPTION) {
-            $codeceptionConfigPath = $this->configLocator->locate(TestFrameworkTypes::CODECEPTION);
+        $extensionsClassExists = class_exists('\Infection\ExtensionInstaller\GeneratedExtensionsConfig');
```
theofidry: You can use `::class` here still

src/TestFramework/Factory.php
```diff
@@ -105,19 +106,33 @@ public function create(string $adapterName, bool $skipCoverage): TestFrameworkAd
...
-                $this->tmpDir,
-                $codeceptionConfigPath,
-                null,
-                $this->jUnitFilePath,
-                $this->projectDir,
-                $this->infectionConfig->getSourceDirectories(),
-                $skipCoverage
-            );
+        if ($extensionsClassExists) {
+            $installedExtensions = \Infection\ExtensionInstaller\GeneratedExtensionsConfig::EXTENSIONS;
```
theofidry: should be imported like other symbols

src/TestFramework/Factory.php
```diff
@@ -105,19 +106,33 @@ public function create(string $adapterName, bool $skipCoverage): TestFrameworkAd
...
-                (new TestFrameworkFinder('codecept'))->find(),
-                $this->tmpDir,
-                $codeceptionConfigPath,
-                null,
-                $this->jUnitFilePath,
-                $this->projectDir,
-                $this->infectionConfig->getSourceDirectories(),
-                $skipCoverage
-            );
+        if ($extensionsClassExists) {
```
theofidry: instead of having:

```php
if ($extensionsCLassExists) {
  // ...
}

throw new InvalidArgumentException();
```

You could invert the condition to avoid one nesting level:

```php
if (!$extensionsCLassExists) {
  throw new InvalidArgumentException();
}

// ...
```

src/TestFramework/Factory.php
```diff
@@ -105,19 +106,33 @@ public function create(string $adapterName, bool $skipCoverage): TestFrameworkAd
...
+            $installedExtensions = \Infection\ExtensionInstaller\GeneratedExtensionsConfig::EXTENSIONS;
+
+            foreach ($installedExtensions as $packageName => $installedExtension) {
+                $factory = $installedExtension['extra']['class'];
```
theofidry: if it's from a generated file why not providing the class directly instead?

src/TestFramework/Factory.php
```diff
@@ -105,19 +106,33 @@ public function create(string $adapterName, bool $skipCoverage): TestFrameworkAd
...
+
+                Assert::classExists($factory);
+
+                if (!is_a($factory, TestFrameworkAdapterFactory::class, true)) {
```
theofidry: shouldn't that throw an error instead?

src/TestFramework/Factory.php
maks-rafalko: It's a convention to use `extra.infection` object to store additional data, and for now it's only a `class`.

Later, when we add additional extension points like Mutators, custom OutputFormatters etc, this can be not only a `class`. 

So, `extension-installer` just copies all the information from the plugin, and it's a responsibility of the client code to decide how to use it

src/TestFramework/Factory.php
maks-rafalko: No, because `extension-installer` is not only about about Test framework Adapters. It can autodiscover any plugin with type `infection-extension` (in the future it will be any extension point).

That's why here we have to filter out not needed classes.

We can't do filtering in the `extension-installer` plugin itself because in this case plugin will depend on Test Framework Adapter, Mutator, etc. which is not what is created for.

src/TestFramework/Factory.php
```diff
@@ -105,19 +107,27 @@ public function create(string $adapterName, bool $skipCoverage): TestFrameworkAd
             );
         }
 
-        if ($adapterName === TestFrameworkTypes::CODECEPTION) {
-            $codeceptionConfigPath = $this->configLocator->locate(TestFrameworkTypes::CODECEPTION);
+        foreach (GeneratedExtensionsConfig::EXTENSIONS as $packageName => $installedExtension) {
```
BackEndTea: Would it be worthwhile to inject `GeneratedExtensionsConfig::EXTENSIONS` into the factory? That would allow us to test it.

maks-rafalko: Well, this unfortunately requires to rewrite `TestFrameworkFinder` and inject it as well, because

* it's created inside factory like `new TestFrameworkFinder(...)`
* test framework name is passed to its constructor so it's not reusable for different adapters -> we need to rewrite `TestFrameworkFinder` to probably pass adapter name to `find()` rather than to constructor

I'm not sure I want to do it in this PR.. What do you think?

BackEndTea: I'm not sure i follow. The factory is created from inside the container, so we should be able to inject the extension into the constructor there if i'm not mistaken.

maks-rafalko: I did it locally, but tests failing because of

```
Infection\FileSystem\Finder\Exception\FinderException: Unable to locate a dummy executable on local system. Ensure that dummy is installed and available.
```

Look at this line:

https://github.com/infection/infection/blob/b70eb6b88fe362797bf2ecaa2326a89d8dcfdc64/src/TestFramework/Factory.php#L113

`TestFrameworkFinder` is hardcoded and also should be injected to be able to mock it. And since it gets test framework parameter to *constructor* rather than to `find()` method. - we can't mock it correctly for all test framework. This class IMO should be completely rewritten as I mentioned in the prev comment

maks-rafalko:
@BackEndTea @theofidry adressed your comments

PR #1021, opened 2020-01-24
Add tests to `MutationTestingRunner`

theofidry:
This PR:

- Is another take of #991
- Introduces back the `EventDispatcher` interface since we are implementing another one here. The previous implementation has been renamed `SyncEventDispatcher`: not sure it's the best name so I'm open to propositions
- Add tests to `MutationTestingRunner`


src/Process/Runner/MutationTestingRunner.php
```diff
@@ -75,13 +76,11 @@ public function __construct(
      */
     public function run(array $mutations, int $threadCount, string $testFrameworkExtraOptions): void
     {
-        $mutantCount = count($mutations);
```
maks-rafalko: Why is this change needed (oncluding change on 96 line)?

theofidry: It's no longer necessary, leftovers of reworking the previous PR

theofidry: Well actually can be kept: for now there is a 1-1 mapping between the number of mutations and created mutant processes so it avoids to do the count twice. It's a micro-optimization though so not necessary per se

maks-rafalko: it's a minor thing, up to you. I would kept it as it was just to not touch already existing code and not counting twice

PR #1023, opened 2020-01-26
Use location on interceptor

BackEndTea:
This makes use of the new location const in the interceptor.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,7 +49,7 @@
         "composer/xdebug-handler": "^1.3.3",
         "infection/abstract-testframework-adapter": "^0.2.3",
         "infection/codeception-adapter": "^0.3.0",
-        "infection/include-interceptor": "^0.2.0",
+        "infection/include-interceptor": "^0.2.1",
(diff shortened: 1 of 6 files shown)


maks-rafalko:
We definitely should use a new tag https://github.com/infection/include-interceptor/releases/tag/0.2.1

BackEndTea:
Bumped the version

composer.json
```diff
@@ -49,7 +49,7 @@
         "composer/xdebug-handler": "^1.3.3",
         "infection/abstract-testframework-adapter": "^0.2.3",
         "infection/codeception-adapter": "^0.3.0",
-        "infection/include-interceptor": "^0.2.0",
+        "infection/include-interceptor": "^0.2.1",
```
theofidry: @infection/core since for now we went for the polyrepo approach, how about going all the way and have PRs on them being reviewed as well?

maks-rafalko: It's fine yes, but there was a PR though

PR #1025, opened 2020-01-27
Fix test warning that prevented its execution

BackEndTea:
This PR fixes the `test_it_dispatches_events_even_when_no_mutations_is_given` test.

It gave a phpunit warning that the return type of the mock ans the real class don't match. This prevented the method from being executed any further, as the test itself was also faulty.

### DIFF
--- a/tests/phpunit/Process/Runner/MutationTestingRunnerTest.php
+++ b/tests/phpunit/Process/Runner/MutationTestingRunnerTest.php
@@ -183,23 +183,28 @@ final class MutationTestingRunnerTest extends TestCase
         $threadCount = 4;
         $testFrameworkExtraOptions = '--filter=acme/FooTest.php';
 
+        $this->processBuilderMock
+            ->expects($this->never())
+            ->method($this->anything())
+        ;
+
         $this->mutantFactoryMock
-            ->expects($this->once())
-            ->method('create')
-            ->with($mutations, $testFrameworkExtraOptions)
-            ->willReturn($processes = [])
+            ->expects($this->never())
+            ->method($this->anything())
         ;
 
         $this->parallelProcessRunnerMock
             ->expects($this->once())
             ->method('run')
-            ->with($processes, $threadCount)
+            ->with([], $threadCount)
         ;
 
         $this->runner->run($mutations, $threadCount, $testFrameworkExtraOptions);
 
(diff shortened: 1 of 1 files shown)


tests/phpunit/Process/Runner/MutationTestingRunnerTest.php
```diff
@@ -183,23 +183,28 @@ public function test_it_dispatches_events_even_when_no_mutations_is_given(): voi
         $threadCount = 4;
         $testFrameworkExtraOptions = '--filter=acme/FooTest.php';
 
+        $this->processBuilderMock
```
theofidry: My understanding was that it was the default behaviour of mocks wasn't it?

BackEndTea: I'm 95% sure that unconfigured mocks instead will allow any number of calls on all methods, and they try to return a 'default' value.

theofidry: then can we check? Because I removed it elsewhere as well so that would require to revert that + IIRC there was mention of a `$this->createStub()` in the new versions which would help with that as well

BackEndTea: I've tested it, and my assumption was correct about the default behaviour of phpunit mocks:
The follow test code passes.
```php
<?php

namespace Fooa;

class A{
    public function foo(){}
}

class B{
    /**
     * @var A
     */
    private $a;

    public function __construct(A $a)
    {
        $this->a = $a;
    }

    public function foo(){$this->a->foo();}
}

class ATest extends TestCase
{
    public function testMock(): void
    {
        $a = $this->createMock(A::class);

        $b = new B($a);
        $b->foo();
        $b->foo();
        $b->foo();
        $b->foo();
        $b->foo();
        $this->assertTrue(true);

    }
}

```

BackEndTea:
Thanks @maks-rafalko & @theofidry

tests/phpunit/Process/Runner/MutationTestingRunnerTest.php
theofidry: I'll try to double check our codebase then because we most likely have that issue in other places

PR #1028, opened 2020-01-27
Update/stan

### LINKED ISSUE #788: We can't add PHPStan plugins, because we use the phar, and the plugins require to use a composer dependency

_Originally posted by @BackEndTea in https://github.com/infection/infection/pull/745#issuecomment-511195465_

What about using composer package instead? This will allow us use so many useful plugins, including 

* https://github.com/thecodingmachine/phpstan-safe-rule (see [why](https://github.com/infection/infection/pull/745#pullrequestreview-261577586))
* https://github.com/thecodingmachine/phpstan-strict-rules
* https://github.com/phpstan/phpstan-phpunit
* and any other

### LINKED ISSUE #882: Improve PHPStan inspection level
Get to a higher level of inspection with PHPStan and in the process fix the leftovers undocumented `array` cases left in the codebase

* See https://github.com/infection/infection/pull/875#discussion_r355182996

BackEndTea:
This PR upgrades phpstan to `0.12.8`, moves it to composer and fixess all issues found by the new version.

Fixes #882
Fixes #788


devTools/phpstan-src.neon
```diff
@@ -1,11 +1,16 @@
...
     ignoreErrors:
+        - '#.* value type specified in iterable type Symfony\\Component\\Process\\Process#'
+        - '#generic class ReflectionClass (but )?does not specify its types#'
         - '#expects .*, array<string>\|bool\|string\|null given#'
         - '#cast array<string>\|bool\|string\|null to#'
         -
-            path: '%currentWorkingDirectory%/src/Process/Runner/Parallel/ParallelProcessRunner.php'
+            path: '../src/Process/Runner/Parallel/ParallelProcessRunner.php'
             message: '#Left side of [\&]{2} is always true#'
-        - '#Method Infection\Configuration\Schema\SchemaConfigurationFile::getDecodedContents() should return stdClass but returns stdClass|null.#'
```
BackEndTea: The `|` was unescaped, so it ignored a lot of other errors as well.

src/Command/BaseCommand.php
```diff
@@ -47,7 +47,13 @@
  */
 abstract class BaseCommand extends Command
 {
+    /**
+     * @var InputInterface
+     */
```
BackEndTea: These annotations (and others) got added back, because phpstan only infers types for private properties that get set in the constructor.

So if they have a default, and dont get set in the constructor we need to add them, or if they are protected/public

src/TestFramework/SafeQuery.php
```diff
@@ -33,12 +33,25 @@
 
 declare(strict_types=1);
 
-namespace Infection\Config\Guesser;
+namespace Infection\TestFramework;
 
-/**
- * @internal
- */
-interface Guesser
```
BackEndTea: Too bad about the ugly diff here 😢 

I've opted to drop the `Guesser` interface. It was never hinted against, and the implementations return differnt things.

src/TestFramework/SafeQuery.php
```diff
@@ -33,12 +33,25 @@
...
+use DOMElement;
+use DOMNodeList;
+use DOMXPath;
+use Webmozart\Assert\Assert;
+
+trait SafeQuery
```
BackEndTea: The method this trait provides was already implemented in two classes, so i've extracted it into a trait instead, as this PR adds it in more places

devTools/phpstan-tests.neon
```diff
@@ -1,7 +1,11 @@
 parameters:
+    level: 4
```
theofidry: can't it be max?

src/Command/BaseCommand.php
```diff
@@ -47,7 +47,13 @@
...
     protected $input;
+    /**
```
theofidry: missing space in-between

src/Command/ConfigureCommand.php
```diff
@@ -148,6 +148,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
         return 0;
     }
 
+    /**
+     * @param array<string> $sourceDirs
```
theofidry: `string[]` please (same elsewhere)

src/Configuration/ConfigurationFactory.php
```diff
@@ -187,6 +187,8 @@ private function retrievePhpUnit(SchemaConfiguration $schema, string $configDir)
     }
 
     /**
+     * @param array<string, array<string>|bool> $schemaMutators
```
theofidry: I would not add this one or keep `array<string, mixed>`. The downstream type directly depends on the config we allow in `schema.json` and can get more complex than that without raising any warning here, resulting in a false assumption

theofidry: It's actually already incorrect in the case of `MBString`: which is `array<string, string[]|array<string, boolean>>` instead of `array<string>|bool`

src/Console/OutputFormatter/AbstractOutputFormatter.php
```diff
@@ -44,6 +44,9 @@
  */
 abstract class AbstractOutputFormatter implements OutputFormatter
 {
+    /**
+     * @var int
```
theofidry: I don't get it: it has a default value no? Isn't that enough to infer the type?

src/Container.php
```diff
@@ -89,7 +89,6 @@
 use Infection\TestFramework\PhpUnit\Config\XmlConfigurationHelper;
 use Infection\TestFramework\PhpUnit\Coverage\IndexXmlCoverageParser;
 use InvalidArgumentException;
-use function is_callable;
```
theofidry: Why removing this one? Isn't it used in `offsetAdd()`?

src/Container.php
```diff
@@ -37,7 +37,7 @@
 
 use function array_filter;
 use function array_key_exists;
-use Closure;
+use function assert;
```
theofidry: where is that one being used?

src/Container.php
```diff
@@ -739,29 +744,35 @@ public function getConfiguration(): Configuration
...
             throw new InvalidArgumentException(sprintf('Unknown service "%s"', $id));
         }
 
         if (array_key_exists($id, $this->values)) {
-            return $this->values[$id];
+            $value = $this->values[$id];
+        } else {
+            $value = $this->values[$id] = $this->factories[$id]($this);
         }
+        assert($value instanceof  $id);
```
theofidry: weird CS here: also I don't get this assertion

src/Logger/FileLogger.php
```diff
@@ -48,9 +48,21 @@
  */
 abstract class FileLogger implements MutationTestingResultsLogger
 {
+    /**
+     * @var MetricsCalculator
+     */
     protected $metricsCalculator;
+    /**
```
theofidry: missing line return at each of them, and other places as well

src/Mutation/FileParser.php
```diff
@@ -61,7 +61,7 @@ public function __construct(Parser $parser)
     public function parse(SplFileInfo $fileInfo): array
     {
         try {
-            return $this->parser->parse($fileInfo->getContents());
+            return $this->parser->parse($fileInfo->getContents()) ?? [];
```
theofidry: why that change?

src/Mutator/ProfileList.php
```diff
@@ -433,12 +433,12 @@ private function __construct()
     }
 
     /**
-     * @return string[]
+     * @return array<int, string>
```
theofidry: it is `string[]` here...

src/TestFramework/SafeQuery.php
```diff
@@ -33,12 +33,28 @@
...
  */
-interface Guesser
+trait SafeQuery
 {
-    public function guess();
+    /**
+     * @return DOMNodeList|DOMElement[]
+     * @phpstan-return DOMNodeList<DOMElement>
+     */
+    private static function safeQuery(DOMXPath $xPath, string $query): DOMNodeList
```
theofidry: What do you think about making it a static helper instead of a trait?

devTools/phpstan-tests.neon
BackEndTea: It was already level 4 before hand, so i've decided not to change it for now

src/Console/OutputFormatter/AbstractOutputFormatter.php
BackEndTea: Default value isn't enough to infer a type, it needs to be set in the constructor for that.

src/Container.php
BackEndTea: This is needed to understand that the `get` method returns an instance of the `id` class string. It means that phpstan knows this is correct.

```php
    public function getConfiguration(): Configuration
    {
        return $this->get(Configuration::class);
    }
```

src/Mutator/ProfileList.php
BackEndTea: There are places that depend on the keys being integers here.

src/Container.php
BackEndTea: Nope this is no longer needed

src/Mutation/FileParser.php
BackEndTea: This isn't needed, ive instead ignored the error in phpstan, as that can never occur with our settings.

BackEndTea:
I've adressed the CS issues and a few other minor things.

src/TestFramework/SafeQuery.php
BackEndTea: For now i'm happy to leave this as a trait.

BackEndTea:
@theofidry It looks like php-scoper has issues with the `assert` function.

composer.json
```diff
@@ -69,6 +69,9 @@
     },
     "require-dev": {
         "helmich/phpunit-json-assert": "^3.0",
+        "phpstan/extension-installer": "^1.0",
+        "phpstan/phpstan": "^0.12.8",
```
localheinz: Shall we require `phpstan/phpstan:012.9`?

devTools/phpstan-src.neon
```diff
@@ -1,11 +1,20 @@
 includes:
-    - phar://%currentWorkingDirectory%/.tools/phpstan/conf/bleedingEdge.neon
+    - ../vendor/phpstan/phpstan/conf/bleedingEdge.neon
 
 parameters:
+    level: max
+    paths:
+        - ../src
+    inferPrivatePropertyTypeFromConstructor: true
```
localheinz: What do you think about keeping keys sorted by name here?

devTools/phpstan-src.neon
```diff
@@ -1,11 +1,20 @@
...
+        - '#generic class ReflectionClass (but )?does not specify its types#'
         - '#expects .*, array<string>\|bool\|string\|null given#'
         - '#cast array<string>\|bool\|string\|null to#'
         -
-            path: '%currentWorkingDirectory%/src/Process/Runner/Parallel/ParallelProcessRunner.php'
+            path: '../src/Process/Runner/Parallel/ParallelProcessRunner.php'
             message: '#Left side of [\&]{2} is always true#'
-        - '#Method Infection\Configuration\Schema\SchemaConfigurationFile::getDecodedContents() should return stdClass but returns stdClass|null.#'
+        -
+            path: '../src/Mutation/FileParser.php'
```
localheinz: Shall we use the baseline feature instead?

devTools/phpstan-tests.neon
```diff
@@ -1,7 +1,11 @@
...
+    paths:
+        - ../tests/phpunit
     excludes_analyse:
         - %currentWorkingDirectory%/tests/e2e/*
         - %currentWorkingDirectory%/tests/phpunit/Fixtures/*
     ignoreErrors:
         - '#Return type \(void\) of method .*[Visitor|Traverser].* should be compatible with return type .* of method PhpParser\\Node.*#'
```
localheinz: Shall we use the baseline feature instead?

Makefile
```diff
@@ -59,8 +58,8 @@ cs: $(PHP_CS_FIXER)
 
 .PHONY: phpstan
 phpstan: vendor $(PHPSTAN)
-	$(PHPSTAN) analyse src --level=max --configuration ./devTools/phpstan-src.neon --no-interaction --no-progress
-	$(PHPSTAN) analyse tests/phpunit --level=4 --configuration ./devTools/phpstan-tests.neon --no-interaction --no-progress
+	$(PHPSTAN) analyse --configuration devTools/phpstan-src.neon --no-interaction --no-progress
+	$(PHPSTAN) analyse --configuration ./devTools/phpstan-tests.neon --no-interaction --no-progress
```
localheinz:
```suggestion
	$(PHPSTAN) analyse --configuration devTools/phpstan-tests.neon --no-interaction --no-progress
```

src/Configuration/Schema/SchemaConfigurationFactory.php
```diff
@@ -70,8 +70,8 @@ public function create(string $path, stdClass $rawConfig): SchemaConfiguration
     private static function createSource(stdClass $source): Source
     {
         return new Source(
-            (array) self::normalizeStringArray($source->directories ?? []),
-            (array) self::normalizeStringArray($source->excludes ?? [])
+            self::normalizeStringArray($source->directories ?? []),
+            self::normalizeStringArray($source->excludes ?? [])
```
localheinz:
```diff
diff --git a/src/Configuration/Schema/SchemaConfigurationFactory.php b/src/Configuration/Schema/SchemaConfigurationFactory.php
index 4d554ce4..6f20f2c9 100644
--- a/src/Configuration/Schema/SchemaConfigurationFactory.php
+++ b/src/Configuration/Schema/SchemaConfigurationFactory.php
@@ -70,8 +70,8 @@ class SchemaConfigurationFactory
     private static function createSource(stdClass $source): Source
     {
         return new Source(
-            self::normalizeStringArray($source->directories ?? []),
-            self::normalizeStringArray($source->excludes ?? [])
+            isset($source->directories) ? self::normalizeStringArray($source->directories) : [],
+            isset($source->excludes) ? self::normalizeStringArray($source->excludes) : []
         );
     }
```

Why bother invoking a method when we already know we don't need to normalize an empty array?

theofidry: given the actual code, purely stylistic here and I find the change more expressive tbh

theofidry:
Closing in favour of #1040

PR #1030, opened 2020-02-04
Use sebastian/environment to better detect non-interactive mode

sanmai:
This PR:

- [x] Make Infection detect a non-interactive mode more reliable
- [x] Covered by tests

Non-interactive mode detection [methods used by Symfony's Application](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Console/Application.php#L870) are not exhaustive, which leads to a possibility for Infection to stuck in a state where it waits for user input indefinitely. All while `sebastian/environment` offers [an additional detection method ](https://github.com/sebastianbergmann/environment/blob/master/src/Console.php#L93-L95) which can see thought the case where I had the above issue, when running Infection with [chronic, from moreutils](https://joeyh.name/code/moreutils/), which can also happen under as simple as output redirection.

```
rm infection.json*
php vendor/bin/infection 2>&1 | grep MSI
# will stall indefinitely...
```

Here I propose to use an interactive mode detection provided by `sebastian/environment` as an additional safeguard, which add almost no additional dependencies since most projects that use PHPUnit depend on that package indirectly anyway.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,6 +54,7 @@
         "nikic/php-parser": "^4.2.2",
         "ocramius/package-versions": "^1.2",
         "sebastian/diff": "^3.0.2",
+        "sebastian/environment": "^4.2",
         "seld/jsonlint": "^1.7",
         "symfony/console": "^3.4.29 || ^4.0 || ^5.0",
         "symfony/filesystem": "^3.4.29 || ^4.0 || ^5.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,28 +4,28 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "2f4805ddce5cd4d7c9a1a2f7fa7f08c7",
+    "content-hash": "defa1a14ef1c0a0917f06906ed361f73",
     "packages": [
         {
             "name": "composer/xdebug-handler",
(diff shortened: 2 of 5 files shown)


maks-rafalko:
I'm fine with it as long as it fixes some issue 👍

sanmai:
I've added a test case where output is redirected to a file. Simply put, without this change Infection will stall indefinitely if you happen to run it like so, from an unconfigured state:
```
rm infection.json*
php vendor/bin/infection 2>&1 | grep MSI
```

sanmai:
CI says there's something wrong with some E2E tests, which I'll try to get to the bottom of. Let's not merge this for now.

sanmai:
Well, apparently we either have `expect` to automate testing, or we have this enhancement. I'd rather have Infection working under `expect` than otherwise, so I'll be shelving this for now.

theofidry:
@sanmai what about attempting to fix Symfony there for accounting for that detection method? IIRC PHPUnit & Symfony have a couple of somewhat duplicated code mostly for historic reasons to avoid conflicts (the less they rely on each other the less issues there is with major releases)

sanmai:
I don't think it'll be the best idea to add that detection method to Symfony or elsewhere for that matter. It kinda works and sort of efficient, but it breaks common automation tools like `expect` which is a great bummer in terms of end-to-end testing. So I think in our case much better solution would be to add timeouts for inputs.

PR #1031, opened 2020-02-05
Enhancement: Extract StrykerApiKeyResolver

localheinz:
This PR

* [x] extracts a `StrykerApiKeyResolver`
* [x] fixes the order of fields in `BadgeLogger` so it is consistent with the order of constructor parameters (and assignments)

💁‍♂ Related to #939.

### DIFF
--- a/src/Environment/StrykerApiKeyResolver.php
+++ b/src/Environment/StrykerApiKeyResolver.php
@@ -1,0 +1,67 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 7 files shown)


src/Environment/StrykerApiKeyResolver.php
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * @internal
+ */
+final class StrykerApiKeyResolver
+{
+    /**
+     * @throws CouldNotResolveStrykerApiKey
+     */
+    public function resolve(array $environment): string
```
maks-rafalko: What do you think about specifying the type of the array?

localheinz: 👍

src/Environment/StrykerApiKeyResolver.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        $names = [
+            'INFECTION_BADGE_API_KEY',
+            'STRYKER_DASHBOARD_API_KEY',
+        ];
+
+        foreach ($names as $name) {
+            if (!array_key_exists($name, $environment) || !is_string($environment[$name])) {
```
localheinz: With increasing `phpstan` levels, we will probably observe complaints about the second part of this condition always evaluating to `false` (since we have now specified an expected value type for elements of the `$environment` array).

localheinz:
Thank you, @maks-rafalko!

PR #1040, opened 2020-02-11
Update PHPStan

### LINKED ISSUE #788: We can't add PHPStan plugins, because we use the phar, and the plugins require to use a composer dependency

### LINKED ISSUE #882: Improve PHPStan inspection level

theofidry:
Replacement for #1028

Fixes #882
Fixes #788


theofidry:
@BackEndTea since I'm the source of the main conflicts I attempted to take over. I wish I could push the test config to max level but unfortunately looks hard to do so

theofidry:
Merging this one since it's mostly @BackEndTea's work which had already been approved

PR #1042, opened 2020-02-11
Allow the initial test suite to be skipped

duncan3dc:
This PR:

- [x] Adds new feature ...
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/154

In an effort to reduce our build times we started to use the `--coverage` option to use previously generated code coverage stats, however we noticed that Infection still runs the entire test suite before beginning mutations, effectively doubling our build time.

I searched through the docs and the only thing I could find that explained this is [here](https://infection.github.io/guide/command-line-options.html#filter):
_"This in no way restricts the initial Infection check on the overall test suite which is still executed in full to ensure all tests are passing correctly before proceeding."_

If Infection doesn't need any other information from this initial run other than the coverage, then it should be possible to bypass it, which is what this PR allows.

I'll happily add tests and a docs PR, but I didn't want to do too much work in case there was something I'd missed that means this isn't a practical option


sanmai:
This is a very dangerous proposition because if tests are not in the passing state where you think they are, even valid mutations will be looking like they were caught.

maks-rafalko:
> This is a very dangerous proposition because if tests are not in the passing state where you think they are, even valid mutations will be looking like they were caught.

That's true. Even if your tests pass from original PHPUnit run, they can fail when being executed by Infection since this is subprocess, different config, etc.

At the same time, I should admit that I needed this option many times, and had to comment out initial tests run right in the code.

What if we add this option with a big red warning in terminal, explaining all the consequences?

Also, this option can be allowed to use *only* with provided coverage (`--coverage`). Otherwise, it does not make any sense.

sanmai:
I think a warning is a must.

duncan3dc:
I will carry on with this when I get some time then, fix the unit tests and add an e2e test for this functionality. Do you want me to prepare a docs PR too, or would you rather decide on the wording yourselves?

maks-rafalko:
Please go ahead with the docs as well 👍  Thank you

src/Engine.php
```diff
@@ -112,6 +112,13 @@ public function execute(int $threads): bool
 
     private function runInitialTestSuite(): void
     {
+        if ($this->config->shouldSkipInitialTestSuite()) {
+            $this->consoleOutput->logSkippingInitialTestSuite();
+            $this->assertCodeCoverageExists(null, $this->config->getTestFramework());
```
maks-rafalko: Probably, it makes sense to create 2 methods and don't use a nullable argument with `if/else`. This would be more readable.

src/Engine.php
```diff
@@ -112,6 +112,13 @@ public function execute(int $threads): bool
 
     private function runInitialTestSuite(): void
     {
+        if ($this->config->shouldSkipInitialTestSuite()) {
```
maks-rafalko: As I mentioned above, it should not be allowed to use `--skip-initial-test-suite` **without** `--coverage`.

Now it's possible:

zsh 2020-02-16 18-11-33

The existing red message is confusing for this case. We need another one that will say exactly about this issue: that `--skip-initial-test-suite` option requires `--coverage`

src/Engine.php
duncan3dc: Is this ok: https://github.com/infection/infection/pull/1042/files#diff-5a8536ad09be922aea0b4725411fea45R180

src/Engine.php
duncan3dc: image

src/Container.php
```diff
@@ -297,7 +297,8 @@ public static function create(): self
                 $config = $container->getConfiguration();
 
                 return new CoverageRequirementChecker(
-                    $config->getCoveragePath() !== '',
+                    $config->shouldSkipCoverage(),
```
duncan3dc: I noticed the wrong argument was being passed here so I corrected it, would you rather this was in a separate PR?

duncan3dc:
`make cs` looks good locally, although `PrettyCI` seems very upset

theofidry:
Might be cache related: try removing the PHP-CS-Fixer cache and running it again

src/Command/InfectionCommand.php
```diff
@@ -174,6 +174,12 @@ protected function configure(): void
                 InputOption::VALUE_REQUIRED,
                 'Extra php options for the initial test runner. Will be ignored if --coverage option presented.'
             )
+            ->addOption(
+                'skip-initial-test-suite',
```
theofidry: `skip-initial-tests`?

src/Command/InfectionCommand.php
```diff
@@ -174,6 +174,12 @@ protected function configure(): void
...
+                null,
+                InputOption::VALUE_NONE,
+                'If --coverage option has been used then this bypasses the initial test suite.'
```
theofidry: `Skips the initial test runs - requires the coverage to be provided via the --coverage option`

src/Command/InfectionCommand.php
theofidry: Would need to reflect the name of the variables used downstream as well

src/Console/ConsoleOutput.php
```diff
@@ -129,4 +129,13 @@ public function logNotInControlOfExitCodes(): void
             'It is your responsibility to disable xdebug/phpdbg unless needed.',
         ]);
     }
+
+    public function logSkippingInitialTestSuite(): void
+    {
+        $this->io->warning([
+            'Skipping the initial test suite run can be very dangerous.' . PHP_EOL .
```
theofidry: I think you can just pass an array of strings here, no need for `PHP_EOL`

sanmai:
Would you mind checking that "Alllow edits from maintainers" checkbox?

Else please update the branch. Thanks.

duncan3dc:
I've done both :+1:

maks-rafalko:
Could you look into failing tests please? I will merge it asap when it's green

duncan3dc:
I _think_ we're all good now, thanks for bearing with me

maks-rafalko:
This is a great contribution, thank you @duncan3dc!

PR #1047, opened 2020-02-12
Enhancement: Extract BuildContextResolver

localheinz:
This PR

* [x] extracts a `BuildContext` value object and a `BuildContextResolver`

Follows #1031.
Related to #939.

💁‍♂ Possible next steps:

- rename `BuildContextResolver` to `TravisCiResolver`
- extract a `BuildContextResolver` interface
- implement a `ChainBuildContextResolver`
- implement and register additional implementations of `BuildContextResolver`

What do you think?

/cc @drupol

### DIFF
--- a/src/Environment/BuildContext.php
+++ b/src/Environment/BuildContext.php
@@ -1,0 +1,61 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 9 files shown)


drupol:
Amazing, nothing to say besides what Travis says:

```
Environment/BuildContextResolver.php                                   
Line 61:
Function sprintf is unsafe to use. It can return FALSE instead of      
throwing an exception. Please add 'use function Safe\sprintf;' at the  
beginning of the file to use the variant provided by the               
'thecodingmachine/safe' library.
```

But I don't know if we are following this.

maks-rafalko:
Thank you @localheinz 👍

localheinz:
Thank you, @drupol and @maks-rafalko!

PR #1048, opened 2020-02-13
Assume non-interactive mode when running in a CI environment

sanmai:
Non-interactive mode detection [methods used by Symfony's Application](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Console/Application.php#L870) are not exhaustive, which leads to a possibility for Infection to stuck in a state where it waits for user input indefinitely. 

```
rm infection.json*
php vendor/bin/infection 2>&1 | grep MSI
# will stall indefinitely...
```

Here I propose to use detection of common CI environment variables as an additional safeguard. This way Infection would not stuck waiting on CI for input if it so happens there's no config file.

This PR:

- [x] Covered by tests
- [x] Fix failed builds if any
- [x] Blocked by #1135

Another take on the problem from #1030


maks-rafalko:
are you going to do something else here or are we happy to merge?

sanmai:
I think I might try to test this.

theofidry:
Shouldn't those environments rely on `SHELL_INTERACTIVE` instead?

sanmai:
[It isn't defined](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) on Travis by default, as I see it.

theofidry:
If that's the case (maybe we should do an `echo $SHELL_INTERACTIVE ` to see?) wouldn't it be simpler to export that value in the travis build?

sanmai:
I've updated the description. Please see if it is any more clear now why we're doing this.

theofidry:
It is clear but I feel like it's easier to "patch" the CI directly by exporting the appropriate environment variable than adding additional safeguard in our code

src/Console/Application.php
```diff
@@ -163,6 +163,10 @@ protected function configureIO(InputInterface $input, OutputInterface $output):
         $output->getFormatter()->setStyle('low', new OutputFormatterStyle('red', null, ['bold']));
         $output->getFormatter()->setStyle('medium', new OutputFormatterStyle('yellow', null, ['bold']));
         $output->getFormatter()->setStyle('high', new OutputFormatterStyle('green', null, ['bold']));
+
+        if ($input->isInteractive() && (getenv('CI') === 'true' || getenv('CONTINUOUS_INTEGRATION') === 'true')) {
```
localheinz: Perhaps it makes sense to use the `BuildContextResolver` here?

After all, when a `BuildContext` could be resolved, we must be running in a continuous integration environment.

localheinz: Ha, looking at #1067, that's not entirely true.

PR #1050, opened 2020-02-13
Remove PhpUnitExtraOptions/PhpSpecExtraOptions implementations and move to adapter level

maks-rafalko:
Why?

* because we have 3 test frameworks, but 2 implementations of extra options ([PHPSpec's one used for Codeception by mistake ;)](https://github.com/infection/infection/pull/1050/files#diff-f20466d79712e5e751a8819581cc4787L251-L254)), and in reality, we need only PHPUnit's one
* While working on extracting PHPSpec adapter to a separate package, I noticed that class `PhpSpecExtraOptions` is just a piece of bad design that really does nothing for PHPSpec
* this logic must be on Adapter level, and this PR does exactly that - moves it to PHPUnit adapter
* PhpSpec now has nothing to do with these extra options - so we got rid of useless class `PhpSpecExtraOptions`


maks-rafalko:
Thank you @sanmai

src/Configuration/ConfigurationFactory.php
```diff
@@ -243,14 +240,8 @@ private static function retrieveCoverageBasePath(
...
-    ): TestFrameworkExtraOptions {
-        $extraOptions = $testFrameworkExtraOptions ?? $schema->getTestFrameworkExtraOptions();
-
-        return $testFramework === TestFrameworkTypes::PHPUNIT
-            ? new PhpUnitExtraOptions($extraOptions)
-            : new PhpSpecExtraOptions($extraOptions)
-        ;
+        SchemaConfiguration $schema
+    ): string {
+        return $testFrameworkExtraOptions ?? $schema->getTestFrameworkExtraOptions() ?? '';
```
theofidry: Wouldn't it be better to force `getTestFrameworkExtraOptions()` to be a `string` here instead of `string|null`? That would avoid the necessity of the last `??`

theofidry: Also if it's a single line might be worth to just inline this method

src/TestFramework/TestFrameworkExtraOptionsFilter.php
```diff
@@ -43,37 +43,21 @@
...
  * @internal
  */
-abstract class TestFrameworkExtraOptions
+final class TestFrameworkExtraOptionsFilter
 {
     /**
-     * @var string
+     * @param string[] $initialRunOnlyOptions
+     *
+     * @throws \Safe\Exceptions\PcreException
```
theofidry: should use import statements here, but I would argue those exceptions should not be shown

maks-rafalko:
I will address your comments @theofidry in a separate PR, thanks

PR #1052, opened 2020-02-16
Extract PhpSpec Test Framework adapter to a separate package

maks-rafalko:
This PR:

* extracts PhpSpec adapter from `infection/infection`
* uses [`infection/phpspec-adapter`](https://github.com/infection/phpspec-adapter/) package

Next step (separate PRs):

Remove `infection/codeception-adapter` and `infection/phpspec-adapter` from `infection/infection` and write a custom "package autoloader" that will be used during configuration step (`ConfigureCommand`)


sanmai:
I would advise against that "package autoloader" because it'll go very far out of way from what a composer package usually does, causing unnecessarily surprise, and will require further ado during the setup stage. It is much more obvious and transparent when a user would explicitly the packages they need, whilst requiring only a two-line mention in the docs, and two-line check in the code.

But if I'm wrong about complexities involved here, please forgive me.

maks-rafalko:
>  would advise against that "package autoloader" because it'll go very far out of way from what a composer package usually does, causing unnecessarily surprise, and will require further ado during the setup stage.

I thought about something like question-answer to load additional dependencies:

`composer require infection/infection`

[...]

```bash
bin/infection --test-framework-phpspec

# it detects that phpspec adapter is not installed

> Would you like to install `infection/phpspec-adapter`? [y]

# user presses `y`

# Infection runs `composer require infection/phpspec-adapter` *in the same process*

# continue execution of `bin/infection` without interrupting, and now with `infection/phpspec-adapter` installed
```

This is not so easy as I can see. I will prepare an RFC where we will discuss a couple of different approaches.

Thank you @sanmai

PR #1053, opened 2020-02-17
Cosmetic changes of PHPUnit adapter

theofidry:
- a few CS changes
- add a few more tests (increase coverage and mutation score)

My goal is to start to go through the test framework part. This is a first round just checking a couple of things, I'll submit more PRs starting to change more stuff afterwards


src/TestFramework/AbstractTestFrameworkAdapter.php
```diff
@@ -187,8 +186,7 @@ protected function buildMutationConfigFile(
     private function getCommandLine(
         string $configPath,
         string $extraOptions,
-        array $phpExtraArgs,
-        bool $skipCoverage
```
theofidry: `$skipCoverage` was unused here

src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
```diff
@@ -72,6 +73,11 @@ public function __construct(
         array $srcDirs,
         bool $skipCoverage
     ) {
+        Assert::notEmpty(
```
theofidry: Add a guard against this case since otherwise we get a cryptic error instead of more "graceful" one L96

src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
```diff
@@ -54,29 +55,31 @@
...
     ) {
-        $this->tempDirectory = $tempDirectory;
+        $this->tmpDir = $tmpDir;
         $this->projectDir = $projectDir;
 
+        $this->originalXmlConfigContent = $originalXmlConfigContent;
         $this->xmlConfigurationHelper = $xmlConfigurationHelper;
         $this->jUnitTestCaseSorter = $jUnitTestCaseSorter;
-
-        $this->dom = new DOMDocument();
```
theofidry: There is no reason to do this in the constructor which adds to the cost of instantiating the builder, instead I moved it in a memoized getter to evaluate this code lazily

src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
```diff
@@ -88,9 +91,7 @@ public function build(
         string $mutationHash,
         string $mutationOriginalFilePath
     ): string {
-        // clone the dom document because it's mutated later
-        $dom = clone $this->dom;
```
theofidry: Could not find evidence that the `clone` was necessary here. Would be happy to add it back if we can find a test capturing this scenario

src/TestFramework/PhpUnit/Config/XmlConfigurationHelper.php
```diff
@@ -122,6 +125,9 @@ public function removeExistingPrinters(DOMXPath $xPath): void
     public function validate(DOMXPath $xPath): bool
     {
         if (self::safeQuery($xPath, '/phpunit')->length === 0) {
+            // TODO: should have the PHPUnit config path passed otherwise we blindly assume
```
theofidry: For another PR

tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapterTest.php
```diff
@@ -35,7 +35,9 @@
 
 namespace Infection\Tests\TestFramework\PhpUnit\Adapter;
 
-use Infection\AbstractTestFramework\MemoryUsageAware;
+use function array_map;
+use Generator;
+use Infection\PhpParser\Visitor\IgnoreNode\PhpUnitCodeCoverageAnnotationIgnorer;
```
theofidry: I'm not very happy with this test:

- It's a pure unit test for a third-party adapter, IMO it should warrant an integration test (either besides this one or instead of this one)
- We do not test anything coming from the abstract adapter which is a bit of a miss IMO

So I'm thinking of making this an integration test and capture the behaviour inherited from the abstract adapter. Do you have any objection or something that I missed?

src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
theofidry: So it is important, at least that's why the build is failing. I'm however unable to find a way to capture this issue, does someone has an idea?

theofidry:
@infection/core PR ready for review despite the failure, as per the comments

src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
theofidry: Found it

tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapterTest.php
maks-rafalko: TBH, I didn't get you. We have tens of e2e tests for PHPUnit adapter. And here is a unit tests. What's wrong with it?

tests/phpunit/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilderTest.php
```diff
@@ -40,33 +40,41 @@
...
-        $builder = new ArgumentsAndOptionsBuilder();
 
         $this->assertSame(
             [
                 '--configuration',
                 $configPath,
-                '--verbose',
-                '--debug',
             ],
-            $builder->build($configPath, '--verbose --debug')
```
maks-rafalko: Why the second parameter is emptied?

tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapterTest.php
theofidry: I feel it would be better to have an integration test for the `PhpUnitAdapter` where you can test all the above - that's fine, but also the following:

- get the version
- get the command lines

Sure an e2e test eventually guarantees that overall it's working, but it's also nice to be able to capture some of that behaviour in a more isolated environment and having a strictly unit test kind of prevents that

src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
```diff
@@ -88,9 +91,7 @@ public function build(
...
-
+        $dom = clone $this->getDom();
```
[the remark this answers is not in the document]
theofidry: Actually found a way to remove it :)

tests/phpunit/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilderTest.php
theofidry: It's a swap around: I moved the "without extra options" as the first one

PR #1057, opened 2020-02-19
Pass the config file path being used to provide more friendly error messages

theofidry:
So far we unconditionally assume the PHPUnit configuration file is `phpunit.xml` (without any path) which is both wrong and very confusing. Instead this PR proposes to always pass the path to provide a more helpful error message


theofidry:
ping @infection/core

PR #1060, opened 2020-02-19
Remove unnecessary @covers annotations

theofidry:
Thanks to the PHPUnit bridge `Symfony\Bridge\PhpUnit\CoverageListener` registered in `phpunit.xml.dist`, we do not need those `@covers` annotations

### DIFF
--- a/tests/phpunit/Environment/BuildContextTest.php
+++ b/tests/phpunit/Environment/BuildContextTest.php
@@ -38,9 +38,6 @@ namespace Infection\Tests\Environment;
 use Infection\Environment\BuildContext;
 use PHPUnit\Framework\TestCase;
 
-/**
- * @covers \Infection\Environment\BuildContext
- */
 final class BuildContextTest extends TestCase
 {
     public function test_constructor_sets_values(): void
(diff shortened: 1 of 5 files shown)


localheinz:
@theofidry 

What do you think about applying this patch

```diff
diff --git a/.php_cs.dist b/.php_cs.dist
index 8959771c..e3feb8db 100644
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -98,6 +98,11 @@ return Config::create()
         'concat_space' => ['spacing' => 'one'],
         'final_static_access' => true,
         'fully_qualified_strict_types' => true,
+        'general_phpdoc_annotation_remove' => [
+            'annotations' => [
+                'covers',
+            ],
+        ],
         'global_namespace_import' => [
             'import_classes' => true,
             'import_constants' => true,
```

instead and running

```
$ make cs
```

?

theofidry:
There is still cases where it's necessary, for example in the auto-review test suite or for test cases covering multiple classes. So we cannot rely on this PHP-CS-Fixer setting since it would remove all of them

PR #1061, opened 2020-02-19
Use `$this` instead of `self` for assertions

theofidry:
I don't feel strongly about either conventions but given that we use `$this->assertX()` everywhere in the codebase, I don't think we should introduce `self::assertX()` in a few places: let's keep only one convention and if we want to change it let's change it everywhere.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -141,6 +141,9 @@ return Config::create()
             'case' => 'snake',
             'style' => 'prefix',
         ],
+        'php_unit_test_case_static_method_calls' => [
+            'call_type' => 'this',
+        ],
         'self_static_accessor' => true,
         'single_line_throw' => false,
         'static_lambda' => true,
--- a/tests/phpunit/Configuration/Schema/SchemaConfigurationFileLoaderTest.php
+++ b/tests/phpunit/Configuration/Schema/SchemaConfigurationFileLoaderTest.php
@@ -105,7 +105,7 @@ final class SchemaConfigurationFileLoaderTest extends TestCase
     private static function createRawConfigWithPathArgument(string $path): Constraint
     {
         return new Callback(static function (SchemaConfigurationFile $config) use ($path) {
-            $this->assertSame($path, $config->getPath());
+            self::assertSame($path, $config->getPath());
 
             return true;
         });
--- a/tests/phpunit/Configuration/Schema/SchemaConfigurationFileLoaderTest.php
+++ b/tests/phpunit/Configuration/Schema/SchemaConfigurationFileLoaderTest.php
@@ -105,7 +105,7 @@ final class SchemaConfigurationFileLoaderTest extends TestCase
     private static function createRawConfigWithPathArgument(string $path): Constraint
     {
         return new Callback(static function (SchemaConfigurationFile $config) use ($path) {
-            self::assertSame($path, $config->getPath());
+            $this->assertSame($path, $config->getPath());
 
             return true;
         });
--- a/tests/phpunit/Environment/BuildContextTest.php
+++ b/tests/phpunit/Environment/BuildContextTest.php
@@ -53,7 +53,7 @@ final class BuildContextTest extends TestCase
             $branch
         );
 
-        self::assertSame($repositorySlug, $buildContext->repositorySlug());
-        self::assertSame($branch, $buildContext->branch());
+        $this->assertSame($repositorySlug, $buildContext->repositorySlug());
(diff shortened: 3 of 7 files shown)


localheinz:
@theofidry 

What do you think about applying this patch

```diff
diff --git a/.php_cs.dist b/.php_cs.dist
index 8959771c..2ed07596 100644
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -141,6 +141,9 @@ return Config::create()
             'case' => 'snake',
             'style' => 'prefix',
         ],
+        'php_unit_test_case_static_method_calls' => [
+            'call_type' => 'this',
+        ],
         'self_static_accessor' => true,
         'single_line_throw' => false,
         'static_lambda' => true,
```

instead and running

```
$ make cs
```

?

theofidry:
Did think to check that one :) Thanks

maks-rafalko:
I'm ok with this decision, but why not `self`?

Methods are static, and by default PHPStan complains when they are called in an object context `$this->`

```
public static function assertSame($expected, $actual, string $message = ''): void
```

Any arguments?

sanmai:
I think this is the same argument why we use spaces vs. tabs: because that's what majority in the PHP ecosystem use. Resp. if you program in Go, you're stupid if you're not using tabs because that's the way it is in the Go land. So there goes `$this->assert...`, even the PHPUnit manual itself uses this style.

And there are some extra benefits, e.g. if you happen to need to specify your own `assertSame` as an instance method, you can certainly do that.

localheinz:
@sanmai 

> And there are some extra benefits, e.g. if you happen to need to specify your own assertSame as an instance method, you can certainly do that.

A custom assertion could also be implemented as a `static` method. 

Personally, I tend to use `static` access for methods which

- are `protected` or `private`
- do not need access to instance fields or methods

sanmai:
It can, but if you _need_ it to be an instance method for all the right reasons, and you have `self::` everywhere, tough luck. We dodge this bullet completely when we follow the traditional approach of `$this->assert...()`.

localheinz:
> It can, but if you need it to be an instance method for all the right reasons, and you have `self::` everywhere, tough luck. We dodge this bullet completely when we follow the traditional approach of `$this->assert...()`.

Can you show an example where you need an assertion to be an instance method?

sanmai:
I'm pretty sure I can, but this example will miss the major points of this dispute: consistency and lack of surprise. If there's a reason to continue using `$this->...`, that is it.

maks-rafalko:
> what majority in the PHP ecosystem use

Unfortunately, majority of PHP ecosystem blindly copied this approach from PHPUnit docs (and me as well).

```
 ------ ----------------------------------------------------------------------- 
  Line   tests/phpunit/Config/InitialYamlConfigurationTest.php                  
 ------ ----------------------------------------------------------------------- 
  120    Dynamic call to static method PHPUnit\Framework\Assert::assertSame().  
 ------ ----------------------------------------------------------------------- 
```

This is what you get with a `phpstan-strict-rules` config.

Let's see what the [documentation says](https://phpunit.readthedocs.io/en/8.5/assertions.html):

> A common question, especially from developers new to PHPUnit, is whether using $this->assertTrue() or self::assertTrue(), for instance, is “the right way” to invoke an assertion. The short answer is: there is no right way. And there is no wrong way, either. It is a matter of personal preference.

> For most people it just “feels right” to use $this->assertTrue() because the test method is invoked on a test object. The fact that the assertion methods are declared static allows for (re)using them outside the scope of a test object. Lastly, the global function wrappers allow developers to type less characters (assertTrue() instead of $this->assertTrue() or self::assertTrue()).

Let's leave the correctnes of these statements to authors, I'm not the person who can judge them.

But IMO Static Analyzers - this is what we need to listen to. If there is a rule [DynamicCallOnStaticMethodsRule](https://github.com/phpstan/phpstan-strict-rules/blob/4c0f97818480edca5d8c48e64d17c1a2de3acd6c/src/Rules/StrictCalls/DynamicCallOnStaticMethodsRule.php) and it's violated, PHPUnit assertions should not be an exclusion for it.

---

I don't say we need to change anything here, just wanted to open a discussion. But personally, I will use static methods in other repos, as I did for extracted adapters. Otherwise you will end up with *ignoring violations of PHPStan*.

localheinz:
@maks-rafalko 

Not a problem with https://github.com/phpstan/phpstan-phpunit!

😉

maks-rafalko:
Is this extension "hides" the error above? :)

PR #1065, opened 2020-02-21
Remove only covered

### LINKED ISSUE #1064: [RFC] Remove only-covered option, make it the default
Currently the [`--only-covered` option](https://infection.github.io/guide/command-line-options.html#only-covered) has the following influence:

- before parsing a file, if the file does not have any test, we skip that file
- given a mutation, it may or may not be covered
- for each mutation we create a mutant and its process
- when executing the mutant processes, if the process is not covered by tests, we skip it
- we log the not covered mutants in the various logs

### Issue °1: semantics

Right now we are dividing "escaped mutants" in two categories: 

- "escaped mutants": that is covered by tests but was not killed by them
- "not covered mutants": that is not covered by tests hence cannot be killed by them

But ultimately, both are just "escaped mutants". The distinction is only interesting if you either care about performance (because you can treat non-covered mutants differently) or if you are interested in code coverage.

But I think if you are interested in improving code coverage, mutation testing is not the solution: it is way too expansive for that and will do a poor job at it (you'll only see some bits from the mutated code but it's not like we mutate all code either). So I don't think it provides any value or even makes things more confusing to expose that information in our loggers.

### Issue °2: performance

All the following operations:

- parsing a file
- creating a mutant process for a mutation
- evaluating a process

Are expansive operations. The last one we already avoid it for non-covered mutations, but that still leaves both previous points still.

### Proposal

At first I was thinking `--only-covered` should be the default, but now I'm more leaning towards this option is not needed at all (and making its behaviour the default). When it was in Humbug, I remember asking Padraìc why it was there, and his answer was along the lines of "this is more out of curiosity" - Humbug was in the very early stage and so was mutation testing in PHP, so it makes sense. Infection however is much more mature and used.

theofidry:
Closes #1064

This PR proposes to remove `--only-covered` and all its downstream implications. The reasons are being laid out in the RFC so the goal of this PR is simply to show the impact on the code.

A few additional notes:

- nothing prevents us to have a "covered MSI" score. But we are better relying on the code coverage score for calculating that metric rather than mutations. In other words, MSI 76% of 43% (code coverage) would result in 32.68%. To be done in another PR though
- I did not change anything regarding the usage of `--min-msi` and `--min-covered-msi`. IMO it makes sense to keep them if we introduce back the "covered MSI score"


sanmai:
Also will require removing plains MSI reporting, since it effectively becomes covered code MSI: https://github.com/infection/infection/blob/3794e423f558848cb328469d11f383c40d9417ad/src/Console/ConsoleOutput.php#L80

theofidry:
We could keep it. We have:

A) Code coverage score
B) MSI `nbr of mutants killed / total nbr of mutants`

So we can have:

1. just the above completely discarding the uncovered mutations
2. The above but including the code coverage score. In which case we have:
  - (A) a strict mutation score (e.g. 76%)
  - (B) a mutation score made relative to the coverage (e.g. 76% of 43% i.e. 32.68%)

The only problematic would be how to name those two. Until now we had (_sort of_):

(A): Covered MSI
(B): MSI

We can instead have:

(A): MSI (mutation score indicator)
(B): mutation coverage?

Open to suggestions. I'm also checking with the other MT tools to see if there is a sort of convention yet

theofidry:
Hm will go with removing it for now: if we introduce it back it will be done in a quite different way so it doesn't make sense to try to keep that code around in-between. Also since the way we are going to provide that value is likely different, I think a dedicated PR to introduce it back if we wish to is more appropriate

src/Process/Runner/TestRunConstraintChecker.php
```diff
@@ -44,10 +44,16 @@ final class TestRunConstraintChecker
 {
     public const MSI_FAILURE = 'min-msi';
 
+    /**
+     * @deprecated
+     */
     public const COVERED_MSI_FAILURE = 'min-covered-msi';
```
duncan3dc: Just dropping a use case here in case you're not aware of it. We require that all mutations are prevented from escaping. We use `min-covered-msi=100` so that if code isn't covered at all then the build can still be green.

We then report back the code coverage stats in pull requests so that reviewers can manually decide if the code coverage is high enough (we expect 100% but allow some exceptions in certain circumstances).

I _think_ if the `min-covered-msi` option was removed then we'd have to remove the 100% limit, and therefor rely on reviews to manually check if any mutants had escaped, and then in cases where coverage wasn't 100% to figure out whether it's acceptable they escaped or not, this would be much worse than our process today.

theofidry: That ability should not be removed don't worry! It may however change a bit (e.g. a change of option name) depending on the decision taken in #1064

theofidry:
Closing for now since it is unlikely we will remove it right away. This PR served its purpose of showing a bit what implications it would have code-wise, let's carry on the discussion in #1064

PR #1068, opened 2020-02-22
Fix: Back up globals in tests that modify environment variables

localheinz:
This PR

* [x] backs up globals in tests that modify environment variables using the `@backupGlobals` annotation

💁‍♂ For reference, see https://phpunit.readthedocs.io/en/8.2/annotations.html#backupglobals.


theofidry:
@localheinz what happened there?

theofidry:
Wouldn't PHPUnit `backupGlobals` to the job?

localheinz:
@theofidry 

🤦‍♂ 

Not sure why I never even thought about that.

Thank you!

Shall I adjust?

theofidry:
@maks-rafalko if that's the case I believe having an alias would be enough:

```php
function unset_env_var(string $value): void {
  putenv($value.'=');
}
```

localheinz:
Tests are failing.

Not sure if and how that is related to using `@backupGlobals enabled`.

maks-rafalko:
I've restarted build on `master`: pass https://travis-ci.org/infection/infection/builds/654177204
I've restarted build on this PR: fail https://travis-ci.org/infection/infection/builds/654280987?utm_source=github_status&utm_medium=notification

so probably it's related, but I didn't dig to understand why and how

sanmai:
Apparently it doesn't play well with Travis (?) 
```

.............................................sh: 4: php: not found
F

There was 1 failure:
1) Infection\Tests\Mutator\Boolean\TrueValueTest::test_it_can_mutate with data set "It mutates inside function call when function is a variable" ('<?php\n\n$a = 'foo';\n$a(true);', '<?php\n\n$a = 'foo';\n$a(false);')
Mutator TrueValue produces invalid code
Failed asserting that 127 is identical to 0.
```

theofidry:
Closed by #1118

PR #1069, opened 2020-02-22
Fix: Keep rules and rule sets sorted in .php_cs.dist

localheinz:
This PR

* [x] keeps rules and rule sets sorted in `.php_cs.dist`

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -67,12 +67,12 @@ $finder = Finder::create()
 return Config::create()
     ->setRiskyAllowed(true)
     ->setRules([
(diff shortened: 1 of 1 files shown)


theofidry:
how does it plays out with the profiles? Doesn't the order have an importance?

theofidry:
Nevermind you kept them all at the top

localheinz:
Thank you, @maks-rafalko and @theofidry!

> how does it plays out with the profiles? Doesn't the order have an importance?

The order does not matter, as far as I can tell.

PR #1071, opened 2020-02-23
Switch to a name convention based event dispatcher

theofidry:
This PR removes the need for `EventDispatcher::getSubscribedEvents()`. The downside is, right now, relies on reflection in an inefficient way. I have a few ways to optimize it but it appears that since we are not relying heavily on events anyway, even the current implementation is barely noticeable on Infection itself (+1s).

I honestly prefer the style proposed in the PR, but still not quite satisfied with the potential performance impact. So I'll leave it up to you to tell what you think of it.

A possible improvement would be, IMO, to force events to implement an `Event` interface and from that we can build a static map (on composer dump-autoload for example and which can have 
a fallback just in case). In other words pretty much auto-generating `EventDispatcher::getSubscribedEvents()` (and completely hiding it)


src/Event/EventDispatcher/SyncEventDispatcher.php
```diff
@@ -37,38 +37,33 @@
...
-     * @var callable[][]
+     * @var EventSubscriber[]
      */
-    private $listeners = [];
+    private $subscriber = [];
 
     public function dispatch(object $event): void
     {
-        $name = get_class($event);
+        $subscription = 'on' . substr((string) strrchr(get_class($event), '\\'), 1);
```
sanmai: Could event itself know this method name? And memoize it.

src/Event/Subscriber/CiInitialTestsConsoleLoggerSubscriber.php
```diff
@@ -55,13 +55,6 @@ public function __construct(OutputInterface $output, TestFrameworkAdapter $testF
         $this->testFrameworkAdapter = $testFrameworkAdapter;
     }
 
-    public function getSubscribedEvents(): array
-    {
-        return [
-            InitialTestSuiteWasStarted::class => [$this, 'onInitialTestSuiteWasStarted'],
```
sanmai: I agree that this is not nice at all.

tests/phpunit/AutoReview/Event/SubscriberProvider.php
```diff
@@ -130,4 +113,24 @@ public static function subscriberSubscriptionMethodsProvider(): Generator
...
+                return strpos($method->getName(), 'on') === 0;
+            }
+        );
+
+        return array_map(
+            static function (ReflectionMethod $method): string {
+                return $method->getName();
+            },
+            $methods
+        );
```
sanmai: How about if:
- Each event will declare in a constant which interface must subscribers implement.
- The loop in `dispatch()` above for each event will grab its expected subscriber interface, and check every subscriber against this interface.
- Subscribers would have a single entry point for events, where they would shuffle incoming events into handlers according to their type.
- Thus it seems we could avoid most introspection.

Another approach:
- Since our `on...` methods are typed, here you can make a list not only of methods, but also of their argument's type. Say, a map with argument type as a key. So this will be the only place to use reflection. Up above in `dispatch()` it should be possible to pattern-match incoming events into handlers.
- This method still requires reflection, but avoids adding a bunch of extra interfaces.

src/Event/EventDispatcher/SyncEventDispatcher.php
theofidry: That could be doable but I think it would be better done in an abstract class `Event` which all events should extend then

tests/phpunit/AutoReview/Event/SubscriberProvider.php
theofidry: I think I would prefer going with generating a map. But that brings the question on how you keep that map up to date (in a composer script when dumping the autoload for example?)

sanmai: Alright, here you go #1084

theofidry: Will try to compare performance wise; I otherwise I think both approach are fine it's just a question of flavour so up to @infection/core team to choose.

But otherwise I was more thinking of generating a static map, rather than doing so at run-time

sanmai: Performance wise it won't make much sense this way or another because this is a one-off operation.

sanmai: But if you want to look at this this way, consider that in #1084 we run Reflection only once per registration, where here you want to run it slightly simpler, but still more complex operation than just a map lookup, on each `dispatch()` call.

maks-rafalko:
I'm sorry I will be skeptical here, but I'm starting losing understanding why we are sometimes changing the things that already perfectly work.

> This PR removes the need for `EventDispatcher::getSubscribedEvents()`. The downside is, right now, relies on reflection in an inefficient way.

First of all, was the `getSubscribedEvents()` method an issue? IMO, no. And here we have a downside (reflection), changing already working code.

We had such methods in subscribers for example in Symfony for ages. I mean, we got used to it, it just works...

Now, there is a convention that every subscriber should have a method that starts with `on***`. Well, not a big deal, but still.

---

If you guys think it worths the change, ok. Regarding 2 proposed solutions, I like https://github.com/infection/infection/pull/1084 more because of issues of this PR described in https://github.com/infection/infection/pull/1071#discussion_r385702072

Having a static map of event => subscribed classes and rely on composer here for me sounds like a not needed complexity without any benefit comparing to what we have now in `master`

theofidry:
>We had such methods in subscribers for example in Symfony for ages. I mean, we got used to it, it just works...
>
>Now, there is a convention that every subscriber should have a method that starts with on***. Well, not a big deal, but still.

The issue though is that you rely on string based callables which are not properly understood by an IDE or SA. As for the naming convention: it's a nitpick, but every now on then you will end up with `onFoo(Bar $event)` or something like that - a name that no longer match the event class. Hence the auto-review tests to enforce it. This PR just takes advantage of that convention to remove the need to declare the subscriptions

maks-rafalko:
> which are not properly understood by an IDE or SA

PHPStan understands it: https://phpstan.org/r/f55623f4-5154-4163-a474-19c2a7e97466

But I doubt that it understands such code:

```
$subscription = 'on' . substr((string) strrchr(get_class($event), '\\'), 1);
```

theofidry:
Neither phpStorm or its PHP Inspection plugin does though :(

>But I doubt that it understands such code:

No but at least when you refactor a method, you don't have to look for its string usages as well.

But overall, this PR or #1084 goal is the same: leverage an existing convention (be it a naming convention or typehint one) to remove a additional configuration step

maks-rafalko:
Closing as https://github.com/infection/infection/pull/1084 is merged

PR #1073, opened 2020-02-24
Introduce a SingletonContainer for tests

theofidry:
We have some stateless services used in our test codebase. Since they are used repeatedly, we sometimes try to "cache" them to avoid instantiating them over and over. However this is:

- verbose: requires a static property, static getter (for the right return type)
- error prone & inconsistent: sometimes we use `setUpBeforeClass()` sometimes a getter
- not that efficient: if you have two test cases implementing that strategy you are still instantiating that service 2 times

This PR introduces a `SingletonContainer` to replace `Container` within the tests as well as a few services used for the tests.


tests/phpunit/AutoReview/IntegrationGroup/IntegrationGroupProvider.php
```diff
@@ -88,7 +74,9 @@ public static function provideIoTestCaseTuple(): Generator
         }
 
         self::$ioTestCaseClassesTuple = array_values(array_filter(array_map(
-            [self::class, 'checkIoOperations'],
+            static function (string $className): ?array {
```
theofidry: Not necessary per se but I was fed up with the unused method warning...

tests/phpunit/SingletonContainer.php
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * @var PHPDocParser|null
+     */
+    private static $phpDocParser;
+
+    public static function getContainer(): Container
+    {
+        if (self::$container === null) {
+            self::$container = Container::create();
```
maks-rafalko: can't we use `??` here as well?

theofidry: Done in a few more places as well

sanmai:
Is there a fixer for `??` operator?  I recall PHP-CS-Fixer replacing it all over the place.

theofidry:
For some reasons the PHP-CS-Fixer is updating the e2e tests... hence the failure of PrettyCI. Not gonna hold the PR because of it but we need to take a look

maks-rafalko:
Looks like PrettyCI ignores `exclude` section:

https://github.com/infection/infection/blob/1420efee99166f32ba09649a4b35cc207229c957/.php_cs.dist#L49-L58

ping @mnapoli

maks-rafalko:
and it tries to add not our header:

```
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
```

:|

theofidry:
@sanmai I think it's already applied:

```
ternary_to_null_coalescing [@PHP70Migration, @PHP71Migration, @PHP73Migration]

```

maks-rafalko:
> and it tries to add not our header:

https://twitter.com/matthieunapoli/status/1232637552282292224?s=20

mnapoli:
Hi, since yesterday I pushed some updates, let me know if you still have issues.

PR #1074, opened 2020-02-24
Accounting For Codeception Cest Tests In JUnit File.

Fenikkusu:
This PR:

- [x] Adds new feature ...
- [x] Covered by tests

Due the way Codeception 'Cests' or 'Codeception Tests' work, code coverage xml generated by the cest files do not exactly match the format that Infection requires for the files to work. Cest files, by their nature, have no extension. When PHP-Coverage runs, it expectedly checks that the file extends TestCase to properly generate the JUnit file. Since Cest tests do not, they lines get added with a single `:` instead of two. This causes a FQNS errors when infection tries to run against the file. This is easily fixed in infection by search for the single `:` instead of the double.

Note: Still testing this change internally. Just getting this prepped for the moment.

### DIFF
--- a/src/TestFramework/Coverage/XmlReport/PhpUnitXmlCoverageFactory.php
+++ b/src/TestFramework/Coverage/XmlReport/PhpUnitXmlCoverageFactory.php
@@ -120,7 +120,7 @@ class PhpUnitXmlCoverageFactory
         CoverageLineData $test,
         TestFileDataProvider $testFileDataProvider
     ): void {
-        $class = explode('::', $test->testMethod, 2)[0];
+        $class = explode(':', $test->testMethod, 2)[0];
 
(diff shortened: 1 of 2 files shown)


Fenikkusu:
I believe everything is in order here. We've been running off a patched version with this change and no issues have arisen thus far.

theofidry:
Will wait in @maks-rafalko review since he is more familiar with Codeception than me

Fenikkusu:
@maks-rafalko , Yes, I'm successfully using Infection with Codeception in 3 projects. The only issue we've had was the issue this MR resolved. Outside of that, everything has been running well.

I'm not worried about porting it to 0.15. Don't really have the time to right now. We are just going to create a docker image that builds off master.

PR #1024, opened 2020-01-27
Tweak MemoryLimiter

theofidry:
- Move `MemoryLimiter` from `Resource\Limiter` to `Resource\Memory` (alternatively could be moved up to `Resource`?)
- Cosmetic changes to `MemoryLimiter`
- Make `MemoryLimiterTest` leverage `FilesystemTestCase` and do some cosmetic changes on it

PR #1076, opened 2020-02-26
MemoryLimiterTest needs an update

sanmai:
This test is broken currently, let's see if CI would notice.

- Previous change to this test: #1024 
- I reviewed and didn't see any obvious bugs, hoping CI would catch anything not-so-obvious.
- Yet this test fail when I run it locally:
```
1) Infection\Tests\Resource\Memory\MemoryLimiterTest::test_it_applies_memory_limit_if_possible
Expectation failed for method name is "appendToFile" when invoked 1 time(s).
Method was expected to be called 1 times, actually called 0 times.
```
- Aaand CI sees nothing wrong [because it skips the test](https://travis-ci.org/infection/infection/jobs/655174307#L579).

- [x] This PR should make the test run even during CI
- [x] Test fixed

### DIFF
--- a/tests/phpunit/Resource/Memory/MemoryLimiterEnvironmentTest.php
+++ b/tests/phpunit/Resource/Memory/MemoryLimiterEnvironmentTest.php
@@ -1,0 +1,75 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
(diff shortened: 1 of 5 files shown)


maks-rafalko:
As we agreed, can we increase MSI for new files?

https://travis-ci.org/infection/infection/jobs/655204240#L883

sanmai:
Some builds are failing because of escaping mutations inside MemoryLimiterEnvironment. But that class is hard to guard because it requires running tests with and without phpdbg to test thoroughly.

There are other failures, but I didn't see they're related.

sanmai:
I updated `MemoryLimiterTest` to reflect that it should not require specific environment properties now.

maks-rafalko:
Then let's merge it.

Thank you @sanmai

sanmai:
Thank you @maks-rafalko

tests/phpunit/Resource/Memory/MemoryLimiterEnvironmentTest.php
```diff
@@ -0,0 +1,75 @@
...
+
+use Composer\XdebugHandler\XdebugHandler;
+use Infection\Resource\Memory\MemoryLimiterEnvironment;
+use const PHP_SAPI;
+use PHPUnit\Framework\TestCase;
+use function Safe\ini_get;
+
+/**
+ * @group integration Requires some I/O operations
+ * @covers \Infection\Resource\Memory\MemoryLimiterEnvironment
```
theofidry: The `@covers` should not be necessary here

tests/phpunit/Resource/Memory/MemoryLimiterEnvironmentTest.php
```diff
@@ -0,0 +1,75 @@
...
+            $this->markTestSkipped('This test requires PHPDBG');
+        }
+
+        $environment = new MemoryLimiterEnvironment();
+        $this->assertTrue($environment->isUsingSystemIni());
+    }
+
+    public function test_it_detects_xdebug_handler(): void
+    {
+        if (PHP_SAPI === 'phpdbg') {
```
theofidry: Since this is used in a couple of other places in the tests as well, what about moving it to a static helper or trait to do `$this->requiresPhpDbg()`?

tests/phpunit/Resource/Memory/MemoryLimiterEnvironmentTest.php
sanmai: TBH I added this because of a habit. Are we arrange these things to happen magically, or just don't care? I saw you removing this.

theofidry: I would like an automated check or fixer for it... but there isn't one and I'm not really willing to invest in it...

PR #1077, opened 2020-02-26
Make check_trailing_whitespaces.sh ignore untracked files

sanmai:
Should be a no-brainer.

Verify:
```
touch devTools/foobar
find devTools/ -type f -not -exec git ls-files --error-unmatch {} 2>/dev/null \;
rm devTools/foobar
```
Should not have `devTools/foobar` in the output.


theofidry:
Would also be lovely to have some sort of caching like PHP-CS-Fixer, its quite slow on my machine :(

sanmai:
From the first glance it should be possible to parallelize this operation. Will try another day.

PR #1078, opened 2020-02-26
Introduce MutantExecutionResult

theofidry:
As of now, all our reporting (loggers) are dependent on `MutantProcess` which exposes a lot of mutable state. It also has the downside of requiring a `MutantProcess` for each mutation to work.

I the future, I would like avoid to create the `Mutant` and `MutantProcess` (which are not free) if possible, which does not necessarily exclude the mutation from being reported. 

For this reason, I propose to decouple the reporting from `MutantProcess` by introducing `MutantExecutionResult`. It can be created from a `MutantProcess` for convenience, but it would be easy to create it only from `Mutation` information as well.


src/Mutant/MutantExecutionResult.php
```diff
@@ -0,0 +1,127 @@
...
+            $process->getMutant()->getDiff(),
+            $process->getMutant()->getMutation()->getMutatorName(),
+            $process->getOriginalFilePath(),
+            $process->getOriginalStartingLine(),
+        );
+    }
+
+    public function __construct(
+        string $processCommandLine,
+        string $processOutput,
```
sanmai: Shouldn't these things be optional? Say, for uncovered mutations.

theofidry: I don't know if it should or be `''` yet: I literally just tried to introduce the object and some usages but did not have a look at the tests yet

sanmai: I mean, there are some arguments which are absolutely required for a mutation result, and some that are somewhat more optional. I understood that this is WIP.

theofidry:
That would work but it would result in the same issue being `MutantProcess` needs to accommodate for two usages: one being configured & evaluated as a process and the other being providing the necessary information for reporting. I was instead thinking of really splitting both usages which would allow to drastically simplify `MutantProcess`

theofidry:
@sanmai I'll go with that for now and simplify `MutantProcess`. If we want to, we can always try to  make `MutantExecutionResult` an interface instead and get `MutantProcess` implement it later see if that is better

PR #1080, opened 2020-02-26
Provide a more user-friendly error when the schema path is invalid

theofidry:


tests/phpunit/TestFramework/PhpUnit/Config/XmlConfigurationManipulatorTest.php
```diff
@@ -504,6 +508,46 @@ public function test_it_consider_as_valid_a_PHPUnit_XML_configuration_without_XS
...
+    foo="bar"
+>
+    <invalid></invalid>
+</phpunit>
+XML
+        );
+
+        set_error_handler(
+            static function (int $type, string $message, string $file, string $line): void {
+                // Silence!
```
theofidry: This is necessary to get around the PHPUnit error handler otherwise whenever `validateSchema()` fails, it will fail there instead of continuing & collecting the errors as we expect it to do

PR #1081, opened 2020-02-27
Make ParallelProcessRunner iterable-type agnostic

sanmai:
This PR:

- [x] Make ParallelProcessRunner iterable-type agnostic. It can now consume not only arrays, but all kinds of iterators/generators. Also it is kind of simpler now.
- [x] Covered by tests - tests weren't changed

### DIFF
--- a/src/Process/Runner/Parallel/ParallelProcessRunner.php
+++ b/src/Process/Runner/Parallel/ParallelProcessRunner.php
@@ -42,7 +42,6 @@ use Infection\Process\MutantProcess;
 use Symfony\Component\Process\Exception\LogicException;
 use Symfony\Component\Process\Exception\ProcessTimedOutException;
 use Symfony\Component\Process\Exception\RuntimeException;
-use Webmozart\Assert\Assert;
 
 /**
  * @internal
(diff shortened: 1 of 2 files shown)


sanmai:
Looks like I made the gods of CI happy from the first try. That's a truly pleasant surprise!

sanmai:
Thank you @maks-rafalko

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -75,57 +74,55 @@ public function __construct(EventDispatcher $eventDispatcher)
      * @throws RuntimeException
      * @throws LogicException
      */
-    public function run(array $processes, int $threadCount, int $poll = 1000): void
+    public function run(iterable $processes, int $threadCount, int $poll = 1000): void
```
theofidry: any reason why we have `iterable` here? I'm not too fond of it because it's really a wide and tricky type: you can fit arrays, traversables, generators (rewindable or not)

sanmai: Yes, the reason is in the making. It is a great type because it fits arrays, unlike generators, which makes a breeze mocking methods typically returning a generator.

theofidry: But why do we need to expand the type if we can use only arrays which is already easy to mock?

PR #1067, opened 2020-02-22
Enhancement: Implement GitHubActionsResolver

localheinz:
This PR

* [x] implements a `GitHubActionsResolver`

Related to #939.

❗️ Blocked by #1068.
💁‍♂ Also see https://github.com/infection/site/pull/155.

PR #1082, opened 2020-02-27
Concurrent mutation generator

sanmai:
- [x] Doc PR https://github.com/infection/site/pull/157

This PR has a goal of solving these two problems:
- Currently Infection spends around 20-30 seconds at the start, or even more depending on where it runs, just to parse relevant files and assemble a list of mutations. Yet it spends a lot more time just waiting for a mutator process to finish running. And it spends [even more time checking for uncovered mutations](https://github.com/infection/infection/issues/1064#issuecomment-589581680), one reason why we have #1064 and #1065.
- At the same time Infection will buffer in an array all files and all mutations. This problem isn't the root problem for #705 because we have other places where we buffer a lot of things, but this PR is one step towards solving problems like #750.

### How's it done?
Can we can spread these 20 seconds while Infection waits for mutated tests to finish? Yes, we can, and that's what this PR is trying to archive. 

Here we add an additional "streaming" mode of operation, when no progress is expected to be reported, which is when running with `--no-progress` option, to be used primarily during CI. Instead of counting files and mutations we stream them directly through a chain of generator functions and some iterators right as mutation testing happens.

For example, with this PR it takes around same 50-60 seconds *less* time to run Infection on itself. Where before I almost never could make Infection finish for faster that a minute and a half because of that delay, now I can make it finish easy under one minute.

As a side effect, with these changes it takes about same time to run Infection both considering and ignoring uncovered files:
```
vendor/bin/phpunit --coverage-xml=build/logs/coverage-xml --log-junit=build/logs/junit.xml --stop-on-failure
bin/infection --skip-initial-tests --coverage=build/logs -j32 --only-covered --no-progress
bin/infection --skip-initial-tests --coverage=build/logs -j32 --no-progress
```
Otherwise put, we no longer have one of two reasons for #1064. This isn't a performance issue any more.

Surely, nothing fancy comes free. Applying this streaming approach we're can no longer know exact number of mutations we're testing against before the fact (when running with `--no-progress`), and therefore we can no longer show how much work is ahead. But since `--no-progress` implies that end result is more important, it should not be a problem. Live mutation report becomes similar to the following, with just the current count of mutations:
```
M..M.M.......M............ME.MM.....M..E........M.   ( 1900)
..M..M.......M........M..M.M......M.........M...M.   ( 1950)
...M....................M.....M.....M..M......M...   ( 2000)
MMMMM........M....................MMMMMMMMMMMMM.MM   ( 2050)
MM.MMM..M.M.M.MM.MMMMMMMMMMMMMMMMMMMMMM....MMMMMMM   ( 2100)
MMMMMMMMMMMMMMMMMMMM

2416 mutations were generated:
    1585 mutants were killed
     310 mutants were not covered by tests
     514 covered mutants were not detected
       4 errors were encountered
       3 time outs were encountered

Metrics:
         Mutation Score Indicator (MSI): 65%
         Mutation Code Coverage: 87%
         Covered Code MSI: 75%

Dashboard report has not been sent: Build context could not be resolved.

Time: 58s. Memory: 192.02MB
```

Memory usage is about _twice as less_ when using  `--no-progress`. As this PR unlocks for us possibility to clean unneeded parts of our pipeline as we go, we will have a chance to have a more definite solution to the memory problem from #705 later.

### Try it

A picture's worth a thousand words. Check out this branch, `composer install`, then:
``` 
vendor/bin/phpunit --coverage-xml=build/logs/coverage-xml --log-junit=build/logs/junit.xml --stop-on-failure
bin/infection --skip-initial-tests --coverage=build/logs -j8 
bin/infection --skip-initial-tests --coverage=build/logs -j8 --no-progress
```
Note the time and memory usage difference. Repeat the same commands with master. 

Builds on Travis are not the fastest, and with a low `-jN`, but compare:
- Recent build on master. [Time: 6m 53s. Memory: 122.01MB](https://travis-ci.org/infection/infection/jobs/655935255#L1005)
- A build on this branch. [Time: 5m 49s. Memory: 64.00MB](https://travis-ci.org/infection/infection/jobs/656284463#L1024)

[Blackfire Profile](https://blackfire.io/profiles/compare/9218c7e2-8e38-4763-8ca5-86a470cd427a/graph) courtesy Theo seemingly confirms these observations.

### Concurrency is not parallelism 
We're not parallelising the whole thing yet, unfortunately. This is kinda hard to do in PHP, and generally hard to do. The whole process is entirely sequential as it was before, and the only things running in parallel are the mutant processes. But we make some effort to ensure we're not waiting for processes to finish doing absolutely nothing.

While we're quite far from parallelising the whole thing, we're a making a step towards solving the problem from #705. For one, because we no longer buffer a lot of heavy stuff in number of places. We still do, but no longer in these "hot" places. For example, as a next step, we can stream reports of escaped mutations into a file, removing the need to keep them in memory until the end of the program.

---
Please note that comments below may not be relevant to the current version of this text.


src/Process/Runner/MutationTestingRunner.php
```diff
@@ -74,26 +74,34 @@ public function __construct(
     /**
      * @param Mutation[] $mutations
      */
-    public function run(array $mutations, int $threadCount, string $testFrameworkExtraOptions): void
+    public function run(iterable $mutations, int $threadCount, string $testFrameworkExtraOptions): void
     {
-        $this->eventDispatcher->dispatch(new MutantsCreationWasStarted(count($mutations)));
+        $this->eventDispatcher->dispatch(new MutantsCreationWasStarted(0));
+        $this->eventDispatcher->dispatch(new MutationTestingWasStarted(0));
```
sanmai: I kept these event in place in sake of backward compatibility with output formatters

src/Mutation/FileMutationGenerator.php
```diff
@@ -73,15 +73,15 @@ public function __construct(
...
+     * @return iterable|Mutation[]
      */
     public function generate(
         SplFileInfo $fileInfo,
         bool $onlyCovered,
         LineCodeCoverage $codeCoverage,
         array $mutators,
         array $nodeIgnorers
-    ): array {
+    ): iterable {
```
sanmai: Here and below we use `iterable` typehint for generators, for one reason because it hides implementation details, and for another because it makes a lot easier to mock these methods: any array is an iterable, so one doesn't need to go through the hoops faking up generators.

src/Mutator/NodeMutationGenerator.php
```diff
@@ -86,27 +85,21 @@ public function __construct(
...
-        return array_reduce(
-            $this->mutators,
-            function (array $mutations, Mutator $mutator) use ($node): array {
-                return $this->generateForMutator($node, $mutator, $mutations);
-            },
-            []
-        );
+        yield from take($this->mutators)->map(function (Mutator $mutator) use ($node) {
+            yield from $this->generateForMutator($node, $mutator);
+        });
```
sanmai: Technically we can return the pipeline here as it is, as it corresponds to the `iterable` typehint, but I want to avoid exposing this implementation detail.

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -51,25 +48,25 @@
...
-     * @var MutantProcess[]
+     * @var ProcessBearer[]
      */
     private $currentProcesses = [];
 
-    public function __construct(EventDispatcher $eventDispatcher)
+    public function __construct(Closure $processHandler)
     {
-        $this->eventDispatcher = $eventDispatcher;
+        $this->processHandler = $processHandler;
```
sanmai: ParallelProcessRunner does need to know about EventDispatcher, as we may want to reuse ParallelProcessRunner in other places.

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -121,19 +118,11 @@ private function cleanFinished(): bool
...
 
-        if (!$mutant->isCoveredByTest()) {
-            $this->eventDispatcher->dispatch(new MutantProcessWasFinished($mutantProcess));
-
-            return false;
-        }
-
-        $mutantProcess->getProcess()->start();
-
-        $this->currentProcesses[] = $mutantProcess;
```
sanmai: A process manager doesn't have to care about types of processes.

sanmai:
AppVeyor build wants to install PHP 7.3+ package on PHP 7.2. Not sure if that's OK.

theofidry:
>Otherwise put, we no longer have one of two reasons for #1064. This isn't a performance issue any more.

Just to clarify, the performance issue is a side-effect that caused the investigation of #1064, but it is not the primary reason. I want to, in the future, make infection works fine without any coverage data. I'm not against providing more information (such as the coverage) or optimization strategies when coverage is available, but I don't want it to be necessary either. For that reason, having an "only covered" option just gets in the way and unnecessarily complicate things by among other things introducing this notion of "skipped mutants" instead of keeping "escaped mutants".

<hr/>

IMO for this PR: I thought a lot of that approach as well before. So here's a few pros/cons:

Pros:

- allows to leverage parallelism earlier for parsing, creating mutations, their mutants, mutant processes and evaluating those processes
- makes the API simpler in a sense: you have an input of files and it get streamed into being mutated, it is more straightforward

Cons:

- loose all possibility to log intermediate state (e.g. number of mutations or mutants)
- more confusing & riskier experience: if an error occurs in the process of parsing to evaluating the mutant process, for whatever reason, it will happen in the middle of other stuff which means possible in the middle of the run after a few seconds or minutes even. That can be a frustrating experience
- loose the possibility implementing any mutation optimization. Indeed there is quite a few possible optimizations given a set of mutations: detect redundant ones, have a "top down" hierarchy e.g. if you have mutationA and mutation1 & mutation2 which are sub-nodes of mutationA, a possible optimization is to skip mutation1 & mutation2 if mutationA is escaped. A trivial example would be removing a method body: if that mutation is escaped, then any sub-mutation will be escaped as well.

I had a few back and forth on this, and ultimately what I would like to do is rather a bit more in the middle of the road of what is currently proposed in this PR and what is done right now.

The current steps are:

1. parse files
2. collection mutations
3. create mutants
4. create mutant processes
5. evaluate mutant processes (parallel) 

This PR proposes to parallelise the whole thing, parsing a file being the start of a process.

An alternative solution would be:

1. 2.: executed in parallel, but retrieve all the mutations back in the main process. This part can be done with the current architecture without too much trouble and should not even affect the logging of this part

Merge 3, 4 & 5: for each of those the starting point is a mutation. It means we would still loose some logging information there e.g. the procress of mutant & mutant process creation, but IMO it's not critical information either so this could be shoved into a process to be executed in parallel

In other words, instead of parallelising the whole thing we could parallelise two distinct steps, allowing the collect all the mutations in the middle and still give the possibility to perform some optimization strategies on them

src/Process/Runner/MutationTestingRunner.php
```diff
@@ -74,26 +74,34 @@ public function __construct(
...
-            },
-            $mutations
-        );
+            return $process;
+        });
 
         $this->eventDispatcher->dispatch(new MutantsCreationWasFinished());
 
-        $this->eventDispatcher->dispatch(new MutationTestingWasStarted(count($processes)));
+        // We filter these here because it is beyond responsibilities of a process manager to care about types of processes
```
sanmai: Adding something like this above this line:
```
$processes = take($processes->toArray());
```
We can make the entire process as "sequential" as before. E.g. it'll buffer until it collects all mutations.

sanmai:
We're not parallelising the whole thing yet, unfortunately. This is kinda hard to do in PHP. The whole process is entirely sequential as it was before, and the only thing is running in parallel are the mutant processes. Otherwise put, there can't be a situation where something happens in the middle of other stuff.

Yes, we're sort of loosing possibility implementing any mutation optimization, but we're not. It is a [matter of a single line of code](https://github.com/infection/infection/pull/1082/files#r385044613) to make everything go step-by-step as before.

theofidry:
If that's the case why not just reducing this PR transformation at the `MutationTestingRunner` and leave everything upstream unchanged?

sanmai:
Because we're buffering results not only inside `MutationTestingRunner`, but also on previous stages. E.g. inside `MutationsCollectorVisitor` we normally have an array with all mutations.

theofidry:
If you do a `toArray()` though you nullify that buffering right?, so I'm not sure it's a win to do so

sanmai:
Yes, we'd nullify if we do. But we don't do that now.

I was saying what we can do a `toArray()` if we ever want to add some extra mutation optimizations you were mentioning.

sanmai:
I'm trying to run Infection over Psalm now, and it sort of runs:
```

SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS   (   50)
SSSSSTTTTTSSSSSSTTTTTSSTTTTTTTTTTTTSSSSSSSSSTTTTTT   (  100)
TTT.........T...........M.......SSSSSSSSSSSSSS....   (  150)
....................T....TT....TT............S.S..   (  200)
..S.TTTSSSSSSSSSTTTTTS..........SSTTTTT..STSSSSS.S   (  250)
SSSSSSSSSSSSSTTTTTTSSSTSSSSSSSSSSSSSSSSSSSSSSSSSST   (  300)
TSTTTTTTTTTTTSTTTTTTTTTTTTTSSSSSSSSSSSSSSSSSSSSSSS   (  350)
```

I'm not entirely sure it will finish, but it's not like before when you could wait and wait until Infection would generate all mutants. TBH I never had the patience to outwait it.

theofidry:
K I'll let the other express their opinions then.

For me the loss of mutation progress is a big downside and I would almost say deal-breaker, but I also think it doesn't invalidate the PR because it can still be done even with the above changes

sanmai:
Right, how about if I add a switch, which would toggle this new "streaming" approach? But by default Infection would show the progress as before. It won't be exactly same, but you would be able to see the total number of mutations.

theofidry:
even with the streaming approach, you can still see the mutation progress and count if you trigger the `iterator_to_array()` at the right place

sanmai:
We do track progress. Try it.

theofidry:
You no longer have the progress bar about parsing the files and displaying the number of mutations found, that is what I mean

sanmai:
True because they’re redundant. Once we parse a file and get a mutation, we start testing against it. And you’ll see a normal status for a mutation.

theofidry:
Sorry I don't get how to see the progress of the file parsing and total number of mutations is redundant...

sanmai:
Because there isn’t a separate “parsing files” stage anymore.

theofidry:
Yes I know, hence: 

>K I'll let the other express their opinions then.

>For me the loss of mutation progress is a big downside and I would almost say deal-breaker, but I also think it doesn't invalidate the PR because it can still be done even with the above changes

I think it is important to keep the progress of it, but again, that's my opinion and would like to hear the voice of the others on it

sanmai:
There’s a progress indicator going. Try it, really. A picture worth a thousand words.

theofidry:
I did, I don't see any:

Screen Shot 2020-02-27 at 14 03 19

sanmai:
Well, these `SSSSS....` isn't an indication of progress? And there you can see a number of mutations found, on the right.

By the way, judging by a lot of skipped mutations in the beginning, aren't you running it with [this line added](https://github.com/infection/infection/pull/1082/files#r385044613)?

theofidry:
If we take the one we have right now on master, we have:

```

    ____      ____          __  _

Infection - PHP Mutation Testing Framework dev-master@a66ff970e8578f4ae7eab1c2f076eb3e12c0c06f

Running initial test suite...

PHPUnit version: 8.3.5

 3579 [============================] 46 secs

Generate mutants...

Processing source code files: 334/334
Creating mutated files and processes: 2471/2471

SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS   (  50 / 2471)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS....   ( 100 / 2471)
................[...]
```

Your PR, from this report PoV, removes:

```
Generate mutants...

Processing source code files: 334/334
Creating mutated files and processes: 2471/2471
```

I would however love to rather see something like:

```
Generating mutations...
Processing source code files: 334/334

SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS   (  50 / 2471)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS....   ( 100 / 2471)
```

i.e. keep the parsing progress and the total mutation count

>By the way, judging by a lot of skipped mutations in the beginning, aren't you running it with this line added?

Ah indeed I did not remove it for the screenshot

sanmai:
Please try without it. You will see there's no delay now where before were "Processing source code files" and "Creating mutated files and processes" counting their numbers.

theofidry:
Yes but you're loosing that total mutations count and as soon as you try to get it you will have this lag in-between again. And IMO, this total mutations count is as important for knowing the overall progress that accessing those mutations to manipulate them (for optimizations for example)

theofidry:
But as said: we are really debating about only half of the PR there which is a bit frustrating because I think there is half of it that has benefits and can go in without any debate

Fenikkusu:
Just my two cents: Rather than make the generation multi-threaded, wouldn't it be better (and maybe even easier) to instead cache the mutations? Right now, unless you run in debug mode, all mutations are deleted after the process finishes. I would think it better to leave the mutations, and then use a system similar to PHP-CS-Fixer where in infection tracks the last modified time of each file and only generates mutations is none exist or the file has changed since the last run.

I know PHP-CS-Fixer creates a 'cache' file that you can specify on running that stores the relative file path and the last modified time. When you run it again, it compares the last modified time to the cache and skips the file completely if they match.

This would become very beneficial in a CI environment where you can cache the initial results. For example, I have a project running infection that has 390 files and generates 5550 mutations currently. The process takes about 20 minutes to finish. If I were able to cache the generated mutations and only reanalyze the files that changed in the merge request, this would only take the full 20 minutes the first run and would likely shave several minutes for all additional runs.

theofidry:
@Fenikkusu those are not exclusive solutions though :) I would say however infection should almost always be used incrementally rather than a full run

sanmai:
@Fenikkusu if you could open a dedicated issue, so we can discuss this idea there?

Fenikkusu:
> @Fenikkusu how about if you could open a dedicated issue, so we can discuss the idea there? It is not to say it is bad or anything, just not to mix things up too much. Thank you.

Issue #1085 Created

sanmai:
@theofidry please see if it is less annoying now. Should work about as before unless you run it with `--no-progress`.

tests/phpunit/Console/OutputFormatter/DotFormatterTest.php
```diff
@@ -110,6 +114,58 @@ public function test_not_covered_correctly_in_console(): void
         $dot->advance($this->createMutantExecutionResultsOfType(MutantProcess::CODE_NOT_COVERED)[0], 10);
     }
 
+    public function test_it_prints_total_number_of_pending_mutations(): void
+    {
+        $totalMutations = (int) (self::DOTS_PER_ROW * 2 + self::DOTS_PER_ROW / 2);
```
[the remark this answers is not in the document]
sanmai: I want to make sure it doesn't span the entire last line, and it is silly to ask someone changing `DOTS_PER_ROW` to recalculate an arbitrary number somewhere else.

src/Process/Runner/MutationTestingRunner.php
```diff
@@ -58,45 +59,82 @@ final class MutationTestingRunner
...
+            $mutant = $this->mutantFactory->create($mutation);
 
-                $this->eventDispatcher->dispatch(new MutantWasCreated());
+            $process = $this->processBuilder->createProcessForMutant($mutant, $testFrameworkExtraOptions);
 
-                return $process;
-            },
-            $mutations
-        );
+            $this->eventDispatcher->dispatch(new MutantWasCreated());
```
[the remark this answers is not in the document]
sanmai: We can remove the entire event in this case, how do you think?

src/Process/Runner/MutationTestingRunner.php
```diff
@@ -35,19 +35,20 @@
 
 namespace Infection\Process\Runner;
 
-use function array_map;
-use function count;
```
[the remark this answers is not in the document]
sanmai: I have no idea why PHP-CS-Fixer won't add, and will remove these. I'm pretty sure I didn't edit these by hand.

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -120,22 +145,38 @@ private function cleanFinished(): bool
...
 
-            return false;
-        }
+        return true;
+    }
 
-        $mutantProcess->getProcess()->start();
+    /**
+     * @param ProcessBearer[] $bucket
+     * @param Generator|ProcessBearer[] $input
```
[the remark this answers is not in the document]
sanmai: Won't work for my IDE :(

tests/phpunit/Console/OutputFormatter/DotFormatterTest.php
```diff
@@ -110,6 +114,58 @@ public function test_not_covered_correctly_in_console(): void
...
+            <<<'TXT'
+
+.: killed, M: escaped, S: uncovered, E: fatal error, T: timed out
+
+..................................................   ( 50 / 125)
+..................................................   (100 / 125)
+.........................                            (125 / 125)
+TXT
+            ),
+            strip_tags($buffer->fetch())
```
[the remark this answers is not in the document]
sanmai: Done

tests/phpunit/Console/OutputFormatter/DotFormatterTest.php
```diff
@@ -110,6 +114,58 @@ public function test_not_covered_correctly_in_console(): void
...
+
+        $buffer = new BufferedOutput();
+        $dot = new DotFormatter($buffer);
+        $dot->start($totalMutations);
+
+        for ($i = 0; $i < $totalMutations; ++$i) {
+            $dot->advance($this->createMutantExecutionResultsOfType(MutantProcess::CODE_KILLED)[0], $totalMutations);
+        }
+
+        $this->assertSame(str_replace("\n", PHP_EOL,
```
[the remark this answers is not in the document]
sanmai: Done

tests/phpunit/Console/OutputFormatter/DotFormatterTest.php
```diff
@@ -110,6 +114,58 @@ public function test_not_covered_correctly_in_console(): void
         $dot->advance($this->createMutantExecutionResultsOfType(MutantProcess::CODE_NOT_COVERED)[0], 10);
     }
 
+    public function test_it_prints_total_number_of_pending_mutations(): void
```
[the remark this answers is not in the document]
sanmai: Done

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -96,20 +123,18 @@ public function run(iterable $processes, int $threadCount, int $poll = 1000): vo
     private function cleanFinished(): bool
```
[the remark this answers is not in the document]
sanmai: That works too

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -120,22 +145,38 @@ private function cleanFinished(): bool
         return false;
     }
 
-    private function startProcess(MutantProcess $mutantProcess): bool
+    private function startProcess(ProcessBearer $processBearer): bool
```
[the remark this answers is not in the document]
sanmai: Done

src/Container.php
```diff
@@ -392,11 +399,14 @@ public static function create(): self
                 );
             },
             MutationTestingRunner::class => static function (self $container): MutationTestingRunner {
+                $config = $container->getConfiguration();
```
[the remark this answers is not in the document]
sanmai: Fixed

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -120,22 +145,38 @@ private function cleanFinished(): bool
...
+        }
+    }
 
-        return true;
+    /**
+     * @param iterable|ProcessBearer[] $input
+     *
+     * @return Generator|ProcessBearer[]
+     */
+    private static function makeGenerator(iterable &$input): Generator
```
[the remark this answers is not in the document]
sanmai: Done

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -96,20 +123,18 @@ public function run(iterable $processes, int $threadCount, int $poll = 1000): vo
     private function cleanFinished(): bool
     {
         // remove any finished process from the stack
-        foreach ($this->currentProcesses as $index => $mutantProcess) {
-            /** @var MutantProcess $mutantProcess */
-            $process = $mutantProcess->getProcess();
+        foreach ($this->currentProcesses as $index => $processBearer) {
+            /** @var ProcessBearer $processBearer */
```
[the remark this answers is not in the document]
sanmai: Removed

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -49,41 +47,70 @@
...
+        $bucket = [];
+
+        // Load the first process from the queue to buy us some time.
+        self::fillBucket($bucket, $generator, 1);
+
         $threadCount = max(1, $threadCount);
 
         // start the initial batch of processes
-        foreach ($processes as $process) {
+        while ($process = array_shift($bucket)) {
```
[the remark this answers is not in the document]
sanmai: Done

sanmai:
I think I've addressed anything there's to address. Please have a look.

src/Process/Runner/MutationTestingRunner.php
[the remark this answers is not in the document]
theofidry: Yep

sanmai:
> or if it wouldn't be better to try to parallelise the parsing of the files instead

That's the plan. To streamline file parsing following the same pattern. Thought I haven't looked in this very close: it might be not as easy.

src/Process/Runner/Parallel/ParallelProcessRunner.php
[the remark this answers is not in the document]
theofidry: yes but at least it's correct, otherwise it's interpreted as `Generator<mixed>|ProcessBearer[]` which is... not good. But it works sometimes for me so I wonder if the PHP extensions plugin has anything to do with this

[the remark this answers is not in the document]
sanmai: What interprets it as such? I recall `Iterator|Object[]` was invented and/or popularised by PHPStorm folks, so virtually every other tool can read it as `Iterator<Object>`. More importantly I think PHPStan is fine with this.

tests/phpunit/Console/OutputFormatter/DotFormatterTest.php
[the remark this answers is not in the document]
sanmai: Alright, I made it a prime number. Should suit the purpose just fine.

sanmai:
If I "parallelize" parsing on top of these changes I end up with:

> Memory: 42.00MB

The same branch running for Psalm tops at about 200-300Mb, where before it was gigabytes, not only it took forever to run. I don't want to complicated this PR even more with parallelized parsing, tbh. Let's see how it goes here first.

src/Process/Builder/SubscriberBuilder.php
```diff
@@ -208,6 +198,7 @@ private function getInitialTestsConsoleLoggerSubscriber(TestFrameworkAdapter $te
 
     private function shouldSkipProgressBars(): bool
     {
+        // This really does not belong here. How's that going to be tested?..
         return $this->noProgress
             || getenv('CI') === 'true'
             || getenv('CONTINUOUS_INTEGRATION') === 'true';
```
sanmai: Need to be dealt with once the CI resolver from the replacement to #1067 comes into play.

tests/phpunit/AutoReview/ProjectCode/ProjectCodeProvider.php
```diff
@@ -98,7 +97,6 @@ final class ProjectCodeProvider
         ComposerExecutableFinder::class,
         BadgeApiClient::class,
         ResultsLoggerTypes::class,
-        MutantCreatingConsoleLoggerSubscriber::class,
```
maks-rafalko: Why did we remove it?

src/Mutation/FileMutationGenerator.php
```diff
@@ -73,15 +73,15 @@ public function __construct(
      *
      * @throws UnparsableFile
      *
-     * @return Mutation[]
+     * @return iterable|Mutation[]
```
theofidry:
```suggestion
     * @return iterable<Mutation>
```

src/Mutation/MutationGenerator.php
```diff
@@ -95,14 +95,12 @@ public function __construct(
      *
      * @return Mutation[]
```
theofidry:
```suggestion
     * @return iterable<Mutation>
```

src/Mutator/NodeMutationGenerator.php
```diff
@@ -86,27 +85,21 @@ public function __construct(
     /**
      * @return Mutation[]
```
theofidry:
```suggestion
     * @return iterable<Mutation>
```

src/Mutator/NodeMutationGenerator.php
```diff
@@ -86,27 +85,21 @@ public function __construct(
...
+    public function generate(Node $node): iterable
     {
-        return array_reduce(
-            $this->mutators,
-            function (array $mutations, Mutator $mutator) use ($node): array {
-                return $this->generateForMutator($node, $mutator, $mutations);
-            },
-            []
-        );
+        yield from take($this->mutators)->map(function (Mutator $mutator) use ($node) {
```
theofidry: I feel like in this case `nikic/iter` would be a better fit here:

```php
yield from map(
    map(function () { ... },
    $this->mutators
);
```

I'm checking the nikic alternative elsewhere: it doesn't make sense to use pipe _and_ iter. But I admit I have a liking for iter whenever dealing with iterables

src/Mutator/NodeMutationGenerator.php
```diff
@@ -86,27 +85,21 @@ public function __construct(
...
+            yield from $this->generateForMutator($node, $mutator);
+        });
     }
 
     /**
-     * @param Mutation[] $mutations
-     *
-     * @return Mutation[]
+     * @return iterable|Mutation[]
```
theofidry:
```suggestion
     * @return iterable<Mutation>
```

src/PhpParser/Visitor/MutationsCollectorVisitor.php
```diff
@@ -46,9 +46,9 @@
 final class MutationsCollectorVisitor extends NodeVisitorAbstract
 {
     /**
-     * @var Mutation[]
+     * @var iterable[]
      */
-    private $mutations = [];
+    private $iterators = [];
```
theofidry:
```suggestion
    private $mutationChunks = [];
```

? (terminology taken after `array_chunk()`) - requires to refactor the occurrences below as well including the variables in the `foreach` statements

src/PhpParser/Visitor/MutationsCollectorVisitor.php
```diff
@@ -59,25 +59,25 @@ public function __construct(NodeMutationGenerator $mutationGenerator)
...
-            $this->mutations[] = $mutation;
-        }
+        $this->iterators[] = $this->mutationGenerator->generate($node);
 
         return null;
     }
 
     /**
-     * @return Mutation[]
+     * @return iterable|Mutation[]
```
theofidry:
```suggestion
     * @return iterable<Mutation>
```

src/Process/Builder/SubscriberBuilder.php
```diff
@@ -130,40 +128,41 @@ public function registerSubscribers(
...
-                    $this->infectionConfig->getLogs(),
-                    $output
-                )
-            ),
-            new PerformanceLoggerSubscriber(
-                $this->stopwatch,
-                $this->timeFormatter,
-                $this->memoryFormatter,
+    ): iterable {
+        yield $this->getInitialTestsConsoleLoggerSubscriber($testFrameworkAdapter, $output);
```
theofidry: why are the subscriber now yielded?

src/Process/Runner/MutationTestingRunner.php
```diff
@@ -58,45 +60,82 @@ final class MutationTestingRunner
...
     ) {
         $this->processBuilder = $mutantProcessBuilder;
         $this->mutantFactory = $mutantFactory;
         $this->parallelProcessManager = $parallelProcessManager;
         $this->eventDispatcher = $eventDispatcher;
+        $this->runConcurrently = $runConcurrently;
     }
 
     /**
      * @param Mutation[] $mutations
```
theofidry:
```suggestion
     * @param iterable<Mutation> $mutations
```

src/Process/Runner/MutationTestingRunner.php
```diff
@@ -58,45 +60,82 @@ final class MutationTestingRunner
...
+            return false;
+        });
 
         $this->parallelProcessManager->run($processes, $threadCount);
 
         $this->eventDispatcher->dispatch(new MutationTestingWasFinished());
     }
+
+    /**
+     * @param iterable|mixed[] $subjects
```
theofidry:
```suggestion
     * @param iterable<Mutation> $mutations
```

src/Process/Runner/MutationTestingRunner.php
```diff
@@ -58,45 +60,82 @@ final class MutationTestingRunner
...
+     */
+    private function bufferAndCountIfNeeded(iterable &$subjects): int
+    {
+        if ($this->runConcurrently) {
+            return 0;
```
theofidry: I'm not sure to understand: why `0` here?

src/Process/Runner/MutationTestingRunner.php
```diff
@@ -58,45 +60,82 @@ final class MutationTestingRunner
...
+
+        $buffer = [];
+
+        // iterator_to_array wants \Traversable, we can have anything else
+        foreach ($subjects as $subject) {
+            $buffer[] = $subject;
+            // TODO in PHP 7.4 use [...$subjects];
+        }
+
+        $subjects = $buffer;
```
theofidry: why is that step necessary? is it because the iterator might not be rewindable?

src/Process/Runner/MutationTestingRunner.php
```diff
@@ -58,45 +60,82 @@ final class MutationTestingRunner
...
-        $this->eventDispatcher->dispatch(new MutantsCreationWasStarted(count($mutations)));
+        $numberOfMutants = $this->bufferAndCountIfNeeded($mutations);
 
-        $processes = array_map(
-            function (Mutation $mutation) use ($testFrameworkExtraOptions): MutantProcess {
-                $mutant = $this->mutantFactory->create($mutation);
+        $this->eventDispatcher->dispatch(new MutantsCreationWasStarted($numberOfMutants));
 
-                $process = $this->processBuilder->createProcessForMutant($mutant, $testFrameworkExtraOptions);
+        $processes = take($mutations);
```
theofidry:
```php
$processes = take($mutations)
    ->map(function (...) {
        // ...
    })
    ->filter(function (...) {
        // ...
    })
;
```
? (CS nitpick - so up to you)

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -49,69 +48,95 @@
...
     {
+        /*
+         * It takes about 100000 ms for a mutated process to finish, where it takes
+         * about 5000 ms to make it. Therefore instead of just waiting we can produce
+         * new processes so that when a process or several finish, we would have
+         * additional jobs on hand, without a need to wait for them to be created.
+         *
+         * For our purposes we need to make sure we only see one process only once.
+         */
+        $generator = self::toGenerator($processes);
```
theofidry:
```suggestion
        $generator = iter\toIter($processes);
```

src/Process/Runner/Parallel/ParallelProcessRunner.php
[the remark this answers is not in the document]
theofidry: >What interprets it as such?

phpStorm, or maybe it's inferring from somewhere else but that's what I get quite often in those cases :/

It's a convention pushed by both PHPStan and psalm and IMO make a lot more sense and that we adopted so far elsewhere in the codebase

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -120,22 +145,36 @@ private function cleanFinished(): bool
...
-            $this->eventDispatcher->dispatch(new MutantProcessWasFinished(
-                MutantExecutionResult::createFromProcess($mutantProcess)
-            ));
+        $this->runningProcesses[] = $processBearer;
+    }
 
-            return false;
+    /**
+     * @param ProcessBearer[] $bucket
+     * @param Generator|ProcessBearer[] $input
```
theofidry:
```suggestion
     * @param Generator<ProcessBearer> $input
```

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -120,22 +145,36 @@ private function cleanFinished(): bool
...
-        $this->currentProcesses[] = $mutantProcess;
+        while ($input->valid() && count($bucket) < $level) {
+            $bucket[] = $input->current();
+            $input->next();
+        }
+    }
 
-        return true;
+    /**
+     * @param iterable|ProcessBearer[] $input
```
theofidry:
```suggestion
     * @param iterable<ProcessBearer> $input
```

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -120,22 +145,36 @@ private function cleanFinished(): bool
...
+     *
+     * @return Generator|ProcessBearer[]
```
theofidry:
```suggestion
     * @return Generator<ProcessBearer>
```

src/Process/Runner/Parallel/ProcessBearer.php
```diff
@@ -33,36 +33,16 @@
...
-    }
-
-    public function onMutantsCreationWasStarted(MutantsCreationWasStarted $event): void
-    {
-        $this->output->writeln([
-            '',
-            sprintf('Creating mutated files and processes: %s', $event->getMutantCount()),
-        ]);
-    }
+    public function markTimeout(): void;
```
theofidry: `markAsTimedOut`?

tests/phpunit/Mutation/FileMutationGeneratorTest.php
```diff
@@ -102,6 +102,8 @@ public function test_it_generates_mutations_for_a_given_file(): void
             []
         );
 
+        $mutations = iterator_to_array($mutations, false);
```
theofidry: why is it necessary to convert it to an array before?

theofidry:
@sanmai my comments are mostly nitpicks at this point but there is a few questions left

tests/phpunit/AutoReview/ProjectCode/ProjectCodeProvider.php
sanmai: We're not logging these events any more because we create mutants as we go in either mode.

src/Mutator/NodeMutationGenerator.php
sanmai: `nikic/iter` isn't solving the problem of `array_map` hell. E.g it doesn't have a fluent interface. It's a deal breaker for me.

src/Mutator/NodeMutationGenerator.php
sanmai: We discussed in another PR that this isn't an issue, [and PHPStan agrees](https://phpstan.org/r/4d335ef9-f576-430a-82ba-30e10b744fd4). More like `iterable` is redundant here and it should be just `@return Mutation[]`

src/Process/Runner/Parallel/ParallelProcessRunner.php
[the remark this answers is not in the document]
sanmai: We can see that [PHPStan is fine with iterables](https://phpstan.org/r/4d335ef9-f576-430a-82ba-30e10b744fd4), but other Iterators are lacking such understanding. I'm going to use the syntax you propose for them as you propose.

src/Process/Runner/Parallel/ProcessBearer.php
sanmai: It was like that. Isn't that going to be unrelated change?

tests/phpunit/Mutation/FileMutationGeneratorTest.php
sanmai: It is treated like an array there. E.g. there's an `assertArrayHasKey` few lines below.

src/Mutator/NodeMutationGenerator.php
sanmai: OK, nevemind. I gave it a second though and I think it doesn't matter if PHPStan can dig it all right. It is more important that humans don't have these powers. For human reader `iterable<Mutation>` works best. Definitely quote me on that.

src/PhpParser/Visitor/MutationsCollectorVisitor.php
sanmai: They're chunks indeed.

src/Process/Builder/SubscriberBuilder.php
sanmai: Why do we need to buffer then in an array?

src/Process/Runner/MutationTestingRunner.php
sanmai: No problem, that's going to work.

src/Process/Runner/MutationTestingRunner.php
sanmai: It'll be fed to `MutationTestingWasStarted` and later fed to an `OutputFormatter` like `DotFormatter` which wants an `int`.   This isn't the cleanest solution, sure, but that's avoids changing the event system even further.

src/Process/Runner/MutationTestingRunner.php
sanmai: It might not be an iterator. We're given an iterable, which can be any of the things.

src/Process/Runner/Parallel/ParallelProcessRunner.php
sanmai: [I'll be rewindable.](https://github.com/nikic/iter/blob/master/src/iter.php#L1024) Generators are not. I want to be very explicit about this here.

`NoRewindIterator` is another can of worms if you ask.

tests/phpunit/AutoReview/ProjectCode/ProjectCodeProvider.php
sanmai: I also removed MutantsCreationWasFinished, MutantsCreationWasStarted

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -49,69 +48,95 @@
...
      */
     public function run(iterable $processes, int $threadCount, int $poll = 1000): void
     {
+        /*
+         * It takes about 100000 ms for a mutated process to finish, where it takes
+         * about 5000 ms to make it. Therefore instead of just waiting we can produce
+         * new processes so that when a process or several finish, we would have
+         * additional jobs on hand, without a need to wait for them to be created.
+         *
+         * For our purposes we need to make sure we only see one process only once.
```
sanmai:
```suggestion
         * For our purposes we need to make sure we only see one process only once. Thus,
         * we use a generator here which is both non-rewindable, and will fail loudly if tried.
```

sanmai:
@theofidry thank you for the review. I addressed most of your comments, and left a reply where it wasn't addressed in code. Please see if things any more clear now.

sanmai:
@maks-rafalko thank you for the review. I glad this PR was to your liking!

Remaining concerns are the following:
 - Docs need an update. I do think performance concerns deserve a separate discussion, so I would stick with an update to `--no-progress` for now, if you wouldn't mind.
 - This is not a final state of affairs, there's a lot more to do to make Infection work flawlessly as it can be with big projects like Psalm. But I'd like to finish with this PR first because further work seems to require somewhat more effort and consideration, and I do want to invest into it only if we have an agreement about this line of work.
- There are some messy places that may require some refactoring. E.g. DotFormatter and ParallelProcessRunner, I don't think these deficiencies should be holding us up, but they're worth mentioning in case someone want to give a hand with them.
- Theo seems to be concerned with the extra complexity iterables bring on the table. His concerns are not unwarranted, but that's the price to pay to make things even more efficient that they were. I looking towards them being resolved so that we're all on the same page here.

tests/phpunit/AutoReview/ProjectCode/ProjectCodeProvider.php
theofidry: And `MutantWasCreated`

src/Mutator/NodeMutationGenerator.php
theofidry: I would also argue that when a convention is in place in the codebase, it's better to challenge it in a dedicated PR and change it everywhere :)

src/Process/Runner/Parallel/ProcessBearer.php
theofidry: Fair I'll propose it in a dedicated PR

src/Process/Runner/MutationTestingRunner.php
theofidry: Would `-1` do the job as well? Because `0` is a legit values so `-1` would have the merit to be a bit more conventional to say "there's no count"

theofidry: If it's not I'll change that in a dedicated PR after this one is merge, no biggie

src/Process/Runner/MutationTestingRunner.php
theofidry: I mean why is it necessary to convert `iterable $mutations` into `array $mutations` here

sanmai: We need to count them to display the number in the progress bar.

theofidry: but is the assignment necessary (which is what requires a pass by reference)? you are already returning the `count()`

sanmai: Returning to previous question: yes, we have several non-rewindable generators upstream.

sanmai:
@maks-rafalko 

Build is green. Anytime you're ready.

maks-rafalko:
Awesome! Thank you, @sanmai

sanmai:

PR #1083, opened 2020-02-27
Parallel mutant process

theofidry:
Sub-PR of #1082 since I'm the one annoying @sanmai about it.

That said after profiling... I'm surprised to not see any improvements at all, even the opposite:

- https://blackfire.io/profiles/compare/94127bb4-ef7b-4de5-a57f-5bdb2d2eef4e/graph
- https://blackfire.io/profiles/compare/0fc7e7b5-0210-4e84-b232-aa3fe88db6c7/graph


sanmai:
If you consider the mutation process alone it ain't gonna work faster, because the amount of work stays the same. There are all the same files, all the same coverage report. I recall we even adding some additional work: these generators are not as cheap as arrays. I'm sure there are ways to mutation process *actually* faster (as opposed to mutation testing), but this isn't the goal of this proposal.

But that's not the point. It does not matter if a certain process takes longer time to run. There's always a some sort of trade-off when trying to parallelize things. The point is from the user's perspective the entire, whole, process finishes faster.

theofidry:
>The point is from the user's perspective the entire, whole, process finishes faster.

IMO if you remove the total mutation count, although the end result is slightly faster, it will be a regression from a user perspective: you have absolutely no idea of the progress. You also prevent any ability to do a "dry run" i.e. just see the number of mutations generated

theofidry:
@sanmai maybe we can remove "unnecessary" changes from this PR then? I still think the removal of the "generating mutant processes" log entry is nice: it's a lot of code for not much and don't bring any value to the user

sanmai:
Yes, we gonna do something like that, but let me finish with my morning coffee :)

theofidry:
I'm on my last infusion so might have to catch up with your changes tomorrow evening ;)

sanmai:
That's what I thought. Good night to you!

sanmai:
"generating mutant processes" is gone

PR #1084, opened 2020-02-28
Use Reflection to infer EventSubscriber methods

sanmai:
- [x] Replace #1071: it made a valid point that we overcomplicate code that isn't "hot" at all
- [x] Add an extra test
- [x] Remove several long and complicated tests we no longer need
- [x] Remove any need to maintain `getSubscribedEvents()`
- [x] Remove `EventDispatcher::getSubscribedEvents()`


src/Event/EventDispatcher/SyncEventDispatcher.php
```diff
@@ -59,11 +62,45 @@ public function dispatch(object $event): void
...
+    private function inferSubscribedEvents(EventSubscriber $eventSubscriber): iterable
+    {
+        $class = new ReflectionClass($eventSubscriber);
+        $methods = $class->getMethods(ReflectionMethod::IS_PUBLIC);
+
+        foreach ($methods as $method) {
+            /** @var ReflectionMethod $method */
+            if (strpos($method->name, 'on') !== 0) {
+                continue;
+            }
```
sanmai: Though we can consider all and every method except for constructors, removing this convention, I think it may bite us later on. Here we're playing it safe.

On the other hand, we don't care how the methods named. We only care about types of arguments.

theofidry: I find keeping it a bit confusing, also the naming convention is guarded by the auto-review tests so we're good there

maks-rafalko: > Also the naming convention is guarded by the auto-review

when you write the code, you have no idea about this autoreview tests and conventions.  My 2 cents is to remove the `on***` convention, if we will go with this PR.

As long as you have a typehinted method - you are done. E.g. Symfony Messenger has no convention - you can use any name for the method that handles a Message, it just needs to have type declaration of this Message.

sanmai: I'm not holding this convention dear, not at all. Should I remove the corresponding autoreview test also?

What I hate most is to argue about inessential problems. And conventions are the primary method to avoid the issue. Say, we have a convention to use spaces for indentation, and no one in their sane mind will argue about using tabs. For the same reason I didn't vote to remove this convention from the start: who knows how many lances were broken during a birth of this convention? If you're sure this convention is unwarranted, let's remove it then.

theofidry: IMO this convention makes total sense: it provides a no-brainer and expressive way to name the subscription plus also help with refactoring, if you had:

```php
function onFoo(Foo $event): void {}
```

And refactor `Foo` to `Bar`, it is very easy to forget to rename `onFoo()` to `onBar()`. Also since this convention is relatively natural, it is unlikely you would ever really run into an issue. At least not more than forgetting `final` in a new class or test class or `@internal`...

sanmai:
`getSubscribedEvents` is gone

tests/phpunit/Fixtures/Event/NullSubscriber.php
```diff
@@ -8,12 +8,11 @@
...
+    public $count = 0;
 
-    /**
-     * @return array<string, callable>
-     */
-    public function getSubscribedEvents(): array
+    public function __construct(UserWasCreated $event)
     {
-        return [];
+        $this->count ++;
```
sanmai:
```suggestion
        ++$this->count; // Should never be called more than once
```

maks-rafalko:
Thank you @sanmai

PR #1087, opened 2020-02-29
Improve IgnoreConfig

sanmai:
- Profiling shows that IgnoreConfig in a very hot class, and casual observation confirms that indeed it has a method potentially called for every mutator, for every line in every file.
- Update it to return early if nothing is ignored. This is the most common case.
- Use `array_key_exists` on a map instead of `in_array` which is a linear search.


theofidry:
👌

PR #1088, opened 2020-02-29
Improve JUnitTestCaseSorter

sanmai:
- This is also a class with a very hot method, called for literally every covered line.
- Don't do any sorting etc if there's only one input to get 5% speed up in a benchmark on live data
- Use a map lookup instead of `in_array`.

~15% faster in total.

This new version normally requires more than one assertion, where re-adding assertions will void all and any gain.


src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
```diff
@@ -51,8 +51,18 @@ final class JUnitTestCaseSorter
      *
      * @return string[]
      */
-    public function getUniqueSortedFileNames(array $coverageTestCases): array
+    public function getUniqueSortedFileNames(array $coverageTestCases): iterable
```
theofidry: The return type should be returned to `iterable<string>`

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
```diff
@@ -51,8 +51,18 @@ final class JUnitTestCaseSorter
...
     {
+        if (count($coverageTestCases) === 1) {
+            foreach ($coverageTestCases as $coverageLineData) {
```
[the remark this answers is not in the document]
sanmai: We don't know if there are numeric keys, and we don't know yet if they're sequential starting from 0. `foreach` is the easiest way to dodge this issue.

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
sanmai: Why do you think this is a problem? PHPStan disagrees, [says this is already understood](https://phpstan.org/r/4d335ef9-f576-430a-82ba-30e10b744fd4).

sanmai:
I don't have a blackfire profile nor profiler at hand, and I am not incurious to see it, but I can tell you from just looking at the code: arrays are hash maps in PHP, and a hash map implies somewhat intricate procedure to hash keys and store them. This isn't a free operation. For a single `array_map` you're doing it twice because you have to index an input array, and it will make another indexed array for you. Where in case of generators nothing like that is going on.

And we're counting nickels here tbh, this way or another at our scale it might be just the same. I more care making a peace of mind. Using `array_map` here is like nuking a fly. It is not necessary at all.

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
```diff
@@ -51,8 +51,18 @@ final class JUnitTestCaseSorter
      *
      * @return string[]
```
sanmai:
```suggestion
     * @return iterable<string>
```

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
sanmai: Fixed.

theofidry:
>And we're counting nickels here tbh, this way or another at our scale it might be just the same. I more care about a peace of mind. Using array_map here is like nuking a fly. It is not necessary at all.

Yes except if you are iterating over all the elements right away though, then you are using a generators which is more error-prone and tricky (as much as I like them!) unnecessarily

I'll try to get a profile: the benefits of hashmap is obvious, I'm not asking a comparison because I question your PR, but because I'm interested in the results

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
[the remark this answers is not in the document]
theofidry: `current()` would do no?

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
```diff
@@ -49,10 +49,20 @@ final class JUnitTestCaseSorter
     /**
      * @param CoverageLineData[] $coverageTestCases
      *
-     * @return string[]
+     * @return iterable<string>
      */
-    public function getUniqueSortedFileNames(array $coverageTestCases): array
+    public function getUniqueSortedFileNames(array $coverageTestCases): iterable
```
theofidry: After taking another look, this `iterable` case does not make sense to me. There is two paths:

- yielding only one result
- yield multiple results

In the second, the yields are preceded by a `usort()` on an array. So there is really no benefits from yielding here from a performance or memory PoV.

Looking at the parent `MutationConfigBuilder`, it also iterates over all the results right away and it does not appear it can take advantage of an iterator either

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
```diff
@@ -63,31 +73,25 @@ static function (CoverageLineData $a, CoverageLineData $b) {
...
+        }
     }
 
     /**
      * @param CoverageLineData[] $coverageTestCases
      *
      * @return CoverageLineData[]
      */
-    private function uniqueByTestFile(array $coverageTestCases): array
+    private function uniqueByTestFile(iterable $coverageTestCases): array
```
theofidry: AFAICS it's `array` here, it's coming from `getUniqueSortedFileNames(array $coverageTestCases)` so there is no need to expand the type

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
```diff
@@ -63,31 +73,25 @@ static function (CoverageLineData $a, CoverageLineData $b) {
             }
         );
 
-        return array_map(
```
theofidry: One improvement I do see there is that we could do the mapping to `testFilePath` directly in `uniqueByTestFile()` to avoid a second looping

theofidry:
Forgot the profile that made me take a second look at the PR: https://blackfire.io/profiles/compare/bd72bf2d-3a9b-4fc8-8aa6-685f312c2a35/graph :(

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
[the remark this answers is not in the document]
sanmai: `foreach` is a shortcut to `reset()` and `current()` IIRC.

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
sanmai: Isn't it going to need a `time` field to sort by it later?

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
sanmai: Fair. Let's not expand.

sanmai:
```suggestion
    private function uniqueByTestFile(array $coverageTestCases): array
```

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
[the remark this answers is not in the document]
sanmai: OK, I got you. `current()` should do.

sanmai:
I don't get this profile tbh. Does it say we added 7000+ calls `array_key_exists`, but did not remove a single call `in_array`?

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
sanmai: How about now?

The point is `MutationConfigBuilder` doesn't need an array.

theofidry:
@sanmai maybe it would be more interesting to use a script mimicking the calls made in infection to better highlight this part: a Blackfire profile is not a call graph, it will omit parts that are insignificant and as you can see in the diff, the wall time difference is already quite insignificant

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
theofidry: damn right, nevermind... then

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
theofidry: >The point is MutationConfigBuilder doesn't need an array.

Sure, it just needs an iterable, but array is always much simpler to deal with than `iterable` or `Generator`, so if the sub-type is enough why expanding it? It's neither a case where there would be a performance benefit or make a more friendly/robust API

sanmai:
@theofidry on the live data benchmark a variant with iterators is 4% faster.

Overall this new version is 15% faster.

sanmai:
@maks-rafalko can you see if these new changes make any sense?

src/TestFramework/Coverage/XmlReport/JUnitTestCaseSorter.php
```diff
@@ -84,10 +88,10 @@ private function uniqueByTestFile(array $coverageTestCases): array
         $usedFileNames = [];
         $uniqueTests = [];
 
-        foreach ($coverageTestCases as $coverageTestCase) {
-            if (!in_array($coverageTestCase->testFilePath, $usedFileNames, true)) {
-                $uniqueTests[] = $coverageTestCase;
-                $usedFileNames[] = $coverageTestCase->testFilePath;
+        foreach ($coverageTestCases as $coverageLineData) {
+            if (!array_key_exists($coverageLineData->testFilePath, $usedFileNames)) {
```
maks-rafalko: would `(string) $converageLineData->testFilePath` do the job instead of ingoring in phpstan config `'#array_key_exists expects int\|string, string\|null given#'`?

sanmai: I was very surprised to learn that `(string) $var` is... 7% slower. I did test this several times to reconfirm.

And yes, we know it is never null here. It's from JUnit report.

maks-rafalko: interesting. Probably this will work faster

```php
assert($coverageLineData->testFilePath !== null);
```

Anyway, thanks for sharing. The current implementation is fine then

sanmai: `assert()` here is about as slow :man_shrugging:

maks-rafalko: I see, thanks for checking!

sanmai: Shall we merge this?

For posterity, here's the benchmark I used.

<details>
  <summary>benchmark.php</summary>

```php
<?php
declare(strict_types=1);

use Infection\TestFramework\Coverage\XmlReport\JUnitTestCaseSorter;
use Infection\TestFramework\Coverage\XmlReport\JUnitTestCaseSorterArrayMap;
use Infection\TestFramework\Coverage\XmlReport\JUnitTestCaseSorterCast;
use Infection\TestFramework\Coverage\XmlReport\JUnitTestCaseSorterOld;

include 'vendor/autoload.php';

// To log, inside getUniqueSortedFileNames() add:
// file_put_contents('/tmp/JUnitTestCaseSorter.log', base64_encode(serialize($coverageTestCases)).PHP_EOL, FILE_APPEND | LOCK_EX);

$testCases = iterator_to_array((static function () {
    $file = fopen('/tmp/JUnitTestCaseSorter.log', 'r');

    while (($line = fgets($file)) !== false) {
        yield unserialize(base64_decode($line));
    }
    fclose($file);
})());

$sorters = [
    new JUnitTestCaseSorterArrayMap(),
    new JUnitTestCaseSorterCast(),
    new JUnitTestCaseSorter(),
    new JUnitTestCaseSorterOld(),
];

echo "Warm up... \n";

foreach ($sorters as $caseSorter) {
    foreach ($testCases as $coverageTestCases) {
        foreach ($caseSorter->getUniqueSortedFileNames($coverageTestCases) as $filename) {
        }
    }
}

echo "Testing...\n";

$results = [];

foreach ($sorters as $caseSorter) {
    $start = microtime(true);

    for ($i = 0; $i < 10; ++$i) {
        foreach ($testCases as $coverageTestCases) {
            foreach ($caseSorter->getUniqueSortedFileNames($coverageTestCases) as $filename) {
                $filename;
            }
        }
    }

    $total = microtime(true) - $start;

    $class = get_class($caseSorter);
    $results[$class] = $total;
    echo "$class: {$total}\n";
}

echo 'New vs Old: ' . round(100 * $results[JUnitTestCaseSorter::class] / $results[JUnitTestCaseSorterOld::class]) . "\n";
echo 'Iter vs Array: ' . round(100 * $results[JUnitTestCaseSorter::class] / $results[JUnitTestCaseSorterArrayMap::class]) . "\n";
echo 'Ignore vs Cast: ' . round(100 * $results[JUnitTestCaseSorter::class] / $results[JUnitTestCaseSorterCast::class]) . "\n";

```
</details>

maks-rafalko: > Shall we merge this?

Sure, thought you can do it yourself since I removed PrettyCI integration completele. But now I see builds are still in progress.

Let's get it merged

maks-rafalko:
Thank you @sanmai

sanmai:

PR #1089, opened 2020-02-29
Rename method from `showProgress` to `noProgress` to avoid confusion 

maks-rafalko:
During reviewing of https://github.com/infection/infection/pull/1082 I noticed that Concurrent Processing should be enabled only when `--no-progress` is used, while the method that returns the value of this option is named `showProgress()` which is the opposite by its meaning

https://github.com/infection/infection/pull/1082/files#diff-5772c249f61507a501a28c2087d7d2a8R407

So here is an update.

### DIFF
--- a/src/Configuration/Configuration.php
+++ b/src/Configuration/Configuration.php
@@ -246,7 +246,7 @@ class Configuration
         return $this->formatter;
     }
 
-    public function showProgress(): bool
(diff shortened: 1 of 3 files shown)


maks-rafalko:
> I suppose it's named correctly in `ConfigurationFactory`?

no issues there as far as I can see

PR #1090, opened 2020-02-29
Tweak the BadgeApiClient

theofidry:
- Move the request part of `BadgeApiClient` into a dedicated `JsonClient` class which allows to more easily test `BadgeApiClient`
- Rename `BadgeApiClient` to `StrykerDashboardClient`
- Uncouple `BadgeApiClient` from the console (`OutputInterface`) in favour of `Psr\LoggerInterface`


tests/phpunit/Http/StrykerDashboardClientTest.php
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * @var DummyLogger
+     */
+    private $logger;
+
+    /**
+     * @var StrykerDashboardClient
+     */
+    private $dashboadClient;
```
localheinz:
```suggestion
    private $dashboardClient;
```

tests/phpunit/Http/StrykerDashboardClientTest.php
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    public function test_it_can_send_a_report(): void
+    {
+        $this->clientMock
+            ->expects($this->once())
+            ->method('request')
+            ->with(
+                'https://dashboard.stryker-mutator.io/api/reports',
+                json_encode(json_decode(<<<'JSON'
```
localheinz: 🤔

src/Http/JsonClient.php
```diff
@@ -0,0 +1,70 @@
...
+use function Safe\curl_exec;
+use function Safe\curl_getinfo;
+use function Safe\curl_init;
+use function Safe\curl_setopt;
+
+/**
+ * @internal
+ * @final
+ */
+class JsonClient
```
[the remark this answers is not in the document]
theofidry: Because `request(string $json)`

src/Http/Response.php
```diff
@@ -0,0 +1,67 @@
...
+
+    public function __construct(string $body, int $statusCode)
+    {
+        Assert::range($statusCode, 200, 599);
+
+        $this->body = $body;
+        $this->statusCode = $statusCode;
+    }
+
+    public function getBody(): string
```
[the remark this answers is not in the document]
theofidry: I'm not a big fan of omitting it: 
- it can create some confusion at whether you mean the property or method
- it is the convention so far in the codebase
- using a public property here would mean loosing type and making this VO mutable

src/Http/StrykerDashboardClient.php
```diff
@@ -35,30 +35,25 @@
...
 use function Safe\sprintf;
-use Symfony\Component\Console\Output\OutputInterface;
 
 /**
  * @internal
  */
-class BadgeApiClient
+class StrykerDashboardClient
 {
     private const STRYKER_DASHBOARD_API_URL = 'https://dashboard.stryker-mutator.io/api/reports';
```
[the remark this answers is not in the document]
theofidry: I do prefer having magic constant values as actual constant rather than being inlined: they are easier to find and make it clear that they are acting as a sort of config

src/Http/Response.php
```diff
@@ -0,0 +1,67 @@
...
+namespace Infection\Http;
+
+use Webmozart\Assert\Assert;
+
+/**
+ * @internal
+ */
+final class Response
+{
+    public const CREATED_RESPONSE_CODE = 201;
```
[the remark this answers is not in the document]
theofidry: Yes since we are checking against it, it doesn't hurt to have something more expressive than `201` :)

tests/phpunit/Http/StrykerDashboardClientTest.php
theofidry: I want to keep a readable JSON, but the JSON passed is a compacted one hence the process

src/Http/Response.php
```diff
@@ -0,0 +1,67 @@
...
+
+    private $body;
+    private $statusCode;
+
+    public function __construct(string $body, int $statusCode)
+    {
+        Assert::range($statusCode, 200, 599);
```
localheinz: What do you think about adding tests for this guard clause?

src/Http/Response.php
```diff
@@ -0,0 +1,67 @@
...
+ * @internal
+ */
+final class Response
+{
+    public const CREATED_RESPONSE_CODE = 201;
+
+    private $body;
+    private $statusCode;
+
+    public function __construct(string $body, int $statusCode)
```
localheinz: What do you think about flipping arguments here?

```diff
diff --git a/src/Http/Response.php b/src/Http/Response.php
index aff73beb..0a5ecef4 100644
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -44,24 +44,24 @@ final class Response
 {
     public const CREATED_RESPONSE_CODE = 201;

-    private $body;
     private $statusCode;
+    private $body;

-    public function __construct(string $body, int $statusCode)
+    public function __construct(int $statusCode, string $body)
     {
         Assert::range($statusCode, 200, 599);

-        $this->body = $body;
         $this->statusCode = $statusCode;
+        $this->body = $body;
     }

-    public function getBody(): string
+    public function getStatusCode(): int
     {
-        return $this->body;
+        return $this->statusCode;
     }

-    public function getStatusCode(): int
+    public function getBody(): string
     {
-        return $this->statusCode;
+        return $this->body;
     }
 }
diff --git a/tests/phpunit/Http/ResponseTest.php b/tests/phpunit/Http/ResponseTest.php
index 02204314..e3bfdefa 100644
--- a/tests/phpunit/Http/ResponseTest.php
+++ b/tests/phpunit/Http/ResponseTest.php
@@ -44,18 +44,18 @@ final class ResponseTest extends TestCase
     /**
      * @dataProvider valueProvider
      */
-    public function test_it_can_be_instantiated(string $body, int $statusCode): void
+    public function test_it_can_be_instantiated(int $statusCode, string $body): void
     {
-        $response = new Response($body, $statusCode);
+        $response = new Response($statusCode, $body);

-        $this->assertSame($body, $response->getBody());
         $this->assertSame($statusCode, $response->getStatusCode());
+        $this->assertSame($body, $response->getBody());
     }

     public function valueProvider(): Generator
     {
-        yield 'empty' => ['', 200];
+        yield 'empty' => [200, ''];

-        yield 'nominal' => ['body', 200];
+        yield 'nominal' => [200, 'body'];
     }
 }
```

src/Http/Response.php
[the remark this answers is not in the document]
localheinz: >I'm not a big fan of omitting it:
>
>* it can create some confusion at whether you mean the property or method

We have static code analysis for this.

>* it is the convention so far in the codebase

There are classes that do not have the `get` prefixes for getters, for example, [`BuildContext`](https://github.com/infection/infection/blob/5f3094ca95593454f47382992ee7ac4d78463c9b/src/Environment/BuildContext.php#L41-L60).

src/Http/StrykerDashboardClient.php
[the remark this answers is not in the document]
localheinz: Using a `private` constant when we could use the literal value instead adds overhead - you need to navigate to the constant to find out about the value.

src/Http/Response.php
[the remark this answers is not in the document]
localheinz: Perhaps use [`teapot/status-code`](https://github.com/teapot-php/status-code) - or extract a separate class `Http\StatusCode` with a constant `CREATED`, then?

tests/phpunit/Http/StrykerDashboardClientTest.php
localheinz: Why not use this, then?

```diff
diff --git a/tests/phpunit/Http/StrykerDashboardClientTest.php b/tests/phpunit/Http/StrykerDashboardClientTest.php
index 50c75e7e..0eceb633 100644
--- a/tests/phpunit/Http/StrykerDashboardClientTest.php
+++ b/tests/phpunit/Http/StrykerDashboardClientTest.php
@@ -81,15 +81,12 @@ final class StrykerDashboardClientTest extends TestCase
             ->method('request')
             ->with(
                 'https://dashboard.stryker-mutator.io/api/reports',
-                json_encode(json_decode(<<<'JSON'
-{
-    "apiKey": "foo",
-    "repositorySlug": "infection/infection",
-    "branch": "master",
-    "mutationScore": 80.31
-}
-JSON
-                ))
+                json_encode([
+                    "apiKey"=> "foo",
+                    "repositorySlug"=> "infection/infection",
+                    "branch"=> "master",
+                    "mutationScore"=> 80.31,
+                ])
             )
             ->willReturn(new Response('Report received!', 201))
         ;
@@ -124,15 +121,12 @@ EOF
             ->method('request')
             ->with(
                 'https://dashboard.stryker-mutator.io/api/reports',
-                json_encode(json_decode(<<<'JSON'
-{
-    "apiKey": "foo",
-    "repositorySlug": "infection/infection",
-    "branch": "master",
-    "mutationScore": 80.31
-}
-JSON
-                ))
+                json_encode([
+                    "apiKey"=> "foo",
+                    "repositorySlug"=> "infection/infection",
+                    "branch"=> "master",
+                    "mutationScore"=> 80.31,
+                ])
             )
             ->willReturn(new Response('Report invalid!', 400))
         ;
```

theofidry: Because then you have the exact same code as the source code, which makes the test completely irrelevant

src/Http/Response.php
theofidry: Don't really want to, because they are here to compensate for the lack of typehint and the same way I don't want to test a typehint, I don't really want to test a guard. An exception I can think of is when you want to provide a more helpful error message, then you want to test the guard to ensure the error message is the one you expect.

Erratum: I'm actually gonna put a more helpful message just in case so you'll have your test :)

src/Http/Response.php
theofidry: A bit annoying but I guess it makes sense since you tend to show the status code first

src/Http/Response.php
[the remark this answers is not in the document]
theofidry: I think both suggested solutions are overkill in the current state of affairs: it's only one occurrence

src/Http/StrykerDashboardClient.php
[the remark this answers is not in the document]
theofidry: It's an extra overhead to navigate to find out the value, but it's an overhead less to have an expressive name instead of a value to decrypt

src/Http/Response.php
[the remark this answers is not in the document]
theofidry: Which is pretty much the exception, brought by you, rather than the norm :)

>We have static code analysis for this.

I don't want to require an SA to read. It's minor granted, but it's still a mental overhead which is unnecessary to me since it does not make it more expressive IMO.

theofidry:
>I wonder why don't we use something like Guzzle here. It is perfectly mockable, and you wouldn't need to deal with curl-anything. You won't even need curl.

If there is other cases where we need to do an HTTP call why not, otherwise I think this is way overkill for the case

tests/phpunit/Http/ResponseTest.php
```diff
@@ -0,0 +1,76 @@
...
+        $this->assertSame($body, $response->getBody());
+    }
+
+    public function test_it_provides_a_user_friendly_error_if_the_status_code_is_not_a_valid_HTTP_status_code(): void
+    {
+        try {
+            new Response(102, '');
+
+            $this->fail();
+        } catch (InvalidArgumentException $exception) {
```
localheinz: What do you think about using `expectException()` and `expectExceptionMessage()` here?

```diff
diff --git a/tests/phpunit/Http/ResponseTest.php b/tests/phpunit/Http/ResponseTest.php
index 8034fe51..9311530a 100644
--- a/tests/phpunit/Http/ResponseTest.php
+++ b/tests/phpunit/Http/ResponseTest.php
@@ -55,16 +55,15 @@ final class ResponseTest extends TestCase

     public function test_it_provides_a_user_friendly_error_if_the_status_code_is_not_a_valid_HTTP_status_code(): void
     {
-        try {
-            new Response(102, '');
+        $statusCode = 102;

-            $this->fail();
-        } catch (InvalidArgumentException $exception) {
-            $this->assertSame(
-                'Expected an HTTP status code. Got "102"',
-                $exception->getMessage()
-            );
-        }
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage(sprintf(
+            'Expected an HTTP status code. Got "%d"',
+            $statusCode
+        ));
+
+        new Response($statusCode, '');
     }

     public function valueProvider(): Generator
```

theofidry: I think the current form is more "traditional" and less "PHPUnit magic". IIRC there was also a time Sebastian wanted to remove it but was considered a bit too extreme.

I also much prefer this form since you in some cases it's clearer:

```php
function test_something(): void {
  // some setup

  try {
    // only the statement that fail
    $this->fail();
  } catch (Exception $e) {
     // ...
  }
}
```

Whereas if you have:

```php
function test_something(): void {
  $this->expectException($exception)

  // other stuff: what statement fails exactly here?
}
```

But anyway, it's something that maybe we should address in an RFC, write down in the ADR document and fix in the whole codebase once for all

localheinz: It has already been written down in this blog post: https://thephp.cc/news/2016/02/questioning-phpunit-best-practices.

theofidry: Fine, let's change it globally though not just here

theofidry:
Any concern left for this PR? So far the comments were a lot of nitpicking but still no approval :)

src/Logger/BadgeLogger.php
```diff
@@ -95,25 +95,23 @@ public function log(): void
...
+        $this->strykerDashboardClient->sendReport(
             $apiKey,
             'github.com/' . $buildContext->repositorySlug(),
             $buildContext->branch(),
             $this->metricsCalculator->getMutationScoreIndicator()
         );
     }
 
-    private function showInfo(string $message): void
+    private function logError(string $message): void
```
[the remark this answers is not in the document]
theofidry: The downstream corresponding log level for `LoggerInterface` is `warning`. But even so `logWarning` would be more accurate. Maybe just gonna go with `logMessage()` which removes confusion. I prefer "log" over "show" here because show is as if you are showing the message: we're not necessarily doing so

theofidry:
>Part of the builds still fails due to MSI I guess

Yes, related to the guards added with Webmoart assertions. I would like to disable those mutations at some point but didn't have a look yet

src/Logger/BadgeLogger.php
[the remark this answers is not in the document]
maks-rafalko: `logMessage()` is 👌

tests/phpunit/Http/StrykerDashboardClientTest.php
SatoAoaka: ✅ 修正しました！

コミット cf09c4e で対応しました。
`server_default=text("ONE_SHOT_GENERATION")` に変更し、シングルクォートを削除しました。

ご指摘ありがとうございました。

PR #1092, opened 2020-03-01
Make E2ETest a member of integration group

sanmai:
- Make E2ETest a member of integration group, because it is very IO intensive.
- Update `IntegrationGroupProviderTest` to include an explicit check for E2ETest. It is important that we do not include E2ETest when running Infection from the very test.

`vendor/bin/phpunit --exclude-group=integration` should be much faster

### DIFF
--- a/tests/phpunit/Console/E2ETest.php
+++ b/tests/phpunit/Console/E2ETest.php
@@ -69,6 +69,7 @@ use Symfony\Component\Process\Exception\ProcessTimedOutException;
 use Symfony\Component\Process\Process;
 
 /**
+ * @group e2e
  * @group integration Requires some I/O operations
  */
 final class E2ETest extends TestCase
--- a/tests/phpunit/Console/E2ETest.php
+++ b/tests/phpunit/Console/E2ETest.php
@@ -69,11 +69,14 @@ use Symfony\Component\Process\Exception\ProcessTimedOutException;
 use Symfony\Component\Process\Process;
 
 /**
- * @group e2e
- * @group integration
+ * @group integration Requires some I/O operations
  */
 final class E2ETest extends TestCase
 {
+    /**
+     * This group must be excluded from E2E testing to avoid endless recursive testing loop situation.
+     */
+    private const EXCLUDED_GROUP = 'integration';
     private const MAX_FAILING_COMPOSER_INSTALL = 5;
     private const EXPECT_ERROR = 1;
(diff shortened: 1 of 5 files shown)


tests/phpunit/Console/E2ETest.php
```diff
@@ -120,14 +121,23 @@ protected function tearDown(): void
         chdir($this->cwd);
     }
 
+    /**
+     * Ensure this test belongs to e2e group. We exclude it during E2E testing
+     * from this class to avoid causing endless recursive testing loop.
+     */
+    public function test_it_belongs_to_e2e_group(): void
```
theofidry: I'm not convinced this one is necessary... You are basically testing that this class is marked with `@group e2e`. Unlike the `@group integration` which aimed to test that if detected an IO/network operation, it should be marked as `integration`

sanmai: Yes, that’s correct. It is marked as such. But also has to be marked e2e. Think this is wrong?

theofidry: I think if marked as `@group e2e` `@group integration` should be unnecessary: I admit did not think that one through when I added the integration checks.

I would remove the need for that test and also make more sense IMO

sanmai: Alright, I’ll try to remove this e2e group designation altogether then.

tests/phpunit/Console/E2ETest.php
```diff
@@ -69,10 +69,14 @@
 use Symfony\Component\Process\Process;
 
 /**
- * @group e2e
```
maks-rafalko: What's the reason for removing this group?

I find it useful to run `--exclude-group=e2e` and run e2e tests separately by `tests/e2e bin/infection` or `tests/e2e build/infection.phar`.

Now, this is impossible.

tests/phpunit/Console/E2ETest.php
sanmai: It turns out @maks-rafalko needs `e2e` group because it is convenient. And it is. And I want this to be member `@group integration` because it is convenient and makes sense since this is a very IO intensive test. How do we do from here?

tests/phpunit/Console/E2ETest.php
sanmai: No problem, I got it back.

maks-rafalko:
Thank you @sanmai

PR #1094, opened 2020-03-01
Move individual logger factories out of LoggerFactory

theofidry:
This is a bit more of a controversial side of my logger refactoring hence the dedicated PR.

For context: in the refactoring I'm doing, I'm kind of moving away from away from inheriting `FileLogger`: the reason is that it requires a few parameters and force all the downstream loggers to use them although they might not need any (most of them just need the metrics calculator). And if most loggers requires a common parameter, e.g. the metrics calculator, it's moved upstream to `FileLogger` because it makes it more convenient...

In the mist of all of that, `LoggerFactory` suffers the most: any new parameter you need to pass to any of the logger needs to be added to `LoggerFactory` as state.

So the proposal is the following: have a dedicated factory for each loggers. Then `FactoryLogger` can focus on creating the right loggers without worrying about the parameters of each loggers. It is more verbose, but allows to refactor an individual logger without touching `LoggerFactory`.

That said I must say I'm not myself 100% convinced on that one, so submitting for review & opinion

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -46,6 +46,7 @@ use Infection\Configuration\Schema\SchemaConfigurationFactory;
 use Infection\Configuration\Schema\SchemaConfigurationFileLoader;
 use Infection\Configuration\Schema\SchemaConfigurationLoader;
 use Infection\Configuration\Schema\SchemaValidator;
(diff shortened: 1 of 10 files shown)


theofidry:
Nay not happy with it

PR #1095, opened 2020-03-01
Refactor the loggers

theofidry:
Whenever I took a look at the loggers I couldn't help but to find they could be made a bit simpler. I am not sure the overall PR is simpler, but here are the key changes:

- `FileLogger` is its own final dedicated class, no longer a base abstract class. This allows to test and pin its behaviour once and only once, not in all its inherited classes as well
- As a consequence of the above, all the previous file loggers are now plain loggers that can be unit tested (see their test differences and you will see how much simpler and enjoyable they are to test)
- A new `LoggerRegistry` which allows `SubscriberBuilder` to consume a simple logger instead of an array of loggers
- Simplify `LoggerFactory`: avoid a double looping and remove the dependence on `ResultsLoggerTypes` (the changed code just do fine without)


src/Logger/LoggerFactory.php
```diff
@@ -72,121 +74,109 @@ public function __construct(
...
-        );
+        return $badge === null
+            ? null
+            : new BadgeLogger(
+                $output,
+                new ChainBuildContextResolver(new TravisCiResolver()),
+                new StrykerApiKeyResolver(),
+                new StrykerDashboardClient(
+                    new JsonClient(),
+                    new ConsoleLogger($output)
```
sanmai: I wish BadgeLogger would know these details itself. Say within a `::build()` method.

src/Logger/TextFileLogger.php
```diff
@@ -36,27 +36,46 @@
...
 use function strlen;
 
 /**
  * @internal
  */
-final class TextFileLogger extends FileLogger
+final class TextFileLogger implements LineMutationTestingResultsLogger
 {
-    protected function getLogLines(): array
+    private $metricsCalculator;
```
sanmai: I wish these loggers would be responsible for the result collection themselves, not MetricsCalculator. Metrics are metrics, logging be logging. Also MetricsCalculator won't collect logs nobody will ask for.

This is probably for another time.

src/Logger/LoggerFactory.php
theofidry: I'm still uncertain about the fate of `LoggerFactory` and attempted something in #1094. So I prefer to not touch that part in this PR

src/Logger/TextFileLogger.php
theofidry: I would also like to split the metrics calculator in two. But I got to admit it kind of complicate its consumption for no good reason... Maybe we can discuss that in another issue or in #1099

PR #1096, opened 2020-03-02
Remove MutantWasCreated

### LINKED ISSUE #1091: Removal of MutantWasCreated
After #1082 I've left [MutantWasCreated](https://github.com/infection/infection/pull/1082/files#diff-34bd975c68ddd342c37abe44b3541baaR91) dangling. It isn't used anywhere, but unlike other (removed) events it may be of use.

Say, we can assign every mutation a sequential ID, and pass it along with `MutantWasCreated` and `MutantProcessWasFinished`, later collecting both events and reconciling the time it took to run a mutation. 

Or, better, we can add `MutantProcessWasAdded`, and use `MutantWasCreated` to estimate a lag we have launching processes, making a suggestion to use, say, a different thread count. Or simply include this metric in a some sort of internal performance report.

- Shall we keep it for time being? 
- Shall we remove it and re-add it later? 
- Is it any good for these purposes?

sanmai:
Fixes #1091


sanmai:
Thank you @maks-rafalko

PR #1097, opened 2020-03-02
Parallel source file collector

sanmai:
Makes sure SourceFileCollector does not buffer and count files unless asked.

Follow-up to #1082


src/Process/Runner/IterableBuffer.php
```diff
@@ -0,0 +1,68 @@
...
+ */
+trait IterableBuffer
+{
+    /**
+     * @param iterable<mixed> $subjects
+     */
+    private static function bufferAndCountIfNeeded(iterable &$subjects, bool $runConcurrently): int
+    {
+        if ($runConcurrently) {
+            // This number is typicaly fed to ProgressFormatter/ProgressBar, or variants.
```
sanmai:
```suggestion
            // This number is typically fed to ProgressFormatter/ProgressBar or variants.
```

src/Process/Runner/IterableBuffer.php
```diff
@@ -0,0 +1,68 @@
...
+            // In progress bar lingo 0 stands for unknown number of steps.
```
sanmai:
```suggestion
            // In progress bar lingo 0 stands for an unknown number of steps.
```

src/Process/Runner/IterableBuffer.php
```diff
@@ -0,0 +1,68 @@
...
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Process\Runner;
```
theofidry: Instead of having it in the `Process\Runner` namespace which makes it really weird to depend on if used in `SourceFileCollector`, what about:

- Naming it `IterableCounter` and moving it at the top level
- Making it a final class, with static method and private constructor instead

The last point is more of a stylistic choice I would say: have a "function" (a static method of a non-instantiable class is a more autoloading friendly option that actual functions in PHP) instead of a trait. Both are perfectly valid though

src/Process/Runner/IterableBuffer.php
```diff
@@ -0,0 +1,68 @@
...
+
+        $buffer = [];
+
+        // iterator_to_array wants \Traversable, we can have anything else
+        foreach ($subjects as $subject) {
+            $buffer[] = $subject;
+            // TODO in PHP 7.4 use [...$subjects];
+        }
+
+        $subjects = $buffer;
```
theofidry: if in a class instead, you could have a private static method `toArray()` instead so reducing it to `$subjects = toArray($subjects)` which makes it a bit more readable IMO

src/Process/Runner/IterableBuffer.php
```diff
@@ -0,0 +1,68 @@
...
+{
+    /**
+     * @param iterable<mixed> $subjects
+     */
+    private static function bufferAndCountIfNeeded(iterable &$subjects, bool $runConcurrently): int
+    {
+        if ($runConcurrently) {
+            // This number is typically fed to ProgressFormatter/ProgressBar or variants.
+            // In progress bar lingo 0 stands for an unknown number of steps.
+            return 0;
```
theofidry: Forgot to double check, would `-1` work as well there? It's a better fit at "no count value" than `0`

sanmai: It says in several places [to use 0 if number of items is unknown](https://github.com/symfony/console/blob/master/Helper/ProgressBar.php#L290).

src/Process/Runner/IterableBuffer.php
sanmai: `IterableCounter` is a better name for sure. At the top level you mean at `Infection\Process`?

src/Process/Runner/IterableBuffer.php
sanmai: I'm afraid a class, that wouldn't make temporary copies of the same array, might not be as concise as this single function. Though no biggie since we mostly deal with objects, I'd rather not to jump into this rabbit hole if possible, as of now.

sanmai: There are other ways to solve this. Say, one can exploit a generator and its ability to return values. But this seems to me even more gross that it is now.

src/Process/Runner/IterableBuffer.php
theofidry: `Infection`: it's not specific of anything so until we have enough utils that could warrant a dedicated namespace I think it's fine to be at the root like `Engine` for example

sanmai: Alright, I made a class with a static method. And added a test.

src/IterableCounter.php
```diff
@@ -0,0 +1,62 @@
...
+            // This number is typically fed to ProgressFormatter/ProgressBar or variants.
+            // In progress bar lingo 0 stands for an unknown number of steps.
+            return 0;
+        }
+
+        // TODO in PHP 7.4 use [...$subjects];
+        $subjects = take($subjects)->toArray();
+
+        return count($subjects);
+    }
```
sanmai: Best-case we're dealing with an iterator like `\Generator` here. Then `toArray()` will call `iterator_to_array` and be done. Worst-case are arrays, but they're mostly contained to the tests, where extra few copies of an array isn't a reason to be concerned.

theofidry: I think it's fine. I'll leave to you to do two runs (and same for regular tests) to compare the memory impact, but if it's like 1MB not even I don't think we care.

src/Process/Runner/IterableBuffer.php
theofidry: 👍

src/IterableCounter.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    /**
+     * @param iterable<mixed> $subjects
+     */
+    public static function bufferAndCountIfNeeded(iterable &$subjects, bool $runConcurrently): int
+    {
+        if ($runConcurrently) {
+            // This number is typically fed to ProgressFormatter/ProgressBar or variants.
+            // In progress bar lingo 0 stands for an unknown number of steps.
+            return 0;
```
maks-rafalko: just an idea - probably it would be more clear to name it `UNKNON_COUNT` and use in ProgressFormatter/ProgressBar/etc this constant. Then it's even not necessary to be zero - any value

up to you

theofidry: Since we already nagged a lot in this PR though I think it would be better in a separate PR, especially since it may affect a few more files. But :+1: on a constant name there it makes it clearer

sanmai: That is an excellent idea.

sanmai: Implemented. It might be not the right place. Please tell me what you think.

src/IterableCounter.php
sanmai: This way or another no difference.

maks-rafalko:
Thank you @sanmai

sanmai:

PR #1098, opened 2020-03-02
Introduce SafeDOMXPath, replacement for SafeQuery

sanmai:
- Replaces slightly annoying `self::safeQuery()` with an instance method `$xPath->query()`
- Check for `DOMDocument` return value. (I think I missed one or two, will check again later.)
- `SafeQuery` ~~s kept `@deprecated` (is it used by sub-packages?)~~ removed


src/TestFramework/SafeDOMXPath.php
```diff
@@ -0,0 +1,92 @@
...
+        $success = $document->loadXML($content);
+
+        Assert::true($success);
+
+        return new self($document);
+    }
+
+    /**
+     * @return DOMNodeList|DOMElement[]
+     * @phpstan-return DOMNodeList<DOMElement>
```
theofidry: Is the `phpstan` prefix necessary here? Why not just doing `@return DOMNodeList<DOMElement>`?

sanmai: Good point. Let's see.

src/TestFramework/Coverage/XmlReport/JUnitTestFileDataProvider.php
```diff
@@ -120,7 +121,9 @@ private static function createXPath(string $jUnitPath): SafeDOMXPath
         }
 
         $dom = new DOMDocument();
-        $dom->load($jUnitPath);
+        $success = @$dom->load($jUnitPath);
+
+        Assert::true($success);
```
theofidry: is there any errors we should up here in case of failure? (same in the other places)

src/TestFramework/Coverage/XmlReport/JUnitTestFileDataProvider.php
```diff
@@ -118,15 +114,17 @@ private function getXPath(): DOMXPath
...
-    private static function createXPath(string $jUnitPath): DOMXPath
+    private static function createXPath(string $jUnitPath): SafeDOMXPath
     {
         if (!file_exists($jUnitPath)) {
             throw CoverageDoesNotExistException::forJunit($jUnitPath);
         }
 
         $dom = new DOMDocument();
-        $dom->load($jUnitPath);
+        $success = @$dom->load($jUnitPath);
```
maks-rafalko: what's the case when we need `@` here?

sanmai: It'll emit a warning, stopping us from testing against this and other errors.

src/TestFramework/Coverage/XmlReport/JUnitTestFileDataProvider.php
sanmai: Normally we could expect valid coverage reports, so this is just for a case where we hit something very rare. So I don't think there's enough evidence to warrant a specific error type.

sanmai:
PrettyCI seems to be disabled, but branch protection wants its status. Can't merge anything even if I wanted.

maks-rafalko:
> PrettyCI seems to be disabled, but branch protection wants its status. Can't merge anything even if I wanted.

strange, I disabled it. Will look into it

Thank you @sanmai

sanmai:
Look into settings -> protected branches

PR #1104, opened 2020-03-03
Allow to configure ignore globally

theofidry:
As of now, if you want to add an `ignore` setting, you need to add it on a mutator basis. This however gets seriously tedious, repetitive and hard to keep in sync if you have a lot of mutators listed.

This PR proposes to allow a global ignore config which will be enriched (not overwritten) by each mutators.

- [x] Docs: https://github.com/theofidry/site/pull/1/files

### DIFF
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -112,6 +112,13 @@
             "additionalProperties": false,
             "description": "Contains the settings for different mutations and profiles",
             "properties": {
+                "ignore": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    }
+                },
+
                 "@arithmetic": { "$ref": "#/definitions/default-mutator-config" },
                 "@boolean": { "$ref": "#/definitions/default-mutator-config" },
                 "@conditional_boundary": { "$ref": "#/definitions/default-mutator-config" },
@@ -129,7 +136,9 @@
                 "@cast": { "$ref": "#/definitions/default-mutator-config" },
                 "@unwrap": { "$ref": "#/definitions/default-mutator-config" },
                 "@extensions": { "$ref": "#/definitions/default-mutator-config" },
+
                 "@default": { "$ref": "#/definitions/default-mutator-config" },
+
                 "Assignment": { "$ref": "#/definitions/default-mutator-config" },
                 "AssignmentEqual": { "$ref": "#/definitions/default-mutator-config" },
                 "BitwiseAnd": { "$ref": "#/definitions/default-mutator-config" },
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -112,7 +112,7 @@
             "additionalProperties": false,
             "description": "Contains the settings for different mutations and profiles",
             "properties": {
-                "ignore": {
+                "global-ignore": {
                     "type": "array",
                     "items": {
                         "type": "string"
--- a/src/Mutator/MutatorResolver.php
+++ b/src/Mutator/MutatorResolver.php
@@ -112,7 +112,7 @@ final class MutatorResolver
      * @param mixed[]|bool $settings
      * @param array<string, string[]> $globalSettings
      *
-     * @return array<string, string[]>|bool
+     * @return array<string, mixed[]>|bool
      */
     private static function resolveSettings($settings, array $globalSettings)
     {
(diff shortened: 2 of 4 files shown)


resources/schema.json
```diff
@@ -112,6 +112,13 @@
             "additionalProperties": false,
             "description": "Contains the settings for different mutations and profiles",
             "properties": {
+                "ignore": {
```
sanmai: Something like `ignore-all` can't be a mutator name, doesn't have to be reserved or tested against.

src/Mutator/MutatorResolver.php
```diff
@@ -89,6 +104,25 @@ public function resolve(array $mutatorSettings): array
         return $mutators;
     }
 
+    /**
+     * @param array<string, string>|bool $settings
+     * @param array<string, string> $globalSettings
+     *
+     * @return array<string, string>|bool
+     */
+    private static function resolveSettings($settings, array $globalSettings)
```
sanmai:
```suggestion
    private static function resolveSettings($settings, array $globalSettings): array
```

src/Mutator/MutatorResolver.php
```diff
@@ -166,7 +200,7 @@ private static function registerFromClass(
         if ($settings === false) {
             unset($mutators[$mutatorClassName]);
         } else {
-            $mutators[$mutatorClassName] = $settings === true ? [] : (array) $settings;
+            $mutators[$mutatorClassName] = (array) $settings;
```
sanmai: Cast needed?

tests/phpunit/Mutator/MutatorResolverTest.php
```diff
@@ -62,14 +63,14 @@ protected function setUp(): void
         $this->mutatorResolver = SingletonContainer::getContainer()->getMutatorResolver();
     }
 
-    public function test_it_resolves_no_mutator_if_no_profile_or_mutator_is_passed(): void
+    public function test_it_resolve_no_mutator_if_no_profile_or_mutator_is_passed(): void
```
sanmai:
```suggestion
    public function test_it_resolves_no_mutator_if_no_profile_or_mutator_is_passed(): void
```

src/Mutator/MutatorResolver.php
theofidry: It can be a boolean

src/Mutator/MutatorResolver.php
theofidry: We do a cast in any case, I'm just removing the unnecessary ternary statement

src/Mutator/MutatorResolver.php
sanmai: Aw, right

src/Mutator/MutatorResolver.php
sanmai: Right. It is `false` to `[]` type of cast.

resources/schema.json
theofidry: `global-ignore`?

sanmai: Yes, that'd be fine too. Everything with a dash works like that.

theofidry: went for `global-ignore` since it's more in line with the names in `MutatorResolver`

BackEndTea:
Is it not possible atm to set an `ignore` on the `@default` mutator set?

theofidry:
@BackEndTea it is but the default is not necessarily all the mutators

src/Mutator/MutatorResolver.php
```diff
@@ -155,8 +193,8 @@ private static function registerFromName(
     }
 
     /**
-     * @param array<string, string>|bool|stdClass $settings
-     * @param array<string, array<string, string>> $mutators
```
maks-rafalko: Why was it changed?

maks-rafalko:
Hm.. I know why I hasn't noticed docs PR, its for you fork: https://github.com/theofidry/site/pull/1/files. I've updated first message to include information about docs, please update with `infection/site` url once ready

src/Mutator/MutatorResolver.php
theofidry: It was incorrect :)

theofidry:
_sight_ stupid me, thanks will update the PR tonight

src/Mutator/MutatorResolver.php
theofidry: Hm actually lemme double check that again

PR #1105, opened 2020-03-03
Fix the order in which the infection configuration files are loaded

theofidry:
The expected order is:

- custom file if specified
- `infection.json` file if available
- `infection.json.dist` file

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -337,10 +337,10 @@ final class InfectionCommand extends BaseCommand
     {
         try {
             $locator->locateOneOf([
-                SchemaConfigurationLoader::DEFAULT_DIST_CONFIG_FILE,
                 SchemaConfigurationLoader::DEFAULT_CONFIG_FILE,
+                SchemaConfigurationLoader::DEFAULT_DIST_CONFIG_FILE,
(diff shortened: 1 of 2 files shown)


maks-rafalko:
I guess this is BC break. Nothing bad here as we are on 0.x, but let's mark such issues so that we can list it on the blog post for 0.16 version / changelog

theofidry:
Will check it out locally tonight

theofidry:
Fair

sanmai:
This isn’t a BC break, it’s a bug fix. Hard to BC break something that wasn’t working.

maks-rafalko:
Why so? If you had 2 files (with and without `.dist`) and they had different content, they will be swapped right now, no?

sanmai:
If you had `infection.json` before together with `infection.json.dist` the first was never ever considered.

maks-rafalko:
> If you had `infection.json` before together with `infection.json.dist` the first was never ever considered.

right, and now (after this PR) it will be considered.

sanmai:
Right, means I'm wrong that this isn't a BC break. It is both bug fix and BC break.

maks-rafalko:
Exactly. Thanks ;)

PR #1106, opened 2020-03-03
Use coverage report as a primary source of files to mutate

### LINKED ISSUE #705: Infection fails with "Allowed memory size ... exhausted" for a project with a really big coverage

sanmai:
This PR is a continuation of #1082. Primary goal of this PR is memory optimization. Fixes #705.

With big projects with huge coverage reports Infection still has a lot of difficulty. Major part of this problem stems from the fact that Infection loads virtually all PHPUnit coverage reports at the start, and then goes one by one over source files looking up each file in the reports to understand if it need mutating, where, how, and so on. Not only it is expensive to load all reports at once, but also working with them isn't fast. Basically, it is one big hash table, where Infection tries to access rows from here and from there. 
- Yes, it may worth [adding a cache](https://github.com/infection/infection/pull/1125), because, say, methods of  `XMLLineCodeCoverage` are called sequentially over data for the same file, but this isn't what I want to propose. 
- Yes, with some careful `ArrayAccess` application you can make [the `$coverage` array clever to load data only as needed](https://github.com/sanmai/infection/commit/e83d1511f2b8cf2c233b3eb28dd4ce4182eb5cab#diff-668eec250a6cee30c97758097868c31cR105), and neither this is what I want to propose.

### What's the idea?

If you'd look at the PHPUnit coverage reports, and how we parse them, you would notice that the coverage reports have the same file names we need. Therefore, instead of iterating over files on disk, we can iterate over coverage reports, parsing them one by one, and discarding them once we finish mutating a file.

There are some problems:

- We not only consider these reports, we also consider JUnit reports, joining them together. This procedure had to be split into two. First we load coverage files, select files we want to be mutated (covered, uncovered, filtered), and only then we add JUnit timings, proceeding with mutation.
- For BC reasons we have to consider files outside of coverage reports, therefore we collect them too, adding missing files at the end, but only if needed. Suffice to say if we'd went half-way towards proposal #1064, this part could be removed from our pipeline, chances are together with the configuration step where user enters source paths.

Overall I had to reshuffle a lot of things, while trying to not to rename them too much. E.g. old tests for the new parser should probably work, and so on.

### Numbers
Enough words, let's the numbers speak:

- [BlackFire diff for a partial Psalm run](https://blackfire.io/profiles/compare/87b519d7-db00-44b0-b507-a68c052e3331/graph)
- [BlackFire diff for a Psalm run with `--filter=src/Psalm/DocComment.php`](https://blackfire.io/profiles/compare/33bb5fb7-93a3-4ba3-8f24-588e904b2703/graph)

```diff
-Time: 38m 17s. Memory: 2.56GB
+Time: 31m 2s. Memory: 553.82MB
```
Isn't that nice for a change?

If you want to try it, try this with something big, like Psalm. If you like me don't have all the time in the world, note the setup instructions below. PCOV is a must there, otherwise:

```
php vendor/bin/phpunit --coverage-xml=build/logs/coverage-xml --log-junit=build/logs/junit.xml --stop-on-failure
php path/to/bin/infection -j8 --skip-initial-tests --coverage=build/logs --no-progress --only-covered
```
Notice how fast it goes about business. Try with the master version, also with `--no-progress`. 

<details>
  <summary>How to set up Psalm</summary>

- With Psalm you need to either use a .phar, or update `vendor/ocramius/package-versions/src/PackageVersions/Versions.php` (inside Psalm checkout) to return something instead of `\OutOfBoundsException`.

- For simplicity sake best to remove some integration tests:

    ` rm -f tests/TypeReconciliation/ConditionalTest.php tests/DocumentationTest.php tests/UnusedVariableTest.php tests/FunctionCallTest.php tests/TypeReconciliation/RedundantConditionTest.php tests/ReturnTypeTest.php tests/ConstantTest.php tests/ClassStringTest.php tests/MethodCallTest.php tests/AssertAnnotationTest.php tests/TraitTest.php tests/ValueTest.php tests/AnnotationTest.php tests/CallableTest.php tests/Template/FunctionTemplateTest.php tests/FileManipulationReturnTypeManipulationTest.php tests/UnusedVariableTest.php tests/TypeReconciliation/RedundantConditionTest.php tests/Loop/ForeachTest.php tests/TypeReconciliation/TypeTest.php tests/PropertyTypeTest.php tests/TypeReconciliation/TypeAlgebraTest.php tests/IssetTest.php tests/Template/ClassTemplateTest.php tests/ArrayAssignmentTest.php tests/PropertyTypeTest.php tests/ArrayFunctionCallTest.php tests/Template/ClassTemplateExtendsTest.php tests/FunctionCallTest.php tests/UnusedVariableTest.php tests/DocumentationTest.php tests/TypeReconciliation/ConditionalTest.php tests/Config/ConfigTest.php`

- I used the following `infection.json`:
```json
{
    "timeout": 50,
    "source": {
        "directories": [
            "src"
        ]
    },
    "logs": {
        "text": "build\/infection.log"
    },
    "mutators": {
        "@removal": true,
        "@return_value": true,
        "@zero_iteration": true
    }
}
```

- In Linux try running Infection with `systemd-run --scope -p MemoryLimit=8000M php ...` to limit the fallout in case a mutation triggers a nasty bug in one of the mutation processes.

</details>

### What's next?

First of all, this is a practical proposal. I'd like to see that Infection still works, and my limited testing shows it pretty much does. Let me be corrected, though, if you find it doesn't.

Second, I'm looking to see a consensus about this proposal, and if there's a consensus that this is good and proper, I'll proceed with the remaining tasks. Getting tests up to speed, etc.

If it'd get rejected, well, then at least I didn't spend time fixing tests for a feature nobody needs, so please bear with me.

#### Task list

- [x] It works
- [x] Profile it
- [x] Write explanation
- [x] Fix PHPStan warnings for `src/`
- [x] Fix old tests, cleanup, rinse and repeat
    - [x] Tests from #1131 
    - [x] Tests from #1125 
- [x] Add new tests
    - [x] tests/phpunit/FileSystem/SourceFileFilterTest.php                              
    - [x] tests/phpunit/TestFramework/Coverage/CoveredFileDataFactoryTest.php            
    - [x] tests/phpunit/TestFramework/Coverage/CoveredFileDataTest.php                   
    - [x] tests/phpunit/TestFramework/Coverage/CoveredFileNameFilterTest.php             
    - [x] tests/phpunit/TestFramework/Coverage/JUnit/TestFileDataAdderTest.php           
    - [x] tests/phpunit/TestFramework/Coverage/XmlReport/FileCodeCoverageProviderTest.php
    - [x] tests/phpunit/TestFramework/PhpUnit/Coverage/XmlCoverageParserTest.php     
- [x] Fix coverage issues, cleanup, rinse and repeat
- [x] Remove `@see`
- [x] removeNamespace() should be moved in a XPathFactory
- [x] Doc PR: not needed
- [x] Remove CoveredFileNameFilter
- [ ] rename LineCodeCoverage to Tracer?
- [ ] CoveredFileData could be named Trace?

Post-merge clean up:
- [ ] FileCodeCoverageProvider no longer needed, but heavily used in tests
- [ ] FileCodeCoverage and SourceFileData might be possible to merge

#### Review tips

- Start with MutationGenerator
- Next you'll see SourceFileDataFactory
- It feeds from PhpUnitXmlCoveredFileDataProvider, which is a thin wrapper around IndexXmlCoverageParser
- SourceFileDataFactory also uses SourceFileFilter, and JUnitTestExecutionInfoAdder
- And a feed of files from configuration.


sanmai:
ping @infection/core 

If you could try this build on anything, even on Infection's master, to see if there are obvious UX/DX issues, which shouldn't be, I’d really appreciate it. 

There's a lot more to do, I didn't even start to review what I've done, and if I could know about any new UX issues, e.g. when something is obviously broken or works unexpectedly, this will be a big help.

theofidry:
@sanmai any way to get a blackfire diff?

sanmai:
@theofidry Here's one: https://blackfire.io/profiles/compare/7b0142d7-b38d-4e98-b00f-444a9daa58fe/graph

sanmai:
Here's another one for Psalm: https://blackfire.io/profiles/compare/fc62b217-2368-422d-b178-4527486ea23c/graph

No doubt there's a lot to improve here speed-wise, but look at the peak memory change.

src/Configuration/ConfigurationFactory.php
```diff
@@ -122,14 +123,21 @@ public function create(
...
+        );
+
         return new Configuration(
             $schema->getTimeout() ?? self::DEFAULT_TIMEOUT,
             $schema->getSource()->getDirectories(),
             $this->sourceFileCollector->collectFiles(
                 $schema->getSource()->getDirectories(),
                 $schema->getSource()->getExcludes(),
-                $filter
+                $sourceFileFilter->getFilterArray(),
```
sanmai:
```suggestion
                $sourceFileFilter->getFilterArray()
```

src/Container.php
```diff
@@ -154,11 +157,33 @@ public static function create(): self
...
+            TestFileDataAdder::class => static function (self $container): TestFileDataAdder {
+                return new TestFileDataAdder(
+                    $container->getTestFrameworkAdapter(),
+                    $container->getMemoizedTestFileDataProvider()
+                );
+            },
             XMLLineCodeCoverageFactory::class => static function (self $container): XMLLineCodeCoverageFactory {
                 return new XMLLineCodeCoverageFactory(
                     $container->getConfiguration()->getCoveragePath(),
                     $container->getIndexXmlCoverageParser(),
```
sanmai:
```suggestion
                    $container->getIndexXmlCoverageParser()
```

src/FileSystem/SourceFileFilter.php
```diff
@@ -0,0 +1,119 @@
...
+        array $sourceDirectories,
+        array $excludeDirectories,
+        string $filter
+    ) {
+        $this->sourceDirectories = $sourceDirectories;
+        $this->excludeDirectories = $excludeDirectories;
+        $this->filters = take(explode(',', $filter))
+            ->map('trim')
+            ->filter()
+            ->toArray();
```
theofidry:
```suggestion
        $this->filters = array_filter(array_map(
            'trim',
            explode(',', $filter)
        ));
```

No need for a fancy wrapper here although I agree it's more elegant, I prefer to reduce the usage of it (like I do with `nikic/iter` when I use it) as it's still an increase of learning curve and of coupling

src/FileSystem/SourceFileFilter.php
```diff
@@ -0,0 +1,119 @@
...
+            $this->iterableToIterator($input),
+            $this->filters,
+            []
+        );
+    }
+
+    /**
+     * @param iterable<SplFileInfo&CoveredFileData> $input
+     *
+     * @return \Iterator<SplFileInfo&CoveredFileData>
```
theofidry:
```suggestion
     * @return Iterator<SplFileInfo&CoveredFileData>
```
(it has already been imported)

src/Mutation/Mutation.php
```diff
@@ -80,7 +80,7 @@ class Mutation
     /**
      * @param Node[] $originalFileAst
      * @param array<string|int|float> $attributes
-     * @param CoverageLineData[] $tests
+     * @param array<CoverageLineData> $tests
```
theofidry: why that change here?

src/Process/Runner/Parallel/ParallelProcessRunner.php
```diff
@@ -52,14 +52,6 @@
  */
 class ParallelProcessRunner
```
theofidry: Those changes seem unrelated no? Would it be possible it to extract it in a dedicated PR to evaluate its impact there?

src/TestFramework/Coverage/CoveredFileData.php
```diff
@@ -0,0 +1,138 @@
...
+     * @var SplFileInfo
+     */
+    private $sourceFile;
+
+    /**
+     * @var CoverageFileData|null
+     */
+    private $coverageFileData;
+
+    /** @var iterable<CoverageFileData> */
```
theofidry: can you keep them in the multi-line form? It especially don't make sense when you use the other form just above :)

I wish there was a fixer for that one as well

src/TestFramework/Coverage/CoveredFileData.php
```diff
@@ -0,0 +1,138 @@
...
+    public function getCoverageFileData(): CoverageFileData
+    {
+        if ($this->coverageFileData !== null) {
+            return $this->coverageFileData;
+        }
+
+        foreach ($this->lazyCoverageFileData as $coverageFileData) {
+            $this->coverageFileData = $coverageFileData;
+
+            break;
```
theofidry: this condition looks really weird: it's as if you're iterating only once

src/TestFramework/Coverage/CoveredFileDataFactory.php
```diff
@@ -0,0 +1,128 @@
...
+namespace Infection\TestFramework\Coverage;
+
+use Infection\Mutation\MutationGenerator;
+use Infection\TestFramework\Coverage\JUnit\TestFileDataAdder;
+use Infection\TestFramework\Coverage\XmlReport\PhpUnitXmlCoverageFactory;
+use Symfony\Component\Finder\SplFileInfo;
+use Webmozart\Assert\Assert;
+
+/**
+ * Assembles a ready feed of CoveredFileData from different sources. Feeds data into MutationGenerator. Does not known about differences between adapters and what not.
```
theofidry: Can we get some line returns to say not exceed a 100 lines limit here?

src/TestFramework/Coverage/CoveredFileDataFactory.php
```diff
@@ -0,0 +1,128 @@
...
+     */
+    public function createCoverage(): iterable
+    {
+        $coverageFeed = $this->primaryCoverageProvider->createCoverage();
+
+        $filteredCoverageFeed = $this->filter->filter($coverageFeed);
+
+        $readyCoverageFeed = $this->testFileDataAdder->addTestExecutionInfo($filteredCoverageFeed);
+
+        if (is_array($this->sourceFiles) && $this->sourceFiles === []) {
```
theofidry: missing import for `is_array()`

src/TestFramework/Coverage/CoveredFileDataFactory.php
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        // Since these are sorted sets, there should be a way to optimize.
+
+        foreach ($this->sourceFiles as $splFileInfo) {
+            $sourceFilePath = $splFileInfo->getRealPath();
+
+            Assert::string($sourceFilePath);
+
+            if (array_key_exists($sourceFilePath, $seenFiles)) {
```
theofidry: missing import for `array_key_exists()`

src/TestFramework/Coverage/CoveredFileDataFactory.php
```diff
@@ -0,0 +1,128 @@
...
+        $seenFiles = [];
+
+        /** @var CoveredFileData $data */
+        foreach ($coverage as $data) {
+            $seenFiles[$data->getSplFileInfo()->getRealPath()] = true;
+
+            yield $data;
+        }
+
+        // Since these are sorted sets, there should be a way to optimize.
```
theofidry: I'm a bit confused at the comment

src/TestFramework/Coverage/CoveredFileDataFactory.php
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    /**
+     * @param iterable<CoveredFileData> $coverage
+     *
+     * @return iterable<CoveredFileData>
+     */
+    private function appendUncoveredFiles(iterable $coverage): iterable
+    {
+        $seenFiles = [];
```
theofidry: `$checkedFiles`?

src/TestFramework/Coverage/CoveredFileDataProvider.php
```diff
@@ -0,0 +1,49 @@
...
+ * @internal
+ */
+interface CoveredFileDataProvider
+{
+    /**
+     * @throws CoverageDoesNotExistException
+     *
+     * @return iterable<CoveredFileData>
+     */
+    public function createCoverage(): iterable;
```
theofidry: `provideFiles()`?

src/TestFramework/Coverage/CoveredFileNameFilter.php
```diff
@@ -0,0 +1,64 @@
...
+ * @see CoveredFileDataFactory
+ *
+ * @internal
+ */
+final class CoveredFileNameFilter
+{
+    private $filter;
+
+    public function __construct(
+        SourceFileFilter $filter
```
theofidry: no need for a multi-line constructor here

src/TestFramework/Coverage/JUnit/MemoizedTestFileDataProvider.php
```diff
@@ -33,11 +33,13 @@
 
 declare(strict_types=1);
 
-namespace Infection\TestFramework\Coverage\XmlReport;
+namespace Infection\TestFramework\Coverage\JUnit;
```
theofidry: AFAICS those changes of namespaces could be done in a separate PR for a bit of clarity

src/TestFramework/Coverage/JUnit/MemoizedTestFileDataProvider.php
```diff
@@ -33,11 +33,13 @@
...
 
 use function array_key_exists;
 
 /**
+ * Caches parsed rows from JUnit report.
```
theofidry: Is that comment necessary? That is what memoizing is about

src/TestFramework/Coverage/XmlReport/PhpUnitXmlCoverageFactory.php
```diff
@@ -36,19 +36,22 @@
...
 /**
+ * Source of primary coverage data. Used by CoveredFileDataFactory.
+ *
+ * @see CoveredFileDataFactory
+ *
  * @internal
  * @final
  */
-class PhpUnitXmlCoverageFactory
+class PhpUnitXmlCoverageFactory implements CoveredFileDataProvider
```
theofidry: `PhpUnitXmlCoveredFileDataFactory`?

src/TestFramework/Coverage/XmlReport/XMLLineCodeCoverage.php
```diff
@@ -36,131 +36,99 @@
...
 use Infection\TestFramework\Coverage\LineCodeCoverage;
 use Infection\TestFramework\Coverage\NodeLineRangeData;
-use function range;
 
 /**
+ * Used by CoveredFileData to implement LineCodeCoverage methods.
+ *
  * @internal
+ *
+ * @see CoveredFileData
```
theofidry: Do you think those `@see` annotations are really necessary? I find it a bit confusing, if I want to see the usage I can look up for the usage, at least it will be guaranteed to be up to date data whereas those `@see` are unlikely to be kept up to date

src/TestFramework/Coverage/XmlReport/XMLLineCodeCoverageFactory.php
```diff
@@ -36,47 +36,39 @@
...
+        IndexXmlCoverageParser $coverageXmlParser
     ) {
         $this->coverageDir = $coverageDir;
         $this->coverageXmlParser = $coverageXmlParser;
-        $this->testFileDataProvider = $testFileDataProvider;
     }
 
     public function create(
         string $testFrameworkKey,
         TestFrameworkAdapter $adapter
```
theofidry: The parameter `$adapter` is not used and could be removed.

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
```diff
@@ -74,30 +65,33 @@ public function __construct(string $coverageDir)
...
+            ->map(static function (CoveredFileData $data) {
+                yield $data->getSplFileInfo()->getRealPath() => $data->getCoverageFileData();
+            });
 
-        return $data;
+        return iterator_to_array($coverage, true);
     }
 
-    private static function createXPath(string $coverageContent): SafeDOMXPath
+    public static function createXPath(string $coverageContent): SafeDOMXPath
```
theofidry: Maybe this piece and its `removeNamespace()` should be moved in a `XPathFactory` utility class? (in the same namespace)

src/TestFramework/PhpUnit/Coverage/XmlCoverageParser.php
```diff
@@ -0,0 +1,243 @@
...
+            $this->coverageDir . '/' . $this->relativeCoverageFilePath
+        ));
+
+        $sourceFileInfo = self::retrieveSourceFileInfo($xPath, $this->relativeCoverageFilePath, $this->projectSource);
+
+        return new CoveredFileData($sourceFileInfo, $this->getCoverageFileData($xPath));
+    }
+
+    /** @return iterable<CoverageFileData> */
+    private function getCoverageFileData(SafeDOMXPath $xPath): iterable
```
theofidry: `retrieveCoverageFileData`?

src/Process/Runner/Parallel/ParallelProcessRunner.php
sanmai: Yes, let's try this.

src/TestFramework/Coverage/CoveredFileData.php
sanmai: You know, typing fast trying to get things done. I wish there was a fixer. Anyway, thanks for the reminder, I'll get these in order in the end.

src/TestFramework/Coverage/CoveredFileData.php
sanmai: Exactly. `lazyCoverageFileData` is an iterator with a single element.

src/TestFramework/Coverage/CoveredFileDataFactory.php
sanmai: No prob.

src/TestFramework/Coverage/CoveredFileDataFactory.php
sanmai: Say you have a sequence of files from disk:

    A B C D E F

And have a sequence of covered files:

    A B C _ E F

See, here D is missing e.g. because it is excluded from the whitelist. We can detect these gaps as we go. But we need to verify that these are actually are sorted sets. So it's out of scope.

src/TestFramework/Coverage/JUnit/MemoizedTestFileDataProvider.php
sanmai: See, I didn't know that when I wrote this comment. And it wasn't obvious it is related to JUnit report. Can we keep it?

Or, say, rename the class so it'd be more obvious what it does.

src/TestFramework/Coverage/XmlReport/XMLLineCodeCoverage.php
sanmai: Well, they really help be navigate the code, find what comes with what. I will certainly keep them for now, but later they are easy to remove once we're done here.

sanmai:
@theofidry thanks for the review! I will address remaining comments in code a bit later.

theofidry:
>See, I didn't know that when I wrote this comment. And it wasn't obvious it is related to JUnit report. Can we keep it?

 I guess then

src/TestFramework/Coverage/JUnit/MemoizedTestFileDataProvider.php
sanmai: Sure, #1131

src/Mutation/Mutation.php
sanmai:
```suggestion
     * @param CoverageLineData[] $tests
```

src/Mutator/NodeMutationGenerator.php
```diff
@@ -134,6 +134,8 @@ private function generateForMutator(Node $node, Mutator $mutator): iterable
 
         $mutationByMutatorIndex = 0;
 
+        Assert::fileExists($this->filePath);
+
```
sanmai:
```suggestion
```

src/TestFramework/Coverage/CoveredFileData.php
```diff
@@ -0,0 +1,138 @@
...
+     * @see TestFileDataAdder
+     */
+    public function retrieveCoverageFileData(): CoverageFileData
+    {
+        if ($this->coverageFileData !== null) {
+            return $this->coverageFileData;
+        }
+
+        foreach ($this->lazyCoverageFileData as $coverageFileData) {
+            $this->coverageFileData = $coverageFileData;
```
sanmai:
```suggestion
            $this->coverageFileData = $coverageFileData; // is a \Generator with one yield, thus it'll only trigger here
```

src/TestFramework/Coverage/CoveredFileData.php
```diff
@@ -0,0 +1,138 @@
...
+     * @see TestFileDataAdder
+     */
+    public function retrieveCoverageFileData(): CoverageFileData
+    {
+        if ($this->coverageFileData !== null) {
+            return $this->coverageFileData;
+        }
+
+        foreach ($this->lazyCoverageFileData as $coverageFileData) {
+            $this->coverageFileData = $coverageFileData; // is a \Generator with one yield, thus it'll only trigger here
```
sanmai:
```suggestion
            $this->coverageFileData = $coverageFileData; // is a Generator with one yield, thus it'll only trigger here (or an array with one element)
```

src/TestFramework/Coverage/JUnit/MemoizedTestFileDataProvider.php
```diff
@@ -38,6 +38,8 @@
...
+ *
```
sanmai:
```suggestion
```

src/Mutation/MutationGenerator.php
```diff
@@ -99,15 +93,18 @@ public function __construct(
...
 
-        foreach ($this->sourceFiles as $fileInfo) {
+        /** @var CoveredFileData $fileData */
+        foreach ($coveredFiles as $fileData) {
             yield from $this->fileMutationGenerator->generate(
-                $fileInfo,
+                $fileData,
                 $onlyCovered,
-                $this->coverageProvider->createFor($fileInfo),
+                new FileCodeCoverage($fileData->retrieveCoverageFileData()),
```
sanmai: This isn't good.

theofidry:
@sanmai just in case: I'll be waiting for you to ping me again before doing another review

sanmai:
@theofidry sure, no prob. But if you can see if there's a method or a class that isn't aptly named, I'd appreciate if you could do that before a full review. Thanks.

src/FileSystem/SourceFileFilter.php
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * @return iterable<SplFileInfo>
+     */
+    public function __construct(string $filter)
+    {
+        $this->filters = array_filter(array_map(
+            'trim',
+            explode(',', $filter)
+        ));
```
sanmai: I understand one can argue that with a single combo of `array_filter(array_map())` it may not make sense to use any wrappers, even if the wrapper does the same `array_filter` and `array_map`. Conversely, I will tend to disagree if we're talking several `array_*` calls in one place. They're a DX disaster, and parsing them is not what one wants to spend their invaluable minutes on, at least as I see it.

To make a point against learning curve and vendor lock-in arguments, I invite you as a contributor. Let's see if I can change your mind, and if not from the start, at least we might learn if there's anything more to be done about that.

sanmai:
@theofidry can you see if I missed anything from #1107? I know I am missing several tests, but in regard of `src/` I'm not entirely sure.

theofidry:
Ok will try to have a look tonight or tomorrow

sanmai:
@theofidry @maks-rafalko Build is green.

src/TestFramework/Coverage/CoveredFileDataFactory.php
sanmai: Well, we don't do any checking with them.

src/FileSystem/SourceFileFilter.php
```diff
@@ -0,0 +1,106 @@
...
+     * @return Iterator<SplFileInfo&CoveredFileData>
+     */
+    private function iterableToIterator(iterable $input)
+    {
+        if ($input instanceof Iterator) {
+            // Generator is an iterator, means most of the time we're done right here.
+            return $input;
+        }
+
+        // Clause for all other cases, e.g. when testing
```
maks-rafalko: just to understand this bit of the code, could you please explain what it will be used for? Any example?

src/TestFramework/Coverage/CoveredFileData.php
```diff
@@ -0,0 +1,138 @@
...
+     */
+    private $coverageFileData;
+
+    /**
+     * @var iterable<CoverageFileData>
+     */
+    private $lazyCoverageFileData;
+
+    /**
+     * @var ?FileCodeCoverage
```
maks-rafalko: makes sense to use only one way of documenting nullable types: `?Type` or `Type|null`.

For example, line 56 uses `Type|null`

src/TestFramework/Coverage/JUnit/TestFileDataAdder.php
```diff
@@ -33,71 +33,63 @@
...
     }
 
     /**
-     * @param CoverageFileData[] $coverage
+     * @param iterable<CoveredFileData> $coverage
+     *
+     * @return iterable<CoveredFileData>
      */
-    private function addTestExecutionInfo(array $coverage): void
+    private function testExecutionInfoAdder(iterable $coverage): iterable
```
maks-rafalko: not sure about the name of the method. IMO, previous version `addTestExecutionInfo()` is more self explained.

The new name does not mean we return or add anything.

sanmai: I'll bring it back.

src/TestFramework/Coverage/CoveredFileData.php
sanmai:
```suggestion
     * @var FileCodeCoverage|null
```
Fair.

src/TestFramework/Coverage/JUnit/TestFileDataAdder.php
sanmai: `addTestExecutionInfo` is still there, a public method line 69 above. So we can have this renamed, but not into `addTestExecutionInfo`

src/TestFramework/Coverage/CoveredFileData.php
sanmai: Fixed this and in another place.

src/Mutation/MutationGenerator.php
```diff
@@ -99,15 +96,18 @@ public function __construct(
      */
     public function generate(bool $onlyCovered, array $nodeIgnorers): iterable
     {
-        $numberOfFiles = IterableCounter::bufferAndCountIfNeeded($this->sourceFiles, $this->runConcurrently);
+        $coveredFiles = $this->coveredFileDataProvider->provideFiles();
```
maks-rafalko: It's a little bit confusing that variable and class named `covered...`, but actually we can provide uncovered files here as well:

```php
# CoveredFileDataFactory

// ...

if ($this->onlyCovered === true) {
    // The case where only covered files are considered
    return $readyCoverageFeed;
}

return $this->appendUncoveredFiles($readyCoverageFeed);
              ^^^
```

I would suggest to rename it to something that covers both covered an uncovered cases.

src/TestFramework/Coverage/CoveredFileData.php
```diff
@@ -0,0 +1,138 @@
...
+use Symfony\Component\Finder\SplFileInfo;
+use Webmozart\Assert\Assert;
+
+/**
+ * Workhorse AKA envelope for all things coverage-related in regard of mutators.
+ *
+ * @internal
+ * @final
+ */
+class CoveredFileData implements LineCodeCoverage
```
maks-rafalko: New file `CoveredFileData` has a very similar name to already existing `CoverageFileData`, I even mixed them up during code review several times. Could we think about renaming it/them if possible?

src/TestFramework/Coverage/CoveredFileNameFilter.php
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Infection\TestFramework\Coverage;
+
+use Infection\FileSystem\SourceFileFilter;
+
+/**
+ * @internal
+ * @final
+ */
+class CoveredFileNameFilter
```
maks-rafalko: Do we need this file? Or is it for better/easier testing?

Looks redundant for me

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
```diff
@@ -74,50 +59,35 @@ public function __construct(string $coverageDir)
...
-     * Remove namespace to work with xPath without a headache
-     */
-    private static function removeNamespace(string $xml): string
-    {
-        /** @var string $cleanedXml */
-        $cleanedXml = preg_replace('/xmlns=\".*?\"/', '', $xml);
+            $fileInfoProvider = new SourceFileInfoProvider($this->coverageDir, $relativeCoverageFilePath, $projectSource);
 
-        Assert::string($cleanedXml);
+            $parser = new XmlCoverageParser($fileInfoProvider);
```
maks-rafalko: What do you think about creating this class 1 time before the loop and pass `$fileInfoProvider` to the `parse()` method?

maks-rafalko:
Also, question. Can we somehow make sure that all those generators are work as developed and not "unbuffered" by mistake in the future? Some kind of automatic test. This is out of the scope of this PR but something would be nice to have.

What I would like to avoid, is a case when all this hard work and benefits of using generators will somehow be "break" by mistake / not careful usage of them.

maks-rafalko:
`bin/infection --only-covered --no-progress -j4`

0.15.3:

```
1984 mutations were generated:
Covered Code MSI: 87%
Time: 4m 29s. Memory: 92.01MB
```

master:

```
2030 mutations were generated:
Covered Code MSI: 86%
Time: 2m 58s. Memory: 40.00MB
```

I have to say - these are really interesting results.

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
sanmai: It is really hard to see it from the diff here, but $fileInfoProvider is also [called once per node](https://github.com/infection/infection/blob/bf44abc1ed5c94998ef7d7490c443d30f6069496/src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php#L82-L89).

tests/phpunit/FileSystem/SourceFileFilterTest.php
```diff
@@ -0,0 +1,140 @@
...
+        $actual = take($actual)
+            ->map(static function (SplFileInfo $fileInfo) {
+                return $fileInfo->getRealPath();
+            })
+            ->toArray();
+
+        $this->assertSame($expected, $actual);
+    }
+
+    private function arrayToSplFileInfoTraversable(array $input): Traversable
```
theofidry: Maybe I misunderstood but isn't the downstream API requiring an `iterable` here in which case we could just do a simple `array_map` and return an `array` instead?

src/Mutation/MutationGenerator.php
sanmai: Yes, true, I probably left these named this way after `CoveredFileData` objects, which, as I see it, should have a cousin `UnoveredFileData` following the same interface, but with slightly simplified internals. 

How does that feel if we'd name them `SourceFileDataSomething`?

src/TestFramework/Coverage/CoveredFileData.php
sanmai: How about `CoverageReportData` instead of `CoverageFileData`?

src/TestFramework/Coverage/CoveredFileNameFilter.php
sanmai: Let me see about removing it.

src/Mutation/FileMutationGenerator.php
```diff
@@ -85,21 +85,18 @@ public function generate(
...
-        $initialStatements = $this->parser->parse($fileInfo);
+        $sourceFilePath = $fileData->getSplFileInfo()->getPathname();
+
+        $initialStatements = $this->parser->parse($fileData->getSplFileInfo());
 
         $mutationsCollectorVisitor = new MutationsCollectorVisitor(
             new NodeMutationGenerator(
                 $mutators,
-                $filePath,
+                $sourceFilePath,
```
theofidry: Minor change to avoid using `getSplFileInfo()` two times

```suggestion
        $fileInfo = $fileData->getSplFileInfo();

        $initialStatements = $this->parser->parse($fileInfo);

        $mutationsCollectorVisitor = new MutationsCollectorVisitor(
            new NodeMutationGenerator(
                $mutators,
                $fileInfo->getPathname(),
                $initialStatements,
                $codeCoverage,
                $onlyCovered,
                $this->lineRangeCalculator
            )
        );
```

src/Mutation/MutationGenerator.php
theofidry: In this case here we can name `$coveredFiles` to `$files` though

sanmai:
> Can we somehow make sure that all those generators are work as developed and not "unbuffered" by mistake in the future?

This is a valid concern. Even if I made best effort to protect such cases by testing with a "never" iterator, I didn't do without instrumentation to guide my efforts. E.g. I had to know that I'm moving in the right direction. 

This metric I was optimising for was time-to-first-mutation. It was hacked together, to be honest, and not worthy present here, so it isn't, but it is not to say it can't be added in a clean way. I will make an issue of it, it this metric appears to make sense.

Another issue we might have is that these "never" iterators are not exactly obvious. You have to know what you're looking at.

```php
$mutations = $this->createMock(Iterator::class);
$mutations
    ->expects($this->never())
    ->method($this->anything())
 ;
```
It may be worth making these cases to use a consistent and visible approach, say, using a concrete `NeverIterator` class fixture.

src/TestFramework/Coverage/CoveredFileData.php
theofidry: Maybe we can do that change in another PR but I would like to rename `LineCodeCoverage` to `Tracer` which is a more ubiquitous term for this concept of _tracing_ the source code to the tests or vice-versa. Hence `CoveredFileData` could be named `Trace here`

src/TestFramework/Coverage/CoveredFileData.php
```diff
@@ -0,0 +1,138 @@
...
+        // There's no point to have it parsed right away as we may not need it, e.g. because of a filter
+        $this->lazyCoverageFileData = $lazyCoverageFileData;
+    }
+
+    public function getSplFileInfo(): SplFileInfo
+    {
+        return $this->sourceFile;
+    }
+
+    public function getRealPath(): string
```
theofidry: is it worth having considering we are using it only in tests?

src/TestFramework/Coverage/CoveredFileData.php
```diff
@@ -0,0 +1,138 @@
...
+
+        Assert::string($realPath);
+
+        return $realPath;
+    }
+
+    /**
+     * Accessor used to update CoverageFileData with TestFileTimeData.
+     */
+    public function retrieveCoverageFileData(): CoverageFileData
```
theofidry: I'm not quite sure about that `FileCodeCoverage <-> CoveredFileData` relationship/API, will get back to it later

src/TestFramework/Coverage/CoveredFileDataFactory.php
theofidry: `$filesSeen`?

src/TestFramework/Coverage/JUnit/TestFileDataAdder.php
```diff
@@ -33,71 +33,63 @@
 
 declare(strict_types=1);
 
-namespace Infection\TestFramework\Coverage\XmlReport;
+namespace Infection\TestFramework\Coverage\JUnit;
```
theofidry: I thought that move was done in another PR?

src/TestFramework/Coverage/JUnit/TestFileDataAdder.php
theofidry: I guess it can be inlined here though since the name is not that compelling?

src/TestFramework/Coverage/JUnit/TestFileDataAdder.php
```diff
@@ -33,71 +33,63 @@
...
+use Infection\TestFramework\Coverage\CoveredFileData;
 
 /**
+ * Adds test execution info to selected covered file data object.
+ *
  * @internal
  * @final
  */
-class PhpUnitXmlCoverageFactory
+class TestFileDataAdder
```
theofidry: `JunitTestExecutionInfoAdder`
?

src/TestFramework/Coverage/XmlReport/FileCodeCoverageProvider.php
```diff
@@ -35,43 +35,16 @@
...
-
-    private $coverageFactory;
-
-    public function __construct(PhpUnitXmlCoverageFactory $coverageFactory)
-    {
-        $this->coverageFactory = $coverageFactory;
-    }
-
-    public function provideFor(SplFileInfo $fileInfo): FileCodeCoverage
+    public function provideFor(CoveredFileData $fileData): FileCodeCoverage
```
theofidry: Is that provider still really needed?

src/TestFramework/Coverage/XmlReport/FileCodeCoverageProvider.php
```diff
@@ -35,43 +35,16 @@
...
-            ? $fileInfo->getPathname()
-            : $fileInfo->getRealPath()
-        ;
-
-        if (!array_key_exists($filePath, $this->coverage)) {
-            return new FileCodeCoverage(new CoverageFileData());
-        }
-
-        return new FileCodeCoverage($this->coverage[$filePath]);
+        return new FileCodeCoverage($fileData->retrieveCoverageFileData());
```
theofidry: I'm not quite sure to understand why we need both `FileCodeCoverage` and `CoveredFileData`

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
```diff
@@ -74,50 +59,35 @@ public function __construct(string $coverageDir)
...
+            $relativeCoverageFilePath = $node->getAttribute('href');
 
-    /**
-     * Remove namespace to work with xPath without a headache
-     */
-    private static function removeNamespace(string $xml): string
-    {
-        /** @var string $cleanedXml */
-        $cleanedXml = preg_replace('/xmlns=\".*?\"/', '', $xml);
+            $fileInfoProvider = new SourceFileInfoProvider($this->coverageDir, $relativeCoverageFilePath, $projectSource);
```
theofidry:
```suggestion
            $fileInfoProvider = new SourceFileInfoProvider(
                $this->coverageDir, 
                $relativeCoverageFilePath, 
                $projectSource
            );
```

src/TestFramework/PhpUnit/Coverage/XPathFactory.php
```diff
@@ -0,0 +1,74 @@
...
+
+use DOMDocument;
+use Infection\TestFramework\SafeDOMXPath;
+use function Safe\preg_replace;
+use Webmozart\Assert\Assert;
+
+/**
+ * @internal
+ */
+final class XPathFactory
```
theofidry: can that be introduced in a separate PR?

src/TestFramework/Coverage/JUnit/TestFileDataAdder.php
sanmai: It did. It is Git thinks this file has something to do with PhpUnitXmlCoverageFactory

src/TestFramework/Coverage/JUnit/TestFileDataAdder.php
sanmai: `addTestExecutionInfo` isn't a generator, where this method is. So, no, can't be inlined, at least without a closure.

sanmai: OK, I made it an explicit generator. It as actually adding generator. Makes sense?

src/FileSystem/SourceFileFilter.php
```diff
@@ -0,0 +1,115 @@
...
+        $this->filters = array_filter(array_map(
+            'trim',
+            explode(',', $filter)
+        ));
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getFilters(): array
```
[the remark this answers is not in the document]
sanmai: Well, I missed this. I guess it really makes so.

src/TestFramework/Coverage/CoveredFileDataFactory.php
sanmai: Done

src/TestFramework/Coverage/JUnit/TestFileDataAdder.php
sanmai: Done

src/TestFramework/Coverage/XmlReport/FileCodeCoverageProvider.php
sanmai: Yes and no. The whole purpose of this file is to save my time on conflict resolution. I guess I can add it to some post-merge clean up list, how do you think?

tests/phpunit/FileSystem/SourceFileFilterTest.php
```diff
@@ -0,0 +1,140 @@
...
+            'src/Foo/Test.php',
+            'src/Bar/Baz.php',
+            'src/Example/Test.php',
+        ], [
+            'src/Foo/Test.php',
+            'src/Bar/Baz.php',
+        ]];
+    }
+
+    private function assertFilter(string $filter, iterable $input, array $expected): void
```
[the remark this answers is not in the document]
sanmai: OK, done

tests/phpunit/FileSystem/SourceFileFilterTest.php
sanmai: Yes, here we can see that downstream can take not only arrays, but interables in general.

src/Mutation/FileMutationGenerator.php
sanmai: Fixed

src/TestFramework/Coverage/CoveredFileNameFilter.php
sanmai: CoveredFileNameFilter is no more.

src/Mutation/MutationGenerator.php
sanmai: Alright, done

src/Configuration/Configuration.php
```diff
@@ -92,6 +93,7 @@ public function __construct(
         int $timeout,
         array $sourceDirectories,
         iterable $sourceFiles,
+        string $filter,
```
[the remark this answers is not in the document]
sanmai: OK, done

tests/phpunit/FileSystem/SourceFileFilterTest.php
```diff
@@ -0,0 +1,140 @@
...
+
+        yield ['', [
+            'src/Foo/Test.php',
+            'src/Bar/Baz.php',
+            'src/Example/Test.php',
+        ], [
+            'src/Foo/Test.php',
+            'src/Bar/Baz.php',
+            'src/Example/Test.php',
+        ]];
```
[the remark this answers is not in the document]
sanmai: This is what you get when you have only 12 inch display. Fixing this now.

src/TestFramework/Coverage/CoveredFileData.php
sanmai: Removed

src/TestFramework/Coverage/XmlReport/FileCodeCoverageProvider.php
sanmai: The reason they are so is because I wanted to keep tests as verbatim as possible. I don't recall other reason so they might be merged later just as well.

theofidry:
I feel like this PR will never end if we keep going :) So I would suggest to merge it as is and address the comments later on.

I'm usually not fond of that approach but hey, either we try to improve the underlying bits to enable that PR to be smaller, but that's gonna take a lot of efforts and time, or we go as is which is still a satisfactory state and fix the issues later on

sanmai:

PR #1102, opened 2020-03-02
Automatically install Test Framework adapter if needed

maks-rafalko:
Implements https://github.com/infection/infection/issues/1054

This is the last big step from decoupling Test Frameworks from Infection core.

We

* no longer depend on PHPSpec and Codeception adapters
* no longer depend on `symfony/yaml` (only in `require-dev`, since we are linting YAML files on CI)
* no longer downloads the PHPSpec / Codeception related code by default

When the user tries to run Infection with PHPSpec or Codeception, Infection asks confirmation to automatically install it **in the same process** and continue mutation process.

See 

[asciicast](https://asciinema.org/a/Pc3PdYOaa1ldaNNdF4SKgohiA)

PR #1107, opened 2020-03-03
Rework the coverage checker

theofidry:
There is a few things I don't like about the current state of the coverage checking:

- Some exceptions don't make sense. A non-exhaustive list:
  -  `CoverageDoesNotExistsException` thrown when the issue is a code coverage tool is not available
  - static factory for exceptions when there is no dynamic parameter and no re-use
  - extending `Exception` which makes it really confusing to know in what category to place it, even more confusing that it is never caught (so `LogicException` would be more fitter for example)
- Some validations are done early in the process (`InfectionContainer::startUp()`), later in the `Engine` (polluting a bit its dependencies) and in the coverage parsing
- Some error messages are not that helpful

So in this PR, I propose to:

- Move more coverage checks from `Engine` to `CoverageRequirementChecker`
- Rename `CoverageRequirementChecker` to `CoverageChecker`
- Remove `CoverageDoesNotExistsException` in favour of:
  - `InvalidCoverage`, used when parsing the coverage for example when it is likely that the coverage is outdated
  - `CoverageNotFound`, used when checking the existence of the coverage
- Rework almost all error messages to be more user-friendly
- Consolidate the checks done:
  - in the parsers, simplify some checks thanks to more checks done upstream
  - when checking the coverage, check the JUnit file as well instead of just the XML index file
- No longer document `InvalidCoverage` and `CoverageNotFound`: neither are meant to be caught at the moment


theofidry:
There is a failure I'm having trouble to fix, for some reasons I cannot run the e2e tests locally:

```
1) Infection\Tests\Console\E2ETest::test_it_runs_an_e2e_test_with_success with data set "Code_Coverage_Ignore" ('...Ignore')
Unexpected exit code. Command output was:
---
You are running Infection with Xdebug enabled.

 [WARNING] Infection cannot control exit codes and unable to relaunch itself.
           It is your responsibility to disable xdebug/phpdbg unless needed.


    ____      ____          __  _

Infection - PHP Mutation Testing Framework dev-refactor/coverage-checker@537362b39151457cd8a8d39c91433d36f6c6327a
Infection config generator requires an interactive mode.

In ConfigurationException.php line 45:

  [Infection\Console\Exception\ConfigurationException]
  Configuration aborted
```

sanmai:
Might be related to #1102

theofidry:
@sanmai master is green though, but [this](https://travis-ci.org/infection/infection/jobs/658148129#L671) is really weird

src/TestFramework/Coverage/CoverageChecker.php
```diff
@@ -0,0 +1,194 @@
...
+        $this->initialTestPhpOptions = $initialTestPhpOptions;
+        $this->coveragePath = $coveragePath;
+        $this->jUnitPath = $jUnitPath;
+    }
+
+    public function checkCoverageRequirements(): void
+    {
+        if ($this->skipInitialTests && !$this->skipCoverage) {
+            throw new CoverageNotFound(
+                'The initial test suite run is being skipped. The XML and JUnit coverage '
```
maks-rafalko: > and JUnit coverage

phpspec does not need JUnit coverage, so it's not correct for all the cases

src/TestFramework/Coverage/CoverageChecker.php
```diff
@@ -0,0 +1,194 @@
...
+            throw new CoverageNotFound(sprintf(
+                'Could not find the file "%s". Please ensure that the XML coverage report has been '
+                . 'properly generated at the right place. If using PHPUnit for example, the option '
+                . 'for the path given is "--coverage-xml=%s"',
+                $coverageIndexFilePath,
+                $this->coveragePath . '/coverage-xml'
+            ));
+        }
+
+        if (!file_exists($this->jUnitPath)) {
```
maks-rafalko: It won't work for PHPSpec (`TestFrameworkAdapter::hasJUnitReport()`).

Should be tested somehow to avoid these mistakes.

src/TestFramework/Coverage/CoverageChecker.php
```diff
@@ -0,0 +1,212 @@
...
+                $this->jUnitPath
+            );
+
+            if ($this->testFrameworkAdapter instanceof PhpUnitAdapter) {
+                $message .= sprintf(
+                    ' The PHPUnit option for the path given is "--log-junit=%s/junit.xml"',
+                    $this->coveragePath
+                );
+            } elseif ($this->testFrameworkAdapter instanceof CodeceptionAdapter) {
+                $message .= ' The Codeception option for the path given is "--xml"';
```
theofidry: I'm pretty sure this is incorrect, @maks-rafalko could you give me the right commend for codeception?

src/TestFramework/Coverage/CoverageChecker.php
```diff
@@ -0,0 +1,212 @@
...
+    {
+        $coverageIndexFilePath = Path::canonicalize(
+            $this->coveragePath . DIRECTORY_SEPARATOR . PhpUnitXmlCoverageFactory::COVERAGE_INDEX_FILE_NAME
+        );
+
+        if (!file_exists($coverageIndexFilePath)) {
+            throw new CoverageNotFound(sprintf(
+                'Could not find the file "%s". Please ensure that the XML coverage report has been '
+                . 'properly generated at the right place. If using PHPUnit for example, the option '
+                . 'for the path given is "--coverage-xml=%s"',
```
theofidry: @maks-rafalko what would be the command for phpspec & codeception?

tests/phpunit/TestFramework/Coverage/CoverageCheckerTest.php
```diff
@@ -0,0 +1,516 @@
...
+
+    public function test_it_does_not_pass_existence_check_if_JUnit_file_is_missing_with_JUnit_report_with_Codeception_test_framework_adapter(): void
+    {
+        $checker = new CoverageChecker(
+            false,
+            false,
+            '',
+            self::$coveragePath,
+            '/invalid/path/to/junit.xml',
+            new CodeceptionAdapter()
```
theofidry: I'm a bit in a pinch here. I tried to be smart and "mock" it by adding it at the end of the file and it works but seem to be deal breaker for phpstan :/

Maybe an alternative is to just pass the "has JUnit report" and test framework adapter name instead. That would make it easier but I'm a bit concerned about testing against the adapter name which we have no control over from here.

@maks-rafalko would I be wrong to add back the adapters as dev dependencies?

sanmai: Another way would be to add an interface for each of the external adapters, and make them use it. E.g.
```
class CodeceptionAdapter implements CodeceptionAdapterInterface {}
```
Where `CodeceptionAdapterInterface` located in the root package. Also solves problem with string constants inside AdapterInstallationDecider.

src/TestFramework/Coverage/CoverageChecker.php
maks-rafalko: See 

https://github.com/infection/codeception-adapter/blob/aa4db3c5748ab4236938efc71f81b08c8f793473/src/CodeceptionAdapter.php#L162-L163

But `infection/infection` must not depend on `use Infection\TestFramework\Codeception\CodeceptionAdapter;`, never. It's not correct from dependencies POV. What `infection/infection` should know is only about `TestFrameworkAdapter` interface.

If this message is so important (which I doubt), it should be on each adapter implementation

tests/phpunit/TestFramework/Coverage/CoverageCheckerTest.php
maks-rafalko: `infection/infection` should not test adapters. Codeception, PHPSpec is not different to X, Y or Z adapter that can be added in the future. It's a responsibility of adapter to test how it works with the core.

IMO, what `infection/infection` should do is only rely on `TestFrameworkAdapter` inserface, no matter what implementation we have.

If some thing is on the interface - we test it in an abstract way. If not - we don't write "hardcoded" code for particular adapter.

> @maks-rafalko would I be wrong to add back the adapters as dev dependencies?

yep, it's incorrect and will lead to cycles in dependencies

---

We already have [`getInitialTestsFailRecommendations()`](https://github.com/infection/abstract-testframework-adapter/blob/master/src/TestFrameworkAdapter.php#L72) method for each adapter which does pretty much the same - it returns custom messages.

sanmai: Otherwise put, to move forward with this PR it is imperative to update TestFrameworkAdapter to include an additional method?

I don't think this is bad or anything, but rather to clarify.

theofidry: It's not that important of a message and really depends on how infection is being used, so I don't think it's the responsibility of the adapter to know that context.

I'll just inject the test framework name and has JUnit report instead: less clean but will work. There is the risk that the test framework adapter changes of names and mess up there but hey, it's not that critical so I think it's ok

tests/e2e/Provide_Existing_Coverage/infection-coverage/junit.xml
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <testsuites>
-  <testsuite name="Test Suite" tests="2" assertions="2" errors="0" failures="0" skipped="0" time="0.010499">
-    <testsuite name="Namespace_\Test\SourceClassTest" file="tests/SourceClassTest.php" tests="2" assertions="2" errors="0" failures="0" skipped="0" time="0.010499">
-      <testcase name="test_it_adds" class="Namespace_\Test\SourceClassTest" classname="Namespace_.Test.SourceClassTest" file="tests/SourceClassTest.php" line="10" assertions="1" time="0.007975"/>
-      <testcase name="test_it_returns_true" class="Namespace_\Test\SourceClassTest" classname="Namespace_.Test.SourceClassTest" file="tests/SourceClassTest.php" line="16" assertions="1" time="0.002524"/>
+  <testsuite name="Test Suite" tests="2" assertions="2" errors="0" failures="0" skipped="0" time="0.015839">
+    <testsuite name="ProvideExistingCoverage\Test\SourceClassTest" file="/Users/tfidry/Project/Humbug/infection/tests/e2e/Provide_Existing_Coverage/tests/SourceClassTest.php" tests="2" assertions="2" errors="0" failures="0" skipped="0" time="0.015839">
```
maks-rafalko: 1. why the path changed from relative to absolute?
2. humbug? :)

theofidry: no idea... I tried to generate it again because the existing one was giving a failure

theofidry: forgive me :P I have way more than I should do under that namespace:

Screen Shot 2020-03-06 at 22 57 20

maks-rafalko: what kind of failure?

maks-rafalko: I mean, if new code started failing for existing tests, isn't it a marker of introduced bug?

JUnit report may be different for different PHPUnit versions, as well as for different test frameworks, unfortunately

theofidry: It appears so, think it is fixed now

theofidry:
Allelouya

PR #1112, opened 2020-03-04
PHPUnit group name accepts no comments

sanmai:
Apparently PHPUnit cannot handle `@group name Comment` format. For example, this command should not execute any tests in this file:
```
php vendor/bin/phpunit  --debug --exclude-group=integration tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
```

But it does. This PR solves this problem.


theofidry:
Does that mean that it's interpreted as the group `name Comment`?

PR #1063, opened 2020-02-20
Add sarb logging option

BackEndTea:
This PR:

- [x] Adds the option to log a json file readable by [sarb](https://github.com/DaveLiddament/sarb), allowing a baseline feature

@DaveLiddament, could you look if the output is correct?

@theofidry Im gonna need some help with the failing tests, as i'm not sure how they work

PR #1114, opened 2020-03-04
Add sarb logger

theofidry:
Replacement for #1063.

Adds the an option to log a json file readable by [sarb](https://github.com/DaveLiddament/sarb), allowing a baseline feature


tests/phpunit/Logger/SarbLoggerTest.php
```diff
@@ -0,0 +1,97 @@
...
+    public function metricsProvider(): Generator
+    {
+        yield 'no mutations' => [
+            new MetricsCalculator(),
+            <<<'JSON'
+[]
+JSON
+        ];
+
+        yield 'all mutations' => [
```
theofidry: @DaveLiddament, could you look if the output is correct? The file path is expected to be an absolute path here, that is to be fixed

maks-rafalko:
Any explanation of how the final user can use it, benefits, etc.?

theofidry:
ping @BackEndTea

PR #1117, opened 2020-03-04
Remove large group (#1115)

theofidry:
As far as I can see this group is not actually used and this test is run as part of `e2e`. @maks-rafalko can you confirm?

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -4,6 +4,8 @@
          bootstrap="./vendor/autoload.php"
          colors="true"
          executionOrder="random"
+         failOnWarning="true"
+         failOnRisky="true"
          verbose="true"
 >
 
@@ -25,12 +27,6 @@
             <directory>src</directory>
         </whitelist>
     </filter>
-
-    <groups>
-        <exclude>
-            <group>large</group>
-        </exclude>
-    </groups>
 
     <listeners>
         <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
--- a/tests/phpunit/Console/E2ETest.php
+++ b/tests/phpunit/Console/E2ETest.php
@@ -125,22 +125,13 @@ final class E2ETest extends TestCase
         chdir($this->cwd);
     }
 
-    /**
-     * Longest test: runs under about 160-200 sec
-     *
-     * To be run with:
-     *
-     * php -dmemory_limit=128M vendor/bin/phpunit --group=large
-     *
-     * @large
-     */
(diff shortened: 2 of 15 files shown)


maks-rafalko:
in `master`:

`vendor/bin/phpunit --group=e2e` does *not**execute `test_it_runs_on_itself` test

So this PR is quite dangerous, this very large test will be executed with other `e2e` tests while previously it wasn't

We should ask @sanmai why it was added. Probably we don't need it?

theofidry:
When do we ever execute it then? I could not find any occurrence anywhere

tests/phpunit/Console/E2ETest.php
```diff
@@ -125,22 +125,13 @@ protected function tearDown(): void
...
-     * @large
-     */
     public function test_it_runs_on_itself(): void
     {
         if (ini_get('memory_limit') === '-1') {
             $this->markTestSkipped(implode("\n", [
                 'Refusing to run Infection on itself with no memory limit set: it is dangerous.',
                 'To run this test with a memory limit set please use:',
-                'php -dmemory_limit=128M vendor/bin/phpunit --group=large',
+                'php -dmemory_limit=128M vendor/bin/phpunit --group=e2e',
```
sanmai: When this was written, the entire test suite wasn't marked as e2e. So it was correct back then.

sanmai:
The build is failing with:

> Cannot open file ""--exclude-group=integration".php".

tests/phpunit/Console/E2ETest.php
```diff
@@ -125,22 +125,13 @@ protected function tearDown(): void
...
     }
 
-    /**
-     * Longest test: runs under about 160-200 sec
-     *
-     * To be run with:
-     *
-     * php -dmemory_limit=128M vendor/bin/phpunit --group=large
-     *
-     * @large
```
sanmai: I'd rather keep this designation. And explanation to its existence.

phpunit.xml.dist
```diff
@@ -26,12 +28,6 @@
         </whitelist>
     </filter>
 
-    <groups>
-        <exclude>
-            <group>large</group>
-        </exclude>
-    </groups>
```
sanmai: And so is best to ignore this too, IMO.

sanmai:
About removing, this is a fine test, for example, to run when you want to find dead code.

As it is only few lines of code, I propose to keep it.

theofidry:
Closing in favour of #1120

PR #1118, opened 2020-03-05
Back up globals in tests that modify environment variables

theofidry:
This PR:

- Introduces a new component `EnvBackup` to easily save & restore environment variables
- A trait leveraging `EnvBackup` to use its usage
- An env manipulator detector to do a smoke test that the trait is used in test cases manipulating environment variables

Replaces #1068. I cannot guess what the error there, I tried to dig it up but it's literally faster to just work around it than fixing it.


tests/phpunit/Env/EnvBackup.php
```diff
@@ -0,0 +1,90 @@
...
+
+namespace Infection\Tests\Env;
+
+use function array_key_exists;
+use function getenv;
+use function Safe\putenv;
+use function Safe\sprintf;
+use Webmozart\Assert\Assert;
+
+final class EnvBackup
```
localheinz: On one hand you argue that we do not need an extra package, but now we reinvent what is already installed.

Why not use [`sebastianbergmann/global-state`](https://github.com/sebastianbergmann/global-state)?

tests/phpunit/Env/BackupEnvVariables.php
```diff
@@ -0,0 +1,63 @@
...
+use Webmozart\Assert\Assert;
+
+trait BackupEnvVariables
+{
+    /**
+     * @var EnvBackup
+     */
+    private $snapshot;
+
+    private function createEnvBackup(): void
```
localheinz:
```suggestion
    private function backupEnvironmentVariables(): void
```

tests/phpunit/Env/BackupEnvVariables.php
```diff
@@ -0,0 +1,63 @@
...
+    {
+        $this->snapshot = EnvBackup::createSnapshot();
+    }
+
+    private function restoreEnvBackup(): void
```
localheinz:
```suggestion
    private function restoreEnvironmentVariables(): void
```

tests/phpunit/Env/BackupEnvVariables.php
```diff
@@ -0,0 +1,63 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Tests\Env;
+
+use Webmozart\Assert\Assert;
+
+trait BackupEnvVariables
```
localheinz:
```suggestion
trait BacksUpEnvironmentVariables
```

tests/phpunit/AutoReview/EnvChecker/EnvManipulatorsTest.php
```diff
@@ -0,0 +1,70 @@
...
+                <<<'TXT'
+Expected the test case "%s" to be using the "%s" trait as environment variable manipulations have 
+been found in the file "%s".
+TXT
+                ,
+                $testCaseClassName,
+                BackupEnvVariables::class,
+                $fileWithEnvManipulations
+            )
+        );
```
localheinz: Is there any chance we can stop using `assertFalse()`, `assertTrue()`, and its counterparts, and instead start using dedicated assertions?

```suggestion
        $import = sprintf(
            'use %s;',
            BackupEnvVariables::class
        );

        $this->assertContains($import, file_get_contents($fileWithEnvManipulations), sprintf(
            <<<'TXT'
Expected the test case "%s" to be using the "%s" trait as environment variable manipulations have
been found in the file "%s".
TXT
            ,
            $testCaseClassName,
            BackupEnvVariables::class,
            $fileWithEnvManipulations
        ));
```

tests/phpunit/Env/EnvBackup.php
theofidry: Because it is way more complex since is handling a lot more cases, and also is probably what, somehow, causes the issue in your PR. So I'm fine with declaring a tiny class to avoid that dependency, there is nothing to complex there

tests/phpunit/AutoReview/EnvChecker/EnvTestCasesProvider.php
```diff
@@ -0,0 +1,189 @@
...
+use Generator;
+use Infection\Tests\AutoReview\ProjectCode\ProjectCodeProvider;
+use Infection\Tests\AutoReview\SourceTestClassNameScheme;
+use function iterator_to_array;
+use PHPUnit\Framework\TestCase;
+use ReflectionClass;
+use function Safe\file_get_contents;
+use Webmozart\Assert\Assert;
+
+final class EnvTestCasesProvider
```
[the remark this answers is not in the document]
theofidry: I wanted to avoid to conflict with `Infection\Environment` that contains different stuff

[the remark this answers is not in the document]
localheinz: If we replaced every occasion of `env` in

- class names
- trait names
- method names
- variable names
- field names

of code added in this pull request with `environment` - where would we see a conflict?

tests/phpunit/Env/EnvBackup.php
localheinz: Why not write our own test framework then, since we do not need all of the assertions in `phpunit/phpunit`?

tests/phpunit/AutoReview/EnvChecker/EnvCodeDetector.php
```diff
@@ -0,0 +1,84 @@
...
+declare(strict_types=1);
+
+namespace Infection\Tests\AutoReview\EnvChecker;
+
+use function Safe\sprintf;
+use function strpos;
+
+final class EnvCodeDetector
+{
+    private const FUNCTIONS = [
```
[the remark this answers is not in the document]
sanmai: Please let's keep it. Next time you'd spent minutes looking for a list of functions you remember you've seen somewhere, somewhere, somewhere. It's annoying to no end to spend time on these things.

tests/phpunit/Env/EnvBackup.php
localheinz: I'm a bit concerned because a lot of code has been added recently that has nothing to do with the primary concern of `infection/infection`.

theofidry: >Why not write our own test framework then, since we do not need all of the assertions in phpunit/phpunit?

Sorry how is that even remotely relevant? For starters we are heavily using PHPUnit and its features and nothing warrants to use another test framework.

Also for the records: it's not like I'm adding this because I don't like PHPUnit, but because we cannot make use of the globals back-up and because you and I could not figure out why and I don't want to spend 3 hours looking for it when a 50LoC class does the job.

tests/phpunit/AutoReview/EnvChecker/EnvTestCasesProvider.php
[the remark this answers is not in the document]
theofidry: We won't have a code conflict, but with already have the namespace `Infection\Environment` which is for quite a different type of "environment". So I did not want to mix the two in there. Maybe I could go with `EnvManipulation` or `EnvironmentManipulation` instead though?

tests/phpunit/AutoReview/EnvChecker/EnvManipulatorsTest.php
theofidry: When an acceptable alternative is available sure :)

tests/phpunit/AutoReview/EnvVariableManipulation/EnvManipulatorCodeDetectorTest.php
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * @covers \Infection\Tests\AutoReview\EnvVariableManipulation\EnvManipulatorCodeDetector
+ */
+final class EnvManipulatorCodeDetectorTest extends TestCase
+{
+    /**
+     * @dataProvider codeProvider
+     */
+    public function test_it_can_detect_IO_operations(string $code, bool $expected): void
```
localheinz: As far as I can tell, the name of the test method is misleading - or does the `EnvManipulatorCodeDetector` also detect modifications of the file system?

tests/phpunit/EnvVariableManipulation/EnvBackup.php
```diff
@@ -0,0 +1,90 @@
...
+
+                continue;
+            }
+
+            putenv($name);
+        }
+
+        foreach ($snapshot as $name => $value) {
+            putenv(sprintf('%s=%s', $name, $value));
+        }
```
localheinz:
```suggestion
        foreach (getenv() as $name => $value) {
            if (array_key_exists($name, $this->environmentVariables)) {
                continue;
            }
            
            putenv($name);
        }

        foreach ($this->environmentVariables as $name => $value) {
            putenv(sprintf('%s=%s', $name, $value));
        }
```

theofidry:
```
Phpunit_Bat_Wrapper/
In CoverageDoesNotExistException.php line 65:
                                                                               
  Code Coverage does not exist. File /home/travis/build/infection/infection/t  
  ests/e2e/Phpunit_Bat_Wrapper/./infection/coverage-xml/index.xml is not foun  
  d. Check phpunit version Infection was run with and generated config files   
  inside /home/travis/build/infection/infection/tests/e2e/Phpunit_Bat_Wrapper  
  /./infection.                                                                
  Command line: '/home/travis/build/infection/infection/tests/e2e/Phpunit_Bat  
  _Wrapper/phpunit.bat' '--configuration' '/home/travis/build/infection/infec  
  tion/tests/e2e/Phpunit_Bat_Wrapper/./infection/phpunitConfiguration.initial  
  .infection.xml'                                                              
  Process Output: PHPUnit 8.5.2 by Sebastian Bergmann and contributors.
```

Shouldn't this e2e test only be run on AppVeyor?

tests/phpunit/AutoReview/EnvVariableManipulation/EnvManipulatorCodeDetectorTest.php
theofidry: no just a bad copy paste. most of the detector stuff was from the FS one

tests/phpunit/EnvVariableManipulation/EnvBackup.php
theofidry: The second suggestion does not work. For the former, why are you against preventing a write if unnecessary? 

Good idea to release the variables

theofidry: I'll do the "release" in the trait though

theofidry: Is there a need to release it though? it's just an array of string in the end, it doesn't take any space and is cleaned up by PHPUnit already

tests/phpunit/AutoReview/EnvVariableManipulation/EnvTestCasesProvider.php
```diff
@@ -0,0 +1,189 @@
...
+
+            return;
+        }
+
+        self::$envTestCaseClassesTuple = array_values(array_filter(array_map(
+            static function (string $className): ?array {
+                return self::envTestCaseTuple($className);
+            },
+            iterator_to_array(ProjectCodeProvider::provideSourceClasses(), true)
+        )));
```
sanmai:
```php
take(ProjectCodeProvider::provideSourceClasses())
    ->map(static function (string $className): ?array {
       return self::envTestCaseTuple($className);
    })
    ->filter()
    ->toArray();
```

tests/phpunit/Logger/BadgeLoggerTest.php
```diff
@@ -118,6 +88,11 @@ protected function setUp(): void
         );
     }
 
+    protected function tearDown(): void
+    {
+        $this->restoreEnvironmentVariables();
+    }
+
```
sanmai: Nice!

tests/phpunit/AutoReview/EnvVariableManipulation/EnvTestCasesProvider.php
sanmai: Not the best fit for `iterator_to_array($foo, true)`

PR #1121, opened 2020-03-05
Automatically adjust wait times and bucket size

sanmai:
This PR:

- [x] Makes ParallelProcessRunner automatically adjust wait times and bucket size
- [x] Removes arbitrary constants

I do not expect this to make much difference in master because master does very little in these gaps between processes, but with #1106 it really shines.

### DIFF
--- a/src/Process/Runner/Parallel/ParallelProcessRunner.php
+++ b/src/Process/Runner/Parallel/ParallelProcessRunner.php
@@ -52,14 +52,6 @@ use function usleep;
  */
 class ParallelProcessRunner
 {
-    /**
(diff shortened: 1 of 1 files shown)


theofidry:
Will try to get a profiling tonight just to see 👍

sanmai:
[Here's a diff.](https://blackfire.io/profiles/compare/a07c885c-94b4-4003-86a0-21a23b211fd1/graph
) 

Not really an improvement in regard of master, just as I thought it won't matter here.

theofidry:
Thanks!

PR #1122, opened 2020-03-05
Improve PerMutator logger

theofidry:
- Indicate the score units (%)
- Round and format the scores to ensure they are 2 significant digit instead of integers
- Pad the cells to create a more readable table


sanmai:
What I really would've loved to see in this table, is the average test run time per mutation. Then you can export this table, sort it, and get several "fastest" mutation, which is a great thing when you want to add mutation testing to some elaborate project with lots of test and huge coverage, where running with `@default` takes forever and makes no sense. Once you get it up and running, you can trivially extend the list of used mutations.

It this an idea worthy of a issue, what do you think?

src/Logger/PerMutatorLogger.php
```diff
@@ -58,30 +70,115 @@ public function __construct(MetricsCalculator $metricsCalculator)
...
+        return $sizes;
+    }
+
+    private static function padCell(string $value, int $size, bool $toLeft): string
+    {
+        return str_pad($value, $size, ' ', $toLeft ? STR_PAD_LEFT : STR_PAD_RIGHT);
+    }
+
+    /**
+     * @var int[]
```
sanmai:
```suggestion
     * @param int[] $columnSizes
     *
     * @var int[]
```

theofidry:
How would you handle the sorting?

I agree it would be very useful to have a report purely for internal metrics analysis, but I was more thinking of a CSV format rather than text with a pretty table

sanmai:
I'd use Excel or Sheet, this isn't a problem.

theofidry:
But wouldn't it be preferable to import that data Excel/Sheet ready rather than requiring you to copy it over?

sanmai:
Oh, yes, that'd be nicer. CSV or TSV is definitely better. 

TBH I can't remember why we have Markdown here.

theofidry:
Alright let's open an issue for that then :) I'm still not quite sure what some loggers are used for (e.g. #1123) and if we want one to collect internal metrics it might be worth having a discussion about the data collected in there

maks-rafalko:
Sigh.. GitHub should disable the approve button if builds are failing. 

https://travis-ci.org/infection/infection/jobs/658672709#L402

theofidry:
Yeah gonna fix that and the AppVeyor issue as well before merging

PR #1132, opened 2020-03-06
Drop support for PHP 7.2

sanmai:
This PR:

- [x] Removes PHP 7.2 from supported versions
- [x] Removes PHP 7.2 from CI
- [x] Updates dependencies
- [x] Decreases CI delays

Rationale:
- PHP 7.2 is [past active support deadline](https://www.php.net/supported-versions.php), soon reaching EOL.
- Infection isn't a unit-testing framework, but rather a CI tool, supposed to be run only once per build. If an app supports PHP 7.3, it can run Infection too in one of the jobs. 
- It appears to me PHP 7.2 wasn't as widespread. E.g. Debian jumped 7.3 right from 7.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ before_install:
 jobs:
   include:
     - stage: Static Code Analysis
-      php: 7.2
+      php: 7.3
       install:
         - composer install $COMPOSER_FLAGS
       script:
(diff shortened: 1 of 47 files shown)


theofidry:
@maks-rafalko the current master already contain a bunch of BC break so what about making it as part of 0.16 instead of 0.17?

maks-rafalko:
👍

sanmai:

PR #1134, opened 2020-03-06
Draft: Count timeouts as escapes

### LINKED ISSUE #1133: Add config option to count timeouts as mutant escapes
I'm concerned that mutation runs on CI environments may inadvertently be overstating MSI scores.

I keep seeing a wide disparity in number of timeouts between running Infection locally (2 timeouts out of ~1200 mutants) vs running on CI (98 timeouts out of ~1000 mutants - that job only mutates files that were changed in the PR) - a timeout rate nearly 60x greater on CI.  This is most likely due to the differing processing power available to each case.

As I understand it, Infection currently counts a mutant against the run's MSI score if it didn't produce some visible misbehaviour (test failure, fatal error, timeout).

I'm asking for an option, settable in either infection-json.dist or supplied via command line, that alters the sense of MSI calculation.  This would simply not count timeouts as "visible misbehaviour".

Referring back to the sample Travis job I attached, only the 871 killed mutants (and the zero fatal errors) would count towards the overall MSI (which would be 87.6% under my proposal) and covered-code MSI (which would be 89.1%), not the reported values of 97% and 98%, respectively.

Although my initial use case arises out or running under a CI environment, I would be reluctant to default it to being set for CI runs.  That feels too much like setting policy, not merely providing mechanism.

- Dug through Infection docs, couldn't find any way to kitbash this together.
- Bump timeout on CI runs.  Didn't do this because mutation runs on Travis take long enough already.

**Additional context**
Sample CI timeout-laden run:
https://travis-ci.org/Algo-Web/POData-Laravel/jobs/659174053?utm_medium=notification&utm_source=github_status

CyberiaResurrection:
This PR:

- [x] Adds new feature ...
- [x] Covered by tests

Fixes https://github.com/infection/infection/issues/1133

This PR adds a boolean flag to both the Calculator and MetricsCalculator classes in the Infection\Mutant namespace that, per my issue #1133 , counts timed-out mutants as escapes/failures (with resulting effect on calculated MSI.

I haven't charged ahead into the config side hookup both because I got a little lost tracking it down, and I wanted to solicit workable names for the configuration and/or command line options that would be used to enable this behaviour.  I suspect what I've used internally, countTimeoutsAsEscapes, might be a bit too long.

### DIFF
--- a/src/Mutant/Calculator.php
+++ b/src/Mutant/Calculator.php
@@ -105,6 +105,7 @@ final class Calculator
 
         $score = 0.;
         $coveredTotal = $this->killedCount + $this->errorCount;
+
         if (!$this->treatTimeoutsAsEscapes) {
             $coveredTotal += $this->timedOutCount;
         }
@@ -149,6 +150,7 @@ final class Calculator
         $score = 0.;
         $testedTotal = $this->totalCount - $this->notTestedCount;
         $coveredTotal = $this->killedCount + $this->errorCount;
+
         if (!$this->treatTimeoutsAsEscapes) {
             $coveredTotal += $this->timedOutCount;
         }
--- a/src/Mutant/Calculator.php
+++ b/src/Mutant/Calculator.php
@@ -89,7 +89,8 @@ final class Calculator
             $calculator->getErrorCount(),
             $calculator->getTimedOutCount(),
             $calculator->getNotTestedCount(),
-            $calculator->getTotalMutantsCount()
+            $calculator->getTotalMutantsCount(),
+            $calculator->getTreatTimeoutsAsEscapes()
         );
     }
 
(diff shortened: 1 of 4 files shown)


src/Mutant/Calculator.php
```diff
@@ -137,7 +149,11 @@ public function getCoveredCodeMutationScoreIndicator(): float
 
         $score = 0.;
         $testedTotal = $this->totalCount - $this->notTestedCount;
-        $coveredTotal = $this->killedCount + $this->timedOutCount + $this->errorCount;
+        $coveredTotal = $this->killedCount + $this->errorCount;
+
+        if (!$this->treatTimeoutsAsEscapes) {
+            $coveredTotal += $this->timedOutCount;
+        }
```
sanmai: I had some time to think and now I think this should be the default. I'll write it out later.

sanmai:
I wonder if there's a chance you can see if #1171 will satisfy your initial requirements for the problem.

theofidry:
This has been added via https://github.com/infection/infection/pull/2817 and is available as of [0.32.3](https://github.com/infection/infection/releases/tag/0.32.3).

PR #1139, opened 2020-03-06
Move dumping the mutant out of the factory

theofidry:
If you look at this PR, you might give it a very confusing look as to why the heck we have a class which is literally one dump. Actually as I'm writing this I think I will replace that by injecting the filesystem instead: I wanted a class instead of a direct call for mocking purposes, to keep the test IO-less.

Back to the topic, what do I want to achieve here? In short, reduce the cost of the `mutation -> mutant execution result` process.

Right now the process is:

- mutation
- create a mutant
  - get mutated code (AST parsing + printing)
  - get a diff (mutated code - original code)
  - dump (I/O)
- create a process
- execute the process
  - execution skipped for non-covered processes
- create mutant execution result

There is two scenarios where there is unnecessary steps:

### Scenario1: non-covered mutations

since the process is never actually used, we neither need to create the process or dump the mutant

### Scenario2: skipped mutations

In the case we have an order about mutations to skip some if one is failed (e.g. the top-down optimization strategy), we have the dump + process creation that is unnecessary

That is two scenarios (one not in use yet) where we can reduce a bit the cost of the `mutation -> mutant execution result` process.

Let's keep in mind that the optimization above is still within the realm of mini if not micro-optimization. So the solution _needs_ to be simple: otherwise it's not worth the trade-off.

### Proposed solution

Introduce `Mutant::getCode()` (the mutated one in case it's not clear). This way we remove the I/O step from getting a mutant but in case we don't ever execute the mutant, we have all the necessary information to create the mutation execution result. This would allow later on to filter _first_ the mutations (by whether they are covered or not) and then map them to create processes instead of vice-versa, without changing anything downstream.

WDYT? /cc @sanmai @maks-rafalko 

If you agree with the optimization above, I'll finish this PR. I realise this should more appropriate as an issue but oh well...


src/Mutant/MutantFactory.php
```diff
@@ -80,28 +77,16 @@ public function create(Mutation $mutation): Mutant
...
-    {
-        if (is_readable($mutantFilePath)) {
-            return file_get_contents($mutantFilePath);
-        }
-
-        $mutantCode = $this->mutantCodeFactory->createCode($mutation);
-
-        file_put_contents($mutantFilePath, $mutantCode);
-
-        return $mutantCode;
```
sanmai: yield here

src/Mutant/Mutant.php
```diff
@@ -46,16 +46,22 @@ class Mutant
 {
     private $mutantFilePath;
     private $mutation;
+    private $code;
     private $diff;
 
-    public function __construct(string $mutantFilePath, Mutation $mutation, string $diff)
```
sanmai: iterable $mutantFilePath

src/Mutant/Mutant.php
```diff
@@ -46,16 +46,22 @@ class Mutant
 {
     private $mutantFilePath;
```
sanmai: `@var string|iterable`

src/Mutant/Mutant.php
```diff
@@ -46,16 +46,22 @@ class Mutant
...
         $this->mutantFilePath = $mutantFilePath;
         $this->mutation = $mutation;
+        $this->code = $code;
         $this->diff = $diff;
     }
 
-    public function getMutantFilePath(): string
+    public function getFilePath(): string
     {
         return $this->mutantFilePath;
```
sanmai:
```php
if (!is_string($this->mutantFilePath)) {
  foreach ($this->mutantFilePath as $mutantFilePath) {
    $this->mutantFilePath = $mutantFilePath;
    break;
  }
  Assert::string($this->mutantFilePath);
}
```

sanmai:
Hang on, I'd do it to confirm.

theofidry:
>I'd defer the whole combo of createMutantCode and createMutantDiff, but this will require something more than a single yield. Probably this'll need something like MutantCodeFactory that will provide diff and write to the file only when needed. Another question if this is going to be a valid optimization.

I'm thinking at some point to merge the current `Mtuation` and `Mutant`, but making the calculated property lazily. But that's for later :)

PR #1144, opened 2020-03-09
Remove --prefer-dist

sanmai:
E.g. see: https://travis-ci.org/infection/infection/jobs/660189536#L242


sanmai:
Nevermind. This isn't Composer. It's something with Travis CI.

Issue #1145, opened 2020-03-09
Track different sorts of timeouts differently


sanmai:
**Is your feature request related to a problem? Please describe.**

It is a lot said these days about "timeout" mutations.

- Some say that if Infection changes the code for something that makes it time out, it's not something a consumer should be bother with.
- Others might say that timeout mutations are testable and should be grouped together with other escapee. There's even a proposal to make this happed with a switch: #1133.

I gave it a long though, and, after swinging back and forth, I left with a mixed feeling about accepting or rejecting either approach. They're both *correct*, depending on circumstances.

The problem here is that there are different timeout mutations with different causes.

- Sometimes tests are just too long, and CI machine is busy, so a test can naturally time out not because it wouldn't succeed, but because Infection isn't configured to wait that long. A user can hardly do anything about these. (There is a way to work around these, but let's think like there isn't for now.)

- Sometimes code really goes bonkers, goes onto rampage in an infinite loop, and such. This is a different situation. In my experience I was able to isolate and these these mutations, sometimes by simply mocking a producer. It is unfair to group these "timeouts" with others since these are worthy of actionable feedback from the user. These are actual test defects Infection promised to help find.

**Describe the solution you'd like.**

I'd like Infection to discern between different types of timeouts. Actual test-originated timeouts may be left alone, where code-originated timeout can be grouped as **runaway mutations**, or as `R` in the report. Runaways in a sense that they not just escaped, but went really long way across the imaginable country doing their worst deeds. Think Bonnie and Clyde, but mutated code.

**Ways to do it.**

We can do as, say, Travis CI does to detect stuck jobs: by tracking output. If there's no output for a specified period of time, it is a runaway. Yes, a broken test may output something themselves, but there's `--disallow-test-output` to make PHPUnit fail gracefully.

It may worth exploiting timeout tracking capabilities PHPUnit offers with `--enforce-time-limit` and `--default-time-limit=N`, although it may be better to instruct a consumer to routinely apply these during testing. There's also `--disallow-resource-usage` but it may not work with untagged tests.

If we'd go deep into processes territory, it should be possible exactly how much a process uses, setting some limits. PHPUnit does that somehow.

maks-rafalko:
I like how these 2 groups are split, makes sense. Explaining this to our users is a nice thing as well.

Didn't dig into all these options of PHPUnit and have no idea how possible it is to implement, but in general idea of introducing `R` and differentiate these timeouts sounds good to me.

Ocramius:
[...]

PR #1152, opened 2020-03-11
Increase niceness for mutant processes

sanmai:
A runaway mutation (#1145) can cause a system resource exhaustion. Therefore, Infection may not be able to secure enough CPU cycles to stop this malignant mutation on time.

- We can work around this issue if we make Infection run with a higher priority by lowering priority of mutant processes. (Raising own priority usually is impossible, and not advisable in principle.)
- Downside of running anything with lower priority is that OS will not spin up CPU as much as it would for a higher- or normal-priority processes. Therefore Infection may run slower on energy-limited devices like laptops.
- This approach [works on Windows too](https://www.php.net/manual/en/function.proc-nice.php#refsect1-function.proc-nice-changelog).

This PR also add syntax validation, which we can keep if the above idea does not find approval.

### DIFF
--- a/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
+++ b/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
@@ -164,6 +164,9 @@ class MutationConfigBuilder extends ConfigBuilder
             <<<'PHP'
 <?php
 
+if (function_exists('proc_nice')) {
+    proc_nice(10);
+}
 %s
 require_once '%s';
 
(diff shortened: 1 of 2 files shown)


sanmai:
Thank you @theofidry

maks-rafalko:
Could you please explain in different words how it was working before and what is changed with this function call? I don't think I understood the issue and solution, sorry

sanmai:
After starting mutation processes Infection checks in a loop if they're out of time, with a slight delay. The problem is that this delay isn't guaranteed in non-RTOS. If the system is under a heavy load, it may take quite some time until Infection is given control back to stop these "timed out" processes. 

But if we lower priority for these processes, we can reasonably expect that the operating system will prioritize Infection main process over these rogue mutant processes. 

The quicker Infection can react to them, the less harm is done, and smoother things go.

maks-rafalko:
Rocket since. Thank you @sanmai

PR #1153, opened 2020-03-11
Account median testing time

sanmai:
Goal is to show users why MT is slow on their project.

We mostly have every class covered by a single test. Therefore for us MT is fairly quick. Median time is on range 0.01 s. which is very good, if my calculations aren't wrong. Basically this means that for the half of all mutations there's only 0.01 s. of tests to check if it escapes, not including test dependencies.

Yet there could be outliers, hence single metric won't work. Min, mean, sd, median, and max for the values needed to be shown. 68–95–99.7 percentiles are of interest too.

This PR:

- [x] Adds new feature:  test time reporting
- [ ] Covered by tests

Report of interest from ApacheBench:
```
Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        4   81  71.5     49     222
Processing:     1   18  10.5     17      53
Waiting:        0   12   9.5      9      52
Total:          7   99  80.4     67     270

Percentage of the requests served within a certain time (ms)
  50%     67
  66%     82
  75%    170
  80%    179
  90%    244
  95%    250
  98%    258
  99%    258
 100%    270 (longest request)
```


sanmai:
I don't think this is any longer relevant post #1171.

PR #1156, opened 2020-03-12
Simplify parsing of source file info

theofidry:


src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
```diff
@@ -77,29 +79,24 @@ private function parseNodes(string $coverageIndexPath, SafeDOMXPath $xPath): ite
...
-                $this->coverageDir,
-                $relativeCoverageFilePath,
-                $projectSource
+            yield $this->xmlCoverageParser->parse(
+                new SourceFileInfoProvider(
+                    $coverageIndexPath,
+                    $this->coverageDir,
+                    $relativeCoverageFilePath,
+                    $projectSource
+                )
```
sanmai: There's no need for `IndexXmlCoverageParser` to be even aware of presence of `XmlCoverageParser`. `IndexXmlCoverageParser` produces `SourceFileInfoProvider`, and `XmlCoverageParser` consumes. They can be connected by some other entity, like `PhpUnitXmlCoveredFileDataProvider`.

src/TestFramework/PhpUnit/Coverage/XmlCoverageParser.php
```diff
@@ -49,31 +49,20 @@
...
     {
         return new SourceFileData(
-            $this->provider->provideFileInfo(),
-            $this->lazilyRetrieveCoverageReport(
-                $this->provider->provideXPath()
-            )
+            $provider->provideFileInfo(),
+            (static function (SafeDOMXPath $xPath) {
+                yield self::retrieveCoverageReport($xPath);
+            })($provider->provideXPath())
```
sanmai: IIRC one-off closures are slightly slower than plain old static functions, but this might be not hot enough place to show it. Not mentioning that adding return type declaration to closures is an annoyance.

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
```diff
@@ -53,9 +55,6 @@
 
 /**
  * @group integration
- * @covers \Infection\TestFramework\PhpUnit\Coverage\IndexXmlCoverageParser
```
sanmai: We'll have to re-add this one line later if we'd want to practice what we preach about performance of MT.

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
```diff
@@ -65,14 +64,15 @@ final class IndexXmlCoverageParserTest extends TestCase
...
      */
     private $parser;
 
     protected function setUp(): void
     {
-        $this->parser = new LegacyXmlCoverageParserAdapter(
-            new IndexXmlCoverageParser(XmlCoverageFixtures::FIXTURES_COVERAGE_DIR)
+        $this->parser = new IndexXmlCoverageParser(
+            XmlCoverageFixtures::FIXTURES_COVERAGE_DIR,
+            new XmlCoverageParser()
```
sanmai: Again, dependence of IndexXmlCoverageParser on XmlCoverageParser makes this test integration, where it could have been pure unit.

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
theofidry: I'm not sure to understand you: `IndexXmlCoverageParser` is _already_ aware of `XmlCoverageParser`. The only change here is instead of making the later stateful, requiring it to be instantiated for each file info provider, it is now stateless

src/TestFramework/PhpUnit/Coverage/XmlCoverageParser.php
theofidry: It was mainly to get it of `lazilyRetrieveCoverageReport()` tbh... I was not fond of the name but could not come up with a better one and since it's a one-liner method...

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
theofidry: What do you mean? Remember we have the PHPUnit bridge listener which maps the `@covers`. It's not like we are on a free-well and it covers everything it's using.

Also each of those classes have their own dedicated test classes, so I don't think there is a need for this test class to cover them as well

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
```diff
@@ -241,49 +274,31 @@ public function noCoveredLineReportProviders(): Generator
...
-                    $incorrectCoverageSrcDir,
-                    Path::canonicalize(XmlCoverageFixtures::FIXTURES_COVERAGE_DIR)
-                ),
-                $exception->getMessage()
-            );
-            $this->assertSame(0, $exception->getCode());
-            $this->assertNull($exception->getPrevious());
+        foreach ($sourceFilesData as $data) {
+            /* @var SourceFileData $data */
+            $normalizedFilesData[$data->getSplFileInfo()->getRealPath()] = $data->retrieveCoverageReport();
```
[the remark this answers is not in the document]
theofidry: Not sure to follow you: that transformation here is to do the mapping which is not the job of `convertToArray()` IMO

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
sanmai: PHPUnit bridge listener does that for us? If so that's great.

theofidry: Yes: https://symfony.com/doc/current/components/phpunit_bridge.html#code-coverage-listener sorry I thought you were aware

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
sanmai: It does, but it shouldn't be. It's was a legacy decision made just to make `IndexXmlCoverageParser` be compatible with old tests. `IndexXmlCoverageParser`'s only true job is to produce `SourceFileInfoProvider`s, hence we have them in the first place.

src/TestFramework/PhpUnit/Coverage/XmlCoverageParser.php
sanmai: Well, for a reason, see? 

There's another. We don't use Psalm yet, but if we'd ever do, this will be another thing to document with types. Where `lazilyRetrieveCoverageReport()` was already documented properly.

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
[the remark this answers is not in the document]
sanmai: It did that before. The method's signature is unchanged, it still accepts two iterables. `convertToArray` will be happy to accept an iterable, as it was. So what's for these two arrays?

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
sanmai: Beautiful, thanks. Shall we not have an autoreview test to guard against this?

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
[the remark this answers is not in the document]
theofidry: I'm not doing the conversion for transforming it into an array, but for correcting the mapping because here what we compare is the coverage reports, not the source file info datas. This is an alternative to have an adapter and require to use the adapter everywhere in this test

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
theofidry: I honestly fail to see the issue here: I'm not trying to get a step further and correcting a legacy or anything, but just switch a service from being stateful to stateless which avoids to instantiate it several times

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
sanmai: We're already testing `XmlCoverageParser` in a dedicated test, so if to simplify `IndexXmlCoverageParser`,  this testing of `XmlCoverageParser` inside `IndexXmlCoverageParserTest` should be removed as redundant.

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
theofidry: I briefly tried but gave up. It would be nice to have fore sure. Note that `@covers` is required in some cases though so we cannot just blindly remove it

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
theofidry: I'm not changing anything there: I just got rid of the "LegacyAdapter" nothing more. I'm not saying there is further refactoring to do, e.g. test more those `CoverageReport` produced in `XmlCoverageParserTest` instead of `IndexXmlCoverageParserTest`, but this is not the goal of this PR

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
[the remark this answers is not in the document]
sanmai: This method was doing the same job before, and didn't buffer one of the outputs. Now it buffers both. This is inefficient.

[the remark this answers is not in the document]
theofidry: @sanmai I think you misread the code there. I'm not doing that for the buffering, but for the mapping which was previously done in the "LegacyAdapter""

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
sanmai: Well, then how could I guess that. We'll do the next later then.

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
sanmai: I wish I could read them, and we'd avoid this conversation.

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
theofidry: Yes sorry I should have tried to make it a bit more clear :)

theofidry:
I didn't not fully realised how skewed `IndexXmlCoverageParser` was. So the goal here is just to make a stateful service into a stateless one: a very tiny change.

I'll move some tests that are in `IndexXmlCoverageParserTest` to `XmlCoverageParser`, or maybe try to clear up the naming issue first mentioned in the TODOs...

PR #1157, opened 2020-03-12
Remove dependency on `OutputInterface` for the loggers

theofidry:
I don't think there is any reason for the loggers to depend on the output especially considering that:

- the output is used purely for logging purposes (in this context)
- it is in every way less flexible than `LoggerInterface` (you need to mock an output, for a logger you can put anything behind it)

Now I'll grant you it's not a big deal, but I find it a bit cleaner so here's the PR with the diff.


theofidry:
I'm sorry what?! For some reasons I can't see the Codecov report anymore without logging in and when I try to log in I have the following:

Screen Shot 2020-03-12 at 14 55 40

Why would codecov need my write commit permissions?

PR #1159, opened 2020-03-12
Show relative MSI

theofidry:
Replacement for #1066.

@sanmai I would like to double check if you think the approach is alright with you before going further

### DIFF
--- a/src/Mutant/MetricsCalculator.php
+++ b/src/Mutant/MetricsCalculator.php
@@ -72,6 +72,7 @@ class MetricsCalculator
      * @var int
      */
     private $totalMutantsCount = 0;
+
     private $killedExecutionResults;
     private $errorExecutionResults;
     private $escapedExecutionResults;
@@ -83,6 +84,11 @@ class MetricsCalculator
      * @var Calculator|null
      */
     private $calculator;
+
+    /**
+     * @var float|null
+     */
+    private $totalLineCodeCoverage;
 
     public function __construct()
     {
@@ -218,6 +224,16 @@ class MetricsCalculator
         return $this->allExecutionResults->getSortedExecutionResults();
     }
 
+    public function registerTotalCoverage(float $totalCoverage): void
+    {
+        $this->totalLineCodeCoverage = $totalCoverage;
+    }
+
+    public function getTotalLineCodeCoverage(): ?float
+    {
+        return $this->totalLineCodeCoverage;
+    }
+
     /**
      * Mutation Score Indicator (MSI)
      */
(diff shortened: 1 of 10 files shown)


src/TestFramework/PhpUnit/Coverage/PercentageParser.php
```diff
@@ -0,0 +1,38 @@
...
+
+        if ($percentage === '') {
+            return .0;
+        }
+
+        Assert::numeric($percentage);
+
+        $percentage = (float) $percentage;
+
+        return $percentage;
```
sanmai: Why not just cast to float? Any cases where this cast wouldn't work?
```
return (float) $percentage;
```

src/Mutant/MetricsCalculator.php
```diff
@@ -242,6 +255,14 @@ public function getCoveredCodeMutationScoreIndicator(): float
         return $this->getCalculator()->getCoveredCodeMutationScoreIndicator();
     }
 
+    /**
+     * Mutation Score Indicator relative to the total lines code coverage
+     */
+    public function getRelativeCoveredCodeMutationScoreIndicator(): float
+    {
+        return $this->getCalculator()->getRelativeCoveredCodeMutationScoreIndicator();
+    }
```
sanmai: Is this correct that this isn't used anywhere yet?

Is it going to be reported as a separate metric? It's okay to hide another in this case.

maks-rafalko:
Please explain the idea, goal, what does this PR change, probably some examples.

theofidry:
@maks-rafalko since not once seem very keen to jump on with #1064, @sanmai proposed to show that "MSI relative to code coverage" first to gauge if it's good enough or not.

I however wanted to check that the way this metric was retrieved made sense first before proceeding further

src/TestFramework/PhpUnit/Coverage/PercentageParser.php
theofidry: I missed your update on that comment, I mean the fact that it should just work with a cast. Will double check that

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParser.php
```diff
@@ -65,7 +69,9 @@ public function parse(string $coverageIndexPath, string $xmlIndexCoverageContent
     {
         $xPath = XPathFactory::createXPath($xmlIndexCoverageContent);
 
-        self::assertHasCoverage($xPath);
+        $totalCoverage = self::retrieveTotalCoverage($xPath);
+
+        $this->eventDispatcher->dispatch(new TotalLineCodeCoverageWasCalculated($totalCoverage));
```
sanmai: I cannot think of a better approach so I guess this'll go. I'm doing something very similar in #1159 FWIW.

theofidry: in what?

theofidry: Ah #1153

PR #1160, opened 2020-03-13
Simplify IndexXmlCoverageParser, remove FileCodeCoverageProviderFactory

sanmai:
This PR:

- [x] Unlinks IndexXmlCoverageParser from XmlCoverageParser; they no longer depend on each other
- [x] Removes FileCodeCoverageProviderFactory for no added value
- [x] Simplifies CoverageChecker to work without a constant
- [x] Covered by tests

Follow-up to #1156


tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageReaderTest.php
```diff
@@ -33,23 +33,25 @@
 
 declare(strict_types=1);
 
-namespace Infection\Tests\TestFramework\Coverage\XmlReport;
+namespace Infection\Tests\TestFramework\PhpUnit\Coverage;
```
maks-rafalko: We discussed it previously in many details, and I still think that coverage reader should not be in `TestFramework\PhpUnit` namespace, because it's used for all the test frameworks. So it's generic, not PHPUnit specific. 

Codeception generates xml report
PHPUnit
and PhpSpec does

sanmai: No prob, I'll move it a second.

sanmai: Right. The thing is I didn't move it in the first place. It is GitHub confuses it with `FileCodeCoverageProviderFactory` I removed. They were very similar, so this can be understood. Yet `IndexXmlCoverageReader` is exactly where `IndexXmlCoverageParser` is in the master.

sanmai: Please confirm. It is not a big deal to move around, but there should be no mistake if this is really required.

maks-rafalko: You are right. I didn't notice that moment when `IndexXmlCoverageParser` was placed under `PhpUnit` namespace.

So the changes in this PR looks good to me.

PS: my concern is based on the following: if we ever happen to extract PHPUnit out of the core, these classes should not be extract together with the test framework, because they are part of the core and used for all adapters

sanmai: I cannot say I agree with their current location either. Shall we move them then?

I'll open a PR.

maks-rafalko: 👍 

Thank you!

sanmai:
Thank you @maks-rafalko

src/Container.php
```diff
@@ -159,12 +160,15 @@ public static function create(): self
             IndexXmlCoverageParser::class => static function (self $container): IndexXmlCoverageParser {
                 return new IndexXmlCoverageParser(
                     $container->getConfiguration()->getCoveragePath(),
-                    new XmlCoverageParser()
                 );
             },
+            XmlCoverageParser::class => static function (self $container): XmlCoverageParser {
```
theofidry: `$container` is not used and can be removed

src/TestFramework/Coverage/CoverageChecker.php
```diff
@@ -222,4 +217,11 @@ private function isPcovIncludedInInitialTestPhpOptions(): bool
             $this->initialTestPhpOptions
         );
     }
+
+    private function getCoverageIndexFilePath(): string
+    {
+        return Path::canonicalize(
+            (new IndexXmlCoverageReader($this->coveragePath))->getIndexXmlPath()
```
theofidry: I feel this `IndexXmlCoverageReader` should be injected instead. Also since you are moving this in a method, might be worth memoizing that call

src/TestFramework/Coverage/XmlReport/PhpUnitXmlCoveredFileDataProvider.php
```diff
@@ -71,9 +70,12 @@ public function __construct(
...
     {
-        $coverageIndexPath = $this->coverageDir . '/' . self::COVERAGE_INDEX_FILE_NAME;
-        $coverageIndexContent = file_get_contents($coverageIndexPath);
-
-        return $this->parser->parse($coverageIndexPath, $coverageIndexContent);
+        foreach ($this->indexParser->parse(
+            $this->indexReader->getIndexXmlPath(),
+            $this->indexReader->getIndexXmlContent()
+        ) as $infoProvider) {
+            // TODO It might be benificial to filter files at this stage, rather than later. SourceFileDataFactory does that.
```
theofidry: I'm not sure to understand this TODO

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageReader.php
```diff
@@ -33,32 +33,33 @@
...
+    public function __construct(string $coverageDir)
+    {
         $this->coverageDir = $coverageDir;
-        $this->coverageXmlParser = $coverageXmlParser;
     }
 
-    public function create(): SourceFileDataProvider
+    public function getIndexXmlPath(): string
+    {
+        return $this->coverageDir . DIRECTORY_SEPARATOR . self::COVERAGE_INDEX_FILE_NAME;
```
theofidry: Can't we return a canonalized path here and memoize the call?

tests/phpunit/ContainerTest.php
```diff
@@ -109,4 +110,39 @@ public function test_it_can_build_dynamic_services(): void
 
         $this->addToAssertionCount(1);
     }
+
+    public function test_it_can_build_source_file_data_factory(): void
```
theofidry: I'm confused by that test: it can build but this ends as a fail?

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
```diff
@@ -273,31 +220,60 @@ public function noCoveredLineReportProviders(): iterable
...
-
-        foreach ($sourceFilesData as $data) {
-            /* @var SourceFileData $data */
-            $normalizedFilesData[$data->getSplFileInfo()->getRealPath()] = $data->retrieveCoverageReport();
+    /**
+     * @param iterable<XmlCoverageFixture> $fixtures
+     *
+     * @return Traversable<string>
+     */
+    private static function xmlCoverageFixturesToList(iterable $fixtures): Traversable
```
theofidry: if you pass an iterable is there a point to return a `Traversable` instead of `iterable`? Same for above

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageReaderTest.php
```diff
@@ -33,23 +33,25 @@
...
-final class FileCodeCoverageProviderFactoryTest extends TestCase
+/**
+ * @group integration
+ */
+final class IndexXmlCoverageReaderTest extends TestCase
 {
-    public function test_it_can_create_an_XMLLine_code_coverage_instance(): void
+    private const COVERAGE_DIR = __DIR__ . '/../../../Fixtures/Files/phpunit/coverage/coverage-xml';
+
+    public function test_it(): void
```
theofidry:
```suggestion
    public function test_it_can_give_the_path_to_the_PHPUnit_index_XML_report(): void
````

src/TestFramework/Coverage/CoverageChecker.php
sanmai: Yes, it'll good to inject it.

src/TestFramework/Coverage/XmlReport/PhpUnitXmlCoveredFileDataProvider.php
sanmai: Have a look at SourceFileDataFactory. There's another comment on this subject.

src/TestFramework/PhpUnit/Coverage/IndexXmlCoverageReader.php
sanmai: I'm sure we can, but for a method that is called only two times per run, is it worth it?

sanmai: OK, maybe four times if include CoverageChecker. I'm not so sure about memorization here.

tests/phpunit/ContainerTest.php
sanmai: Yes, because it returns a lazy iterator, which is only trigger a failure upon use.

tests/phpunit/TestFramework/PhpUnit/Coverage/IndexXmlCoverageParserTest.php
sanmai: Yes, because we use `iterator_to_array`, and it wants `Traversable`. It'll still work with `iterable`, but IIRC Psalm wasn't thrilled about it.

tests/phpunit/ContainerTest.php
theofidry: what I mean is that the test is a bit confusing, at least from it's title. Maybe adding a comment where the failure occurs as to why this is expected to fail?

sanmai: True. I'll rename it.

PR #1163, opened 2020-03-13
Improve usage of IndexXmlCoverageReader

theofidry:
- Inject the `IndexXmlCoverageReader`
- Canonicalize the index file path
- Calculate the index file path only once


sanmai:
AppVeyor isn't happy about different `DIRECTORY_SEPARATOR`

PR #1164, opened 2020-03-13
Rename LineCodeCoverage to Trace

theofidry:
After @sanmai 's PR #1106, I think it is time to introduce this new Tracer terminology. Reasons:

- It is a term used in mutant (at least the private version) so it provides common grounds with other mutation testing tools (well at least one...)
- I think it the term is understable enable enough: a trace is the association between a piece of source code and its associated tests
- It provides a good alternative to "covered" which is in our case horribly confusing: covered by tests, covered from a coverage report, covered by the mutation testing tool... And all the usages of "coverage" and "covered" coming from manipulating coverage reports

**This PR contains only naming changes.** There might be a few CS tweaks here and there, but no class has been removed, methods parameters changed or others.

There is a big TODO worth reading if you get confused as to why we have `ProxyTrace` in some places: IMO this is a leaking implementation detail that could be removed by slightly tweaking the `Trace` interface.

There is also a renaming change to do in the test framework adapter package. An update of it in infection/infection can be later though so it's not blocking for this PR.

I invite everyone to review this and I hope that you share the feeling that it helps to clarify some elements. At least to me, it revealed a few axes of improvements, e.g. the fact that we have currently two traces implementations depending on one another with a leaking abstraction.

I also added the nomenclature file, it's by no means finished, in fact I only added the tracing related terms. I think however I'll complete it more and more here instead of a random file locally. So for now it's more for internal usage and we can decide to publish it later, either to the website or to have it peer-reviewed with other mutation testing tool maintainers

### DIFF
--- a/src/FileSystem/SourceFileFilter.php
+++ b/src/FileSystem/SourceFileFilter.php
@@ -39,7 +39,9 @@ use function array_filter;
 use function array_map;
 use function explode;
 use Infection\FileSystem\Finder\Iterator\RealPathFilterIterator;
+use Infection\TestFramework\Coverage\ProxyTrace;
 use Iterator;
+use SplFileInfo;
 
 /**
  * @internal
@@ -73,9 +75,9 @@ class SourceFileFilter
     }
 
     /**
-     * @param iterable<object> $input
+     * @param iterable<SplFileInfo&ProxyTrace> $input
      *
-     * @return iterable<object>
+     * @return iterable<SplFileInfo&ProxyTrace>
      */
     public function filter(iterable $input): iterable
     {
@@ -91,9 +93,9 @@ class SourceFileFilter
     }
 
     /**
-     * @param iterable<object> $input
+     * @param iterable<SplFileInfo&ProxyTrace> $input
      *
-     * @return Iterator<object>
+     * @return Iterator<SplFileInfo&ProxyTrace>
      */
     private function iterableToIterator(iterable $input): Iterator
     {
--- a/src/FileSystem/SourceFileFilter.php
+++ b/src/FileSystem/SourceFileFilter.php
@@ -41,7 +41,7 @@ use function explode;
 use Infection\FileSystem\Finder\Iterator\RealPathFilterIterator;
 use Infection\TestFramework\Coverage\ProxyTrace;
 use Iterator;
-use SplFileInfo;
+use Symfony\Component\Finder\SplFileInfo;
 
 /**
  * @internal
--- a/src/FileSystem/SourceFileFilter.php
+++ b/src/FileSystem/SourceFileFilter.php
@@ -39,9 +39,7 @@ use function array_filter;
 use function array_map;
 use function explode;
 use Infection\FileSystem\Finder\Iterator\RealPathFilterIterator;
-use Infection\TestFramework\Coverage\SourceFileData;
 use Iterator;
-use Symfony\Component\Finder\SplFileInfo;
 
 /**
  * @internal
@@ -75,9 +73,9 @@ class SourceFileFilter
     }
 
     /**
-     * @param iterable<SplFileInfo&SourceFileData> $input
+     * @param iterable<object> $input
      *
-     * @return iterable<SplFileInfo&SourceFileData>
+     * @return iterable<object>
      */
     public function filter(iterable $input): iterable
     {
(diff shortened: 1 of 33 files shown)


src/TestFramework/Coverage/XmlReport/TestTrace.php
```diff
@@ -37,29 +37,38 @@
...
+ * its associated source files.
+ *
+ * TODO: FileMutationGenerator::generate() ends up having to rely on both SourceFileData & Trace due
+ *  to the fact that a Trace right now only exposes a test-related API
+ *  (`hasTests()`, `getAllTestsForMutation()`). Maybe a cleaner solution would be expose the
+ *  associated source file API to `Trace` as well, i.e. moving up `SourceFileData::getSplFileInfo()`
+ *  to `Trace`. This current class `TestsTrace` could then become a simple helper for `SourceFileData`
+ *  instead of a full-pledge `Trace` implementation. And FileMutationGenerator as a result would
+ *  also be simplified to not have to rely on SourceFileData which is `Trace` implementation and
+ *  depend on `Trace` only.
```
theofidry: Big TODO

src/FileSystem/SourceFileFilter.php
```diff
@@ -75,9 +73,9 @@ public function getFilters(): array
     }
 
     /**
-     * @param iterable<SplFileInfo&SourceFileData> $input
+     * @param iterable<object> $input
```
maks-rafalko: What's the reason of this change?

theofidry: I couldn't really see a required type downstream. I assume SplFileInfo? But at the end couldn't see it so I dont see the point to couple it to SourceFileData

theofidry:
>Seems like we need to get used to it and make this doc public eventually. Without explanation - it is really hard to understand what the Trace is.

I agree it is not intuitive at the very least. But I hope that it clears up some confusing with the sense of "coverage" that is used so much in our codebase yet designating very different things. That's also why I added the nomenclature!

>Also, having "Trace" and "TestTrace" is confusing, but as I understand this will be changed after fixing that TODO.

Definitely. I just didn't want to tackle it here since it's going to produce a few more changes I think

theofidry:
I'll wait on @sanmai review as well since it's touching a lot of stuff he did recently :)

src/FileSystem/SourceFileFilter.php
maks-rafalko: From what I see in `SourceFileDataProvider` (in master), it is only `iterable<SourceFileData>`.

More strict - better IMO.

theofidry:
But the filter itself is quite generic so is there a need to couple it to
that?

maks-rafalko:
> But the filter itself is quite generic so is there a need to couple it to that?

Filter class is named `SourceFileFilter`, isn't it supposed to filter only `SourceFileData`, but not all `object`s?

Having more strict API will help SA tools (with more strict configs) to check that we pass iterable of `SourceFileData` there, but not integers for example.

theofidry:
I guess I'll work on decoupling it if necessary in a separate PR

src/FileSystem/SourceFileFilter.php
sanmai: I like `iterable<SplFileInfo&SourceFileData>` better because it comes up when you look up usage. Else you will be looking at SourceFileData and wonder why is there this funny `getSplFileInfo` and where does it go? Thank you for getting it back.

src/FileSystem/SourceFileFilter.php
```diff
@@ -39,9 +39,9 @@
 use function array_map;
 use function explode;
 use Infection\FileSystem\Finder\Iterator\RealPathFilterIterator;
-use Infection\TestFramework\Coverage\SourceFileData;
+use Infection\TestFramework\Coverage\ProxyTrace;
 use Iterator;
-use Symfony\Component\Finder\SplFileInfo;
+use SplFileInfo;
```
sanmai: Different SplFileInfo

src/TestFramework/Coverage/XmlReport/TestTrace.php
sanmai: I'd argue merging them together will cause a conflict of concerns. The original Trace object has nothing to do with mutated files. As an alternative approach this proxy could be made a full-fledged trace provider, with hopefully minimal changes to the consumers.

sanmai: IIRC the reason this is proxy is convenience sake, no more.

theofidry:
@sanmai I'll do a PR and we'll see if it's gonna cause a conflict of concerns (in which case we won't adopt it an opt for a more appropriate name) or if it solves it. It should in any case be a better discussion starting point than this TODO

PR #1165, opened 2020-03-14
Upgrade to test-framework-adapter 0.3.0

theofidry:
This PR is strictly about this upgrade. There is further refactoring in `TestLocations` that I would like to carry on before tackling https://github.com/infection/infection/pull/1164#issuecomment-599021929 but this will be done in a separate PR for better clarity and ease of review.


theofidry:
And damnit all test frameworks need to be updated

maks-rafalko:
https://travis-ci.org/github/infection/infection/jobs/662330996#L473

I would say this is a nice error for PHAR. Indeed, when Infection is bundled with all external adapters, they should use 1 compatible interface. 

On the other side, currently, adapters can work with the "previous" version of `infection/infection` before upgrading, which for me proves that evering is ok package-wise.

> And damnit all test frameworks need to be updated

I assume you like it! 🤣 

---

LGTM as long as build is fixed

PR #1166, opened 2020-03-14
ProxyTrace isn't a Trace

sanmai:
Apparently nothing calls these methods.


theofidry:
I actually would like it to be, but I’m also not certain that my PR will
work out as I desire so depending of the result, this PR might be more
appropriate.

So I’m +1 for the change but I would like this PR to wait a bit to see how
the other direction goes

PR #1167, opened 2020-03-15
Tweak TestLocations

theofidry:
Followup of #1164 and #1165.

- Rename `MethodLocationData` to `SourceMethodRange`
- Consolidate the tests of `TestLocations`
- Remove the comment of `TestLocations` in favour of a more self-describing test
- Rename `TestLocation#byLine` to `testLocationsBySourceLine`
- Rename `TestLocation#byMethod` to `sourceMethodRangeByMethod`


theofidry:
ping @sanmai ?

tests/phpunit/TestFramework/Coverage/TestLocationsNormalizerTest.php
```diff
@@ -136,7 +138,7 @@ public function locationsProvider(): iterable
                         ],
                     ],
                     [
-                        '__construct' => new MethodLocationData(
+                        '__construct' => new SourceMethodRange(
```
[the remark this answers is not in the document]
theofidry: 👍

src/TestFramework/Coverage/JUnit/JUnitTestExecutionInfoAdder.php
```diff
@@ -83,11 +83,12 @@ public function addTestExecutionInfo(iterable $traces): iterable
...
     {
         foreach ($traces as $trace) {
-            foreach ($trace->retrieveTestLocations()->byLine as &$testsLocations) {
+            foreach ($trace->retrieveTestLocations()->getTestsLocationsBySourceLine() as &$testsLocations) {
                 foreach ($testsLocations as $line => $test) {
-                    $testsLocations[$line] = self::createCompleteTestLocation($test);
+                    $testsLocations[$line] = $this->createCompleteTestLocation($test);
                 }
             }
+            unset($testsLocations);
```
sanmai: This loop with references is a trouble brewing, but I guess it's another time to fix.

src/TestFramework/Coverage/TestLocations.php
```diff
@@ -39,48 +39,41 @@
...
+        array $testLocationsBySourceLine = [],
+        array $sourceMethodRangeByMethod = []
+    ) {
+        $this->byLine = $testLocationsBySourceLine;
+        $this->byMethod = $sourceMethodRangeByMethod;
+    }
 
     /**
-     * TODO: use a getter "get"
+     * This method needs to be able to return a reference for performance reasons.
```
sanmai: I wonder if there's actual penalty here.

theofidry: Definitely as otherwise you would create a new array reference at the very least here and would need to create a new object upstream to assign the modified one. That said since the content is all objects, their references are kept and not copied, so maybe the performance impact won't be as much.

Might be worth to revisit

PR #1168, opened 2020-03-15
Enrich Trace

theofidry:
Follow up of #1167 and alternative to #1166 

- Enrich the `Trace` interface to be aware of:
  - The source file information
  - allow to list all the known tests (instead of just "has test" and "all tests for given node range" which it current has)
- Rename `TestTrace` as `TestLocator` and use it as an internal utility in `ProxyTrace`

Be judge with the diff after a rebase on #1167. IMO it's a good compromise of what we had before:

- we have a simpler trace than before although more rich, making the consuming API simpler
- we still take advantage of the previous `TestTrace` (which is far from being trivial) to not bloat `ProxyTrace`


src/TestFramework/Coverage/ProxyTrace.php
```diff
@@ -79,15 +79,12 @@ public function __construct(SplFileInfo $sourceFile, iterable $lazyTestLocations
...
+    public function getSourceFileInfo(): SplFileInfo
     {
         return $this->sourceFile;
     }
 
-    /**
-     * Used by RealPathFilterIterator
-     */
-    public function getRealPath(): string
+    public function getSourceRealPath(): string
```
sanmai: `RealPathFilterIterator` wants exactly `getRealPath()` because it expects `SplFileInfo`. Here we're duck-typing `ProxyTrace` to match.

theofidry: I see, welp definitely need to add a test for that then

sanmai: Alternatively we can update `RealPathFilterIterator`'s `current()` method to handle these differences for us. It'll require yet another SplFileInfo-like wrapper, and tbh I can't say it looks exactly better the duck-typing approach we take here.

theofidry: The duck-typing is fine it just needed a test which was not hard to do based on your code :)

tests/phpunit/FileSystem/SourceFileFilterTest.php
```diff
@@ -37,10 +37,11 @@
 
 use function array_values;
 use Infection\FileSystem\SourceFileFilter;
+use Infection\TestFramework\Coverage\Trace;
 use IteratorIterator;
 use PHPUnit\Framework\TestCase;
 use function Pipeline\take;
-use Symfony\Component\Finder\SplFileInfo;
+use SplFileInfo;
```
sanmai:
```suggestion
use Symfony\Component\Finder\SplFileInfo;
```

tests/phpunit/TestFramework/Coverage/TestLocationsTest.php
```diff
@@ -109,8 +109,8 @@ public function test_it_can_expose_its_tests_locations_by_reference(): void
                 ],
             ],
             [
-                'mutate' => new SourceMethodLineRange(12, 16),
-                'createNode' => new SourceMethodLineRange(32, 33),
+                'mutate' => new SourceMethodRange(12, 16),
+                'createNode' => new SourceMethodRange(32, 33),
```
sanmai: SourceMethodRange sneaked through

tests/phpunit/FileSystem/SourceFileFilterTest.php
theofidry: I would actually prefer `SplFileInfo` to be honest: it's like when you have a class coupled to a type, it if works with the type interface it's better to typehint the interface (provided that doesn't complexify things).

In other words: since it doesn't cost anything, I don't see why we should couple `SourceFileFilter` to `Symfony\Component\Finder\SplFileInfo` instead of `SplFileInfo`

tests/phpunit/TestFramework/Coverage/TestLocationsTest.php
theofidry: Good catch

tests/phpunit/FileSystem/SourceFileFilterTest.php
sanmai: `\SplFileInfo` is definitely better if we're not calling Symfony's version methods.

PR #1170, opened 2020-03-16
Delegate all filtering operations to FilteredEnrichedTraceProvider

sanmai:
Currently we do filtering for `--filter=foo` in two fairly distant places: in `SourceFileCollector` and in `FilteredEnrichedTraceProvider`, making type requirements for `SourceFileFilter` less obvious. Here  we move all filtering responsibility to `FilteredEnrichedTraceProvider`.

This PR:

- [x] Simplifies SourceFileCollector
- [x] Updates FilteredEnrichedTraceProvider
- [x] Covered by tests

Overall we add only about three to four lines in terms of code, and two lines in terms of tests, while removing quite a bit of redundant integration testing.


src/FileSystem/SourceFileCollector.php
```diff
@@ -67,21 +63,9 @@ public function collectFiles(
             ->exclude($excludeDirectories)
             ->in($sourceDirectories)
             ->files()
+            ->name('*.php')
         ;
 
-        if ($filter === '') {
-            $finder->name('*.php');
-        } else {
-            $finder->filterFiles(
```
maks-rafalko: 1. Am I right saying this moving won't impact performance / memory usage? Is it possible that any code in the middle (between `SourceFileCollector` and `FilteredEnrichedTraceProvider` can use this files results, without filters applied?

2. Seems like `FilterableFinder::filterfiles()` is no longer used (and `FilterableFinder` in general)

sanmai: 1. FilterableFinder uses the same RealPathFilterIterator we used in `FilteredEnrichedTraceProvider` before to filter coverage-sourced files, therefore we have no reason to believe the performance will be affected. 
2. `FilterableFinder` is indeed no longer needed. I'll remove it.

maks-rafalko: Sorry, for performance question, I meant this

> Is it possible that any code in the middle (between `SourceFileCollector` and `FilteredEnrichedTraceProvider` can use these file results, without filters applied?

Just want to be sure not filerered collection of files won't be used anywhere before `FilteredEnrichedTraceProvider`

sanmai: I made `SourceFileCollector` return a generator, so we can be 100% certain it is not going to be used more than once.

maks-rafalko: 👍

sanmai:
Thank you @maks-rafalko

tests/phpunit/FileSystem/SourceFileCollectorTest.php
```diff
@@ -109,107 +102,9 @@ public function sourceFilesProvider(): iterable
             ],
         ];
 
-        yield 'filter by file name, no excludes' => [
```
theofidry: We're delegating it to `FilteredEnrichedTraceProvider` but it feels like the later is not as extensively tested as the previous version of `SourceFileCollector`... Should we add back those tests to `FilteredEnrichedTraceProvider`?

sanmai: `RealPathFilterIteratorTest` has these covered already IIRC

sanmai: Anyway, there isn't any apparent loss of coverage or MSI degradation.

sanmai: And, yes, `SourceFileCollectorTest` has a fairly extensive list of fixtures, which look somewhat similar to what we have removed here.

theofidry: 👍

PR #1171, opened 2020-03-16
Exclude mutations that are over specified time limit

sanmai:
Looking at Psalm's median testing time (#1153) of about 10 minutes, and looking at the anticipated number of timed out mutations for these circumstances, I can't help but notice that since we know about how much time a mutation needs to be tested thoroughly, we can exclude these definitely out -of-time-budget mutations from even running.

This change considerably speeds up MT for projects with incoherent test coverage, while also lowering detection rates (which is bad), while leaving the door open either to raise the timeout, or to factor in `@covers` annotations (which is good).

For Psalm I get 6m 17s. of runtime with `-j2` on a complete test suite. Before it was unfeasible to run.

### Psalm results for `@default` set

```json
{
    "timeout": 15,
    "source": {
        "directories": [
            "src"
        ]
    },
    "logs": {
        "text": "build\/infection.log"
    },
    "mutators": {
        "@default": true
    }
}

```  
  
```
.: killed, M: escaped, U: uncovered, E: fatal error, T: timed out, S: skipped

...SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSMSSSSSSSSSSSSSS   (   50)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS   (  100)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS   (  150)

...

4441 mutations were generated:
      47 mutants were killed
      18 covered mutants were not detected
       0 errors were encountered
      79 time outs were encountered
    4297 mutants required more time than configured

Metrics:
         Mutation Score Indicator (MSI): 87%
         Mutation Code Coverage: 100%
         Covered Code MSI: 87%

Time: 6m 16s. Memory: 1.28GB
```

This PR:

- [x] Adds time limit limiting
- [x] Blocked by #1233 
- [x] Covered by tests 
- [x] Doc PR: https://github.com/infection/site/pull/167 (just a mention)

Related: #1145,  #1133


theofidry:
How do you think those should be reported?

I've been thinking for a while that we might need a `DetectionStatus::SKIPPED` in which case providing a reason would be appropriate

sanmai:
I don't think it makes sense to report them differently in the console report (`T` is fine), but having them as a different number in the final report is a good idea, UX-wise.

sanmai:
@maks-rafalko what's your take on this idea?

sanmai:
We can report these time-constrained mutations as, naturally, `C` for constrained.

src/Mutation/Mutation.php
```diff
@@ -164,6 +166,18 @@ public function getAllTests(): array
...
     }
 
+    /**
+     * Overall time needed to run known tests for a mutation, excluding dependencies.
+     */
+    public function getTimeToTest(): float
+    {
+        // TODO this needs to be memoized
+
+        return array_sum(array_map(static function (TestLocation $data) {
```
maks-rafalko: what do you think about renaming `$data` to `$test`?

theofidry:
Also not 100% sure to understand the relationship with the configurable timeout or how it should be

sanmai:
> Should the user of Infection treat it like a call to action? 

Most definitely. We'll report these `C` mutations mentioning the configured timeout:

```
4441 mutations were generated:
      47 mutants were killed
      18 covered mutants were not detected
       0 errors were encountered
      79 time outs were encountered
    4297 mutants required more time than configured

Metrics:
         Mutation Score Indicator (MSI): 87%
         Mutation Code Coverage: 100%
         Covered Code MSI: 87%

Time: 6m 20s. Memory: 1.28GB
```
So that a user would (hopefully) see these mutations were constrained by the timeout they set. Next they either raise a timeout, or be directed to the docs for further instruction.

Note this above was a run over Psalm, and it took only modest 6 minutes.

sanmai:
> Also not 100% sure to understand the relationship with the configurable timeout or how it should be

This is the user choice. If they know they have several very long tests, they can sacrifice some mutations by shortening time needed for the MT process.

E.g. if one has a 15 second long integration test, setting time limit to 10 second will effectively exclude all mutations covered by the test. Now one has a choice of either raising the timeout, or tagging this test with `@coversNothing`.

src/Console/OutputFormatter/DotFormatter.php
```diff
@@ -66,7 +66,8 @@ public function start(int $mutationCount): void
             . '<escaped>M</escaped>: escaped, '
             . '<uncovered>S</uncovered>: uncovered, '
             . '<with-error>E</with-error>: fatal error, '
-            . '<timeout>T</timeout>: timed out',
+            . '<timeout>T</timeout>: timed out, '
+            . '<constrained>C</constrained>: constrained',
```
theofidry: I admit I'm not sold on the term "constrained". I feel like having "Skipped" with an explanation as to why (e.g.: not tested, timeout exceed limit, etc.) would be both more expressive and clearer: the user would not have to wonder _what_ "constrained" means, but rather wonder _why_ the mutation was skipped.

This also fits better for future changes where say we skip mutations through another optimization strategy. The typical example being the top-down strategy: you skip the nested mutations if the top one escaped. When implementing it, we would only have to use the existing "skipped" and provide another reason. Whereas if we go with "constrainted", it's really unclear what we should do there

src/Mutation/Mutation.php
```diff
@@ -164,6 +166,16 @@ public function getAllTests(): array
         return $this->tests;
     }
 
+    /**
+     * Overall time needed to run known tests for a mutation, excluding dependencies.
+     */
+    public function getTimeToTest(): float
+    {
+        return $this->timeToTest ?? $this->timeToTest = array_sum(array_map(static function (TestLocation $testLocation) {
```
theofidry:
```suggestion
        return $this->timeToTest ?? $this->timeToTest = array_sum(array_map(
            static function (TestLocation $testLocation) {
                return $testLocation->getExecutionTime();
            }, 
            $this->tests
        ));
```

src/Mutation/Mutation.php
```diff
@@ -164,6 +166,16 @@ public function getAllTests(): array
         return $this->tests;
     }
 
+    /**
+     * Overall time needed to run known tests for a mutation, excluding dependencies.
+     */
+    public function getTimeToTest(): float
```
theofidry:
```suggestion
    public function getTotalTestExecutionTime(): float
```

src/Process/Builder/MutantProcessFactory.php
```diff
@@ -58,7 +58,7 @@ class MutantProcessFactory
     // TODO: is it necessary for the timeout to be an int?
     public function __construct(
         TestFrameworkAdapter $testFrameworkAdapter,
-        int $timeout,
+        float $timeout,
```
theofidry: I would not do that change in this PR and require the `Configuration::getTimeOut()` to return a `float`, but if you wish to I can do it

src/Process/Runner/MutationTestingRunner.php
```diff
@@ -100,6 +103,17 @@ public function run(iterable $mutations, string $testFrameworkExtraOptions): voi
 
                 return false;
             })
+            ->filter(function (Mutant $mutant) {
```
theofidry: any reason why this should be done in another `filter()` call instead of the previous one?

src/Console/OutputFormatter/DotFormatter.php
sanmai: OK, I don't mind "skipped", but we used `S` for uncovered. How do you propose we tag these skipped in the progress log?

theofidry: `U` for uncovered? (should be changed in a different PR though)

src/Process/Builder/MutantProcessFactory.php
sanmai: It doesn't require. PHP will be happy to [cast int to float for us](https://3v4l.org/pHN1K).

src/Mutation/Mutation.php
sanmai: This is a tricky moment. We can't say it is execution time because one may think this is actual test execution time, where it isn't. For example, this timing does not include dependent tests, therefore actual execution time might be larger.

src/Process/Runner/MutationTestingRunner.php
sanmai: Yes, this distinction points at a refactoring angle. E.g. you can factor out this filter into another filter-adding class or method to test it separately.

src/Mutation/Mutation.php
theofidry: Is there a way we could ignore those dependent tests when executing them?

sanmai: [I don't think this is feasible.](https://phpunit.readthedocs.io/en/9.0/writing-tests-for-phpunit.html#writing-tests-for-phpunit-examples-stacktest2-php)

theofidry: `getTestExecutionTime` then? `time to test` sounds a bit weird, and I think keeping a comment to mention it does not count the dependent tests is fine

sanmai: Renamed into `getNominalTestExecutionTime`

src/Console/OutputFormatter/DotFormatter.php
sanmai: OK, prints `S: skipped` now

sanmai:
@maks-rafalko the build is green

sanmai:
We're mentioning this new class of mutations here in the final report:

```
4441 mutations were generated:
...
    4297 mutants required more time than configured
```

You're right that docs would need more than just a mention. How do you think if we'd get back at this closer to the release time? 

Another place where this new feature should be mentioned is the configuration master. It does not have a step to configure the timeout, where it really could. And there we can explain what's the deal with timeouts in great detail. I think I can get to this next thing.

maks-rafalko:
> How do you think if we'd get back at this closer to the release time?

Sounds good. Will need your help though ;)

> Another place where this new feature should be mentioned is the configuration master.

this is also a good idea.

Thank you, @sanmai.

src/Mutation/Mutation.php
```diff
@@ -62,6 +63,10 @@ class Mutation
     private $originalFileAst;
     private $tests;
     private $coveredByTests;
+    /**
+     * @var float|null
+     */
+    private $nominalTimeToTest;
```
DaveLiddament, line 69: What is the unit for this time, seconds, milliseconds, etc? 

Suggest:
1. update name to `nominalTimeInSecondsToTest` or whatever unit actually is.
2. or better still create a value object `Seconds`, 'MiliSeconds` to aid clarity here.

sanmai: It's all seconds here.

src/Process/Runner/MutationTestingRunner.php
```diff
@@ -100,6 +103,17 @@ public function run(iterable $mutations, string $testFrameworkExtraOptions): voi
...
+                if ($mutant->getMutation()->getNominalTestExecutionTime() < $this->timeout) {
```
sanmai, line 107: @danepowell if we add return false above this line we should be able to see that we are getting significantly better results if your issue https://github.com/infection/infection/issues/1965 is indeed a bug

PR #1172, opened 2020-03-17
Optimize JUnit test lookups to stop on the first element

sanmai:
Offers modest [3.5% improvement](https://blackfire.io/profiles/compare/6022a84b-a6f0-4f3d-865e-1443cbc69585/graph) on an example project with 1.4M junit.xml.

### DIFF
--- a/src/TestFramework/Coverage/JUnit/JUnitTestExecutionInfoAdder.php
+++ b/src/TestFramework/Coverage/JUnit/JUnitTestExecutionInfoAdder.php
@@ -48,6 +48,9 @@ use Infection\TestFramework\Coverage\Trace;
  */
 class JUnitTestExecutionInfoAdder
 {
+    /**
+     * @var TestFileDataProvider|MemoizedTestFileDataProvider|JUnitTestFileDataProvider
+     */
     private $testFileDataProvider;
 
     private $adapter;
--- a/src/TestFramework/Coverage/JUnit/JUnitTestExecutionInfoAdder.php
+++ b/src/TestFramework/Coverage/JUnit/JUnitTestExecutionInfoAdder.php
@@ -49,7 +49,7 @@ use Infection\TestFramework\Coverage\Trace;
 class JUnitTestExecutionInfoAdder
 {
     /**
-     * @var TestFileDataProvider|MemoizedTestFileDataProvider|JUnitTestFileDataProvider
+     * @var TestFileDataProvider
      */
     private $testFileDataProvider;
 
--- a/src/TestFramework/Coverage/JUnit/JUnitTestFileDataProvider.php
+++ b/src/TestFramework/Coverage/JUnit/JUnitTestFileDataProvider.php
@@ -80,6 +80,8 @@ final class JUnitTestFileDataProvider implements TestFileDataProvider
             );
         }
 
+        Assert::same($nodes->length, 1);
+
         return new TestFileTimeData(
             $nodes[0]->getAttribute('file'),
             (float) $nodes[0]->getAttribute('time')
--- a/src/TestFramework/Coverage/JUnit/JUnitTestFileDataProvider.php
+++ b/src/TestFramework/Coverage/JUnit/JUnitTestFileDataProvider.php
@@ -78,6 +78,8 @@ final class JUnitTestFileDataProvider implements TestFileDataProvider
             }
         }
 
+        Assert::notNull($nodes);
+
         if ($nodes->length === 0) {
             throw TestFileNameNotFoundException::notFoundFromFQN(
                 $fullyQualifiedClassName,
--- a/src/TestFramework/Coverage/JUnit/JUnitTestFileDataProvider.php
+++ b/src/TestFramework/Coverage/JUnit/JUnitTestFileDataProvider.php
@@ -90,14 +90,14 @@ final class JUnitTestFileDataProvider implements TestFileDataProvider
 
     private static function testCaseMapGenerator(string $fullyQualifiedClassName): iterable
     {
-        // Similar format for <testsuite>
-        yield '(//testsuite[@name="%s"])[1]' => $fullyQualifiedClassName;
+        // A default format for <testsuite>
+        yield '//testsuite[@name="%s"][1]' => $fullyQualifiedClassName;
 
         // A format where the class name is inside `class` attribute of `testcase` tag
-        yield '(//testcase[@class="%s"])[1]' => $fullyQualifiedClassName;
+        yield '//testcase[@class="%s"][1]' => $fullyQualifiedClassName;
 
         // A format where the class name is inside `file` attribute of `testcase` tag
-        yield '(//testcase[contains(@file, "%s")])[1]' => preg_replace('/^(.*):+.*$/', '$1.feature', $fullyQualifiedClassName);
+        yield '//testcase[contains(@file, "%s")][1]' => preg_replace('/^(.*):+.*$/', '$1.feature', $fullyQualifiedClassName);
     }
 
     private function getXPath(): SafeDOMXPath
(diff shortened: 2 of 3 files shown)


src/TestFramework/Coverage/JUnit/JUnitTestFileDataProvider.php
```diff
@@ -91,12 +80,26 @@ public function getTestFileInfo(string $fullyQualifiedClassName): TestFileTimeDa
...
 
+        Assert::same($nodes->length, 1);
+
         return new TestFileTimeData(
             $nodes[0]->getAttribute('file'),
             (float) $nodes[0]->getAttribute('time')
         );
     }
 
+    private static function testCaseMapGenerator(string $fullyQualifiedClassName): iterable
```
theofidry:
```suggestion
    /**
     * @return iterable<string, string>
     */
    private static function testCaseMapGenerator(string $fullyQualifiedClassName): iterable
```
? (not entirely sure it's the syntax for PHPStan & Psalm)

src/TestFramework/Coverage/JUnit/JUnitTestExecutionInfoAdder.php
```diff
@@ -48,6 +48,9 @@
  */
 class JUnitTestExecutionInfoAdder
 {
+    /**
+     * @var TestFileDataProvider|MemoizedTestFileDataProvider|JUnitTestFileDataProvider
```
theofidry: I don't understand this change: why should `JUnitTestExecutionInfoAdder` care about what implementation can be used here? It's using a `TestFileDataProvider ` that's all it needs to know right?

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestFileDataProviderTest.php
```diff
@@ -110,4 +110,43 @@ public function test_it_works_with_feature_junit_format(): void
 
         $this->assertSame('/codeception/tests/bdd/FeatureA.feature', $testFileInfo->path);
     }
+
+    /**
+     * @dataProvider xmlProvider
+     */
+    public function test_it_does_not_trigger_count_assertion(string $xml): void
+    {
+        file_put_contents($this->tempfile, $xml);
```
theofidry: Missing an import (`Safe\file_put_contents`)

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestFileDataProviderTest.php
```diff
@@ -110,4 +110,43 @@ public function test_it_works_with_feature_junit_format(): void
...
+
+        $provider = new JUnitTestFileDataProvider($this->tempfile);
+        $provider->getTestFileInfo('ExampleTest');
+        $this->addToAssertionCount(1);
+    }
+
+    public static function xmlProvider(): iterable
```
theofidry: PHPUnit data providers don't need to be static

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestFileDataProviderTest.php
```diff
@@ -110,4 +110,43 @@ public function test_it_works_with_feature_junit_format(): void
 
         $this->assertSame('/codeception/tests/bdd/FeatureA.feature', $testFileInfo->path);
     }
+
+    /**
+     * @dataProvider xmlProvider
+     */
+    public function test_it_does_not_trigger_count_assertion(string $xml): void
```
theofidry: Not sure to understand the test case

src/TestFramework/Coverage/JUnit/JUnitTestFileDataProvider.php
```diff
@@ -65,23 +65,12 @@ public function getTestFileInfo(string $fullyQualifiedClassName): TestFileTimeDa
...
-            // try another format where the class name is inside `file` attribute of `testcase` tag
-            $nodes = $xPath->query(
-                sprintf('//testcase[contains(@file, "%s")]', $feature)
-            );
+            if ($nodes->length !== 0) {
+                break;
+            }
         }
 
         if ($nodes->length === 0) {
```
theofidry: Scrutinizer:

>The variable $nodes seems to be defined by a foreach iteration on line 68. Are you sure the iterator is never empty, otherwise this variable is not defined?

src/TestFramework/Coverage/JUnit/JUnitTestExecutionInfoAdder.php
sanmai: Pure convenience. It's a one click jump for my IDE.

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestFileDataProviderTest.php
sanmai: Broken query triggers an assertion I added in the tested class.

src/TestFramework/Coverage/JUnit/JUnitTestFileDataProvider.php
sanmai: Yeah, I'll fix these once I get done with git bisect from #1173

src/TestFramework/Coverage/JUnit/JUnitTestFileDataProvider.php
sanmai: We're 100% sure about that.

src/TestFramework/Coverage/JUnit/JUnitTestExecutionInfoAdder.php
theofidry: It is necessary to add it then? Because we don't do that anywhere else and I think it's more confusing than anything

sanmai: I'd add `@see` in this case, would you mind? 

It really saves my day.

theofidry: isn't going to the interface then the implementation just 1 more click away?

The reason I'm not fond of it is that we have absolutely no mean to keep it in sync and it will get out of sync, adding to the confusion

sanmai: Not in my IDE.

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestFileDataProviderTest.php
sanmai: Neither it needs to access `$this`. Isn't the same rule applies here as in case of static closures?

theofidry: I guess, but IIRC PHPUnit is not attempting to call it statically so I don't think there is any benefit (whereas for the static closure there is a slight perf bonus)

src/TestFramework/Coverage/JUnit/JUnitTestExecutionInfoAdder.php
theofidry: That's for sure, but I also don't feel like this is a reliable way to do this and unless we have a tooling to automate this, nobody else in the team is really paying attention to this so it will get out of sync and become incorrect or missing in 99% of the places

I'm happy to let it here (as long as it doesn't add an import, so FQCN might be needed) but just want to be clear: that is neither reliable or sustainable in its current form at least

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestFileDataProviderTest.php
sanmai: Even if you call something as something non-static, it will still have a static scope.

PR #1177, opened 2020-03-17
Benchmark for JUnitTestCaseSorter::uniqueByTestFile

sanmai:
This PR:

- [x] Adds a bechmark for JUnitTestCaseSorter, based on live data
- [x] Adds optional bucket sort
- [x] Covered by tests
- [ ] Is `isset()` required for PHP 7.4?
- [ ] Standard benchmark profile

Looking just at the [Big O notation for the bucket sort](https://www.growingwiththeweb.com/2015/06/bucket-sort.html), it seems to be more effective in general, but for smaller number of items and/or for larger number of buckets [Quicksort is more effective](https://en.wikipedia.org/wiki/Quicksort#Average-case_analysis), again, looking at the notation only. 

Here's a Big O graph for 23 buckets:

graph for 23 buckets

Blue is for quicksort. Orange is for bucket sort.

For 25 buckets we're aiming the sort becomes theoretically more effective after 15 elements. Considering that we don't do the second stage of intra-bucket sorting, this effectiveness boundary might be closer for us, yet still far.

If we consider the distribution of overly tested mutations, we can see that our bucket sort will be applied in roughly 40% of cases in case of Psalm.

## Bucket selection algorithm

Since effectiveness of the algorithm is in direct relationship with the number of buckets, and since our data is effectively skewed towards a specific mode at around 1/3 of a second, it does not help if we use a fixed number of buckets. 

Test time distribution gistagram

Instead we gradually lower precision of our timings starting with 8th of a second, later lowering it to 4 seconds. Also we pre-sort bucket array for the first second worth of buckets. By using calculated buckets we try to avoid looping around bucket numbers.

<details>
  <summary>Example of a bucket distribution</summary>

| Bucket | Test times |
| ----- | ---- |
| 0 | 0.01, 0.08, 0.06 ... everything under 1/8 of a second goes here |
| 1 |  (nothing fell here, skipping these hereafter)
| 2 | 0.26, 0.31, 0.28, 0.36, 0.35 |
| 4 | 0.56, 0.53 |
| 5 | 0.75 |
| 7 | 0.99, 0.95, 0.94 |
| 8 | 1.06 |
| 10 | 1.25 |
| 11 | 1.45, 1.44 |
| 14 | 1.8, 1.86 |
| 16 | 2.11, 2.05 |
| 18 | 2.28 |
| 19 | 2.45 |
| 23 | 2.93 |
| 24 | 3.03 |
| 25 | 3.17, 3.21 |
| 28 | 3.62, 3.53 |
| 30 | 3.75 |
| 32 | 7.81, 5.93, 4.42, 4.37, 4.22, 6.09, 4.27, 4.4, 6.75, 7.11, 4.93, 6.79, 6.1, 4.95, 6.66, 7.88, 7.69, 5.24, 5.8, 5.58, 5.99 |
| 64 | 8.57, 9.34, 8.92, 8.97, 8.22, 11.77, 8.61, 9.98, 9.13 |
| 96 | 15.25, 12.75, 12.48, 12.9, 13.7, 15.06, 14.85, 14.22, 15.08, 12.22 |
| 128 | 19.57, 17.9, 17.94, 18.89 |
| 160 | 20.73, 23.94, 22.72, 20.55, 20.31 |
| 192 | 26.77 |
| 224 | 29.58 |
| 256 | 32.26, 32.39 |
| 352 | 45.48, 46.91 |

</details>

Further optimization begs the following questions:
- Should we care about tests longer than one seconds? E.g. if not, we can leave only several buckets under one second, and leave everything else above one second unsorted.
- Aren't we overfitting our optimization procedure to Psalm only? Might be better to verify our approach against a greater dataset.
- Is this really the bottleneck we should care about? (No doubt this was a fun exercise.)


src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
```diff
@@ -36,15 +36,35 @@
...
+    /**
+     * For 25 buckets QS becomes theoretically less efficient on average at and after 15 elements.
+     * Exposed for testing purposes.
+     */
+    public const USE_BUCKET_SORT_AFTER = 15;
+
+    /**
+     * Milliseconds in a second. То stop Infection from mutating a constant;
+     */
+    private const MS_IN_S = 1000;
```
theofidry: Doesn't appear to be used

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
```diff
@@ -98,9 +156,11 @@ private function uniqueByTestFile(array $testLocations): array
         $uniqueTests = [];
 
         foreach ($testLocations as $testLocation) {
+            /** @var string $filePath */
             $filePath = $testLocation->getFilePath();
 
-            if (!array_key_exists($filePath, $usedFileNames)) {
+            // isset() is 20% faster than array_key_exists()
+            if (!isset($usedFileNames[$filePath])) {
```
theofidry: I thought this was no longer the case on PHP 7.3?

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
```diff
@@ -98,9 +156,11 @@ private function uniqueByTestFile(array $testLocations): array
         $uniqueTests = [];
 
         foreach ($testLocations as $testLocation) {
+            /** @var string $filePath */
```
theofidry: on theory there is no guarantee that this is non-null, and even if is it won't mess up the code bellow either, so can be removed?

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
```diff
@@ -74,15 +93,54 @@ public function getUniqueSortedFileNames(array $tests): iterable
...
 
-        foreach ($uniqueTestLocations as $testLocation) {
-            yield $testLocation->getFilePath();
+            $buckets[$msTime][] = $location;
+        }
+
+        ksort($buckets);
+
+        foreach ($buckets as $bucket) {
+            foreach ($bucket as $value) {
```
theofidry:
```suggestion
            yield from $bucket;
```

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
```diff
@@ -74,15 +93,54 @@ public function getUniqueSortedFileNames(array $tests): iterable
...
+            2 => [],
+            3 => [],
+            4 => [],
+            5 => [],
+            6 => [],
+            7 => [],
+        ];
+
+        foreach ($uniqueTestLocations as $location) {
+            // Quicky drop off lower bits, reducing precision to 8th of a second
```
theofidry:
```suggestion
            // Quick drop off lower bits, reducing precision to 8th of a second
```

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -96,4 +100,113 @@ public function test_it_returns_unique_and_sorted_by_time_test_cases(): void
...
+        // Sanity check
+        $this->assertNotTrue(self::orderConstraintsValid($uniqueTestLocations));
+
+        $tries = 100;
+
+        // Benchmark bucket sort
+        $totalBucketSort = 0;
+
+        for ($i = 0; $i < $tries; ++$i) {
+            $start = microtime(true);
```
theofidry: missing import for `microtime`

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -96,4 +100,113 @@ public function test_it_returns_unique_and_sorted_by_time_test_cases(): void
...
+
+        return $uniqueTestLocations;
+    }
+
+    /**
+     * We assume locations should be ordered within an order of magnitude.
+     *
+     * @return bool
+     */
+    private static function orderConstraintsValid(array $sortedTestLocations)
```
theofidry:
```suggestion
    /**
     * @param TestLocation[] $sortedTestLocations
     */
    private static function isOrderConstraintsValid(array $sortedTestLocations): bool
```
?

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -96,4 +100,113 @@ public function test_it_returns_unique_and_sorted_by_time_test_cases(): void
...
+
+        $sortedTestLocations = iterator_to_array(JUnitTestCaseSorter::sort($uniqueTestLocations));
+        $this->assertTrue(self::orderConstraintsValid($sortedTestLocations), 'Bucket sort failed order check');
+
+        // Another sanity check
+        $sortedTestLocations = self::quicksort($uniqueTestLocations);
+        $this->assertTrue(self::orderConstraintsValid($uniqueTestLocations), 'Quicksort failed order check');
+    }
+
+    public function test_it_sorts_faster_than_quicksort(): void
```
theofidry: So it basically makes sure this is indeed a performance improvement?

theofidry:
I would also love us to add a benchmark to profile the mutations -> mutant process, but I don't think it should hold this PR either

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
sanmai: I'll check if it so starting PHP 7.4. [In 7.3 it is like this.](https://gist.github.com/sanmai/d866617fed5e4aa1e3eee15c78abf39a)

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
sanmai: Path for TestFileTimeData is not optional. It is never a null.

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
sanmai: It'll break the order  :man_facepalming:

sanmai:
I'd love to test it with the new benchmarks, but this isn't particularly easy because we test only so much mutations, not the whole range. I'll see to it anyway a bit later.

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
sanmai: Yes, that's right.

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
theofidry: How do you mean?

sanmai: Wish I knew! Try it: check it out, change, run the test.

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
```diff
@@ -36,53 +36,134 @@
...
+    }
+
+    /**
+     * Sorts tests to run the fastest first. Exposed for benchmarking purposes.
+     *
+     * @param TestLocation[] $uniqueTestLocations
+     *
+     * @return iterable<TestLocation>
+     */
+    public static function bucketSort(array $uniqueTestLocations): iterable
```
theofidry: If it needs to be exposed, we might as well extract it in a dedicated class (which could have a private constructor for)?

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
theofidry: Actually the answer is silly :) With `yield from` you also yield the keys, but hey you know what's the default value of `iterator_to_array` second argument? `true`: https://3v4l.org/XWc5G

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
theofidry: Fair enough, it's a hotpath anyway so I don't mind in this case

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -53,9 +61,8 @@ public function test_it_returns_first_file_name_if_there_is_only_one(): void
...
         $sorter = new JUnitTestCaseSorter();
 
-        $uniqueSortedFileNames = iterator_to_array(
-            $sorter->getUniqueSortedFileNames($coverageTestCases)
-        );
+        /** @var array $uniqueSortedFileNames */
+        $uniqueSortedFileNames = $sorter->getUniqueSortedFileNames($coverageTestCases);
 
         $this->assertCount(1, $uniqueSortedFileNames);
         $this->assertSame('/path/to/test-file-1', $uniqueSortedFileNames[0]);
```
theofidry: Might be better to check the whole array here? (`assertSame()`) - that also removes the need of adding `/** @var array $uniqueSortedFileNames */` above

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -96,4 +103,133 @@ public function test_it_returns_unique_and_sorted_by_time_test_cases(): void
         $this->assertCount(3, $uniqueSortedFileNames);
         $this->assertSame('/path/to/test-file-3', $uniqueSortedFileNames[0]);
```
theofidry: Same here: maybe worth checking against the whole array, or if the values are annoying we can do:

```php
$this->assertSame(
    [
        '/path/to/test-file-1',
        '/path/to/test-file-2',
        '/path/to/test-file-3',
    ],
    array_keys($uniqueSortedFileName)
);
```

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -96,4 +103,133 @@ public function test_it_returns_unique_and_sorted_by_time_test_cases(): void
...
+        $this->assertLessThan(
+            // Quicksort's average O(n log n)
+            JUnitTestCaseSorter::USE_BUCKET_SORT_AFTER * log(JUnitTestCaseSorter::USE_BUCKET_SORT_AFTER),
+            // Bucket Sort's average O(n + k)
+            JUnitTestCaseSorter::USE_BUCKET_SORT_AFTER + JUnitTestCaseSorter::BUCKETS_COUNT
+        );
+    }
+
+    /**
+     * @dataProvider locationsArrayProvider
```
theofidry:
```suggestion
     * @dataProvider locationsArrayProvider
     *
     * @param TestLocation[] $uniqueTestLocations
```

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -96,4 +103,133 @@ public function test_it_returns_unique_and_sorted_by_time_test_cases(): void
...
+     */
+    public function test_it_sorts_correctly(array $uniqueTestLocations): void
+    {
+        $sortedTestLocations = iterator_to_array(JUnitTestCaseSorter::bucketSort($uniqueTestLocations));
+        $this->assertTrue(self::isOrderConstraintsValid($sortedTestLocations), 'Bucket sort failed order check');
```
theofidry:
```suggestion
        $sortedTestLocations = iterator_to_array(
            JUnitTestCaseSorter::bucketSort($uniqueTestLocations),
            false
        );
        
        $this->assertTrue(
            self::isOrderConstraintsValid($sortedTestLocations),
            'Bucket sort failed order check'
        );
```

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -96,4 +103,133 @@ public function test_it_returns_unique_and_sorted_by_time_test_cases(): void
...
+    /**
+     * Sanity check
+     *
+     * @dataProvider locationsArrayProvider
+     */
+    public function test_quicksort_sorts_correctly(array $uniqueTestLocations): void
+    {
+        self::quicksort($uniqueTestLocations);
+
+        $this->assertTrue(self::isOrderConstraintsValid($uniqueTestLocations), 'Quicksort failed order check');
```
theofidry:
```suggestion
     * @dataProvider locationsArrayProvider
     *
     * @param TestLocation[] $uniqueTestLocations
     */
    public function test_quicksort_sorts_correctly(array $uniqueTestLocations): void
    {
        self::quicksort($uniqueTestLocations);

        $this->assertTrue(
            self::isOrderConstraintsValid($uniqueTestLocations), 
            'Quicksort failed order check'
        );
```

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -96,4 +103,133 @@ public function test_it_returns_unique_and_sorted_by_time_test_cases(): void
...
+    }
+
+    /**
+     * @dataProvider locationsArrayProvider
+     */
+    public function test_it_sorts_faster_than_quicksort(array $uniqueTestLocations): void
```
theofidry:
```suggestion
    /**
     * @dataProvider locationsArrayProvider
     *
     * @param TestLocation[] $uniqueTestLocations
     */
    public function test_it_sorts_faster_than_quicksort(array $uniqueTestLocations): void
```

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -96,4 +103,133 @@ public function test_it_returns_unique_and_sorted_by_time_test_cases(): void
...
+            self::quicksort($locationsCopy);
+            $totalQuickSort += microtime(true) - $start;
+        }
+
+        $this->assertGreaterThanOrEqual(0.01, abs($totalQuickSort - $totalBucketSort));
+    }
+
+    public static function locationsArrayProvider(): iterable
+    {
+        $locations = array_map(
```
theofidry: missing import for `array_map()`

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -96,4 +103,133 @@ public function test_it_returns_unique_and_sorted_by_time_test_cases(): void
...
+            static function (float $executionTime): TestLocation {
+                return new TestLocation('', '', $executionTime);
+            },
+            JUnitTimes::JUNIT_TIMES
+        );
+
+        yield [array_slice($locations, 0, JUnitTestCaseSorter::USE_BUCKET_SORT_AFTER * 10)];
```
theofidry: missing import for `array_slice`

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -96,4 +103,133 @@ public function test_it_returns_unique_and_sorted_by_time_test_cases(): void
...
+            // Updates by reference
+            $locationsCopy = $uniqueTestLocations;
+            self::quicksort($locationsCopy);
+            $totalQuickSort += microtime(true) - $start;
+        }
+
+        $this->assertGreaterThanOrEqual(0.01, abs($totalQuickSort - $totalBucketSort));
+    }
+
+    public static function locationsArrayProvider(): iterable
```
theofidry:
```suggestion
    /**
     * @return iterable<TestLocation>
     */
    public static function locationsArrayProvider(): iterable
```

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -96,4 +103,133 @@ public function test_it_returns_unique_and_sorted_by_time_test_cases(): void
...
+            },
+            JUnitTimes::JUNIT_TIMES
+        );
+
+        yield [array_slice($locations, 0, JUnitTestCaseSorter::USE_BUCKET_SORT_AFTER * 10)];
+
+        yield [$locations];
+    }
+
+    private static function quicksort(&$uniqueTestLocations): void
```
theofidry:
```suggestion
    /**
     * @param TestLocation[] $uniqueTestLocations
     */
    private static function quicksort(array &$uniqueTestLocations): void
```

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -96,4 +103,133 @@ public function test_it_returns_unique_and_sorted_by_time_test_cases(): void
...
+    }
+
+    /**
+     * We assume locations should be ordered within an order of magnitude.
+     *
+     * @param TestLocation[] $sortedTestLocations
+     *
+     * @return bool
+     */
+    private static function isOrderConstraintsValid(array $sortedTestLocations)
```
theofidry:
```suggestion
     */
    private static function isOrderConstraintsValid(array $sortedTestLocations): bool
```

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
sanmai: Well, that's silly for sure. Fixed this.

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
sanmai: That's iterable of array of array of locations. Not sure if it makes sense to document this since we're not using this method directly.

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
sanmai: Fair. How's TestLocationBucketSorter working for a name?

sanmai:
1. 40% of Psalm's mutations have more than 15 tests. [This is the graph.](https://user-images.githubusercontent.com/139488/77132338-6762dd00-6aa2-11ea-9a1a-6bc1a29ed4d6.png)
2. This graph is for Big O for quicksort and bucket sort.
3. I cannot make a straight on benchmark on Psalm because Blackfire requires some impossible amounts of RAM I don't have. Our standard benchmarks show no particular improvement, but neither a regression. Might be because of insufficient depth.

Third point is the major point for me. Even if local benchmarks show improvement, global improvement might be below the noise threshold. I'd like to see them, and to show them, but it appears to be hard for one reason or another.

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
theofidry: no you're right, we don't for providers...

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
theofidry: 👌

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -37,6 +37,8 @@
 
 use Infection\AbstractTestFramework\Coverage\TestLocation;
 use Infection\TestFramework\Coverage\JUnit\JUnitTestCaseSorter;
+use function iterator_to_array;
+use function log;
```
theofidry: unnecessary here no?

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
```diff
@@ -81,19 +80,33 @@ public function test_it_returns_unique_and_sorted_by_time_test_cases(): void
...
         );
 
-        $this->assertCount(3, $uniqueSortedFileNames);
-        $this->assertSame('/path/to/test-file-3', $uniqueSortedFileNames[0]);
+        $this->assertSame([
+            '/path/to/test-file-3',
+            '/path/to/test-file-4',
+            '/path/to/test-file-1',
+            '/path/to/test-file-2',
+        ], $uniqueSortedFileNames);
```
theofidry:
```suggestion
        $this->assertSame(
            [
                '/path/to/test-file-3',
                '/path/to/test-file-4',
                '/path/to/test-file-1',
                '/path/to/test-file-2',
            ], 
            $uniqueSortedFileNames
        );
```

tests/phpunit/TestFramework/Coverage/JUnit/TestLocationBucketSorterTest.php
```diff
@@ -0,0 +1,196 @@
...
+     *
+     * @param TestLocation[] $uniqueTestLocations
+     */
+    public function test_it_sorts_faster_than_quicksort(array $uniqueTestLocations): void
+    {
+        if (extension_loaded('xdebug') || PHP_SAPI === 'phpdbg') {
+            $this->markTestSkipped('Benchmarks under xdebug or phpdbg are brittle');
+        }
+
+        if (self::isOrderConstraintsValid($uniqueTestLocations) === true) {
```
theofidry: unnecessary `=== true` here

tests/phpunit/TestFramework/Coverage/JUnit/TestLocationBucketSorterTest.php
```diff
@@ -0,0 +1,196 @@
...
+            }
+        );
+    }
+
+    /**
+     * We assume locations should be ordered within an order of magnitude.
+     *
+     * @param TestLocation[] $sortedTestLocations
+     */
+    private static function isOrderConstraintsValid(array $sortedTestLocations): bool
```
theofidry: `areConstraintsOrderValid()`?

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestCaseSorterTest.php
sanmai: They're used to confirm the math used to select bucket sort, which is done in this class.

tests/phpunit/TestFramework/Coverage/JUnit/TestLocationBucketSorterTest.php
sanmai: OK, removed

tests/phpunit/TestFramework/Coverage/JUnit/TestLocationBucketSorterTest.php
sanmai: Fair

Issue #1176, opened 2020-03-17
[RFC] Performance monitoring


theofidry:
Given #1173, the more frequent recent works on performances, I think it is time to step up our game. Just for a bit more context, here is the current situation:

- we profile occasionally in our corners each one on different projects
- profiling is not necessarily easy as we have a lot of "noisy" or unreliable parts of the applications (in terms of profiling), e.g.:
  - a lot of I/O operations: read files, dump files
  - process handling: for the tests, for the mutation analysis
  - parsing of PHP files into AST and traversing: it's often a dominant part of the profile yet not something we can do much about

What I would like to achieve is an easier experience to profile, provide more reliable and robust results.

In order to achieve that, I think there is several actionable:

- Create dedicated small cases, e.g.:
  - the initial test run process creation (no execution): this would allow to profile the command creation
  - file parsing and traverse to get mutations (this is mostly PhpParser core and a bit of our stuff)
  - mutation -> mutant process creation which includes the coverage parsing + creation of the framework config file
  - the report creation
- Set up some benchmarks: with the few cases above it would be easier to have a real benchmark, not profile, e.g. via https://github.com/phpbench/phpbench
- A dry-run mode: i.e. a mode where we skip the process handling and reporting (which is a huge part of instability due to the processes and usleep calls). This would be an ideal complement with the `--coverage` option
- A utility to ease profiling: there is often a few recurring points to do for a profile, e.g.:
  - make sure xdebug is disabled (huge overhead)
  - make sure pcov is disabled (skews blackfire giving weird profiles)
  - make sure the autoloader is dumped in authoritative mode
  - make sure the dependencies are up to date (easy to forget if you check out two different versions)
  - generate the coverage, make sure that the infection command includes `--quiet` for example
- Request an OSS license for Blackfire? (not sure it's doable, but hey we can ask)

I don't expect all the above to solve 100% of our problems or to find big performance improvements from any of them. Also some might be too hard to implement in the current form. For example, wouldn't it be amazing that at each PR for all the scenarios above are compared with both versions to get the difference?

In other words, my goal here is to list some actionnables in order to climb the ladder. I do not think we need to try to reach the top of it right away. For example having just a few small scenarios which are more "profile friendly" or a dry-run mode would already be a huge step forward IMO.

theofidry:
Note from @sanmai, includes #1150 as well

theofidry:
[...]

PR #1178, opened 2020-03-17
Add tracing profiling

theofidry:
Related to #1176 

How does it works? You can add profiling "scenarios" in `tests/benchmark`. The code there is autoloaded as usual and use the regular autoloader. This PR introduces two scenarios:

- `MutationGenerator`:
  - `generate-mutations-closure.php` contains the main code: configure the container, get the fixtures and returns the closure for which we want to profile its execution (of the closure) which is parsing the source files to generate the mutations
  - `generate-mutations.php` simply call the closure above, useful locally for debugging
  - `profile.php`: the code on which to execute the `blackfire run` command
- `Tracing`: profiles the (enriched) traces, i.e. parsing the XML coverage & JUnit report and append uncovered files to create traces. The structure is analogue to `MutationGenerator`

A few additional notes:

- I introduced `BlackfireInstrumentator` which facilitates its usages, what it does is:
  -  check that you have the appropriate environment to profile, i.e. have the blackfire extension enabled, Xdebug & pcov disabled and that the composer autoloader is dumped with classmap authoritative
  - enable the probe for the closure executed, which means anything that happens outside of this scope is completely ignored. This allows to profile only what we desire without a care of how "heavy" or "noisy" setting up the scenario is
- The coverage & source codes are compressed in order to be able to push them to GitHub (limit of 100MB)
- `make profile` just makes it easy to run all the profiles

There is obviously a lot more that could be done, but I believe this is a step in the right direction without bringing an atrocious level of complexity to solve our issues.

Here's the two profiles generated after running the command:

- https://blackfire.io/profiles/d8e5faa7-7a92-4e11-85f2-e7dd8213bad9/graph
- https://blackfire.io/profiles/14eee7a9-9a00-4532-9d0f-71321ef27b6f/graph

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ profile: vendor tests/benchmark/MutationGenerator/sources tests/benchmark/Tracin
 		--title="Tracing" \
 		--metadata="commit=$(COMMIT_HASH)" \
 		php tests/benchmark/Tracing/profile.php
-	composer dump --classmap-authoritative
+	composer dump
 
 
 .PHONY: autoreview
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ DOCKER_RUN_74=$(FLOCK) devTools/*php74*.json $(DOCKER_RUN) infection_php74
 DOCKER_RUN_74_IMAGE=devTools/Dockerfile-php74-xdebug.json
 
 FLOCK=./devTools/flock
+COMMIT_HASH=$(shell git rev-parse --short HEAD)
 
 
 #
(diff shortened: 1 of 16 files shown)


Makefile
```diff
@@ -65,6 +66,23 @@ phpstan: vendor $(PHPSTAN)
...
+		--samples=5 \
+		--title="MutationGenerator" \
+		--metadata="commit=$(COMMIT_HASH)" \
+		php tests/benchmark/MutationGenerator/profile.php
+	blackfire run \
+		--samples=5 \
+		--title="Tracing" \
+		--metadata="commit=$(COMMIT_HASH)" \
+		php tests/benchmark/Tracing/profile.php
+	composer dump --classmap-authoritative
```
maks-rafalko: why do we need to dump at the end again?

theofidry: completely wrong, I meant to revert it

theofidry:
>How is it supposed to be used? Manually running locally?

For now yes, locally. It won't help past commits, but my hope is that it stays stable so that it stays relevant for the new ones.

>in which cases do we need to provide profiles?

Probably as soon as we do a PR performance related or on a hotpath where performances are an issue.

>does the author of the PR need to run it on master and on a new branch to compare?

Yes

So as you can see from those answers: it's not a magic solution at all. The idea though is that we have a few scoped scenarios (i.e. less noisy & more consistent & easier to run). It's only step though.

A typical example is for the last few days sanmai and I tried on our ways to profile a specific path (actually the "Tracing" scenario here) with very different results and sometimes really unhelpful profiles.

You can also see from the `BlackfireInstrumentator::check()` that there is a couple of annoying steps that are best being detected, rather than waiting 5min or more for a profile to complete to find out and start over

sanmai:
Wanted to try this, but where from sources go?

Issue #1150, opened 2020-03-11
Add option to dry-run Infection


sanmai:
When testing Infection on a big project you don't want to actually MT that project. Because it'll take a lot of time, and because you are testing Infection, but not that project.

So far I had to comment out `start()` call inside `ParallelProcessRunner`, and it works. But this interferes with testing. I have to un-comment it back to run PHPUnit, and comment it again to benchmark Infection. 

Another problem here is that the report gets filled with `E`'s. Therefore, some parts of our processing pipeline become ignored and unused. Say, we don't necessarily collect errored mutations for future logging. What if we are hitting a bottleneck if we do?

What I want is to be able to run Infection with `--dry-run[=N]`, where N is an optional random seed, not unlike that of PHPUnit. Then Infection would use a stand-in dummy process that will fail/err/escape pseudo-randomly according to the seed.

theofidry:
I've been thinking the same since we started to profile more regularly: executing the processes seriously skews the results. Generally the more I/O and network we have the less the profiles are reliable.

So I'm totally up for a `--dry-run` option. Maybe the solution is to replace the parallel runner which will consume the provided generator (to trigger its content) but won't fire a process. It may just not dispatch any `MutantProcessWasFinished` and we could skip the whole logging with a `NullLogger` for example.

I think we if are more interesting in the downstream part, e.g. the loggers, we are better off with finding a way to benchmark it without all the initial run

sanmai:
As of writing we're collecting 100% of execution results, even if we can tell for certain we won't need some of them, or all of them, because of disabled logging. This is a definite area of improvement, but with the current state of `--dry-run` it is impossible to see any benefits from changes in the area.

theofidry:
@sanmai I do not understand what is missing here. Could you detail it further?

sanmai:
PR #1183 implemented `--dry-run` by skipping the actual process execution, but the code still:  

- Collects all of execution results                                                                                                                                                                                                                                                            
- Processes all of them through the result pipeline                                                                                                                                                                                                                                             
- Does all kinds of downstream work (event, metrics, etc.)    

Ideally we'd want to be able to profile mutation generation in isolation, without the noise of result collection. And possibly result collection without mutation generation.

```php
enum DryRunMode {                                                                 
    case Off;                    // Normal execution                              
    case SkipExecution;          // Current behavior: skip processes, still collec
    case SkipCollection;         // Skip processes AND skip event dispatch        
}                                                                                 
```

[...]

PR #1179, opened 2020-03-18
Move threadCount to Configuration

theofidry:
As I was looking into #1150, I was blocked by this question: where should this parameter go? Should it stored in the `Configuration` be part of service constructor(s) or passed around as a parameter?

If we take a closer look at Infection, the reality is as follows:

- we have a lot of parameters (see `Configuration`)
- most services depends on those parameters (i.e. depends on `Configuration`)

And it totally makes sense. As much as `Container::withDynamicParameters()` might be a bit confusing at first, it's neither an out of ordinary pattern and neither a bad fit: the alternative means passing the config to 3/4 of our services and that does not make any sense.

So back to the initial question. I think the best way to approach this is to ask ourselves: is that parameter an implementation detail of this class, or a generic parameter? For better clarity let's take our example here from this PR:

```php
// Prior to this PR
class ParallelProcessRunner {
    function run($processes, $threadCount)
}
```

`ParallelProcessRunner` is a concrete class without any parents. But if we imagined it had an interface, e.g. `ProcessRunner`, do the parameters `$processes` and `$threadCount` have a place there or are they an implementation detail of `ParallelProcessRunner`? IMO `$process` has a place yes, but not `$threadCount`.

Another example: `MutationTestingResultsLogger`:

```php
interface MutationTestingResultsLogger
{
    /**
     * Logs results of Mutation Testing to somewhere
     */
    public function log(): void;
}
```

Would it make sense to have a config object passed to `log()` for each type of logger? IMO while doable, it does not really makes sense either here (it could in other contexts maybe?).

I might revisit some similar cases in the future, but for now I'm fixing this one and I will add the `dryRun` parameter to `Configuration` in a similar fashion

### DIFF
--- a/tests/phpunit/Fixtures/Process/DummyProcessBearer.php
+++ b/tests/phpunit/Fixtures/Process/DummyProcessBearer.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Infection\Tests\Fixtures\Process;
+
+use Infection\Process\Runner\Parallel\ProcessBearer;
+use PHPUnit\Framework\Assert;
+use Symfony\Component\Process\Process;
+use function Safe\sprintf;
+
+final class DummyProcessBearer implements ProcessBearer
+{
+    private $process;
+    private $expectTimeOut;
+
+    public function __construct(Process $process, bool $expectTimeOut)
+    {
+        $this->process = $process;
+        $this->expectTimeOut = $expectTimeOut;
+    }
+
+    public function getProcess(): Process
+    {
+        return $this->process;
+    }
+
+    public function markAsTimedOut(): void
+    {
+        if (!$this->expectTimeOut) {
+            Assert::fail(sprintf(
+                'Did not expect "%s()" to be called',
+                __FUNCTION__
+            ));
+        }
+    }
+}
--- a/tests/phpunit/Process/Runner/Parallel/ParallelProcessRunnerTest.php
+++ b/tests/phpunit/Process/Runner/Parallel/ParallelProcessRunnerTest.php
@@ -36,11 +36,14 @@ declare(strict_types=1);
 namespace Infection\Tests\Process\Runner\Parallel;
 
 use Closure;
+use Generator;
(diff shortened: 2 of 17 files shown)


tests/phpunit/Process/Runner/Parallel/ParallelProcessRunnerTest.php
```diff
@@ -36,148 +36,175 @@
...
-        $runner = new ParallelProcessRunner(self::makeEventAccounter($eventDispatcher));
-        $runner->run($processes, 4, 0);
+        $runner->run($processes);
     }
 
-    public function test_it_checks_for_timeout(): void
+    public function test_it_checks_if_the_executed_processes_time_out(): void
     {
-        $processes = (function (): iterable {
+        $processes = (function (): Generator {
```
sanmai: iterable?

tests/phpunit/Fixtures/Process/DummyProcessBearer.php
```diff
@@ -0,0 +1,37 @@
...
+    public function markAsTimedOut(): void
+    {
+        if (!$this->expectTimeOut) {
+            Assert::fail(sprintf(
+                'Did not expect "%s()" to be called',
+                __FUNCTION__
+            ));
+        }
+    }
+}
```
sanmai: Easier than a mock?

theofidry: I think so. In general I don't lick mocks that much, if they are appropriate why not, but if you can easily do without I prefer without. It has several advantages:

- Rots less easily
- Easier to refactor
- Less magic (so more accessible)
- Faster (mocking is slow)

I also got the impression that [jakzal](https://github.com/jakzal), [frankdejonge](https://github.com/frankdejonge) & [sebastian](https://github.com/sebastianbergmann) recommend the same. I would not say as far as blindly do without mocks at all, but I'm considering hunting down some cases where I feel we should not need mock

sanmai: I'm not a big fan of mocks tbh, so it's just surprising to see a different approach for once.

PR #1183, opened 2020-03-19
Implements dry-run mode

### LINKED ISSUE #1150: Add option to dry-run Infection

theofidry:
Closes #1150

### DIFF
--- a/src/Command/InfectionCommand.php
+++ b/src/Command/InfectionCommand.php
@@ -191,6 +191,12 @@ final class InfectionCommand extends BaseCommand
                 InputOption::VALUE_NONE,
                 'Debug mode. Will not clean up Infection temporary folder.'
             )
+            ->addOption(
+                'dry-run',
+                null,
+                InputOption::VALUE_NONE,
+                'Dry run. Will not clean up Infection temporary folder.'
+            )
         ;
     }
 
@@ -307,7 +313,8 @@ final class InfectionCommand extends BaseCommand
             $testFramework === '' ? null : $testFramework,
             $testFrameworkExtraOptions === '' ? null : $testFrameworkExtraOptions,
             trim((string) $input->getOption('filter')),
-            (int) $this->input->getOption('threads')
+            (int) $this->input->getOption('threads'),
+            (bool) $this->input->getOption('dry-run')
         );
     }
 
--- a/devTools/phpstan-src.neon
+++ b/devTools/phpstan-src.neon
@@ -36,6 +36,9 @@ parameters:
             path: '../src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php'
             message: '#array_key_exists expects int\|string, string\|null given#'
             count: 1
+        -
+            path: '../src/FileSystem/DummyFileSystem.php'
+            message: '#Infection\\FileSystem\\DummyFileSystem#'
     level: max
     paths:
         - ../src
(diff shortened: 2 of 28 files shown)


src/Command/InfectionCommand.php
```diff
@@ -191,6 +191,12 @@ protected function configure(): void
                 InputOption::VALUE_NONE,
                 'Debug mode. Will not clean up Infection temporary folder.'
             )
+            ->addOption(
+                'dry-run',
+                null,
+                InputOption::VALUE_NONE,
+                'Dry run. Will not clean up Infection temporary folder.'
```
sanmai: Why keep it?

sanmai (approved):

src/Command/InfectionCommand.php
theofidry: Wrong copy/paste

maks-rafalko:
Wanted to try it locally

```
gh pr checkout 1183
composer install

bin/infection --dry-run 
```

Initial tests fails

```
vendor/bin/phpunit
```

Test fails

```
  1) Infection\Tests\BenchmarkTest::test_all_the_benchmarks_can_be_executed with data set "Tracing" 


Error Output:
  ================
  PHP Warning:  file_get_contents(infection/tests/benchmark/Tracing/coverage/coverage-xml/index.xml): failed to open stream: No such file or directory in infection/vendor/thecodingmachine/safe/generated/filesystem.php on line 250
  PHP Fatal error:  Uncaught Safe\Exceptions\FilesystemException: file_get_contents(infection/tests/benchmark/Tracing/coverage/coverage-xml/index.xml): failed to open stream: No such file or directory in infection/vendor/thecodingmachine/safe/generated/Exce
  ptions/FilesystemException.php:9
  Stack trace:
```

After commenting out that tests - everything works as expected

Who is supposed to use this feature? only developers of Infection? I mean, do we need to document it?

theofidry:
hm will check with the test run, I forgot about it and just ran it with skipping the initial tests.

>Who is supposed to use this feature? only developers of Infection? I mean, do we need to document it?

Mostly us for profiling, but I guess it can still be useful to get and idea of the number of mutations you get with your config

theofidry:
Hm test pass. I think you need to do a `make profile` to untar the appropriate files for `BenchmarkTest`

sanmai:
> Hm test pass. I think you need to do a `make profile` to untar the appropriate files for `BenchmarkTest`

maks-rafalko:
> Hm test pass. I think you need to do a `make profile` to untar the appropriate files for `BenchmarkTest`

that's the point, there shouldn't be any dependency. Anyway, @sanmai fixed it :)

maks-rafalko (approved):

src/Command/InfectionCommand.php
theofidry: Fixed in #2436.

PR #1186, opened 2020-03-20
Tweak Makefile

theofidry:
@maks-rafalko #1183 just makes it a bit more obvious to me: I don't think any of us is really using the `make test` command, at least not very often.

IMO the docker-based tests are nice to have since it really helps out in some situations. But most of the time you are happy to just run the basic tests and leave the rest to the CI. For that reason I would like to:

- rename the test targets to `*-docker` when they use Docker
- introduce local test targets

In other words, you can have `make test` and `make test-docker`. The advantages of using a Makefile here is like before: it takes care of some sanity check you might otherwise forget (installing/updating deps, untarring some files...)

Also removed 7.2 which we no longer supports :)


theofidry:
Also note we are missing pcov for completeness...

PR #1187, opened 2020-03-20
Tweak JUnitTestCaseSorter

sanmai:
This PR:

- [x] Removes null checks as the path for TestFileTimeData is not optional
- [x] Uses semantically incorrect [but faster isset()](https://gist.github.com/sanmai/d866617fed5e4aa1e3eee15c78abf39a)


src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
```diff
@@ -59,11 +59,11 @@ public function getUniqueSortedFileNames(array $tests): iterable
             /** @var TestLocation $testLocation */
             $testLocation = current($uniqueTestLocations);
 
+            /** @var string $filePath */
             $filePath = $testLocation->getFilePath();
 
-            if ($filePath !== null) {
-                yield $filePath;
-            }
+            // Path for TestFileTimeData is not optional. It is never a null.
```
theofidry: I would slightly tweak the comment to reflect that then otherwise the comment is confusing as it contradicts the type

theofidry: Also might be more appropriate to put L62 with the type

sanmai: Can you see if it is better now?

src/TestFramework/Coverage/JUnit/JUnitTestCaseSorter.php
```diff
@@ -59,11 +59,16 @@ public function getUniqueSortedFileNames(array $tests): iterable
...
+             * Therefore we don't need to make any type checks here.
+             */
+
+            /** @var string $filePath */
             $filePath = $testLocation->getFilePath();
 
-            if ($filePath !== null) {
-                yield $filePath;
-            }
+            yield $filePath;
```
theofidry: LGTM, can you also update the return type?  `@return iterable<string>`

sanmai: Done

PR #1189, opened 2020-03-20
Tweak the MinMsiChecker to throw instead of returning a boolean

theofidry:
Depends on #1188 

- Rename `TestRunConstraintChecker` to `MinMsiChecker`
- Move metrics related classes under the `Metrics` namespace together
- Change the pattern of usage of the `MinMsiChecker`: instead of returning a boolean with failuretype, and let the `Engine` log accordingly and return a boolean, we do nothing & return nothing when everything is fine and throw on failure.

Why this change? There is several reasons:

- Reduce dependency in shared mutable state (`MetricsCalculator`, `TestRunConstraintChecker#failureType`
- Simplify the code: rather than using a boolean return value which requires additional mutable state to provide context, it is easier to let the callee instead of the caller do the logging appropriately

Note that although I plan to change more, I sticked to the above which is already quite a big change IMO. So I did not change the error messages or when they should be displayed for example.

One more thing: I now throw `MinMsiCheckFailed` exception instead of logging with an `error` block, so the result is now like this:

Screen Shot 2020-03-20 at 14 47 04


src/Console/ConsoleOutput.php
```diff
@@ -77,40 +76,40 @@ public function logBadMsiErrorMessage(MetricsCalculator $metricsCalculator, floa
...
-        $msiDifference = $msi - $minMsi;
+    public function logMinCoveredCodeMsiCanGetIncreasedNotice(float $minMsi, float $coveredCodeMsi): void
+    {
+        $typeString = 'Covered Code MSI';
 
         $this->io->note(
             sprintf(
                 self::MIN_MSI_CAN_GET_INCREASED_NOTICE,
                 $typeString,
-                $msiDifference,
```
maks-rafalko: IMO, removing this variable makes the code less understandable

theofidry: Fair, i'll introduce it back, but after since I'm going to get rid of most of those in the coming PR after rebase :)

PR #1190, opened 2020-03-20
Round MSI scores

theofidry:
- Better min-score and min-covered-score validation
- Round min-score & min-covered-score values
- Round calculated MSI & covered code MSI


sanmai:
IIRC the original proposal explicitly left the whole precision intact arguing that for sufficiently large project leaving only two digits past the decimal point might not be enough.

theofidry:
I'm fine with increasing it by one more digit for example, but having 10 of them is ridiculous :)

sanmai:
How do you think if we'd detect the precision as provided by a user, then display them as such?

maks-rafalko:
> IIRC the original proposal explicitly left the whole precision intact arguing that for sufficiently large project leaving only two digits past the decimal point might not be enough.

exactly. Not sure why we reverting it here. What's the point? You see this only on CI, and only when it fails.

theofidry:
>exactly. Not sure why we reverting it here. What's the point? You see this only on CI, and only when it fails.

3 reasons:

- it looks absolutely ridiculous:

image

- you don't want a failure because of a 0.0000001% difference: I'm already getting pissed at Codecov when it gives me a -0.01% code coverage...

>How do you think if we'd detect the precision as provided by a user, then display them as such?

I thought about it, but there is the risk that people just use `93` instead of `93.00`.

What about having a more sensible rounding, like 3 digits? I doubt it makes sense to have more. If you have such a large codebase, you probably are not running it on the whole codebase anyway

sanmai:
> there is the risk that people just use 93 instead of 93.00

Cool. Then we display it with no digits past decimal point, right?

theofidry:
#694 describes a different problem though: rounding without any digit past the decimal point at all whereas this PR is rounding with 2 digits past the decimal point

How about we try to get the precision from the user but default to 2?

sanmai:
> How about we try to get the precision from the user but default to 2?

As in `max(2, $userPrecision)`? That'd be perfect.

theofidry:
Updated :)

src/Metrics/MetricsCalculator.php
```diff
@@ -98,6 +102,10 @@ public function __construct()
 
     public function collect(MutantExecutionResult ...$executionResults): void
     {
+        if (count($executionResults) > 0) {
```
maks-rafalko: what do you think about adding some comment here? not clear why do we need it

tests/phpunit/Metrics/CalculatorTest.php
```diff
@@ -139,16 +140,16 @@ public function metricsProvider(): iterable
             2,
             2,
             0,
-            84.61538461538461,
+            84.62,
```
maks-rafalko: we can leave on of the data set with previous precision (84.61538461538461) to test it still works if user wants it

PR #1193, opened 2020-03-21
Auto-discovery of the JUnit file

theofidry:
One thing that annoys me with the current `--coverage` option is its rigidity. It is cumbersome and unfriendly. Even worse, [our docs on it are wrong](https://infection.github.io/guide/command-line-options.html#coverage) referencing `phpunit.junit.xml` instead of `junit.xml`.

#1107 is an improvement, it helps to:

- fail early if there is a misconfiguration or the coverage does not look good
- provide helpful error messages in case of failure

Although it is an improvement, I think we can go a step further and try to locate automatically the JUnit file and the XML report index file in the provided directory.

Since performances can still be an issue (we would do more I/O lookups), I propose to keep the current expected structure as the default and fallback to a more detailed lookup in case of failure.

This PR is only about the JUnit file. More PRs will be done in order to reach the goal mentioned above.


theofidry:
Travis failures are due to one mutant escaping - which is fine, it's about caching the result so there is not much to do there iMO

src/TestFramework/Coverage/JUnit/JUnitReportLocator.php
```diff
@@ -0,0 +1,122 @@
...
+    {
+        if ($this->jUnitPath !== null) {
+            return $this->jUnitPath;
+        }
+
+        // This is the JUnit path enforced before. It is also the one recommended by the
+        // CoverageChecker hence it makes sense to try this one first before attempting any more
+        // expensive lookup
+        if (file_exists($this->defaultJUnitPath)) {
+            return $this->jUnitPath ?? $this->defaultJUnitPath;
```
maks-rafalko: do we need to check `$this->jUnitPath` here? If it's not `null`, we will early return on line 76 I guess

src/TestFramework/Coverage/JUnit/JUnitReportLocator.php
```diff
@@ -0,0 +1,122 @@
...
+                        $files
+                    )
+                )
+            ));
+        }
+
+        $junitFileInfo = current($files);
+
+        if ($junitFileInfo !== false) {
+            return $this->jUnitPath ?? Path::canonicalize($junitFileInfo->getPathname());
```
maks-rafalko: Same question: do we need to check $this->jUnitPath here? If it's not `null`, we will early return on line 76 I guess

src/TestFramework/Coverage/JUnit/JUnitReportLocator.php
maks-rafalko: hm... `$this->jUnitPath` is never assigned, probably you meant to use `??=` here instead of `??`: https://3v4l.org/gpRX9

Then the test that checks memozing does not work correctly, I guess with current state of this PR nothing is cached

maks-rafalko:
> Even worse, [our docs on it are wrong](https://infection.github.io/guide/command-line-options.html#coverage) referencing `phpunit.junit.xml` instead of `junit.xml`.

This page does not contain `phpunit.junit.xml`. Probably something is cached on your browser?

src/TestFramework/Coverage/JUnit/JUnitReportLocator.php
theofidry: >Then the test that checks memozing does not work correctly, I guess with current state of this PR nothing is cached

Might not be limited to here... Will double check that elsewhere as well in another PR

theofidry: Nvm just need to be `=` here

src/TestFramework/Coverage/JUnit/JUnitReportLocator.php
theofidry: Yes simple `=` is more appropriate here!

theofidry:
>Probably something is cached on your browser?

Indeed :(

PR #1194, opened 2020-03-22
Auto-discovery of the index XML coverage report file

theofidry:
Follow up of #1193: auto-discover the index XML file.

With #1193 and this PR, I will be able in the next PR to slightly simplify the way we deal with the passed coverage file path allowing to break away from the currently expected file structure.


tests/phpunit/TestFramework/Coverage/XmlReport/IndexXmlCoverageLocatorTest.php
```diff
@@ -0,0 +1,166 @@
...
+        $expected = normalizePath(realpath($this->tmp . DIRECTORY_SEPARATOR . $indexRelativePath));
+
+        $this->assertSame($expected, $this->locator->locate());
+        // Call second time to check the cached result
+        $this->assertSame($expected, $this->locator->locate());
+    }
+
+    public function test_it_cannot_locate_the_index_file_if_the_result_is_ambiguous(): void
+    {
+        touch('index.xml');
```
maks-rafalko: didn't get this case. Shouldn't this loginc from locator immediately return `index.xml` and do not search in nested folders:

```php
if (file_exists($this->defaultIndexPath)) {
    return $this->indexPath = $this->defaultIndexPath;
}
```

theofidry: the default path is `coverage/coverage-xml/index.xml`, here it is respectively `coverage/index.xml` and `coverage/sub-dir/index.xml`

PR #1197, opened 2020-03-23
Move the process termination callback from the process runner to the process bearer

theofidry:
While I believe it makes sense that a process runner is aware of a closure to call at the end of the process run, I do not think the closure should come directly from the runner but rather the process itself: on theory that could allow each process to have a very different closure, rather than a generic one.

In other words, the following configuration offers a bit more flexibility. Granted however it is but a nitpick at this point


src/Process/Builder/MutantProcessBuilder.php
```diff
@@ -49,11 +52,20 @@ class MutantProcessBuilder
 {
     private $testFrameworkAdapter;
     private $timeout;
+    private $eventDispatcher;
+    private $resultFactory;
 
-    public function __construct(TestFrameworkAdapter $testFrameworkAdapter, int $timeout)
-    {
+    // TODO: is it necessary for the timeout to be an int?
```
sanmai: Not at all. Process supports floats.

src/Process/Builder/MutantProcessBuilder.php
```diff
@@ -75,6 +87,21 @@ public function createProcessForMutant(Mutant $mutant, string $testFrameworkExtr
...
+
+        $mutantProcess->registerTerminateProcessClosure(static function () use (
+            $mutantProcess,
+            $eventDispatcher,
+            $resultFactory
+        ): void {
+            $eventDispatcher->dispatch(new MutantProcessWasFinished(
+                $resultFactory->createFromProcess($mutantProcess))
+            );
+        });
```
sanmai: Do you think MutantProcess can't be aware of the events? E.g. it can do what it should without closure DI.

I'm not sure this is the right way, but adding a callback with an event here, which events are used to avoid these callbacks, looks a tiny bit messy.

sanmai: I mean, would it be simpler to inject `eventDispatcher` and `resultFactory` into the `MutantProcess`, rather then pulling these tricks with closures and such?

src/Process/Builder/MutantProcessBuilder.php
theofidry: Will probably try to change it in another PR then :)

src/Process/Builder/MutantProcessBuilder.php
sanmai: Not to say we should go the way right this moment. For one because I come to thinking it'll be better to inject some sort of a result handler, which, in turn, would dispatch the events.

theofidry: My reasoning was that:

- The current dependency does not add up: the runner needs to be aware of the underlying `ProcessBearer` implementation - its callback registered only works with one
- It is common for a process to have a callback (in fact the Symfony one has one, it's just awkward to try to extend it)

But I would be lying to say I'm 100% confident on that one... I just think it's slightly more correct, without feeling that I nailed it either

PR #1198, opened 2020-03-23
Introduce UnsupportedMethod

theofidry:
Small exception to be a bit more expressive when throwing an exception in a method when the method is not supported, e.g. in our "fake implementations".


maks-rafalko:
Please fix the build https://travis-ci.org/github/infection/infection/jobs/665939304#L321

PR #1199, opened 2020-03-24
Log initial test failure instead of throwing

theofidry:


src/Command/RunCommand.php
```diff
@@ -243,7 +244,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
         try {
             $engine->execute();
-        } catch (MinMsiCheckFailed $exception) {
+        } catch (InitialTestsFailed | MinMsiCheckFailed $exception) {
             // TODO: we can move that in a dedicated logger later and handle those cases in the
             // Engine instead
             // TODO/ do the same for TestFails
```
maks-rafalko: this todo should be removed now I guess

PR #1141, opened 2020-03-08
Bundle test framework adapters into PHAR

maks-rafalko:
This PR:

- [x]  closes https://github.com/infection/infection/issues/1129
- [x] bundles adapters into PHAR
- [x] Doc PR: https://github.com/infection/site/pull/159

PR #1201, opened 2020-03-24
Increase verbosity for Travis CI builds

sanmai:
This PR:

- [ ] Fixes CI build for PHP 7.4 + xdebug, hopefully

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ sudo: false
 
 env:
   global:
-    - INFECTION_FLAGS='--threads=$(nproc) --min-msi=63.949716139497 --min-covered-msi=74 --coverage=coverage --log-verbosity=none'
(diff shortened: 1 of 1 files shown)


maks-rafalko:
I would say this is quite a dangerous change for us.

I'm thinking about the cases like in this PR https://github.com/infection/infection/pull/1141 we got exactly "initial tests failed" error while I mistakenly thought that the issue is in the other place

Is it only for 0.16? We will merge 0.16. then back to master after release

sanmai:
I rolled it back. In the end we only add `--skip-initial-tests` because apparently tests we causing Travis to timeout the build.

maks-rafalko:
> because apparently tests we causing Travis to timeout the build.

is it because of the process does not output anything for a long time?

sanmai:
It might be it, as we're running it with `--no-progress`. Yet it only failed for PHP 7.4 + xdebug for some unknown reason. I think `--skip-initial-tests` here offers a good opportunity to speed up our own builds.

PR #1202, opened 2020-03-24
Update PHPStan, fix warnings

sanmai:
This PR:

- [x] Updates PHPStan
- [x] Fixes two new warnings


src/Command/RunCommand.php
```diff
@@ -266,12 +266,14 @@ private function initContainer(InputInterface $input): void
         $testFrameworkExtraOptions = trim((string) $this->input->getOption('test-framework-options'));
         $initialTestsPhpOptions = trim((string) $input->getOption('initial-tests-php-options'));
 
+        /** @var ?string $minMsi */
```
theofidry: I think `string|null` is more correct

sanmai: Corrected

PR #1203, opened 2020-03-24
Restrict Docker build context to only what is needed

sanmai:
Try running something on Docker:

```
make test-infection-xdebug-74-docker
```
Observe "Sending build context..." counting hundreds of megabytes. This PR makes sure this does not happen, because Docker only needs one directory for the context, nothing else.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@
 /.box_dump/
 /.ci/keys.asc
 /.composer/
-/.dockerignore
 /.idea
 /.php_cs.cache
 /.tools/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -6,6 +6,7 @@
 /.box_dump/
 /.ci/keys.asc
 /.composer/
+/.dockerignore
```
maks-rafalko: why do we need it in `.gitignore`?

sanmai: We 100% don't need it :man_facepalming:

maks-rafalko:
Thank you, @sanmai

PR #1204, opened 2020-03-24
Rename MutantProcessBuilder to MutantProcessFactory

theofidry:
I think it is more appropriate to call it `Factory` than `Builder`, since it's:

- closer to what it does
- a more standard name for that role/pattern
- can be confusion with the actual [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern)

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -449,7 +449,7 @@ final class Container
             },
             MutationTestingRunner::class => static function (self $container): MutationTestingRunner {
                 return new MutationTestingRunner(
-                    $container->getMutantFactoryBuilder(),
+                    $container->getMutantProcessFactory(),
                     $container->getMutantFactory(),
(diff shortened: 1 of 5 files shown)


theofidry:
Failure is due ot codecov

src/Container.php
```diff
@@ -828,9 +828,9 @@ public function getInitialTestsRunner(): InitialTestsRunner
         return $this->get(InitialTestsRunner::class);
     }
 
-    public function getMutantProcessBuilder(): MutantProcessBuilder
+    public function getMutantFactoryBuilder(): MutantProcessFactory
```
maks-rafalko: `getMutantProcessFactory((` ?

theofidry: good catch

PR #1207, opened 2020-03-24
Mutant & Mutation tweaks

theofidry:
I don't think there is a need to keep a duplicated method hence the removal of: 

- `Mutant::isCoveredByTests()`
- `Mutant::getTests()`

I also propose to rename `Mutation::isCoveredByTests()` to `hasTests()` or alternatively `isExecutedByTests()` to avoid the "coverage" notion confusion.

`Mutation::getAllTests()` also has been renamed `getTests()`, since it's unclear otherwise whether it's all the tests for the mutations or all the tests in general. Removing just implies the tests for the mutation

### DIFF
--- a/src/Metrics/MetricsCalculator.php
+++ b/src/Metrics/MetricsCalculator.php
@@ -68,7 +68,7 @@ class MetricsCalculator
     /**
      * @var int
      */
-    private $notCoveredByTestsCount = 0;
+    private $notExecutedByTestsCount = 0;
 
     /**
      * @var int
@@ -110,7 +110,7 @@ class MetricsCalculator
                     break;
 
                 case DetectionStatus::NOT_COVERED:
-                    $this->notCoveredByTestsCount++;
+                    $this->notExecutedByTestsCount++;
                     $this->notCoveredExecutionResults->add($executionResult);
 
                     break;
@@ -164,7 +164,7 @@ class MetricsCalculator
 
     public function getNotTestedCount(): int
     {
-        return $this->notCoveredByTestsCount;
+        return $this->notExecutedByTestsCount;
     }
 
     public function getTotalMutantsCount(): int
--- a/src/Mutation/Mutation.php
+++ b/src/Mutation/Mutation.php
@@ -158,7 +158,7 @@ class Mutation
     /**
      * @return TestLocation[]
      */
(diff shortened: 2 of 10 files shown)


src/Metrics/MetricsCalculator.php
```diff
@@ -110,7 +110,7 @@ public function collect(MutantExecutionResult ...$executionResults): void
                     break;
 
                 case DetectionStatus::NOT_COVERED:
-                    $this->notCoveredByTestsCount++;
+                    $this->notExecutedByTestsCount++;
```
maks-rafalko: I'm -1 on this renaming

* it's clear what `covered` means, I don't think there is any confusion in _"coverage notion"_
* now it's consfusing why the property is named `notExecutedByTestsCount`, but the detection status is `NOT_COVERED`

src/Mutation/Mutation.php
```diff
@@ -150,16 +150,15 @@ public function getMutatedNode(): MutatedNode
         return $this->mutatedNode;
     }
 
-    // TODO: hasTest()?
-    public function isCoveredByTest(): bool
+    public function hasTests(): bool
```
maks-rafalko: same here

src/Metrics/MetricsCalculator.php
theofidry: >it's clear what covered means,

It's not, because it is _not_ covered. The line code coverage only guarantees that a line is _executed_ by a given test, it does not mean this test covers this line.

Also there was a discussion recently with the other mutation testing tools about an alternative to the kill/escape vocabulary and the resounding candidate was covered/uncovered which is going to be even more confusing if we keep those words with the line coverage

I totally agree with you about the detection status, I would like to change it, but that requires a bit more refactoring affecting the UI as well hence I left it alone for now.

theofidry:
Closed in favour of #1209

src/Metrics/MetricsCalculator.php
maks-rafalko: > The line code coverage only guarantees that a line is executed by a given test, it does not mean this test covers this line.

I don't agree. I know the difference between types of coverage, and PHPUnit has *line coverage*, which exactly means that line is covered as soon as **any** of the opcode is executed.

PR #1209, opened 2020-03-24
Refactor/mutant less

theofidry:
This PR:

- [ ] Adds new feature ...
- [ ] Covered by tests

### DIFF
--- a/src/Mutant/MutantExecutionResultFactory.php
+++ b/src/Mutant/MutantExecutionResultFactory.php
@@ -57,14 +57,13 @@ class MutantExecutionResultFactory
     public function createFromProcess(MutantProcess $mutantProcess): MutantExecutionResult
     {
         $process = $mutantProcess->getProcess();
-        $mutant = $mutantProcess->getMutant();
-        $mutation = $mutant->getMutation();
+        $mutation = $mutantProcess->getMutation();
 
         return new MutantExecutionResult(
             $process->getCommandLine(),
             $this->retrieveProcessOutput($process),
             $this->retrieveDetectionStatus($mutantProcess),
-            $mutant->getDiff(),
+            $mutation->getDiff(),
             $mutation->getMutatorName(),
             $mutation->getOriginalFilePath(),
             $mutation->getOriginalStartingLine()
@@ -86,7 +85,7 @@ class MutantExecutionResultFactory
 
     private function retrieveDetectionStatus(MutantProcess $mutantProcess): string
     {
-        if (!$mutantProcess->getMutant()->getMutation()->hasTests()) {
+        if (!$mutantProcess->getMutation()->getMutation()->hasTests()) {
             return DetectionStatus::NOT_COVERED;
         }
 
--- a/src/Mutant/MutantExecutionResultFactory.php
+++ b/src/Mutant/MutantExecutionResultFactory.php
@@ -86,7 +86,7 @@ class MutantExecutionResultFactory
 
     private function retrieveDetectionStatus(MutantProcess $mutantProcess): string
     {
-        if (!$mutantProcess->getMutant()->isCoveredByTest()) {
+        if (!$mutantProcess->getMutant()->getMutation()->hasTests()) {
             return DetectionStatus::NOT_COVERED;
         }
 
--- a/src/Mutation/Mutation.php
+++ b/src/Mutation/Mutation.php
@@ -37,13 +37,10 @@ namespace Infection\Mutation;
 
 use function array_intersect_key;
 use function array_keys;
+use Closure;
 use function count;
-use function implode;
 use Infection\AbstractTestFramework\Coverage\TestLocation;
 use Infection\Mutator\ProfileList;
-use Infection\PhpParser\MutatedNode;
-use function md5;
-use PhpParser\Node;
 use function Safe\array_flip;
 use Webmozart\Assert\Assert;
 
@@ -64,33 +61,26 @@ class Mutation
 
     private $originalFilePath;
     private $mutatorName;
-    private $mutatedNodeClass;
-    private $mutatedNode;
-    private $mutationByMutatorIndex;
     private $attributes;
-    private $originalFileAst;
     private $tests;
     private $executedByTests;
+    private $calculateState;
 
     /**
-     * @var string|null
+     * @var MutationCalculatedState|null
      */
-    private $hash;
+    private $calculatedState;
 
     /**
-     * @param Node[] $originalFileAst
-     * @param array<string|int|float> $attributes
      * @param TestLocation[] $tests
+     * @param Closure(): MutationCalculatedState $calculateState
      */
     public function __construct(
         string $originalFilePath,
-        array $originalFileAst,
         string $mutatorName,
         array $attributes,
-        string $mutatedNodeClass,
-        MutatedNode $mutatedNode,
-        int $mutationByMutatorIndex,
-        array $tests
+        array $tests,
+        Closure $calculateState
     ) {
         Assert::oneOf($mutatorName, array_keys(ProfileList::ALL_MUTATORS));
 
@@ -99,14 +89,16 @@ class Mutation
         }
 
         $this->originalFilePath = $originalFilePath;
-        $this->originalFileAst = $originalFileAst;
         $this->mutatorName = $mutatorName;
         $this->attributes = array_intersect_key($attributes, array_flip(self::ATTRIBUTE_KEYS));
-        $this->mutatedNodeClass = $mutatedNodeClass;
-        $this->mutatedNode = $mutatedNode;
-        $this->mutationByMutatorIndex = $mutationByMutatorIndex;
         $this->tests = $tests;
         $this->executedByTests = count($tests) > 0;
+        $this->calculateState = $calculateState;
+    }
+
+    public function getFilePath(): string
+    {
+        return $this->getCalculatedState()->getMutationFilePath();
     }
 
     public function getOriginalFilePath(): string
@@ -114,25 +106,9 @@ class Mutation
         return $this->originalFilePath;
     }
 
-    /**
-     * @return Node[]
-     */
-    public function getOriginalFileAst(): array
-    {
-        return $this->originalFileAst;
-    }
-
     public function getMutatorName(): string
     {
         return $this->mutatorName;
-    }
-
-    /**
-     * @return (string|int|float)[]
-     */
-    public function getAttributes(): array
-    {
-        return $this->attributes;
     }
 
     public function getOriginalStartingLine(): int
@@ -140,19 +116,19 @@ class Mutation
         return (int) $this->attributes['startLine'];
     }
 
-    public function getMutatedNodeClass(): string
-    {
-        return $this->mutatedNodeClass;
-    }
-
-    public function getMutatedNode(): MutatedNode
-    {
-        return $this->mutatedNode;
-    }
-
     public function hasTests(): bool
     {
         return $this->executedByTests;
+    }
+
+    public function getMutatedCode(): string
+    {
+        return $this->getCalculatedState()->getMutatedCode();
+    }
+
+    public function getDiff(): string
+    {
+        return $this->getCalculatedState()->getDiff();
     }
 
     /**
@@ -165,21 +141,11 @@ class Mutation
 
     public function getHash(): string
     {
-        return $this->hash ?? $this->hash = $this->createHash();
+        return $this->getCalculatedState()->getMutationHash();
     }
 
-    private function createHash(): string
+    private function getCalculatedState(): MutationCalculatedState
     {
-        $hashKeys = [
-            $this->originalFilePath,
-            $this->mutatorName,
-            $this->mutationByMutatorIndex,
-        ];
-
-        foreach ($this->attributes as $attribute) {
-            $hashKeys[] = $attribute;
-        }
-
-        return md5(implode('_', $hashKeys));
+        return $this->calculatedState ?? $this->calculatedState = ($this->calculateState)();
     }
 }
--- a/src/Mutation/Mutation.php
+++ b/src/Mutation/Mutation.php
@@ -158,7 +158,7 @@ class Mutation
     /**
      * @return TestLocation[]
      */
-    public function getAllTests(): array
+    public function getTests(): array
     {
         return $this->tests;
     }
--- a/src/Mutation/Mutation.php
+++ b/src/Mutation/Mutation.php
@@ -70,7 +70,7 @@ class Mutation
     private $attributes;
     private $originalFileAst;
     private $tests;
-    private $coveredByTests;
+    private $executedByTests;
 
     /**
      * @var string|null
@@ -106,7 +106,7 @@ class Mutation
         $this->mutatedNode = $mutatedNode;
         $this->mutationByMutatorIndex = $mutationByMutatorIndex;
         $this->tests = $tests;
-        $this->coveredByTests = count($tests) > 0;
+        $this->executedByTests = count($tests) > 0;
     }
 
     public function getOriginalFilePath(): string
@@ -150,10 +150,9 @@ class Mutation
         return $this->mutatedNode;
     }
 
-    // TODO: hasTest()?
-    public function isCoveredByTest(): bool
+    public function hasTests(): bool
     {
-        return $this->coveredByTests;
+        return $this->executedByTests;
     }
 
     /**
--- a/src/Mutation/MutationFactory.php
+++ b/src/Mutation/MutationFactory.php
@@ -33,10 +33,13 @@
 
 declare(strict_types=1);
 
-namespace Infection\Mutant;
+namespace Infection\Mutation;
 
+use function implode;
+use Infection\AbstractTestFramework\Coverage\TestLocation;
 use Infection\Differ\Differ;
-use Infection\Mutation\Mutation;
+use Infection\PhpParser\MutatedNode;
+use function md5;
 use PhpParser\Node;
 use PhpParser\PrettyPrinterAbstract;
 use function Safe\sprintf;
@@ -45,7 +48,7 @@ use function Safe\sprintf;
  * @internal
  * @final
  */
-class MutantFactory
+class MutationFactory
 {
     private $tmpDir;
     private $differ;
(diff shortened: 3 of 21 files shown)


src/Engine.php
```diff
@@ -155,12 +155,15 @@ private function runMutationAnalysis(): void
...
-            : $actualExtraOptions;
+        $mutationExtraOptions = $this->adapter instanceof ProvidesInitialRunOnlyOptions
+            ? $this->testFrameworkExtraOptionsFilter->filterForMutationProcess(
+                $extraOptions,
+                $this->adapter->getInitialRunOnlyOptions()
+            )
+            : $extraOptions;
 
-        $this->mutationTestingRunner->run($mutations, $filteredExtraOptionsForMutant);
+        $this->mutationTestingRunner->run($mutations, $mutationExtraOptions);
```
theofidry: The real change here is `filterForMutantProcess` -> `filterForMutationProcess`, the rest is pure CS

src/Mutation/Mutation.php
```diff
@@ -99,88 +89,64 @@ public function __construct(
         }
 
         $this->originalFilePath = $originalFilePath;
-        $this->originalFileAst = $originalFileAst;
         $this->mutatorName = $mutatorName;
+        // TODO: move this up
```
theofidry: TODO

sanmai:
I'm not so sure `MutationCalculatedState` is a way better than a `Mutant`.

theofidry:
> It is somewhat like calling a doorknob a handle-you-pull-to-open-a-door.

Still better than calling it a pullable-door :P

I'm happy to take another name. Keep in mind that though that it is strictly internal to `Mutation` (and by extension `MutationFactory)` unlike `Mutant` was

src/Mutation/MutationFactory.php
```diff
@@ -0,0 +1,220 @@
...
+            ): MutationCalculatedState {
+                return $this->calculateState(
+                    $originalFilePath,
+                    $originalFileAst,
+                    $mutatorName,
+                    $attributes,
+                    $mutatedNodeClass,
+                    $mutatedNode,
+                    $mutationByMutatorIndex
+                );
```
sanmai: Why a callback here is better than a DI we use elsewhere?

theofidry:
>That's a lot of changes here. I wish I could review functional changes separately from name changes.

I wanted that as well, but I also needed the final state to judge if was worth it or not. But my recommendation is to checkout locally `Mutation`, `MutationFactory`, `MutantCodeFactory` & `MutationTestingRunner`: those are the ones that changed. All the others are:

- renaming
- using `Mutation` instead of `Mutant`

and frankly are quite minimal and uninteresting from a review PoV. It was really useful for me though to judge the final result and get an early feedback about the design

src/Mutation/MutationFactory.php
theofidry: for laziness. An alternative would be to either:

- calculate everything upfront, but then we would have a performance regression
- inject the required services in `Mutation` and push back the complexity of calculating the fields in `Mutation`

The later is the most sensible option if you don't like this closure, but I found the closure way easier to handle (and in the tests as well since you can just create a dummy calculated state, you don't require any services)

sanmai: How much gain do you think we get from this additional laziness?

theofidry: Can give it a go but I suspect 5-10%.

Note that it won't change much IMO:

- in the factory you pass the values directly instead of the closure, ok this part is simpler
- in `Mutation` you pass the values directly instead of the closure, ok a bit simpler
- everywhere else: not much difference

The closure returns a simple object, so instead of having:

```php
static function () {
     return new MutationCalculatedState('hash', 'changedCode', 'diff');
}
```

You will have:

```php
'hash', 
'changedCode', 
'diff'
```

theofidry: +12% https://blackfire.io/profiles/compare/37e0648d-2fd7-4b21-92ba-13797432a618/graph

sanmai: > but I suspect 5-10%.

I did some benchmarks, and didn't notice anything like that. More like to the contrary. If you can find a proof there's a positive impact, it'd be great. Else we'd complicating things without a reason.

theofidry: - the hash is used only if you create a process, which we skip for uncovered mutations
- the diff for now is used only for the reporting, so it depends on the reports used
- same for mutated code

theofidry: In other words: if you use the debug report where we display the diff regardless of the status of the mutation, then it won't make a difference indeed.

In any other scenario, it will

theofidry: Removed for now

theofidry:
Update as per @sanmai comments, this PR has changed.

You can see it with the following rationale: `Mutant` = `Mutation` + some calculated state, but since at no point we skip any `Mutation`, we have a 1:1 relationship between `Mutant` & `Mutation`. As a result it makes sense to just merge both and avoid any confusion.

I hope this makes the reasoning a bit easier, but the PR needs no more tweaking for being reviewed :)

maks-rafalko:
I'm afraid I don't understand this renaming. Every resource about mutation testing have a definition of **Mutant**, which is now just removed.

A couple of well known ones:

* Equivalent **Mutant**
* High Order **Mutant**
* Second Order **Mutant**
* etc.

Also

* [Wikipedia's article](https://en.wikipedia.org/wiki/Mutation_testing) has this definition
* [Mutation Testing Repository ](http://crestweb.cs.ucl.ac.uk/resources/mutation_testing_repository/)'s articles use this definition
* https://leanpub.com/mutationtesting book [distinguish](https://user-images.githubusercontent.com/3725595/77860469-38720700-7218-11ea-80d3-0dcfdd4cbbb7.png) Mutant and Mutation

So, merging these 2 different things makes the things even more confusing IMO (these are my thought not looking into the code yet).

Also, didn't see any mentions of `Mutagenesis` before in any of the "official" docs. Where does it come from?

---

Another concern is about HOM (High Order Mutations).

> Higher order mutations are the combinations of two or more single mutations. 
>
> E.g. a second order mutation is the combination of two mutations, a third order mutation is the combination of three mutations. 
> [...]
> Most cases, HOM just creates a new mutant [...]

So, HOM is a *combination* of several mutations (N) which results in 1 Mutant. N:1 relations, not 1:1 which is conflicts with your statement:

> ... we have a 1:1 relationship between Mutant & Mutation

---

I still don't agree with the changes from https://github.com/infection/infection/pull/1207 that are presented here as well.

Didn't have enough time to learn the code (though spent 20 minutes already), but at this point I'm -1 on these renaming.

theofidry:
>I'm afraid I don't understand this renaming. Every resource about mutation testing have a definition of Mutant, which is now just removed.

Because there it implies an applied changed or at best is a context where mutation & mutant are interchangeable. 

>Also, didn't see any mentions of Mutagenesis before in any of the "official" docs. Where does it come from?

https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/46584.pdf

>So, merging these 2 different things makes the things even more confusing IMO (these are my thought not looking into the code yet).

My answer to this statement is:

>At the end of the day, my conclusion was: either "mutation" and "mutant" are synonyms in the context of mutation testing, or if we consider them being different, the difference lies in the fact that the mutant designates the mutated program, i.e. where the mutation has been applied, and mutation the transformation to apply (but not yet applied). However, in infection this is not the case, since Mutant is a kind of Mutation with some calculated state.

>So, HOM is a combination of several mutations (N) which results in 1 Mutant. N:1 relations, not 1:1 which conflicts with your statement:

I'll check if done in mutant. If not done there, I highly doubt we'll ever have to worry about that.

But supposedly, that would still be perfectly doable by instead of having a class `Mutation`, an abstraction with `PlainMutation` and `HighOrderMutation` for example.

>Didn't have enough time to learn the code (though spent 20 minutes already), but at this point I'm -1 on these renaming.

As mentioned, the main change really is `Mutation` & `MutationFactory`. The rest is smooth & boring adaptations

theofidry:
Maybe let's put the problem another way: I don't think the current distinction between `Mutation` & `Mutant` _in infection_ has any sense (code-wise) and compares to the more general purpose definitions we are referring to. Hence the proposal to merge it. If you think `Mutant` would be a better name than `Mutation`, I'm happy to go for that as well

theofidry:
@sanmai @maks-rafalko I've updated the description as an attempt to better express myself and address some of the concerns mentioned.

If there is any specific piece (e.g. @maks-rafalko you mentioned the renaming cases from #1207) I'll be happy to address them (in the case of those renaming cases @maks-rafalko I'll revert it as soon as I update that PR).

I would however like a first review/consensus before spending more time updating this PR :)

src/Configuration/ConfigurationFactory.php
```diff
@@ -214,8 +213,6 @@ private static function retrieveCoverageBasePath(
         string $configDir,
         string $tmpDir
     ): string {
-        Assert::nullOrStringNotEmpty($existingCoveragePath);
```
theofidry, line 217: We do not care about its existence here: this responsibility is moved to the CoverageChecker

theofidry:
Rebased & reverted the namespace changes: so it requires another PR to fix some naming inconsistencies but that also lighten (a bit) the amount of stuff to review and opens up a dedicated space to discuss some naming concerns instead of doing it here.

I'll appreciate another attentive review on it, it was also very time consuming for me to review, revert & rebase :)

theofidry:
ping @infection/core

BackEndTea:
👍 I think this makes the language of the project a bit simpler, as well as the code

src/Mutant/MutantExecutionResultFactory.php
```diff
@@ -86,7 +85,7 @@ private function retrieveProcessOutput(Process $process): string
 
     private function retrieveDetectionStatus(MutantProcess $mutantProcess): string
     {
-        if (!$mutantProcess->getMutant()->isCoveredByTest()) {
+        if (!$mutantProcess->getMutation()->hasTests()) {
```
maks-rafalko, line 88: I still not convinced by this change. It is confusing IMO (`isCoveredByTest()` -> `hasTests()`).

src/Mutant/MutationFactory.php
```diff
@@ -66,35 +74,103 @@ public function __construct(
...
+
+        $hash = self::createHash(
+            $originalFilePath,
+            $mutatorName,
+            $attributes,
+            $mutationByMutatorIndex
+        );
+
+        $mutationFilePath = sprintf(
+            '%s/mutation.%s.infection.php',
```
maks-rafalko, line 110: I have to say I'm lost here. 

Even from the nomenclature added in this PR, Mutation is:

> The result of applying a mutator **to the AST** of a subject and represents **a change to be applied**.

But the result source code after applying a _Mutation_ is called **Mutant**. 

> New program that differs from the original by applying a mutation.

IMO this conflicts with definitions from Mutation Testing theory (not even in Infection in particular)

---

After more digging into this patch, here is what we do in this PR:

```php
$mutatedCode = $codeFactory->createCode();

# later in `MutationTestingRunner.php` we dump it to FS

$filesystem->dumpFile('mutation.%s.infection.php', $mutatedCode)
```

How it was before (from my POV - much better)

```php
$mutatedCode = $mutantCodeFactory->createCode();

# later in `MutationTestingRunner.php` we dump it to FS

$filesystem->dumpFile('mutant.%s.infection.php', $mutatedCode)
```

Let's also see the name of `$mutationFilePath` variable. How can Mutation have a file path? From nomenclature, Mutation _"... represents a change to be applied"_, while Mutant is _"New program that differs from the original by applying a mutation."_

New program can have a file (where this program is located).
Change to be applied - not. (Mutation is _"The result of applying a mutator to the AST of a subject"_).

We store AST in memory, and we dumpt the result of mutated AST (which is a Mutant) to a file.

tests/phpunit/Mutator/BaseMutatorTestCase.php
```diff
@@ -81,19 +81,19 @@ protected function setUp(): void
             $this->fail('Input code cant be the same as mutated code');
         }
 
-        $mutants = $this->mutate($inputCode, $settings);
+        $mutatedCodes = $this->mutate($inputCode, $settings);
```
maks-rafalko, line 84: again, the result of mutation is Mutant. As for me, the previous variable name is perfectly ok

tests/phpunit/Process/Runner/MutationTestingRunnerTest.php
```diff
@@ -145,46 +135,24 @@ public function test_it_applies_and_run_the_mutations(): void
...
-            )
-        ;
-
         $this->fileSystemMock
             ->expects($this->exactly(2))
             ->method('dumpFile')
             ->withConsecutive(
-                ['/path/to/mutant0', 'mutated code 0'],
-                ['/path/to/mutant1', 'mutated code 1']
+                ['/path/to/mutation0.php', 'mutated code 0'],
```
maks-rafalko, line 142: same here - dumped *new* mutated code is called Mutant, not Mutation

src/Mutant/MutationFactory.php
```diff
@@ -66,35 +74,103 @@ public function __construct(
         $this->tmpDir = $tmpDir;
         $this->differ = $differ;
         $this->printer = $printer;
-        $this->mutantCodeFactory = $mutantCodeFactory;
+        $this->codeFactory = $mutantCodeFactory;
```
maks-rafalko, line 77: I'm not sure this renaming is needed. This is exactly *mutant* code factory

maks-rafalko:
> 'll check if done in mutant. If not done there, I highly doubt we'll ever have to worry about that.

Are you saying that Mutant (ruby lib) is the source of truth for us? I doubt. If there is something is (not) implemented, I have no idea why we need to (not) do the same.

src/Mutant/MutantExecutionResultFactory.php
theofidry: Hm thought I reverted it

theofidry:
>Are you saying that Mutant (ruby lib) is the source of truth for us? I doubt. If there is something is (not) implemented, I have no idea why we need to (not) do the same.

It's not a source of truth. It is however a reliable source of a more mature version (9yo now?). I also like to take pitest as a reference.

So that mutant does or doesn't do something doesn't translate in a _need_ for us to imitate. It can however provide some insight on a usage. But to more broadly: we have yet to have anything concrete on higher order mutants, so trying to introduce an abstraction for the sole purpose of it at this point would be IMO unwise

src/Mutant/MutationFactory.php
theofidry: I'm really thinking hard on this, so let's see if that makes sense.

The way I see it, the changed AST, a string diff, or a dumped file, is not the mutant, but the mutation.

I interpret _"New program that differs from the original by applying a mutation."_ more as "the process with the mutated code loaded", i.e. in our case, the "process in which we intercept the autoloader to load the mutated code instead of the original".

Maybe this is too much of a stretch? But honestly this makes more sense to me and also has another merit: it is not specific to PHP/Infection. Dumping the mutated code to the FS is a Infection specificity due to how PHP works, I don't think calling _that_ a mutant is correct

theofidry:
>I have a feeling that 2 things are mixed up here which really confuse me:

@maks-rafalko I totally get that feeling :) It's unfortunately a case where I feel both the point of contortion and confusion is about the naming. Because objectively code-wise this PR removes an extra abstraction we have no use of (at the very least for now) and (mostly as a result of that removal) simplifies things

maks-rafalko:
> I interpret "New program that differs from the original by applying a mutation." more as "the process with the mutated code loaded", i.e. in our case, the "process in which we intercept the autoloader to load the mutated code instead of the original".

> Calling that [the dumped file] a mutant does not sound correct to me

^ this is where we have a conflicted understanding of what the Mutant is.

For example, here is how Wikipedia [treat it](https://en.wikipedia.org/wiki/Mutation_testing):

mutant_wiki

Another example from Mutation Testing book:

mutant_book

theofidry:
### Pitest

<details>

<summary>Mutant.java</summary>

```java
/**
 * A fully generated mutant
 */
public final class Mutant {

  private final MutationDetails details;
  private final byte[]          bytes;

  public Mutant(final MutationDetails details, final byte[] bytes) {
    this.details = details;
    this.bytes = bytes;
  }

  /**
   * Returns a data relating to the mutant
   *
   * @return A MutationDetails object
   */
  public MutationDetails getDetails() {
    return this.details;
  }

  /**
   * Returns a byte array containing the mutant class
   *
   * @return A byte array
   */
  public byte[] getBytes() {
    return this.bytes;
  }

}
```

</details>

But no `Mutation.java`

### mbj/mutant

<details>

<summary>mutation.rb</summary>

```rb
# Represent a mutated node with its subject
  class Mutation
    include AbstractType, Adamantium::Flat
    include Concord::Public.new(:subject, :node)

    CODE_DELIMITER = "\0"
    CODE_RANGE     = (0..4).freeze

    # Identification string
    #
    # @return [String]
    def identification
      "#{self.class::SYMBOL}:#{subject.identification}:#{code}"
    end
    memoize :identification

    # Mutation code
    #
    # @return [String]
    def code
      sha1[CODE_RANGE]
    end
    memoize :code

    # Normalized mutation source
    #
    # @return [String]
    def source
      Unparser.unparse(node)
    end
    memoize :source

    # The monkeypatch to insert the mutation
    #
    # @return [String]
    def monkeypatch
      Unparser.unparse(subject.context.root(node))
    end
    memoize :monkeypatch

    # Normalized original source
    #
    # @return [String]
    def original_source
      subject.source
    end

    # Test if mutation is killed by test reports
    #
    # @param [Result::Test] test_result
    #
    # @return [Boolean]
    def self.success?(test_result)
      self::TEST_PASS_SUCCESS.equal?(test_result.passed)
    end

    # Insert mutated node
    #
    # @param kernel [Kernel]
    #
    # @return [Loader::Result]
    def insert(kernel)
      subject.prepare
      Loader.call(
        binding: TOPLEVEL_BINDING,
        kernel:  kernel,
        source:  monkeypatch,
        subject: subject
      )
    end

  private

    # SHA1 sum of source and subject identification
    #
    # @return [String]
    def sha1
      Digest::SHA1.hexdigest(subject.identification + CODE_DELIMITER + source)
    end
    memoize :sha1

    # Evil mutation that should case mutations to fail tests
    class Evil < self

      SYMBOL            = 'evil'
      TEST_PASS_SUCCESS = false

    end # Evil

    # Neutral mutation that should not cause mutations to fail tests
    class Neutral < self

      SYMBOL            = 'neutral'
      TEST_PASS_SUCCESS = true

    end # Neutral

    # Noop mutation, special case of neutral
    class Noop < Neutral

      SYMBOL            = 'noop'
      TEST_PASS_SUCCESS = true

    end # Noop

  end # Mutation
```
</details>

But no `Mutant`


### Stryker

<details>

<summary>Mutant.ts</summary>

```ts
interface Mutant {
  mutatorName: string;
  fileName: string;
  range: Range;
  replacement: string;
}
```
</details>

No `Mutation` interface.

You also mention two sources from mutation testing book, mine comes from another. So I don't think there is a consensus here and admittedly, if you go to someone and say we have:

```php
$x = $a + $b;
```

and there is a "X":

```php
$x = $a - $b
```

Instead of "X", what is the more fitted? "mutation" or "mutant"? Do you think many people have a strong opinion about it and more importantly does it impact the understanding the the above.

Out of curiosity I submitted that part of the nomenclature to the mutation testing slack to see some other opinions. Will report back...

@maks-rafalko just so we move forward a bit: are we disagreeing only on the mutant/mutation definition or are we also disagreeing in merging the two classes into one? (If that's the case could you expand on to why?)

theofidry:
Reporting on the poll: so far the following projects agree on that piece of the nomenclature:

- mull
- mutant
- mutmut

theofidry:
Had a brief read again... Personally my opinion didn't change tbh and I feel this simplifies the code too. But if you feel differently this should be closed? I can otherwise rebase it, but I don't want to do so if it's just for it to stay open for years.

theofidry:
Also fun fact that's the first link of the research paper you've recently linked (the one you linked is slightly more recent and detailed however)

theofidry:
FYI there is a couple more papers listed [here](https://github.com/theofidry/awesome-mutation-testing?tab=readme-ov-file#papers) if you're interested

theofidry:
@sanmai if @maks-rafalko is open to the idea, I can rebase it, but from the diff you can see this was simply removing the `Mutant` class itself and simplifying a bit the code IMO

theofidry:
Also maybe there is more gradual changes that could be extracted to ease the review...

theofidry:
Experimented more recently. If this changes goes through it will be done a different way so as is I do not think it is worth the effort to try to get those changes in.

PR #1210, opened 2020-03-25
Introduce MutationAttributeKeys

theofidry:
This is part of https://github.com/infection/infection/pull/1209 but I'm wondering if it really belongs there. Basically with have those attribute keys which are used in Mutation & the lexer, so maybe it's better to extract it.


sanmai:
There are several places where the attributes can be accessed as constants. E.g. in `getOriginalStartingLine`

theofidry:
I'll double check it in a separate PR since I don't want to create too many conflicts with #1209

PR #1216, opened 2020-03-26
Tweak the memory limiter

theofidry:
Just some cleanup/polishing

### DIFF
--- a/src/Engine.php
+++ b/src/Engine.php
@@ -64,7 +64,7 @@ final class Engine
     private $coverageChecker;
     private $eventDispatcher;
     private $initialTestsRunner;
-    private $memoryLimitApplier;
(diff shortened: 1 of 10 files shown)


src/Engine.php
```diff
@@ -143,7 +143,9 @@ private function runInitialTestSuite(): void
             $initialTestSuitProcess->getOutput()
         );
 
-        $this->memoryLimitApplier->applyMemoryLimitFromProcess($initialTestSuitProcess, $this->adapter);
+        // Limit the memory used for the mutation processes based on the memory used for the initial
+        // test run
```
maks-rafalko: don't think it should be on the second line

theofidry: trying to stick to 100

PR #1218, opened 2020-03-26
Tweak PhpProcess

theofidry:
- Move PhpProcess from Infection\Console\Util to Infection\Process
- Rename PhpProcess to XdebugProcess


theofidry:
Because it's about manipulating XdebugHandler

sanmai:
Well, at what point it does anything with XdebugHandler? True, it does use a class from XdebugHandler to manipulate environment variables, but they may or may not change a thing in terms of Xdebug usage. Even you mention in the comment that "for example" it'd modify environment to reinstate Xdebug if it was.

If it would be 100% launching a process with Xdebug, I would understand this name change. But it isn't.

theofidry:
Then `XdebugHandlerProcess`? `PhpProcess` does not makes sense to me: it's not more "php" than the parent `Process` and does not manipulate PHP directly either

maks-rafalko:
What about `OriginalPhpProcess` / `OriginalPhpConfigProcess` or something like that?

PR #1219, opened 2020-03-26
Tweak the initial test run process factory

theofidry:
- Rename `InitialTestRunBuilder` to `InitialTestsRunFactory` (with the `s` to be consistent with the runner `InitialTestsRunner`)
- Remove the optional parameters from `InitialTestRunBuilder` & `InitialTestsRunner`: they do not need to be optional
- Move the factories from `Infection\Process\Builder` to `Infection\Process\Factory`
- Consolidate the tests. Notable change: I no longer mock `Process` in `InitialTestsRunner` but instead favoured an integration test there


src/Engine.php
```diff
@@ -130,8 +130,8 @@ private function runInitialTestSuite(): void
 
         $initialTestSuitProcess = $this->initialTestsRunner->run(
             $this->config->getTestFrameworkExtraOptions(),
-            $this->config->shouldSkipCoverage(),
-            explode(' ', (string) $this->config->getInitialTestsPhpOptions())
+            explode(' ', (string) $this->config->getInitialTestsPhpOptions()),
```
maks-rafalko: I'm not sure this change is needed (swapping parameters)

theofidry: no but I also dislike having the parameters declared in a certain order in the method and another order for the properties declaration, assignment and method orders, so just made sure they are all in the same order

theofidry: Actually that's not the case here so nvm, will revert that

theofidry: It was to keep the same order as `TestFrameworkAdapter::getInitialTestRunCommandLine`

PR #1221, opened 2020-03-27
Tweak the FullyQualifiedClassNameVisitor

theofidry:
- Introduce `FullyQualifiedClassNameManipulator` which is a tiny wrapper around the node attributes API to provide typehint safety
- Restrict FQCN to `FullyQualified` instead of `Name`
- Consolidate the tests


theofidry:
If you are uncomfortable with a new utility class, a friendly alternative is to add those methods in the visitor itself

src/PhpParser/Visitor/FullyQualifiedClassNameVisitor.php
```diff
@@ -36,33 +36,34 @@
...
 
     public function enterNode(Node $node): ?Node
     {
-        if ($node instanceof Stmt\Namespace_) {
+        if ($node instanceof Node\Stmt\Namespace_) {
             $this->namespace = $node->name;
-        } elseif ($node instanceof Stmt\ClassLike) {
-            // TODO: check the cases of multiple namespaces
-            $node->setAttribute(self::FQN_KEY, $node->name ? Name::concat($this->namespace, $node->name->name) : null);
+        } elseif ($node instanceof Node\Stmt\ClassLike) {
```
sanmai:
```suggestion

            return null;
        }

        if ($node instanceof Node\Stmt\ClassLike) {
```

PR #1224, opened 2020-03-27
Tweak ParentVisitor

theofidry:
Rework the test of the parent visitor: I think the previous one just kind of make sure all nodes have a parent, but not that the parent is the correct one. The current test, while less automatic, is better at letting us double checking that the result is the expected one.

Also moved out the nodes handling to a dedicated utility class `ParentConnector` like done in #1221


src/Mutator/Boolean/TrueValue.php
```diff
@@ -92,8 +92,8 @@ public function canMutate(Node $node): bool
             return false;
         }
 
-        $parentNode = $node->getAttribute(ParentConnectorVisitor::PARENT_KEY);
-        $grandParentNode = $parentNode !== null ? $parentNode->getAttribute(ParentConnectorVisitor::PARENT_KEY) : null;
+        $parentNode = ParentConnector::findParent($node);
+        $grandParentNode = $parentNode !== null ? ParentConnector::findParent($parentNode) : null;
```
sanmai: `if ($parentNode === null) return null; ...` ?

src/Mutator/Operator/Continue_.php
```diff
@@ -77,12 +77,8 @@ public function canMutate(Node $node): bool
...
 
-        $parentNode = $node->getAttribute(ParentConnectorVisitor::PARENT_KEY);
+        $parentNode = ParentConnector::findParent($node);
 
-        if ($parentNode instanceof Node\Stmt\Case_) {
-            return false;
-        }
-
-        return true;
+        return !($parentNode instanceof Node\Stmt\Case_);
```
sanmai: Are parentheses required here? It was a thing in older versions of PHP IIRC.

src/PhpParser/Visitor/ReflectionVisitor.php
```diff
@@ -140,26 +140,22 @@ private function isPartOfFunctionSignature(Node $node): bool
...
 
-        if (!$node->hasAttribute(ParentConnectorVisitor::PARENT_KEY)) {
-            return false;
-        }
-
-        $parent = $node->getAttribute(ParentConnectorVisitor::PARENT_KEY);
+        $parent = ParentConnector::findParent($node);
 
-        return $parent instanceof Node\Param || $node instanceof Node\Param;
+        return $parent !== null && ($parent instanceof Node\Param || $node instanceof Node\Param);
```
sanmai: `if ($parent === null) {return false;} ...`?

src/PhpParser/Visitor/ParentConnectorVisitor.php
```diff
@@ -59,20 +57,23 @@ public function beforeTraverse(array $nodes): ?array
...
+    }
+
+    public function enterNode(Node $node)
     {
         $stackCount = count($this->stack);
 
-        if ($stackCount !== 0) {
-            $node->setAttribute(self::PARENT_KEY, $this->stack[$stackCount - 1]);
-        }
+        ParentConnector::setParent($node, $this->stack[$stackCount - 1] ?? null);
```
sanmai: I like the explicit `if ($stackCount !== 0) {}` better. Easier to read.

src/Mutator/Boolean/TrueValue.php
theofidry: it was a ternary before, I don't really fancy replacing it; If we want to get rid of ternary operators it would be better to replace them everywhere IMO

src/Mutator/Operator/Continue_.php
theofidry: they are not, it's just clearer IMO. But might just be me

src/PhpParser/Visitor/ParentConnectorVisitor.php
theofidry: The difference here is not the count check, but that the attribute is set (to null) even if no parent is found

src/Mutator/Operator/Break_.php
```diff
@@ -77,8 +77,8 @@ public function canMutate(Node $node): bool
             return false;
         }
 
-        $parentNode = $node->getAttribute(ParentConnectorVisitor::PARENT_KEY);
+        $parentNode = ParentConnector::findParent($node);
 
-        return !$parentNode instanceof Node\Stmt\Case_;
+        return !($parentNode instanceof Node\Stmt\Case_);
```
sanmai: parentheses?

src/Mutator/Operator/Continue_.php
sanmai: As of current master, we have about 37 cases where parentheses are not used in a similar situation, and only 6 where are.

src/Mutator/Boolean/TrueValue.php
sanmai: Fair enough.

src/PhpParser/Visitor/ParentConnectorVisitor.php
sanmai: So there are cases where a parent has to be unregistered, is this right?

theofidry: In practice it doesn't make any difference – when it was not set we default to `null` from `getAttribute()`, but i thought that we might as well have all our "metadata" available so that upon inspection you see clearly everything

src/Mutator/Operator/Continue_.php
theofidry: I'll do it in another PR

src/Mutator/Operator/Break_.php
maks-rafalko: I agree it looks strange

theofidry: what is strange to me to is to have `!` applying on the expression for `instanceof` instead of the variable like everywhere else :)

Anyway I don't want to debate about this CS: we'll stick to the convention used I'll just update it in a separate PR everywhere rather than just here

maks-rafalko: That's the point, since we didn't agree with any of the format, we should continue using the *current* one, to avoid inconsistencies. But let's forget about this case and don't waste everyone's time...

theofidry: sorry I didn't realise this was on another thread. As said [there](https://github.com/infection/infection/pull/1224#discussion_r400072709), I'll fix it

maks-rafalko: 👍

PR #1225, opened 2020-03-27
Reduce RunCommand state

theofidry:
This PR:

- Removes `RunCommand::initialize()`: the rationale being we don't need it to be there. This method is a hook to request more user input _before_ the input is validated, i.e. if you wanna fiddle with it. We don't have that case though, so might as well not abuse of it
- Removes `RunCommand#container`: it is as easy to pass the container around in our case
- Removes `RunCommand#consoleOutput`: it is as easy to pass it

The benefit of it is that now if you look at `RunCommand::initialize()`:

- the restart & init container appears clearly
- the startup no longer initialize state of the command that can be re-used later (so less-error prone)


theofidry:
Looks like you cannot skip those arguments :(

theofidry:
ping @infection/core

src/Command/BaseCommand.php
```diff
@@ -57,11 +58,15 @@ abstract class BaseCommand extends Command
...
     protected $output;
 
+    /**
+     * @var SymfonyStyle
+     */
+    protected $io;
+
     protected function initialize(InputInterface $input, OutputInterface $output): void
     {
-        parent::initialize($input, $output);
```
maks-rafalko, line 62: not calling parent's method is IMO a bad practice.

Yes, now the parent method is empty, but it does exist. Nothing prevents Symfony's developers to add some logic there later, which in this case won't be executed.

theofidry: Yet you don't for `execute` or `setUp()` & `tearDown()` in PHPUnit :)

I understand where this comes from, but it's not always a good idea:

- Symfony _cannot_ add logic there later: that would be a BC (it's the same in several places in the Symfony codebase and a recurrent issue of offering an inheritance base extension point)
- If Symfony do add some logic there (BC or not), what makes you think that we should call it? As there is no more guarantee than calling this is going to or not break things than not calling it

However all of that said, I also don't feel strongly about it either. So if you prefer that call here I'll add it back

Issue #197, opened 2018-02-27
What are the lowest supported versions of each test framework that infection supports?


BackEndTea:
TODO:

* [ ] Find out exactly what the lowsest versions are of each supported test framework
* [ ] Add documentation for that on https://github.com/infection/site
* [ ] Add an e2e test that requires the use of that version, so we know if we no longer support that version
* [ ] Add a constraint to `composer.json`

borNfreee:
The lowest versions I worked with using infection are:

* PHPUnit 5.5.7
* phpspec 3.4.0

But I don't exactly the minimum ones

theofidry:
We also should have a conflict rule in the `composer.json` for it. It won't help with the PHAR (yet at least) but still worth it IMO :)

BackEndTea:
@theofidry I'm sorry i dont follow what you mean

theofidry:
I mean [something like this](https://github.com/nelmio/alice/blob/master/composer.json#L37-L39). It's not a good example because it should prevent Symfony 5. But when you install Infection as a library it will prevent you to be able to install it if you have the wrong versions

sidz:
lowest version for the different versions of php interpreter will be different. For example phpUnit < 6 with php 7.2 will throw Fatal error:

```
PHP Fatal error:  Declaration of SebastianBergmann\Comparator\DOMNodeComparator::assertEquals($expected, $actual, $delta = 0, $canonicalize = false, $ignoreCase = false) must be compatible with SebastianBergmann\Comparator\ObjectComparator::assertEquals($expected, $actual, $delta = 0, $canonicalize = false, $ignoreCase = false, array &$processed = Array) in phar:///var/www/infection/phpunit-5.7.9.phar/sebastian-comparator/DOMNodeComparator.php on line 110
PHP Stack trace:
PHP   1. {main}() /var/www/infection/phpunit-5.7.9.phar:0

Fatal error: Declaration of SebastianBergmann\Comparator\DOMNodeComparator::assertEquals($expected, $actual, $delta = 0, $canonicalize = false, $ignoreCase = false) must be compatible with SebastianBergmann\Comparator\ObjectComparator::assertEquals($expected, $actual, $delta = 0, $canonicalize = false, $ignoreCase = false, array &$processed = Array) in phar:///var/www/infection/phpunit-5.7.9.phar/sebastian-comparator/DOMNodeComparator.php on line 110

Call Stack:
[...]

PR #1227, opened 2020-03-28
PoC: check the test framework adapter version before proceeding

theofidry:
I think it would be beneficial to provide an API for the test framework adapter to check that the test framework used is a valid version. Here is the example with this PoC with PHPUnit 5:

Screen Shot 2020-03-28 at 12 58 46
Screen Shot 2020-03-28 at 13 00 26

(verbose mode with the trace cut off)


src/TestFramework/AbstractTestFrameworkAdapter.php
```diff
@@ -123,34 +124,19 @@ public function getMutantCommandLine(
         );
     }
 
+    /**
+     * @throws UnsupportedTestFrameworkVersion
```
theofidry: I think it's better to throw when we cannot identify the version, rather than carrying on with the unknown version... Up for debate though

src/TestFramework/AbstractTestFrameworkAdapter.php
```diff
@@ -123,34 +124,19 @@ public function getMutantCommandLine(
...
-        try {
-            $version = $this->versionParser->parse($process->getOutput());
-        } catch (InvalidArgumentException $e) {
-            $version = 'unknown';
-        } finally {
-            $this->version = $version;
-        }
-
-        return $this->version;
+        return $this->version ?? $this->version = $this->retrieveVersion();
```
theofidry: Tiny refactoring to move the "retrieve code" in a clear place and just show clearly that we memoize the result here

src/TestFramework/AbstractTestFrameworkAdapter.php
```diff
@@ -123,34 +124,19 @@ public function getMutantCommandLine(
...
     }
 
+    /**
+     * @throws UnsupportedTestFrameworkVersion
+     */
+    abstract public function checkVersion(): void;
```
theofidry: this should be pushed in the `TestFrameworkAdapter` interface. The goal IMO is for it to be called by the test framework factory (not the adapter specific one, but the `Factory` one like done in this PR)

src/TestFramework/AbstractTestFrameworkAdapter.php
```diff
@@ -190,6 +176,34 @@ private function getCommandLine(
...
+            ['--version']
+        );
+
+        $process = new Process($testFrameworkVersionExecutable);
+        $process->mustRun();
+
+        try {
+            return $this->versionParser->parse($process->getOutput());
+        } catch (InvalidVersion $exception) {
+            throw new UnsupportedTestFrameworkVersion(sprintf(
```
theofidry: change here from the refactoring: `VersionParser` now throws a friendly runtime-error instead of an `InvalidArgumentException` (which should never be caught - it's a logic exception). We now catch it to throw a more friendly error

src/TestFramework/Factory.php
```diff
@@ -85,6 +85,15 @@ public function __construct(
     }
 
     public function create(string $adapterName, bool $skipCoverage): TestFrameworkAdapter
+    {
+        $adapter = $this->constructAdapter($adapterName, $skipCoverage);
+
+        $adapter->checkVersion();
```
theofidry: the check version I was talking about earlier

src/TestFramework/Factory.php
```diff
@@ -85,6 +85,15 @@ public function __construct(
...
+
+        return $adapter;
+    }
+
+    private function constructAdapter(string $adapterName, bool $skipCoverage): TestFrameworkAdapter
```
theofidry: `construct`? `instantiate`? we need an alternative to `create` here

src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
```diff
@@ -52,6 +53,8 @@ class PhpUnitAdapter extends AbstractTestFrameworkAdapter implements IgnoresAddi
 {
     public const COVERAGE_DIR = 'coverage-xml';
 
+    private const MINIMUM_SUPPORTED_VERSION = '6.0.0';
```
theofidry: version up for debate: I would argue it may be time to push for a higher version, e.g. one that at least supports pcov

src/TestFramework/AbstractTestFrameworkAdapter.php
maks-rafalko: 1. note that this class in not used by any other adapter
2. we need to make it inside `abstract-testframework-adapter` first, not here

src/TestFramework/Factory.php
maks-rafalko: .. `buildAdapter` / `initializeAdapter`

any option, just with `Adapter` postfix, please

src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
maks-rafalko: `pcov` support is added in `8.0`. However, `8.0` has a very annoying BC break - removed [`assertArraySubset`](https://github.com/sebastianbergmann/phpunit/issues/3494#issuecomment-464073900) which from what I know prevents some devs from upgrading (there is a separate package for it though).

Personally, I'm ok even with bumping it to `8.0`, but I can't answer to myself why would we do it.

No infection works perfectly with PHPUnit `5.2`, the only issue I found is that `pcov` is not supported.

src/TestFramework/AbstractTestFrameworkAdapter.php
maks-rafalko: I would not add it to the main `TestFrameworkAdapter` interface, but instead create a new one, like `MemoryUsageAware` for example. Not sure all test frameworks need it.

src/TestFramework/Version/VersionParser.php
```diff
@@ -46,16 +45,20 @@ final class VersionParser
...
     {
         $matches = [];
         $matched = preg_match(self::VERSION_REGEX, $content, $matches);
 
-        Assert::notSame(
-            $matched,
-            0,
-            sprintf('Expected "%s" to be contain a valid SemVer (sub)string value.', $content)
-        );
+        if ($matched === 0) {
```
maks-rafalko: `$matched !== 1` ? since negative return values are `0`, `false`

src/TestFramework/AbstractTestFrameworkAdapter.php
theofidry: yes, but that was a PoC :)

src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
theofidry: I would like to provide a better experience than waiting for the test framework to fully run to get an error saying that the coverage could not be generated :/

src/TestFramework/Version/VersionParser.php
theofidry: since it's the safe function it's going to be `0` or `1` and throw an error on `false`

src/TestFramework/AbstractTestFrameworkAdapter.php
theofidry: I feel they should: maybe the version we support is very low and that's fine, but I would really prefer us to throw with a friendly error if the adapter is using an unsupported version rather than letting the user in the dark with a potentially very weird error

src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
theofidry: Maybe it's just too optimistic of me: since PHPUnit does not support it, I don't think we should either

maks-rafalko: so what version are you suggesting to stick to and why? 6.0.0? 8.0.0?

src/TestFramework/AbstractTestFrameworkAdapter.php
maks-rafalko: do you have an idea what the min version for PhpSpec?

maks-rafalko: by the way, Codeception's min version is 3.1.1, and this is controlled by composer itself, so you won't even be able to install it:

https://github.com/maks-rafalko/infection-codeception-adapter/blob/9321a43f57f4ef9ac980bb4cd74376fe900e59a5/composer.json#L56-L58

theofidry: >do you have an idea what the min version for PhpSpec?

No not really :/

>by the way, Codeception's min version is 3.1.1, and this is controlled by composer itself, so you won't even be able to install it:

Not sure to understand what you mean there: the issue is with the PHAR were we instrument a tool (PHPUnit, Codeception & co.) we have no control over

sanmai:
This looks like a good idea on the one hand, but on the other hand I don't see a substantial evidence backing the need for such drastic measures. 

Yes, there are couple of places where we have to track for older formats of coverage, but they can be (if not) thoroughly unit-tested, therefore there isn't a parsing problem per se.

src/TestFramework/AbstractTestFrameworkAdapter.php
maks-rafalko: > Not sure to understand what you mean there

I mean that *in addition* to this check, we should add the same `conflict` setting to PhpSpec's adapter to fail as earlier as possible for composer-based install. PHAR distribution will be handled by this new `checkVersion` runtime method.

But again, since you are not sure (and me as well) what is the minimum PhpSpec version supported by Infection, and we don't even know criteria, I still think it shouldn't be on the `TestFrameworkAdapter` level, but as a separate interface

theofidry:
@sanmai it's not only about the coverage reports (so far even supporting PHPUnit 5.x is still ok), but _also_ the test framework instrumentation which includes:

- retrieving the version
- generating the command line for the tests
- generating the command line for the tests for mutations
- generating the config file (in some cases) for the mutations
- parsing the output to know if the test failed or not

I'll be honest here: I'm not quite sure we are thorough in testing all of those either so I would not be surprised if we have a few non-working cases.

<hr/>

>But again, since you are not sure (and me as well) what is the minimum PhpSpec version supported by Infection, and we don't even know criteria, I still think it shouldn't be on the TestFrameworkAdapter level, but as a separate interface

What minimum version we want to support means what minimum version _we are willing to look at_ in case of a bug. So it also depends on your, personally I never used neither phpspec or codeception so I'm not keen to look at it even on recent versions at all.

But I admit if that was only based on that, I would be a lot more aggressive in the minimum version too. So maybe a solution is to go first for a permissive arbitrary version and we can always adapt it later, e.g:

After all, we don't want to break stuff for users. So one thing is say we are in theory pushing for phpspec 4.0.0 minimum and users come at us saying they would still like to have it work with phpspec 3.0.0, then we can lower that option and we know at least what minimum version to look at for our testing.

- phpspec: 3.0.0 (https://packagist.org/packages/phpspec/phpspec#3.0.0) which is the first release to allow PHP7
- codeception 3.0.0 (https://packagist.org/packages/codeception/codeception#3.0.0): same – although is it compatible with it @maks-rafalko?
- PHPUnit 5.0

I would also like at some point us to be more through with the tests (see the full list above) and have this unit/integration tests at least for each major version, (e.g. if we support PHPUnit 5.x, then it means 5.x, 6.x, 7.x, 8.x, 9.x). Is that a lot? For sure, but also realistically between each major version something can break there.

<hr/>

@maks-rafalko provided we go for a minimum version for each major framework (and if you disagree on the specific versions proposed up there I propose 1.0.0 for them...), would you agree to add this to `TestFrameworkAdapter`?

maks-rafalko:
> would you agree to add this to TestFrameworkAdapter?

sure, thanks, proposed versions sounds good to me

> codeception 3.0.0 (https://packagist.org/packages/codeception/codeception#3.0.0): same – although is it compatible with it @maks-rafalko?

When I had the first implementation of codeception adapter, the minimum version with my fixes inside `codeception/codeception` was 3.1.1. But after that, we completely changed implementation and those fixes in `3.1.1` are no longer needed for Infection, so I *think* that 3.0.0 would work, however I had no chance to check it

BackEndTea:
This also fixed #197

src/TestFramework/AbstractTestFrameworkAdapter.php
BackEndTea: What kind of version output do you get if you run a dev version of phpunit? I wouldn't prohibit users from using infection, just because we cannot parse the version info. We should probably output a big warning though.

theofidry: Hm fair

src/TestFramework/AbstractTestFrameworkAdapter.php
theofidry: Good idea 👍

src/TestFramework/AbstractTestFrameworkAdapter.php
theofidry: Thinking about this again: it might actually be sure way better to, even in much lower versions, stick to displaying a warning about using an unsupported version rather than bailing out at all

src/TestFramework/AbstractTestFrameworkAdapter.php
```diff
@@ -127,11 +131,32 @@ public function getVersion(): string
         return $this->version ?? $this->version = $this->retrieveVersion();
     }
 
+    /**
+     * @throws UnsupportedTestFrameworkVersion
+     *                                         TODO: add this to the interface upstream
```
maks-rafalko, line 136: is this TODO still relevant?

src/TestFramework/Version/VersionParser.php
```diff
@@ -46,16 +46,20 @@ final class VersionParser
...
+            throw new InvalidVersion(sprintf(
+                'Expected "%s" to be contain a valid SemVer (sub)string value.',
```
maks-rafalko, line 59: probably just `to contain`? Just see such construction for the first time :)

PR #1232, opened 2020-03-30
InstanceOf_ mutator

sanmai:
- [x] Doc PR: https://github.com/infection/site/pull/164

Quite simple one. Replaces:
```php
$foo instanceof Anything
```
With either `true` or `false`.
- When `true` we assume the value always the of same type, and the check is redundant.
- When `false` we assume the testing isn't thorough.

Report for this mutator:

```
338 mutations were generated:
     316 mutants were killed
      22 covered mutants were not detected
       0 errors were encountered
       0 time outs were encountered

```


maks-rafalko:
A couple of questions:

1. How a developer should kill it? Or is this mutator should show the redundant check (like `public -> protected` indicating that `$a instance of $b` is not needed? Can we add such info to description of the mutator?

2. I would suggest instead of producing 2 mutations (`true`, `false`) do the following
    ```diff
    -$a instance of $b
    +! $a instance of $b
    ```
    
    Why? Because producing 2 mutations (`true`, `false`) duplicates original value (being it `true` of `false`). With the suggested above mutation, we will have only one new mutation - the opposite to the original. What do you think?

sanmai:
> How a developer should kill it?

By ensuring this check is actually required by using appropriately typed arguments. I think we have to come up with a style of explanation before writing any of them since stating the obvious might not be as useful.

I went with the negation approach you suggested, can you have a look?

src/Mutator/Boolean/InstanceOf_.php
```diff
@@ -0,0 +1,74 @@
...
+     * @return iterable<Node\Expr\ConstFetch>
+     */
+    public function mutate(Node $node): iterable
+    {
+        yield new Node\Expr\ConstFetch(new Node\Name('true'));
+
+        yield new Node\Expr\ConstFetch(new Node\Name('false'));
+    }
+
+    public function canMutate(Node $node): bool
```
theofidry: I think `canMutate()` should be placed before `mutate()`, but since it's hardly something we can enforce don't mind me if you don't agree with that

tests/phpunit/Fixtures/SimpleMutationsCollectorVisitor.php
```diff
@@ -43,7 +43,7 @@ public function leaveNode(Node $node)
 
         // It is important to not rely on the keys here. It might otherwise result in some elements
         // being overridden, see https://3v4l.org/JLN73
-        foreach($this->mutator->mutate($node) as $mutatedNode) {
+        foreach ($this->mutator->mutate($node) as $mutatedNode) {
```
theofidry: I thought the CS fixer would pick that up...

src/Mutator/Boolean/InstanceOf_.php
sanmai: Well, I took another mutator for a template, and that's how it was there.

tests/phpunit/Fixtures/SimpleMutationsCollectorVisitor.php
sanmai: It doesn't fix fixtures for some reason I didn't yet have time to investigate.

Issue #1238, opened 2020-04-05
Updated to Infection 0.16.0. Exclude param in the configuration file is ignored 


ecourtial:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.16.1@1928f71f526a3078ae2a2e83ca8f250cb2407346
| Test Framework version | PHPUnit 9.0.1
| PHP version | 7.4.x
| Platform    | e.g. Debian Alpine (for CI)/MacOS (for dev)
| Github Repo | https://github.com/ecourtial/Fighting-Steel-Dynamic-Campaign-Manager

- With the previous version (0.15), the Symfony src/Kernel.php file was properly ignored, and I had a 100% MSI. Now it dropped to 92% because this file is no longer ignored as it should as defined in the configuration file.

- Clone the repo. Do a composer install. Launch Infection (make infection or _vendor/bin/infection_.

<details>
 <summary>phpunit.xml</summary>

 ```xml
<?xml version="1.0" encoding="UTF-8"?>

<!-- https://phpunit.readthedocs.io/en/latest/configuration.html -->
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
         backupGlobals="false"
         colors="true"
         bootstrap="tests/bootstrap.php"
         executionOrder="random"
         resolveDependencies="true"
>
    <php>
        <ini name="error_reporting" value="-1" />
        <server name="APP_ENV" value="test" force="true" />
        <server name="SHELL_VERBOSITY" value="-1" />
        <server name="SYMFONY_PHPUNIT_REMOVE" value="" />
        <server name="SYMFONY_PHPUNIT_VERSION" value="7.5" />
        <env name="SYMFONY_DEPRECATIONS_HELPER" value="weak" />
    </php>

    <testsuites>
        <testsuite name="Project Test Suite">
            <directory>tests</directory>
        </testsuite>
    </testsuites>

    <filter>
        <whitelist processUncoveredFilesFromWhitelist="true">
            <directory suffix=".php">src</directory>
            <exclude>
                <file>src/Kernel.php</file>
            </exclude>
        </whitelist>
    </filter>

    <listeners>
        <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
    </listeners>
</phpunit>

 ```
</details>

<details>
 <summary>infection.json.dist</summary>

 ```json
{
[...]

PR #1239, opened 2020-04-07
Return back `$finder->notPath($excludedDir)` to filter out excluded files configured in `infection.json`

maks-rafalko:
Fixes https://github.com/infection/infection/issues/1238

Unfortunately, 2 issues here:

1. Here https://github.com/infection/infection/commit/b2f6ddb4eb752d271d5e359c7e5d3d4016b72af6 `$finder->notPath(..)` was removed and it caused finder not respect excluded files from `infection.json`

2. Here https://github.com/infection/infection/pull/1106 we introduced another issue: when we get the files from coverage report, we don't respect `source.excludes` config from `infection.json`, and mutate all files that present in the coverage report regardless of the exclusion configuration

I have fixed the first item here, but have difficulties with the second, so @sanmai I need your help here.

I believe the best (from performance POV) place to filter out excluded files is in this file where I left `TODO` https://github.com/infection/infection/compare/bugfix/exclude-file?expand=1#diff-feed5b44802aab03ea332afc7740b86dR83

Since calling `SourceFileInfoProvider::provideFileInfo()` is quite an expensive thing, we need to filter out files here to reduce the result array of files (iterator).

However, I'm not sure how to do it better. In Symfony's finder we just call `->notPath($excludedFirOrFile)`, but here we have only path that we read from the XML coverage report.

Any ideas?

PS

1. The first issue is unit tested (pass)
2. For the second point, I've added a functional test you can play with (or at least have a look at). It's currently failing, obviously.

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -178,7 +178,8 @@ final class Container
             },
             SourceFileFilter::class => static function (self $container): SourceFileFilter {
                 return new SourceFileFilter(
-                    $container->getConfiguration()->getSourceFilesFilter()
+                    $container->getConfiguration()->getSourceFilesFilter(),
+                    $container->getSchemaConfiguration()->getSource()->getExcludes()
                 );
             },
             JUnitTestExecutionInfoAdder::class => static function (self $container): JUnitTestExecutionInfoAdder {
--- a/src/FileSystem/SourceFileFilter.php
+++ b/src/FileSystem/SourceFileFilter.php
@@ -42,6 +42,7 @@ use Infection\FileSystem\Finder\Iterator\RealPathFilterIterator;
 use Infection\TestFramework\Coverage\Trace;
 use Iterator;
 use SplFileInfo;
+use Symfony\Component\Finder\Iterator\PathFilterIterator;
 
 /**
  * @internal
@@ -54,12 +55,22 @@ class SourceFileFilter
      */
     private $filters;
 
-    public function __construct(string $filter)
+    /**
+     * @var string[]
+     */
+    private $excludeDirectories;
+
+    /**
+     * @param string[] $excludeDirectories
+     */
+    public function __construct(string $filter, array $excludeDirectories)
     {
         $this->filters = array_filter(array_map(
             'trim',
             explode(',', $filter)
         ));
+
+        $this->excludeDirectories = $excludeDirectories;
     }
 
     /**
@@ -81,15 +92,21 @@ class SourceFileFilter
      */
     public function filter(iterable $input): iterable
     {
-        if ($this->filters === []) {
-            return $input;
+        $iterator = $this->iterableToIterator($input);
+
+        if ($this->filters !== []) {
+            $iterator = new RealPathFilterIterator(
+                $iterator,
+                $this->filters,
+                []
+            );
         }
 
-        return new RealPathFilterIterator(
-            $this->iterableToIterator($input),
-            $this->filters,
-            []
-        );
+        if ($this->excludeDirectories !== []) {
+            $iterator = new PathFilterIterator($iterator, [], $this->excludeDirectories);
+        }
+
+        return $iterator;
     }
 
     /**
--- a/src/TestFramework/Coverage/Trace.php
+++ b/src/TestFramework/Coverage/Trace.php
@@ -60,6 +60,8 @@ interface Trace
      */
     public function getRealPath(): string;
 
+    public function getRelativePathname(): string;
+
     public function hasTests(): bool;
 
     public function getTests(): TestLocations;
--- a/src/FileSystem/SourceFileCollector.php
+++ b/src/FileSystem/SourceFileCollector.php
@@ -66,6 +66,10 @@ class SourceFileCollector
             ->name('*.php')
         ;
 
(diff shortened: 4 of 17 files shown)


sanmai:
This is somewhat unfortunate that we have to do this. Can we consider dropping this feature?

For example, in the current state of affairs this feat can be implemented by excluding a file from coverage [by means provided by phpunit.xml](https://stackoverflow.com/a/52036485/93540).

theofidry:
There is two scenarios:

- a regular run
- coverage provided

In the 2nd case, I don't think there is much we can do, we need to do as listed above. In the first case though, maybe we can consider whitelisting or blacklisting the collected files which can have two advantages:

- the filtering is easier (only the files we want are "indexed")
- dumping the coverage may be faster (less coverage to dump)

However doing so will likely generate a bigger overhead in generating the PHPUnit config and it's not excluded that the new overhead is bigger than the benefit

maks-rafalko:
> Can we consider dropping this feature?

Personally, I don't like it. 2 points:

* we had this feature, it is useful, and our internal implementation (in this case - using coverage report as a source of mutations) should not be a reason to delete it

* this is a BC for no reason IMO

---

> In the 2nd case, I don't think there is much we can do, we need to do as listed above.

Didn't get it. Before this feature (let's say in 0.15.0) - we generated coverage report for the whole set of source files, but mutated only those that were not excluded in `infection.json`. So these sets were separated and independent.

Now, the set of whitelisted files in coverage report erases `excludes` files from `infection.json`.

Ok, I've managed to make it work same as before with using `PathFilterIterator` that is used by Symfony's Finder under the hood when we called `notPath()`.

If anyone want to go with the XML config instead (adding excludes there) - we can do it in `master`, but for now I'm happy with the fix for 0.16

sanmai:
> If anyone wants to go with the XML config instead

Considering this is a three line fix basically, I don't think it makes sense.

src/Container.php
```diff
@@ -178,7 +178,8 @@ public static function create(): self
             },
             SourceFileFilter::class => static function (self $container): SourceFileFilter {
                 return new SourceFileFilter(
-                    $container->getConfiguration()->getSourceFilesFilter()
+                    $container->getConfiguration()->getSourceFilesFilter(),
+                    $container->getSchemaConfiguration()->getSource()->getExcludes()
```
theofidry, line 182: That doesn't look right to me: either we should move it up in `Configuration` or find something else. The schema configuration is meant to be a 1:1 dependency with the configuration file, which is a "low level" thing: it is not something we want the "domain" of Infection to depend on

src/FileSystem/SourceFileFilter.php
```diff
@@ -54,12 +55,22 @@ class SourceFileFilter
...
-    public function __construct(string $filter)
+    /**
+     * @var string[]
+     */
+    private $excludeDirectories;
+
+    /**
+     * @param string[] $excludeDirectories
+     */
+    public function __construct(string $filter, array $excludeDirectories)
```
theofidry, line 66: Note to myself: check the usage of filters, maybe we do that trim + explode operation more than once

src/FileSystem/SourceFileFilter.php
```diff
@@ -81,15 +92,21 @@ public function getFilters(): array
...
+        if ($this->filters !== []) {
+            $iterator = new RealPathFilterIterator(
+                $iterator,
+                $this->filters,
+                []
+            );
+        }
+
+        if ($this->excludeDirectories !== []) {
+            $iterator = new PathFilterIterator($iterator, [], $this->excludeDirectories);
```
theofidry, line 106: is this right? So if you have both `filters` and `excludes`, you and up with ignoring filters here

src/TestFramework/Coverage/Trace.php
```diff
@@ -60,6 +60,8 @@ public function getSourceFileInfo(): SplFileInfo;
      */
     public function getRealPath(): string;
 
+    public function getRelativePathname(): string;
```
theofidry, line 63: why is this needed? If its for the same reasons as `getRealPath` I would add a comment as well. Otherwise this is confusing.

Note to self: might be worth extracting it in a separate interface for clarity, which `Trace` could extend from

src/FileSystem/SourceFileFilter.php
maks-rafalko: why?

src/TestFramework/Coverage/Trace.php
maks-rafalko: `PathFilterIterator` requires it

src/FileSystem/SourceFileFilter.php
theofidry: ah I see, nevermind then

src/TestFramework/Coverage/Trace.php
maks-rafalko: See https://github.com/infection/infection/pull/1241

src/Container.php
maks-rafalko: See https://github.com/infection/infection/pull/1241

PR #1240, opened 2020-04-10
Bugproof: fails in @dataProvider not recognised as such

Slamdunk:
This PR:

- [x] Covered by tests

References: https://github.com/psr7-sessions/storageless/pull/119#issuecomment-611368754

Hi, I would expect that a mutation affecting `@dataProvider` would be same as the same mutation affecting the test itself, but it is not so.

Is this expected or a bug?

### DIFF
--- a/tests/e2e/PhpUnit_DataProvider/tests/SourceClassTest.php
+++ b/tests/e2e/PhpUnit_DataProvider/tests/SourceClassTest.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace PhpUnit_DataProvider\Test;
+
+use PhpUnit_DataProvider\SourceClass;
+use PHPUnit\Framework\TestCase;
+
+class SourceClassTest extends TestCase
+{
+    /**
+     * @dataProvider instancesProvider
+     */
+    public function testAlwaysTrueFromFactoryMethod(SourceClass $class)
+    {
+        // $class = SourceClass::factoryMethod();
+        self::assertTrue($class->getValue());
+    }
+
+    public function instancesProvider(): array
+    {
+        return [
+            [SourceClass::factoryMethod()],
+        ];
+    }
+}
--- a/tests/e2e/PhpUnit_DataProvider/README.md
+++ b/tests/e2e/PhpUnit_DataProvider/README.md
@@ -1,0 +1,5 @@
+# Title
+
+## Summary
+
+TODO
--- a/tests/e2e/PhpUnit_DataProvider/composer.json
+++ b/tests/e2e/PhpUnit_DataProvider/composer.json
@@ -1,0 +1,17 @@
+{
+    "autoload": {
+        "psr-4": {
+            "PhpUnit_DataProvider\\": "src/"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "PhpUnit_DataProvider\\Test\\": "tests/"
+        }
+    },
+    "require": {
+    },
+    "require-dev": {
+        "phpunit/phpunit": "^8"
+    }
+}
--- a/tests/e2e/PhpUnit_DataProvider/expected-output.txt
+++ b/tests/e2e/PhpUnit_DataProvider/expected-output.txt
@@ -1,0 +1,7 @@
+Total: 3
+
+Killed: 2
+Errored: 0
+Escaped: 1
+Timed Out: 0
+Not Covered: 0
--- a/tests/e2e/PhpUnit_DataProvider/infection.json
+++ b/tests/e2e/PhpUnit_DataProvider/infection.json
@@ -1,0 +1,16 @@
+{
+    "timeout": 25,
+    "source": {
+        "directories": [
+            "src"
+        ]
+    },
+    "logs": {
+        "summary": "infection.log"
+    },
+    "tmpDir": ".",
+    "mutators": {
+        "TrueValue": true,
+        "PublicVisibility": true
+    }
+}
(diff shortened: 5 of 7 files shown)


tests/e2e/PhpUnit_DataProvider/tests/SourceClassTest.php
```diff
@@ -0,0 +1,25 @@
...
+use PHPUnit\Framework\TestCase;
+
+class SourceClassTest extends TestCase
+{
+    /**
+     * @dataProvider instancesProvider
+     */
+    public function testAlwaysTrueFromFactoryMethod(SourceClass $class)
+    {
+        // $class = SourceClass::factoryMethod();
```
Slamdunk, line 15: Uncommenting this line makes the tests passes

sanmai:
[As per comment](https://github.com/psr7-sessions/storageless/pull/119#issuecomment-611908372), we have yet to see this as a bug. Infection mostly takes the coverage provided by the test framework, and it does its job right most of the time.

Slamdunk:
Ok. I'll leave this PR open and give the maintainers the decision on how to handle this.

If is would not be considered as a bug, could we at least document the behavior somewhere?

sanmai:
Can you please clarify what is the problematic behavior?

Slamdunk:
Infection reports a mutation is uncovered, but if I manually apply the mutation the SUT fails, so the mutation is actually covered.

sanmai:
I can't reproduce the issue with 0.16. 

Therefore: https://github.com/psr7-sessions/storageless/pull/120

Slamdunk:
> I can't reproduce the issue with 0.16.

I don't get your reply: this PR reproduce the issue, is targeted against master, and the build fails, so the issue is therefore reproducible

sanmai:
I'm sorry to break for you but that's not how it works. If you take a closer look at the coverage report, you would see that there are several uncovered lines:

```
$ php vendor/bin/phpunit --coverage-text
...
 Summary:                  
  Classes: 66.67% (2/3)    
  Methods: 94.12% (32/34)  
  Lines:   85.12% (103/121)

\PSR7Sessions\Storageless\Http::PSR7Sessions\Storageless\Http\SessionMiddleware
  Methods:  81.82% ( 9/11)   Lines:  76.00% ( 57/ 75)
\PSR7Sessions\Storageless\Session::PSR7Sessions\Storageless\Session\DefaultSessionData
  Methods: 100.00% (12/12)   Lines: 100.00% ( 32/ 32)
\PSR7Sessions\Storageless\Session::PSR7Sessions\Storageless\Session\LazySession
  Methods: 100.00% (11/11)   Lines: 100.00% ( 14/ 14)
```
And that's what Infection tells you.

Slamdunk:
Oh, now I see master is behind:

image

sanmai:
Again, this is not a bug in Infection, be it master or 0.16.x. There's a hole in your test coverage, and that's why you have several uncovered mutations.

Slamdunk:
Got it: infection relies on CC provider to know if a mutation is covered or not, and not strictly to the output of the test. `@dataProvider`s don't count for CC, so any mutation occurring in them is never covered. That's not a fault in infection but in the underlying tools.

Hope I got i right.

sanmai:
> That's not a fault in infection but in the underlying tools.

That's about right. 

There's some logic in the fact that PHPUnit won't include what data providers doing in the coverage report.

sanmai:
Do you want to continue to pursue this PR?

Slamdunk:
No, but I'd like to be sure no one would ever spend 2 hours like me into understand this "issue", if we can call it so

sanmai:
What do you think will have helped you avoid these difficulties? 

For example, we can update the text Infection outputs upon finishing.

Slamdunk:
It appears this is a long standing misunderstanding:

1. https://github.com/sebastianbergmann/phpunit/issues/1450#issuecomment-57428548
1. https://stackoverflow.com/questions/47621926/phpunit-dataproviders-vs-coverage

Ever thou not an Infection "fault", it would be good to pinpoint it in Infection's doc too.

> For example, we can update the text Infection outputs upon finishing.

Mmm, cli output sounds not the right place to me. I propose a `Caveats` chapter inside `Documentation` one in https://infection.github.io/guide/

sanmai:
If you feel like it you can open a PR here: https://github.com/infection/site

Starting here: https://github.com/infection/site/blob/master/src/guide/index.md

theofidry:
Hm I'm not sure I follow everything exactly.

So if I understand correctly, a mutation is marked as escaped but if you apply it manually the tests fail?

If so indeed this is likely due to the coverage report. For example:

- infection is using it wrong (that would be a legit bug)
- the way the coverage is configured is wrong (an issue on your side)
- the coverage generator generates an incorrect coverage report (legit bug but in the coverage generator, e.g. pcov)

So far we have yet to see that it's the first case.

If it's the second case though it has likely to do with the `@covers` annotations or something along those lines, because data providers don't influence the way the coverage works (cf. [this demo](https://github.com/theofidry/coverage-dataprovider-demo)).

In this scenario however, I believe there is something we could do in Infection, not to fix the issue but to debug it more easily. An example would be:

```
infection info <mutation-hash>

- diff
- mutator name
- tests detected
```

But it requires some groundwork before getting there

Slamdunk:
@theofidry as mentioned in the previous comment https://github.com/infection/infection/pull/1240#issuecomment-611937864 there's no bug anywhere.

But it's just very tricky to understand that's not a bug, for the normal user.

You write a test class, you use a code in a `@dataProvider`, you test that code, Infection mutates the code used in the `@dataProvider`: the test fails, but Infection tells the mutation isn't tested because PHPUnit Code-Coverage runs only for the test but not for the `@dataProvider`.

I plan to document this behavior in the doc.

theofidry:
Ha I see, so you are executing the SUT outside the test, which indeed doesn't sound like a good idea :) If there is any way any doc can be improved in this regard that would be good

Slamdunk:
> you are executing the SUT outside the test, which indeed doesn't sound like a good idea :)

Heh, indeed, but not easy to spot, see:

https://github.com/psr7-sessions/storageless/blob/53f4b403e10710f3b477f32945b2fcd1a3d40ab8/test/StoragelessTest/Http/SessionMiddlewareTest.php#L533

Slamdunk:
Doc proposed in https://github.com/infection/site/pull/165

PR #1242, opened 2020-04-12
Removed some dead code

BackEndTea:
This PR cleans up some dead code.

* MemoryUsageAware was moved to a package
* The exception wasn't used anymore
* Some unused variables


src/Logger/DebugFileLogger.php
```diff
@@ -123,7 +123,7 @@ private function getResultsLine(
 
         $separateSections = false;
 
-        foreach ($executionResults as $index => $executionResult) {
+        foreach ($executionResults as $executionResult) {
```
theofidry, line 126: may be an idea of a mutation?

theofidry:
@maks-rafalko I think the failures are a bit weird, I cannot reproduce them locally with no cache enabled

src/Logger/DebugFileLogger.php
maks-rafalko: this mutation will in most cases be killed because of "undefined variable", no?

sanmai: That, unless `$index` in unused.

theofidry: Yeah would be nice to find a way to not blindly do it which would create too many errors. So a potential idea to explore but may not be doable

PR #1244, opened 2020-04-13
Move xdebug-filter under tests

theofidry:
- Easier to find
- Taken care of by the CS fixer with the current config
- Remove the unnecessary leading backslashes (we are in the global namespace already)


maks-rafalko:
Thank you @theofidry

PR #1252, opened 2020-04-16
Introduce IO object

theofidry:
There is a recurrent pain with dealing with IO objects in infection. Indeed we often need not only the input and output, but also the SymfonyStyle. That's 3 objects to pass around everywhere which are quite close to each other.

This PR proposes to encapsulate the 3 with the `IO` object which extends the `SymfonyStyle` for convenience whilst still exposing the input & output.


sanmai:
CI errors with:

>  Cannot retrieve the IO object before the command was initialized

PR #1254, opened 2020-04-16
Move no-progress detection to the ConfigurationFactory

theofidry:
- A few subscribers are relying on `skipProgressBar` which currently is detected in their factories (when registering them in `Container`)
- `skipProgressBar` was relying on an "old" way to detect the CI environment, this PR now leverages the `CiDetector` introduced instead
- The `MutationGenerator` and `MutationTestingRunner` are relying on `noProgress` in order to know if they should run concurrently or not. With this PR, they will run concurrently if a CI environment is detected as well

### DIFF
--- a/src/Configuration/ConfigurationFactory.php
+++ b/src/Configuration/ConfigurationFactory.php
@@ -47,6 +47,7 @@ use Infection\Mutator\MutatorFactory;
 use Infection\Mutator\MutatorParser;
 use Infection\Mutator\MutatorResolver;
 use Infection\TestFramework\TestFrameworkTypes;
+use function getenv;
 use function Safe\sprintf;
 use function sys_get_temp_dir;
 use Webmozart\PathUtil\Path;
@@ -142,7 +143,7 @@ class ConfigurationFactory
             $debug,
             $onlyCovered,
             $formatter,
-            $noProgress,
+            self::retrieveNoProgress($noProgress),
             self::retrieveIgnoreMsiWithNoMutations($ignoreMsiWithNoMutations, $schema),
             self::retrieveMinMsi($minMsi, $schema),
             $showMutations,
@@ -230,6 +231,14 @@ class ConfigurationFactory
         SchemaConfiguration $schema
     ): string {
         return $testFrameworkExtraOptions ?? $schema->getTestFrameworkExtraOptions() ?? '';
+    }
+
+    private static function retrieveNoProgress(bool $noProgress): bool
+    {
+        return $noProgress
+            || getenv('CI') === 'true'
+            || getenv('CONTINUOUS_INTEGRATION') === 'true'
+        ;
     }
 
     private static function retrieveIgnoreMsiWithNoMutations(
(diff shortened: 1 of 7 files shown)


theofidry:
>I'm not sure we need to decide for a developer to run Infection with lost statistics of how many mutants are already processed ((33/1044)), which is now the case.

It is not lost since you see it after the run is done. It just assumes that "--no-progress" (which is enabled by default when a CI is detected) the number is not shown at the intermediate steps

theofidry:
Note that it's is possible to add it back if you feel strongly about it, I left it out because it makes it more complex. I was also thinking of displaying this information nonetheless in debug verbosity.

That said IMO, I feel the issue lies more with the auto-CI detection which cannot be opted out.

sanmai:
This is unfortunate that we have to decide for a developer to whenever display remaining number of mutants in a CI environment, but running with this option gives _the best_ outcome, as appropriate for CI environment. Speed and memory usage wise.

maks-rafalko:
Ok, we can add new info to the docs, explaining all the performance win, we can show a red banner in the CI logs that `--no-progress` is much faster when CI env var is available, but I still not agree that we are deciding it by the user without possibility to change it the way how Infection works on CI.

Probably, we can run it by default with `--no-progress` in CI, but

* informing user about it in the docs (so doc PR is need) and/or in console (to be discussed)
* make it possible to opt-out

theofidry:
>but I still not agree that we are deciding it by the user without possibility to change it the way how Infection works on CI.

Note that this is something we already do. The difference here is we no longer display the mutation count in the middle of the run.

I agree with you though: it is nice to have an auto-CI detection, but there is several things I don't like currently:

- We are pretty much doing (and after this PR exactly doing) `--no-progress` automatically if a CI is detected. Which means in the CI `--no-progress` should not be necessary yet the users are not informed about it
- There is, at least documented, no easy way to disable the CI detection (I guess it comes down to removing the CI global but that might have quite a few side-effects so not ideal)
- There is no logging about this CI detection

As noted earlier though, those are already existing problems. I think they should be addressed, but this is not what this PR is about

maks-rafalko:
> Note that this is something we already do.

I don't get it, tbh. `master` branch automatically detects `CI` env var and *only* disables progress bars. That's it. No more influence.

This PR now automatically enables `--no-progress` behavior, where mutant generator works in a different way, resulting to skipping actual mutated count on each step (row). That is the difference.

IMO this PR does 2 different things:

1. it changes the way CI is detected for *old* places (`$skipProgressBar`)
2. it introduces new behavior when `CI` is found for mutation generator, which concerns me

sanmai:
> make it possible to opt-out

That's an important bit. IIRC this isn't possible currently. E.g. one can circumvent this CI detection.

theofidry:
>IMO this PR does 2 different things:
>
>it changes the way CI is detected for old places ($skipProgressBar)
>it introduces new behavior when CI is found for mutation generator, which concerns me

Yep, but again the new behaviour is not displaying the mutation count, nothing more.

The legit concern here IMO is to be able to force the "progress" behaviour even in CI, but that's something I would like to address in a separate PR

theofidry:
ping @infection/core 

@maks-rafalko I created https://github.com/infection/infection/pull/1261 to ready a way to opt-out from the CI detection

PR #1255, opened 2020-04-17
Tweak Travis & AppVeyor configuration

theofidry:
The main difference is that we now only check the min MSI score in PHP 7.4 with pcov. The reasons being:

- we have tests skipped below 7.4, hence if we want to check against "most" (not all since we have some driver specific tests) the codebase, it should be this build
- pcov is both the fastest and more reliable coverage driver

We are definitely loosing some checks against other drivers, but I feel that the value provided by the min check in those is very limited.

There is a few changes that I'm commenting in the PR


.ci/travis-functions.sh
```diff
@@ -56,7 +56,6 @@ function xdebug_enable() {
 # Globals:
 #   TRAVIS_PULL_REQUEST
 #   TRAVIS_BRANCH
-#   MIN_MSI
```
theofidry, line 59: we were no longer using this one

.ci/travis-functions.sh
```diff
@@ -65,21 +64,18 @@ function xdebug_enable() {
 #   String
 #######################################
 function get_infection_pr_flags() {
-    local flags="--no-progress";
```
theofidry, line 68: this flag has been moved to the travis configuration instead: makes it less "hidden" and also ensures it is used even in a regular non-PR build

.ci/travis-functions.sh
```diff
@@ -65,21 +64,18 @@ function xdebug_enable() {
...
+    local flags="";
     local changed_files;
-    local min_msi;
 
     if ! [[ "${TRAVIS_PULL_REQUEST}" == "false" ]]; then
         git remote set-branches --add origin "$TRAVIS_BRANCH";
         git fetch;
 
         changed_files=$(git diff origin/"$TRAVIS_BRANCH" --diff-filter=A --name-only | grep src/ | paste -sd "," -);
-        min_msi=$(($(grep -o 'min-msi=[0-9]*' .travis.yml | head | cut -f2 -d=) + 1))
```
theofidry, line 77: Is there a reason to _bump_ the MSI score in PRs?

.ci/travis-functions.sh
```diff
@@ -65,21 +64,18 @@ function xdebug_enable() {
...
-
-        >&2 echo "Assumed minimal MSI: $min_msi%"
 
         if [ -n "$changed_files" ]; then
-            flags="--filter=${changed_files} --ignore-msi-with-no-mutations --only-covered --min-msi=${min_msi} --show-mutations $flags";
+            # Set those flags only if there is any changed files detected
+            flags="--filter=${changed_files} --ignore-msi-with-no-mutations --only-covered --show-mutations ${flags}";
```
theofidry, line 78: Should `--only-covered` be used only in PRs?

.travis.yml
```diff
@@ -41,7 +42,7 @@ jobs:
     - &STANDARD_TEST_JOB
       stage: Test
       php: 7.3
-      env: DRIVER="xdebug"
+      env: DRIVER="none"
```
theofidry, line 45: minor change: I'm not changing the build overall here just declaring the script on the none driver instead of Xdebug

.travis.yml
```diff
@@ -56,6 +57,12 @@ jobs:
             if [[ "$DRIVER" == "pcov" ]]; then
               pecl install pcov;
             fi
+          - INFECTION_FLAGS="${INFECTION_FLAGS} $(get_infection_pr_flags)"
```
theofidry, line 60: difference with the previous approach: now we build `INFECTION_FLAGS` directly no more concatenation further down in the process

appveyor.yml
```diff
@@ -49,7 +49,7 @@ install:
 test_script:
     - cd c:\projects\workspace
     - vendor\bin\phpunit --prepend=tests/xdebug-filter.php --coverage-clover=clover.xml --coverage-xml=coverage/coverage-xml --log-junit=coverage/junit.xml
-    - php bin\infection --threads=4 --log-verbosity=none --coverage=coverage
+    - php bin\infection --threads=%NUMBER_OF_PROCESSORS% --log-verbosity=none --coverage=coverage --no-progress --only-covered --ignore-msi-with-no-mutations
```
theofidry, line 52: slight change, let's see if that works

theofidry:
Forgot to leave my comments :P

.ci/travis-functions.sh
maks-rafalko: why not? to force increasing MSI

.travis.yml
```diff
@@ -119,22 +126,18 @@ jobs:
       <<: *STANDARD_TEST_JOB
       php: nightly
       env: DRIVER="phpdbg"
-
       script:
-        # We don't have xdebug on the nightly build, so we always run with phpdbg.
-        - make compile;
-        - ./tests/e2e_tests build/infection.phar;
+        # We don't have Xdebug on the nightly build, so we always run with PHPDBG.
```
maks-rafalko, line 130:
```suggestion
        # We don't have Xdebug on the nightly build, so we always run with phpdbg.
```

this is how it's named in https://github.com/krakjoe/phpdbg#installation

appveyor.yml
maks-rafalko: why do we need `--only-covered` here, and in travis as well?

.ci/travis-functions.sh
theofidry: That's cool if it's a PR with new changes, otherwise when it's refactoring something where there is nothing to improve it's gonna be a red build for nothing.

On a side note I find way more damaging to have non-actionable build failures (e.g. Codecov) because it makes you complacent towards red builds

appveyor.yml
theofidry: To make it faster, and on Travis it was already there (and is only for the PRs)

.ci/travis-functions.sh
maks-rafalko: We have a `--diff-filter=A` that means only added files are counted for MSI checks, so refactoring of the existing files should not fail the build.

theofidry: More than once though git understood a file as bien added instead of moved though; maybe that's why I got that impression a lot

PR #1256, opened 2020-04-17
Simplify container

theofidry:
- Ensures all the services are always configured (note that this won't guarantee everything will work still)
- Better document the default values
- Make easier to re-configure the container

### DIFF
--- a/src/Command/RunCommand.php
+++ b/src/Command/RunCommand.php
@@ -78,20 +78,21 @@ final class RunCommand extends BaseCommand
                     'Name of the Test framework to use ("%s")',
                     implode('", "', TestFrameworkTypes::TYPES)
                 ),
-                ''
+                Container::DEFAULT_TEST_FRAMEWORK
             )
             ->addOption(
                 'test-framework-options',
                 null,
                 InputOption::VALUE_REQUIRED,
-                'Options to be passed to the test framework'
+                'Options to be passed to the test framework',
+                Container::DEFAULT_TEST_FRAMEWORK_EXTRA_OPTIONS
             )
             ->addOption(
                 'threads',
                 'j',
                 InputOption::VALUE_REQUIRED,
                 'Number of threads to use by the runner when executing the mutations',
-                '1'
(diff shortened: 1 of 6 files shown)


src/Command/RunCommand.php
```diff
@@ -78,20 +78,21 @@ protected function configure(): void
                     'Name of the Test framework to use ("%s")',
                     implode('", "', TestFrameworkTypes::TYPES)
                 ),
-                ''
```
maks-rafalko, line 81: it's now `null`, not the empty string. Is it expected change?

src/Command/RunCommand.php
```diff
@@ -78,20 +78,21 @@ protected function configure(): void
...
+                Container::DEFAULT_TEST_FRAMEWORK
             )
             ->addOption(
                 'test-framework-options',
                 null,
                 InputOption::VALUE_REQUIRED,
-                'Options to be passed to the test framework'
+                'Options to be passed to the test framework',
+                Container::DEFAULT_TEST_FRAMEWORK_EXTRA_OPTIONS
```
maks-rafalko, line 88: Why do we need defaut values in the input, where they are `null`s? Looks strange to me and in order to understand what is the value - we need to find a value of this const, which is `null`, meaning "there is no default value"

src/Command/RunCommand.php
theofidry: Yes: in the end it does not change the result thanks to how we process the value (when we take it from the input). So although this change does not change anything downstream, I thought it had the merit of pointing "hey there's a default that depend on that" and that when refactoring this value, you would be lead to think of the default container value

src/Command/RunCommand.php
theofidry: Clarified some bits in https://github.com/infection/infection/pull/1256/files#r410943399, "we need to find a value of this const," is precisely what I'm interested in since changing the default value of the option can result in the default being different than the container known default. Having this here is a nudge to double check that container default

PR #1257, opened 2020-04-17
Introduce an application Logger

theofidry:
This introduces a logger as a service (default being `NullLogger`) which more easily allows to have services relying on a logger which can easily be logging to the output.

The goal is to ultimately get rid of `OutputConsole` – at least in its current form. It's undeniable that what it is doing is necessary, I however would like however to change the way it does it.

### DIFF
--- a/.ci/travis-functions.sh
+++ b/.ci/travis-functions.sh
@@ -56,7 +56,6 @@ function xdebug_enable() {
 # Globals:
 #   TRAVIS_PULL_REQUEST
 #   TRAVIS_BRANCH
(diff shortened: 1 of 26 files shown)


theofidry:
ping @infection/core

theofidry:
>From this statement, it is not clear why it is needed to change anything. For me it looks like moving the code from one place to another.

Indeed :) It's a bit hard to articulate properly without going in big debates, so I prefer to just state my goal without going into details. I hope the small increments I'm doing to get there makes sense in themselves

PR #1258, opened 2020-04-17
Log the debuggers found via the logger instead of ConsoleOutput

theofidry:
I'm trying to play with the logs displayed as I think there is a few things that could be improved there. However I believe this small change could add to the noise hence is worth extracting in a separate PR.

As a result here is this PR which is a small tweak on the ConsoleLogger. What does it do?

In 0.16.3:

Screen Shot 2020-05-01 at 22 43 29

In master (due to #1263):

Screen Shot 2020-05-01 at 22 45 22

Now:

Screen Shot 2020-05-01 at 22 45 55

So in other words:

- the "notice" output (without "IO block") is a very plain message, pretty much what you get with a plain `writeln()`
- the other outputs (other log level) are as "normal" – standard Symfony Console
Screen Shot 2020-05-01 at 22 47 34
 (since #1263 those are like so):

Screen Shot 2020-05-01 at 22 47 27


theofidry:
@sanmai @maks-rafalko do you know where I should look at to see how infection resolves which debugger to use?

maks-rafalko:
What is a "debugger" in this context? :)

theofidry:
sorry I meant code coverage generator... Guess I was too tired :D

sanmai:
I don't think it does. This is a testing tool's job. E.g. PHPUnit's.

maks-rafalko:
> I don't think it does. This is a testing tool's job. E.g. PHPUnit's.

Exactly. I guess it's here

https://github.com/sebastianbergmann/php-code-coverage/blob/28efe6950c2c3ceb86f0fc63aa077617950c78c5/src/CodeCoverage.php#L891-L908

sanmai:
Also note it wasn't like this [as of late 2019](https://github.com/sebastianbergmann/php-code-coverage/commit/8198f748870693e87fe40e19452768cef00869ea). Back then XDebug was used over PCOV if both were present.

tests/phpunit/Logger/ConsoleLoggerTest.php
```diff
@@ -210,22 +232,95 @@ static function (): void {},
...
+        yield [
+            LogLevel::INFO,
+            'message',
+            '[info] message',
+        ];
+
+        yield [
+            LogLevel::DEBUG,
+            'message',
+            '[debug] message',
```
theofidry, line 323: note that this test doesn't capture the formatter tags e.g. here the message that we write is `<debug>[debug] message</debug>` where the tags are handled by the output formatter. I'm don't know however how to test that...

maks-rafalko (approved):
LGTM.

What do you think about moving

`Xdebug detected`

below the LOGO?

IMO it's prettier when we have Infection logo first, then all the rest information, as shown here

Screen Shot 2020-05-01 at 22 45 22

Or it's impossible?

theofidry:
Good idea, done

sanmai (approved):

tests/phpunit/Logger/ConsoleLoggerTest.php
```diff
@@ -208,22 +230,95 @@ static function (): void {},
...
 
         yield [LogLevel::DEBUG, OutputInterface::VERBOSITY_VERY_VERBOSE, false];
 
         yield [LogLevel::DEBUG, OutputInterface::VERBOSITY_DEBUG, true];
+    }
 
-        yield [LogLevel::ALERT, OutputInterface::VERBOSITY_QUIET, false];
+    public static function logLevelProvider(): iterable
+    {
+        yield [
```
theofidry, line 276: this is a new method; should use the enum bucket!

theofidry: cannot: it is not an enum and is from psr/log

staabm:
> I think there is a few things that could be improved there

do you have your ideas written down somewhere (e.g. in a issue)?

I would add to this list, that it would be great if infection could stream output of escaped mutations while running, instead of output nothing for sometimes 15-20 minutes and in the end print a huge wall of text

theofidry:
@staabm I need to review both exactly what this PR does and your suggestion.

I only half-heartedly rebased this PR which is 5y old because I was feeling bad to keep creating PRs and thought I should have a little look to try to close already opened ones 😰

staabm:
ohh.. I didn't realize this PR was that old :)

theofidry:
Nw, your comment is still relevant. I don't know if I would address it here or document it in a new or existing issue, I will see!

PR #1259, opened 2020-04-18
Introduce FormatterName

theofidry:
- Introduce the enum `FormatterName` to list all the known formatter names
- Inject the output as a service
- Leverage the output as a service to introduce a `FormatterFactory`
- Register the formatter as a service and inject directly in the services using a formatter instead of requiring them to instantiate the formatter
- Remove the formatter name from the configuration


theofidry:
ping @infection/core

PR #1261, opened 2020-04-25
Add option to opt-out the CI detection

theofidry:
- Add a new `--force-progress` which disables the CI detection, effectively ensuring the user will see all the progress bars and the mutation count (during the run – the ones in the summary are _always_ displayed independently from noProgress)
- Ensures the options `forceProgress` and `noProgress` cannot be set together both from the CLI or by manipulation the container directly
- Replaces the regular usage of `CiDetector` by `MemoizedCiDetector` which memoize the CI detection call


src/Command/RunCommand.php
```diff
@@ -110,7 +111,13 @@ protected function configure(): void
                 'no-progress',
                 null,
                 InputOption::VALUE_NONE,
-                'Do not output progress bars'
+                'Do not output progress bars and mutation count during progress. Automatically enabled if a CI is detected'
+            )
+            ->addOption(
+                'progress',
```
maks-rafalko, line 117: what about `--with-progress`?

```
--no-progress
--with-progress
```

but the current one is also ok, just another idea :)

tmp
```diff
@@ -0,0 +1,35 @@
+Progress =
```
maks-rafalko, line 1: to be removed?

src/Command/RunCommand.php
theofidry: `force-progress`? It's more in line with the description and if we have `with-progress` I would think of `with-no-progress` or `without-progress` rather than `no-progress`

maks-rafalko: `--force-progress` 👌

maks-rafalko:
We forgot about documenting this new option. Could you please add a PR to the docs?

PR #1262, opened 2020-04-25
Install test framework adapters as dev dependencies

maks-rafalko:
Since Infection is for development purposes, adapter must be installed as `dev` dependencies as well.

php-cs-fixer's failure is unrelated and fixed in master


maks-rafalko:
Thank you @theofidry & @BackEndTea

PR #1267, opened 2020-05-05
Move retrieve version in a dedicated method

theofidry:
Extracted from #1227. This allows to better show the "retrieve version" logic and the "getVersion" which memoize the call.


theofidry:
Hm indeed, although to note this is a very unlikely error so in practice we do not expect any change from the user experience

maks-rafalko:
Unfortunately, we had this issue at least on Travis many times before (not sure about the current state), when Travis' PHPUnit was used instead of the one from the `vendor` folder. 

But yeah, I think this will be a very rare case.

theofidry:
I see; Well I hope it's not a case poping up anymore and otherwise it will be an opportunity to revisit our approach

Issue #1273, opened 2020-06-18
Escaped Mutant does not escape when mutation is tested manually


sebastianbergmann:
I started to use Infection in one of my projects, [`code-unit`](https://github.com/sebastianbergmann/code-unit), and noticed something weird. Could be a bug, or I may be confused -- in which case this might be a support issue.

This is my `infection.json` configuration file:
```json
{
    "source": {
        "directories": [
            "src"
        ]
    },
    "mutators": {
        "@default": true
    },
    "minMsi": 100,
    "minCoveredMsi": 100
}
```

Here is the output I get:
```
./tools/infection --show-mutations
You are running Infection with PCOV enabled.

    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/

Infection - PHP Mutation Testing Framework 0.16.3@52d597af80429d52dd6218fcb9766fc7653ec88c

Running initial test suite...

PHPUnit version: 9.2.3

   52 [============================] < 1 sec

Generate mutants...

Processing source code files: 15/15
.: killed, M: escaped, S: uncovered, E: fatal error, T: timed out

.......................................SS....SS..S   ( 50 / 127)
SSSSSSSSS............MM.M.........M...MM.......M..   (100 / 127)
......M.......S.SS.SS......                          (127 / 127)
Escaped mutants:
================


1) /usr/local/src/code-unit/src/CodeUnitCollectionIterator.php:31    [M] DecrementInteger

--- Original
+++ New
@@ @@
     }
     public function rewind() : void
     {
-        $this->position = 0;
+        $this->position = -1;
     }
     public function valid() : bool
     {


2) /usr/local/src/code-unit/src/CodeUnitCollectionIterator.php:31    [M] DecrementInteger

--- Original
+++ New
@@ @@
     }
     public function rewind() : void
     {
-        $this->position = 0;
+        $this->position = -1;
     }
     public function valid() : bool
     {


3) /usr/local/src/code-unit/src/CodeUnitCollectionIterator.php:51    [M] Increment

--- Original
+++ New
@@ @@
     }
     public function next() : void
     {
-        $this->position++;
+        $this->position--;
     }
 }


[...]

PR #1280, opened 2020-07-06
Set `failOnRisky`, `failOnWarning` to `true` if parameters are not already set

### LINKED ISSUE #1274: Update phpunit.xml to fail on risky tests by default

#1273 showed that there are cases where a mutation may cause a test to become risky, yet these tests are not a cause of failure for PHPUnit by default.

The fix is a easy as adding [`failOnRisky="true"` to `phpunit.xml`](https://github.com/sebastianbergmann/code-unit/commit/33dcbf5a8ae6a555248e7e3c7b02eb3cbbd74b87), yet it is not as obvious as one may think it is.

Since PHPUnit won't treat new risky tests as failures by default, and since this behavior causes an unnecessary confusion even for experienced users (see #1273), I suggest that we add `failOnRisky="true"` and `failOnWarning="true"` by default unless a conflicting directive is present, just like we do with `executionOrder="random"`.

That is, if there's already `failOnRisky="false"` then it should not be changed.

Another way would be to parse PHPUnit's output to see if there a risky tests, and give some feedback to the user adding a line or two to the mutation log. This looks like a support headache tbh.

maks-rafalko:
Fixes #1274

> Since PHPUnit won't treat new risky tests as failures by default, and since this behavior causes an unnecessary confusion even for experienced users (see #1273), we add failOnRisky="true" and failOnWarning="true" by default unless a conflicting directive is present, just like we do with executionOrder="random".

> That is, if there's already failOnRisky="false" then it should not be changed.

### DIFF
--- a/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
+++ b/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
@@ -100,6 +100,7 @@ class InitialConfigBuilder implements ConfigBuilder
 
         $this->addCoverageFilterWhitelistIfDoesNotExist($xPath);
         $this->addRandomTestsOrderAttributesIfNotSet($version, $xPath);
+        $this->addFailOnAttributesIfNotSet($version, $xPath);
         $this->configManipulator->replaceWithAbsolutePaths($xPath);
         $this->configManipulator->setStopOnFailure($xPath);
         $this->configManipulator->deactivateColours($xPath);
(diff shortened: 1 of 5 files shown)


src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
```diff
@@ -202,7 +203,7 @@ private function createNode(DOMDocument $dom, string $nodeName): DOMElement
 
     private function addRandomTestsOrderAttributesIfNotSet(string $version, SafeDOMXPath $xPath): void
     {
-        if (!version_compare($version, '7.2', '>=')) {
+        if (version_compare($version, '7.2', '<')) {
```
sanmai, line 206: It seems like this change is missing a test. Or there is?..

maks-rafalko: This method is covered by tests, I just simplified the condition:

https://github.com/infection/infection/blob/3aafcd43cec699620348cd5c40a197956308c37a/tests/phpunit/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilderTest.php#L410-L447

maks-rafalko:
Thank you, @sanmai.

maks-rafalko:
@sanmai it looks like I've implemented this feature for the **initial** tests run, while from what I understand we wanted this for the mutant phpunit execution, so that if Infection changes the code and tests became risky/with warning, then Mutant becomes killed.

What do you think about if I add the same to `phpunit.xml` for Mutant process? (`MutationConfigBuilder`)

Related to https://github.com/infection/infection/issues/1620

sanmai:
Yeah, looks like this is going to make sense.

Issue #1270, opened 2020-06-12
Decrement mutator makes array indexes negative


maks-rafalko:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.16
| Test Framework version | any
| PHP version | x.y.z (`php -v`)

This should not happen

```diff
10) /SourceClass.php:14    [M] DecrementInteger

--- Original
+++ New
@@ @@
         $c = [1, 2, 3];
-        $d = $c[0];
+        $d = $c[-1];
```

sanmai:
[Negative indexes in an array are not forbidden.](https://3v4l.org/GpRED) You can totally have an array with a negative index.

maks-rafalko:
right, but from my experience with this mutator, each time it decrements `0` to `-1` as an index of the array, it leads to undefined index notice

```
Notice: Undefined offset: -1
```

which in its turn always leads to killed Mutant.

So, I think this is a useless mutation in almost 100% cases. I'm speaking only about `0` to `-1`, not about `-99` to `-100`.

LoopKarma:
thank you for labelling issues - I will try to pick up this bug +)  
@maks-rafalko can you please explain why mutation should not be applied if `the number is an index of the array`?

maks-rafalko:
sure, as I said:

> from my experience with this mutator, each time it decrements `0` to `-1` as an index of the array, it leads to undefined index notice
> 
> ```
> Notice: Undefined offset: -1
> ```


So, there are many cases when I have similar code to this:


```php
$items = $object->getItems();

Assert::greaterThan(0, $items); // here we know that there is at least 1 item

$firstItem = $items[0];
```

So, when we have a list of items (array indexed from `0` to `N`) - there is no point in mutating `0` to `-1`, because it will always lead to 

```
Notice: Undefined offset: -1
```

PR #1281, opened 2020-07-07
DecrementInteger mutator should not decrement zero indexes for array access

LoopKarma:
This PR:
- [x] Covered by tests
Fixes https://github.com/infection/infection/issues/1270


src/Mutator/Number/DecrementInteger.php
```diff
@@ -98,6 +98,10 @@ private function isAllowedComparison(Node\Scalar\LNumber $node): bool
 
         $parentNode = ParentConnector::getParent($node);
 
+        if ($parentNode instanceof Node\Expr\ArrayDimFetch && $node->value === 0) {
```
maks-rafalko, line 101: I don't think this should be inside `isAllowedComparison()` method, because this is not related to `comparison`

Could you please move it e.g. before line 86 (`if ($this->isPartOfSizeComparison($node)) {`)

maks-rafalko:
I pushed a fix for e2e tests, now CI is happy.

Thank you for this contribution, @LoopKarma!

LoopKarma:
@maks-rafalko happy to help! Strange but I got errors in another e2e tests than Travis. How you came up that specifically this test should be improved?

maks-rafalko:
> How you came up that specifically this test should be improved?

I ran `tests/e2e_tests bin/infection` and analyzed the output. There was something about difference between expected and actual output

LoopKarma:
@maks-rafalko interesting, I ran the same command but got nothing for local php but got errors in other tests under docker setup. For sure I missed something here, since I'm pretty green here )

Issue #1231, opened 2020-03-30
Exclude certain lines or sections of code from MT not excluding them from coverage


sanmai:
**Is your feature request related to a problem? Please describe.**

As of now, the only way to stop Infection from mutating certain chunks of code without a config file is to use `@codeCoverageIgnoreStart` and related annotation. This will disable all and every mutation, where in may cases I've seen only some mutations are to be avoided.

**Describe the solution you'd like**

I'd like to be able to disable certain mutations for some parts of code on line or chunk basis, in a style similar to the above annotations. For example:

```php
/** @infection-disable GreaterThan */
while ($a > $b) {

}
```

**Describe alternatives you've considered**

It is possible to accomplish similar task [by using a configuration directive](https://infection.github.io/guide/how-to.html#Disable-in-particular-class-or-method-or-line) as of #79, but this is tedious and requires to update configuration every time a line moves up or down.

**Additional context**

- An additional visitor can tag nodes as non-mutable by certain or all mutators.
- Configuration blacklist check is a very hot path, working without it should be faster.

theofidry:
We would also probably need a finer control:

```php
// disable all on the next node
/** @infection-disable GreaterThan */
while ($a > $b) {
     // ...
     // won't be ignored
     while ($c > $d) {
          // ...
     }
}
```

```php
// disable all on the next node
/** @infection-disable GreaterThan */
while (
    $a > $b
    || $c > $d
    || $e > $f
) {
     // ...
}
```


```php
// disable only $c > $d
while (
    $a > $b
    /** @infection-disable GreaterThan */
    || $c > $d
    || $e > $f
) {
     // ...
}
```

```php
// Disable multiple mutators
/** @infection-disable GreaterThan, LessThan */
while ($a > $b) {

}
```

maks-rafalko:
For future contributor: we can see how this is implemented in Psalm (https://psalm.dev/docs/running_psalm/dealing_with_code_issues/#docblock-suppression)

While I hate polluting the code with tool-specific annotations *inside* methods, this is definitely more useful that adding an exception by particular line (`SomeClass::method:34`)

sanmai:
> We would also probably need a finer control

That'll be difficult, if not impossible. So far I propose to do it like Psalm does: on statement and method levels.

niconoe-:
This way to exclude some snippets for valid reasons is much better than using the configuration file, so thank you very much for this implementation.

[...]

Issue #697, opened 2019-06-07
Allow adding mutation exclusions for specific calls


NGTOne:
**Is your feature request related to a problem? Please describe.**
It's often the case that production codebases include support/monitoring/logging code that, if mutated, has little-to-no impact (or, alternatively, isn't worth testing the result of - e.g. calls to a logging function). If your codebase has lots of logging, this can generate large numbers of needless false positives and greatly slow down the mutation test run.

**Describe the solution you'd like**
Perhaps an addition to the `mutators` element in `infection.json`, along the lines of:
```json
{
    "mutators": {
        "ArrayItem": {
            "excludeCallsTo": [
                "Namespace\\Class::writeToLog",
                "OtherNamespace\\SomeOtherClass"
            ]
        }
    }
}
```

Obviously this wouldn't necessarily be 100% given PHP's extremely loose ways of calling functions (for instance, it might not catch something like `$level = "info"; Log::$level(...)`), but it would still save a lot of false positives over code like this.

**Describe alternatives you've considered**
Excluding every single log line, line-by-line, using the existing exclusion syntax. Doesn't scale well at all, and extremely brittle.

BackEndTea:
I'm assuming this would mostly be for the `FunctionCallRemoval` and `MethodCallRemoval` mutators?

I'd be okay with adding a filter, which works along the lines of how `ignore` currently works for excluding mutation of certain classes/namespaces/methods. e.g. use the [`fnmatch`](https://www.php.net/manual/en/function.fnmatch.php) syntax. (As that is used under the hood). We regex could also be an option, but i feel `fnmatch` is simpler, and good enough for what it is going to do.

NGTOne:
@BackEndTea I'm thinking a _bit_ more general - for instance, my use case also includes `ArrayElement` and `RemoveArrayElement` (our log lines are given a "context" array, to keep them easily `grep`able), both of which fall into the "it doesn't really affect anything if this breaks" category. And, even though I don't do it myself, I can imagine something like array-to-object casts falling into that bucket as well.

sanmai:
Infection itself would have used this for `Assert::` calls, I think.

And, yes, right now this can be done only on per-mutator basis, and not to say this is bad. Just a limitation.

theofidry:
@sanmai from what I understood, this appears to only work to ignore namespaces, classes or methods: not peculiar lines

sanmai:
[...]

PR #1285, opened 2020-07-28
Do not mutate clean up functions

exussum12:
Functions like Close and Free should not be removed by the function call
removals. Removing them does not change the tested behavior, but does
free the memory before the script finished / garbage collection is
invoked

This PR:

- [ ] Adds new feature ...
- [x] Covered by tests

### DIFF
--- a/src/Mutator/Removal/FunctionCallRemoval.php
+++ b/src/Mutator/Removal/FunctionCallRemoval.php
@@ -48,6 +48,18 @@ final class FunctionCallRemoval implements Mutator
 {
     use GetMutatorName;
 
+    /** @var string[] */
+    private $doNotRemoveFunctions = [
+        'assert',
+        'closedir',
+        'curl_close',
+        'curl_multi_close',
+        'fclose',
+        'mysqli_free_result',
+        'mysqli_close',
+        'socket_close',
+    ];
+
     public static function getDefinition(): ?Definition
     {
         return new Definition(
@@ -85,6 +97,6 @@ final class FunctionCallRemoval implements Mutator
             return true;
         }
 
-        return $name->toLowerString() !== 'assert';
+        return !in_array($name->toLowerString(), $this->doNotRemoveFunctions);
     }
 }
--- a/tests/phpunit/Mutator/Removal/FunctionCallRemovalTest.php
+++ b/tests/phpunit/Mutator/Removal/FunctionCallRemovalTest.php
@@ -165,13 +165,21 @@ $a = 3;
 PHP
         ];
 
-        yield 'It does not remove an assert() call' => [
+        yield 'It does not remove disallowed calls' => [
             <<<'PHP'
 <?php
 
 assert(true === true);
 aSsert(true === true);
 \assert(true === true);
+fclose($fileHandle);
+closedir($close);
+curl_close($curlHandle);
+fclose();
+mysqli_free_result();
+mysqli_close();
+socket_close();
+
 $a = 3;
 PHP
         ];


src/Mutator/Removal/FunctionCallRemoval.php
```diff
@@ -48,6 +48,18 @@ final class FunctionCallRemoval implements Mutator
 {
     use GetMutatorName;
 
+    /** @var string[] */
+    private $whitelisted = [
```
sanmai, line 52: whitelist probably is the name best to be avoided; say, it can be `doNotRemoveList` instead

src/Mutator/Removal/MethodCallRemoval.php
```diff
@@ -48,6 +48,13 @@ final class MethodCallRemoval implements Mutator
 {
     use GetMutatorName;
 
+    /** @var string[] */
+    private $whitelisted = [
+        'close',
+        'free',
+        'free_result',
```
sanmai, line 55: I'm not terribly excited about a special list for instance methods. I wish we could have this either configurable, or completed with the means supposed to exists under #1231

src/Mutator/Removal/FunctionCallRemoval.php
exussum12: Updated the names and the PR name

src/Mutator/Removal/MethodCallRemoval.php
exussum12: If it was configurable you would need this everywhere ? 
How about checking that all of these in the method versions take no argument ? eg `$someClass->free()` would be ignored but `$someClass->free($someVariable)` would be allowed to mutate ?

theofidry: I think doing so against native calls is ok but extending it to arbitrary classes is not: a user will likely be able to test that its `$someClass->free()` is called or not

tests/phpunit/Mutator/Removal/FunctionCallRemovalTest.php
```diff
@@ -165,13 +165,21 @@ public function mutationsProvider(): iterable
 PHP
         ];
 
-        yield 'It does not remove an assert() call' => [
+        yield 'It does not remove whitelisted calls' => [
```
sanmai, line 168:
```suggestion
        yield 'It does not remove disallowed calls' => [
```

src/Mutator/Removal/MethodCallRemoval.php
exussum12: They return void though ? https://www.php.net/manual/en/mysqli-result.free.php or https://www.php.net/manual/en/event.free could check if void return and 0 arguments and called one of these things ?

theofidry: and how do you tell the difference between `$mysqliResult->free()` and `$repository->free()`? We do not have type inference at the moment so we cannot distinguish the two and the later may deserve to be covered by infection

exussum12: I agree there will be methods missed by doing this, I don't have the stats but most methods I have seen called one of these are methods which have no real observable behaviour. 

Could have this as a user config option instead at project level instead of line level ?

sanmai: I don't think a user config option for this list is a bad idea.

maks-rafalko:
Since we have different opinions on object's methods calls, here is what I suggest:

* leave only native functions call ignore list in this PR and merge it
* think about method calls in a separate issue. But: I recomment to read this thread: https://github.com/infection/infection/issues/697. By implementing the feature that is being discussed there will cover your needs (method calls)

exussum12:
> Since we have different opinions on object's methods calls, here is what I suggest:

Sounds good to me, I have updated the PR to be just the native functions side of things here.

maks-rafalko:
Thank you, @exussum12!

lcobucci:
Does it make sense to add `openssl_free_key()`  to the list too?

maks-rafalko:
sounds good to me. Feel free to open a PR targeted to 0.17, so we will release it ASAP as 0.17.1

PR #1294, opened 2020-08-13
Refrain from adding filter tags if there are existing new-style coverage / include tags

### LINKED ISSUE #1283: PHPUnit 9.3 compatibility issue
| Question    | Answer
| Infection version | 0.16.x, master
| Test Framework version | PHPUnit 9.3
| PHP version | n/a
| Platform    | n/a

PHPUnit 9.3 comes [with a change of terminology and structure](https://github.com/sebastianbergmann/phpunit/blob/a0d6b21c6c8f6564212a1a14292d230ee35eba6d/ChangeLog-9.3.md#configuration-of-code-coverage-and-logging-in-phpunitxml), mainly going away from blacklist/whitelist, but not only.

Infection has to change provided `phpunit.xml` to run each and every mutation. So far I can tell [these changes are in the clear](https://github.com/infection/infection/blob/7ba6b0b0ef41fd80c0a3534af5cd15bd5d6e0955/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php#L107-L122). 
The problem is in [the changes Infection adding to collect coverage](https://github.com/infection/infection/blob/ddbcf37cb7437a428b8dd91ffa9ee9d2d4255969/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php#L112-L115) during the initial test. Specifically [it will add `<filter>` tag](https://github.com/infection/infection/blob/ddbcf37cb7437a428b8dd91ffa9ee9d2d4255969/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php#L149-L169) like below, even if there is PHPUnit 9.3 compatible `<coverage>` directive:

```diff
  <coverage>
    <include>
      <directory>path/to/src</directory>
    </include>
  </coverage>
+  <filter>
+    <whitelist>
+      <directory>path/to/src</directory>
+    </whitelist>
+  </filter>
```

Signs of legacy configuration:
 
```xml
 <phpunit disableCodeCoverageIgnore="...">
 <phpunit ignoreDeprecatedCodeUnitsFromCodeCoverage="...">
 <phpunit cacheTokens="...">
 <filter><whitelist>...
 <logging><log type="...">
 ```
 Signs of updated configuration:

 ```xml
 <coverage>
 <logging><*> <!-- where * isn't 'log' -->
```
### Possible solution

There are several consideration:

 - One way to tackle this issue is to check for `<coverage>` element before adding `<filter>` element. If it so happens that this `phpunit.xml` is being used with earlier version of PHPUnit that isn't aware of this new element, then this will be caught during validation against xsd.
- Additionally we have to deal with the case where there isn't `<coverage>` or `<filter>` elements. In this case we can check if `vendor/phpunit/phpunit/phpunit.xsd` mentions `<coverage>`  element, and then decide to add either, but to be honest I'd rather see Infection fail with an explanation. 
- Since Infection does not depend on PHPUnit, we are not guaranteed to find `phpunit.xsd`. This does not say that we cannot have a negative dependency, that is a future version of Infection [can be in conflict](https://github.com/infection/infection/blob/e1beb5ce19e7fef992ed8dbaa6ff78dfbda04570/composer.json#L68) with earlier versions of PHPUnit by standard means offered by `composer.json`.

### Alternative solution

As for failing with a message, Infection could check if there are `<coverage>` or `<filter>` elements, and if not, fail with an explanation instead of adding such elements.

### Workaround

Current workaround is to run PHPUnit to collect coverage before running Infection like so:
```bash
vendor/bin/phpunit --coverage-xml=build/logs/coverage-xml --log-junit=build/logs/junit.xml
vendor/bin/infection --skip-initial-tests --threads=$(nproc) --coverage=build/logs --show-mutations --no-interaction 
```
I'd like to think this is the right way to run Infection during CI, but there might be other opinions.

### Another workaround

If you do not plan to use PHPUnit older than 9.3, there's another workaround. One problem here is that Infection tries to validate updated `phpunit.xml`, but it uses newest schema where PHPUnit will be happy to use older one. When we instruct Infection to use older schema, it just works. Just like so:

```diff
-    xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
+    xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/schema/9.2.xsd"
```
This workaround requires the least amount of effort if you don't mind being nagged by PHPUnit about "--migrate-configuration"  now and then.

If you _do_ plan to use PHPUnit 9.2 or earlier together with 9.3, then you would probably need to inject something like this somewhere in your CI pipeline:
```bash
test -f vendor/phpunit/phpunit/schema/9.2.xsd || 
    mkdir -p vendor/phpunit/phpunit/schema && 
    cp vendor/phpunit/phpunit/phpunit.xsd vendor/phpunit/phpunit/schema/9.2.xsd
```

----

Thanks to @terabytesoftw for pointing at these changes.

sanmai:
Since current PHPUnit will accept old-style `filter` / `whitelist` tags with a warning, it is should be enough to have Infection running if we avoid adding old-style tags if there are existing new-style  `coverage` / `include` tags. 

This PR:

- [x] Part to #1283
- [x] Covered by tests


maks-rafalko:
Is it a temp solution just to release ASAP?

Are we going to fix it properly, adding `coverage` / `include` tags as we do it with `filter` / `whitelist` tags now, depending on the PHPUnit version?

sanmai:
Yes, we are going to fix this as properly as we can. I'll update the description. Sorry about the confusion.

lcobucci:
I think we have to do something similar to the `log` node, I still find issues when running infection with PHPUnit 9.3 (after applying this patch):

```
  The code coverage generated by the initial test run is invalid. Please report the issue on the                                                     
  infection repository "https://github.com/infection/infection".                                                                                     
                                                                                                                                                                                                                                                                                            
  $ '/home/lcobucci/projects/libs/clock/vendor/phpunit/phpunit/phpunit' '--configuration' '/tmp/infection/phpunitConfiguration.initial.infection.xml'
  PHPUnit 9.3.7 by Sebastian Bergmann and contributors.                                                                                              
                                                                                                                                                     
  Runtime:       PHP 7.4.9                                                                                                                           
  Configuration: /tmp/infection/phpunitConfiguration.initial.infection.xml                                                                           
  Random Seed:   1597823922                                                                                                                          
                                                                                                                                                     
    Warning - The configuration file did not pass validation!                                                                                        
    The following problems have been detected:                                                                                                       
                                                                                                                                                     
    Line 17:                                                                                                                                         
    - Element 'log': This element is not expected.                                                                                                   
                                                                                                                                                     
    Test results may not be as expected.                                                                                                             
                                                                                                                                                     
                                                                                                                                                     
  ....                                                                4 / 4 (100%)                                                                   
                                                                                                                                                     
  Time: 00:00.008, Memory: 4.00 MB                                                                                                                   
                                                                                                                                                     
  OK (4 tests, 10 assertions)                                                                                                                        
                                                                                                                                                     
                                                                                                                                                    
  Issue(s):                                                                                                                                          
  - The file "index.xml" could not be found: Could not find any "index.xml" file in "/tmp/infection"                                                 
```

sanmai:
Can you get a coverage report normally?

sebastianbergmann:
> I think we have to do something similar to the `log` node

Likely, as `<log>` was also [changed](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-9.3.md#configuration-of-code-coverage-and-logging-in-phpunitxml).

lcobucci:
@sanmai PHPUnit is behaving properly but, as @sebastianbergmann said, things have changed regarding that bit too (making the current code incompatible with the newer schema).

lcobucci:
A quick (a.k.a. **extremely hacky and, hopefully, temporary**) workaround would be relying on PHPUnit's config migration flag instead of providing a forward compatibility layer. I don't really like it but at least would make infection work without changing things drastically.

sanmai:
I don't like this `--migrate-configuration` in the slightest. 

@lcobucci Would you be minding to test another set of changes?

lcobucci:
> I don't like this --migrate-configuration in the slightest...

@sanmai 100% understandable. Happy to test, send it over!

sanmai:
I have no problem running these together:
```
$ php vendor/bin/infection --version
Infection - PHP Mutation Testing Framework version 0.17.0@201c8ea39a3748552bfadb109ca3bc41f3259cd2
$ php vendor/bin/phpunit --version
PHPUnit 9.3.7 by Sebastian Bergmann and contributors.
```
Is there actually an issue here?

sanmai:
@lcobucci 
Am I correct in understanding that you propose a user to run `phpunit --migrate-configuration` to update their configuration before running Infection?

lcobucci:
@sanmai I'm using the very same versions, here is [my PHPUnit config](https://github.com/lcobucci/clock/blob/master/phpunit.xml.dist).

This is the error I'm seeing: https://travis-ci.com/github/lcobucci/clock/jobs/375224476 (it's also weird that infection is exiting with `0` instead of an error)

lcobucci:
@sanmai my suggestion was to run on Infection's generated config file because that is the incorrect one.

sanmai:
It is indeed strange that Infection does not exit with an error status, but the fact that Infection does fail here isn't as strange. You have risky tests, and Infection configures PHPUnit to fail on them as per #1274. Try setting `failOnRisky="false"` in your phpunit.xml.dist.

lcobucci:
@sanmai the risky tests only appear when using the config file Infection generates:

```xml
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd" colors="false" verbose="true" beStrictAboutOutputDuringTests="true" beStrictAbout
TestsThatDoNotTestAnything="true" beStrictAboutTodoAnnotatedTests="true" beStrictAboutChangesToGlobalState="true" beStrictAboutCoversAnnotation="true" beStrictAboutResourceUsageDuringSmallTests="true" forceCover
sAnnotation="true" executionOrder="random" resolveDependencies="true" failOnRisky="true" failOnWarning="true" stopOnFailure="true" cacheResult="false" stderr="false">
  <php>
    <ini name="date.timezone" value="UTC"/>
  </php>
  <testsuites>
    <testsuite name="unit">
      <directory>/home/lcobucci/projects/libs/clock/test</directory>
    </testsuite>
  </testsuites>
  <coverage processUncoveredFiles="true">
    <include>
      <directory suffix=".php">/home/lcobucci/projects/libs/clock/src</directory>
    </include>
  </coverage>
  <filter>
    <whitelist>
      <directory>/home/lcobucci/projects/libs/clock/src</directory>
    </whitelist>
  </filter>
  <logging>
    <log type="coverage-xml" target="/tmp/infection/coverage-xml"/>
    <log type="junit" target="/tmp/infection/junit.xml"/>
  </logging>
</phpunit>
```

@sebastianbergmann is possible that having the old and new formats is confusing PHPUnit?

sanmai:
> risky tests only appear when using the config file Infection generates

That's exactly right. See #1274.

sebastianbergmann:
> @sebastianbergmann is it possible that having the old and new formats is confusing PHPUnit?

Everything is possible. If you think you found an issue in PHPUnit then please open a ticket with a minimal, self-contained, reproducing test case.

lcobucci:
@sanmai @sebastianbergmann alright, I found the cause for the risky tests.

Since my config uses the newer schema, Infection creates the `filter` with a different configuration (`processUncoveredFilesFromWhitelist="true"` isn't set).

After applying this patch, Infection fails with the message I sent earlier (coverage file not created because the config schema changes).

lcobucci:
@sanmai feel free to clone the repo and play with it, it might be easier for us to address this.

lcobucci:
@maks-rafalko @sanmai I've sent a PR to @sanmai's fork, extending this patch to make things compatible with PHPUnit 9.3 (still with a workaround, though).

That code makes things work 100% for me and shouldn't affect the behaviour when using older PHPUnit versions.

lcobucci:
Folks, is there anything I can do to help this moving forward?

sanmai:
1. I'm at the point where I stopped to understand the nature of the issue. I can run Infection with PHPUnit 9.3 just fine, but I also can't ignore that this is _not_ true for all cases. Which are they I do not know, yet.
2. This approach we're taking here with requesting coverage reports from the configuration opens a can of worms. Might be better to [follow the route suggested](https://github.com/infection/infection/issues/1283#issuecomment-670330697) by @sebastianbergmann. Steps to do that are not clear.

If you can explore on any of the points, this might clear things up.

lcobucci:
@sanmai alright, I agree with @sebastianbergmann's points. Will try to work on something :+1:

sanmai:
@maks-rafalko this should be good to go.

maks-rafalko (approved):

sanmai:
Thank you @maks-rafalko for the review.

Thank you @lcobucci for your efforts. We're almost there.

PR #1299, opened 2020-08-19
Infection must return appropriate exit status when failing

sanmai:
This PR fixes the unfortunate issue where Infection wouldn't exit with an error status in case of an error (e.g. low MSI).

Tests pending.

### DIFF
--- a/src/Command/BaseCommand.php
+++ b/src/Command/BaseCommand.php
@@ -75,7 +75,7 @@ abstract class BaseCommand extends Command
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        return $this->executeCommand($this->getIO()) ? self::SUCCESS : self::FAILURE;
+        return $this->executeCommand($this->getIO()) ? 0 : 1;
     }
 
     abstract protected function executeCommand(IO $io): bool;
--- a/src/Command/BaseCommand.php
+++ b/src/Command/BaseCommand.php
@@ -75,12 +75,10 @@ abstract class BaseCommand extends Command
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        $this->executeCommand($this->getIO());
-
-        return 0;
+        return $this->executeCommand($this->getIO()) ? self::SUCCESS : self::FAILURE;
     }
 
-    abstract protected function executeCommand(IO $io): void;
+    abstract protected function executeCommand(IO $io): bool;
 
     final protected function getIO(): IO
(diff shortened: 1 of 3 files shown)


maks-rafalko (approved):
can we target it to 0.17 branch?

sanmai:
Sure!

maks-rafalko:
> PHP Fatal error:  Uncaught Error: Undefined class constant 'SUCCESS' in /home/travis/build/infection/infection/src/Command/BaseCommand.php:78

one job has failed with such error, restarted (I'm not sure why only one ;) )

maks-rafalko:
still no luck https://travis-ci.com/github/infection/infection/jobs/375253541#L658

src/Command/BaseCommand.php
```diff
@@ -75,12 +75,10 @@ protected function initialize(InputInterface $input, OutputInterface $output): v
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        $this->executeCommand($this->getIO());
-
-        return 0;
+        return $this->executeCommand($this->getIO()) ? self::SUCCESS : self::FAILURE;
```
sanmai, line 78:
```suggestion
        return $this->executeCommand($this->getIO()) ? 0 : 1;
```

maks-rafalko:
Thank you, @sanmai

src/Command/RunCommand.php
```diff
@@ -213,7 +213,7 @@ protected function configure(): void
         ;
     }
 
-    protected function executeCommand(IO $io): void
+    protected function executeCommand(IO $io): bool
```
theofidry, line 216: how about keeping `void` here? In practice we tend to throw an exception in case of an error instead of a boolean value

sanmai: Do you propose that we re-throw the exceptions?

theofidry: We actually already do since we don't wrap the whole call in with a `catch (Throwable $t)` (and Symfony already does it). So I don't think forcing this return code is providing anything tbh

sanmai: Well, we do wrap the other call for two very important exceptions.

theofidry:
But to log the error and carry on right?

sanmai:
Well, not during CI. One expects that Infection will break the CI build if MSI is lower than required.

theofidry:
Fair enough

PR #1313, opened 2020-08-25
Allow fractional values for timeout

sanmai:
This PR:

- [x] Allows fractional values for timeout; say, half a second
- [x] Doc PR https://github.com/infection/site/pull/174


maks-rafalko (approved):
I would suggest adding 2 words in the docs, so that it's clear for users

PR #1319, opened 2020-08-31
Declare incompatbility with PHPUnit 9.3+

sanmai:
This PR:

- [x] Marks PHPUnit 9.3+ as non-installable together with Infection 0.17.x, with hopes that we'll sort these problems by 0.18.

Rationale is the following:
- PHPUnit 9.3 causes very nasty [interop problems](https://github.com/sebastianbergmann/phpunit/blob/a0d6b21c6c8f6564212a1a14292d230ee35eba6d/ChangeLog-9.3.md#configuration-of-code-coverage-and-logging-in-phpunitxml) where Infection may work or it may not, [and it is hard to understand what the problem is](https://github.com/ergebnis/playground/runs/1040078747#step:9:19). This is a major inconvenience for our users.
- All the while PHPUnit 9.3 has very little new apart above mentioned incompatibilities. [Only three or four new features](https://github.com/sebastianbergmann/phpunit/blob/a0d6b21c6c8f6564212a1a14292d230ee35eba6d/ChangeLog-9.3.md#added) and no critical issues resolved. 

It is unfortunate that we have to do this, but if we care about our users, and if we want to save their time debugging this problem and waiting for a solution that may not come very soon, I think this is the best course of action for the released branch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -67,6 +67,7 @@
     },
     "conflict": {
         "phpunit/php-code-coverage": ">9 <9.1.4",
+        "phpunit/phpunit": ">=9.3",
         "symfony/console": "=4.1.5"
     },
     "require-dev": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,20 +4,89 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "e87aa701107a3562b138043f91aa8eac",
+    "content-hash": "22372a1e32713547073a1dd3138fa473",
     "packages": [
         {
-            "name": "composer/xdebug-handler",
-            "version": "1.4.2",
+            "name": "composer/package-versions-deprecated",
+            "version": "1.11.99",
             "source": {
                 "type": "git",
-                "url": "https://github.com/composer/xdebug-handler.git",
-                "reference": "fa2aaf99e2087f013a14f7432c1cd2dd7d8f1f51"
+                "url": "https://github.com/composer/package-versions-deprecated.git",
+                "reference": "c8c9aa8a14cc3d3bec86d0a8c3fa52ea79936855"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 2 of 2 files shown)


maks-rafalko (approved):

maks-rafalko:
Absolutely agree. Thank you @sanmai

maks-rafalko:
Ported to master https://github.com/infection/infection/commit/34671ab5478b4d2d5967661f718aba446b3c14b7

had to recommit since cherry-pick / merge was impossible due to conflicts in lock file

sanmai:
Fair enough. Thank you!

Slamdunk:
Hi, in [`paratest`](https://github.com/paratestphp/paratest) we faced similar issues. Like infection, paratest needs deep knowledge about PHPUnit configuration, and many times we ended up with incompatibilities and bugs already fixed in PHPUnit XML handlers.
So in paratest I have decided for a steep turn: rely on PHPUnit classes to load and handle everything, even though they are marked as `@internal`:

https://github.com/paratestphp/paratest/blob/90b23bdddaa640837ea8af79ca522f72d02b96f2/src/Runners/PHPUnit/Options.php#L336-L338

Cons: we occasionally must pin exact PHPUnit version (although we still use caret operator to allow future versions), as something could break between patch versions.
Pros: no code, no maintenance, no bugs upon our shoulders.

I'm not telling you to do this as well; I just wanted to share the stress relieve of such solution.

theofidry:
I believe you should also add this change to the version checker so that this applies to the PHAR as well

rieschl:
Hi, is this really a problem with infection? PHPUnit 9.3 changed its configuration XML XSD, so users who use 9.3 should update their config. PHPUnit even notifies the user with a warning when an old config file was encountered and offers a migration tool.
Okay maybe the old config should have been marked as deprecated and not instantly be removed, but that's to be discussed in an issue in PHPUnit's issue tracker.

I run infection version `0.17.3` with PHPUnit `9.3.8` on multiple packages without any problems and no different behavior after updating.

Maybe we should reach out to @sebastianbergmann and the PHPUnit team and discuss re-adding the old entities into the XSD file? Or maybe infection could use an own internal phpunit.xsd to validate the schema?

sanmai:
@rieschl yes, you can use a different schema, see #1283

PR #1326, opened 2020-09-07
Add an ability to ignore Mutators for source code line(s) by regular expression

maks-rafalko:
- [x] Fixes https://github.com/infection/infection/issues/697
- [x] Covered by tests 
- [x] Doc PR: https://github.com/infection/site/pull/177

Allows to avoid mutating such lines of code as:

```diff
- Assert::notNull($variable);
```

or

```diff
- $this->logger->error($message, ['user' => $user]);
+ $this->logger->error($message, []);
```

and similar if developers are not interested in them.

Example of new config setting usage:

```json
{
    "mutators": {
        "MethodCallRemoval": {
            "ignoreSourceCodeByRegex": [
                "Assert::.*",
                "\\$this->logger.*"
            ]
        }
    }
}
```

Also, there is a new setting for `mutators` - `global-ignoreSourceCodeByRegex` that will be added to *all* enabled mutators if present. Works similar to [`global-ignore`](https://infection.github.io/guide/profiles.html#main)

```json
{
    "mutators": {
        "global-ignoreSourceCodeByRegex": [
            "Assert::.*"
        ]
    }
}
```

From an implementation point of view, it's not possible to use `IgnoreConfig` and `IgnoreMutator` to ignore mutators for configured lines of code by regex, because, at the step of applying Mutator, we don't have any information about the source line of code. 

While we technically can get it (using PrettyPrinter's [`prettyPrintExpr($expression)`](https://github.com/nikic/PHP-Parser/blob/f66a32e2df7866d593fa7a1bb582348f772511b0/lib/PhpParser/PrettyPrinterAbstract.php#L211) function or by passing source code to each mutator and finding the needed line), for me it seems like not the right solution.

This PR tries to filter out mutants before creating Mutant processes, similar to what is being discussed in `pre-`/`post-` hooks here https://github.com/infection/infection/issues/1323

So, before deciding whether we need to ignore mutator or no, we are applying regular expression (if any) to the diff of Mutant. Easy, yet so powerful.

I will continue adding tests/docs after we agree with the implementation.

I see this feature is most requested during the last months, so I would like to release it asap.

Regarding the suggested plugin system - https://github.com/infection/infection/issues/1323. When we will do it, this implementation can be redeveloped using plugin system/hooks, while for users configuration will be the same, so no BC here.

PS I'm aware of the failed build, it's ok for now since no tests have been updated

### DIFF
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -122,6 +122,12 @@
                         "type": "string"
                     }
                 },
+                "global-ignoreSourceCodeByRegex": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    }
+                },
 
                 "@arithmetic": { "$ref": "#/definitions/default-mutator-config" },
                 "@boolean": { "$ref": "#/definitions/default-mutator-config" },
@@ -502,6 +508,12 @@
                     "additionalProperties": false,
                     "properties": {
                         "ignore": {
+                            "type": "array",
+                            "items": {
+                                "type": "string"
+                            }
+                        },
+                        "ignoreSourceCodeByRegex": {
                             "type": "array",
                             "items": {
                                 "type": "string"
--- a/src/Configuration/ConfigurationFactory.php
+++ b/src/Configuration/ConfigurationFactory.php
@@ -37,7 +37,6 @@ namespace Infection\Configuration;
 
 use function array_fill_keys;
 use function array_key_exists;
-use function count;
 use function dirname;
 use Infection\Configuration\Entry\PhpUnit;
 use Infection\Configuration\Schema\SchemaConfiguration;
@@ -169,9 +168,9 @@ class ConfigurationFactory
      *
      * @return array<class-string<Mutator&ConfigurableMutator>, mixed[]>
      */
-    public function resolveMutators(array $schemaMutators, string $mutatorsInput): array
+    private function resolveMutators(array $schemaMutators, string $mutatorsInput): array
     {
-        if (count($schemaMutators) === 0) {
+        if ($schemaMutators === []) {
             $schemaMutators = ['@default' => true];
         }
 
--- a/src/Configuration/ConfigurationFactory.php
+++ b/src/Configuration/ConfigurationFactory.php
@@ -39,11 +39,12 @@ use function array_fill_keys;
 use function array_key_exists;
 use function count;
 use function dirname;
-use function explode;
 use Infection\Configuration\Entry\PhpUnit;
 use Infection\Configuration\Schema\SchemaConfiguration;
 use Infection\FileSystem\SourceFileCollector;
 use Infection\FileSystem\TmpDirProvider;
+use Infection\Mutator\ConfigurableMutator;
+use Infection\Mutator\Mutator;
 use Infection\Mutator\MutatorFactory;
 use Infection\Mutator\MutatorParser;
 use Infection\Mutator\MutatorResolver;
@@ -166,7 +167,7 @@ class ConfigurationFactory
     /**
      * @param array<string, mixed> $schemaMutators
      *
-     * @return array<string, mixed[]>
+     * @return array<class-string<Mutator&ConfigurableMutator>, mixed[]>
      */
     public function resolveMutators(array $schemaMutators, string $mutatorsInput): array
     {
@@ -273,10 +274,8 @@ class ConfigurationFactory
 
         foreach ($resolvedMutatorsMap as $mutatorClassName => $config) {
             if (array_key_exists('ignoreSourceCodeByRegex', $config)) {
-                $classNameParts = explode('\\', $mutatorClassName);
-                $mutatorName = end($classNameParts);
+                $mutatorName = MutatorFactory::getMutatorNameForClassName($mutatorClassName);
 
-                Assert::string($mutatorName);
                 Assert::isArray($config['ignoreSourceCodeByRegex']);
 
                 $map[$mutatorName] = $config['ignoreSourceCodeByRegex'];
--- a/src/Configuration/ConfigurationFactory.php
+++ b/src/Configuration/ConfigurationFactory.php
@@ -36,13 +36,14 @@ declare(strict_types=1);
 namespace Infection\Configuration;
 
 use function array_fill_keys;
+use function array_key_exists;
 use function count;
 use function dirname;
+use function explode;
 use Infection\Configuration\Entry\PhpUnit;
 use Infection\Configuration\Schema\SchemaConfiguration;
 use Infection\FileSystem\SourceFileCollector;
 use Infection\FileSystem\TmpDirProvider;
-use Infection\Mutator\Mutator;
 use Infection\Mutator\MutatorFactory;
 use Infection\Mutator\MutatorParser;
 use Infection\Mutator\MutatorResolver;
@@ -50,6 +51,7 @@ use Infection\TestFramework\TestFrameworkTypes;
 use OndraM\CiDetector\CiDetector;
 use function Safe\sprintf;
 use function sys_get_temp_dir;
+use Webmozart\Assert\Assert;
 use Webmozart\PathUtil\Path;
 
 /**
@@ -121,6 +123,11 @@ class ConfigurationFactory
             $namespacedTmpDir
         );
 
+        $resolvedMutatorsArray = $this->resolveMutators($schema->getMutators(), $mutatorsInput);
+
+        $mutators = $this->mutatorFactory->create($resolvedMutatorsArray);
+        $ignoreSourceCodeMutatorsMap = $this->retrieveIgnoreSourceCodeMutatorsMap($resolvedMutatorsArray);
+
         return new Configuration(
             $schema->getTimeout() ?? self::DEFAULT_TIMEOUT,
             $schema->getSource()->getDirectories(),
@@ -134,7 +141,7 @@ class ConfigurationFactory
             $logVerbosity,
             $namespacedTmpDir,
             $this->retrievePhpUnit($schema, $configDir),
-            $this->retrieveMutators($schema->getMutators(), $mutatorsInput),
+            $mutators,
             $testFramework,
             $schema->getBootstrap(),
             $initialTestsPhpOptions ?? $schema->getInitialTestsPhpOptions(),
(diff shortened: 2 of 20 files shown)


src/Process/Runner/MutationTestingRunner.php
```diff
@@ -103,6 +112,22 @@ public function run(iterable $mutations, string $testFrameworkExtraOptions): voi
...
+            ->filter(function (Mutant $mutant) {
+                $mutatorName = $mutant->getMutation()->getMutatorName();
+
+                if (!array_key_exists($mutatorName, $this->ignoreSourceCodeMutatorsMap)) {
+                    return true;
+                }
+
+
+                foreach ($this->ignoreSourceCodeMutatorsMap[$mutatorName] as $sourceCodeRegex) {
+                    if (preg_match("/-(.*?){$sourceCodeRegex}/", $mutant->getDiff()) === 1) {
```
sanmai, line 124: What's the catch here?

maks-rafalko: not really. Since we are applying reges on the diff, we want to apply it on the *originall* code, this is why we have a `-` on the begging.

in the e2e test of this PR, with the following config

```
"ignoreSourceCodeByRegex": [
    "getString"
]
```

we are ignoring such mutation:

```diff
- public function getString()
+ public function getString()
```

as you can see, this is not the beginning of the string

sanmai, line 124:
```suggestion
                    if (preg_match("/^-(.*?){$sourceCodeRegex}/", $mutant->getDiff()) === 1) {
```

Then it has to be like this, what do you think?

maks-rafalko: no, because `$mutant->getDiff()` returns:

```
--- Original
+++ New
@@ @@
         $this->getString();
         return 'hello';
     }
-    public function getString()
+    protected function getString()
     {
         return 'string';
     }
 }
```

so, with `^` it won't match

sanmai: Well, I'm more concerned about it matching a dash somewhere else:

https://3v4l.org/QvZ2G

Slamdunk: I agree with @sanmai : the current version is prone to error when dash is elsewhere.
I have the following suggestions:

1. use `^` with `m` modifier to only and exact match the original; the theoretical false positive of `--- Original` part of the diff, even if possible, is negligible
1. use `u` modifier, since UTF8 code is allowed
1. drop `(.*?)` as is prohibits exact matches to the user
1. add `$` for the same reason, to allow exact matches
1. be permissive only to indentations, so add `\s*` at start

All in all, the final form should be:

```suggestion
                    if (preg_match("/^-\s*{$sourceCodeRegex}$/mu", $mutant->getDiff()) === 1) {
```

So that user is allowed to choose between `getString` and `.*getString.*` regexes.

Slamdunk:
I'm very happy with the results. I miss only one functionality: a `global-ignoreSourceCodeByRegex`, like the already existing `global-ignore`, as many ignores may be specific to the environment of the mutation, instead of the specific mutation, like the `assert` one.

maks-rafalko:
@Slamdunk @sanmai great suggestions, thank you very much. Implemented.

Updated description. Also, added `global-ignoreSourceCodeByRegex` as you requested.

Ready for review.

src/Differ/DiffSourceCodeMatcher.php
```diff
@@ -0,0 +1,49 @@
...
+use function Safe\preg_match;
+
+/**
+ * @internal
+ */
+final class DiffSourceCodeMatcher
+{
+    public function matches(string $diff, string $sourceCodeRegex): bool
+    {
+        return preg_match("/^-\s*{$sourceCodeRegex}$/mu", $diff) === 1;
```
sanmai, line 47: `preg_quote` might be missing around here. Else we need to document the delimiter used.

Slamdunk: No quote! How can user provide regexes if we quote them?

Delimiter must be documented though, yes.

sanmai: Right. Shall we escape only the delimiter and save the time on explaining it?

Slamdunk: This can make sense

resources/schema.json
```diff
@@ -122,6 +122,12 @@
                         "type": "string"
                     }
                 },
+                "global-ignoreSourceCodeByRegex": {
```
sanmai, line 125: Why not call it just `ignoreSourceCodeByRegex` ?

maks-rafalko: to be consistent with the `global-ignore`

Slamdunk: To keep consistency with `ignore`/`global-ignore`?

src/Differ/DiffSourceCodeMatcher.php
sanmai:
```php
$sourceCodeRegex = str_replace('/', '\/', $sourceCodeRegex);
```

src/Configuration/ConfigurationFactory.php
```diff
@@ -255,4 +261,28 @@ private static function retrieveMinCoveredMsi(?float $minCoveredMsi, SchemaConfi
...
+     *
+     * @return array<string, array<int, string>>
+     */
+    private function retrieveIgnoreSourceCodeMutatorsMap(array $resolvedMutatorsMap): array
+    {
+        $map = [];
+
+        foreach ($resolvedMutatorsMap as $mutatorClassName => $config) {
+            if (array_key_exists('ignoreSourceCodeByRegex', $config)) {
+                $classNameParts = explode('\\', $mutatorClassName);
```
sanmai, line 276: There's a trait `GetMutatorName`, if `getName()` can be made static, and I think it is, then this could be greatly simplified by a call to `$mutatorClassName::getName()`

src/Differ/DiffSourceCodeMatcher.php
maks-rafalko: https://github.com/infection/infection/pull/1326/commits/bddf72fb00798efcbd9aefddf5ee02d16fd8d103

src/Configuration/ConfigurationFactory.php
maks-rafalko: > Make `Mutator::getName()` non-static #989 

🤣 

I'm personally ok with 2 duplicated lines here

sanmai: Well, one can totally do just that: https://3v4l.org/XNklO

sanmai: #1331

tests/e2e/Exclude_Mutations_By_Regex/infection.json
```diff
@@ -0,0 +1,31 @@
...
+            "ignore": [
+                "App\\*"
+            ],
+            "ignoreSourceCodeByRegex": [
+                ".*getString.*"
+            ]
+        },
+        "MethodCallRemoval": {
+            "ignoreSourceCodeByRegex": [
+                "Assert::.*",
```
sanmai, line 26: Shouldn't we dog-food this feature already?

sanmai (approved):

tests/e2e/Exclude_Mutations_By_Regex/infection.json
Slamdunk: As a non-native English speaker, may I ask you what does this mean?

sanmai: Sure. I'm slightly surprised to find that there's even [a Wikipedia article for the term](https://github.com/symfony/symfony/pull/38191).

> Eating your own dog food or dogfooding is the practice of an organization using its own product.

src/Configuration/ConfigurationFactory.php
maks-rafalko: updated

src/Configuration/ConfigurationFactory.php
```diff
@@ -151,10 +159,33 @@ public function create(
...
+            $ignoreSourceCodeMutatorsMap
         );
     }
 
+    /**
+     * @param array<string, mixed> $schemaMutators
+     *
+     * @return array<class-string<Mutator&ConfigurableMutator>, mixed[]>
+     */
+    public function resolveMutators(array $schemaMutators, string $mutatorsInput): array
```
sanmai, line 172:
```suggestion
    private function resolveMutators(array $schemaMutators, string $mutatorsInput): array
```
I don't see it used anywhere else.

sanmai (approved):

sanmai:
If we can extend #1327 to allow injecting arbitrary configuration for a plugin, this feature is a great candidate for a standard plugin. Nevertheless I think this idea shouldn't be holding anything up since it should be possible to refactor this ignorer into a plugin without changing the configuration schema in a bit.

src/Configuration/ConfigurationFactory.php
maks-rafalko: thanks, fixed

src/Configuration/ConfigurationFactory.php
```diff
@@ -151,10 +159,33 @@ public function create(
...
+    {
+        if (count($schemaMutators) === 0) {
```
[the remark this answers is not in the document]
maks-rafalko: fixed

sanmai:
CS Fixer is going nuts :disappointed:

tests/e2e/Exclude_Mutations_By_Regex/infection.json
maks-rafalko: > Shouldn't we dog-food this feature already?

if we need to - we certainly can. I didn't do any analysis for this though

sanmai: It can certainly wait.

maks-rafalko:
> CS Fixer is going nuts 😞

yeah, it's because it's downloading v2.15.8 instead of v2.16.4 as the latest PHAR for "https://cs.sensiolabs.org/download/php-cs-fixer-v2.phar" URL

2.16.4 works as expected locally. I will try to resolve it as a separate issue somewhere later today

maks-rafalko:
> If we can extend #1327 to allow injecting arbitrary configuration for a plugin, this feature is a great candidate for a standard plugin.

Agree. That was my plan - to release this one ASAP, and then, when we are ready with the plugin system (which can take some time) - we will refactor internal implementation.

maks-rafalko:
Regarding the php-cs-fixer: there is an open issue, let's see how they will resolve it: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5099

I think if they wont' fix it today, we can think about downloading particular hardcoded version for the time being

sanmai:
I did `composer require --dev friendsofphp/php-cs-fixer` for local CS works.

sanmai:
Shall we merge this?

maks-rafalko:
Thank you @sanmai & @Slamdunk for great review

PR #1331, opened 2020-09-15
Dedicated static function to parse out mutator name

sanmai:
This PR:

- [x] Adds a dedicated static function to parse out mutator name
- [x] Covered by tests

### DIFF
--- a/src/Mutator/MutatorFactory.php
+++ b/src/Mutator/MutatorFactory.php
@@ -92,4 +92,11 @@ final class MutatorFactory
 
         return $mutators;
     }
+
+    public static function getMutatorNameForClassName(string $className): string
+    {
+        $parts = explode('\\', $className);
+
+        return (string) end($parts);
+    }
 }
--- a/src/Mutator/MutatorFactory.php
+++ b/src/Mutator/MutatorFactory.php
@@ -35,9 +35,10 @@ declare(strict_types=1);
 
 namespace Infection\Mutator;
 
-use function in_array;
+use function end;
+use function explode;
+use function is_a;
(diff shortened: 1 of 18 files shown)


src/Mutator/MutatorFactory.php
```diff
@@ -92,4 +92,11 @@ public function create(array $resolvedMutators): array
 
         return $mutators;
     }
+
+    public static function getMutatorNameForClassName(string $className): string
+    {
+        $parts = explode('\\', $className);
+
+        return (string) end($parts);
+    }
```
sanmai, line 101: Closest place where `Mutator::getName()` is used.

maks-rafalko (approved):

theofidry:
Looks a bit weird to me that it's the factory responsible for giving the mutator name (by default)

sanmai:
It is. Should be easy to refactor. What do you have in mind?

theofidry:
What was wrong with how it was before? I saw there was a bunch of PRs so maybe I missed the overall picture

sanmai:
It was only an instance method before (still is), but sometimes you want to get a name for a mutator having only a class name for it.

theofidry:
Do you have a place where this is the case? I think it would help to understand what's going on because indeed it's a bit tricky

sanmai:
Sure, [right about here](https://github.com/infection/infection/commit/ee40e434a0b3ec8385e3d1d5591c2013b72a7720#diff-f20466d79712e5e751a8819581cc4787R270-R285).

theofidry:
Replied there

PR #1340, opened 2020-10-17
Drop support for PHP 7.3

maks-rafalko:
php_roadmap

Active support for PHP 7.3 is finished on 6 Dec 2020. Here is the first of several PRs to drop support for PHP 7.3 in Infection.

Infection 0.18, which I will release soon, will be the last Infection version with PHP 7.3 support.

Nexts PRs:

- Update the docs (https://github.com/infection/site/pull/178)
- Run Rector to upgrade to PHP 7.4 features (typed properties / short functions, etc.)

PR #1341, opened 2020-10-17
Switch to use ocramius/package-versions

sanmai:
Use of compatibility package is no longer needed on PHP 7.4+

Follow-up to #1340 

(This will break builds on PHP 7.3)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,7 @@
         "infection/include-interceptor": "^0.2.4",
         "justinrainbow/json-schema": "^5.2",
         "nikic/php-parser": "^4.9 <4.10",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -52,7 +52,7 @@
         "infection/include-interceptor": "^0.2.4",
         "justinrainbow/json-schema": "^5.2",
         "nikic/php-parser": "^4.9 <4.10",
-        "composer/package-versions-deprecated": "^1.2 || ^2.0",
+        "ocramius/package-versions": "^1.2 || ^2.0",
```
maks-rafalko, line 55: I believe `composer.lock` must be updated

maks-rafalko (approved):

PR #1344, opened 2020-10-18
Add autoreview workflow

sanmai:
This PR:

- [x] Adds GitHub Workflow with a set of autoreview tests and checks

### DIFF
--- a/.github/workflows/autoreview.yaml
+++ b/.github/workflows/autoreview.yaml
@@ -1,0 +1,49 @@
+# yamllint disable rule:line-length
+# yamllint disable rule:braces
+
+name: Autoreview
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - master
+
+jobs:
+  tests:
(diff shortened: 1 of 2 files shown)


maks-rafalko (approved):
Should we remove it from TravisCI then

sanmai:
That's the idea, but that needs slightly more consideration. 

Scrutinizer is getting annoying. Now it uses PHP 7.3, and then it uses PHP 7.4. Can we get it sorted? I don't think I have the knobs to do it.

maks-rafalko:
> Scrutinizer is getting annoying

I have to admit I didn't use Scrutinizer report for ages (years). It doesn't prevent from merging and we don't use any of the complaints from it (I suppose)

Do we really need it? :|

sanmai:
TBH I hardly remember any actionable feedback from it... Remove it then?

maks-rafalko:
ok, removed

Issue #1345, opened 2020-10-18
Don't mutant $limit 0, -1 in preg_split 


vjik:
In function [preg_split](https://www.php.net/manual/en/function.preg-split.php) argument `limit` with values 0 and -1 behaves the same way. It would be cool if there was no mutation 0 to -1 and -1 to 0.

maks-rafalko:
Hello, totally agree that this shouldn't be mutated.

Feel free to help us and fix it, shouldn't be hard. Mutator that does it is:

https://github.com/infection/infection/blob/81099171f3e43fc600e73f515e80858a49d240fb/src/Mutator/Number/DecrementInteger.php#L48

maks-rafalko:
Fixed in https://github.com/infection/infection/pull/1347

PR #1347, opened 2020-10-18
Don't mutate $limit argument from 0 to -1 and from -1 to 0 in `preg_split` function

sidz:
This PR stops mutate $limit argument from 0 to -1 and from -1 to 0 in `preg_split` function.

Requested in https://github.com/infection/infection/issues/1345

### DIFF
--- a/tests/phpunit/Mutator/Number/DecrementIntegerTest.php
+++ b/tests/phpunit/Mutator/Number/DecrementIntegerTest.php
@@ -470,5 +470,41 @@ PHP
 $b = $a[0];
 PHP
         ];
+
+        yield 'It does not decrement limit argument of preg_split function when it equals to 0' => [
+            <<<'PHP'
+<?php
+
+preg_split('//', 'string', 0);
+PHP
+        ];
+
+        yield 'It does decrement limit argument of preg_split function when it greater than 0' => [
+            <<<'PHP'
+<?php
+
+preg_split('//', 'string', 1);
+PHP
+            ,
+            <<<'PHP'
+<?php
+
+preg_split('//', 'string', 0);
+PHP
+        ];
+
+        yield 'It does decrement limit argument of preg_split function when it equal to -1' => [
+            <<<'PHP'
+<?php
+
+preg_split('//', 'string', -1);
+PHP
+            ,
+            <<<'PHP'
+<?php
+
+preg_split('//', 'string', -2);
+PHP
+        ];
     }
 }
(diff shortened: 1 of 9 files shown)


tests/phpunit/Mutator/Number/DecrementIntegerTest.php
```diff
@@ -192,211 +192,10 @@ public function mutationsProvider(): iterable
             ,
         ];
 
-        yield 'It does not decrement zero when it is compared as greater than count() on the right side' => [
+        yield 'It does not decrement limit argument of preg_split function when it equals to 0' => [
             <<<'PHP'
 <?php
-
```
maks-rafalko, line 198: were these tests removed by mistake? If so, please revert

maks-rafalko (approved):

vjik:
It is wrong to completely deny mutations.
There must be mutations more 0 and less -1.

sidz:
@vjik added more tests to show that it works.

But found an issue with Decrement mutator when
`preg_split('//', 'string', 1);`

https://github.com/infection/infection/blob/5682c2e186fe1fb4ad5569e031717e1f31d35796/src/Mutator/Number/DecrementInteger.php#L82

and similar with Increment mutator when:
`preg_split('//', 'string', 0);`

https://github.com/infection/infection/blob/f8cf148e95a24bd69320e272c65bbbf5e46bd32d/src/Mutator/Number/IncrementInteger.php#L72

vjik:
See https://github.com/infection/infection/pull/1350

IncrementInteger work fine without fixes. Problems with DecrementInteger and OneZeroInteger.

sidz:
@vjik `IncrementInteger` doesn't work - https://infection-php.dev/r/1lr (proof)

vjik:
> @vjik `IncrementInteger` doesn't work - https://infection-php.dev/r/1lr (proof)

IncrementInteger don't work for `preg_split('//', 'string', -1);` => `preg_split('//', 'string', 0);` — that is right. 
In other cases IncrementInteger it works — thas is also right.

maks-rafalko (approved):
👌

sidz:
@vjik I suppose all possible cases were covered. PR is ready for review.

> IncrementInteger don't work for preg_split('//', 'string', -1); => preg_split('//', 'string', 0); — that is right.

It didn't work right as mutators had an issue with negative values mutations. This behavior was fixed in https://github.com/infection/infection/pull/1351. So current master will mutate values from -1 to 0.

sanmai (approved):
Tests are sensible.

maks-rafalko:
Thank you @sidz :+1:

PR #1349, opened 2020-10-18
Use alpine CLI instead of ubuntu. Significantly decreases image size.

sidz:
Use alpine CLI instead of ubuntu in the devTools. Significantly decreases image size.

Additional improvements:
- Add PHP 8 container.
- Switch from Ubuntu to Alpine
- Drop `git` as dependency
- switch to `--prefer-dist` instead of `--prefer-source`

p.s. Does anybody use these containers?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,8 @@ INFECTION=./build/infection.phar
 DOCKER_RUN=docker run --tty --rm --volume "$$PWD":/opt/infection --workdir /opt/infection
 DOCKER_RUN_74=$(FLOCK) devTools/*php74*.json $(DOCKER_RUN) infection_php74
 DOCKER_RUN_74_IMAGE=devTools/Dockerfile-php74-xdebug.json
+DOCKER_RUN_80=$(FLOCK) devTools/*php80*.json $(DOCKER_RUN) infection_php80
+DOCKER_RUN_80_IMAGE=devTools/Dockerfile-php80-xdebug.json
 
 FLOCK=./devTools/flock
 COMMIT_HASH=$(shell git rev-parse --short HEAD)
@@ -106,11 +108,15 @@ test-unit: $(PHPUNIT)
 
 .PHONY: test-unit-docker
 test-unit-docker:	## Runs the unit tests on the different Docker platforms
-test-unit-docker: test-unit-74-docker
+test-unit-docker: test-unit-80-docker
 
 .PHONY: test-unit-74-docker
 test-unit-74-docker: $(DOCKER_RUN_74_IMAGE) $(PHPUNIT)
 	$(DOCKER_RUN_74) $(PHPUNIT) --group default
+
+.PHONY: test-unit-80-docker
+test-unit-80-docker: $(DOCKER_RUN_80_IMAGE) $(PHPUNIT)
+	$(DOCKER_RUN_80) $(PHPUNIT) --group default
 
 .PHONY: test-e2e
 test-e2e: 	 	## Runs the end-to-end tests on the different Docker platforms
@@ -126,24 +132,34 @@ test-e2e-docker: 	## Runs the end-to-end tests on the different Docker platforms
 test-e2e-docker: test-e2e-phpdbg-docker test-e2e-xdebug-docker
 
 .PHONY: test-e2e-phpdbg-docker
-test-e2e-phpdbg-docker: test-e2e-phpdbg-74-docker
+test-e2e-phpdbg-docker: test-e2e-phpdbg-74-docker test-e2e-phpdbg-80-docker
 
 .PHONY: test-e2e-phpdbg-74-docker
 test-e2e-phpdbg-74-docker: $(DOCKER_RUN_74_IMAGE) $(INFECTION)
 	$(DOCKER_RUN_74) $(PHPUNIT) --group integration,e2e
 	$(DOCKER_RUN_74) env PHPDBG=1 ./tests/e2e_tests $(INFECTION)
 
+.PHONY: test-e2e-phpdbg-80-docker
+test-e2e-phpdbg-80-docker: $(DOCKER_RUN_80_IMAGE) $(INFECTION)
+	$(DOCKER_RUN_80) $(PHPUNIT) --group integration,e2e
+	$(DOCKER_RUN_80) env PHPDBG=1 ./tests/e2e_tests $(INFECTION)
+
 .PHONY: test-e2e-xdebug-docker
-test-e2e-xdebug-docker: test-e2e-xdebug-74-docker
+test-e2e-xdebug-docker: test-e2e-xdebug-74-docker test-e2e-xdebug-80-docker
 
 .PHONY: test-e2e-xdebug-74-docker
 test-e2e-xdebug-74-docker: $(DOCKER_RUN_74_IMAGE) $(INFECTION)
 	$(DOCKER_RUN_74) $(PHPUNIT) --group integration,e2e
 	$(DOCKER_RUN_74) ./tests/e2e_tests $(INFECTION)
 
+.PHONY: test-e2e-xdebug-80-docker
+test-e2e-xdebug-80-docker: $(DOCKER_RUN_80_IMAGE) $(INFECTION)
+	$(DOCKER_RUN_80) $(PHPUNIT) --group integration,e2e
+	$(DOCKER_RUN_80) ./tests/e2e_tests $(INFECTION)
+
 .PHONY: test-infection
 test-infection:		## Runs Infection against itself
-test-infection:
+test-infection:MakefileTes
 	$(INFECTION) --threads=4
 
 .PHONY: test-infection-docker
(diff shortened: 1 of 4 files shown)


maks-rafalko (approved):
> p.s. Does anybody use these containers?

Me personally - sometimes

sanmai (commented):
I don't think resulting container size matters much (not only because it uses a reusable layer, but primarily because this isn't something anyone should deploy), but increased complexity is not very cool (e.g. `apk` mumbo-jumbo instead of simple `apt-get`)

sanmai:
Lots of stuff use Ubuntu containers, therefore if one uses any of the Ubuntu-based containers, for them effect from this change will be insubstantial. But please correct me if I'm wrong.

sidz:
> but increased complexity is not very cool (e.g. apk mumbo-jumbo instead of simple apt-get)

to be honest I do not see any complexity overhead between `apk` and `apt`. Difference only which letters are you typing to install anything `apt install` vs `apk add`. IMO it sounds like holy war between `yum/dnf` and `apt`.

> Lots of stuff use Ubuntu containers, therefore if one uses any of the Ubuntu-based containers, for them effect from this change will be insubstantial. 

Api-platform and Sylius based on alpine even for development purposes for example.

> But please correct me if I'm wrong.

You are right. There is no other benefits except final size.

Also I'd like to save some space (no I do not want to by new SSD/HDD/whatever :P ) because I have tons of containers and images on my local machine.

sanmai:
Container layers are reusable. If anything you touch, just anything, uses Ubuntu containers, then there will be no space savings.

```
$docker system df -v
Images space usage:

REPOSITORY                        TAG                     IMAGE ID            CREATED       (wraps below)
infection_php74                   latest                  1a3d42ceee18        6 months ago        

SIZE                SHARED SIZE         UNIQUE SIZE         CONTAINERS
445.6MB             405MB               40.55MB             0
```

The image takes only 40 MB extra above what the original image takes. With Alpine Linux you have to install daily necessities like bash and git, which, with relevant libs, might take a lot more than these mere 40 megs.

sidz:
> If anything you touch, just anything, uses Ubuntu containers, then there will be no space savings.
I see your point you just trying to say that you are using images based on Ubuntu but you can't say that everybody use it as base layer. Personally I'm not and probably @maks-rafalko as well.

> but not with increased complexity of the script, needed to setup the image.
I still do not see where is complexity sorry :)

p.s. I'm ok to close this PR if you still do not want to see it in master

sanmai:
I'd like to have a second opinion in the light on the new arguments present. I might too much biased toward simplicity.

What the space savings by the way?

sidz:
> What are the space savings by the way?

if we are talking about `infection_php74` image only - not so big. Around 12MB.

But if we are talking about `infection_php74` + base image (that is what I originally meant when said "significantly") - smth around 335MB.

```
$ docker system df -v                                           
Images space usage:

REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE                SHARED SIZE         UNIQUE SIZE         CONTAINERS
infection_php74     latest              20f03eb620fe        45 seconds ago      110.7MB             82.4MB              28.27MB             0
infection_php74     ubuntu              a358a7df6e0f        4 minutes ago       445.2MB             405.1MB             40.11MB             0
php                 7.4-cli             e4c48aebbce3        6 days ago          405.1MB             405.1MB             0B                  0
composer            latest              ea5fa9a44267        2 weeks ago         176.4MB             82.4MB              93.97MB             0
php                 7.4-cli-alpine      9e1753f1efbe        2 weeks ago         82.4MB              82.4MB              0B                  0
```

> Also I wonder if there's a pre-built Alpine-based image with Xdebug? 

never seen it before. Maybe unofficial only. Not an option IMO.

maks-rafalko:
Guys, up to you to decide what to use here, ubuntu or alpine based.

I don't see any issues with alpine as I've been using alpine images for 2 years. But my voice shouldn't be counted as I use docker for Infection not very often. Use whatever is convenient for you :)

sanmai:
This PR adds a whole bunch of commands just to save some hundred megs on a base image (which is reusable), where `apt-get` would simply work. With all honesty I don't think it's all worth it. 

Let me reiterate. It's not about Alpine, its about saving time just on the off-chance someone would need to change this Dockerfile for whatever reason: `apt-get` something is much easier to change than this `apk add` mumbo-jumbo.

sidz:
> It's not about Alpine, its about saving time just on the off-chance someone would need to change this Dockerfile for whatever reason: `apt-get` something is much easier to change than this `scanelf --needed --nobanner` mumbo-jumbo.

1. I've never seen that anybody touch this Dockerfile in Infection (the same as dockerfile for php7.3) once it created.
2. You don't need to touch `scanlf` at all as long as you use Alpine. There is no reason to touch this part.
3. And the last but not least: If anybody doesn't know what to do - google it. 

What if I don't have a clue what is `apt-get` because I came from FreeBSD or Gentoo world? :) IMO simplicity is not an argument here.

sanmai:
> If anybody doesn't know what to do - google it.

Great! Do you imply that I should use my _very_ sparse free time googling about `scanelf` instead of working on, say, #1283? I don't feel like this will be the best use of it tbh.

sanmai (changes requested):
> I've never seen that anybody touch this Dockerfile in Infection (the same as dockerfile for php7.3) once it created.

It's a good tradition IMO. Let's keep it that way, shall we?

sidz:
> Great! Do you imply that I should use my very sparse free time googling about scanelf flags and options instead of working on, say, #1283? 

You may continue with https://github.com/infection/infection/issues/1283. I did everything for you here.

> It's a good tradition IMO. Let's keep it that way, shall we?

It was in the meaning of that you do not need to care about it.

sanmai:
You can say what you want but words are just not enough here. The problem here is in upcoming PHP 8.0 release. If you could add a Dockerfile for it (say, using `php:8.0-rc-alpine`) with xdebug and stuff, then I can _really, really,_ forget about ever having to deal with `scanelf` flags and options. And then I'll be perfectly fine with the use of Alpine here. Make sense?

sidz:
> The problem here is in upcoming PHP 8.0 release. If you could add a Dockerfile for it (say, using php:8.0-rc-alpine) with Xdebug and stuff, then I can really, really, forget about ever having to deal with scanelf flags and options. And then I'll be perfectly fine with the use of Alpine here. Make sense?

Done.

Makefile
```diff
@@ -126,44 +132,61 @@ test-e2e-docker: 	## Runs the end-to-end tests on the different Docker platforms
...
 
+.PHONY: test-e2e-xdebug-80-docker
+test-e2e-xdebug-80-docker: $(DOCKER_RUN_80_IMAGE) $(INFECTION)
+	$(DOCKER_RUN_80) $(PHPUNIT) --group integration,e2e
+	$(DOCKER_RUN_80) ./tests/e2e_tests $(INFECTION)
+
 .PHONY: test-infection
 test-infection:		## Runs Infection against itself
-test-infection:
+test-infection:MakefileTes
```
sanmai, line 162: Something is off here.

devTools/Dockerfile-php80-xdebug
```diff
@@ -0,0 +1,33 @@
+reFROM php:8.0-rc-cli-alpine
```
sanmai, line 1: reFROM?

devTools/Dockerfile-php80-xdebug
```diff
@@ -0,0 +1,33 @@
...
+        zip
+
+ARG XDEBUG_VERSION=3.0.0beta1
+
+RUN set -eux; \
+    apk add --no-cache --virtual .build-deps \
+        ${PHPIZE_DEPS} \
+        build-base \
+    ; \
+    mkdir -p /usr/src/php/ext/xdebug && curl -fsSL https://pecl.php.net/get/xdebug-${XDEBUG_VERSION} | tar -xvz -C "/usr/src/php/ext/xdebug" --strip 1; \
```
sidz, line 15: since php8 `pecl` and `pear` are not available anymore: https://externals.io/message/103977

sanmai: Yeah, that's one reason of my concerns.

sidz:
```
$ docker system df -v
Images space usage:

REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE                SHARED SIZE         UNIQUE SIZE         CONTAINERS
infection_php80     latest              72543361cc72        9 minutes ago       91.35MB             80.72MB             10.63MB             0
```

devTools/Dockerfile-php80-xdebug
sidz: there is an option to use https://packagist.org/packages/friendsofphp/pickle instead but I do not think that this is good idea.

sanmai: That's correct.

sanmai (approved):

sanmai:
@sidz Great work!

Fun fact: running test takes one-third of a time less on PHP 8 than on PHP 7.4.

sanmai:
Merging now [to debug this issue](https://github.com/infection/infection/runs/1295793063#step:12:28).

PR #1351, opened 2020-10-18
Change DecrementInteger and IncrementInteger mutators behaviour

sidz:
Found an interesting issues while worked with https://github.com/infection/infection/pull/1347 that:

- DecrementInteger mutator generates wrong mutation with negative values: -1 was mutated to -0 instead of -2
- IncrementIngeger mutator generates wrong mutation with negative values: -1 was mutated to -2 instead of 0

This PR fixes behavior for both mutators and now DecrementInteger will return -2 for case when original value is -1 and IncrementIngeger returns 0 for case when original value is -1

### DIFF
--- a/src/Mutator/Number/DecrementInteger.php
+++ b/src/Mutator/Number/DecrementInteger.php
@@ -74,7 +74,15 @@ final class DecrementInteger extends AbstractNumberMutator
      */
     public function mutate(Node $node): iterable
     {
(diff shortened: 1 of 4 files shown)


sanmai (approved):

maks-rafalko (approved):

maks-rafalko:
👍 Thanks @sidz

PR #1356, opened 2020-10-19
Add more tests for DecrementInteger, IncrementInteger and OneZeroInteger mutators

sidz:
I've decided to add more tests for DecrementInteger, IncrementInteger and OneZeroInteger mutators as I have a strong feeling that I may break current behavior in https://github.com/infection/infection/pull/1347.

The idea is to cover all possible scenarios in current master and before I'll start to do changes in https://github.com/infection/infection/pull/1347 in order to show how exactly these mutators currently do their job to avoid possible questions in those PR.

### DIFF
--- a/tests/phpunit/Mutator/Number/DecrementIntegerTest.php
+++ b/tests/phpunit/Mutator/Number/DecrementIntegerTest.php
@@ -62,15 +62,6 @@ PHP
             ,
         ];
 
-        yield 'It does not decrement the number one' => [
-            <<<'PHP'
-<?php
-
-$a = 1;
-PHP
-            ,
-        ];
-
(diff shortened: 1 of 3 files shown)


tests/phpunit/Mutator/Number/DecrementIntegerTest.php
```diff
@@ -62,15 +62,6 @@ public function mutationsProvider(): iterable
             ,
         ];
 
-        yield 'It does not decrement the number one' => [
```
sidz, line 65: duplicates https://github.com/infection/infection/blob/master/tests/phpunit/Mutator/Number/DecrementIntegerTest.php#L337-L345

maks-rafalko (approved):

PR #1362, opened 2020-10-22
Use GitHub Actions to test on Windows

sanmai:
The number one reason to do this is the extreme slowness of AppVeyor builds. They take almost forever to queue, and as long to execute one after another. That's a free service so one can't really complain, but for GH workflows there's no such issue. And it's a free service too.

This PR:

- [x] Uses GitHub Actions to test on Windows
- [x] ~~And on macOS, for good measure~~
- [x] And with both Xdebug and PCOV

FWIW PHPUnit takes the same approach, [though in a much more elaborate way](https://github.com/sebastianbergmann/phpunit/blob/b2a9914fd2057a72ec65bc9b3cd963c831a1b14b/.github/workflows/ci.yml#L88-L118).

### DIFF
--- a/.github/workflows/autoreview.yaml
+++ b/.github/workflows/autoreview.yaml
@@ -30,14 +30,19 @@ jobs:
           php-version: ${{ matrix.php-version }}
           tools: composer:v2
 
+      - name: Get composer cache directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
       - name: Cache dependencies
         uses: actions/cache@v2
         with:
-          path: ~/.cache/composer
-          key: composer-${{ matrix.php-version }}-${{ hashFiles('**/composer.*') }}
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('**/composer.*') }}-${{ matrix.dependencies }}
           restore-keys: |
-            composer-${{ matrix.php-version }}-${{ hashFiles('**/composer.*') }}-
-            composer-${{ matrix.php-version }}-
+            composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('**/composer.*') }}-
+            composer-${{ runner.os }}-${{ matrix.php-version }}-
+            composer-${{ runner.os }}-
             composer-
 
       - name: Install dependencies
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -57,7 +57,7 @@ jobs:
         run: |
           composer update --no-interaction --prefer-dist --no-progress ${{ matrix.dependencies }}
 
-      - name: Run tests
+      - name: Run tests using Makefile
         if: runner.os != 'Windows'
         run: |
           make test-unit
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -62,6 +62,11 @@ jobs:
         run: |
           make test-unit
 
+      - name: Run tests without Makefile
+        if: runner.os == 'Windows'
+        run: |
+          vendor/bin/phpunit --group default
+
       - name: Run Infection
         run: |
           php bin/infection -j2 --test-framework-options="--group=default" --ignore-msi-with-no-mutations
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -67,4 +67,4 @@ jobs:
 
       - name: Run Infection
         run: |
-          ./bin/infection -j$(nproc) --test-framework-options="--group=default" --only-covered --ignore-msi-with-no-mutations
+          ./bin/infection -j2 --test-framework-options="--group=default" --only-covered --ignore-msi-with-no-mutations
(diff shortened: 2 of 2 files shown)


.github/workflows/autoreview.yaml
```diff
@@ -30,14 +30,19 @@ jobs:
           php-version: ${{ matrix.php-version }}
           tools: composer:v2
 
+      - name: Get composer cache directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
```
sanmai, line 35: `~/.cache/composer` is Linux/macOS specific. This different approach accounts for Windows too.

.github/workflows/ci.yaml
```diff
@@ -33,27 +35,33 @@ jobs:
...
 
       - name: Install dependencies
         run: |
           composer update --no-interaction --prefer-dist --no-progress ${{ matrix.dependencies }}
 
-      - name: Run tests
+      - name: Run tests using Makefile
+        if: runner.os != 'Windows'
         run: |
           make test-unit
```
sanmai, line 63: Running PHPUnit on Windows is super annoying. It's either a .BAT, or some other BS I don't think it's worth dealing right now. More importantly Infection itself knows how to work around this, therefore this step is actually redundant. But let's keep it for now, not when under Windows.

.github/workflows/ci.yaml
```diff
@@ -33,27 +35,33 @@ jobs:
...
 
       - name: Run Infection
         run: |
-          ./bin/infection -j$(nproc) --test-framework-options="--group=default"
+          php bin/infection -j2 --test-framework-options="--group=default" --only-covered --ignore-msi-with-no-mutations
```
sanmai, line 67: These last two bits are from AppVeyor config.

.github/workflows/autoreview.yaml
```diff
@@ -30,14 +30,19 @@ jobs:
...
-          path: ~/.cache/composer
-          key: composer-${{ matrix.php-version }}-${{ hashFiles('**/composer.*') }}
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('**/composer.*') }}-${{ matrix.dependencies }}
           restore-keys: |
-            composer-${{ matrix.php-version }}-${{ hashFiles('**/composer.*') }}-
-            composer-${{ matrix.php-version }}-
+            composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('**/composer.*') }}-
+            composer-${{ runner.os }}-${{ matrix.php-version }}-
+            composer-${{ runner.os }}-
```
sanmai, line 45: This section had to be amended to match ci.yaml to improve cache hit ratio.

sanmai (commented):
I think we can pretty much depose AppVeyor at this point.

.github/workflows/ci.yaml
maks-rafalko: At the end of all the changes, does it mean we won't run unit tests on Windows? If so - why? currently we run them on AppVeyor and I would say we should continue doing that.

.github/workflows/ci.yaml
```diff
@@ -12,18 +12,20 @@ on:
 
 jobs:
   tests:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.operating-system }}
 
     strategy:
       matrix:
+        operating-system: [ubuntu-latest, windows-latest, macos-latest]
```
maks-rafalko, line 19: not sure we need macos, but the only argument I have is the increased time of the build. Is there anything we get from running the build on macos?

up to you, minor for me

.github/workflows/ci.yaml
maks-rafalko: do we really want to run with `--only-covered` for all the matrix combination?

.github/workflows/ci.yaml
sanmai: IDK, let's remove it.

.github/workflows/ci.yaml
sanmai: Not at all! On the contrary, Infection will run them as it does. The initial test run, remember?

.github/workflows/ci.yaml
sanmai: Best guess it was to speed up the build, but let's remove it.

.github/workflows/ci.yaml
sanmai: Here: https://github.com/infection/infection/runs/1290189378#step:8:15

sanmai:
One possible next step would be to build a .phar, and then use it to run E2E tests on all supported platform. I'm not so sure this will be possible with PHP 8 right now, but worth a try anyway.

.github/workflows/ci.yaml
sanmai: The only real downside here is that we won't see skipped-incomplete tests in the log for Windows builds. Think it's a blocker?

sanmai: Right, we're running tests on Windows too now.

maks-rafalko: sorry for the delay, good point with the initial tests run. But since you made an update to run tests on windows, let it be so ;)

maks-rafalko (approved):
Cool!

PR #1365, opened 2020-10-22
Remove AppVeyor, run E2E tests during CI

sanmai:
This PR:

- [x] Runs integration and end-to-end tests on Windows using GitHub Actions.
- [x] Adds a separate mutation testing workflow, with only PHP 7.4/PCOV and Linux/Windows
- [x] Disables AppVeyor (will report success all the time)
- [x] Runs full set of E2E tests on Ubuntu, without a .phar yet.

The one bit I cannot do: protected branch config needs new Windows checks to set required. Also, IIRC version branches are not protected, they might need some love too. 

There's no need to drop a requirement for AppVeyor: it is still used by earlier branches.


.github/workflows/ci.yaml
```diff
@@ -57,16 +59,48 @@ jobs:
...
-          vendor/bin/phpunit --group default
+          ls tests/e2e/*/composer.json | xargs dirname |
+             xargs -I{} composer --working-dir={} update --no-interaction --prefer-dist --no-progress ${{ matrix.dependencies }}
 
-      - name: Run Infection
+      - name: Run a subset of E2E tests
+        if: runner.os == 'Windows'
+        shell: bash
+        run: |
+          make test-e2e-phpunit PHP='phpdbg -qrr' E2E_PHPUNIT_GROUP=e2e BENCHMARK_SOURCES=
```
sanmai, line 94: It just wouldn't work with Xdebug. [It seems installed, but then it is not.](https://github.com/infection/infection/pull/1365/checks?check_run_id=1297037605#step:11:1)

.github/workflows/autoreview.yaml
```diff
@@ -38,9 +38,9 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ${{ steps.composer-cache.outputs.dir }}
-          key: composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('**/composer.*') }}-${{ matrix.dependencies }}
+          key: composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('composer.*') }}-${{ matrix.dependencies }}
           restore-keys: |
-            composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('**/composer.*') }}-
+            composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('composer.*') }}-
```
sanmai, line 43: We only need to consider two files here, no others in `tests/e2e` or otherwise.

.github/workflows/ci.yaml
```diff
@@ -57,16 +59,48 @@ jobs:
...
         if: runner.os == 'Windows'
+        uses: actions/cache@v2
+        with:
+          path: tests/e2e/*/vendor
+          key: e2e-vendor-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('tests/e2e/*/composer.json') }}-${{ matrix.dependencies }}
+          restore-keys: |
+            e2e-vendor-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('tests/e2e/*/composer.json') }}-
+            e2e-vendor-${{ runner.os }}-${{ matrix.php-version }}-
+            e2e-vendor-${{ runner.os }}-
+            composer-
```
sanmai, line 82: Installing anything with Composer is insanely slow in Windows, just compare [these 4 seconds on Linux](https://github.com/infection/infection/pull/1365/checks?check_run_id=1296609944#step:6:1) with [these 34 seconds on Windows](https://github.com/infection/infection/pull/1365/checks?check_run_id=1296609961#step:6:1). The solution is to cache the dependencies, but not only archives, the whole `vendor` folders. Note that we don't cache `composer.lock` here: it takes far less time to verify that all is in order.

This might be a mistake, but let's see.

Makefile
```diff
@@ -39,6 +40,8 @@ BENCHMARK_SOURCES=tests/benchmark/MutationGenerator/sources \
 				  tests/benchmark/Tracing/coverage \
 				  tests/benchmark/Tracing/sources
 
+E2E_PHPUNIT_GROUP=integration,e2e
+PHPUNIT_GROUP=default
```
sanmai, line 44: To let us override these from outside.

tests/e2e/SymfonyFlex/run_tests.bash
```diff
@@ -0,0 +1,23 @@
...
+
+    if [ "$DRIVER" = "phpdbg" ]
+    then
+        phpdbg $PHPARGS -qrr $INFECTION
+    else
+        php $PHPARGS $INFECTION
+    fi
+
+    diff -u -w expected-output.txt infection.log
+}
```
sanmai, line 23: SymfonyFlex needs a special environment variable, and it uses old version of PHPUnit which is not compatible with the one we use, so we can't really run this test from `E2ETest`. Adding this file prevents this test from running from  `E2ETest`, and also solves env variable issue.

tests/phpunit/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilderTest.php
```diff
@@ -98,6 +98,10 @@ public function test_it_builds_and_dump_the_xml_configuration(): void
 
     public function test_it_preserves_white_spaces_and_formatting(): void
     {
+        if (PHP_EOL === "\r\n") {
+            $this->markTestSkipped('Test fixture uses Unix line endings');
+        }
+
```
sanmai, line 104: We haven't run this test on Windows before, apparently.

appveyor.yml
```diff
@@ -1,55 +1,2 @@
...
-    - ps: .ci\appveyor_install_xdebug.ps1
-
-test_script:
-    - cd c:\projects\workspace
-    - vendor\bin\phpunit --prepend=tests/xdebug-filter.php --coverage-clover=clover.xml --coverage-xml=coverage/coverage-xml --log-junit=coverage/junit.xml
-    - php bin\infection --threads=%NUMBER_OF_PROCESSORS% --log-verbosity=none --coverage=coverage --no-progress --only-covered --ignore-msi-with-no-mutations
-
-matrix:
-  fast_finish: true
+clone_depth: 1
```
maks-rafalko, line 2: why do we need to keep this file?

tests/phpunit/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilderTest.php
```diff
@@ -98,6 +98,10 @@ public function test_it_builds_and_dump_the_xml_configuration(): void
 
     public function test_it_preserves_white_spaces_and_formatting(): void
     {
+        if (PHP_EOL === "\r\n") {
```
maks-rafalko, line 101: minor:

we already have:

```php
if (PHP_OS_FAMILY === 'Windows') {
        $this->markTestSkipped('Not interested in profiling on Windows');
}
```
and 

```php
if ('\\' === DIRECTORY_SEPARATOR) {
```

what do you think about using one of those methods to detect windows and not introduce the 3rd one?

maks-rafalko (approved):

maks-rafalko:
> The one bit I cannot do: protected branch config needs new Windows checks to set required. Also, IIRC version branches are not protected, they might need some love too.

let's merge and I will set it up

tests/phpunit/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilderTest.php
sanmai: This is not to detect Windows, but to test specifically against `PHP_EOL`. Windows build under WSL might have proper line endings, for example. Makes sense?

appveyor.yml
sanmai: AppVeyor will fail the build without it.

maks-rafalko: so will we remove once 0.18 is not maintained anymore?

tests/phpunit/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilderTest.php
maks-rafalko: yep, thanks

appveyor.yml
sanmai: Yes, that's the idea. Remove it once we don't need it anywhere.

Issue #1322, opened 2020-09-03
checkstyle output log format


sanmai:
It looks like this:

```
<?xml version="1.0" encoding="ISO-8859-15"?>
<checkstyle version="6.5">
  <file name="src/Example.php">
    <error line="172" severity="error" message="Error message" source="MutatorType"/>
    <error line="218" severity="error" message="Another error message" source="AnotherMutator"/>
  </file>
</checkstyle>

```
[XSD for the format.](https://github.com/K-Phoen/PHP-CS-Fixer/blob/dd96fc45c0b44df9236f1dbdad9815a8865c2b72/doc/checkstyle.xsd)

Then with help of [cs2pr](https://github.com/staabm/annotate-pull-request-from-checkstyle) we can [annotate a PR with comments](https://github.com/staabm/annotate-pull-request-from-checkstyle#annotate-a-pull-request-based-on-a-checkstyle-xml-report) for each escaped mutation. [Format for these annotations.](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-a-warning-message)

With [Phan](https://github.com/phan/phan/commit/91a56adc15dfd3c65a1ad2b4b9ba5ec480d9a1bd#diff-706ee59d555b2e31d5e8ee4f842ba5fbR35) people use it like this: 
```
vendor/bin/phan --no-progress-bar --output-mode checkstyle | cs2pr --graceful-warnings --colorize
```

We can use it by running Infection normally, then feeding the report to `cs2pr` from a file.

PR #1368, opened 2020-10-22
Add github logger to be able to use Annotations on GitHub Actions

maks-rafalko:
This PR:

- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/183

Fixes https://github.com/infection/infection/issues/1322

### Description

This feature allows adding errors/warnings for particular commits/PRs - GitHub Annotations

Basically, this feature adds escaped mutants information (diff) to particular lines where modified source code wasn't detected by tests:

checkstyle

### Things to note

* Unfortunately, GitHub Annotations does not allow using syntax highlighting inside annotations, so we can't display a pretty diff, as well as using `<details>` tag to be able to show/hide huge text. 
* Currently, I've added a separate GH Action to run Infection just for the added files in order to display warnings.
* Displayed warnings does not fail the build

### Goal

This will greatly improve review process as it will be immediately clear what parts of the code are not covered by tests properly (thus escaped mutants)

Suggestions are appreciated.


sanmai:
Just a quick note that there’s hardly any reason why Infection can’t output GitHub compatible annotations straight away.

sanmai:
I mean we can skip this XML and cs2pr altogether by outputting the annotations right from Infection. This will also solve the problem I’ll mention in the review comments in a moment.

.github/github-functions.sh
```diff
@@ -0,0 +1,24 @@
...
+
+    if ! [[ "${PR_NUMBER}" == "" ]]; then
+        git remote set-branches --add origin "$GITHUB_HEAD_REF";
+        git fetch;
+
+        changed_files=$(git diff master --diff-filter=A --name-only | grep src/ | paste -sd "," -);
+
+        if [ -n "$changed_files" ]; then
+            # Set those flags only if there is any changed files detected
+            flags="--filter=${changed_files} --ignore-msi-with-no-mutations --only-covered ${flags}";
```
sanmai, line 18:
```suggestion
            flags="--filter=${changed_files} --ignore-msi-with-no-mutations --min-msi=100.0 --min-covered-msi=100.0 --only-covered ${flags}";
```

This report is primarily important in case of CI failure. But if we make Infection fail the CI, guess what, the next command with `cs2pr` won't run!

infection.json.dist
```diff
@@ -9,7 +9,8 @@
         "text": "infection.log",
         "badge": {
             "branch": "master"
-        }
+        },
+        "checkstyle": "build/checkstyle.xml"
```
sanmai, line 13: Just like it so it could be: 
```json
"github": "/dev/stdout" // or /dev/stderr, whichever works best
```
And the it could output the annotations directly: 
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-a-warning-message

.github/github-functions.sh
maks-rafalko: no worries, I like your idea ;) it makes sense

but I'm not sure about `--min-msi=100.0 --min-covered-msi=100.0` in the comment above. Do we really want to fail the build even with 1 escaped mutant?

well, we can try...

infection.json.dist
sanmai: If you look in the source of cs2pr, these annotations are pretty damn simple:

https://github.com/staabm/annotate-pull-request-from-checkstyle/blob/79e8cb48fb729d31684d02fe856dbf2282b65d6f/cs2pr#L108-L137

It is not to say we should scrap this new checkstyle report, but it won't exactly work here.

.github/github-functions.sh
sanmai: Surely not, but for the sake of an example! The whole schema should work even if Infection fails.

infection.json.dist
sanmai: I think it can be even better with "github" progress reporter.

maks-rafalko: > I think it can be even better with "github" progress reporter.

what is it?

.github/workflows/mutation-testing-annotations.yaml
```diff
@@ -0,0 +1,57 @@
...
+# yamllint disable rule:braces
+
+name: Infection
+
+on:
+    pull_request:
+    push:
+        branches:
+            - main
+            - master
```
sanmai, line 11: This workflow for PRs only so this `push:` section can go. This will also simplify `get_infection_pr_flags()`

.github/workflows/mutation-testing-annotations.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v2
+
+            - name: Setup PHP
+              uses: shivammathur/setup-php@v2
+              with:
+                  php-version: ${{ matrix.php-version }}
+                  tools: composer:v2
```
sanmai, line 31: `coverage: pcov`?

.github/workflows/mutation-testing-annotations.yaml
sanmai: Frankly there will be only one line left, so it'll be as good as it is inlined.

.github/workflows/mutation-testing-annotations.yaml
```diff
@@ -0,0 +1,57 @@
...
+            - name: Install dependencies
+              run: |
+                  composer update --no-interaction --prefer-dist --no-progress
+
+            - name: Run Infection for added files only
+              run: |
+                  source .github/github-functions.sh;
+                  INFECTION_FLAGS=$(get_infection_pr_flags);
+                  echo "INFECTION FLAGS: ${INFECTION_FLAGS}";
+                  php bin/infection -j2 $INFECTION_FLAGS;
```
sanmai, line 57: Assuming this is a PR-only workflow:
```suggestion
                  php bin/infection -j2 --filter=$(git diff origin/"${GITHUB_BASE_REF:-master}" --diff-filter=A --name-only | grep src/ | paste -sd "," -) --ignore-msi-with-no-mutations --only-covered 
```

Alternatively, this can be put in a separate step, and then this step could depend on `if:` previous step had anything.

sanmai:
If I understood correctly, PHPStan enables this reporter automatically (?) on GitHub CI.

Their implementation: https://github.com/phpstan/phpstan-src/blob/c99b7f8c09fe6e319624ef3b0b66f43d30056fe4/src/Command/ErrorFormatter/GithubErrorFormatter.php

maks-rafalko:
> If I understood correctly, PHPStan enables this reporter automatically (?) on GitHub CI.

seems so, but I'm not sure it's a good choice TBH. Without proper filtering (as we do it right now), this will overload the diff with thousands of warnings by default (which MT can produce with even not very big projects)

infection.json.dist
sanmai: Like ProgressFormatter or DotFormatter. Set with `--formatter` option.

sanmai:
> Without proper filtering (as we do it right now), this will overload the diff with thousands of warnings by default

If Infection can detect it is running on GitHub (provided it was asked to do that), then it can fetch and run against changed files just as well.

sanmai:
For example, specifying `--formatter=github` might trigger all this new machinery with changed files etc.

.github/workflows/mutation-testing-annotations.yaml
```diff
@@ -0,0 +1,52 @@
...
+                      composer-
+
+            - name: Install dependencies
+              run: |
+                  composer update --no-interaction --prefer-dist --no-progress
+
+            - name: Run Infection for added files only
+              run: |
+                  git fetch;
+                  php bin/infection -j2 --filter=$(git diff origin/"${GITHUB_BASE_REF:-master}" --diff-filter=A --name-only | grep src/ | paste -sd "," -) --ignore-msi-with-no-mutations --only-covered;
```
sanmai, line 52:
```suggestion
                  php bin/infection -j2 --filter=$(git diff origin/$GITHUB_BASE_REF --diff-filter=A --name-only | grep src/ | paste -sd "," -) --ignore-msi-with-no-mutations --only-covered;
```

.github/workflows/mutation-testing-annotations.yaml
```diff
@@ -0,0 +1,52 @@
...
+                      composer-${{ runner.os }}-
+                      composer-
+
+            - name: Install dependencies
+              run: |
+                  composer update --no-interaction --prefer-dist --no-progress
+
+            - name: Run Infection for added files only
+              run: |
+                  git fetch;
```
sanmai, line 51:
```suggestion
                  git fetch --depth=1 origin $GITHUB_BASE_REF
```

maks-rafalko:
> For example, specifying `--formatter=github` might trigger all this new machinery with changed files etc.

my opinion:

1. `formatter` and `logger` - two different concepts. Formatter is for displaying the MT process, during the run (on each step). Logger - doing smth with the calculated results after MT is done. So here we are dealing with `logger`. We don't want to display GH Annotation on each step.

2.  I don't like idea of making Infection too opionated here. Why should it decide for developer what files to filter? Why only `A`dded files but not `M`odified? Why not both? Why both? And so on.

So, the current state of PR IMO is ideal - we give a developer new logger that can display GH Annotations to a file / `stdout`. Developer decides what files to mutate in order to use it - all the files in the project, just added new files, just modified or a combination.

What I think we should add (but in a separate PR), is an ability to set logger by command line option, like `--logger-checkstyle=php://stdout`, or any other of supported loggers

Why? To be able to *not* use checkstyle report for all the Infection runs in all currently available GH Actions, but to specify it only for `Infection / Mutation Testing Code Review Annotations` one

sanmai:
> Why? To be able to not use checkstyle report for all the Infection runs in all currently available GH Actions, but to specify it only for Infection / Mutation Testing Code Review Annotations one

That's what I was thinking when I proposed `--formatter=github`. This will probably work just as well.

sanmai:
> 2\. I don't like an idea of making Infection too opionated here. Why should it decide for developer what files to filter? Why only `A`dded files but not `M`odified? Why not both? Why both? And so on.

Infection is already very opinionated in a similar way. For example, by choosing to add a whitelist (or what's it called now) to `phpunit.xml` if there is none. Making Infection collect all the necessary details for a user if the run it with `--formatter=github` or `--logger-github` is a step in the same direction.

Imagine you're writing a blog post. What would be better to write, that Infection will now automatically mutate only changed files if run with `--logger-github`, or that Infection can do that only if you're brave enough to sprinkle it with a bunch of hardly readable `git`/`paste` magic?

(I'd take it if you say that Infection shouldn't add `<whitelist>` if there's none, that's a major blocker for our issue with PHPUnit 9.3.)

maks-rafalko:
> What would be better to write, that Infection will now automatically mutate only changed files if run with `--logger-github`

1. how would we do it from PHP code? (filter out changed/added files only)?
2. I still don't like the idea to decide what files to leave after filtering: `A`dded, `M`odified, both. How to decide? Here in Infection we removed `M`odified because of too many failed builds due to the low MSI in the old files. But why should we do the same for all the projects? Then, we need to add settings to enable / disable `M` files? Doesn't sound good

sanmai:
1. `shell_exec('git diff origin/$GITHUB_BASE_REF --diff-filter=AM --name-only')` and take it from there.
2.  It can be `--logger-github=AM` with a default set to `A`

maks-rafalko:
Sounds good. I will give it a try, thanks!

.github/workflows/mutation-testing-annotations.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+            - name: Get composer cache directory
+              id: composer-cache
+              run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+            - name: Cache dependencies
+              uses: actions/cache@v2
+              with:
+                  path: ${{ steps.composer-cache.outputs.dir }}
+                  key: composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('**/composer.*') }}-${{ matrix.dependencies }}
```
sanmai, line 38:
```suggestion
                  key: composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('composer.*') }}-${{ matrix.dependencies }}
```

.github/workflows/mutation-testing-annotations.yaml
```diff
@@ -0,0 +1,52 @@
...
+                  restore-keys: |
+                      composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('**/composer.*') }}-
```
sanmai, line 40:
```suggestion
                      composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('composer.*') }}-
```
To match #1365

maks-rafalko:
Ready for the next round of review.

Now it's possible to use `--log-verbosity=AM` or 

```json
{
    "logs": {
        "github": "A"
    }
}
```

resources/schema.json
```diff
@@ -73,6 +73,10 @@
                             "description": "Mutation score badge for your GitHub project."
                         }
                     }
+                },
+                "github": {
+                    "type": "string",
+                    "description": "GitHub Annotations for escaped Mutants in the added/modifies files"
```
sanmai, line 79: Is this relevant?

src/Logger/GitHub/GitDiffFileProvider.php
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * @final
+ *
+ * @internal
+ */
+class GitDiffFileProvider
+{
+    public function provide(string $diffFilter): string
+    {
+        return (string) shell_exec(sprintf('git diff origin/"${GITHUB_BASE_REF:-master}" --diff-filter=%s --name-only | grep src/ | paste -sd ","', $diffFilter));
```
sanmai, line 49: - `GITHUB_BASE_REF` is always present on GitHub. If not, it should fail rather than substitute it with `master`/`main` (and which btw?).
- Missing `escapeshellarg()`

sanmai (commented):
After a couple of nights sleeping on this idea I came to a realization these are two separate options:
- One set is to select files for mutation by GitHub/CI diff, `--git-diff-filter=A --git-diff-base=master`.
- The other is to output notices in GitHub format, `--log-github`

It is doubtful  `--logger-github` is sensible without the first two, but first two a quite possibly usable outside of GitHub, with `--git-diff-base` defaulting to `$GITHUB_BASE_REF` or whatever it is on other CI services.

E.g. we can make Infection mutate only changed files on Travis CI, without logging anything in a GitHub format.

maks-rafalko:
> What do you think?

It makes sense.  The only one downside of the proposed solution is the number of new options (three), which is not good. And one can only be used with another.

But they still make sense.

So, ok - let's wait 1 day for any other comments and I will implement it, just don't want to rewrite it for the 4th time.

resources/schema.json
maks-rafalko: yes, why not? 

> Now it's possible to use `--logger-github=AM` or
> 
> ```json
> {
>     "logs": {
>         "github": "A"
>     }
> }
> ```

---

hm, now I'm thinking how we will use `--logger-github` option. Without the value? So that if it's used withut `--git-diff-filter=A` - all the files will be mutated and too many annotations will be created. I don't say it's bad, but it's the fact.

sanmai:
> just don't want to rewrite it for the 4th time

I feel your pain. That's why I waited for two days before proposing this change.

resources/schema.json
sanmai: How about let them be? Say, one might have a small project with only minimal number of remaining mutations, and they want to fix them all. Having Infection report all and every will be a huge help.

theofidry:
@maks-rafalko I don't think i'll be able to review it tonight or it is gonna be much later in the evening

sanmai:
Not sure how's you're planning to working with this, but how do you think if we split this PR into two? 

These are essentially two related but distinct features. They'll be easier to review one by one.

maks-rafalko:
To be honest I would like not to do this. Because fixing all the tests that work with `Configuration`, `SchemaConfiguration` and `Container` takes 90% of the time spent on this PR.

BTW, this is something to think about - how to make tests not so painful. Any new parameter added to `Container::withValues()` brings so many changes in the tests which are very hard to understand, look at this:

https://github.com/infection/infection/blob/85091deff6a7415b5552cadfa2b484c04a046b95/tests/phpunit/Configuration/ConfigurationFactoryTest.php#L199-L257

If I find mental helth to spend this time again fixing those test, I will split it, but IMO it's not so big to review.

sanmai:
> BTW, this is something to think about - how to make tests not so painful. Any new parameter added to `Container::withValues()` brings so many changes in the tests which are very hard to understand, look at this:

I had to do this twice already, and I did not like it. At all.

src/Configuration/ConfigurationFactory.php
```diff
@@ -65,27 +67,30 @@ class ConfigurationFactory
      */
     private const DEFAULT_TIMEOUT = 10;
 
-    private $tmpDirProvider;
-    private $mutatorResolver;
-    private $mutatorFactory;
-    private $mutatorParser;
-    private $sourceFileCollector;
-    private $ciDetector;
+    private TmpDirProvider $tmpDirProvider;
```
theofidry, line 70: I think there might be a phpStorm helper to apply that everywhere (I'm saying that as a note not to do it here in this PR)

src/Logger/GitHub/GitDiffFileProvider.php
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @internal
+ */
+class GitDiffFileProvider
+{
+    public const DEFAULT_BASE = 'origin/master';
+
+    public function provide(string $gitDiffFilter, string $gitDiffBase): string
+    {
+        return (string) shell_exec(
```
theofidry, line 51: missing the import of `shell_exect()` & `escapeshellarg()`

src/Logger/GitHubAnnotationsLogger.php
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    public function getLogLines(): array
+    {
+        $lines = [];
+        $currentWorkingDirectory = getcwd();
+
+        foreach ($this->metricsCalculator->getEscapedExecutionResults() as $escapedExecutionResult) {
+            $error = [
+                'line' => (string) $escapedExecutionResult->getOriginalStartingLine(),
```
theofidry, line 60: Legit?

src/Logger/GitHubAnnotationsLogger.php
```diff
@@ -0,0 +1,94 @@
...
+        // newlines need to be encoded
+        // see https://github.com/actions/starter-workflows/issues/68#issuecomment-581479448
+        $message = str_replace("\n", '%0A', $error['message']);
+
+        return "::warning file={$filePath},line={$error['line']}::{$message}\n";
+    }
+
+    private function relativePath(string $currentWorkingDirectory, string $path): string
+    {
+        return str_replace($currentWorkingDirectory . '/', '', $path);
```
theofidry, line 92: missing import for `str_replace`

src/Logger/GitHubAnnotationsLogger.php
```diff
@@ -0,0 +1,94 @@
...
+
+        return $lines;
+    }
+
+    /**
+     * @param array{line: string, message: string} $error
+     */
+    private function buildAnnotation(string $filePath, array $error): string
+    {
+        // newlines need to be encoded
```
theofidry, line 83:
```suggestion
        // New lines need to be encoded
```
?

src/Logger/GitHubAnnotationsLogger.php
```diff
@@ -0,0 +1,94 @@
...
+                'message' => <<<"TEXT"
+Escaped Mutant:
+
+{$escapedExecutionResult->getMutantDiff()}
+TEXT
+            ,
+            ];
+
+            $lines[] = $this->buildAnnotation(
+                $this->relativePath($currentWorkingDirectory, $escapedExecutionResult->getOriginalFilePath()),
```
theofidry, line 70:
```suggestion
                Path::makeRelative($escapedExecutionResult->getOriginalFilePath(), $currentWorkingDirectory),
```

src/Logger/GitHubAnnotationsLogger.php
```diff
@@ -0,0 +1,94 @@
...
+                $error
```
theofidry, line 71: pure not, no need to address this here, but I would recommend to always append a comma when possible as it makes the diffs more readable whenever you add an argument or may make it easier to append and argument sometimes

tests/phpunit/Logger/GitHubAnnotationsLoggerTest.php
```diff
@@ -0,0 +1,76 @@
...
+    public function test_it_logs_correctly_with_mutations(
+        MetricsCalculator $metricsCalculator,
+        array $expectedLines
+    ): void {
+        $logger = new GitHubAnnotationsLogger($metricsCalculator);
+
+        $this->assertSame($expectedLines, $logger->getLogLines());
+    }
+
+    public function metricsProvider(): iterable
```
theofidry, line 61:
```suggestion
    public static function metricsProvider(): iterable
```

theofidry (approved):
Looks quite good to me, now I need to catch up with the thread :P

theofidry:
I'm not sure I follow the whole discussion, it's a bit tedious to see what's still relevant and what's outdated...

Do I understand correctly that the goal is to be able to enable this logger from both the config and command line? From what I can see so far we do not allow a toggle for any logger from the command line so what's the rational there?

>Because fixing all the tests that work with Configuration, SchemaConfiguration and Container takes 90% of the time spent on this PR.
>BTW, this is something to think about - how to make tests not so painful. Any new parameter added to Container::withValues() brings so many changes in the tests which are very hard to understand, look at this:

Agreed. For the record when I added those, I was happy with the result in terms of reading the flow, testability & coverage (which were all quite low before that). I however was already not happy with the experience when changing a new parameter: changing the source code is not that bad but the tests is just horrible. Maybe it's us being too exhaustive there or maybe we are missing a bit of abstraction (within the tests) to make it more expressive & flexible. Food for thoughts I guess!

I must say I quite like the PR otherwise, I did not expect the logger implementation to be this simple! 👍 

I am sure there might be some tweaking in the future, maybe adding the mutator name or something like that, but pretty this can definitely be done later on as we get feedback from the community

src/Configuration/ConfigurationFactory.php
maks-rafalko: well, there is [Rector PHP](https://github.com/rectorphp/rector) that does exactly that - updating/upgrading the code to new features. But yeah, this is for a separate PR. I just changed it where touched the file

src/Logger/GitHubAnnotationsLogger.php
maks-rafalko: seems like yes, tried to fix, let's see

tests/phpunit/Logger/GitHubAnnotationsLoggerTest.php
maks-rafalko: we can't, since we use `$this->createCompleteMetricsCalculator()` inside this method

src/Configuration/ConfigurationFactory.php
theofidry: I admit I still never managed to get Rector working properly for me... (e.g. for using the safe functions when possible)

maks-rafalko:
> Do I understand correctly that the goal is to be able to enable this logger from both the config and command line? From what I can see so far we do not allow a toggle for any logger from the command line so what's the rational there?

Initially, I've implemented this logger in a `infection.json` file only, as all other ones. But we understood that this logger should be used **only** on GH Actions, and it does not make sense to have it inside config under git. So I implemented a CLI option as well. In the future, we will make an update to make it possible to enable all the loggers by CLI or config setting.



> I must say I quite like the PR otherwise, I did not expect the logger implementation to be this simple! +1

:clap: 


then waiting for @sanmai's review and ready to be released!

sanmai (approved):
Let’s merge this!

maks-rafalko:
Thank you all for the great review and ideas!

sanmai:
There's a Symfony's special reporting component now:

https://github.com/symfony/symfony/blob/f0aa4e7df17d40eadf540c6c3d0f9e6feac67648/src/Symfony/Component/Console/CI/GithubActionReporter.php#L21

src/Logger/GitHub/GitDiffFileProvider.php
```diff
@@ -0,0 +1,61 @@
...
+            sprintf(
+                'git diff %s --diff-filter=%s --name-only | grep src/ | paste -sd ","',
```
Slamdunk, line 55: Hardcoded `src/` is likely to be wrong, it should be picked by config.
PR coming soon

PR #1369, opened 2020-10-23
Improve Dockerfiles

sanmai:
- Moves tools install command lower to make sure we don't rebuild the whole thing after every change in the list.
- Adds some essential packages.
- Configures Xdebug on PHP 8.

### DIFF
--- a/devTools/Dockerfile-php80-xdebug
+++ b/devTools/Dockerfile-php80-xdebug
@@ -1,10 +1,5 @@
 FROM php:8.0-rc-cli-alpine
 
-RUN apk add --no-cache \
-        bash \
-        expect \
-        zip
-
 ARG XDEBUG_VERSION=3.0.0beta1
 
 RUN set -eux; \
@@ -25,8 +20,16 @@ RUN set -eux; \
     apk add --no-cache --virtual .phpexts-rundeps ${runDeps}; \
     apk del .build-deps
 
+RUN apk add --no-cache \
+        ncurses \
+        make \
+        bash \
+        expect \
+        zip
+
 COPY --from=composer:latest /usr/bin/composer /usr/bin/composer
 COPY devTools/memory-limit.ini ${PHP_INI_DIR}/conf.d/memory-limit.ini
+COPY devTools/xdebug-coverage.ini ${PHP_INI_DIR}/conf.d/xdebug-coverage.ini
 
 RUN adduser -h /opt/infection -s /bin/bash -D infection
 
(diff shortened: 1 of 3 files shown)


devTools/Dockerfile-php80-xdebug
```diff
@@ -25,8 +20,16 @@ RUN set -eux; \
...
+RUN apk add --no-cache \
+        ncurses \
+        make \
+        bash \
+        expect \
+        zip
+
 COPY --from=composer:latest /usr/bin/composer /usr/bin/composer
 COPY devTools/memory-limit.ini ${PHP_INI_DIR}/conf.d/memory-limit.ini
+COPY devTools/xdebug-coverage.ini ${PHP_INI_DIR}/conf.d/xdebug-coverage.ini
```
sidz, line 32: this can be a bit optimized to decrease amount of useless layers:
`COPY devTools/memory-limit.ini devTools/xdebug-coverage.ini ${PHP_INI_DIR}/conf.d/`

devTools/Dockerfile-php80-xdebug
```diff
@@ -25,8 +20,16 @@ RUN set -eux; \
     apk add --no-cache --virtual .phpexts-rundeps ${runDeps}; \
     apk del .build-deps
 
+RUN apk add --no-cache \
```
sidz, line 23: Maybe for Infection it's ok but usually exactly this part is persistent and does not change very often.

p.s. I'd leave it right after `FROM`.

devTools/Dockerfile-php80-xdebug
```diff
@@ -25,8 +20,16 @@ RUN set -eux; \
...
+        ncurses \
+        make \
```
sidz, line 25: why do we need to have `make` inside container? I didnt' see that we use it elsewhere except on host machine.

sanmai: To run the whole set of E2E tests with benchmarks and stuff: `make test-e2e`
(It is possible to do without `make`, but quite a hassle.)

devTools/Dockerfile-php80-xdebug
sanmai: Well, I had to change it several times already, and I can say that a build for Xdebug takes more than few moments.

devTools/Dockerfile-php80-xdebug
sanmai: Sure, done.

devTools/Dockerfile-php80-xdebug
sidz: fine with me.
Don't expect that we will install other php extensions in the nearest feature.

sidz (approved):
:+1:

sanmai:
Thank you @sidz

PR #1335, opened 2020-10-02
Fix spelling in console output

MidnightDesign:
Fixes the spelling in the "consider increasing the required..." message

PR #1374, opened 2020-10-25
Fix updated PHPStan issues, update composer.lock

sanmai:
Ref: https://github.com/infection/infection/runs/1304815307#step:7:87


sanmai:
Merged as part of #1335

PR #1375, opened 2020-10-26
Request coverage reports with command line args

sanmai:
This PR:

- [x] Request coverage reports from PHPUnit with command line args, thus avoiding working [with the new schema](https://github.com/sebastianbergmann/phpunit/blob/a0d6b21c6c8f6564212a1a14292d230ee35eba6d/ChangeLog-9.3.md#configuration-of-code-coverage-and-logging-in-phpunitxml).
- [x] Removes new-style loggers from the configuration
- [x] Replaces the bulk of #1294
- [x] Covered by tests

What left to do with #1283:

- There has to be a test with new schema.
- New-style coverage "whitelist" adder has to be dealt with (`addCoverageFilterWhitelistIfDoesNotExist`).

### DIFF
--- a/src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
+++ b/src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
@@ -99,7 +99,7 @@ class PhpUnitAdapter extends AbstractTestFrameworkAdapter implements IgnoresAddi
             $extraOptions = trim(sprintf(
                 '%s --coverage-xml=%s --log-junit=%s',
                 $extraOptions,
-                $this->tmpDir . DIRECTORY_SEPARATOR . self::COVERAGE_DIR,
+                $this->tmpDir . '/' . self::COVERAGE_DIR,
                 $this->jUnitFilePath // escapeshellarg() is done up the stack in ArgumentsAndOptionsBuilder
             ));
         }
--- a/src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
+++ b/src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
@@ -38,8 +38,13 @@ namespace Infection\TestFramework\PhpUnit\Adapter;
 use Infection\AbstractTestFramework\MemoryUsageAware;
 use Infection\PhpParser\Visitor\IgnoreNode\PhpUnitCodeCoverageAnnotationIgnorer;
 use Infection\TestFramework\AbstractTestFrameworkAdapter;
+use Infection\TestFramework\CommandLineArgumentsAndOptionsBuilder;
+use Infection\TestFramework\CommandLineBuilder;
+use Infection\TestFramework\Config\InitialConfigBuilder;
+use Infection\TestFramework\Config\MutationConfigBuilder;
 use Infection\TestFramework\IgnoresAdditionalNodes;
 use Infection\TestFramework\ProvidesInitialRunOnlyOptions;
+use Infection\TestFramework\VersionParser;
 use function Safe\preg_match;
 use function Safe\sprintf;
 use function version_compare;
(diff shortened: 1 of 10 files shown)


src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
```diff
@@ -52,11 +57,56 @@ class PhpUnitAdapter extends AbstractTestFrameworkAdapter implements IgnoresAddi
...
+        CommandLineArgumentsAndOptionsBuilder $argumentsAndOptionsBuilder,
+        VersionParser $versionParser,
+        CommandLineBuilder $commandLineBuilder,
+        ?string $version = null
+    ) {
+        parent::__construct($testFrameworkExecutable, $initialConfigBuilder, $mutationConfigBuilder, $argumentsAndOptionsBuilder, $versionParser, $commandLineBuilder, $version);
+
+        $this->tmpDir = $tmpDir;
+        $this->jUnitFilePath = $jUnitFilePath;
+    }
```
sanmai, line 79: Extending constructors does not break LSP.

src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
```diff
@@ -52,7 +52,7 @@ public function build(string $configPath, string $extraOptions): array
                 '--configuration',
                 $configPath,
             ],
-            explode(' ', $extraOptions)
+            explode(' ', $extraOptions) // FIXME might break space-containing paths
```
sanmai, line 55: Not fixing this issue in this PR.

tests/phpunit/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilderTest.php
```diff
@@ -220,35 +217,14 @@ public function test_it_removes_original_loggers(): void
         $this->assertSame(0, $logEntries->length);
     }
 
-    public function test_it_adds_needed_loggers(): void
+    public function test_it_does_not_add_loggers_ever(): void
```
sanmai, line 220:
```suggestion
    public function test_it_does_not_add_coverage_loggers_ever(): void
```

src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
```diff
@@ -52,11 +57,56 @@ class PhpUnitAdapter extends AbstractTestFrameworkAdapter implements IgnoresAddi
...
+     * @param string[] $phpExtraArgs
+     *
+     * @return string[]
+     */
+    public function getInitialTestRunCommandLine(
+        string $extraOptions,
+        array $phpExtraArgs,
+        bool $skipCoverage
+    ): array {
+        if ($skipCoverage === false) {
```
maks-rafalko, line 98: question: is it any better than `!$skiptCoverage`. Found a couple of places and all ones from you. I mean we usually use `!$condition` with booleans, probably worth unifying

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
```diff
@@ -82,9 +82,11 @@ public function replaceWithAbsolutePaths(SafeDOMXPath $xPath): void
     public function removeExistingLoggers(SafeDOMXPath $xPath): void
     {
         foreach ($xPath->query('/phpunit/logging') as $node) {
-            $document = $xPath->document->documentElement;
-            Assert::isInstanceOf($document, DOMElement::class);
-            $document->removeChild($node);
+            $node->parentNode->removeChild($node);
```
maks-rafalko, line 85: what do you think about adding a comment to this method (or method body) than loggers will be added by Infectin through CLI arguments? will be easier to understand the whole picture for future readers

maks-rafalko (approved):

src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
sanmai: For one, `!$condition` is messy, it might trigger on a `null`. For another, use `===` is more generic to all types, more uniform. What do you think? 

Is there a fixer for `!$bool`?

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
sanmai, line 88: How about now?

src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
sanmai: I was very hesitant about `===` if you remember, and I'm still not exactly on board with the arguments I've seen back then. True, it can be ugly when comparing strings and numbers, for example, but the real reason I'm using `===` everywhere effectively _out of habit_ now because it prevents unnecessary and avoidable BC breaks. 

It is not like someone would extend the very this class, but if this happens, identical comparison gets us prepared. For example, if you were using `!$condition` and someone extending extending this class decides put a different falsy value here, you can't really change it to something else without a BC break. Say, you want to give a `null` special meaning? Well, you can't, because `null` is falsy. On the contrary, if you use `$condition === false`, you're protected from said BC breaks. Add any nulls anytime you want! That's the reason. Makes sense?

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
sanmai: How about now?

src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
maks-rafalko: makes sense

I know the reason why it looks weird for me, because I've been using phpstan with strict rules for 2 years on my job, and it's impossible to use non-boolean comparison there.

[Look at this code](https://phpstan.org/r/9e7ecebe-c62f-4e3c-9887-3ef449c61067) with strict rules: 

```php
$null = rand(0, 1) === 0 ? null : true;
		
if (!$null) {
			
}
```

error: `Only booleans are allowed in a negated boolean, true|null given.`. So if it's nullable - makes sense to use `=== true`, but when it's `bool` only - it's safe to use just `!$null` here.

Regarding your example, I guess it's impossible to do it: https://phpstan.org/r/b07764ac-22c8-4c06-ad83-a19ba9e09381 - we can pass only booleans to the parent (original) method.

Anyway, it's not critical, we can leave it. Just not usual usage for me, that's why I asked.

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
maks-rafalko: cool!

src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
sanmai: Sure. It's nice to have PHPStan warn you about improper negation, but it's a liability nevertheless. E.g. where negation can be improper, `===` is always safe to use.

PR #1376, opened 2020-10-26
Integration group annotation does not take comments

sanmai:
I think we discussed this before, and these were dangling from that time.


theofidry:
Could be worth reporting upstream? It's quite annoying tbh :(

theofidry (approved):

sanmai:
Might well just as be.

Thank you @theofidry

PR #1378, opened 2020-10-27
Refactor InitialConfigBuilder

### LINKED ISSUE #1283: PHPUnit 9.3 compatibility issue

sanmai:
This PR:

- [x] Adds heuristics to detect used PHPUnit configuration format.
- [x] Add a method to update PHPUnit configuration with appropriate coverage whitelist/include directives.
- [x] Fixes #1283 to the point we're using Infection with the newest PHPUnit configuration.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/schema/9.2.xsd"
+         xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/phpunit.xsd"
          bootstrap="./vendor/autoload.php"
          colors="true"
          executionOrder="random"
@@ -23,11 +23,11 @@
         </testsuite>
     </testsuites>
 
-    <filter>
-        <whitelist>
+    <coverage>
+        <include>
             <directory>src</directory>
-        </whitelist>
-    </filter>
+        </include>
+    </coverage>
 
     <groups>
         <exclude>
--- a/src/TestFramework/PhpUnit/Adapter/PhpUnitAdapterFactory.php
+++ b/src/TestFramework/PhpUnit/Adapter/PhpUnitAdapterFactory.php
@@ -44,6 +44,7 @@ use Infection\TestFramework\PhpUnit\Config\Builder\InitialConfigBuilder;
 use Infection\TestFramework\PhpUnit\Config\Builder\MutationConfigBuilder;
 use Infection\TestFramework\PhpUnit\Config\Path\PathReplacer;
 use Infection\TestFramework\PhpUnit\Config\XmlConfigurationManipulator;
+use Infection\TestFramework\PhpUnit\Config\XmlConfigurationVersionProvider;
 use Infection\TestFramework\VersionParser;
 use function Safe\file_get_contents;
 use Symfony\Component\Filesystem\Filesystem;
@@ -87,6 +88,7 @@ final class PhpUnitAdapterFactory implements TestFrameworkAdapterFactory
                 $tmpDir,
                 $testFrameworkConfigContent,
                 $configManipulator,
+                new XmlConfigurationVersionProvider(),
                 $sourceDirectories
             ),
             new MutationConfigBuilder(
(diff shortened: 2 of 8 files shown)


maks-rafalko (approved):
Looks good :+1:

sanmai:
Great! Release-worthy, what do you think?

maks-rafalko:
Sure, will do it tonight

PR #1379, opened 2020-10-27
Trigger Travis CI deploy stage only for tagged builds

sanmai:
Should cut about a minute or two of build time.


maks-rafalko (approved):
:+1:

PR #1383, opened 2020-10-28
Remove AppVeyor as we use GitHub actions with Windows

maks-rafalko:
Also, I've removed AppVeyor integration with Infection


sanmai:
Fingers crossed we won't need to patch up older branches! 

BTW, how about them? It is safe to drop a release branches since you can start a new branch from a tag at any time.

sanmai (approved):

maks-rafalko:
> BTW, how about them? It is safe to drop a release branches since you can start a new branch from a tag at any time.

removed all old release branches

PR #1386, opened 2020-10-28
Extend (reuse) JSON definitions to avoid issues where we have to duplicate part of the definitions

maks-rafalko:
This PR is to show how we can reuse definitions and [extend them](https://json-schema.org/understanding-json-schema/structuring.html#extending), by adding custom keys. (follow up to https://github.com/infection/infection/pull/1385)

But here we have a big issue from my POV:

when we are using extending via `allOf`, we can't use `"additionalProperties": false,`, because sub-schemas are applied one by one and the object gets rejected.

So, when we are using `allOf` - we have to remove `additionalProperties`, which introduces a potential bug when users can add not supported key.

More about it: http://json-schema.org/understanding-json-schema/reference/combining.html#allof

---

I don't know how to resolve it. From the link above:

>  There are some proposals to address this in the next version of the JSON schema specification.

Seems like we can't use `allOf` in our case, unfortunately.

But if someone knows how to do it - please let me know.

### DIFF
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -190,34 +190,25 @@
                             "type": "boolean"
                         },
                         {
-                            "type": "object",
-                            "additionalProperties": false,
-                            "properties": {
-                                "ignore": {
-                                    "type": "array",
-                                    "items": {
-                                        "type": "string"
-                                    }
-                                },
-                                "ignoreSourceCodeByRegex": {
-                                    "type": "array",
-                                    "items": {
-                                        "type": "string"
-                                    }
-                                },
-                                "settings": {
-                                    "type": "object",
-                                    "additionalProperties": false,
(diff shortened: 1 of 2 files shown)


sanmai:
I don't have an expression this compromise is worth it. 

Can't we read the schema from a test, and verify that every mutator element contains the `ignoreSourceCodeByRegex` key?

maks-rafalko:
> Can't we read the schema from a test, and verify that every mutator element contains the `ignoreSourceCodeByRegex` key?

It's more complicated than just checking  `ignoreSourceCodeByRegex`, because we should check all the **new** case as well. But seems like we should do smth like that.

I hoped inheritance in JSON Schema specification could do the job, but it's not the case

Issue #1387, opened 2020-10-29
Mutator proposal: Flip null coalescing operator sides


MidnightDesign:
I couldn't find a mutator that mutates `$a ?? $b` to `$b ?? $a`. It would uncover the following case:

```php
function pickOne(?string $a, ?string $b): ?string {
    return $b ?? $a;
}

// The test to kill the mutant:
public function testPrefersBOverA(): void
{
    $picked = pickOne('foo', 'bar');

    self::expectSame('bar', $picked);
}
```

Playground link: https://infection-php.dev/r/v4d

maks-rafalko:
Sounds good. Feel free to implement it, shouldn't be hard. And we are here to help!

sidz:
Implemented in: https://github.com/infection/infection/pull/1389

PR #1389, opened 2020-10-29
Add Flip Coalesce operator mutator

sidz:
This PR:

- [x] Adds new FlipCoalesce mutator
- [x] Covered by tests
- [x] Schema needs an update
- [x] Doc PR: https://github.com/infection/site/pull/185

Mutates:
```diff
$foo = 'foo';
$bar = 'bar';
- $foo ?? $bar;
+ $bar ?? $foo;
```
```diff
$foo = 'foo';
$bar = 'bar';
$baz = 'baz';
- $foo ?? $bar ?? $baz;
+ $foo ?? $baz ?? $bar;
```
```diff
$foo = 'foo';
$bar = 'bar';
$baz = 'baz';
- $foo ?? $bar ?? $baz;
+ $bar ?? $foo ?? $baz;
```

Requested in https://github.com/infection/infection/issues/1387

p.s. From my POV `Infection\Mutator\Operator\Coalesce` mutator is redundant when this PR is merged. Thoughts?

### DIFF
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -247,6 +247,7 @@
                 "Throw_": { "$ref": "#/definitions/default-mutator-config" },
                 "Finally_": { "$ref": "#/definitions/default-mutator-config" },
                 "Coalesce": { "$ref": "#/definitions/default-mutator-config" },
+                "FlipCoalesce": { "$ref": "#/definitions/default-mutator-config" },
                 "PregQuote": { "$ref": "#/definitions/default-mutator-config" },
                 "PregMatchMatches": { "$ref": "#/definitions/default-mutator-config" },
                 "ArrayItemRemoval": {
--- a/src/Mutator/Operator/FlipCoalesce.php
+++ b/src/Mutator/Operator/FlipCoalesce.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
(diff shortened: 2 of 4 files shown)


maks-rafalko (approved):
:+1:

maks-rafalko:
> p.s. From my POV `Infection\Mutator\Operator\Coalesce` mutator is redundant when this PR is merged. Thoughts?

I didn't do any analytics, but from the first look - agree, sounds like redundant

sanmai:
`Coalesce` is redundant indeed.

sanmai (commented):
How about [`self::A ?? self::B`](https://3v4l.org/Vf17i)? And with constants in general. These might be unkillable.

maks-rafalko:
> How about [`self::A ?? self::B`](https://3v4l.org/Vf17i)?

is it a real case? I mean it's strange to use constant in `if` statement. It's either always true or always false

sanmai:
It can be any kind of constant expression. 
```php
define('A_CONST', null);
class B {
    const A = A_CONST;
}
```
No doubt it's a poor programming style, but shall we punish it?

maks-rafalko:
Probably I don't understand the real case and that's why don't see an issue really. Any real examples? I just can'r remember any cases where with such code `self::A ?? self::B`, even with ` self::A ?? $variable`

sanmai:
The issue is that you can't test `self::A ?? self::B`. This could be a convenience method in a generated class.

maks-rafalko:
So what are you suggesting? Do not mutate when the left side is a const?

Alternatively, if the user has such a weird case, they can ignore such line of the code, via `ignoreSourceCodeByRegex`

sanmai:
Yes, I think it's a general enough case to ignore all mutations with a constant fetch on the left.

maks-rafalko:
TBH I'm not convinced, as I have never seen such code. But if you did - then ok.

sanmai:
Let me put this way: in which circumstances do you thing running this mutation against a constant on a left will be _useful_?

Again, it is not about this mutation is harmful, it's not, but rather it is _not_ useful.

sanmai:
If I learnt something from the `IdenticalEqual`/`NotIdenticalNotEqual` debacle is that if a mutation might be seen as a punishment for a poor programming style, you shouldn't do it. Any such mutations make Infection less useful, and that's bad.

sidz:
@sanmai done

sanmai (approved):

PR #1394, opened 2020-10-30
Split schema.json into root schema and mutators schema

sanmai:
This PR:

- [x] Splits schema.json into root schema and mutators schema
- [x] Should help make  other schemas


maks-rafalko:
What is the final goal and how it will help, could you please clarify?

sanmai:
Recall you had this problem with playground where there wasn’t a new mutant. And you told it uses a different schema for good reasons. So now you can include mutants.json into your custom schema, and largely forget about updating it.

maks-rafalko:
> So now you can include mutants.json into your custom schema, and largely forget about updating it.

you mean using public URL and pointing to https://raw.githubusercontent.com/infection/infection/3dc1f8dd9dec65b258657f0c54569ae7b868459a/resources/mutators.json?

sanmai:
This will be quite ineffective. If you're using Infection as a dependency you can include all mutators into another schema like so:
```json
 "$ref": "../vendor/infection/infection/resources/mutators.json#/mutators"
```

sanmai:
Well, and you could actually do it like so, without this PR:
```
 "$ref": "../vendor/infection/infection/resources/schema.json#/mutators"
```

I'll have it closed for now.

PR #1401, opened 2020-10-31
[DEBUG] Test github logger from same repository, not from a fork

maks-rafalko:
this one is for testing how github logger works when PR is created from the same repo, not from a fork as I did on original implementation PR.

Let's see, hopefully, there is no difference.


maks-rafalko:
It works, cool

https://github.com/infection/infection/actions/runs/339412989

PR #1403, opened 2020-11-02
Refactor command to constants

vkhramtsov:
This PR:
- [x] Covered by tests

### DIFF
--- a/src/Command/RunCommand.php
+++ b/src/Command/RunCommand.php
@@ -67,13 +67,82 @@ use function trim;
  */
 final class RunCommand extends BaseCommand
 {
+    /** @var string */
+    private const OPTION_TEST_FRAMEWORK = 'test-framework';
+
+    /** @var string */
+    private const OPTION_TEST_FRAMEWORK_OPTIONS = 'test-framework-options';
+
+    /** @var string */
+    private const OPTION_THREADS = 'threads';
+
+    /** @var string */
+    private const OPTION_ONLY_COVERED = 'only-covered';
+
+    /** @var string */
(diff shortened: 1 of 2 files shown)


src/Command/RunCommand.php
```diff
@@ -159,73 +228,76 @@ protected function configure(): void
                 Container::DEFAULT_FORMATTER_NAME
             )
             ->addOption(
-                'git-diff-filter',
+                self::OPTION_GIT_DIFF_FILTER,
                 null,
                 InputOption::VALUE_REQUIRED,
                 'Filter files to mutate git `--diff-filter` options. A - only for added files, AM - for added and modified.',
```
sanmai, line 234: `--diff-filter`?

vkhramtsov: I think that it is not applicable here because we have `diff-filter` only once in this command

vkhramtsov: I mean in this file

sanmai (commented):
On one hand this is something I'd do if I'd have to make a command from scratch. On other hand I can't help but notice that the code became less readable after this change.

vkhramtsov:
At the same time this change able to help with decreasing possibility of error. Anyway, it is just proposal

maks-rafalko (approved):
I don't think we will often read this code, so as for me - having constants probably better than having a possible error. But that's just my opinion

sanmai:
Fair enough.

sanmai:
Fair enough.

Thank you @vkhramtsov

PR #1405, opened 2020-11-02
Add While expression mutator

sidz:
This PR:

- [x] Adds new While expression mutator
- [x] Covered by tests
- [x] Schema needs an update
- [x] Doc PR: https://github.com/infection/site/pull/196

Mutates:
```diff
$condition = true;
- while ($condition) {
+ while (false) {
```

### DIFF
--- a/devTools/phpstan-src-baseline.neon
+++ b/devTools/phpstan-src-baseline.neon
@@ -666,6 +666,11 @@ parameters:
 			path: ../src/Mutator/ZeroIteration/Foreach_.php
 
 		-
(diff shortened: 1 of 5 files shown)


maks-rafalko (approved):
Cool.

What about `do while()`? should we add similar mutator?

sidz:
@maks-rafalko yes we should and I also have a plan to add it but it is not a ZeroIteration mutator. I'm thinking about deprecate zero_iteration profile in favor of loop (we also have Loop section in documentation).

But not sure how to achieve this with json schema validator so far. (found that latest shcema (https://json-schema.org/draft/2019-09/release-notes.html#meta-data-vocabulary) has "deprecated" property)

Or we may do it a bit aggressive and just rename `zero_iteration` profile to `loop` and realase in the next "major".

what do you think @maks-rafalko, @sanmai ?

sanmai:
I'm for the rename. Next major assumes BC breaks, and that's what it is.

sidz:
thanks @maks-rafalko , @sanmai

PR #1407, opened 2020-11-02
Rename @zero_iteration profile to the @loop

sidz:
This PR renames `@zero_iteration` profile to the `@loop` as agreed in https://github.com/infection/infection/pull/1405.

We would like to add DoWhile mutator which is not a zero iteration mutator but all of them are `loop` mutators


maks-rafalko (approved):
:ok_hand:

sidz:
thanks @maks-rafalko

Issue #1408, opened 2020-11-02
Order of keys in `mutators` config array, leads to different results


canvural:
| Question    | Answer
| ------------| ---------------
| Infection version | `0.19.1@46c17a7fdba19ad05bec7b10e006ce2f00d954a1`, `0.20.0@735506ce5fc35dd504c9ba147d5fe0554898b232`
| Test Framework version | PHPUnit 9.3
| PHP version | 7.4 ?
| Platform    | Playground
| Github Repo | -

[Here](https://infection-php.dev/r/ky2) is playground link which ignores a `DecrementInteger` mutation in `Infected\\SourceClass::add` method. Works as expected.

But if the array key order changed and `@default": true` comes after the ignore rule, it does not work as expected. Can be seen [here.](https://infection-php.dev/r/3y6)

Is this expected? Couldn't find anything in the documentation. And lost 45 minutes to this, trying to figure out why wouldn't it ignore it.

sidz:
as for me it sounds like an issue.
Good catch btw :+1:

sanmai:
I believe a similar issue was fixed in 0.20. Can you check that you can reproduce with it?

canvural:
@sanmai Yes, it also happens with `0.20.0@735506ce5fc35dd504c9ba147d5fe0554898b232`

Although you can't see the actual infection log file, I had a failing (cause MSI was too low due to not ignoring) CI job [here.](https://github.com/canvural/php-openapi-faker/runs/1343464133?check_suite_focus=true)

sanmai:
Thank you for the report! There was another similar issue we managed to discover/fix. Should be in a next release.

canvural:
Thank you @sanmai, @sidz and @maks-rafalko

PR #1409, opened 2020-11-02
Order of keys in `mutators` config array, leads to different results

sidz:
Adds test to proof the issue from https://github.com/infection/infection/issues/1408


sidz:
Closed in favor of https://github.com/infection/infection/pull/1410

Issue #1406, opened 2020-11-02
PcreException should mention which regexp is actually invalid


sstok:
**Is your feature request related to a problem? Please describe.**
When a regex in the `ignoreSourceCodeByRegex` is invalid this crashes the application with:

```
[PHP Warning:  preg_match(): Compilation failed: missing closing parenthesis at offset 58 in [...]/infection/vendor/thecodingmachine/safe/generated/pcre.php on line 590]

In PcreException.php line 19:

  [Safe\Exceptions\PcreException (1)]
  PREG_INTERNAL_ERROR: Internal error
``` 

**Describe the solution you'd like**
This is not very helpful because it doesn't show which regex is invalid. Making debugging rather challenging.

I'm not sure this should be resolved here or in the `thecodingmachine/safe` package?
Instead when this exception is thrown it should show which regexp is invalid. Even better if it's able to mention which config-line is happens at but that's just dreaming 😁

sidz:
Implemented in https://github.com/infection/infection/pull/1412

PR #1412, opened 2020-11-03
Check ignoreSourceCodeByRegex and global-ignoreSourceCodeByRegex have valid syntax

sidz:
PR implemented https://github.com/infection/infection/issues/1406 request.

Check `ignoreSourceCodeByRegex` and `global-ignoreSourceCodeByRegex` config options that both have valid syntax.

`global-ignoreSourceCodeByRegex` is invalid:
image
`ignoreSourceCodeByRegex`  is invalid:
image

### DIFF
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -129,7 +129,8 @@
                 "global-ignoreSourceCodeByRegex": {
                     "type": "array",
                     "items": {
-                        "type": "string"
+                        "type": "string",
+                        "format": "regex"
                     }
                 },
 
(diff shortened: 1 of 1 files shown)


maks-rafalko (approved):
Wow, nice solution.

sanmai (approved):

sidz:
thanks @maks-rafalko and @sanmai

sstok:
Wow this is amazing 🎉  thanks sidz.

sstok:
Sorry to be the bringer of bad news 😅 but when the regexp contains a `/` this will fail the schema validation, by escaping this, it passes the validation but because the https://github.com/infection/infection/blob/master/src/Differ/DiffSourceCodeMatcher.php#L47 escapes all slashes this causes it to be double escaped, meaning the regex becomes invalid.

For now I solved this by using `\x{2f}` but as the regexp is already invalid with `/` unescaped this line can be removed.
(Sorry I didn't create a separate issue, but it's late for me.)

PR #1415, opened 2020-11-07
Use not included delimiter

### LINKED ISSUE #1414: Escaping issues in `ignoreSourceCodeByRexeg` feature
when the regexp contains a `/` this will fail the schema validation, by escaping this, it passes the validation but because the https://github.com/infection/infection/blob/master/src/Differ/DiffSourceCodeMatcher.php#L47 escapes all slashes this causes it to be double escaped, meaning the regex becomes invalid.

For now I solved this by using `\x{2f}` but as the regexp is already invalid with `/` unescaped this line can be removed.

_Originally posted by @sstok in https://github.com/infection/infection/pull/1412#issuecomment-721350069_

sanmai:
This PR:

- [x] Changes DiffSourceCodeMatcher to use a delimiter that isn't included in the regex, instead of escaping the default one. 
- [x] Covered by tests
- [x] Fixes #1414

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1954,16 +1954,16 @@
     "packages-dev": [
         {
             "name": "amphp/amp",
-            "version": "v2.5.0",
+            "version": "v2.5.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/amphp/amp.git",
-                "reference": "f220a51458bf4dd0dedebb171ac3457813c72bbc"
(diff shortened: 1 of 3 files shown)


maks-rafalko (approved):

maks-rafalko:
Thank you @sanmai

sstok:
Thanks.

PR #1420, opened 2020-11-09
Deferify Mutant

sanmai:
This PR is a follow-up to #1422:

- [x] Deferify MutantExecutionResult
- [x] Deferify MutantFactory
- [x] [Blackfire diff](https://blackfire.io/profiles/compare/545bda96-16a1-41ff-9ed7-db1126e2876f/graph) shows **34%** time reduction
- [x] As seen on CI: [7m 10s](https://github.com/infection/infection/runs/1377310526#step:8:1)  on master vs [6m 21s](https://github.com/infection/infection/runs/1378147000#step:8:1) for this PR
- [x] Covered by tests

Running this on Psalm with a config below, without any file logging:
```diff
- Time: 2m 23s. Memory: 2.15GB
+ Time: 1m 42s. Memory: 2.15GB
```
<details>
  <summary>infection.json and command line arguments</summary>

```json
{
    "timeout": 2,
    "source": {
        "directories": [
            "src"
        ]
    },
    "mutators": {
        "@unwrap": true
    }
}
```
Command line:
```
infection.phar --skip-initial-tests --coverage=build/logs -j2
```
</details>

There's almost no difference if logging is enabled, which makes sense: we'll have to unlazify the diffs to log them.
```diff
- Time: 2m 24s. Memory: 2.15GB
+ Time: 2m 23s. Memory: 2.15GB
```
With `--no-progress --show-mutations` and no additional logging I get:
```diff
- Time: 2m 17s. Memory: 0.29GB
+ Time: 1m 38s. Memory: 0.53GB
```
This is 30% faster! 

It is true this lazy evaluation isn't coming without a price, but I tend to think it's worth it. More about it below.

### Discussion

It is clear to see the speed boost we're having, but there are drawbacks:
1. This boost comes at the expense of raised top memory consumption. The program has to hold chunks of AST instead of flat diffs and source code listings. This is possible to remedy by further work (continued in #1425): 
    - Now we're holding on to all mutation results, where it should be possible to discard unneeded results as we go. E.g. when running with `--only-covered` it makes no sense to hold on to uncovered mutations. We might count them, but that should be it.
    - Likewise, when there's no logging other that to show escaped mutations, we might discard other mutations as soon as we count them.
    - And more, we can stream-write to logs as we receive mutation results.
2. Most of the work to render mutants is being postponed, therefore there could be a lag to write to log files upon finishing with all mutations. This could be remedied by adding an extra progress bar indicator for log-writing process.
2. With all this deferred execution the program is slightly less easy to follow. There's no remedy.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,7 +55,7 @@
         "ocramius/package-versions": "^1.2 || ^2.0",
         "ondram/ci-detector": "^3.3.0",
         "sanmai/later": "^0.1.1",
-        "sanmai/pipeline": "^3.1 || ^5.0",
+        "sanmai/pipeline": "^5.1",
         "sebastian/diff": "^3.0.2 || ^4.0",
         "seld/jsonlint": "^1.7",
         "symfony/console": "^3.4.29 || ^4.1.19 || ^5.0",
--- a/composer.json
+++ b/composer.json
@@ -54,6 +54,7 @@
         "nikic/php-parser": "^4.10.2",
         "ocramius/package-versions": "^1.2 || ^2.0",
         "ondram/ci-detector": "^3.3.0",
+        "sanmai/later": "^0.1.1",
         "sanmai/pipeline": "^3.1 || ^5.0",
         "sebastian/diff": "^3.0.2 || ^4.0",
         "seld/jsonlint": "^1.7",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "956a7f651410672f7485130b4cd962a2",
+    "content-hash": "0c171d3b66a8475134d8ffad35618cf2",
     "packages": [
         {
             "name": "composer/package-versions-deprecated",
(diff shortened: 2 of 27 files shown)


theofidry:
>With all this deferred execution the program is slightly less easy to follow. There's no remedy.

Could be easier to follow if we were to get rid of Mutant? (https://github.com/infection/infection/pull/1209)

I can take some time to rebase the PR this weekend if necessary, but I would be curious to see if that can help out with those performance gains as well

sanmai:
> Could be easier to follow if we were to get rid of Mutant? (#1209)

There's no contradiction here. If we'd get rid of Mutant, we'll still have to manage diffing and pretty-printing (#1209 does that in the same order, e.g. long before use). And therefore if we'd do that in a deferred way, there will be the same readability issue.

If we'd consider that a Mutant, or what is in place of it, could do none of the diffing and printing, but instead pass around ASTs, we'll no longer have this problem, but that's a huge responsibility skew with quite a blast radius. I'm not so certain it's going to be worth it, but neither I can say it's a bad idea.

theofidry:
>(#1209 does that in the same order, e.g. long before use)

Yes I recall that, but actually could easily be done lazily as you mentioned although this comes at the cost of holding the whole AST in-memory for longer

I did not mean that there is contradiction, just that there a big overlap and that maybe your PR would look a lot simpler _after_ #1209. But if you don't think so I guess I'll have to take on myself to do a more complex rebase :P

That said I would like us to move forward a bit on #1209 because I don't want to have to rebase it if we don't want that change at all...

sanmai:
> I'll have to take on myself to do a more complex rebase

Well, I feel your pain. Sorry about that!

maks-rafalko (approved):
I don't think #1209 should block anything here. It's

* already stale for a half of the year
* arguable (I still don't agree with the renamings), we don't have an agreement there

sanmai:
Thank you @maks-rafalko

theofidry:
Hm apparently I forgot to push my comment.

>I don't think #1209 should block anything here. It's
>
>already stale for a half of the year
>arguable (I still don't agree with the renamings), we don't have an agreement there

Outch that hurts :(

Anyway I'll try to see if there is anything that can be extracted. I don't want to spend hours refactoring a PR that is not going to be merged

maks-rafalko:
> Outch that hurts :(

didn't mean to hurt you, sorry! But those 2 points are what I see.

PR #1422, opened 2020-11-09
Lazify ProxyTrace with Deferred

sanmai:
This PR:

- [x] Attempts to lazify ProxyTrace [with Deferred](https://github.com/sanmai/later#use)
- [x] Covered by tests

`sanmai/later` is a small, fully type-transparent library (with generics, under PHPStan/Psalm/Phan) from yours truly (full disclosure), invented this spring during the original work on ProxyTrace. I let it lay for several months, used here and there, and I think it's time for a wider adoption. It is quite easy to lazify/deferify almost everything with this library. What is ongoing in #1420.

In this particular case we employed about the same pattern before, therefore, outside of having to manage less code, there will be no speed bump, but I envy #1420 comes with at least some improvement in this regards (and _there is_ some!).

### DIFF
--- a/src/TestFramework/Coverage/ProxyTrace.php
+++ b/src/TestFramework/Coverage/ProxyTrace.php
@@ -36,6 +36,7 @@ declare(strict_types=1);
 namespace Infection\TestFramework\Coverage;
 
 use Infection\TestFramework\Coverage\XmlReport\TestLocator;
+use Later\Interfaces\Deferred;
 use Symfony\Component\Finder\SplFileInfo;
 use Webmozart\Assert\Assert;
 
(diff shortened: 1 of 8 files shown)


src/TestFramework/Coverage/ProxyTrace.php
```diff
@@ -96,25 +95,11 @@ public function hasTests(): bool
...
-            $this->testLocations = $testLocations;
-
-            break;
-        }
-
-        Assert::isInstanceOf($this->testLocations, TestLocations::class);
-        $this->lazyTestLocations = []; // let GC have it
-
-        return $this->testLocations;
+        return new TestLocations();
```
sanmai, line 102: It should be possible to make it return `null`.

maks-rafalko (approved):

PR #1423, opened 2020-11-10
Optimize NodeMutationGenerator

sanmai:
This PR:

- [x] Optimizes `NodeMutationGenerator` to run fastest checks first
- [x] Covered by tests
- [x] [Blackfire diff](https://blackfire.io/profiles/compare/7339bfb2-76f7-4c7b-9023-af765e8b7358/graph) shows 12% time reduction.

 `NodeMutationGenerator` is one of the hottest places we have. Infections spends about _two thirds_ of the time in it. And it operates with O(n<sup>2</sup>) complexity, which we can't really do anything about (best guess). Therefore any small optimization we can make is worth it.

<details>
  <summary>Cachegrind reports</summary>

Here's a screenshot of a Cachegrind report from before this change:

NodeMutationGenerator

And here's after:

NodeMutationGenerator2

</details>

### DIFF
--- a/src/Mutator/NodeMutationGenerator.php
+++ b/src/Mutator/NodeMutationGenerator.php
@@ -37,6 +37,7 @@ namespace Infection\Mutator;
 
 use function count;
 use function get_class;
+use Infection\AbstractTestFramework\Coverage\TestLocation;
 use Infection\Mutation\Mutation;
 use Infection\PhpParser\MutatedNode;
 use Infection\PhpParser\Visitor\ReflectionVisitor;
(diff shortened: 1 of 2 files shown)


sidz (approved):
LGTM

sanmai:
Thank you @sidz

src/Mutator/NodeMutationGenerator.php
```diff
@@ -113,22 +130,7 @@ private function generateForMutator(Node $node, Mutator $mutator): iterable
...
-
-        $tests = $this->trace->getAllTestsForMutation(
-            $this->lineRangeCalculator->calculateRange($node),
-            $isOnFunctionSignature
-        );
-
-        if ($tests instanceof Traversable) {
-            $tests = iterator_to_array($tests, false);
-        }
+        $tests = $this->getAllTestsForMutation();
```
theofidry, line 133: under what circumstances is this called multiple times?

sanmai: We can have multiple mutations for a give node, can't we?

sanmai: This probably be best renamed to something like `getAllTestsForNode`

PR #1425, opened 2020-11-11
Refactor ResultsCollector out of MetricsCalculator

sanmai:
This PR:

- [x] Refactors `ResultsCollector` out of `MetricsCalculator`
- [x] Covered by tests

It's a lot of changed lines, but the bottom line is that :
- We had `MetricsCalculator` doing several jobs at once, and now we have `ResultsCollector` and `MetricsCalculator` doing own parts of their jobs. 
- And there's `MutationTestingResultsCollectorSubscriber` and `MutationTestingConsoleLoggerSubscriber` doing collecting and reporting, instead of the latter doing two jobs at once.

This will let us use different strategies for result accumulation. E.g. if none of the loggers expect uncovered mutants, we can drop them, and likewise can be said about killed mutants. Most of the time they're just noise. Yet final counts will have all and every mutant counted, just like before.

I don't expect any substantial performance or memory consumption change here.

~~Blocked by #1420~~


sanmai:
@theofidry this is a follow-up to #1420, that's why it basically includes all of #1420.

src/Container.php
```diff
@@ -469,11 +475,19 @@ public static function create(): self
                     $container->getConfiguration()->noProgress()
                 );
             },
+            MutationTestingResultsCollectorSubscriberFactory::class => static function (self $container): MutationTestingResultsCollectorSubscriberFactory {
+                return new MutationTestingResultsCollectorSubscriberFactory(
+                    $container->getMetricsCalculator(),
+                    // TODO ResultsCollector can have multiple strategies on results accumulations, e.g. whenever uncovered results need saving
+                    $container->getResultsCollector()
```
sanmai, line 482: Here's the place to inject something, that will stand between `ResultsCollectorSubscriber` and actual `ResultsCollector` doing the filtering of unneeded results.

theofidry:
I can't look into this PR in details but I'm 👍  on the theory, it is indeed an excellent way to reduce unnecessary metrics processing depending on what loggers are used.

src/Event/Subscriber/MutationTestingResultsCollectorSubscriber.php
```diff
@@ -0,0 +1,62 @@
...
+
+use Infection\Event\MutantProcessWasFinished;
+use Infection\Metrics\Collector;
+
+/**
+ * @internal
+ */
+final class MutationTestingResultsCollectorSubscriber implements EventSubscriber
+{
+    /** @var Collector[] */
```
maks-rafalko, line 46:
```suggestion
    /** @var array<int, Collector> */
```

?

src/Event/Subscriber/MutationTestingResultsCollectorSubscriberFactory.php
```diff
@@ -0,0 +1,60 @@
...
+
+use Infection\Metrics\Collector;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @internal
+ */
+final class MutationTestingResultsCollectorSubscriberFactory implements SubscriberFactory
+{
+    /** @var Collector[] */
```
maks-rafalko, line 46:
```suggestion
    /** @var array<int, Collector> */
```

tests/phpunit/Logger/JsonLoggerTest.php
```diff
@@ -210,12 +214,13 @@ public function metricsProvider(): iterable
...
             false,
-            $this->createUncoveredMetricsCalculator(),
+            new MetricsCalculator(2),
+            $this->createUncoveredResultsCollector(),
             [
                 'stats' => [
-                    'totalMutantsCount' => 1,
+                    'totalMutantsCount' => 0,
                     'killedCount' => 0,
-                    'notCoveredCount' => 1,
```
maks-rafalko, line 218: could you please explain why the stats was changed?

src/Event/Subscriber/MutationTestingResultsCollectorSubscriber.php
sanmai: We're not using the key anywhere, so more appropriate definition would be `array<Collector>`. It is also saves us from the trouble declaring the array-key type for PHPStan up the stream. And, well, `array<Collector>` is same as `Collector[]`.

src/Metrics/MetricsCalculator.php
```diff
@@ -44,28 +43,14 @@
...
-
-    private int $escapedCount = 0;
-
-    private int $timedOutCount = 0;
-
-    private int $notCoveredByTestsCount = 0;
+    /**
+     * @var array<string, int>
+     */
+    private array $countByStatus = [];
```
sanmai, line 53: Contrary to the above, the array-key type is significant here. We'll be using it and comparing against strings.

tests/phpunit/Logger/JsonLoggerTest.php
sanmai: I got it back.

src/Event/Subscriber/MutationTestingResultsCollectorSubscriber.php
sanmai: Again, key-type is insignificant here.

maks-rafalko: but what kind of integer keys are there? If they are natural numbers, we can use `list<Collector>`, if they are integers with "wholes", it would be better to use `array<int, Collector>`

IMO, `int` is better than implicit `array-key`

tests/phpunit/Logger/JsonLoggerTest.php
```diff
@@ -246,12 +252,13 @@ public function metricsProvider(): iterable
 
         yield 'Non UTF-8 characters' => [
             false,
-            $this->createNonUtf8CharactersCalculator(),
+            new MetricsCalculator(2),
+            $this->createNonUtf8CharactersCollector(),
             [
                 'stats' => [
-                    'totalMutantsCount' => 1,
```
maks-rafalko, line 252: same question, why was it changed? (I mean probably it's ok, I just didn't get the change)

maks-rafalko (approved):

src/Event/Subscriber/MutationTestingResultsCollectorSubscriber.php
sanmai: What I mean is that the API of `MutationTestingResultsCollectorSubscriber` does not care about keys.

`iterable<Collector>` could work.

tests/phpunit/Logger/JsonLoggerTest.php
sanmai: It tests against empty `MetricsCalculator`, and it does that because it's a test of an edge-case of `JsonLogger`. We do not actually care about these numbers here.

sanmai:
Thank you @maks-rafalko

PR #1426, opened 2020-11-11
Improve MutationTestingRunner

sanmai:
This PR:

- [x] Changes `MutationTestingRunner` to use a more direct `cast()` instead of `map()`; should save a few cycles on Generator checks inside `map()`
- [x] Adds a TODO note

### DIFF
--- a/src/Process/Runner/MutationTestingRunner.php
+++ b/src/Process/Runner/MutationTestingRunner.php
@@ -101,7 +101,7 @@ class MutationTestingRunner
         $this->eventDispatcher->dispatch(new MutationTestingWasStarted($numberOfMutants));
 
         $processes = take($mutations)
-            ->map(function (Mutation $mutation): Mutant {
+            ->cast(function (Mutation $mutation): Mutant {
                 return $this->mutantFactory->create($mutation);
             })
             ->filter(function (Mutant $mutant) {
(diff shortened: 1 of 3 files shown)


src/Process/Runner/MutationTestingRunner.php
```diff
@@ -132,6 +132,7 @@ public function run(iterable $mutations, string $testFrameworkExtraOptions): voi
                 return true;
             })
             ->filter(function (Mutant $mutant) {
+                // TODO refactor this comparison into a dedicated comparer to make it possible to swap strategies
```
sanmai, line 135: For example, if a time limit set to 25 seconds, but a whole test suite takes only 20 seconds, we can tell that any set of tests will take less than the timeout. Therefore we can safely skip this check.

maks-rafalko: > For example, if a time limit set to 25 seconds, but a whole test suite takes only 20 seconds, we can tell that any set of tests will take less than the timeout. Therefore we can safely skip this check.

where we can get this information from, not calling `getNominalTestExecutionTime()`?

maks-rafalko (approved):

src/Process/Runner/MutationTestingRunner.php
sanmai: From the actual time spent on the initial test run, or from the same junit report. There are multiple opportunities.

sanmai:
Thank you @maks-rafalko

src/Process/Runner/MutationTestingRunner.php
theofidry: 👍  I had some along those lines in mind as well

PR #1427, opened 2020-11-12
Address NodeMutationGenerator review

sanmai:
`getAllTestsForCurrentNode()` is not concerned with a mutator or mutation, only with the current node.

Follow-up to #1423

### DIFF
--- a/src/Mutator/NodeMutationGenerator.php
+++ b/src/Mutator/NodeMutationGenerator.php
@@ -130,7 +130,7 @@ class NodeMutationGenerator
             );
         }
 
-        $tests = $this->getAllTestsCurrentNode();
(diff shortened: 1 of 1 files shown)


src/Mutator/NodeMutationGenerator.php
```diff
@@ -169,7 +169,7 @@ private function isInsideFunction(): bool
     /**
      * @return TestLocation[]
      */
-    private function getAllTestsForMutation(): array
+    private function getAllTestsCurrentNode(): array
```
sanmai, line 172: It is not concerned with a mutator or mutation, only with the current node.

sidz (approved):

sanmai:
Thank you @sidz

PR #1430, opened 2020-11-18
Introduce FilteringResultsCollector

sanmai:
This PR:

- [x] Adds `FilteringResultsCollector` that excludes unneeded mutation results depending on logging/debugging configuration.
- [x] Covered by tests
- [x] Benchmarks
  - From current `master` build [there's a 66% memory reduction](https://blackfire.io/profiles/compare/9c9a7a2e-231f-46be-85df-76c710eb95e5/graph).
  - From 0.20.1 [a diff shows](https://blackfire.io/profiles/compare/cc3fb3b9-0d33-4406-82a3-65009dfcaf0c/graph) -46% time reduction, -23% memory reduction.

I'm not too happy about `TargetDetectionStatusesProvider` because it does what individual loggers should provide, but that should do it for now. Changing the loggers to supply this information to `TargetDetectionStatusesProvider` will require fairly substantial refactoring with otherwise uncertain benefits.

Follow-up to #1425 

<details>
  <summary>Blackfire is a bit flacky, here are the screenshots</summary>
  
vs current master
vs 0.20.1
  
</details>

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -78,9 +78,11 @@ use Infection\FileSystem\SourceFileFilter;
 use Infection\FileSystem\TmpDirProvider;
 use Infection\Logger\GitHub\GitDiffFileProvider;
 use Infection\Logger\LoggerFactory;
+use Infection\Metrics\FilteringResultsCollector;
 use Infection\Metrics\MetricsCalculator;
 use Infection\Metrics\MinMsiChecker;
 use Infection\Metrics\ResultsCollector;
+use Infection\Metrics\TargetDetectionStatusesProvider;
 use Infection\Mutant\MutantCodeFactory;
 use Infection\Mutant\MutantExecutionResultFactory;
 use Infection\Mutant\MutantFactory;
@@ -478,8 +480,10 @@ final class Container
             MutationTestingResultsCollectorSubscriberFactory::class => static function (self $container): MutationTestingResultsCollectorSubscriberFactory {
                 return new MutationTestingResultsCollectorSubscriberFactory(
                     $container->getMetricsCalculator(),
-                    // TODO ResultsCollector can have multiple strategies on results accumulations, e.g. whenever uncovered results need saving
-                    $container->getResultsCollector()
+                    new FilteringResultsCollector(
+                        $container->getResultsCollector(),
+                        $container->getTargetDetectionStatusesProvider()->get()
+                    )
                 );
             },
             MutationTestingConsoleLoggerSubscriberFactory::class => static function (self $container): MutationTestingConsoleLoggerSubscriberFactory {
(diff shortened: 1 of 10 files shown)


src/Container.php
```diff
@@ -536,6 +541,18 @@ public static function create(): self
                     $container->getLogger()
                 );
             },
+            TargetDetectionStatusesProvider::class => static function (self $container): TargetDetectionStatusesProvider {
+                $config = $container->getConfiguration();
+
+                return new TargetDetectionStatusesProvider(
+                    $container->getConfiguration()->getLogs(),
```
sanmai, line 548:
```suggestion
                    $config->getLogs(),
```

theofidry (approved):
I'm 👍. Again I cannot focus on the code details atm but the general idea looks good to me.

I however have one reserve but I think it's your plans anyway: I don't think the current `TargetDetectionStatusesProvider` make any sense.

Its role – if I understand correctly, is to provide which status to filter by to the filter for the collector, collector that is tailored for the loggers.

I guess we can add:

```php
interface MutationTestingResultsLogger {
  // ...

  /**
   * @return string[]
   */
  getUsedDetectionStatuses(): array;
}
```

So the logger registry can combine all of them without having to know any details. However the tricky part is to achieve that since we are creating the loggers lazily by passing the factory to the subscriber...

So maybe we will need a LoggerProxy which takes the factory and instantiate the underlying loggers underneath. This way both the subscriber and the provider can consume this proxy (which can implement the interface of the logger even).

I'm also wondering about the name, what do you think of renaming `TargetDetectionStatusesProvider` to `LoggerDetectionStatusDetector`?

I also wonder if it would make sense to start documenting this part of our application, this is starting to be a complex piece of the app with a lot of indirections...

maks-rafalko (approved):
:ok_hand:

sanmai:
@theofidry I'm 100% on board with you on the state of `TargetDetectionStatusesProvider`. I tried to go this `getUsedDetectionStatuses()` route, but found, like you said, even if some of the loggers would implement such method, it is difficult to get to the both because they're created after the fact, and because they're behind several layer of abstraction.

> I'm also wondering about the name, what do you think of renaming `TargetDetectionStatusesProvider` to `LoggerDetectionStatusDetector`?

This is a certainly more appropriate name considering the current state of affairs, but I named it such because it is supposed to feed from loggers. It is not supposed to _detect_ anything in a perfect world. What do you think about `LoggerDetectionStatusProvider`?

I'll update the code to link to your review.

sanmai:
I'm going to merge this for now, to see if I can do more in this direction in a next PR.

theofidry:
>It is not supposed to detect anything, in a perfect world. What do you think about LoggerDetectionStatusProvider?

I was seeing more as "detect/find out what detection status the loggers are gonna use"

>like you said, even if some of the loggers would implement such method, it is difficult to get to them both because they're created after the fact, and because they're behind several layer of abstraction.

But with this LoggerProxy we can fix that right?

Issue #1433, opened 2020-11-20
Both global-ignore and global-ignoreSourceCodeByRegex cannot be set in infection.json file at the same time


lapisangularis:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.20.1
| Test Framework version | PHPUnit 8.5.9
| PHP version | 7.4.11
| Platform    | Linux 6a8c54c6eabc 4.14.116-boot2docker
| Playground | https://infection-php.dev/r/7l7

When setting up both `global-ignoreSourceCodeByRegex` and `global-ignore` settings in `infection.json.dist` file, the second one in order is producing `The profile or mutator "global-ignoreSourceCodeByRegex||global-ignore" was not recognized.` error.

I believe the problem is that there is a `break` statement in MutatorResolver, so the second option is not unset from config, and then treated as a profile or mutator.

<details>
 <summary>phpunit.xml</summary>

 ```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit backupGlobals="false"
         backupStaticAttributes="false"
         bootstrap="vendor/autoload.php"
         colors="true"
         convertErrorsToExceptions="true"
         convertNoticesToExceptions="true"
         convertWarningsToExceptions="true"
         processIsolation="false"
         stopOnFailure="false">
    <testsuites>
        <testsuite name="Unit">
            <directory suffix="Test.php">./tests/Unit</directory>
        </testsuite>

        <testsuite name="Integration">
            <directory suffix="Test.php">./tests/Integration</directory>
        </testsuite>

        <testsuite name="Feature">
            <directory suffix="Test.php">./tests/Feature</directory>
        </testsuite>
    </testsuites>
    <filter>
        <whitelist processUncoveredFilesFromWhitelist="true">
            <directory suffix=".php">./src</directory>
            <exclude>
                <directory suffix="routes.php">./src</directory>
            </exclude>
        </whitelist>
    </filter>
</phpunit>
 ```
</details>

<details>
 <summary>Output with issue</summary>

 ```
bash-5.0$ vendor/bin/infection

In MutatorResolver.php line 113:

  The profile or mutator "global-ignoreSourceCodeByRegex" was not recognized.


[...]

PR #1434, opened 2020-11-20
#1433 Allow to set both global settings in config file at the same time

lapisangularis:
This PR:

- [x] Fixes an issue, when both `global-` settings are set in config, which threw an exception before this fix
- [x] Covered by tests
- [x] Fixes https://github.com/infection/infection/issues/1433


maks-rafalko (approved):
Thank you @lapisangularis for fixing this issue :+1:

PR #1436, opened 2020-11-21
Refactor LoggerFactory into BadgeLoggerFactory/FileLoggerFactory

sanmai:
This PR:

- [x] Renames `LoggerRegistry` into `FederatedLogger` (federated is like united, or centralized), logger of loggers (a new `LoggerRegistry` supposed to be added later)
- [x] Refactors `LoggerFactory` into `BadgeLoggerFactory` and `FileLoggerFactory`
- [x] `BadgeLoggerFactory` is responsible only for badge logger, it will not need to interact with result filters etc. from #1430 
- [x] `FileLoggerFactory` does everything `LoggerFactory` did before except for `BadgeLogger`
- [x] Covered by tests

The general idea is to separate pure line-loggers and the badge logger, which isn't their kind. Badge logger does not require to use  `ResultsCollector`, therefore we are not going to include it in a new `LoggerRegistry` [or `LoggerProxy`](https://github.com/infection/infection/pull/1430#issuecomment-731594817). Having BadgeLogger being cared by a dedicated factory simplifies our work down the line, and the implementations of `LoggerFactory` too.


theofidry:
I have very mixed feelings on this one:

-  "Federated" sounds very weird to me and unlike Registry, Chain, Proxy, I've never seen it used
-  Regarding splitting `LoggerFactory` into `BadgeLoggerFactory` and `FileLoggerFactory`, why not. One one hand creating a badge factory is simpler so moving it out shows that its factory is very simple, but that also means we start to have dedicated factories for the different loggers and atm at least, I'm not sure if this other indirection is worth it. But again, no strong feeling there

>Badge logger does not require to use ResultsCollector, therefore we are not going to include it in a new LoggerRegistry or LoggerProxy

I'm wondering though, does that help at all? If we take a look at the logger interface and logger registry they are dead simple and do not care about the result collector at all. So I'm not sure introduce another abstraction or segmenting further the loggers is really worth it and IMO it is fine to have a logger that use the result collector and another that doesn't. At the end of the day you will get a detection status provider of some kind that will check what results is being used and the one that doesn't use the result collector simply won't use any.

So I feel like the logging process is already quite complex between the subscribers, knowing what is instantiated when and the proxy is going into the direction of simplifying things a bit from a contributor PoV (you can focus more on the logger itself and registering it, less about when it's instantiated, caring about the subscriber & co.). But I don't feel that this PR improves things in that regard

sanmai:
> So I feel like the logging process is already quite complex between the subscribers, knowing what is instantiated when and the proxy is going into the direction of simplifying things a bit from a contributor PoV (you can focus more on the logger itself and registering it, less about when it's instantiated, caring about the subscriber & co.). But I don't get the impression that this PR improves things in that regard

Indeed. What this does is moves `BadgeLogger` out of the way for this future simplification you're mentioning. Have a look at the old implementation of the `LoggerFactory`. If even had a dedicated check for `BadgeLogger`, and it wasn't even following the process for other loggers. 

If we'd keep the `LoggerFactory` the way it was, we'll have to drag `BadgeLogger` all along, checking for it on every action.

theofidry:
But splitting the factory is still doable without having this `FederatedLogger`: we could very well have:

```
return new LoggerRegistry(
    ...array_filter(
        ...$container->getFileLoggerFactory()->createFromLogEntries(
                        $container->getConfiguration()->getLogs()
                    ),
                    $container->getBadgeLoggerFactory()->createFromLogEntries(
                        $container->getConfiguration()->getLogs()
                    ),
    )
);
```

sanmai:
It's the name for the thing you called `LoggerProxy`. What do you think it should do? This logger-proxy-registry thing. 

Registry is something like a place where records are kept. E.g. you can call a registry and get certain information. But we can call it a proxy just as well, since it'll be a stand-in for loggers' method `getUsedDetectionStatuses()`

theofidry:
I'm more thinking of something along the lines of:

```php
MutationTestingResultsLoggerSubscriberFactory::__construct(ProxyLogger $logger) {}

class ProxyLogger implements Logger {
__construct(callable $factory) {
}

function log() {
  $this->logger = $this->logger ?? $this->factory->create();

  $this->logger->log();
}
}
```

and `callable $factory` which is a callable relying on the factories to get the main `LoggerRegistry` instance

theofidry (approved):
Having a deeper look, looks good to me except:

- The renaming from `LoggerRegistry` to `FederatedLogger`: I don't think it's necessary and I don't see an improvement in the naming
- The current `FederatedLogger` service: IMO the service ID should be `MutationTestingResultsLogger` (the implementation differs, but what you want to request is the interface) which later will be an instance of `LoggerProxy` instead – The difference will then be instead of having all the loggers eagerly created like right now, they will be lazily

sanmai:
Good point about the service ID!

sanmai:
> ```
>  function log() {
>       $this->logger = $this->logger ?? $this->factory->create();
> 
>       $this->logger->log();
> }
> ```

Do you mean this factory's result will be memoized?

theofidry:
It's slightly different: notice that the factory callable is a `() => MutationTestingResultsLogger`: it creates the loggers lazily when `log()` is called (which should happen _much_ later). So to create this callable will likely be something like:

```
            MutationTestingResultsLogger::class => static function (self $container): MutationTestingResultsLogger {
                $factory = () => new LoggerRegistry(...array_filter([
                    $container->getFileLoggerFactory()->createFromLogEntries(
                        $container->getConfiguration()->getLogs()
                    ),
                    $container->getBadgeLoggerFactory()->createFromLogEntries(
                        $container->getConfiguration()->getLogs()
                    ),
                ]);

                return new LoggerProxy($factory);
            },
```

sanmai:
Thank you @theofidry

PR #1437, opened 2020-11-21
Allow E2E tests to pass on PHP 8

sanmai:
One of the E2E tests on PHP 8 is failing [with one of the mutations being uncovered](https://github.com/infection/infection/runs/1435444637#step:13:316), where for all other PHP versions it is covered. 

I don't think it's other than a PHP 8 pre-release bug.


maks-rafalko (approved):

maks-rafalko:
Thank you @sanmai

PR #1440, opened 2020-11-25
Add ConcatOperandRemoval mutator (fixes #1439)

### LINKED ISSUE #1439: Mutator suggestion: String concatenation operator
Is there a reason why there is no mutator that mutates `$x = 'foo' . 'bar'` to `$x = 'foo'` and `$x = 'bar'` (and maybe also to `$x = 'bar' . 'foo'`). Seems like a pretty obvious thing to me. Here's a use case:

```php
final class PrefixingStorage implements StorageInterface
{
    private string $prefix;
    private StorageInterface $innerStorage;

    public function __construct(string $prefix, StorageInterface $innerStorage)
    {
        $this->prefix = $prefix;
        $this->innerStorage = $innerStorage;
    }

    public function assign(string $key, string $value): void
    {
        $this->innerStorage->assign($this->prefix . $key, $value); // Is $this->prefix actually used?
    }

    public function get(string $key): ?string
    {
        return $this->innerStorage->get($this->prefix . $key); // Is $this->prefix actually used?
    }
}
```

I'd be happy to make my first code contribution to this project if you think this is something that Infection should be able to do.

MidnightDesign:
This PR:

- [x] Adds a new `ConcatOperandRemoval` mutator that mutates `'foo' . 'bar'` to `'foo'` and `'bar'`.
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/199

fixes #1439

### DIFF
--- a/src/Mutator/Removal/ConcatOperandRemoval.php
+++ b/src/Mutator/Removal/ConcatOperandRemoval.php
@@ -35,7 +35,6 @@ declare(strict_types=1);
 
 namespace Infection\Mutator\Removal;
 
-use function assert;
 use Infection\Mutator\Definition;
 use Infection\Mutator\GetMutatorName;
 use Infection\Mutator\Mutator;
@@ -44,6 +43,8 @@ use PhpParser\Node;
 
 /**
  * @internal
+ *
+ * @implements Mutator<Node\Expr\BinaryOp\Concat>
  */
 final class ConcatOperandRemoval implements Mutator
 {
@@ -84,8 +85,6 @@ TxT
 
     /**
      * @psalm-mutation-free
-     *
-     * @param Node\Expr\BinaryOp\Concat $node
      */
     public function mutate(Node $node): iterable
     {
--- a/src/Mutator/Removal/ConcatOperandRemoval.php
+++ b/src/Mutator/Removal/ConcatOperandRemoval.php
@@ -53,7 +53,7 @@ final class ConcatOperandRemoval implements Mutator
     public static function getDefinition(): ?Definition
     {
         return new Definition(
-            <<<'TXT'
+            <<<'TxT'
 Removes an operand from a string concatenation.
 
 ```php
--- a/src/Mutator/Removal/ConcatOperandRemoval.php
+++ b/src/Mutator/Removal/ConcatOperandRemoval.php
@@ -1,0 +1,96 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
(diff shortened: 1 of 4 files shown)


tests/phpunit/Mutator/Removal/ConcatOperandRemovalTest.php
```diff
@@ -0,0 +1,76 @@
...
+    {
+        $this->doTest($input, $expected);
+    }
+
+    public function mutationsProvider(): iterable
+    {
+        yield 'Removes both operands' => [
+            <<<'PHP'
+<?php
+'foo' . 'bar';
```
sanmai, line 57: How about `$a . $b . $c`?

src/Mutator/Removal/ConcatOperandRemoval.php
```diff
@@ -0,0 +1,102 @@
...
+    {
+        return $node instanceof Node\Expr\BinaryOp\Concat;
+    }
+
+    /**
+     * @psalm-mutation-free
+     */
+    public function mutate(Node $node): iterable
+    {
+        assert($node instanceof Node\Expr\BinaryOp\Concat);
```
sanmai, line 90: I think this should be `Assert::isInstanceOf($node, Node\Expr\BinaryOp\Concat::class);`

sanmai: `use Webmozart\Assert\Assert;`

MidnightDesign: I'd keep the `assert()` since perf penalty in production. We're not really checking the type on other mutators as well. What do you think? (Ideally, the `Mutator` interface would be `@template`d and the type of the argument would be the template type.)

sanmai: Aye, you can use just `@param Node\Expr\BinaryOp\Concat $node`

MidnightDesign: Just came to realize that "production" is silly when talking about a development tool. But still, I think that if we just want to help out the static analyzers (in this case PHPStan), we should use `assert()` instead of "real" code.

But I'd be happy to use the `Assert` class if you have a different opinion. Just wanted to share mine.

sanmai: I see that virtually every `::mutate()` uses `@param` to deal with the issue.

src/Mutator/Removal/ConcatOperandRemoval.php
```diff
@@ -0,0 +1,102 @@
...
+        );
+    }
+
+    public function canMutate(Node $node): bool
+    {
+        return $node instanceof Node\Expr\BinaryOp\Concat;
+    }
+
+    /**
+     * @psalm-mutation-free
```
sanmai, line 86:
```suggestion
     * @psalm-mutation-free
     *
     * @param Node\Expr\BinaryOp\Concat $node
```

MidnightDesign: PHPStan will (righfully) complain about that. That's what I wanted to avoid by using `assert()`.

src/Mutator/Removal/ConcatOperandRemoval.php
MidnightDesign: Yes, and they're all in PHPStan's ignore list. I wanted to avoid adding it there because there's a reason PHPStan complains about it.

src/Mutator/Removal/ConcatOperandRemoval.php
sanmai: It is very peculiar that it does.

sanmai: Right, so we mute this warning for all other mutators in `phpstan-src-baseline.neon`

MidnightDesign: Well, `Mutator#mutate()` says it accepts `Node`. An implementation can't just narrow that type because that breaks the interface's contract.

src/Mutator/Removal/ConcatOperandRemoval.php
MidnightDesign: Again, I think the "real" solution would be to add `@template TNode` to `Mutator` and change the argument type to `TNode` (but keep the type `Node` on `canMutate`, of course.)

sanmai: OK, how do you think if we do `make phpstan-baseline` to ignore the issue, use `@param` here, and deal with this later in a centralized fashion?

sanmai: You can ignore changes to `phpstan-tests-baseline.neon` for time being.

sanmai: I think PHPStan is overthinking here. We didn't change the input type ([that'll be a PHP error](https://3v4l.org/TqNho)), but rather we use `@param` point out exactly which subtype we know we will have here. There's no need to do any kind of assertion because we ensure we get the type we expect in `canMutate()`

src/Mutator/Removal/ConcatOperandRemoval.php
```diff
@@ -0,0 +1,102 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Mutator\Removal;
+
+use function assert;
```
sanmai, line 38:
```suggestion
```

sanmai (approved):
LGTM

sanmai:
I'll let @maks-rafalko make another review because I might have a reason to be biased here. And thank you!

maks-rafalko (approved):
:+1: 

thanks for your contribution @MidnightDesign!

src/Mutator/Removal/ConcatOperandRemoval.php
```diff
@@ -0,0 +1,101 @@
...
+ * @internal
+ */
+final class ConcatOperandRemoval implements Mutator
+{
+    use GetMutatorName;
+
+    public static function getDefinition(): ?Definition
+    {
+        return new Definition(
+            <<<'TxT'
```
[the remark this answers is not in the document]
MidnightDesign: I copied that from another class

maks-rafalko:
Thank you @MidnightDesign

PR #1453, opened 2020-12-06
Limit testing to PHPUnit <9.5

sanmai:
The root issue in PHPUnit-Bridge has been fixed already (https://github.com/symfony/symfony/issues/39329), but not in a release yet. 

We can remove this limitation once PHPUnit-Bridge >5.2.0 comes out.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "938c5a4a9fd0101be4745052152f4d44",
+    "content-hash": "f2ab9038b438fb8f1ff8fdc10932927f",
     "packages": [
         {
             "name": "composer/xdebug-handler",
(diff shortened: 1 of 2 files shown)


composer.lock
```diff
@@ -288,16 +288,16 @@
         },
         {
             "name": "nikic/php-parser",
-            "version": "v4.10.2",
+            "version": "v4.10.3",
```
sidz, line 291: this change will break some our tests. I've already fixed our tests but didn't push code.

sidz (approved):

composer.lock
sidz: fixed in https://github.com/infection/infection/pull/1454

sanmai, line 291:
```suggestion
            "version": "v4.10.2",
```
OK, let's try this.

composer.lock
```diff
@@ -292,12 +292,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/nikic/PHP-Parser.git",
-                "reference": "658f1be311a230e0907f5dfe0213742aff0596de"
+                "reference": "dbe56d23de8fcb157bbc0cfb3ad7c7de0cfb0984"
```
sidz, line 295: reference should be changed as well.

maks-rafalko (approved):
fyi #1454 has been merge

Issue #324, opened 2018-04-28
Implement Regex Mutator


maks-rafalko:
@kalessil suggested a set of interesting mutators for Regular Expressions.

```diff
- preg_match("/test/i", $string));
+ preg_match("/test/", $string));

- preg_match("/test/g", $string));
+ preg_match("/test/", $string));

- preg_match("/^test$/", $string));
+ preg_match("/^test/", $string));
+ preg_match("/test$/", $string));
```

and so on.
Original message:

> E.g. add/remove i-flag, `\d,\w\s` <-> `\D,\W\S`, `{<number>[, {number}]}` - same as hardcoded numbers mutations, `.*` <-> `.+`

thoughts?

-----

UPD: now it's clear that we should implement it https://twitter.com/infection_php/status/990136566726189057

ntzm:
How about removing `preg_quote` calls?

kalessil:
For `preg_quote` will be enough to add/modify 2nd parameter with as `''` (when diffs will be more informative about intention).

BackEndTea:
> For preg_quote will be enough to add/modify 2nd argument as '' (then diffs will be more informative about intention).

From what i can tell this won't work
```php
var_dump(preg_quote('\ab?'));
var_dump(preg_quote('\ab?', ''));
```
yields:
```
string(6) "\\ab\?"
string(6) "\\ab\?"
```

adeptofvoltron:
not to duplicate someone's work
yesterday I just to work on murators:

![image](https://user-images.githubusercontent.com/6844275/41502059-73f7f564-71b1-11e8-8442-caa9f4bd3186.png)
![image](https://user-images.githubusercontent.com/6844275/41502061-788859ac-71b1-11e8-9bfb-9e5de52f67aa.png)

they both work similar whey...If carret is present in regex  it delete one...otherway it put one in to the regex.
Same with ending dollar

I would try to make PR today or tommorow. Regards

maks-rafalko:
other ideas to implement: https://github.com/stryker-mutator/stryker-net/blob/master/docs/RegexMutators.md

PR #388, opened 2018-06-17
Preg match regex mutators

adeptofvoltron:
This PR:

- [x] Adds new feature ...
- [x] Covered by tests
- [x] Fixes #324
- [ ] Documentation PR (like at https://github.com/infection/site/pulls)

These mutators works only if in `preg_match` method regex string is passed as an mutator.
to make it work also with variables, or some concat I need to learn more about ATS.

PR #1455, opened 2020-12-07
Preg match regex mutators #2

maks-rafalko:
This PR is a rebirth of https://github.com/infection/infection/pull/388

Fixes https://github.com/infection/infection/issues/324

- took @adeptofvoltron's work
- fixed all comments and adopted to the current code base
- added more tests
- [x] Doc PR: https://github.com/infection/site/pull/203

2 new mutations here:

```diff
# 1
- preg_match("/^test$/", $string));
+ preg_match("/^test/", $string));

# 2
- preg_match("/^test$/", $string));
+ preg_match("/test$/", $string));
```

Other possible mutations will be done in a separate PRs after this one is merged.


psalm-baseline.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files psalm-version="4.1.1@16bfbd9224698bd738c665f33039fade2a1a3977">
+  <file src="src/Mutator/Regex/PregMatchRemoveCaret.php">
+    <ImpureFunctionCall occurrences="1">
```
maks-rafalko, line 4: Unfortunately, `thecodingmachine/safe` package does not support Psalm's annotations yet, and functions are missing `@pure` annotation, that's why we have issues reported by Psalm.

Tried https://github.com/hectorj/safe-php-psalm-plugin, but it doesn't work properly, so for now it's ok to suppress them

src/Mutator/Regex/AbstractPregMatch.php
```diff
@@ -0,0 +1,107 @@
...
+
+        yield new FuncCall($node->name, [$newArgument] + $arguments, $node->getAttributes());
+    }
+
+    public function canMutate(Node $node): bool
+    {
+        return $node instanceof FuncCall
+            && $node->name instanceof Node\Name
+            && strtolower((string) $node->name) === 'preg_match'
+            && $node->args[0]->value instanceof Node\Scalar\String_
```
BackEndTea, line 75: How will this handle an unpacked array as an argument? Kind of an edge case, but may be worth making sure it doesn't crash on that.,

e.g.
```php
\preg_match(...func_get_args());
```

maks-rafalko: added a test to cover this case

only raw strings are mutated for now, with an unpacked array this mutator will do nothing

BackEndTea: 👍

src/Mutator/Regex/PregMatchRemoveCaret.php
```diff
@@ -0,0 +1,93 @@
...
+        $flags = $matches[4] ?? '';
+
+        return $delimiter . $regexBody . $delimiter . $flags;
+    }
+
+    protected function isProperRegexToMutate(string $pattern): bool
+    {
+        preg_match(self::ANALYSE_REGEX, $pattern, $matches);
+
+        return array_key_exists(2, $matches) && $matches[2] === '^';
```
BackEndTea, line 91: prehaps this could be simplified to `($matches[2] ?? null) === '^'`

BackEndTea (approved):
1 nitpick

maks-rafalko:
Thank you @BackEndTea

PR #1456, opened 2020-12-07
Use different folders for PHPStan cache for src and tests folders

maks-rafalko:
without this patch, `make phpstan` overrides just written cache and always insects all files

with this patch, 2 different phpstan runs for src and tests folders use their own cache folders, so cache feature is used correctly


sidz (approved):
:+1:

maks-rafalko:
Thank you @sidz

PR #1460, opened 2020-12-12
Use unique values in `ignoreSourceCodeByRegex` to not waste resources

maks-rafalko:
Fixes #1459

Issue #1459, opened 2020-12-11
Consolidate $ignoreSourceCodeMutatorsMap


theofidry:
![Screen Shot 2020-12-11 at 20 51 30](https://user-images.githubusercontent.com/5175937/101948665-bb928c00-3bf2-11eb-9111-b409ecc6a337.png)

I think we could do a `array_unique()` on it: the less expressions we have to evaluate the better

PR #1461, opened 2020-12-13
Add schema check to not allow duplicates in ignore and ignoreSourceCodeByRegex config options

sidz:
Add schema check and do not allow to have duplicates in **ignore** and **ignoreSourceCodeByRegex** config options.

image

cover https://github.com/infection/infection/issues/1459 request.

I'd say that we don't need to have https://github.com/infection/infection/pull/1460 with this fix. But in case to have a double check we may leave fix from https://github.com/infection/infection/pull/1460 as well.

### DIFF
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -122,6 +122,7 @@
             "properties": {
                 "global-ignore": {
                     "type": "array",
+                    "uniqueItems": true,
                     "items": {
                         "type": "string"
                     }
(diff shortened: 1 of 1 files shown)


maks-rafalko:
It doesn't cover a case where `global-ignoreSourceCodeByRegex` has the same value as `ignoreSourceCodeByRegex` on a mutator value, where after merging we have duplicates

But for simple cases 👍

maks-rafalko (approved):

maks-rafalko:
Thank you @sidz

PR #1458, opened 2020-12-08
Add a describe command

BackEndTea:
This allows users to get a quick idea of a mutator,
without having to leave the cli.

This PR:

- [x] Adds a describe command, which describes mutators
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/204

PR #1464, opened 2021-01-06
Add mutator diffs to Definition object

maks-rafalko:
* Add diffs for all mutators
* Make diff required argument

Follow up for https://github.com/infection/infection/pull/1458

### DIFF
--- a/src/Mutator/Arithmetic/Assignment.php
+++ b/src/Mutator/Arithmetic/Assignment.php
@@ -59,7 +59,11 @@ assignment operators, i.e. `+=`, `*=`, `.=`, etc., with a plain assignment opera
 TXT
             ,
             MutatorCategory::SEMANTIC_REDUCTION,
-            null
+            null,
+            <<<'DIFF'
+- $a += $b;
++ $a = $b;
+DIFF
         );
     }
 
--- a/src/Mutator/Unwrap/UnwrapArrayUintersectAssoc.php
+++ b/src/Mutator/Unwrap/UnwrapArrayUintersectAssoc.php
@@ -76,7 +76,16 @@ $x = ['baz' => 'bar'];
 TXT
             ,
             MutatorCategory::SEMANTIC_REDUCTION,
-            null
+            null,
+            <<<'DIFF'
+- $x = array_uintersect_assoc(
+-     ['foo' => 'bar'],
+-     ['baz' => 'bar'],
+-     $value_compare_func
+- );
++ $x = ['foo' => 'bar'];
++ $x = ['baz' => 'bar'];
+DIFF
         );
     }
 
--- a/src/Command/DescribeCommand.php
+++ b/src/Command/DescribeCommand.php
@@ -38,6 +38,7 @@ namespace Infection\Command;
 use function array_key_exists;
 use function array_keys;
 use Infection\Console\IO;
+use Infection\Mutator\Definition;
 use Infection\Mutator\Mutator;
 use Infection\Mutator\ProfileList;
 use function Safe\sprintf;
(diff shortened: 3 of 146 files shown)


BackEndTea (approved):

sidz (approved):

src/Mutator/Arithmetic/Assignment.php
```diff
@@ -59,7 +59,11 @@ public static function getDefinition(): ?Definition
 TXT
             ,
             MutatorCategory::SEMANTIC_REDUCTION,
-            null
+            null,
+            <<<'DIFF'
+- $a += $b;
++ $a = $b;
+DIFF
```
theofidry, line 66: You can now do:

```php
            <<<'DIFF'
            - $a += $b;
            + $a = $b;
            DIFF
```

src/Mutator/Unwrap/UnwrapArrayUintersectAssoc.php
```diff
@@ -76,7 +76,16 @@ public static function getDefinition(): ?Definition
...
             MutatorCategory::SEMANTIC_REDUCTION,
-            null
+            null,
+            <<<'DIFF'
+- $x = array_uintersect_assoc(
+-     ['foo' => 'bar'],
+-     ['baz' => 'bar'],
+-     $value_compare_func
+- );
++ $x = ['foo' => 'bar'];
```
theofidry, line 86: is that correct? because this could be simplified to `$x = ['baz' => 'bar'];` right? Or we still do this in case there is a call with side-effects?

theofidry (approved):
👍

src/Mutator/Unwrap/UnwrapArrayUintersectAssoc.php
maks-rafalko: Yes, this is how it works - producing N-1 mutatations where N is the number of arguments:

https://github.com/infection/infection/blob/1f0569f37432228f0ccb1d326d53194d06bb728f/tests/phpunit/Mutator/Unwrap/UnwrapArrayUintersectAssocTest.php#L58-L74

src/Mutator/Arithmetic/Assignment.php
maks-rafalko: I would prefer to run something like Rector to change it in the whole project, if needed, and if this possible ;) but for now leave as is if you don't mind

theofidry: fair

src/Mutator/Unwrap/UnwrapArrayUintersectAssoc.php
theofidry: Would something like that still work?

```
            <<<'DIFF'
- $x = array_uintersect_assoc(
-     ['foo' => 'bar'],
-     ['baz' => 'bar'],
-     $value_compare_func
- );
# Mutation °1
+ $x = ['foo' => 'bar'];
# Mutation °2
+ $x = ['baz' => 'bar'];
```

maks-rafalko: do you mean we need to point the number of mutation in this diffs? If so - makes sense.. can do in another PR

theofidry: Yes I find it a bit confusing otherwise as it's not clear that it's actually two mutations not one. It is a suggestion however, maybe there is a better way to convey it

maks-rafalko: agree, will do, thanks

PR #1472, opened 2021-01-18
Do not log by default

sanmai:
In the light of #1430 it doesn't make sense to keep Infection suggesting logging at all times. Keeping old behavior will mean that Infection will use 66% more memory where it could not.

This PR:

- [x] Updates configuration master to do not log by default.
- [x] Covered by tests


maks-rafalko (approved):

maks-rafalko:
merged to master https://github.com/infection/infection/commit/edfe1bd9308731988156d1d1d0a24d2400576fb2

theofidry:
I wonder if it's wise though: when you kill everything that's all good but as soon as you got one, then what? You need to run infection again with a log file this time

sanmai:
Then just `-s`, no?

theofidry:
WDYM?

sanmai:
I always use Infection with `--show-mutations`, no logging needed.

theofidry:
but because it's not the default I think this will just lead to very confused users. But maybe `--show-mutations ` could be enabled automatically when there is no log file?

sanmai:
This also be confusing, e.g. when a user expects `--show-mutations` to be enabled automatically, add a log file, and sees it no longer enabled without clear reason. This can be alleviated by a adding a note in the output, but then we could consider other similar ways. For example:

- We could add "Tip: to show escaped mutations run Infection with --show-mutations" somewhere after the results.
- We could have `--show-mutations` enabled at all times unconditionally, but showing it above metrics.

maks-rafalko:
I would vote for 

> We could add "Tip: to show escaped mutations run Infection with --show-mutations" somewhere after the results.

Issue #1475, opened 2021-01-24
Stryker dashboard returned an unexpected response code: 200


webarchitect609:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.18.2
| Test Framework version | PHPUnit 9.5.1
| PHP version | 7.3.14
| Platform    | Ubuntu 18.04.5 LTS
| Github Repo | https://github.com/webarchitect609/composer-library-skeleton

Summary
-------

Infection does not expect the `200 OK` result from Stryker Mutator on badge dashboard report send. Minor harmless bug?

Steps to reproduce
------------------

1. Enable [mutation bage](https://infection.github.io/guide/mutation-badge.html) exactly as described in the Infection documentation.
2. Use Travis CI and don't forget to create a secrent environment `INFECTION_BADGE_API_KEY`
3. Use the following command to run Infection(in `.travis.yml` of course):
    ```bash 
    vendor/bin/infection --threads=4 --min-msi=100 --min-covered-msi=100 --no-progress --no-interaction --ansi --skip-initial-tests --coverage=build/coverage-xml
    ```
    (yes, in my case I use composer scripts, but the result command is exactly as shown above)
4. Run the build in Travis CI, and then check the build when Infection was sending the data to Stryker Mutator.

Expected result
---------------

Totally successfull sending.

Actual result
-------------

```
[warning] Sending dashboard report...

[warning] Stryker dashboard returned an unexpected response code: 200

[notice] Dashboard response:

HTTP/1.1 200 OK

Content-Length: 115

Content-Type: application/json; charset=utf-8

ETag: W/"73-XovDObjU2iT2NVTMM2NbmjSuo5U"

X-Powered-By: Express

Set-Cookie: ARRAffinity=cfe9d88b4d05c4f99a90649b783196ca671405478765ab764374675d71877225;Path=/;HttpOnly;Secure;Domain=dashboard.stryker-mutator.io

Set-Cookie: ARRAffinitySameSite=cfe9d88b4d05c4f99a90649b783196ca671405478765ab764374675d71877225;Path=/;HttpOnly;SameSite=None;Secure;Domain=dashboard.stryker-mutator.io

Date: Sun, 24 Jan 2021 17:33:01 GMT

{"href":"https://dashboard.stryker-mutator.io/reports/github.com/webarchitect609/composer-library-skeleton/master"}

```

[...]

PR #1477, opened 2021-01-25
Expect status 200 as well as 201 for Stryker Dashboard

maks-rafalko:
Seems like Stryker Dashboard decided to return 200 status instead of previously returned 201.

Let's expect both in case they decide to change it again

Fixes https://github.com/infection/infection/issues/1475

- [ ] merge to master as well

### DIFF
--- a/src/Logger/Http/Response.php
+++ b/src/Logger/Http/Response.php
@@ -42,8 +42,8 @@ use Webmozart\Assert\Assert;
  */
 final class Response
 {
-    public const OK_RESPONSE_CODE = 200;
-    public const CREATED_RESPONSE_CODE = 201;
+    public const HTTP_OK = 200;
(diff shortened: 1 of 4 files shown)


src/Logger/Http/Response.php
```diff
@@ -42,6 +42,7 @@
  */
 final class Response
 {
+    public const OK_RESPONSE_CODE = 200;
```
sidz, line 45: what do you think if constant will have a name like `HTTP_OK` and `HTTP_CREATED` ?

sidz (approved):
:+1:

src/Logger/Http/Response.php
maks-rafalko: :ok_hand:

PR #1487, opened 2021-02-22
tests(makefile): moved cs-fixer download to GitHub releases

DeepDiver1975:
This PR:

- [ ] Adds new feature ...
- [ ] Covered by tests

When executing make cs cs-fixer was not properly downloaded. 
https://cs.sensiolabs.org/download/php-cs-fixer-v2.phar seems to be offline.
I decided to move over to GitHub releases of cs-fixer.

Let me know if this is the way you want this to be fixed,

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ BOX=./.tools/box
 BOX_URL="https://github.com/humbug/box/releases/download/3.10.0/box.phar"
 
 PHP_CS_FIXER=./.tools/php-cs-fixer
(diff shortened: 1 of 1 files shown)


maks-rafalko (approved):
Well, this is a little bit different behaviour, as previously we downloaded the latest version, but now it's sticked to exact one.

However, for me it's better, because when the new PHP-CS-Fixer version is released, we always have failed builds without any code changes, which is not good.

I'm :+1: here

sanmai (approved):

PR #1491, opened 2021-02-23
Build PHAR distribution using docker

maks-rafalko:
I have PHP 8.0.2 installed locally, and `make compile` does not work there.

I'm not 100% sure we need it, but it was very handy to build Infection PHAR using docker locally, rather than reinstalling PHP and/or looking why it does not work (still useful).

So, since we have other commands executed in docker container, I would like to propose having such new command for docker as well.

### DIFF
--- a/devTools/Dockerfile-php74-xdebug
+++ b/devTools/Dockerfile-php74-xdebug
@@ -22,6 +22,7 @@ RUN apk add --no-cache \
         make \
         bash \
         expect \
+        git \
         zip
 
 COPY --from=composer:latest /usr/bin/composer /usr/bin/composer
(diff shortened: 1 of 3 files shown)


devTools/Dockerfile-php74-xdebug
```diff
@@ -22,6 +22,7 @@ RUN apk add --no-cache \
         make \
         bash \
         expect \
+        git \
```
maks-rafalko, line 25: this is needed for Box to get the tag

sidz (approved):

maks-rafalko:
Thank you @sidz

Issue #876, opened 2019-12-03
composer install --no-scripts installs 0.13.4 instead of 0.15.0


someniatko:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.13.4 (`infection.phar --version`) but intended is 0.15.0
| Test Framework version | PHPUnit 8.4.3
| PHP version | 7.4.0
| Platform    | Alpine Linux, also Windows 10 - results are the same
| Github Repo | -

1. add `infection/infection` `^0.15` to your `composer.json` file
2. remove `vendor` folder and `composer.lock` file for the clean experiment
3. run `composer install --no-scripts` (`composer.lock` file will be generated)
4. observe that in `composer.lock` version is set to correct one (for now it's `0.15.0`)
5. run `vendor/bin/infection --version`.

**Expected:**
Infection - PHP Mutation Testing Framework 0.15.x@\<hash\>

**Actual:**
Infection - PHP Mutation Testing Framework 0.13.4@f0f8327f5e60d1a7d8ab4ea01074c6250ab96f57

theofidry:
maybe a `composer why infection/infection` would tell? Btw it's usually better to do (provided it's an app not a lib) `composer update xx/yy` eventually with the `--with-dependencies` flag than removing the `composer.lock`

someniatko:
@theofidry i don't have a composer.lock file because i am developing a library, therefore the .lock file is not committed to the git (it makes no sense anyway, because the project using the library will define the final list of dependencies). Then in CI i execute `composer install` and then launch tests. However, if i use `--no-scripts` flag, for some reason it installs "partially" or so. Maybe i have indeed the `0.15.0` version, however, for some reason it identifies itself as 0.13.4 if some scripts are not executed.

someniatko:
Yes, it looks like it's indeed so: 
```
user> composer why infection/infection
<my lib name> dev-master  requires (for development)  infection/infection (^0.15)
user> php74 vendor/bin/infection --version
You are running Infection with Xdebug enabled.
Infection - PHP Mutation Testing Framework 0.13.4@f0f8327f5e60d1a7d8ab4ea01074c6250ab96f57
```

however, infection still identifies itself as 0.13.4

maks-rafalko:
We are relying on [`ocramius/package-versions`](https://github.com/Ocramius/PackageVersions) package:

[...]

PR #1496, opened 2021-03-14
Upgrade ocramius/package-versions to 1.11.0

maks-rafalko:
Also, set min PHP version from… 7.4.0 to 7.4.7 as `package-versions` requires

Fixes https://github.com/infection/infection/issues/876

How did I tested it:

* created a new project
* `composer require infection/infection` (loaded 0.21.4)
* `rm -rf vendor`
* `composer install --no-scripts`
* `vendor/bin/infection -V` - 0.21.4 (OK). Note that `ocramius/package-versions`'s `composer.lock` file [containes `0.16.4`](https://github.com/Ocramius/PackageVersions/blob/1.11.x/composer.lock#L982), so the issue is fixed as I see

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
         }
     ],
     "require": {
-        "php": "^7.4 || ^8.0",
+        "php": "^7.4.7 || ^8.0",
         "ext-dom": "*",
         "ext-json": "*",
         "ext-libxml": "*",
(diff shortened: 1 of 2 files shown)


sanmai (approved):

maks-rafalko:
Thank you @sanmai

PR #1499, opened 2021-03-19
pcov.directory autodetection

### LINKED ISSUE #1497: It would be nice if Infection warned that none of the set up directories are covered by the configured pcov
I understand that this should be done by PhpUnit and/or pcov, but they are not talking with each other. So I guess it's up to all the unittest runners to check it for them (maybe I need to open tickets for ~~PhpStorm and~~ Visual Studio Code as well 🤔).

* https://github.com/krakjoe/pcov/issues/49
* https://github.com/sebastianbergmann/php-code-coverage/issues/741
* https://youtrack.jetbrains.com/issue/WI-52914

sanmai:
This PR:

- [x] Add a basic `pcov.directory` autodetection
- [x] Covered by tests
- [x] Fixes #1497

Note an E2E tests that files on current Infection: f25532a

> When `pcov.directory` is left unset, PCOV will attempt to find `src`, `lib` or, `app` in the current working directory, in that order.

[If PCOV sees any of the directories](https://github.com/krakjoe/pcov/blob/develop/README.md), and if it wasn't told otherwise, it would just use first of them. 

E.g. if a project has both `lib` a and `app`, only the first fill be used for coverage, and the other will have 0% coverage. We work around this by forcing not-so-clever default value of `.` i.e. current working directory if we don't see `pcov.directory` set to something more specific.

Future step would be to set `pcov.directory` to a specific value inferred from the configuration.

### DIFF
--- a/src/Config/ValueProvider/PCOVDirectoryProvider.php
+++ b/src/Config/ValueProvider/PCOVDirectoryProvider.php
@@ -1,0 +1,73 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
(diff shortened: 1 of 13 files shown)


maks-rafalko (approved):
:ok_hand:

PR #1502, opened 2021-03-30
Fix file not found exception for Codeception Cest `bug`

agilov:
For Codeception Cests there was TestFileNameNotFoundException because $fullyQualifiedClassName variable contains colon separated feature name but there is no .feature file in junit.xml report.

In  junit.xml  there is entries like so:

`<testcase file="/app/modules/site/controllers/ExampleErrorController_Cest.php" name="test400" class="app\modules\site\controllers\ExampleErrorController_Cest" feature="test400" assertions="3" time="0.010310"/>
`

And $fullyQualifiedClassName is like this `app\modules\site\controllers\ExampleErrorController_Cest:test400`

I think there is no need to open the issue for such a small bug and I've fixed it in my vendor directory and propose the change via pull request.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -21,7 +21,8 @@ granular tests, you can run `make` to see the available commands.
 <br />
 <hr />
 
-« [Go back to the readme](README.md) »
+« [Go back to the readme][readme] »
 
 
 [docker]: https://www.docker.com/get-docker
+[readme]: /README.md
--- a/.github/workflows/autoreview.yaml
+++ b/.github/workflows/autoreview.yaml
@@ -47,7 +47,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          composer update --no-interaction --prefer-dist --no-progress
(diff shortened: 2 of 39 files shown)


maks-rafalko (commented):
Could you please 

1. add a test for this case
2. target 0.21 branch (if you want it to be released asap)

agilov:
> Could you please
> 
> 1. add a test for this case
> 2. target 0.21 branch (if you want it to be released asap)

Added test case and changed branch. But I see conflict message "This branch has conflicts that must be resolved" for file I didn't touch. Is this my fault?

maks-rafalko:
yes, this happened because master and 0.21 are different, unfortunately.

Could you please start your branch from 0.21 and cherry pick your commits? Or just make them again (shouldn't be hard).

So, as a result - we need to see only diff with your changes.

Regarding changes - cool. Will be happy to merge once diff is ready.

agilov:
Started a new branch from 0.21, created this PR: https://github.com/infection/infection/pull/1503

Issue #1509, opened 2021-04-25
Upgrade xdebug-handler to v2


maks-rafalko:
See https://github.com/composer/xdebug-handler/blob/2.0.0/UPGRADE.md

The main point is a support of Xdebug 3

> The default behaviour has changed from always restarting if Xdebug is loaded, to only restarting if Xdebug is active. This has been introduced to support Xdebug3 modes. Xdebug is considered active if it is loaded, and for Xdebug3, if it is running in a mode other than `xdebug.mode=off`.

maks-rafalko:
Implemented https://github.com/infection/infection/pull/1510

Issue #1473, opened 2021-01-19
For recent Xdebug coverage collection needs XDEBUG_MODE=coverage or xdebug.mode=coverage 


JesusValera:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.20.2
| Test Framework version | PHPUnit 9.5.0
| PHP version | PHP 7.4.14 (cli) with Xdebug v3.0.2
| Platform    | Docker (php:7.4-fpm)
| Github Repo | https://github.com/jenshaase/phel-lang

</details>

<details>
 <summary>Output with issue</summary>

 ```

 dev@php:/srv/phel-lang$ vendor/bin/infection

    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/

Infection - PHP Mutation Testing Framework version 0.20.2@6035c1566af6a5a8d833a276351e5e18ed412305


Running initial test suite...

PHPUnit version: 9.5.0

  614 [============================]  1 min

 In CoverageChecker.php line 169:

  The code coverage generated by the initial test run is invalid. Please report the issue on the
  infection repository "https://github.com/infection/infection".

  $ '/srv/phel-lang/vendor/phpunit/phpunit/phpunit' '--configuration' '/srv/phel-lang/data/infection/infection/
  phpunitConfiguration.initial.infection.xml' '-vvv' '--coverage-xml=/srv/phel-lang/data/infection/infection/co
  verage-xml' '--log-junit=/srv/phel-lang/data/infection/infection/junit.xml'
  PHPUnit 9.5.0 by Sebastian Bergmann and contributors.

  Runtime:       PHP 7.4.14
  Configuration: /srv/phel-lang/data/infection/infection/phpunitConfiguration.initial.infection.xml
  Random Seed:   1611098728
  Warning:       XDEBUG_MODE=coverage or xdebug.mode=coverage has to be set

  ...............................................................  63 / 608 ( 10%)
  ............................................................... 126 / 608 ( 20%)
  ............................................................... 189 / 608 ( 31%)
  ............................................................... 252 / 608 ( 41%)
  ............................................................... 315 / 608 ( 51%)
  ............................................................... 378 / 608 ( 62%)
  ............................................................... 441 / 608 ( 72%)
[...]

PR #1510, opened 2021-04-26
Upgrade xdebug-handler to v2 and handle that `coverage` option is set

sidz:
Upgrade xdebug-handler to v2 and handle that `coverage` option is set in case of Xdebug 3

Feature Request: https://github.com/infection/infection/issues/1509
Issue: https://github.com/infection/infection/issues/1473

### DIFF
--- a/src/Console/XdebugHandler.php
+++ b/src/Console/XdebugHandler.php
@@ -52,7 +52,7 @@ final class XdebugHandler
     {
         // We force the color option unconditionally since it is able to detect the --no-ansi option
         // to disable it if necessary
-        (new ComposerXdebugHandler(self::PREFIX, '--ansi'))
+        (new ComposerXdebugHandler(self::PREFIX))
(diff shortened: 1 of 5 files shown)


src/Console/XdebugHandler.php
```diff
@@ -52,7 +52,7 @@ public static function check(LoggerInterface $logger): void
     {
         // We force the color option unconditionally since it is able to detect the --no-ansi option
         // to disable it if necessary
```
maks-rafalko, line 54: is this comment still relevant since no `--ansi` argument is passed anymore?

maks-rafalko (approved):

src/Console/XdebugHandler.php
sidz: removed

maks-rafalko:
Thank you @sidz

Issue #1514, opened 2021-05-03
Multiplication mutator should not mutate when return value is integer


maks-rafalko:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.22.0
| Test Framework version | PHPUnit
| Reproducer | https://infection-php.dev/r/dld0

The following mutation does not make sense:

```diff
public function mul(int $a, int $b): int
{
-    return $a * $b;
+    return $a / $b;
}
```

because it is always killed since result type of new `$a / $b` expression is `float` but `int` is expected.

Such mutant is killed with the following error message:

```
TypeError: Return value of Infected\SourceClass::mul() must be of the type int, float returned
```

PR #1515, opened 2021-05-03
Multiplication mutator should not mutate when return value is integer

sidz:
This PR fixes https://github.com/infection/infection/issues/1514 request.

### DIFF
--- a/src/Mutator/Arithmetic/Multiplication.php
+++ b/src/Mutator/Arithmetic/Multiplication.php
@@ -39,6 +39,9 @@ use Infection\Mutator\Definition;
 use Infection\Mutator\GetMutatorName;
 use Infection\Mutator\Mutator;
 use Infection\Mutator\MutatorCategory;
+use Infection\PhpParser\Visitor\ParentConnector;
+use Infection\PhpParser\Visitor\ReflectionVisitor;
+use function is_string;
 use PhpParser\Node;
 
 /**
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -46,7 +46,7 @@
         "ext-dom": "*",
         "ext-json": "*",
         "ext-libxml": "*",
-        "composer/xdebug-handler": "^1.3.3",
+        "composer/xdebug-handler": "^2.0",
```
maks-rafalko, line 49: seems like this commit should not be in this PR (already in master)

src/Mutator/Arithmetic/Multiplication.php
```diff
@@ -94,7 +96,19 @@ public function canMutate(Node $node): bool
             return false;
         }
 
-        return true;
+        $functionScope = $node->getAttribute(ReflectionVisitor::FUNCTION_SCOPE_KEY);
```
maks-rafalko, line 99: I believe we should also check that the current node is a `return` expression as it's done e.g. here: https://github.com/infection/infection/blob/master/src/Mutator/ReturnValue/This.php#L84

Otherwise, the following code will not be mutated, but it should:

```php
public function mul(int $a, int $b): int
{
    $c = $a * $b;

    return 1;
}
```

btw, could you please add such case to tests so that we are sure it is mutated?

sidz: done

maks-rafalko (approved):
👌

maks-rafalko:
Thank you @sidz

PR #1468, opened 2021-01-10
Add support for `@infection-ignore-all` annotation

sanmai:
See #1231

This PR:

- [x] Adds a possibility to ignore an entire statement by commenting it with  `@infection-ignore-all`
- [x] Covered by an E2E test
- [x] Covered by unit tests
- [x] Doc PR: PR description can be used verbatim

Give this example Infection won't mutate or even look at anything inside this function:

```php
/** @infection-ignore-all */
public function doSomethingNastyButCostlyToRefactor() {

}
```

Likewise, given this annotation Infection won't consider anything in this loop:

```php
/** @infection-ignore-all */
foreach ($foo as $bar) {
    // 
}
```

PR #1517, opened 2021-05-04
Remove CodeCoverageAnnotationIgnorer

sanmai:
This PR:

- [x] Removes CodeCoverageAnnotationIgnorer 
- [x] Covered by tests (removes em')
- [x] Adds a test to ensure `@infection-ignore-all` works on class level too [as we say in the docs](https://github.com/infection/site/pull/215/files).
- [x] Doc PR: https://github.com/infection/site/pull/215

Since we have a very explicit annotation `@infection-ignore-all` (added in #1468) that, again, very explicit on the purpose, and works not only for methods, but for most kinds of statements, let's remove `@codeCoverageIgnore` ignorer, that not only implicit (you may only want to ignore a piece of code temporarily, and only ignore) but also messes up the metrics (hides otherwise-uncovered mutations).


sidz (approved):
:+1:

maks-rafalko (approved):

Issue #1528, opened 2021-07-02
Using the splat operator (`...`) to copy an array


AlexandruGG:
Hello!

**Is your feature request related to a problem? Please describe.**
Copying an array using the `...` operator is different from copying via assignment because the array pointer is maintained in the second case. It would be great if there would be a mutator which switches one for the other. 

**Describe the solution you'd like**
```PHP
$array = [1, 2, 3];
$array2 = [...$array]; // -> this could be mutated to $array2 = $array;
```

**Additional context**
See example here, I hope this clarifies what I mean: https://3v4l.org/gfW5D#v8.0.8

I noticed there is already a [Spread Mutator](https://github.com/infection/infection/blob/master/src/Mutator/Operator/Spread.php) so I thought this could be something similar to that one, let me know if this makes sense.

maks-rafalko:
Hi, thanks for this feature request. You are very welcome to implement it, shouldn't be hard :)

AlexandruGG:
I'm going to try get familiar with the codebase and give it a shot 😄

PR #1529, opened 2021-07-03
Add Mutators: SpreadAssignment, SpreadRemoval

AlexandruGG:
This PR:

- [x] Adds new mutators involving the splat operator (`...`)
- [x] Renames existing `Spread` Mutator to `SpreadOneItem`
- [x] Covered by tests
- [x] Docs PR: https://github.com/infection/site/pull/219

Closes: https://github.com/infection/infection/issues/1528

I manually tested with the `.phar` the behaviour provided as an example in the issue.


src/Mutator/Operator/SpreadAssignment.php
```diff
@@ -0,0 +1,107 @@
...
+            return false;
+        }
+
+        if (count($node->items) !== 1) {
+            return false;
+        }
+
+        Assert::allNotNull($node->items);
+
+        return $node->items[0]->unpack;
```
AlexandruGG, line 105: the assertion is more for PHPStan because from what I understand and see in other examples the items will be `ArrayItem`, but the type [here](https://github.com/nikic/PHP-Parser/blob/master/lib/PhpParser/Node/Expr/Array_.php#L13) is `ArrayItem|null`. Let me know if I'm missing something, it's the first time I'm working with the php parser

AlexandruGG:
@maks-rafalko hey 👋, can you please run the workflows on this? Also, it should be ready for review

src/Mutator/Operator/SpreadRemoval.php
```diff
@@ -0,0 +1,99 @@
...
+```php
+$x = [$collection, 4, 5];
+```
+TXT
+            ,
+            MutatorCategory::SEMANTIC_REDUCTION,
+            null,
+            <<<'DIFF'
+- $x = [...$collection, 4, 5];
++ $x = [$collection, 4, 5];
```
maks-rafalko, line 74: This particular mutator will be killed by static analyzers, which makes it useless.

Please, see the following Psalm examples:

* `$x = [...$collection, 4, 5];` - https://psalm.dev/r/4b159a69f1 - no issues
* `$x = [$collection, 4, 5];` - https://psalm.dev/r/9e9b78e2e1 Psalm error - `11:12 - Argument 1 of acceptsInt expects int, array{1, 2, 3} provided`

in other words, if we are using for example https://github.com/Roave/infection-static-analysis-plugin - this mutant will always be killed, resulting in wasted resources.

So I would vote to not do this mutation. Does it make sense?

maks-rafalko (commented):
Thank you very much for this contribution.

I do like the `SpreadAssignment` mutator, however I have some concerns about `SpreadRemoval`, please detailed comment below.

src/Mutator/Operator/SpreadRemoval.php
AlexandruGG: interesting point, I'm familiar with that plugin, however, in your example we are making an assumption: we assume that the array is a list of a single type, which might not be the case. There are many situations where an array will have mixed types and this mutant becomes valuable. 

Here's the simplest example, based on yours: https://psalm.dev/r/6ba915ee93. I've added the assertions at the end to show that the first element of each new array is not the same.

There will be other cases where the union type isn't even mentioned because an array can have many different types, so the developer might go with `mixed`.

maks-rafalko: ok, makes sense ;)

maks-rafalko (approved):
Overall - looks good to me. 

Let's see, if nobody else has any objections, I'm happy to merge

sanmai (approved):
I'm not the biggest fan of `SpreadRemoval` since it seems like not the most productive mutator, but we have several more of this kind, and yet sometimes they can be of use.

maks-rafalko:
Ok, let's merge then. 

Thank you @AlexandruGG 

---

I'm starting thinking about somehow gathering user's statistics about mutators (not just killing/escaping ratio, but usefulness from developers points of view during working on real projects), as described in this great article https://github.com/infection/infection/discussions/1526

AlexandruGG:
@maks-rafalko @sanmai thank you both for reviewing and merging this so quickly! 🙌

drupol:
Very nice stuff ! :)

tests/phpunit/Mutator/Operator/SpreadAssignmentTest.php
```diff
@@ -0,0 +1,172 @@
...
+declare(strict_types=1);
+
+namespace Infection\Tests\Mutator\Operator;
+
+use Infection\Tests\Mutator\BaseMutatorTestCase;
+
+final class SpreadAssignmentTest extends BaseMutatorTestCase
+{
+    /**
+     * @requires PHP >= 7.4
```
sidz, line 43: I'm late to the party.

`@requires` is not needed as minimum compatible with Infection version of PHP is 7.4

AlexandruGG: Good catch, should probably be removed from other places as well then. I might create a PR specifically for those

sidz: > I might create a PR specifically for those

yes please. Thanks

PR #1533, opened 2021-07-07
Remove `@requires PHP >= 7.4` Annotations

AlexandruGG:
This PR removes annotations which are no longer needed because the project already requires PHP 7.4 or 8: https://github.com/infection/infection/blob/master/composer.json#L45

Addresses: https://github.com/infection/infection/pull/1529#discussion_r664563102


AlexandruGG:
@sidz this is ready, small one :)

sidz (approved):
:+1: 

thanks @AlexandruGG

sanmai (approved):

PR #1535, opened 2021-07-12
Run Infection on PHP 8.1.

sidz:
This PR:
- [ ] Runs Infection on PHP 8.1.
- [ ] Makes PHP 8.0 main version which will be used for running Coding Standards, Mutation testing and so on

### DIFF
--- a/.github/workflows/autoreview.yaml
+++ b/.github/workflows/autoreview.yaml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        php-version: ['7.4']
+        php-version: ['8.0']
 
     name: Autoreview on PHP ${{ matrix.php-version }}
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,15 +17,16 @@ jobs:
     strategy:
       matrix:
         operating-system: [ubuntu-latest]
-        php-version: ['7.4']
+        php-version: ['8.0']
         dependencies: ['']
         coverage-driver: [pcov, xdebug]
         include:
           - { operating-system: 'windows-latest', php-version: '7.4', dependencies: '', coverage-driver: 'xdebug' }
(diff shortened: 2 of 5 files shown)


sidz:
latest release `nikic/PHP-Parser ` still has an issue with parsing parameter passed by reference.
Already fixed in master: https://github.com/nikic/PHP-Parser/commit/c758510a37218d631fd10f67bca5bccbfef864fa

waiting release.

maks-rafalko:
> waiting for release.

https://github.com/nikic/PHP-Parser/releases/tag/v4.12.0

.github/workflows/autoreview.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        php-version: ['7.4']
```
maks-rafalko, line 19: 7.4 is still supported until 28 of November, 2021, and it's used [by ~60% of Infection users](https://github.com/infection/infection/blob/master/composer.json#L45), I think we should not get rid of it now.

Or, alternatively, if we want to get rid (personally, I'm ok with that) - we should bump the version in composer.json to ^8.0, here

https://github.com/infection/infection/blob/master/composer.json#L45

sidz: reverted

.github/workflows/ci.yaml
```diff
@@ -40,14 +41,10 @@ jobs:
...
 
-      - name: Get composer cache directory
-        id: composer-cache
-        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
-
       - name: Cache dependencies
         uses: actions/cache@v2
         with:
-          path: ${{ steps.composer-cache.outputs.dir }}
+          path: ~/.cache/composer
```
sanmai, line 47: This won't work on Windows IIRC, and we're testing on Windows.

sidz:
looks like Github have some issues with builds based on windows. Same on master

PR #1149, opened 2020-03-10
Implement the Stryker HTML report

theofidry:
Closes #1137

Issue #1137, opened 2020-03-06
Add support for Stryker dashboard HTML report


theofidry:
As of now we are naming our badge logger "badge" because we make use of the minimal API to just get the badge with the mutation score.

Stryker dashboard however supports a full HTML report as well. The specifications can be found [here](https://github.com/stryker-mutator/mutation-testing-elements/tree/master/packages/mutation-testing-report-schema).

For the configuration I propose two choices:

- A new logger (at the configuration level), e.g. `dashboard`
- Rename `badge` to `dashboard` or `stryker-dashboard` and have a setting `htmlReport: bool` with opt-in by default

PR #1538, opened 2021-07-14
Made `logs.badge.branch` capable of accepting a regular expression

### LINKED ISSUE #814: Upload MSI for all/multiple branches
The way the config is documented implies mutation score can only be uploaded for one branch. It should be possible to upload multiple (specified) branches and all branches (without specifying).

### LINKED ISSUE #1536: Allow sending stryker-mutator.io reports for multiple branches
Since approximately 2019, I started moving away from the `master` branch naming, and started to use the minor version of libraries as branch name, such as:

 * `1.0.x`
 * `2.1.x`
 * ...

Infection has a configuration for sending information to `stryker-mutator.io`, but only when the branch name matches **precisely**:

https://github.com/infection/infection/blob/f2290b58fff988d17577d43fabf7bb64a1f6049f/src/Logger/BadgeLogger.php#L86

Would a patch to change this into a regex (perhaps new setting, if necessary?) be welcome?

Ocramius:
By configuring `logs.badge.branch` with a `/`-delimited regular expression, we can collect
reports about a number of branches.

```json
{
    "source": {
        "directories": ["src/"]
    },
    "logs": {
        "badge": {
            "branch": "/^release-.*$/"
        }
    }
}
```
    
Note that `logs.badge.matchBranchRegex`, initially proposed as an alternative, has also been removed.

We can safely assume that a `branch` value starting and ending with `/` is a regular expression, because
git branches starting or ending with `/` are not valid anyway.

Fixes #1536
Fixes #814 

Related: #1137
Related: #1149

This PR:

- [x] Adds new feature ...
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/220


tests/phpunit/Configuration/Entry/LogsAssertions.php
```diff
@@ -65,7 +63,8 @@ private function assertLogsStateIs(
             $this->assertNull($badge);
         } else {
             $this->assertNotNull($badge);
-            $this->assertBadgeStateIs($badge, $expectedBadge->getBranch());
```
Ocramius, line 68: Removed this: it made the tests much more confusing than they were, and a simpler solution was to do object-to-object equality.

tests/benchmark/Tracing/provide-traces-closure.php
```diff
@@ -69,7 +69,8 @@
     Container::DEFAULT_DRY_RUN,
     Container::DEFAULT_GIT_DIFF_FILTER,
     Container::DEFAULT_GIT_DIFF_BASE,
-    Container::DEFAULT_USE_GITHUB_LOGGER
+    Container::DEFAULT_USE_GITHUB_LOGGER,
+    true
```
Ocramius, line 73: This parameter was missing and crashing my local e2e tests (`make test`).

Something about no-op mutators.

tests/benchmark/MutationGenerator/generate-mutations-closure.php
```diff
@@ -65,7 +65,8 @@ static function (SplFileInfo $fileInfo): Trace {
 );
 
 $mutators = $container->getMutatorFactory()->create(
-    $container->getMutatorResolver()->resolve(['@default' => true])
+    $container->getMutatorResolver()->resolve(['@default' => true]),
+    true
```
Ocramius, line 69: This parameter was missing and crashing my local e2e tests (`make test`).

Something about no-op mutators.

src/Logger/BadgeLogger.php
```diff
@@ -54,22 +55,25 @@ final class BadgeLogger implements MutationTestingResultsLogger
     private StrykerApiKeyResolver $strykerApiKeyResolver;
     private StrykerDashboardClient $strykerDashboardClient;
     private MetricsCalculator $metricsCalculator;
-    private string $branch;
+    private ?string $exactBranchMatch;
```
Ocramius, line 58: IMO we could perhaps get rid of this long-term: for now, I kept it.

Internals could build a single regex when an `$exactBranchMatch` exists, but it's probably easier to proceed this way for now.

Ocramius: Alternatively, `Branch` could provide a `->matches(string $branch): bool`, which would remove this state proliferation, and reduce test amount.

src/Configuration/Entry/Badge.php
```diff
@@ -35,20 +35,45 @@
...
 {
-    private string $branch;
+    private ?string $exactBranchMatch;
+    private ?string $matchBranchRegex;
+
+    public function __construct(?string $exactBranchMatch, ?string $matchBranchRegex)
+    {
+        if ($matchBranchRegex !== null) {
+            try {
+                // Yes, the `@` is intentional. For some reason, `thecodingmachine/safe` does not suppress the warnings here
```
Ocramius, line 55: Really had to take a shower after this: I was hoping that `thecodingmachine/safe` did deal with this stuff, but nope.

src/Logger/BadgeLogger.php
Ocramius: In fact, I like `Badge#availableFor(string $branchName)`.

I'll probably refactor this, but **after** I have a first review from the maintainers :)

maks-rafalko (commented):
I'm ok with the current state of PR.

One idea to discuss though: what if we make existing (in `master`) config key `branch` more clever to understand regexes?

```json
{
    "logs": {
        "badge": {
            "branch": "master"
        }
    }
}
```

^ BC support, pushes the report for the exact branch - `master` in this case.

```json
{
    "logs": {
        "badge": {
            "branch": "/^foo$/"
        }
    }
}
```

^ new feature. Infection understands this is a regex and matches with `preg_match()`.

In this case we will keep BC and extend the functionality of `branch` config key.

Ocramius:
> In this case we will keep BC and extend the functionality of `branch` config key.

Didn't want to mess with the existing `branch` value, as I don't know what values users have used within it so far.

What I can certainly try to do is say that if the `branch` value starts with `/` and ends with `/`, and is at least 3 characters long, then I treat it as a regex: a bit weak, but could work?

maks-rafalko:
> What I can certainly try to do is say that if the `branch` value starts with `/` and ends with `/`, and is at least 3 characters long, then I treat it as a regex: a bit weak, but could work?

IMO, sounds ok :) I believe the majority of the users use it like `branch: master`

Also, why I like the idea of having just one `branch` config key - it will be similar to PHPUnit's `--filter`, which accepts raw values as well as regular expressions, which makes it quite flexible.

Ocramius:
Ok, will go back at this and simplify the patch: possibly rewrite, since that would be simpler :D

Ocramius:
Thinking further about this, git branches cannot start with `/` nor `/`, so we're potentially good here :+1:

src/Configuration/Entry/Badge.php
Ocramius: Looked into this: `"format": "regex"` would collide with simple values

Ocramius:
@maks-rafalko I've reworked this to hide the fact that a regex or exact match happens inside a `Branch`, of which I do no longer expose the configuration contents.

I'll rework the docs PR.

Small suggestion: smash docs and sources together - easier for atomic diffs ;-)

Ocramius:
Docs PR also updated.

resources/schema.json
```diff
@@ -65,12 +65,22 @@
                 },
                 "badge": {
                     "type": "object",
+                    "description": "Mutation score badge for your GitHub project. If provided, infection will report results for maching branches to an upstream reporting dashboard/collector.",
```
maks-rafalko, line 68:
```suggestion
                    "description": "Mutation score badge for your GitHub project. If provided, Infection will report results for matching branches to an upstream reporting dashboard/collector.",
```

resources/schema.json
```diff
@@ -65,12 +65,22 @@
...
                     "additionalProperties": false,
                     "required": ["branch"],
                     "properties": {
                         "branch": {
                             "type": "string",
-                            "description": "Mutation score badge for your GitHub project."
+                            "description": "Mutation score badge for your GitHub project. If this value starts and ends with \"/\", it will be considered a regular expression.",
```
maks-rafalko, line 74:
```suggestion
                            "description": "Mutation score badge for your GitHub project. If this value starts and ends with '/', it will be considered a regular expression.",
```

up to you, but IMO easier to read

maks-rafalko (approved):
:+1:

Ocramius:
Applied suggested improvements 👍

tests/phpunit/Configuration/Entry/BadgeTest.php
```diff
@@ -36,16 +36,45 @@
...
+        $this->assertFalse($badge->applicableForBranch(' foo'));
+        $this->assertFalse($badge->applicableForBranch('foo1'));
+    }
+
+    public function test_it_rejects_invalid_regex(): void
+    {
+        try {
+            new Badge('/[/');
+
+            $this->fail();
```
theofidry, line 73: the slightly less error prone approach with `expectException()` is probably fine here (and IIRC preferred but don't quote me on this)

tests/phpunit/Configuration/Entry/LogsAssertions.php
theofidry: object-to-object equality is always tricky due to the loose comparison, but if anyone is interested there is https://github.com/webmozarts/strict-phpunit which solves that issue and allows to remove a lot of boilertemplate

theofidry (approved):
👍 Left some comments but I don't think they are critical

tests/phpunit/Configuration/Entry/BadgeTest.php
Ocramius: Ah yes, this is because my initial approach was testing for a previous exception, which I then got rid of.

Indeed, will refactor.

Ocramius:
@theofidry for value objects, object-to-object equality is an acceptable trade-off, so I'll leave that one as-is.

It is true that some edge cases can be troublesome, but they mostly affect floating point arithmetic and deeply nested graphs with recursion, so not really important in here.

tests/phpunit/Configuration/Entry/BadgeTest.php
Ocramius: Never mind: I still have a previous exception, so this kind of test still applies as-is, and `expectException()` is not a sufficient assertion.

tests/phpunit/Configuration/Entry/BadgeTest.php
```diff
@@ -36,16 +36,45 @@
...
 
-        $this->assertBadgeStateIs($badge, 'master');
+        $this->assertTrue($badge->applicableForBranch('master'));
+        $this->assertFalse($badge->applicableForBranch('mast'));
+        $this->assertFalse($badge->applicableForBranch('master '));
+        $this->assertFalse($badge->applicableForBranch(' master'));
+        $this->assertFalse($badge->applicableForBranch('master1'));
+    }
+
+    public function test_it_can_be_instantiated_with_a_regex_branch_match(): void
```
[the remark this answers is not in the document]
Ocramius: 8a37265abde6f6502acab3fdf249b054d4ec2136

theofidry:
@Ocramius definitely, not necessary here for this case

Ocramius:
Rebased and applied @maks-rafalko and @theofidry's suggestions.

Also looking forward to have this merged, so I don't have to wait for upstream approval for CI :shrug: (thanks crypto-bros, for ruining github)

tests/phpunit/Configuration/Entry/BadgeTest.php
```diff
@@ -36,16 +36,54 @@
...
+            [' master ', 'master', false],
+            ['master1', 'master', false],
+            ['foo', '/^(foo|bar)$/', true],
+            ['bar', '/^(foo|bar)$/', true],
+            ['foobar', '/^(foo|bar)$/', false],
+            ['fo', '/^(foo|bar)$/', false],
+            ['ba', '/^(foo|bar)$/', false],
+            ['foo ', '/^(foo|bar)$/', false],
+            [' foo', '/^(foo|bar)$/', false],
+            ['foo1', '/^(foo|bar)$/', false],
```
sanmai, line 73: You can also do `yield 'test name' => ['master', 'master', true];` but I'm nitpicking here.

Ocramius: I prefer to avoid `Generator` when a simpler type is sufficient :+1:

Ocramius:
Thanks for the merge!

@theofidry FWIW, avoid squashing - I put some effort in commit messages, and 0dde91a2a454559a6b1877add9044261ac4bbb9e is really butchered :-\

theofidry:
>FWIW, avoid squashing

We squash the commits in general: the git history of a PR is rarely tidy and it too often provides close to no value for the amount of efforts it requires to keep it tidy (and even more to try to enforce it) when IMO a link to GitHub/GitLab provides a lot more information.

That said I'll be aware to avoid to do that on yours in the future, I did not took notice of it here

PR #1539, opened 2021-07-19
[Performance] Improve Infection performance executed against slow test suites

maks-rafalko:
This PR dramatically improves performance for Infection executed against functional (read - slow) tests.

```diff
- Time: 4m 18s. Memory: 0.05GB
+ Time: 1m 19s. Memory: 0.06GB
```

for one of the PR in my daily job's project.

## Problem

As you know, we always said that Infection runs only those tests that cover mutated line, not the all tests from an application.

It is partially true, but not exactly. PHPUnit's XML file allows us to only [filter original set of tests by `<directory>` or `<file>` ](https://phpunit.readthedocs.io/en/9.5/configuration.html#the-testsuite-element).

And this is what we do starting from 2017. 

<details>
<summary>Example of a `phpunit.xml` file generated by Infection for one of the Mutant (click me)</summary>

```xml
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/schema/9.2.xsd" bootstrap="/infection/tests/e2e/PestTestFramework/./infection/interceptor.autoload.740af9ed4589f3494f0c14ed544eec29.infection.php" colors="false" stopOnFailure="true" cacheResult="false" stderr="false">
  <testsuites>
    <testsuite name="Infection testsuite with filtered tests">
      <file>/infection/tests/e2e/PestTestFramework/tests/CalculatorPhpUnitTest.php</file>
    </testsuite>
  </testsuites>
  <filter>
    <whitelist>
      <directory>/infection/tests/e2e/PestTestFramework/src/</directory>
    </whitelist>
  </filter>
</phpunit>

```

</details>

Notice in the example above, that we tell PHPUnit to run tests only from `CalculatorPhpUnitTest.php`. But this file can contain 1 test that covers mutated line, **and hundreds of other tests** that still will be executed for Mutant. 

It wasn't always noticeable because we (ok, at least I am) used Infection for **unit** tests, which are fast enough.

However, since I'm currently working on a project with 2000+ *functional* tests and run Infection against this huge functional test suite, this became a bottleneck.

Imagine there is a file `ProductApiTest.php` that covers all the API endpoints with functional tests (Symfony WebTestCase).

* In this file, there is only 1 test case that covers mutated line. This test case takes **1s**
* In this file, there are additionally other 100 test cases that together take **15s**

So, with Infection 0.23.0 (current `master`), instead of running 1 test case for 1s, we run all the tests from `ProductApiTest.php` for 1+15=16s.

## Solution

With this PR, for the example explained above, we run only 1 test case for 1s, *saving 15 from 16 seconds*.

### Technical details

Since PHPUnit doesn't allow us to provide test cases on `phpunit.xml` file level, we can only do it on a command line level, using `--filter` option

* Command line on `master`: `'phpunit' '--configuration' '/path-to/phpunitConfiguration.d3b360aa95bfe11dadbe87dbd49b70c4.infection.xml'`
* Command line on this PR: `'phpunit' '--configuration' '/path-to/phpunitConfiguration.d3b360aa95bfe11dadbe87dbd49b70c4.infection.xml' '--filter' '/App\\Test::test_case1|App\\Test::test_case2/'` 

Instead of 1000 words, here is an example of the logs for the escaped mutant:

`master`:

```bash
3) /srv/api/src/DataPersister/TableViewDataPersister.php:32    [M] InstanceOf_

--- Original
+++ New
@@ @@
     }
     public function persist($data, array $context = [])
     {
-        if ($data instanceof TableView && ($context['collection_operation_name'] ?? null) === 'post' && $this->security->getUser() instanceof Employee) {
+        if ($data instanceof TableView && ($context['collection_operation_name'] ?? null) === 'post' && true) {
             /** @var Employee $employee */
             $employee = $this->security->getUser();
             $data->setCreatedBy($employee);

$ '/srv/api/vendor/phpunit/phpunit/phpunit' '--configuration' '/tmp/infection/phpunitConfiguration.1aadd4bc46a4b6ae51fb8e3a0d295e89.infection.xml'
  Test cache cleared
  PHPUnit 9.5.6 by Sebastian Bergmann and contributors.
  
  Random Seed:   1625689971
  
  Testing 
  .............................................................   61 / 1166 (  5%)
  .............................................................  122 / 1166 ( 10%)
  .............................................................  183 / 1166 ( 15%)
  .............................................................  244 / 1166 ( 20%)
  .............................................................  305 / 1166 ( 26%)
  .............................................................  366 / 1166 ( 31%)
  .............................................................  427 / 1166 ( 36%)
  .............................................................  488 / 1166 ( 41%)
  .............................................................  549 / 1166 ( 47%)
  .............................................................  610 / 1166 ( 52%)
  .............................................................  671 / 1166 ( 57%)
  ..............................
```

and this PR:

```bash
3) /srv/api/src/DataPersister/TableViewDataPersister.php:32    [M] InstanceOf_

--- Original
+++ New
@@ @@
     }
     public function persist($data, array $context = [])
     {
-        if ($data instanceof TableView && ($context['collection_operation_name'] ?? null) === 'post' && $this->security->getUser() instanceof Employee) {
+        if ($data instanceof TableView && ($context['collection_operation_name'] ?? null) === 'post' && true) {
             /** @var Employee $employee */
             $employee = $this->security->getUser();
             $data->setCreatedBy($employee);

$ '/srv/api/vendor/phpunit/phpunit/phpunit' '--configuration' '/tmp/infection/phpunitConfiguration.5b66e9f56127cb2a42531283cf8dc120.infection.xml' '--filter' '/App\\Tests\\[...very long filter]/'
  Test cache cleared
  PHPUnit 9.5.6 by Sebastian Bergmann and contributors.
  
  Random Seed:   1625689648
  
  Testing 
  ...............................................................  63 / 192 ( 32%)
  ............................................................... 126 / 192 ( 65%)
  ............................................................... 189 / 192 ( 98%)
  ...                                                             192 / 192 (100%)
```

So it is 192 executed tests (in this branch) instead of 1166 (in `master`). 

## How it impacts unit (fast) test suites

I tried to understand the impact for fast tests, running `master` branch and this PR against Infection itsetlf:

```bash
make test-infection-xdebug-80-docker
```

From my tests, running Infection against Infection becomes *slower* with this PR.

```
master:
  2m 52s. Memory: 0.17GB
  2m 47s. Memory: 0.17GB
  2m 46s. Memory: 0.17GB

PR:
  3m 8s. Memory: 0.19GB
  3m 6s. Memory: 0.19GB
  3m 9s. Memory: 0.19GB
```

* Why was memory usage increased (I don't think it's a problem though)? Because now we need to store all those `--filter '/the long list of test cases/' options
* Why was the execution time increased? From my tests, there are 2 reasons
  * Using `--filter` in PHPUnit itself adds some penalty as I understand, because this option is matched using `preg_match()` inside PHPUnit. But this adds only 5 extra seconds with the "Infection for Infection" example. 
  * Preparing a unique set of test cases and build `--filter '/the long list of test cases/' option takes the most extra time

Keeping that in mind, probably it's better to enable this feature by some option? If so, what the name of this new option you can recommend?

* `--only-covering-test-cases`
* `--slow-tests` (probably it will be some common option to enable not just this feature)
* ...

I don't know.

### Conclusion

The slower the tests, the more benefit / performance boost this PR will give.

---

- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/221

### DIFF
--- a/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
+++ b/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
@@ -35,17 +35,16 @@ declare(strict_types=1);
 
 namespace Infection\TestFramework\PhpUnit\CommandLine;
 
+use function array_key_exists;
 use function array_map;
 use function array_merge;
-use function array_unique;
 use function count;
 use function escapeshellcmd;
 use function explode;
-use function implode;
 use Infection\AbstractTestFramework\Coverage\TestLocation;
 use Infection\TestFramework\CommandLineArgumentsAndOptionsBuilder;
 use function ltrim;
-use function Safe\sprintf;
+use function rtrim;
 
 /**
  * @internal
@@ -71,18 +70,33 @@ final class ArgumentsAndOptionsBuilder implements CommandLineArgumentsAndOptions
         return $options;
     }
 
+    /**
+     * @param TestLocation[] $tests
+     */
     public function buildForMutant(string $configPath, string $extraOptions, array $tests): array
     {
         $options = $this->buildForInitialTestsRun($configPath, $extraOptions);
 
         if (count($tests) > 0) {
-            $escapedTests = array_map(
-                static fn (TestLocation $testLocation): string => escapeshellcmd($testLocation->getMethod()),
-                $tests
-            );
+            $filterString = '/';
+            $usedTestCases = [];
+
+            foreach ($tests as $testLocation) {
+                $testCaseString = $testLocation->getMethod();
+
+                if (array_key_exists($testCaseString, $usedTestCases)) {
+                    continue;
+                }
+
+                $usedTestCases[$testCaseString] = true;
+
+                $filterString .= escapeshellcmd($testCaseString) . '|';
+            }
+
+            $filterString = rtrim($filterString, '|') . '/';
 
             $options[] = '--filter';
-            $options[] = sprintf('/%s/', implode('|', array_unique($escapedTests)));
+            $options[] = $filterString;
         }
 
         return $options;
--- a/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
+++ b/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
@@ -39,11 +39,13 @@ use function array_key_exists;
 use function array_map;
 use function array_merge;
 use function count;
+use function escapeshellarg;
 use function escapeshellcmd;
 use function explode;
 use Infection\AbstractTestFramework\Coverage\TestLocation;
 use Infection\TestFramework\CommandLineArgumentsAndOptionsBuilder;
 use function ltrim;
+use function preg_quote;
 use function rtrim;
 
 /**
@@ -97,7 +99,7 @@ final class ArgumentsAndOptionsBuilder implements CommandLineArgumentsAndOptions
 
                 $usedTestCases[$testCaseString] = true;
 
-                $filterString .= escapeshellcmd($testCaseString) . '|';
+                $filterString .= preg_quote($testCaseString, '/') . '|';
             }
 
             $filterString = rtrim($filterString, '|') . '/';
--- a/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
+++ b/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
@@ -45,6 +45,7 @@ use function implode;
 use Infection\AbstractTestFramework\Coverage\TestLocation;
 use Infection\TestFramework\CommandLineArgumentsAndOptionsBuilder;
 use function ltrim;
+use function Safe\sprintf;
 
 /**
  * @internal
@@ -74,8 +75,6 @@ final class ArgumentsAndOptionsBuilder implements CommandLineArgumentsAndOptions
     {
         $options = $this->buildForInitialTestsRun($configPath, $extraOptions);
 
-//                preg_replace('/\swith data set (.*)/', '', $test->getMethod())
-
         if (count($tests) > 0) {
             $escapedTests = array_map(
                 static fn (TestLocation $testLocation): string => escapeshellcmd($testLocation->getMethod()),
@@ -83,7 +82,7 @@ final class ArgumentsAndOptionsBuilder implements CommandLineArgumentsAndOptions
             );
 
             $options[] = '--filter';
-            $options[] = implode('|', array_unique($escapedTests));
+            $options[] = sprintf('/%s/', implode('|', array_unique($escapedTests)));
         }
 
         return $options;
--- a/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
+++ b/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
@@ -39,8 +39,6 @@ use function array_key_exists;
 use function array_map;
 use function array_merge;
 use function count;
-use function escapeshellarg;
-use function escapeshellcmd;
 use function explode;
 use Infection\AbstractTestFramework\Coverage\TestLocation;
 use Infection\TestFramework\CommandLineArgumentsAndOptionsBuilder;
--- a/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
+++ b/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
@@ -37,7 +37,12 @@ namespace Infection\TestFramework\PhpUnit\CommandLine;
 
 use function array_map;
 use function array_merge;
+use function array_unique;
+use function count;
+use function escapeshellcmd;
 use function explode;
+use function implode;
+use Infection\AbstractTestFramework\Coverage\TestLocation;
 use Infection\TestFramework\CommandLineArgumentsAndOptionsBuilder;
 use function ltrim;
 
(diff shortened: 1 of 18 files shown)


sanmai:
Suggestion: if we can know a certain test suite contains this much tests (a thousand) which take this much seconds to run, we can make an educated guess as to whenever to enable this new option.

maks-rafalko:
> Suggestion: if we can know a certain test suite contains this much tests (a thousand) which take this much seconds to run, we can make an educated guess as to whenever to enable this new option.

thousand of tests does not mean they are slow, to be honest. 1000 unit tests can take less than a second, while 1000 functional tests can take minutes. 

Probably we need to calculate the total time of the covering tests and compare with some constant. I tried it, and enabled `--filter` option dynamically, but my results was very strange or inaccurate. Probably will give it a one more try

src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
```diff
@@ -64,4 +69,36 @@ public function build(string $configPath, string $extraOptions): array
...
     }
+
+    /**
+     * @param TestLocation[] $tests
+     */
+    public function buildForMutant(string $configPath, string $extraOptions, array $tests): array
+    {
+        $options = $this->buildForInitialTestsRun($configPath, $extraOptions);
+
+        if (count($tests) > 0) {
```
sanmai, line 80: I don't think we can take just the number of tests, and test suite time wouldn't necessary give the whole picture. I'd say median test time multiplied on the number of tests will give a better estimate to the time measure because median naturally filters outliers. Or we can go full with [three standard deviations of the mean](https://en.wikipedia.org/wiki/68%E2%80%9395%E2%80%9399.7_rule), which should give possibly even better estimate.

src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
```diff
@@ -64,4 +69,36 @@ public function build(string $configPath, string $extraOptions): array
...
+                if (array_key_exists($testCaseString, $usedTestCases)) {
+                    continue;
+                }
+
+                $usedTestCases[$testCaseString] = true;
+
+                $filterString .= escapeshellcmd($testCaseString) . '|';
+            }
+
+            $filterString = rtrim($filterString, '|') . '/';
```
sanmai, line 96: Are we sure we absolutely need this computation? I mean, is there a chance we won't run this mutant ever? And what about mutants on the same line of the same file?

maks-rafalko: didn't get you, could you please clarify what do you mean?

This code just builds `'/App\\\\Test::test_case1|App\\\\Test::test_case2/'` line, meaning to run only 2 test cases of X. Those tests cases - is 100% of all the test cases that cover particular mutated line (Mutant)

Ocramius:
@maks-rafalko I'm re-reading the description of this issue, and I'm trying to make sense of it.

Just because I'm trying to understand, if I have a test like following:

```php
class MyTest ... {
   function test1() { ... }
   function test2() { ... }
}
```

Does this patch make sure that only either of `test1` or `test2` are run for a specific mutation, if they cover separate code section? Did infection run all of `MyTest` before this patch?

maks-rafalko:
> Does this patch make sure that only either of `test1` or `test2` are run for a specific mutation, if they cover separate code section? Did infection run all of `MyTest` before this patch?

@Ocramius  Exactly. Currently (in `master`), Infection creates a custom `phpunit.xml` for each mutant, and adds **test files** that cover mutated line, not particular test cases inside those files:


```xml
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/schema/9.2.xsd" bootstrap="/infection/tests/e2e/PestTestFramework/./infection/interceptor.autoload.740af9ed4589f3494f0c14ed544eec29.infection.php" colors="false" stopOnFailure="true" cacheResult="false" stderr="false">
  <testsuites>
    <testsuite name="Infection testsuite with filtered tests">
      <file>/infection/tests/e2e/PestTestFramework/tests/CalculatorPhpUnitTest.php</file>
    </testsuite>
  </testsuites>
  <filter>
    <whitelist>
      <directory>/infection/tests/e2e/PestTestFramework/src/</directory>
    </whitelist>
  </filter>
</phpunit>
```

^ as I described in the description, `CalculatorPhpUnitTest.php` test file can have 1000 tests cases, but only 10 cover mutated line. So, we need `--filter 'test_case_1|test_case_2...|test_case_10` to reduce the number of executed tests.

Ocramius:
Excellent++! This is gonna be literal HOURS of mutation test runs saved. For me, the largest mutation test suite (integration tests) takes 2h :D

maks-rafalko:
Yes, this is what I'm experiencing on my test suite as well with 2000+ functional tests ;)

sanmai:
> I tried it, and enabled `--filter` option dynamically, but my results were very strange or inaccurate. Probably will give it one more try

Yes, it is all not that simple, unfortunately. I think this is worthy a dedicated research project, and an option toggle like `--only-covering-test-cases` should be the best course of action now. We can add an auto toggle later.

maks-rafalko:
@sanmai ready for review. I've added `--only-covering-test-cases` option for now as we discussed above, so this new feature will be disabled by default and can be enabled manually

sanmai (approved):
LGTM

maks-rafalko:
Thank you @sanmai for the review

nicojs:
Q: can the performance degradation on the "infection on infection" usecase also be influenced by simply having more tests in this branch? Or did you correct for that in some way?

maks-rafalko:
I avoided the impact of mutating the new code and tests by adding `@infection-ignore-all` on the new code, and commenting out the tests. 

The most impact is really because of building the very long strings for each mutant where running all unit tests is faster than creating this filter

Issue #1524, opened 2021-06-22
Empty diff shown


mimmi20:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.23.0
| Test Framework version | PHPUnit 9.5.5
| PHP version | 7.4
| Platform    | Ubuntu 20.04 (Github Action)
| Github Repo | https://github.com/mimmi20/mezzio-navigation-laminasviewrenderer-bootstrap/pull/36/commits/9fe605b99dd8cfc0a44ba481d99dd4e771e0d837

During a Build in a Github Action an empty diff was shown in the console output and in the Github Action.

![image](https://user-images.githubusercontent.com/1120192/122995220-9bb20100-d3a9-11eb-8ff0-1d9ffdc0aabc.png)
![image](https://user-images.githubusercontent.com/1120192/122994517-d8312d00-d3a8-11eb-806b-818228818843.png)

The line of code was 

```php
$html .= $this->getIndent() . $this->getIndent() . '<ul class="breadcrumb">' . PHP_EOL;
```

With the Parameter "-s" , I could see on the console output that the "concat" Mutator was used. I asume that the Mutator tried to swap the both "$this->getIndent()" pieces.

It would be helpful, if the Github-Log-Output would include the used Mutator, when the Option "-s" is set.

PR #1540, opened 2021-07-20
Add Mutator information to GitHub annotation logger

maks-rafalko:
* It will give more information about what Mutator is used and will be in line with other loggers
* It will make it clear to understand why the diff can be empty, like in https://github.com/infection/infection/issues/1524

Fixes https://github.com/infection/infection/issues/1524


sanmai (approved):

maks-rafalko:
thank you, @sanmai

PR #1544, opened 2021-08-05
Allow Infection test suite to be executed in parallel using Paratest

maks-rafalko:
2 goals here:

1. This is much faster (5 seconds for `vendor/bin/paratest` instead of 24 seconds for `vendor/bin/phpunit`)
2. This is the first step for supporting new Test Framework Adapter - Paratest, to run initial tests suite run using `paratest` instead of `phpunit`

For now, just for the local usage

### DIFF
--- a/src/Command/RunCommand.php
+++ b/src/Command/RunCommand.php
@@ -485,7 +485,7 @@ final class RunCommand extends BaseCommand
     ): void {
         $locator = $container->getRootsFileOrDirectoryLocator();
 
-        if ($customConfigPath = (string) $io->getInput()->getOption(self::OPTION_CONFIGURATION)) {
+        if (($customConfigPath = (string) $io->getInput()->getOption(self::OPTION_CONFIGURATION)) !== '') {
             $locator->locate($customConfigPath);
         } else {
             $this->runConfigurationCommand($locator, $io);
--- a/src/Process/Factory/InitialTestsRunProcessFactory.php
+++ b/src/Process/Factory/InitialTestsRunProcessFactory.php
@@ -38,7 +38,9 @@ namespace Infection\Process\Factory;
 use Infection\AbstractTestFramework\TestFrameworkAdapter;
 use Infection\Process\OriginalPhpProcess;
 use function method_exists;
+use PackageVersions\Versions;
 use Symfony\Component\Process\Process;
+use function version_compare;
 
 /**
  * @internal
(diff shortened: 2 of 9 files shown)


src/Command/RunCommand.php
```diff
@@ -485,7 +485,7 @@ private function startUp(
     ): void {
         $locator = $container->getRootsFileOrDirectoryLocator();
 
-        if ($customConfigPath = (string) $io->getInput()->getOption(self::OPTION_CONFIGURATION)) {
+        if (($customConfigPath = (string) $io->getInput()->getOption(self::OPTION_CONFIGURATION)) !== '') {
```
maks-rafalko, line 488: this is because of the upgraded `symfony/console`, PHPStan complained

tests/phpunit/FileSystem/FileSystemTestCase.php
```diff
@@ -84,9 +86,16 @@ protected function tearDown(): void
 
     final protected static function removeTmpDir(): void
     {
+        $testToken = getenv('TEST_TOKEN');
+
         (new Filesystem())->remove(
             normalizePath(
-                realpath(sys_get_temp_dir()) . '/' . self::TMP_DIR_NAME
+                sprintf(
```
maks-rafalko, line 93: when tests are executed on parallel, different threads started to remove the temp folders of another thread. We need to have different folders for each thread, which is done here

src/Process/Factory/InitialTestsRunProcessFactory.php
```diff
@@ -77,7 +79,7 @@ public function createProcess(
 
         $process->setTimeout(null); // Ignore the default timeout of 60 seconds
 
-        if (method_exists($process, 'inheritEnvironmentVariables')) {
+        if (method_exists($process, 'inheritEnvironmentVariables') && version_compare(Versions::getVersion('symfony/console'), 'v4.4', '<')) {
```
sidz, line 82: why do we need to have an additional condition which uses ocramius/package-versions here? does `method_exists` not cover our case?

maks-rafalko: Tried to explain it in the commit message:

> Call `$process->inheritEnvironmentVariables();` only when it exists AND is not deprecated
> Otherwise, it does not make sense because starting from `syfmony/console` 4.4 environment variable are inhereted automatically, and calling deprecated methods fails the tests

We have a job that does `composer install ... --prefer-lowest`. After some dependencies upgrade, this job started installing `symfony/console:4.4.0` instead of `symfony/console:3.*`.

In `symfony/console:4.4.*` the method  `inheritEnvironmentVariables()` exists, but calling it produces a deprecation notice, which [failed the build](https://github.com/infection/infection/runs/3257362627#step:7:134)

So, technically we need to call this method when it exists and and it does something, but in `4.4.*` id does nothing (see the deprecation message: `...env variables are always inherited`).

The final result

* < 4.4.0 - `inheritEnvironmentVariables()` method exists and we should call it
* = `4.4.*` - `inheritEnvironmentVariables()` method exists but we should not call it
* > `4.4.*` - `inheritEnvironmentVariables()`method does not exist

sidz: ah. I've forgot that we still support symfony packages < 4.4.

Issue #1519, opened 2021-05-10
Speed up Infection runs by remembering which test killed a mutant


MidnightDesign:
As @maks-rafalko pointed out to me on Discord, Infection currently runs the tests for a mutant in order from fastest to slowest, but if the killing test is the slowest, it will always be run last.

It might be beneficial for Infection to remember (i.e. save in a file on disk, for example) which test killed a mutant so it can run that test first the next time. If that test wasn't run the last time or or if it doesn't kill the mutant, it could fall back to the current system.

This strategy would be most beneficial in development environments, but it could potentially also be used in CI by committing the mutant -> test map to the repository.

maks-rafalko:
@MidnightDesign feel free to experiment with the first implementation https://github.com/infection/infection/pull/1549

maks-rafalko:
Implemented in #1549

PR #1547, opened 2021-08-07
For Mutant's phpunit.xml, set `executionOrder="default"` to prevent random ordering of the tests since we need them to be sorted (fastest - first)

maks-rafalko:
During the work on https://github.com/infection/infection/issues/1519, found that we have a critical bug. We always said we run the tests from fastest to slowest, hoping that the fastest tests fail first. It's true, unless the original `phpunit.xml` has an `executionOrder=random` attribute set (closest example - Infection's test suite itself).

---

When Infection builds Mutant's `phpunit.xml` file, we add `<file>` tags, sorted by tests execution time, so that the fastest tests runs first.

But when the original `phpunit.xml` contains `executionOrder="random"`, it was inherited to Mutant's `phpunit.xml` and broke the order, running the tests in a random order each time `infection` was executed.

This also explains why I got different execution time of Infection for the project with *slow functional* tests (there is the same issue for unit tests, but they are fast enough so this issue is not visible).

When Mutant is covered by 1000 functional tests, and tests are executed by random, it can be killed by the first test in run `#1` and can be killed by 1000-th test in run `#2`, if the killing test is executed the last.

With this update, all the tests will be executed in the same orders (fastest - first) in all the infection runs.

For more clarity, this is what happens in `master` right now with executing the same command 2 times:

```bash
'phpunit' '--configuration' 'infection/phpunitConfiguration.96749d7e4a15eec4f82a2413a56b5fba.infection.xml
```

`#1` run:

```bash
PHPUnit 9.5.7 by Sebastian Bergmann and contributors.

Random Seed:   1628355728

Testing 
.............................................................   61 / 1404 (  4%)
.............................................................  122 / 1404 (  8%)
.............................................................  183 / 1404 ( 13%)
.............................................................  244 / 1404 ( 17%)
.............................................................  305 / 1404 ( 21%)
.............................................................  366 / 1404 ( 26%)
.............................................................  427 / 1404 ( 30%)
.............................................................  488 / 1404 ( 34%)
E

Time: 01:05.354, Memory: 459.00 MB

There was 1 error:
```

`#2` run:

```bash
PHPUnit 9.5.7 by Sebastian Bergmann and contributors.

Random Seed:   1628355731

Testing 
.............................................................   61 / 1404 (  4%)
.............................................................  122 / 1404 (  8%)
.............................................................  183 / 1404 ( 13%)
.............................................................  244 / 1404 ( 17%)
..E

Time: 00:49.420, Memory: 349.00 MB

There was 1 error:
```

^ you can see that tests were executed in a random order, and the killing test was 499th in the `#1` run, and 245th in the `#2` run

### DIFF
--- a/src/TestFramework/AbstractTestFrameworkAdapter.php
+++ b/src/TestFramework/AbstractTestFrameworkAdapter.php
@@ -154,7 +154,8 @@ abstract class AbstractTestFrameworkAdapter implements TestFrameworkAdapter
             $tests,
             $mutantFilePath,
             $mutationHash,
-            $mutationOriginalFilePath
+            $mutationOriginalFilePath,
+            $this->getVersion()
         );
     }
 
--- a/src/TestFramework/Config/MutationConfigBuilder.php
+++ b/src/TestFramework/Config/MutationConfigBuilder.php
@@ -55,7 +55,8 @@ abstract class MutationConfigBuilder
         array $tests,
         string $mutantFilePath,
         string $mutationHash,
-        string $mutationOriginalFilePath
+        string $mutationOriginalFilePath,
+        string $version
     ): string;
 
     protected function getInterceptorFileContent(string $interceptorPath, string $originalFilePath, string $mutantFilePath): string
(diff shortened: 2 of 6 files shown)


sidz (approved):
:+1:

PR #1548, opened 2021-08-07
Make the `--filter` option shorter for PHPUnit

maks-rafalko:
This is just an attempt to make the `--filter` shorter. It doesn't fix https://github.com/infection/infection/issues/1545 though.

Explanation:

Difference between `master` and this PR with built `--filter`:

```diff
- /App\\Tests\\Functional\\Api\\Document\\DocumentTemplateTest\:\:test_it_can_update_document_template with data set "Returns error when at leas one of the business programs not in the document type'\''s list"
+ /DocumentTemplateTest\:\:test_it_can_update_document_template with data set "Returns error when at leas one of the business programs not in the document type'\''s list"
```

We can get rid of all the `App\\Tests\\Functional\\Api\\Document\\` and similar prefixes, leaving just `ClassTest::test_case_method` notation (or `ClassTest::test_case_method with data set "#1"` if data provider is used).

Why not just `test_case_method1|test_case_method2`? Because the less unique the filter, the more tests will be executed (which is worse from performance POV). If we have 1000 files with `test_case_method1`, then all the 1000 tests will be executed. Prefixing `test_case_method1` with `ClassTest::` make it hopefully unique.

On my real project test, this PR saves 13,000 characters from 50,000 and saves the performance (runs the same amount of tests, not more).

### DIFF
--- a/tests/phpunit/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilderTest.php
+++ b/tests/phpunit/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilderTest.php
@@ -142,7 +142,7 @@ final class ArgumentsAndOptionsBuilderTest extends TestCase
             [
                 'App\Test::test_case1',
             ],
-            '/App\\\\Test\:\:test_case1/',
+            '/Test\:\:test_case1/',
         ];
 
         yield '2 test cases' => [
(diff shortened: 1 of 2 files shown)


sanmai:
Here's what we can do: 
- We can have several optimization strategies for `--filter`. For example, we can go the way you propose here, or we can use the original approach, or we can go `test_case_method1|test_case_method2`.
- Next we can run a cost/benefit for each of the strategies. For example, if the original strategy leads to a regex shorter than, say, 300 characters, we could probably just use it and not consider anything else.
- We know how much time it takes to run every test, and we can tell how much time it would take to run the most optimal set of tests. Next we can check if there's a substantial difference between different remaining strategies. For example, if the "shortest" strategy such as `test_case_method1|test_case_method2` leads to a test duration within 1% of the optimal time, we can probably just use the shortest.
- If we didn't find the optimal strategy, we can optimize by finding a strategy with the shortest resulting regex (we can probably predict its length) that has the least amount of the deviation from the optimum.
 
Do you think this will be too much trouble to make it worth it?

theofidry:
@sebastianbergmann would there be an alternative solution? For example setting `--filters` as a possible array and if there is multiple entries then all must match?

sebastianbergmann:
Just to make sure I understand your motivation correctly: you would like to define a list of tests to run? Then, I think, finally implementing https://github.com/sebastianbergmann/phpunit/issues/3387 would be the way to go. That, however, would be a new feature that would only be available in PHPUnit >= 10.

theofidry:
>That, however, would be a new feature that would only be available in PHPUnit >= 10

That's totally fine: whichever solution is implemented in PHPUnit will be in a newer version and infection will most likely always have a wider range of PHPUnit versions to support.

I think @maks-rafalko solutions is a good one as it tries to limit the issue but it's a tiny patch at best. @sanmai is also a viable solution but it feels (IMO) like adding a lot of brittle (it tries to find a _compromise_) code for something that IMO should just be possible without workarounds.

@maks-rafalko @sanmai maybe another patch would be to simply remove `--filter` and add a warning in the case where the length could not be reduced enough. It will result in a big performance degradation in those cases but IMO it is better than a failure/false result

maks-rafalko:
@sanmai @theofidry I've added your ideas for further implementing to here https://github.com/infection/infection/issues/1545#issuecomment-912941104, and will keep that issue open to try improve this filtering farther. For now will merge this small patch and test it on my daily job project with the upcoming 0.25.0 release

theofidry:
👍

tests/phpunit/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilderTest.php
```diff
@@ -151,25 +151,50 @@ public function provideTestCases(): Generator
...
+            '/Test\:\:test_case1 with data set "With special character @"|Test\:\:test_case2/',
+        ];
+
+        yield '2 data sets from data provider for the same test case' => [
+            true,
+            [
+                'App\Test::test_case1 with data set "#1"',
+                'App\Test::test_case1 with data set "#2"',
+            ],
+            '/Test\:\:test_case1 with data set "\#1"|Test\:\:test_case1 with data set "\#2"/',
```
staabm, line 181: do I get it right, that we are running

`phpunit --filter='Test::test_case1 with data set "With special character @"'`

?

if so, I think we can shorten this even more. PHPUnit supports these notations:

`--filter "testSuiteSorterDefectsWithDataProviderTest::testOne#1"`
`--filter "testSuiteSorterDefectsWithDataProviderTest::testOne@data-set"`

see e.g. https://github.com/sebastianbergmann/phpunit/pull/6272
(This PR did not introduce this filter feature.. it only adjusted the output rendering)

meaning, we don't need this `with data set` in the pattern

maks-rafalko: > do I get it right, that we are running

yes. And we do it for PHPUnit <10 and PHPUnit >=10, see

https://github.com/infection/infection/blob/5392be12640e90a3f6f2f3f6a73b1f32b746f562/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php#L176-L181

if you think we can do it shorter, great! let's extract to a separate issue.

related

- https://github.com/infection/infection/issues/1825#issuecomment-1423104126
- https://github.com/infection/infection/pull/1854
- https://github.com/infection/infection/issues/1899 and the fix https://github.com/infection/infection/pull/1900

Issue #1542, opened 2021-08-02
No line of code was executed during tests.


alexdodonov:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.21.5
| Test Framework version | PHPUnit 8.5.19
| PHP version | 7.4.4
| Platform    | Windows
| Github Repo | https://github.com/alexdodonov/mezon-router

Hi! I am getting error "No line of code was executed during tests. This could be due to `@covers` annotations or your  
  PHPUnit filters not being set up correctly."

Here is infection config:
```json
{
	"source": {
		"directories": [
			"./Mezon/Router"
		],
		"excludes": [
			"./Mezon/Router/Tests"
		]
	},
	"phpUnit": {
		"customPath": ".\/vendor\/phpunit\/phpunit\/phpunit"
	},
	"logs": {
		"text": ".\/tmp\/infection.txt"
	},
	"mutators": {
		"@default": true
	}
}
```

Here is PHPUnit config:

```xml
<?xml version="1.0" encoding="UTF-8"?>

<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/7.0/phpunit.xsd"
	backupGlobals="false" colors="true" bootstrap='./vendor/autoload.php'
	stopOnError="true" stopOnFailure="true">
	<php>
		<ini name="error_reporting" value="-1" />
	</php>

	<testsuites>
		<testsuite name="unit">
			<directory suffix="UnitTest.php">
				./Mezon/Router/Tests
			</directory>
		</testsuite>
	</testsuites>

	<filter>
		<whitelist>
			<directory>./Mezon/Router</directory>
			<exclude>
				<directory suffix="UnitTest.php">
					./Mezon/Router/Tests
				</directory>
				<directory>./vendor</directory>
				<file>autoload.php</file>
			</exclude>
		</whitelist>
	</filter>

	<logging>
		<log type="coverage-html"
			target="./tmp/coverage-phpunit.html"></log>
		<log type="coverage-text" target="php://stdout"
			showOnlySummary="true"></log>
	</logging>
</phpunit>
``` 

What do I do wrong?

sanmai:
Try running it with `XDEBUG_MODE=coverage`

alexdodonov:
It have not helped (

maks-rafalko:
Thank you very much for providing a repository with reproducible issue, this is how the bug report should look like!

[...]

PR #1550, opened 2021-08-09
Trim "\n" and "\t" characters when replacing relative paths with absolute ones during XML config creation

maks-rafalko:
Fixes https://github.com/infection/infection/issues/1542

When XML config contains "\n" or "\t" characters in the tag with paths, it breaks the path when Infection replaces relative paths with absolute ones.

Solution: just trim it.

Example of the problematic `phpunit.xml`:

https://github.com/alexdodonov/mezon-router/blob/dea1e597aba3dbd86148b9e9fc129048530c3280/phpunit.xml#L13-L15

Reason of the bug in debugger:

tabs-bug

### DIFF
--- a/tests/phpunit/TestFramework/PhpUnit/Config/XmlConfigurationManipulatorTest.php
+++ b/tests/phpunit/TestFramework/PhpUnit/Config/XmlConfigurationManipulatorTest.php
@@ -110,6 +110,36 @@ XML
         );
     }
 
+    public function test_it_replaces_with_absolute_paths_xml_file_with_tabs(): void
+    {
+        $this->assertItChangesXML(
+            <<<'XML'
+<phpunit cacheTokens="true">
+    <testsuites>
+		<testsuite name="All Tests">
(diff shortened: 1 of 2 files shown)


tests/phpunit/TestFramework/PhpUnit/Config/XmlConfigurationManipulatorTest.php
```diff
@@ -110,6 +110,36 @@ static function (XmlConfigurationManipulator $configManipulator, SafeDOMXPath $x
...
 
+    public function test_it_replaces_with_absolute_paths_xml_file_with_tabs(): void
+    {
+        $this->assertItChangesXML(
+            <<<'XML'
+<phpunit cacheTokens="true">
+    <testsuites>
+		<testsuite name="All Tests">
+			<directory suffix="UnitTest.php">
+				./Tests
```
maks-rafalko, line 121: it's not visible here, but `./Tests` is rounded by `\n\t`

sidz (approved):

PR #1554, opened 2021-08-11
Generate `infection.json` (without `.dist` postfix) by default

maks-rafalko:
Generate `infection.json` (without `.dist` postfix) because this is a json file and **any** IDE automatically understands it.

Currently, generated `.dist` file is not recognized as JSON file by PHPStorm, which makes it work like usual text file, where JSON autocompletion won't work

Related to https://github.com/infection/infection/pull/1553


sidz (approved):
shouldn't we rename our own `infection.json.dist` (https://github.com/infection/infection/blob/master/infection.json.dist) to just `infection.json` as part of this PR?

maks-rafalko:
Makes sense, thanks. Updated

PR #1555, opened 2021-08-12
Detect syntax errors during mutation analysis and differentiate them from all errors

maks-rafalko:
Implements https://github.com/infection/infection/issues/262

- [x] Release `infection/abstract-testframework-adapter` before merging

---

From the first look, I didn't want to implement it since we [ensure mutated code is always valid](https://github.com/infection/infection/pull/301), but in the future we will allow users to write their own Mutators and use them *in addition* to Infection's ones, and if users' Mutator produce invalid code - users should be aware of it, because Mutators that produce invalid syntax should be fixed and should not be used since this is a wasting of time/resources.


src/Mutator/ProfileList.php
```diff
@@ -464,6 +464,9 @@ final class ProfileList
         // Extensions
         'BCMath' => Mutator\Extensions\BCMath::class,
         'MBString' => Mutator\Extensions\MBString::class,
+
+        // Internal only usage
```
theofidry, line 468: `Internal usage only`

src/Mutator/SyntaxError.php
```diff
@@ -0,0 +1,81 @@
...
+namespace Infection\Mutator;
+
+use PhpParser\Node;
+
+/**
+ * @internal
+ *
+ * @implements Mutator<Node\Expr\Variable>
+ */
+final class SyntaxError implements Mutator
```
theofidry, line 45: From what i understand it's for testing only right? So why registering the mutator here in source code? I guess we don't have (yet) a way to register mutators only the fly for tests only?

tests/phpunit/TestFramework/PhpUnit/Adapter/PestAdapterTest.php
```diff
@@ -275,6 +287,21 @@ public function outputProvider(): iterable
         yield ['Tests:  1 failed, 8 passed', false];
     }
 
+    public function syntaxErrorOutputProvider(): iterable
+    {
+        yield ['Tests:  1 risked', false];
+
+        yield [
+            <<<TXT
+'ParseError
```
theofidry, line 296: Just in case: you can now indent the heredocs: https://wiki.php.net/rfc/flexible_heredoc_nowdoc_syntaxes

I would also recommend to use nowdocs vs heredocs when possible (i.e. quote `TXT` into `'TXT'`)

theofidry (commented):
Tiny nitpicks, looks good!

src/Mutator/SyntaxError.php
maks-rafalko: it's for testing only, yes

as far as I know, we can't register mutators on the fly atm. If we remove `SyntaxError` from `ProfileList.php`, tests fail with "Uknown Mutator" error raised here

https://github.com/infection/infection/blob/b07e053857b114b312c915aee869392f215179d8/src/Mutator/MutatorFactory.php#L59-L65

src/Mutator/ProfileList.php
maks-rafalko: fixed in https://github.com/infection/infection/pull/1557

tests/phpunit/TestFramework/PhpUnit/Adapter/PestAdapterTest.php
maks-rafalko: fixed in https://github.com/infection/infection/pull/1557

tests/phpunit/Logger/JsonLoggerTest.php
```diff
@@ -102,16 +104,17 @@ public function metricsProvider(): iterable
...
                     'errorCount' => 2,
+                    'syntaxErrorCount' => 2,
                     'skippedCount' => 2,
                     'timeOutCount' => 2,
-                    'msi' => 60,
-                    'mutationCodeCoverage' => 80,
-                    'coveredCodeMsi' => 75,
+                    'msi' => 50,
+                    'mutationCodeCoverage' => 83.33,
+                    'coveredCodeMsi' => 60,
```
sanmai, line 117: Well, this is wrong. If Infection caused an arbitrary program to end with a syntax error, this was a killed mutant. Just like if mutation leads to a type error.

PR #1576, opened 2021-09-13
Make GitHub annotation work when Infection is run in subdirectory

kubawerlos:
When Infection is run in Git's repository root (with `./backend/vendor/bin/infection`) then logged warnings are fine:
> ::warning file=backend/src/Foo.php,line=42::Escaped Mutant for Mutator (...)

However, when Infection is run in subdirectory (`backend`, with command `./vendor/bin/infection`) then logged warnings are like this:
> ::warning file=src/Foo.php,line=42::Escaped Mutant for Mutator (...)

which makes Github annotation not working.

This change fixes that problem.


maks-rafalko (approved):
Thank you @kubawerlos

PR #1583, opened 2021-09-20
Do not use ocramius/package-versions (and do not break building PHAR this time)

kubawerlos:
This PR isn't actually a feature, but revert of [Revert "Do not use ocramius/package-versions"](https://github.com/infection/infection/pull/1581) with fix for building PHAR.


scoper.inc.php
```diff
@@ -42,7 +42,7 @@
 
 return [
     'whitelist' => [
-        \Composer\Autoload\ClassLoader::class,
+        'Composer\*',
```
maks-rafalko, line 45: that's crazy, I tried 

image

and it didn't work, so I gave up :|

maks-rafalko (approved):
Thank you @kubawerlos

scoper.inc.php
kubawerlos: Yeah, I tried that too, also some other random class and was prefixed. There is already [issue](https://github.com/humbug/php-scoper/issues/532) reported.

maks-rafalko:
Unfortunately, had to revert it again, see https://github.com/infection/infection/pull/1586

and 

https://github.com/infection/infection/pull/1586/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR116

so before revert, when we ran `build/infection.phar` - it worked, but after `cd build` and running `./infection.phar` - there was en error (see that PR).

I've updated CI to catch this issue by the line above

PR #1594, opened 2021-10-13
Update PHPStan

kubawerlos:
Already some fixes, and after this continue fixing the issues (especially the one with `InvalidArgumentException` - there is something wrong with declared exceptions and there is unfortunately no test for that code).


maks-rafalko:
Thank you, @kubawerlos

PR #1606, opened 2021-11-30
Allow Symfony 6

julien-boudry:


sanmai (approved):

sidz (commented):
IMO we need to test Infection against all supported versions.

maks-rafalko (commented):
> - The lock file is not up to date with the latest changes in composer.json, it is recommended that you run `composer update` or `composer update <package name>`.

to fix the failed build, please do `composer update --lock`

@julien-boudry and everyone interested: you can use PHAR distribution instead, unless this one is released. PHAR distribution is prefixed and won't have any conflicts with your Symfony application, including 6.0

maks-rafalko:
> IMO we need to test Infection against all supported versions.

agree

let's merge it as is to allow people to upgrade and use it on their own risk for now :)

PR #1611, opened 2021-12-10
Do not run Infection for itself on windows

maks-rafalko:
* It takes too much time (26m vs 6-7m on Ubuntu)
* We already test Infection on Windows in `ci.yaml` in e2e tests

As a result, the whole build will take 6-7 minutes instead of 26 minutes


sanmai (approved):
Yes, this was a mistake

PR #1619, opened 2021-12-21
Treat all function signature nodes as 1-line range to avoid duplicate TestLocations

maks-rafalko:
Finally, caught a bug that leads to a massive `TestLocation[]` duplications. For the very simple example - our [e2e `Example_Test`](https://github.com/infection/infection/tree/master/tests/e2e/Example_Test) - we previously had 4 duplicate objects of `TestLocation` class because of this bug.

infection-non-unique-bug

The more lines a method has, the more duplicates of `TestLocation` objects mutant has.

how to reproduce:

Add a breakpoint here:

https://github.com/infection/infection/blob/af58050714ef00c4c624846a2c3e7a9fb6cea6f9/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php#L89

then

```bash
cd tests/e2e/ExampleTest

INFECTION_ALLOW_XDEBUG=1 XDEBUG_TRIGGER=yes ../../../bin/infection
```

and you will see 4 duplicated test locations for just 1 mutant.

---

### Explanation

Previously, `Node` of class `ClassMethod` (mutation `public -> protected`) returned a range of lines for the whole method. Consider the following code:

```php
class Test {
    public function findMe() // line 4
    {
        // ...
    } // line 7
}
```

for this code, `ClassMethod` node returned a [4, 7] range instead of [4, 4].

Because of this, later *for each line* (4, 5, 6, 7) we found all the test cases that cover these lines.

If we have 1 test case `TestClass::test_method_find_me` that covers `findMe()` method, then we returned 4 exactly the same duplicate test cases for `ClassMethod` mutation, one for each of (4, 5, 6, 7) lines.

For the real-world examples, if we have a method with 10 lines of code that are covered by 3 tests, we have `3 * 10 = 30` `TestLocation`s instead of just 3.

After this update, only 1 set of test cases returned for the function signature mutant

### DIFF
--- a/tests/phpunit/TestFramework/Coverage/ProxyTraceTest.php
+++ b/tests/phpunit/TestFramework/Coverage/ProxyTraceTest.php
@@ -178,7 +178,7 @@ final class ProxyTraceTest extends TestCase
         $this->assertCount(
             1,
             iterator_to_array($trace->getAllTestsForMutation(
-                new NodeLineRangeData(19, 22),
+                new NodeLineRangeData(19, 19),
                 true
             ))
         );
--- a/src/TestFramework/Coverage/LineRangeCalculator.php
+++ b/src/TestFramework/Coverage/LineRangeCalculator.php
@@ -36,6 +36,7 @@ declare(strict_types=1);
 namespace Infection\TestFramework\Coverage;
 
 use Infection\PhpParser\Visitor\ParentConnector;
+use Infection\PhpParser\Visitor\ReflectionVisitor;
 use PhpParser\Node;
 
 /**
@@ -47,7 +48,11 @@ final class LineRangeCalculator
     {
         $node = $this->getOuterMostArrayNode($node);
 
-        return new NodeLineRangeData($node->getStartLine(), $node->getEndLine());
+        $endLine = $node->getAttribute(ReflectionVisitor::IS_ON_FUNCTION_SIGNATURE, false) === true
+            ? $node->getStartLine() // function signature node should always be 1-line range: (start, start)
+            : $node->getEndLine();
+
+        return new NodeLineRangeData($node->getStartLine(), $endLine);
     }
 
     /**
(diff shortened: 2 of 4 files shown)


tests/phpunit/TestFramework/Coverage/ProxyTraceTest.php
```diff
@@ -178,7 +178,7 @@ public function test_it_exposes_its_test_locations(): void
         $this->assertCount(
             1,
             iterator_to_array($trace->getAllTestsForMutation(
-                new NodeLineRangeData(19, 22),
+                new NodeLineRangeData(19, 19),
                 true
```
maks-rafalko, line 182: now - incorrect test. When the mutation is on the function signature, it must be 1-line range

sanmai (approved):
Did you find material improvements from fixing this issue?

maks-rafalko:
> Did you find material improvements from fixing this issue?

For running Infection for itself, the only difference that I see is a small memory improvement:

```diff
- Time: 3m 8s. Memory: 0.26GB
+ Time: 3m 8s. Memory: 0.24GB
```
However, there is a difference in how much work needs to be done to prevent the consequences of this issue. I have 2 places in mind:

1. Building `--filter=test_a|test_b|...` option

  In this piece of code, back in that time I had to manually filter out duplications

  https://github.com/infection/infection/blob/af58050714ef00c4c624846a2c3e7a9fb6cea6f9/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php#L93-L116

  See line `103` - `if (array_key_exists($testCaseString, $usedTestCases)) { `. Now, this loop will iterate on a much less amount of tests

2. Building unique set of tests for HTML report
  
  This is WIP feature, but I already had to filter out duplicates exactly because of this issue, see this code https://github.com/infection/infection/commit/dcc2b6c6e1c4ac2eb24c3b57f34c451940005c2b#diff-01143bfb1cd8c30fa7f368e58fa5b9bebf3f35de2a4bc509123ed81358b59ff8R105-R115

  Again, now this work won't be needed (or if needed - amount of duplicates will be much less).

sanmai:
> In this piece of code, back in that time I had to manually filter out duplications

What if we would fail with a warning if we ever see duplicates there moving forward?

maks-rafalko:
yeah, probably

I will continue investigating duplicates because I rebased "html report" branch on this new commit in master and there are still some duplicates

I will get back with the results

maks-rafalko:
> I will get back with the results

extended the fix: https://github.com/infection/infection/pull/1622

Issue #1620, opened 2021-12-21
False Positive (not finding deprecated warning on Implicit conversion from float to int)


navarr:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.25.5 (`infection.phar --version`)
| Test Framework version | PHPUnit 9.5.10
| PHP version | 8.1.0
| Platform    | Ubuntu LTS 20.04.3 (Github Actions)
| Github Repo | navarr/binary-explode

Github Actions run: https://github.com/navarr/binary-explode/runs/4596804439?check_suite_focus=true

* Problem observed: On PHP 8.1.0, an implicit conversion from a float to an int raises a Deprecated warning.  [3v4l](https://3v4l.org/C74oo#v8.1.0)
* This is covered in my logarithmic binary exploder by casting the output of `log` to an `int` [Line](https://github.com/navarr/binary-explode/blob/main/src/LogarithmicBinaryExploder.php#L22)
* My automated tests contains a test that would trigger this deprecation when run [Line](https://github.com/navarr/binary-explode/blob/main/tests/LogarithmicBinaryExploderTest.php#L49) [3v4l](https://3v4l.org/C74oo#v8.1.0)
* Infection does not detect this deprecation and so the mutant removing the cast escapes

<details>
 <summary>phpunit.xml</summary>

 ```xml
<?xml version="1.0" encoding="utf-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" colors="true" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd">
    <testsuite name="unit">
        <directory>tests</directory>
    </testsuite>
    <coverage cacheDirectory="/tmp">
        <include>
            <directory suffix=".php">src</directory>
        </include>
    </coverage>
</phpunit>
 ```
</details>

<details>
 <summary>Output with issue</summary>

 ```

    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/

Infection - PHP Mutation Testing Framework version 0.25.5


Running initial test suite...

PHPUnit version: 9.5.10

[...]

PR #1621, opened 2021-12-21
Mark Mutant as killed if Test Framework returns non-zero exit code

maks-rafalko:
PHPUnit can return `"Tests passed! OK (10 tests, 32 assertions)"` output, however return code will be non-zero.

This happens when, for example, `symfony/phpunit-bridge` is used with `SYMFONY_DEPRECATIONS_HELPER=max[self]=0`, and it detects outstanding deprecations which fails PHPUnit execution, while all the tests are passing

See https://github.com/infection/infection/issues/1620#issuecomment-999073728

Now, when Test Framework exit code is > 100, Mutant will be treated as `E`rrored (retained behavior).

When Test Framework exit code is non-zero, Mutant will be treated as Killed (new behavior), even if from output's point of view tests pass.

### DIFF
--- a/src/Mutant/MutantExecutionResultFactory.php
+++ b/src/Mutant/MutantExecutionResultFactory.php
@@ -106,7 +106,7 @@ class MutantExecutionResultFactory
 
         $output = $this->retrieveProcessOutput($process);
 
-        if ($this->testFrameworkAdapter->testsPass($output)) {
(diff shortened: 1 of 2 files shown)


sanmai (approved):

maks-rafalko:
thank you, @sanmai

PR #1622, opened 2021-12-22
Return unique tests for the range with more than 1 line

maks-rafalko:
Same test can cover more than 1 line. To avoid many duplicates, we need to return unique tests after accumulating them by each line from the `$lineRange`.

Consider the following example:

```php
public function foo(): string
{
    return $a // line 1, covered by test_case_1, test_case_2
       ? $b   // line 2, covered by test_case_1
       : $c;  // line 3, covered by test_case_2
}
```

on `master`, before https://github.com/infection/infection/pull/1619 (the first fix for non-unique `TestLocation`s) we had:

```bash
3 * 3 * 2 = 18 # (9 duplicates of test_case_1, 9 duplicates of test_case_2)
│   │   └─── number of test cases
│   └─────── number of source code lines
└─────────── number of outer iterations
```
after https://github.com/infection/infection/pull/1619 fix, we had

```bash
1 * 3 * 2 = 6 # (3 duplicates of test_case_1, 3 duplicates of test_case_2)
│   │   └─── number of test cases
│   └─────── number of source code lines
└─────────── number of outer iterations
```

because we removed outer iterations by making range [1-line](https://github.com/infection/infection/blob/30db7f9f3e2b6aa654408dd37587637479ad8b99/src/TestFramework/Coverage/LineRangeCalculator.php#L51-L54) for signature

and now, in this PR, we have

```bash
unique(tests from 3 lines) = 2 # (test_case_1, test_case_2), no duplicates
```

because after all tests for each line in the range are accumulated, we remove duplicates.

No time difference noticed. Memory improvements:

```diff
- Memory: 0.24GB
+ Memory: 0.23GB
```

### DIFF
--- a/src/TestFramework/Coverage/XmlReport/TestLocator.php
+++ b/src/TestFramework/Coverage/XmlReport/TestLocator.php
@@ -36,6 +36,7 @@ declare(strict_types=1);
 namespace Infection\TestFramework\Coverage\XmlReport;
 
 use function array_key_exists;
+use function array_values;
 use Infection\AbstractTestFramework\Coverage\TestLocation;
 use Infection\TestFramework\Coverage\NodeLineRangeData;
 use Infection\TestFramework\Coverage\SourceMethodLineRange;
@@ -96,11 +97,19 @@ class TestLocator
      */
     private function getTestsForLineRange(NodeLineRangeData $lineRange): iterable
     {
+        // same test can cover more than 1 line. To avoid many duplications, we need to return unique tests after
+        // accumulating them by each line from the range
+        $uniqueTestLocations = [];
+
         foreach ($lineRange->range as $line) {
             if (array_key_exists($line, $this->testLocations->getTestsLocationsBySourceLine())) {
-                yield from $this->testLocations->getTestsLocationsBySourceLine()[$line];
+                foreach ($this->testLocations->getTestsLocationsBySourceLine()[$line] as $testLocation) {
+                    $uniqueTestLocations[$testLocation->getMethod()] = $testLocation;
+                }
             }
         }
+
+        yield from array_values($uniqueTestLocations);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tests/phpunit/TestFramework/Coverage/XmlReport/TestLocatorTest.php
```diff
@@ -126,26 +126,6 @@ public function rangeProvider(): iterable
...
-                [
-                    'testMethod' => 'Infection\Acme\FooTest::test_it_can_do_1',
-                    'testFilePath' => '/path/to/acme/FooTest.php',
-                    'testExecutionTime' => 0.456,
-                ],
-                [
-                    'testMethod' => 'Infection\Acme\FooTest::test_it_can_do_0',
-                    'testFilePath' => '/path/to/acme/FooTest.php',
-                    'testExecutionTime' => 0.123,
-                ],
```
maks-rafalko, line 148: better than 1000 words, here is a result of removing duplicates. This test was a proof that we had a bug :)

Now - there are no more duplicates (removed array values are duplicates)

src/TestFramework/Coverage/XmlReport/TestLocator.php
```diff
@@ -96,11 +97,19 @@ private function getTestsForFunctionSignature(NodeLineRangeData $lineRange): ite
...
     private function getTestsForLineRange(NodeLineRangeData $lineRange): iterable
     {
+        // same test can cover more than 1 line. To avoid many duplications, we need to return unique tests after
+        // accumulating them by each line from the range
+        $uniqueTestLocations = [];
+
         foreach ($lineRange->range as $line) {
             if (array_key_exists($line, $this->testLocations->getTestsLocationsBySourceLine())) {
-                yield from $this->testLocations->getTestsLocationsBySourceLine()[$line];
+                foreach ($this->testLocations->getTestsLocationsBySourceLine()[$line] as $testLocation) {
```
sidz, line 106: can't we get rid of `if` here? and do smth like:
```
foreach ($this->testLocations->getTestsLocationsBySourceLine()[$line] ?? [] as $testLocation) {
   //do smth
}
```

sidz (approved):
:+1:

src/TestFramework/Coverage/XmlReport/TestLocator.php
maks-rafalko: fixed here https://github.com/infection/infection/commit/6a6eb17b475786b59adee5572f16d579a5eb4e63

maks-rafalko:
thank you @sidz

PR #1624, opened 2021-12-22
Allow phpstan and infection composer plugins to be executed (Composer 2.2)

maks-rafalko:
Let's see if it works with new Composer 2.2 requirements (https://getcomposer.org/doc/06-config.md#allow-plugins) and if it fixes https://github.com/infection/infection/pull/1623

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -89,7 +89,11 @@
         "platform": {
             "php": "7.4.7"
         },
-        "sort-packages": true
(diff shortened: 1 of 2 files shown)


maks-rafalko:
seems like the failures are from Box project, then merging https://github.com/infection/infection/pull/1623 for now until it's fixed there

PR #1630, opened 2021-12-28
Set `failOnRisky`, `failOnWarning` to `true` if parameters are not already set for mutants

maks-rafalko:
previously, we did the same for **initial** tests run generated `phpunit.xml`.

This update is for `phpunit.xml`, generated for each Mutant.

Related to https://github.com/infection/infection/pull/1280, https://github.com/infection/infection/issues/1620


sanmai (approved):
Makes sense.

maks-rafalko:
thank you @sanmai

Issue #1604, opened 2021-11-18
`ignoreSourceCodeByRegex` with no match should raise an Error


Slamdunk:
Hi, I've used quite few times `ignoreSourceCodeByRegex` (both global and per-mutator) and it may happen that, after a refactor, an ignore regex is not necessary anymore. But Infection doesn't tell me that, so I might forget the regex in place, and that could be very bad if, in a later commit, a new line-of-code matches the forgotten regex by accident.

Just like PHPStan fails when an ignore regex is set but has no match during the check, I'd like Infection too to have an exit code over zero for unmatched `ignoreSourceCodeByRegex`.

sanmai:
Unmatched regexes should not lead to a non-zero exit status by default as there could be version-dependent mutants which do not always appear and trigger a regex, among other things. Showing a notice should be fine, though.

orlandothoeny:
Another possibility would be to make the behaviour configurable, fore example as PHPStan is doing it: https://phpstan.org/user-guide/ignoring-errors#reporting-unused-ignores

theofidry:
Also when applying on a diff, this cannot work.

But yeah maybe a confit entry or a flag could help.

sanmai:
There are also `@infection-ignore-all` annotations. [Read about them here.](https://infection.github.io/guide/usage.html#infection-ignore-all-support)

PR #1631, opened 2021-12-28
Fail Infection execution when at least 1 ignore source code regex wasn't matched

maks-rafalko:
Implements https://github.com/infection/infection/issues/1604

Will be useful to remove stale/non-matching `ignoreSourceCodeByRegex` settings inside `infection.json` and will help debugging them.

### DIFF
--- a/src/Process/Runner/MutationTestingRunner.php
+++ b/src/Process/Runner/MutationTestingRunner.php
@@ -35,6 +35,10 @@ declare(strict_types=1);
 
 namespace Infection\Process\Runner;
 
+use function array_keys;
+use function array_merge;
+use function array_unique;
+use function array_values;
 use Infection\Differ\DiffSourceCodeMatcher;
 use Infection\Event\EventDispatcher\EventDispatcher;
 use Infection\Event\MutantProcessWasFinished;
@@ -47,6 +51,7 @@ use Infection\Mutant\MutantFactory;
 use Infection\Mutation\Mutation;
 use Infection\Process\Factory\MutantProcessFactory;
 use function Pipeline\take;
+use function Safe\array_flip;
 use Symfony\Component\Filesystem\Filesystem;
 
 /**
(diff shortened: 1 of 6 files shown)


src/Process/Runner/MutationTestingRunner.php
```diff
@@ -118,6 +127,13 @@ public function run(iterable $mutations, string $testFrameworkExtraOptions): voi
 
                 return true;
             })
+            ->filter(static function (Mutant $mutant) use ($notMatchedSourceCodeRegexes): bool {
+                if ($notMatchedSourceCodeRegexes !== []) {
+                    throw NotMatchedIgnoreSourceCodeRegexFound::forRegexes(array_keys($notMatchedSourceCodeRegexes));
+                }
+
+                return true;
+            })
```
maks-rafalko, line 136: @sanmai I need your help here.

I feel like `sanmain/pipeline` needs one more method - `tap` (similar to this one https://lodash.com/docs/#tap)

What I want to do is after Mutants are filtered and we completed building `$notMatchedSourceCodeRegexes` - I would like to check it, but there is no possibility to do it since we have only `filter()` method on the `Standard` class, which **iterates over all the items again**.

So, desired logic:

```php
take($mutations)
    ->filter() // filter ignored mutants out and build $notMatchedSourceCodeRegexes
    // (!) new
    ->tap() // check $notMatchedSourceCodeRegexes and throw an exception if needed
    ->filter() // continue existing logic - filter not covered by tests mutants
```

do you have any recommendations here? is it the right way to go?

sanmai: The problem here is that there's no `array $mutants`; there are code paths when we produce them only as required. In other words, worst case there's `interable $mutants` with who knows how many values in it.

If you want to drop arbitrary parts of a sequence (say, first five, or two after ten), there's a handy `slice()` method, but I don't think that's what you're looking for here.

maks-rafalko: I'm looking for a place (method) where I can do some job after the first `filter()` but before the second `filter()`

sanmai: Since you can't know if you're looking at the very last mutant, you can't throw an exception from anywhere here. Neither you should require to iterate over mutants all over again. Therefore I'd like to suggest to track regex usages, and only throw at some later stage. With that said I would much more prefer to see a notice: breaking non-PR builds can be a real pain.

There's another option: you can inject a fake mutant at the end of the stream, and then catch it somewhere else. I can elaborate further, but this will be incredibly scuffed and very opaque. Simple counting should work better.

sanmai (changes requested):
Unmatched regexes should not lead to a non-zero exit status by default as there could be version-dependent mutants which do not always appear and trigger a regex. Showing a notice should be fine, though.

src/Process/Runner/MutationTestingRunner.php
```diff
@@ -99,15 +104,19 @@ public function run(iterable $mutations, string $testFrameworkExtraOptions): voi
...
             ->cast(function (Mutation $mutation): Mutant {
                 return $this->mutantFactory->create($mutation);
             })
-            ->filter(function (Mutant $mutant): bool {
+            ->filter(function (Mutant $mutant) use (&$notMatchedSourceCodeRegexes): bool {
                 $mutatorName = $mutant->getMutation()->getMutatorName();
 
                 foreach ($this->ignoreSourceCodeMutatorsMap[$mutatorName] ?? [] as $sourceCodeRegex) {
                     if ($this->diffSourceCodeMatcher->matches($mutant->getDiff()->get(), $sourceCodeRegex)) {
+                        unset($notMatchedSourceCodeRegexes[$sourceCodeRegex]);
```
sanmai, line 118: Could be:
```php
$matchedSourceCodeRegexes[$sourceCodeRegex] += 1;
```

src/Process/Runner/MutationTestingRunner.php
```diff
@@ -118,6 +127,13 @@ public function run(iterable $mutations, string $testFrameworkExtraOptions): voi
 
                 return true;
             })
+            ->filter(static function (Mutant $mutant) use ($notMatchedSourceCodeRegexes): bool {
+                if ($notMatchedSourceCodeRegexes !== []) {
+                    throw NotMatchedIgnoreSourceCodeRegexFound::forRegexes(array_keys($notMatchedSourceCodeRegexes));
```
sanmai, line 132: This is not going to work because this gets called for every mutant, even for the very first one.

PR #1633, opened 2021-12-30
Do not add methods with 0% coverage to `TestsLocations` to prevent wasting of resources

maks-rafalko:
Consider the following XML coverage file https://github.com/infection/infection/blob/de7d826e770fbb0400c91272e5460e1ff880e65f/tests/phpunit/Fixtures/Files/phpunit/coverage/coverage-xml/FirstLevel/firstLevel.php.xml#L20 where `mutate()` method has 0% coverage.

There is no need to add this method to `TestsLocations` as later no tests will be found for its lines, so that the following code will just waste resources:

https://github.com/infection/infection/blob/de7d826e770fbb0400c91272e5460e1ff880e65f/src/TestFramework/Coverage/XmlReport/TestLocator.php#L117-L130

because since method is not covered - lines inside it also are not covered (check the XML coverage above)

For Infection itself, I see no benefits regarding time and memory used though, however it's clear that less operations are executed during Infection run.

### DIFF
--- a/src/TestFramework/Coverage/XmlReport/XmlCoverageParser.php
+++ b/src/TestFramework/Coverage/XmlReport/XmlCoverageParser.php
@@ -156,6 +156,10 @@ class XmlCoverageParser
         $methodsCoverage = [];
 
         foreach ($methodsCoverageNodes as $methodsCoverageNode) {
+            if ((int) $methodsCoverageNode->getAttribute('coverage') === 0) {
+                continue;
+            }
+
             $methodName = $methodsCoverageNode->getAttribute('name');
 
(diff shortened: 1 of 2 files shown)


sanmai:
I strangely remember there was PHPUnit bug which made it report zero lines covered for a method, but non-zero for the lines inside the method.

maks-rafalko:
It would be nice to have an example, because before opening this PR, I generated XML reports and didn't notice anything like that.

Probably it was for the old versions? Then (probably) this change will cause escaped mutants for someone and we will find it out, and will cover by test case

sanmai:
Yeah, that's a fair bet.

PR #1642, opened 2022-01-11
Add support for latest ondram/ci-detector version

rdotter:
ondram/ci-detector now only supports ^3.3, when upgrading to the latest GrumPHP (https://packagist.org/packages/phpro/grumphp), ondram/ci-detector ^4.0 is required.

This PR:

- [x] Covered by tests (should be because of usage of supporting both lowest and highest dependencies)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,7 +53,7 @@
         "infection/include-interceptor": "^0.2.5",
         "justinrainbow/json-schema": "^5.2.10",
         "nikic/php-parser": "^4.13",
-        "ondram/ci-detector": "^3.3.0",
+        "ondram/ci-detector": "^3.3.0 || ^4.0",
         "sanmai/later": "^0.1.1",
         "sanmai/pipeline": "^5.1 || ^6",
         "sebastian/diff": "^3.0.2 || ^4.0",


rdotter:
@maks-rafalko could you have a look? We want to upgrade to the latest GrumPHP version (for PHP8.1 support) and this is blocking us on getting forward.

Thanks :pray:

maks-rafalko:
this won't work, unfortunately. 

3.x: https://github.com/OndraM/ci-detector/blob/4.0.0/src/Ci/CiInterface.php#L51

4.x: https://github.com/OndraM/ci-detector/blob/3.5.1/src/Ci/CiInterface.php#L47

interfaces are not compatible, and we are currently using `getGitBranch()` method.

What I can suggest is to upgrade `ondram/ci-detector` so that infection uses 4.x version inside. Previous Infection versions will use 3.x then.

Could you please do it? I would be happy to merge and release.

maks-rafalko:
Closing in favor of https://github.com/infection/infection/pull/1669

maks-rafalko:
Released: https://github.com/infection/infection/releases/tag/0.26.5

Issue #1643, opened 2022-01-11
Error `Malformed UTF-8 characters, possibly incorrectly encoded ` on send Stryke dashboard report with tests that contain invalid UTF-8 characters


vjik:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.26.0
| Test Framework version | PHPUnit 9.5.11
| PHP version | 7.4.26
| Platform    | Ubuntu
| Github Repo | -

See here: https://github.com/yiisoft/strings/runs/4777997532?check_suite_focus=true#step:8:198

<details>
<summary>infection.json</summary>

```
{
    "source": {
        "directories": [
            "src"
        ]
    },
    "logs": {
        "text": "php:\/\/stderr",
        "stryker": {
            "report": "master"
            "badge": "master"
        }
    },
    "mutators": {
        "@default": true
    }
}
```

</details>

<details>
 <summary>phpunit.xml</summary>

 ```xml
<?xml version="1.0" encoding="UTF-8"?>

<phpunit bootstrap="vendor/autoload.php"
         colors="true"
         verbose="true"
         failOnRisky="true"
         failOnWarning="true"
         convertErrorsToExceptions="true"
         convertNoticesToExceptions="true"
         convertWarningsToExceptions="true"
         stopOnFailure="false"
         executionOrder="random"
         resolveDependencies="true">
    <php>
        <ini name="error_reporting" value="-1"/>
    </php>

    <testsuites>
        <testsuite name="Yii Strings tests">
            <directory>./tests</directory>
        </testsuite>
    </testsuites>

    <coverage>
        <include>
            <directory>./src</directory>
        </include>
    </coverage>
</phpunit>
 ```
</details>

<details>
 <summary>Output with issue</summary>

 ```
vendor/bin/roave-infection-static-analysis-plugin -j2 --ignore-msi-with-no-mutations --only-covered

    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/

Infection - PHP Mutation Testing Framework version 0.26.0

Notice:  You are running Infection with PCOV enabled.

Running initial test suite...

PHPUnit version: 9.5.11

[...]

PR #1644, opened 2022-01-11
Correctly handle non UTF-8 characters in diff and process output for HTML logger

maks-rafalko:
Fixes https://github.com/infection/infection/issues/1643


devanych:
Thanks. This solved the #1643 and https://github.com/yiisoft/strings/pull/77 problems.

PR #1651, opened 2022-01-18
`thecodingmachine/safe` update

ricardoboss:
In this PR, I updated `thecodingmachine/safe` to version `^2.0` to support the new signatures from PHP 8.0.

I noticed there already is a branch for PHP 8 in infection, so I based this PR upon that one.


maks-rafalko:
I'm ok with the change itself, but the build is failing

maks-rafalko:
Hm, seems like was closed automatically for some reason. Feel free to open a new PR and fix the build

ricardoboss:
Will do!

Issue #1650, opened 2022-01-18
`$basePath` must be a string


ricardoboss:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.26.1
| Test Framework version | PHPUnit 9.5.11
| PHP version | 8.1.1
| Platform    | Ubuntu (GitHub actions)
| Github Repo | https://github.com/elephox-dev/framework

When running a CI pipeline, I want to generate the new HTML report and publish it via Stryker. I set up my account on Stryker and enabled it for the repository linked above. Then, I added the API secret as a secret variable on GitHub and added it to the step running infection.

In case the bug is not obvious: I expected this to work. Can someone explain why it failed (have I configured it wrong?) or if this is an actual bug?

The error says: "In Assert.php line 2060: $basePath must be a string".

<details>
 <summary>GitHub Actions config</summary>

 ```yml
- name: 🧟 Run Infection (auto scale)
  if: ${{ matrix.os == 'ubuntu-latest' }}
  run: composer run ci:mutation-test:autoscale
  env:
    STRYKER_DASHBOARD_API_KEY: ${{ secrets.STRYKER_DASHBOARD_API_KEY }}
 ```
</details>

<details>
<summary>composer script</summary>

```bash
infection --coverage=build/coverage --skip-initial-tests --logger-github --no-interaction -j$(nproc)
```
</details>

<details>
 <summary>Output with issue</summary>

 ```
Warning:  Sending dashboard report...

In Assert.php line 2060:
                              
  $basePath must be a string  
                              

run [--test-framework TEST-FRAMEWORK] [--test-framework-options TEST-FRAMEWORK-OPTIONS] [-j|--threads THREADS] [--only-covered] [-s|--show-mutations] [--no-progress] [--force-progress] [-c|--configuration CONFIGURATION] [--coverage COVERAGE] [--mutators MUTATORS] [--filter FILTER] [--formatter FORMATTER] [--git-diff-filter GIT-DIFF-FILTER] [--git-diff-lines] [--git-diff-base GIT-DIFF-BASE] [--logger-github] [--logger-html LOGGER-HTML] [--noop] [--only-covering-test-cases] [--min-msi MIN-MSI] [--min-covered-msi MIN-COVERED-MSI] [--log-verbosity LOG-VERBOSITY] [--initial-tests-php-options INITIAL-TESTS-PHP-OPTIONS] [--skip-initial-tests] [--ignore-msi-with-no-mutations] [--debug] [--dry-run]

[...]

PR #1652, opened 2022-01-19
Create `MutationExecutionResult` instance with real path always, not mixed

maks-rafalko:
Fixes https://github.com/infection/infection/issues/1650

Previously, we could have a mix of real paths and relative paths in `MutationExecutionResult` instances. 

This broke HTML logger, which assumes that all paths are real (absolute), to find the base common path (`Path::getLongestCommonBasePath`).


sanmai (approved):

maks-rafalko:
Thank you @sanmai

PR #1674, opened 2022-03-07
Stand with Ukraine

maks-rafalko:
image

PHPUnit did the same: https://github.com/sebastianbergmann/phpunit/commit/4634e702b5f05f5e948e531eb8b4fc19be40610c

and many other projects adds information about the current real situation, see https://twitter.com/Ocramius/status/1500054275325579265

see https://github.com/vshymanskyy/StandWithUkraine/issues/4 and for example this one https://github.com/Roave/SecurityAdvisories/pull/91/files

PR #1676, opened 2022-03-19
Update README.md

sanmai:
Follow up to #1674


maks-rafalko (approved):
there are some issues with the build, but they aren't related to this change obviously. We will fix them later when we have more time and energy...

Thank you @sanmai

Issue #669, opened 2019-03-22
[RFC] Split the entire set of mutators in a sub-project


sanmai:
My proposal is to split the entire set of mutators in a sub-project. Not only `Infection\Mutator\` namespace, but the bulk of related machinery too, while also opening up the API for use in other projects (by removing `@private` annotations first and foremost).

I had this idea for some time. For one because I wanted a tool which I could use in place of `PHPUnit` executable to do mutation testing against a single test suite (e.g. `infection-tool tests/Example/SomethingTest.php`) regardless of an effective coverage (just assuming that if a test says it `@covers` something, it does it for 100%). I don't think this feature is impossible, but this appears to take much more effort I can get by with they way I want it to be with the current state of affairs (Infection would need to have quite drastic changes).

For another, because Infection is very tightly integrated with PHPUnit's outputs, and not like this is bad, but this makes it very hard to integrate other testing frameworks which use different means for estimating coverage or even don't have the standard coverage logs at all. If mutators were a sub-project, then there could be a more tightly-integrated framework-specific tool for, say, Behat, that would function very different from Infection, but still use the same time-proven and thoroughly tested mutators.

Last straw for me was [this video of a talk on mutation testing](https://youtu.be/EPU47QqtMiU?t=758) given at Badoo (is it like Tinder on steroids? I'm out of a dating scene for a long time, but it seems like they're kind of big) posted like yesterday, which mentions problems Badoo had with their codebase. This video is in Russian and quite long, so, saving you the trouble watching it, in the nutshell, they couldn't use Infection because collecting PHPUnit's standard coverage takes several hours for their codebase, and it takes around [30 Gb in disk space alone](https://speakerdeck.com/badoo_tech/mutatsionnoie-tiestirovaniie-v-php-vladimir-iants-badoo?slide=43). They also had some other issues with the tooling they use. In the end they took mutators from Infection (because "they're very solid and concisely written", kudos to all of you!), making their own simplified tool ([an obligatory screenshot](https://speakerdeck.com/badoo_tech/mutatsionnoie-tiestirovaniie-v-php-vladimir-iants-badoo?slide=49)). They could also had used mutators from Infection more easily if they were a sub-project, and even contribute to it in code or else.

One of more or less annoying issues when developing Infection is very long test run times. If mutators were a sub-project, then adding a mutator would not require running the whole Infection's test suite with endless E2E tests in multitude of PHP versions, and such (they're for purpose, so no offence!). Only standard unit tests would have to be run.

[...]

PR #1686, opened 2022-04-29
Support custom mutators

vss414:
Infection supports only a set of Mutators which are based on AST and [PHP-Parser](https://github.com/nikic/PHP-Parser) project. In such situation library users can't use their own mutators (which may be useful in their projects but doesn't fit to be a part of the library).
This PR adds the ability to use custom mutators in projects that use the infection library.
Custom mutators should implement `Infection\Mutator\Mutator` interface.

To use custom mutators, user must add them to the mutator list in the infection config. User has to use the class name with the namespace of the custom mutator as a name of mutator in the config. An example of using a custom mutator in the infection config:
```
{
    "mutators": {
        "App\\Infection\\MyFirstCustomMutator": true,
        "App\\Infection\\MySecondCustomMutator": {
            "ignore": [
                "*Controller::*",
            ]
        }
    }
}
```


maks-rafalko:
Overall, I'm ok with a feature itself, however it is much more complex.

Related to issues/comments discussed before

* https://github.com/infection/infection/issues/669
* https://github.com/infection/infection/pull/947#discussion_r363245788
* https://github.com/infection/infection/pull/278#issuecomment-377486976

As it was already pointed out in those comments, we should consider the following cases (this list can be extended and should taken into account/discussed during development of this feature here or in other PRs):

* exposing `Mutator` interface. Now it's `@internal`, but should be "public". I was thinking about moving this interface to a separate package so that developers can depend on highly abstract package to develop their own Mutators (as we did [here](https://github.com/infection/infection/issues/922), but this can be done later). And unfortunately, `Node` interface from PHP-Parses is also sitting in the main package, which is bad from package design
* PHAR distribution. Now, all the classes inside Infection are prefixed during that [PHAR compilation](https://github.com/infection/infection/blob/714a8d299307bc65e21307c96dca71fdc10ff8fd/scoper.inc.php#L44). It means, `Mutator` interface will be prefixed (read as "namespace will be changed") as well, and this will break your custom mutator. We must to stop prefixing `Mutator` interface. For example, PHPStan does the same - see https://github.com/phpstan/phpstan-src/blob/1.7.x/compiler/build/scoper.inc.php#L268-L269. Note that `Node` interface shouldn't also be prefixed, unfortunately.
* e2e test must be added (see `tests/e2e` and `tests/add_new_e2e`) to check custom Mutator works with `infection/infection` as well as with PHAR

---

By the way, what is your custom Mutator? Can you please provide an example? I'm thinking about what are your case, probably we will need to create another packages like `infection-dotrine` similar to what PHPStan does with its plugins.

resources/schema.json
```diff
@@ -135,7 +135,10 @@
         },
         "mutators": {
             "type": "object",
-            "additionalProperties": false,
+            "additionalProperties": {
```
maks-rafalko, line 138: as Theo already mentioned [here](https://github.com/infection/infection/pull/947#discussion_r363245788), this change can lead to incorrect (not so strict) usage of configuration files. Probably we need to move it inside some key, like `mutators.custom`? I don't know

michalbundyra:
@vss414 This is great! Exactly what I was looking for. Can't wait to get my hands on it :) 

@maks-rafalko My custom mutator will be to remove some calls while using fluent interfaces - like:

```php
$query->select('*')->withTrashed();
```

so I'd add custom mutator to remove part of that statement:

Mutations:
```
$query->select('*');
```

and:
```
$query->withTrashed();
```


Ofc if we force our developers to [not use fluent interfaces](https://ocramius.github.io/blog/fluent-interfaces-are-evil/) then it will be not needed.
But the thing right now is that we are using them and mutation like that would be very nice.
Not sure if it can be global mutation, that's why I would be ok to have it in my project.

I had another examples I wanted to use custom mutators, but I forgot now.... 

Thanks!

blackwolf12333:
I can maybe add some examples from our codebase that would require support for custom mutators to be more effective.

Our codebase is using a custom framework in which we have a custom query language that I would like to be able to mutate. For example querying some database table on a field to be greater than 0 would look like: 

`new ApiConditionSimple('field', 0, '>')`

It would be nice to be able to mutate the `'>'` string which infection currently can't do because it's just a string. But it would be nice to be able to apply the `@conditional_boundary` mutations to that for example.

maks-rafalko:
I decided to move forward with this one, as it turned out we are very close to make it possible, and there is a need for this from Ondrej (maintainer of PHPStan), so we will have a good first consumer of our API.

This will be the first PR, in the following PRs I will extract our extension points to a separate package so that anyone who needs to implement custom mutator will require only small package `infection/mutator` (for example) with an interface instead of require the whole Infection.

But I have a problem here, seems like Box/Scoper have a bug or something like that.

@theofidry when I add `Infection\Mutator\Mutator` in the excluded files in `scoper.inc.php`, and run

```bash
cd tests/e2e/Custom_Mutator

../../../build/infection.phar
```

then I have an error:

```
PHP Fatal error:  Uncaught Error: Interface "Infection\Mutator\Mutator" not found in phar:///home/maksrafalko/apps/infection/build/infection.phar/src/Mutator/Arithmetic/Assignment.php:15
Stack trace:
#0 phar:///home/maksrafalko/apps/infection/build/infection.phar/.box/vendor/composer/ClassLoader.php(576): include()
#1 phar:///home/maksrafalko/apps/infection/build/infection.phar/.box/vendor/composer/ClassLoader.php(427): Composer\Autoload\{closure}()
```

which looks weird, as there is a non-scoped interface in the PHAR

theofidry:
>UPD: probably, I misunderstood exclude and expose concepts inside Scoper, so I re-read and updated scoper.inc.php, let's see if it works.

What you likely want there is expose indeed, exclude is to mark a symbol as "native to PHP" (how it is handled is... depends, e.g. you can declare native PHP symbols, for example that is what polyfills are for, or for handling symbols coming from third-party that is not in your vendor for WP plugins).

>The bad thing that PHP-Parser's PhpParser\Node interface is not in the separate package, so we have to depend on the whole nikic/php-parser just to use 1 interface :| and because of that - do not prefix the whole parser package. I tried to not prefix only this interface, but then it doesn't work, probably I'm doing it wrong so @theofidry your review help would be appreciated

I don't think that would be possible, as a consumer you want to be able to consume any node no, so it's more than just the node interface?

I think for now it's best if you expose the whole `PhpParser` namespace. I can't review the code right now will try to do so ASAP

scoper.inc.php
```diff
@@ -35,6 +35,15 @@
...
 return [
     'prefix' => 'Infected',
+    'expose-classes' => [
+        'Infection\Mutator\Definition',
+        'Infection\Mutator\Mutator',
+        'Infection\Mutator\MutatorCategory',
+    ],
+    'expose-namespaces' => [
+        // we have to expose it cause Mutator depends on PhpParser/Node interface, and it's not in a separate package
+        '/^PhpParser/',
```
theofidry, line 45: can be just the name:

```suggestion
        'PhpParser',
```

maks-rafalko:
Replacing `exclude-namespaces` with `expose-namespaces` for PHP-Parser namespaces leads to error during `make compile` (https://github.com/infection/infection/pull/1686/commits/809d8f243e6baf72321e4e6f055be65a76c9b764):

```
PHP Fatal error:  Cannot declare class Infected\PhpParser\Internal\TokenPolyfill, because the name is already in use in phar:///home/runner/work/infection/infection/build/infection.phar/vendor/nikic/php-parser/lib/PhpParser/Internal/TokenPolyfill.php on line 8
```

https://github.com/infection/infection/actions/runs/8953112456/job/24591330644?pr=1686#step:15:339

So for now I will revert it.

src/Mutant/MutantExecutionResult.php
```diff
@@ -121,6 +123,11 @@ public function getMutantHash(): string
         return $this->mutantHash;
     }
 
+    public function getMutatorClass(): string
```
theofidry, line 126: is there a reason to use a public getter instead of making it a public property since it's readonly?

theofidry (commented):
only a glance review; need to do a proper one still

resources/schema.json
maks-rafalko: resolved here: https://github.com/infection/infection/pull/1686/commits/2f6ec735b7633ed40472ecfa03ccc709618a54ac

scoper.inc.php
maks-rafalko: This is what I started with, but running `infection.phar` then fails with:

```
PHP Fatal error:  Uncaught Error: Class "Infection\PhpParser\FileParser" not found in phar:///home/maksrafalko/apps/infection/build/infection.phar/src/Container.php:185
```

It looks like it matches `Infection\PhpParser\FileParser` in the middle

src/Mutant/MutantExecutionResult.php
maks-rafalko: it's for consistency - all other private readonly variables are exposed with public getters. If we need to use public readonly properties, this should be done everywhere in a separate PR

maks-rafalko (approved):
Ok, I need to move forward. If there will be any additional feedback, I will sort it out in separate PRs

maks-rafalko:
thank you @vss414 for pushing it forward and initial work, stay tuned.

maks-rafalko:
@vss414 it's released https://github.com/infection/infection/releases/tag/0.29.0

PR #1688, opened 2022-05-05
Skip analysis when no executed lines with `--git-diff-lines` option

romm:
This PR:
 
* [x]  Adds new feature 
* [x]  Covered by tests

---

When using the `--git-diff-lines` option, there are situations where a
file can be modified but will not yield any test change, for instance
when only a comment is modified in a PHP source file.

This commit prevents an error being thrown when this very situation
occurs; instead, the analysis is skipped.

### DIFF
--- a/src/TestFramework/Coverage/XmlReport/NoLineExecutedInDiffLinesMode.php
+++ b/src/TestFramework/Coverage/XmlReport/NoLineExecutedInDiffLinesMode.php
@@ -1,0 +1,49 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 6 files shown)


romm:
Reference: https://github.com/CuyZ/Valinor/runs/6311117414

src/TestFramework/Coverage/XmlReport/NoLineExecutedInDiffLinesMode.php
```diff
@@ -0,0 +1,49 @@
...
+use UnexpectedValueException;
+
+/**
+ * @internal
+ */
+final class NoLineExecutedInDiffLinesMode extends UnexpectedValueException
+{
+    public static function create(): self
+    {
+        return new self('No files in diff found, skipping mutation analysis.');
```
maks-rafalko, line 47:
```suggestion
        return new self('No covered lines in in diff found, skipping mutation analysis.');
```

or something similar, because the current error message is from another exception :)

maks-rafalko (approved):
Just a minor suggestion, otherwise looks good :+1:

src/TestFramework/Coverage/XmlReport/NoLineExecutedInDiffLinesMode.php
romm: Yeah I kept the message on purpose because I thought it made sense in this situation as well, but there is a slight difference indeed. Thanks for the review!

maks-rafalko:
Thank you @romm

maks-rafalko:
Release as https://github.com/infection/infection/releases/tag/0.26.9

Enjoy!

romm:
That was fast! thank you!!

Issue #1662, opened 2022-02-03
Infection do not working on PHP 8.1.1. 


lvsantos1:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.26.4
| Test Framework version | PHPUnit
| PHP version | 8.1.1
| Platform    | Windows

<details>
 <summary>phpunit.xml</summary>

 ```xml
<?xml version="1.0"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd" backupGlobals="true" backupStaticAttributes="false" bootstrap="phpunit_bootstrap.php" colors="true">
  <coverage/>
  <testsuites>
    <testsuite name="Domain">
      <directory>Domain</directory>
    </testsuite>
  </testsuites>
</phpunit>

 ```
</details>

Infection do not working on PHP 8.1.1. 

<details>
 <summary>Output with issue</summary>

 ```
PHP Fatal error:  Uncaught ValueError: Unknown format specifier "~" in C:\Users\vinic\AppData\Roaming\Composer\vendor\webmozart\assert\src\Assert.php:836
Stack trace:
#0 C:\Users\vinic\AppData\Roaming\Composer\vendor\webmozart\assert\src\Assert.php(836): sprintf()
#1 C:\Users\vinic\AppData\Roaming\Composer\vendor\infection\infection\src\TestFramework\VersionParser.php(57): Webmozart\Assert\Assert::notSame()
#2 C:\Users\vinic\AppData\Roaming\Composer\vendor\infection\infection\src\TestFramework\AbstractTestFrameworkAdapter.php(190): Infection\TestFramework\VersionParser->parse()
#3 C:\Users\vinic\AppData\Roaming\Composer\vendor\infection\infection\src\TestFramework\AbstractTestFrameworkAdapter.php(131): Infection\TestFramework\AbstractTestFrameworkAdapter->retrieveVersion()
#4 C:\Users\vinic\AppData\Roaming\Composer\vendor\infection\infection\src\TestFramework\AbstractTestFrameworkAdapter.php(141): Infection\TestFramework\AbstractTestFrameworkAdapter->getVersion()
[...]

PR #1690, opened 2022-05-11
Escape percent sign in `Assert::notSame` message

villfa:
This PR fixes https://github.com/infection/infection/issues/1662

The 3rd argument of `Webmozart\Assert\Assert::notSame` is passed to `sprintf`:
```php
    public static function notSame($value, $expect, $message = '')
    {
        if ($expect === $value) {
            static::reportInvalidArgument(\sprintf(
                $message ?: 'Expected a value not identical to %s.',
                static::valueToString($expect)
            ));
        }
    }
```

See https://github.com/webmozarts/assert/blob/dc96b67/src/Assert.php#L831-L839

If we call it with a message containing `%~` then PHP thows `ValueError: Unknown format specifier "~"`.

This PR escapes the `%` sign to avoid this error.

### DIFF
--- a/src/TestFramework/VersionParser.php
+++ b/src/TestFramework/VersionParser.php
@@ -37,6 +37,7 @@ namespace Infection\TestFramework;
 
 use function Safe\preg_match;
 use function Safe\sprintf;
+use function str_replace;
 use Webmozart\Assert\Assert;
 
 /**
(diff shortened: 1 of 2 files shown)


maks-rafalko (approved):
Thank you @villfa

maks-rafalko:
Released https://github.com/infection/infection/releases/tag/0.26.10

Issue #1634, opened 2022-01-07
Avoid using `webmozart/path-util`


ricardoboss:
Apparently, [webmozart/path-util](https://packagist.org/packages/webmozart/path-util) is no longer maintained and infection has it as a dependency.

You should consider changing it to a maintained package or the proposed one: [symfony/filesystem](https://packagist.org/packages/symfony/filesystem).

maks-rafalko:
Thanks. `symfony/filesystem` is already [in our composer.json](https://github.com/infection/infection/blob/e55f980bcd73d365d2c15be786cf026cf063e72f/composer.json#L62) / lock file, so feel free to help with getting rid of `webmozart/path-util`

ricardoboss:
Will look into it!

ricardoboss:
The question is, are you willing to drop support for everything below Symfony 5.4? An alternative could be to just "borrow" the small set of functions used (like `isAbsolute` or `canonicalize`) and put them as static functions somewhere in the infection codebase.

What do you think?

maks-rafalko:
> The question is, are you willing to drop support for everything below Symfony 5.4?

hm.. didn't think about it, I guess now, because now we support `"symfony/filesystem": "^3.4.29 || ^4.1.19 || ^5.0 || ^6.0",`

> An alternative could be to just "borrow" the small set of functions used 

then the question is what is better - just continue depending on the abandoned package or copy the code to our codebase? What is the drawbacks of using `webmozart/path-util` except warning message?

ricardoboss:
IMHO it's more of an emotional drawback. It's not really assuring when installing a package and the first thing you see is: "This package has an abadoned dependency!"

Technically it should be fine, I think. (at last as long as it's available on packagist...)

theofidry:
Pretty mch as @ricardoboss pointed out. AFAIK path-util is very stable and didn't require a single patch for several years so code-wise it is unlikely to get out of sync any time soon. So it's mostly about whether the message is prohibitive or not

maks-rafalko:
> IMHO it's more of an emotional drawback. It's not really assuring when installing a package and the first thing you see is: "This package has an abadoned dependency!"

Well, keeping in mind that we can't upgrade to `symfony/filesystem:^5.4` and keeping in mind that _any_ project will see this orange warning message - personally I'm ok to get rid of  `webmozart/path-util` and copy 4 used functions to Infection codebase.

RodideBoer:
[...]

PR #1693, opened 2022-05-30
Replace abandoned `webmozart/path-util` with `symfony/filesystem`

martinssipenko:
This PR:

Replaces abandoned `webmozart/path-util` with `symfony/filesystem`.

Fixes https://github.com/infection/infection/issues/1634

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -59,7 +59,7 @@
         "sebastian/diff": "^3.0.2 || ^4.0",
         "seld/jsonlint": "^1.7",
         "symfony/console": "^3.4.29 || ^4.1.19 || ^5.0 || ^6.0",
-        "symfony/filesystem": "^3.4.29 || ^4.1.19 || ^5.0 || ^6.0",
+        "symfony/filesystem": "^5.4 || ^6.0",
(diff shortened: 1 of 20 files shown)


martinssipenko:
I just noticed the conversation in #1634 about dropping Symfony below `5.4`. Question I have is why do you need to keep those old Symfony versions?

maks-rafalko (approved):
I think we don't need it.

image

All the versions below 5.4 EOLd, so personally I'm ok with the current state of this PR.

Users with older SF versions will use Infection <= 0.26.10

martinssipenko:
@maks-rafalko dunno whether 4.4 works on PHP 8.0, which is required for Infection. Will you merge the PR then? I later can follow up with another one to remove < 5.4 altogether.

maks-rafalko:
> dunno whether 4.4 works on PHP 8.0

according to symfony repostiories and GH Actions, it works even on 8.1

> Will you merge the PR then?

yes, right now

> I later can follow up with another one to remove < 5.4 altogether.

do you think it's needed? for what reason?

maks-rafalko:
Released: https://github.com/infection/infection/releases/tag/0.26.11

PR #1695, opened 2022-05-31
Drop `symfony/*` below `5.4`

martinssipenko:
This PR:

Drops `symfony/*` below `5.4` and allows to install `6.x` versions of dev deps.

I created this as a follow-up of [#1693](https://github.com/infection/infection/pull/1693) where `symfony/filesystem` below `5.4` was dropped, if you think this is not necessary just close the PR.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,10 +58,10 @@
         "sanmai/pipeline": "^5.1 || ^6",
         "sebastian/diff": "^3.0.2 || ^4.0",
         "seld/jsonlint": "^1.7",
-        "symfony/console": "^3.4.29 || ^4.1.19 || ^5.0 || ^6.0",
+        "symfony/console": "^5.4 || ^6.0",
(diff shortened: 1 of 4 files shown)


maks-rafalko:
> if you think this is not necessary just close the PR. 

The problem is that currently I don't know if this is necessary or not :) What is the main benefit of this update? Does it solve any issue?

martinssipenko:
> What is the main benefit of this update?

The only benefit would be keeping things clean and consistent.

> Does it solve any issue?

Not that I know of.

Slamdunk (approved):

maks-rafalko (approved):
Ok, since Symfony <5.4 versions are not support anymore (except 4.4 LTS) and since users who still use outdated versions can use older Infection versions - I think it's good to go from my side.

@martinssipenko may I ask you to rebase and push? Thank you.

maks-rafalko:
Thank you @martinssipenko

Issue #1700, opened 2022-06-21
Running the Infection cannot detect code coverage driver correctly


peter279k:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.26.11
| Test Framework version | PHPUnit 9.5.21
| PHP version | 8.0.20 (`php -v`)
| Platform    | Ubuntu 18.04
| Github Repo | https://github.com/peter279k/php-gender

When enabling the `XDebug` extension to run the `./vendor/bin/infection`, it will get the following error message:

```
The code coverage generated by the initial test run is invalid. Please report the issue on the                                                          
  infection repository "https://github.com/infection/infection".               
                                                                               
                                                                 
  $ '/data/php-gender/vendor/bin/phpunit' '--configuration' '/tmp/infection/p  
  hpunitConfiguration.initial.infection.xml' '--coverage-xml=/tmp/infection/c  
  overage-xml' '--log-junit=/tmp/infection/junit.xml'                          
  PHPUnit 9.5.21 #StandWithUkraine                                             
                                                                               
  Random Seed:   1655794267                                                    
  Warning:       XDEBUG_MODE=coverage or xdebug.mode=coverage has to be set    
                                                                               
  .                                                                   1 / 1 (  
  100%)                                                                        
                                                                               
  Time: 00:00.006, Memory: 8.00 MB                                             
                                                                               
  OK (1 test, 3 assertions)                                                    
                                                                                                                                             
                                                                               
  Issue(s):                                                                    
  - The file "index.xml" could not be found: Could not find any "index.xml" file in "/tmp/infection"                                                      

```

- If adding the `XDEBUG_MODE=coverage` to declare the environment variable, it will not get above error message.
   -  As this [comment](https://github.com/infection/infection/issues/1691#issuecomment-1160471029) says, it's not necessary to do that. It seems to be a bug.

PR #1701, opened 2022-06-21
Try to resolve issue #1700

peter279k:
This PR:

- [X] Closes issue #1700.

Fixes https://github.com/infection/infection/issues/1700

- Finally, I guess I found the problematic codes about above issue.
- By default, the `xdebug.mode` setting will be `develop` in the Ubuntu distribution and I think it should change the condition to evaluate the setting is not `coverage`. And it can cover all of situations about setting the `XDEBUG_MODE` environment variable.


src/Process/OriginalPhpProcess.php
```diff
@@ -89,7 +89,7 @@ private static function shallExtendEnvironmentWithXdebugMode(): bool
         }
 
         // The last case: Xdebug 3+ running inactive.
-        return ini_get_unsafe('xdebug.mode') === '';
+        return ini_get_unsafe('xdebug.mode') !== 'coverage';
```
sanmai, line 92: This should be fine since [Xdebug 2 isn't supported with PHP 7.4](https://xdebug.org/docs/compat)

tests/phpunit/Process/OriginalPhpProcessTest.php
```diff
@@ -64,13 +64,12 @@ public function test_it_injects_xdebug_env_vars(): void
     {
         $process = new OriginalPhpProcess(['env']);
         $process->run(null, ['TESTING' => 'test']);
-
```
sanmai, line 67: You'll need to re-add this line.

sanmai (approved):
I guess it makes sense to change it to compare against any other value since it isn't always defaults to an empty string. Thanks for going forward with this fix!

sanmai (approved):
Still looking good

maks-rafalko (approved):
Makes sense, thanks for your time @peter279k & @sanmai

maks-rafalko:
Released https://github.com/infection/infection/releases/tag/0.26.12

PR #1703, opened 2022-06-22
Fix issue_#1700 again

peter279k:
This PR:

Fixes https://github.com/infection/infection/issues/1700 again.

Since I use the latest version (0.26.12) and setup the same development environment, it still gets the same error that mention in above issue.

I try to collect the `ini_get_unsafe('xdebug.mode')` and `extension_loaded('xdebug')` values, and I found these values are `false`.

I also ensure that my `XDebug` extension is available and I try to `return true` directly. It's worked perfectly without presenting the above issue.

I suspect the internal `XDebug 3.x` enabling mechanism behavior is different from `XDebug 2.x` and that's the reason why these conditions are not worked well.

I also notice that the future Infection will not support old PHP 7.4 version. That is, it will not support `XDeBug 2.x` version in upcoming version.

I think it's time to remove these useless conditions and return `true` directly if the `pcov` and `phpdbg` are not loaded.

### DIFF
--- a/devTools/phpstan-src.neon
+++ b/devTools/phpstan-src.neon
@@ -32,9 +32,6 @@ parameters:
             message: '#^Method Infection\\Container::get.*\(\) should return .* but returns object\.$#'
             count: 1
         -
-            path: '../src/Process/OriginalPhpProcess.php'
-            message: '#Function ini_get is unsafe to use#'
-        -
             path: '../src/TestFramework/Coverage/CoverageChecker.php'
             message: '#Function ini_get is unsafe to use#'
         -
--- a/src/Process/OriginalPhpProcess.php
+++ b/src/Process/OriginalPhpProcess.php
@@ -39,7 +39,6 @@ use function array_merge;
 use Composer\XdebugHandler\PhpConfig;
 use Composer\XdebugHandler\XdebugHandler;
 use function extension_loaded;
-use function ini_get as ini_get_unsafe;
 use const PHP_SAPI;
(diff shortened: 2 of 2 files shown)


peter279k:
To verify the different `XDebug` enabling mode mechanism in different versions, I setup the PHP 7.4.10 with `XDebug 2.9` version.

I fall back to the `Infection 0.26.6` version, and I try to collect the `ini_get_unsafe('xdebug.mode')` and `extension_loaded('xdebug')` values, I found the result is as follows:

- `ini_get_unsafe('xdebug.mode')` is `false` and it can be verified because this value is only available in `XDebug 3.x` version.
- `extension_loaded('xdebug')` is `true` and it's the different value in `XDebug 3.x` version.

Finally, I found the `XDebug` `v3.1.2` version has the problem when getting the `ini_get_unsafe('xdebug.mode')` and `extension_loaded('xdebug')` values.

And I try to use the latest `XDebug` `v3.1.5` version doesn't have the problem when collecting the `ini_get_unsafe('xdebug.mode')` and `extension_loaded('xdebug')` values.

sanmai (approved):

maks-rafalko (approved):
> Finally, I found the `XDebug` `v3.1.2` version has the problem when getting the `ini_get_unsafe('xdebug.mode')` and `extension_loaded('xdebug')` values.

that explains strange behavior, thanks for digging into it

Released https://github.com/infection/infection/releases/tag/0.26.13

Issue #1722, opened 2022-09-02
Support `--threads=max`


dkarlovi:
**Is your feature request related to a problem? Please describe.**
If running on multiple platforms, the user needs to decide which of the ways [documented in the docs](https://infection.github.io/guide/command-line-options.html#threads-or-j) needs to be used. This makes scripting more difficult.

**Describe the solution you'd like**
Add `--threads=max` where `max` represents the value which would have been obtained by using the above linked methods.

**Additional context**
Psalm does it and [even makes `--threads=max` the default](https://github.com/vimeo/psalm/issues/2795#issuecomment-587255089). I'm guessing the count detection code could be stolen from there.

PR #1723, opened 2022-09-03
Introduce `--threads=max` option value to automatically detect the number of CPU cores

maks-rafalko:
Implements https://github.com/infection/infection/issues/1722

Doc PR: https://github.com/infection/site/pull/240

### DIFF
--- a/src/Resource/Processor/CpuCoresCountProvider.php
+++ b/src/Resource/Processor/CpuCoresCountProvider.php
@@ -42,15 +42,10 @@ use function filter_var;
 use function function_exists;
 use function is_int;
 use function is_readable;
-use LogicException;
-use const PHP_OS;
-use const PHP_VERSION;
 use function Safe\file_get_contents;
-use function Safe\ini_get;
 use function Safe\shell_exec;
 use function substr_count;
 use function trim;
-use function version_compare;
 
 /**
  * @internal
@@ -66,18 +61,11 @@ final class CpuCoresCountProvider
             return 1;
         }
 
-        if (ini_get('pcre.jit') === '1'
-            && PHP_OS === 'Darwin'
-            && version_compare(PHP_VERSION, '7.3.0') >= 0
-            && version_compare(PHP_VERSION, '7.4.0') < 0
-        ) {
-            return 1;
-        }
-
         if (!extension_loaded('pcntl') || !function_exists('shell_exec')) {
             return 1;
         }
 
+        // for Linux
         $hasNproc = trim(@shell_exec('command -v nproc'));
 
         if ($hasNproc !== '') {
@@ -89,6 +77,7 @@ final class CpuCoresCountProvider
             }
         }
 
+        // for MacOS
         $ncpu = trim(shell_exec('sysctl -n hw.ncpu'));
         $cpuCount = filter_var($ncpu, FILTER_VALIDATE_INT);
 
@@ -97,14 +86,14 @@ final class CpuCoresCountProvider
         }
 
         if (is_readable('/proc/cpuinfo')) {
-            $cpuinfo = file_get_contents('/proc/cpuinfo');
-            $cpuCount = substr_count($cpuinfo, 'processor');
+            $cpuInfo = file_get_contents('/proc/cpuinfo');
+            $cpuCount = substr_count($cpuInfo, 'processor');
 
             if ($cpuCount > 0) {
                 return $cpuCount;
             }
         }
 
-        throw new LogicException('Failed to detect number of CPUs!');
+        return 1;
     }
 }
--- a/src/Resource/Processor/CpuCoresCountProvider.php
+++ b/src/Resource/Processor/CpuCoresCountProvider.php
@@ -1,0 +1,110 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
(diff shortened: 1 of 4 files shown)


src/Resource/Processor/CpuCoresCountProvider.php
```diff
@@ -0,0 +1,110 @@
...
+    public static function provide(): int
+    {
+        if (defined('PHP_WINDOWS_VERSION_MAJOR')) {
+            return 1;
+        }
+
+        if (ini_get('pcre.jit') === '1'
+            && PHP_OS === 'Darwin'
+            && version_compare(PHP_VERSION, '7.3.0') >= 0
+            && version_compare(PHP_VERSION, '7.4.0') < 0
```
sanmai, line 72: Should we even care about PHP 7.4?

src/Resource/Processor/CpuCoresCountProvider.php
```diff
@@ -0,0 +1,110 @@
...
+        if (is_readable('/proc/cpuinfo')) {
+            $cpuinfo = file_get_contents('/proc/cpuinfo');
+            $cpuCount = substr_count($cpuinfo, 'processor');
+
+            if ($cpuCount > 0) {
+                return $cpuCount;
+            }
+        }
+
+        throw new LogicException('Failed to detect number of CPUs!');
```
sanmai, line 108: Why not `return 1;` in this case?

src/Resource/Processor/CpuCoresCountProvider.php
```diff
@@ -0,0 +1,110 @@
...
+        if ($hasNproc !== '') {
+            $nproc = trim(shell_exec('nproc'));
+            $cpuCount = filter_var($nproc, FILTER_VALIDATE_INT);
+
+            if (is_int($cpuCount)) {
+                return $cpuCount;
+            }
+        }
+
+        $ncpu = trim(shell_exec('sysctl -n hw.ncpu'));
```
sanmai, line 92: It is unclear why we assume `nproc` may not be there while `sysctl` supposed to be always there. Should we even check that `nproc` exists? Because if it isn't clearly we won't be getting a single digit :thinking:

sanmai (changes requested):
Left few comments, mostly nits. The only real deal is `LogicException` at the end of `provide()`.

sidz:
My 2 cents: it is not Infection responsibility. We just make our codebase more complicated. `-j$(nproc)` or `-j$(sysctl -n hw.ncpu)` works fine. End user can use Make or composer scripts if he don't want to type `--threads=`

src/Resource/Processor/CpuCoresCountProvider.php
```diff
@@ -0,0 +1,110 @@
...
+     */
+    public static function provide(): int
+    {
+        if (defined('PHP_WINDOWS_VERSION_MAJOR')) {
+            return 1;
+        }
+
+        if (ini_get('pcre.jit') === '1'
+            && PHP_OS === 'Darwin'
+            && version_compare(PHP_VERSION, '7.3.0') >= 0
```
sidz, line 71: can't we use `PHP_VERSION_ID` which contains an integer value instead of trying to compare strings?

src/Resource/Processor/CpuCoresCountProvider.php
maks-rafalko: this is for MacOS case, where `nproc` is not available so the code will be here, and `sysctl -n hw.ncpu` will return the number of cores. Checked on Linus and Macos machines, added comments

maks-rafalko:
Addressed your comments, initial version was blindly copied from Psalm.

> My 2 cents: it is not Infection responsibility. We just make our codebase more complicated.

Well, yes and no. IMO it's all about "those tiny things for DX". You set `--threads=max` and 

- don't care about any potential cases where `$(nproc)` might not work because of incorrect escaping (in `Jenkinsfile` you will have to prepend it like `sh 'infection ... --threads=\$(nproc)`'`
- don't care and just let anyone copy this command line from Linux to MacOS machines and vice-versa

Yes it's completely not critical and both points can be ignored, but it's a little thing that might be more convenient for someone (which is why this feature request was born from our end user, not me :) ).

sanmai (approved):
I agree this is a very dangerous option, yet if people want this I guess that's what we have to add :thinking:

dkarlovi:
@sanmai what makes it dangerous?

sanmai:
- Infection does surprising things with the code. For one, it creates infinite loops. And could include something consuming memory. It could make you machine unresponsive for prolonged periods of time and generally lead to bad experiences, especially if that's the first time you're using Infection.
- Running Infection with more threads not necessary leads to better performance. Unlike, say, pure CPU-bound tasks such as static analysis. Thus this option could give a false promise of fast execution where in fact it will work in the opposite way.

I would seriously object enabling this option by default, or recommending it to new users.

dkarlovi:
That means the threads upper limit should be restricted / validated regardless of how you specify it, which further proves the idea Infection should know the upper limit.

For example, if max=12, what difference is it if I say --threads=12 or --threads=max, from the user POV this is totally equivalent. If running it like that is "dangerous", infection should disallow or warn the user about it.

Adding --threads=max seems unrelated (with using max threads being "dangerous") to me.

sanmai:
Infection can't know the upper limit. Say, if your test suite is predominantly IO-bound, then it makes sense to run Infection with more threads than you have CPU. 

Do you think there's a sure way what to do in this situation? How can we know if a test suite is CPU-bound or IO-bound to get the best results?

dkarlovi:
This is knowledge you have, the user doesn't. It sounds like Infection could have named threads calculator strategies which the user can try out and use some which make sense.

For example, `conservative`, `per-cpu`, `io-bound`, etc.

dkarlovi:
@maks-rafalko what do you think here?

sanmai:
> threads calculator strategies

How we should approach this idea, given Infection is a PHP app?

dkarlovi:
@sanmai not sure what you mean here?

Since now infection knows the number of CPUs available, I can imagine you could do something like
- `--threads=conservative` = number of CPUs * 0.8
- `--threads=per-cpu` = number of CPUs, which is current max
- `--threads=io-bound` = number of CPUs * 1.5

All names and numbers are illustrational, of course.

src/Resource/Processor/CpuCoresCountProvider.php
```diff
@@ -0,0 +1,99 @@
...
+use function is_readable;
+use function Safe\file_get_contents;
+use function Safe\shell_exec;
+use function substr_count;
+use function trim;
+
+/**
+ * @internal
+ */
+final class CpuCoresCountProvider
```
theofidry, line 53: What do you think of `CpuInfo::getCount()` as a name instead?

src/Resource/Processor/CpuCoresCountProvider.php
```diff
@@ -0,0 +1,99 @@
...
+{
+    /**
+     * Copied and adapter from Psalm project: https://github.com/vimeo/psalm/blob/4.x/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php#L1454
+     */
+    public static function provide(): int
+    {
+        if (defined('PHP_WINDOWS_VERSION_MAJOR')) {
```
theofidry, line 60: Might be worth caching the result? Although I don't expect it to be called twice, you also don't expect the system to change the number of CPU cores available within the same process and it's only a `int` value

src/Resource/Processor/CpuCoresCountProvider.php
```diff
@@ -0,0 +1,99 @@
...
+use function substr_count;
+use function trim;
+
+/**
+ * @internal
+ */
+final class CpuCoresCountProvider
+{
+    /**
+     * Copied and adapter from Psalm project: https://github.com/vimeo/psalm/blob/4.x/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php#L1454
```
theofidry, line 56: Is there any chance Psalm is willing to extract it to a dedicated package? Otherwise provided they would be interested in a package, I'd be happy to provide one

theofidry:
Sorry for the late review, great addition! Happy to do the changes myself

src/Resource/Processor/CpuCoresCountProvider.php
dkarlovi: This might be interesting as some sort of tiny "system information" HAL package which could be reused across the ecosystem indeed. :+1:

Issue #1725, opened 2022-09-08
Do not mutate `$smth ?? null` to `null ?? $smth`


maks-rafalko:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.26.14
| Test Framework version | PHPUnit
| PHP version | 8.1
| Example | https://infection-php.dev/r/p5o1

Looks like `Coalesce` should not do the following mutation, where `null` is used:

```diff
- $var = $smth ?? null;
+ $var = null ?? $smth;
```

because it's impossible to kill it.

PR #1726, opened 2022-09-09
Do not mutate Coalesce operator when `null` is used on the right

maks-rafalko:
Because it's impossible to kill such Mutant

Fixes https://github.com/infection/infection/issues/1725


sidz (approved):
:+1:

maks-rafalko:
Thank you @sidz

Issue #1673, opened 2022-02-27
False Positive on Exception code increments


dominikzogg:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.26.5
| Test Framework version | PHPUnit 9.5.16
| PHP version | 8.0.8
| Platform    |  Ubuntu
| Github Repo | https://github.com/chubbyphp/chubbyphp-framework

As you can see, i get escaped mutants, but i can't reproduce them, doing the same manipulations myself.

https://dashboard.stryker-mutator.io/reports/github.com/chubbyphp/chubbyphp-framework/master#mutant/Router/Exceptions/MissingRouteAttributeOnRequestException.php

```
Not Covered mutants:
====================

1) /home/schreiner/Projects/default/chubbyphp/chubbyphp-framework/src/Router/Exceptions/MissingRouteAttributeOnRequestException.php:25    [M] DecrementInteger

--- Original
+++ New
@@ @@
      */
     public static function create($route) : self
     {
-        return new self(sprintf('Request attribute "route" missing or wrong type "%s", please add the "%s" middleware', get_debug_type($route), RouteMatcherMiddleware::class), 1);
+        return new self(sprintf('Request attribute "route" missing or wrong type "%s", please add the "%s" middleware', get_debug_type($route), RouteMatcherMiddleware::class), 0);
     }
 }


2) /home/schreiner/Projects/default/chubbyphp/chubbyphp-framework/src/Router/Exceptions/MissingRouteAttributeOnRequestException.php:25    [M] IncrementInteger

--- Original
+++ New
@@ @@
      */
     public static function create($route) : self
     {
-        return new self(sprintf('Request attribute "route" missing or wrong type "%s", please add the "%s" middleware', get_debug_type($route), RouteMatcherMiddleware::class), 1);
+        return new self(sprintf('Request attribute "route" missing or wrong type "%s", please add the "%s" middleware', get_debug_type($route), RouteMatcherMiddleware::class), 2);
     }
 }


3) /home/schreiner/Projects/default/chubbyphp/chubbyphp-framework/src/Router/Exceptions/RouteGenerationException.php:32    [M] DecrementInteger

--- Original
+++ New
@@ @@
      */
     public static function create(string $name, string $path, array $attributes, ?\Throwable $previous = null) : self
     {
[...]

PR #655, opened 2019-03-08
Use parent array for coverage

BackEndTea:
This PR:

- [x] Handles array coverage a bit better
- [ ] Covered by tests

As shown here, array coverage is a bit wonky, and not all items are considered covered, except if function/methods are called as part of the array. 
This PR should make sure that any array is properly covered.
image

It isn't properly tested, as the MutationsCollectorVisitor is a bit complex. I'll create a separate PR, to fix that up later.

Fixes https://github.com/infection/infection/issues/652

PR #1734, opened 2022-10-06
Correctly handle exception code mutants

romm:
I've been having this issue but don't really know where to start in order to fix this, so I thought the best way would be to provide a failing test.

When Infection creates the `DecrementInteger` and `IncrementInteger` mutants on an exception code, even if this code is part of a test, the report will mark those two mutants as uncovered. Changing the code manually leads to a failing test.

Any hint on how we could fix this behavior?

### DIFF
--- a/tests/e2e/Exception_Code/README.md
+++ b/tests/e2e/Exception_Code/README.md
@@ -1,0 +1,4 @@
+# Exception code mutants are killed
+
+Tests that `DecrementInteger` and `IncrementInteger` mutants on exception code
+are killed.
--- a/tests/e2e/Exception_Code/composer.json
+++ b/tests/e2e/Exception_Code/composer.json
@@ -1,0 +1,15 @@
+{
+    "require-dev": {
+        "phpunit/phpunit": "^9.5.5"
+    },
+    "autoload": {
+        "psr-4": {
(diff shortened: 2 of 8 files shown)


maks-rafalko:
I think this is the same issue as https://github.com/infection/infection/issues/1673 and the root cause is because coverage driver doesn't mark this line as covered. So in the first place this is not an Infection bug.

However, we did a fix previously for similar issue and if you want to help fixing this issue with exceptions, please look at https://github.com/infection/infection/pull/655

romm:
Thanks for the hint, I'll try to take a look at it. 🙂

romm:
This PR will be solved with #1777 and work done in sebastianbergmann/php-code-coverage#964.

I can close this PR, but we could also rebase on `master` when #1777 is merged, so this library has a new test to prevent future regression with this particular case.

WDYT?

maks-rafalko:
If you have time and energy to wait & rebase and complete this test, that would be awesome IMO

maks-rafalko:
@romm could you please rebase , #1777 is merged

maks-rafalko (approved):
Thank you @romm 

Now it passes

romm:
Thank you for the reminder! 😊

Issue #1735, opened 2022-10-10
[TestLocator::getTestsForFunctionSignature] InvalidArgumentException: Expected an array to contain 1 elements. Got: 8.


Slamdunk:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.26.15
| Test Framework version | PHPUnit 9.5.25
| PHP version | 8.1.11
| Platform    | Ubuntu

### Source code

```php
<?php

namespace Foo;

final class FooBar
{
    private function fooBar(): void
    {
        $var = [
            new class() {
                public function foo(): bool
                {
                    return true;
                }
            },
        ];
    }
}
```

### Error trace

```
Generate mutants...

Processing source code files: 4/9
In Assert.php line 2074:
                                                    
  [Webmozart\Assert\InvalidArgumentException]       
  Expected an array to contain 1 elements. Got: 8.  
                                                    

Exception trace:
  at ./vendor/webmozart/assert/src/Assert.php:2074
 Webmozart\Assert\Assert::reportInvalidArgument() at ./vendor/webmozart/assert/src/Assert.php:780
 Webmozart\Assert\Assert::eq() at ./vendor/webmozart/assert/src/Assert.php:1745
 Webmozart\Assert\Assert::count() at ./vendor/infection/infection/src/TestFramework/Coverage/XmlReport/TestLocator.php:86
 Infection\TestFramework\Coverage\XmlReport\TestLocator->getTestsForFunctionSignature() at n/a:n/a
 iterator_to_array() at ./vendor/infection/infection/src/Mutator/NodeMutationGenerator.php:183
 Infection\Mutator\NodeMutationGenerator->getAllTestsForCurrentNode() at ./vendor/infection/infection/src/Mutator/NodeMutationGenerator.php:132
 Infection\Mutator\NodeMutationGenerator->generateForMutator() at ./vendor/infection/infection/src/Mutator/NodeMutationGenerator.php:109
 Infection\Mutator\NodeMutationGenerator->generate() at ./vendor/infection/infection/src/PhpParser/Visitor/MutationCollectorVisitor.php:77
 Infection\PhpParser\Visitor\MutationCollectorVisitor->getMutations() at ./vendor/infection/infection/src/Mutation/FileMutationGenerator.php:101
 Infection\Mutation\FileMutationGenerator->generate() at ./vendor/infection/infection/src/Mutation/MutationGenerator.php:92
 Infection\Mutation\MutationGenerator->generate() at n/a:n/a
[...]

PR #1745, opened 2022-10-18
Fix the issue with anonymous classes inside arrays

Slamdunk:
This PR shows https://github.com/infection/infection/issues/1735 in action: https://github.com/infection/infection/actions/runs/3272547451/jobs/5383690401#step:13:265

Despite been related to `PublicVisibility` (and `ProtectedVisibility` as well), I have not been able to get the error in a Unit Test within `\Infection\Tests\Mutator\FunctionSignature\PublicVisibilityTest`.

I'd like to fix this and propose it here, but I'm stuck: I have no clue where to start, can you help me please with some hits?

Quick command to get the error locally: `(cd tests/e2e/Github1735/ && php ../../../bin/infection -v)`

Fixes https://github.com/infection/infection/issues/1735

### DIFF
--- a/src/TestFramework/Coverage/LineRangeCalculator.php
+++ b/src/TestFramework/Coverage/LineRangeCalculator.php
@@ -44,15 +44,17 @@ use PhpParser\Node;
  */
 final class LineRangeCalculator
 {
-    public function calculateRange(Node $node): NodeLineRangeData
+    public function calculateRange(Node $originalNode): NodeLineRangeData
     {
-        $node = $this->getOuterMostArrayNode($node);
+        if ($originalNode->getAttribute(ReflectionVisitor::IS_ON_FUNCTION_SIGNATURE, false) === true) {
+            $startLine = $originalNode->getStartLine();
+            // function signature node should always be 1-line range: (start, start)
+            return new NodeLineRangeData($startLine, $startLine);
+        }
 
-        $endLine = $node->getAttribute(ReflectionVisitor::IS_ON_FUNCTION_SIGNATURE, false) === true
-            ? $node->getStartLine() // function signature node should always be 1-line range: (start, start)
-            : $node->getEndLine();
+        $outerMostArrayNode = $this->getOuterMostArrayNode($originalNode);
 
-        return new NodeLineRangeData($node->getStartLine(), $endLine);
+        return new NodeLineRangeData($outerMostArrayNode->getStartLine(), $outerMostArrayNode->getEndLine());
     }
 
     /**
--- a/tests/e2e/Github1735/README.md
+++ b/tests/e2e/Github1735/README.md
@@ -1,0 +1,3 @@
+# Github 1735
+
+* Related to https://github.com/infection/infection/issues/1735
--- a/tests/e2e/Github1735/composer.json
+++ b/tests/e2e/Github1735/composer.json
@@ -1,0 +1,15 @@
+{
+    "require-dev": {
+        "phpunit/phpunit": "^9.5.20"
+    },
+    "autoload": {
+        "psr-4": {
+            "Github1735\\": "src/"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "Github1735\\Test\\": "tests/"
+        }
+    }
+}
--- a/tests/e2e/Github1735/expected-output.txt
+++ b/tests/e2e/Github1735/expected-output.txt
@@ -1,1 +1,10 @@
-# Effects per Mutator
+Total: 2
+
+Killed: 2
+Errored: 0
+Syntax Errors: 0
+Escaped: 0
+Timed Out: 0
+Skipped: 0
+Ignored: 0
+Not Covered: 0
--- a/tests/e2e/Github1735/expected-output.txt
+++ b/tests/e2e/Github1735/expected-output.txt
@@ -1,0 +1,1 @@
+# Effects per Mutator
--- a/tests/e2e/Github1735/infection.json
+++ b/tests/e2e/Github1735/infection.json
@@ -9,7 +9,7 @@
         "PublicVisibility": true
     },
     "logs": {
-        "perMutator": "infection.log"
+        "summary": "infection.log"
     },
     "tmpDir": "."
 }
--- a/tests/e2e/Github1735/infection.json
+++ b/tests/e2e/Github1735/infection.json
@@ -1,0 +1,15 @@
+{
+    "timeout": 3,
+    "source": {
+        "directories": [
+            "src"
+        ]
+    },
+    "mutators": {
+        "PublicVisibility": true
+    },
+    "logs": {
+        "perMutator": "infection.log"
+    },
+    "tmpDir": "."
+}
--- a/tests/e2e/Github1735/phpunit.xml
+++ b/tests/e2e/Github1735/phpunit.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/schema/9.2.xsd"
+         bootstrap="./vendor/autoload.php"
+         colors="true"
+>
+    <testsuites>
+        <testsuite name="Test Suite">
+            <directory>./tests/</directory>
+        </testsuite>
+    </testsuites>
+
+    <filter>
+        <whitelist>
+            <directory>./src/</directory>
+        </whitelist>
+    </filter>
+</phpunit>
--- a/tests/e2e/Github1735/phpunit.xml
+++ b/tests/e2e/Github1735/phpunit.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/schema/9.2.xsd"
+         xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/phpunit.xsd"
          bootstrap="./vendor/autoload.php"
          colors="true"
 >
@@ -10,9 +10,9 @@
         </testsuite>
     </testsuites>
 
-    <filter>
-        <whitelist>
+    <coverage>
+        <include>
             <directory>./src/</directory>
-        </whitelist>
-    </filter>
+        </include>
+    </coverage>
 </phpunit>
--- a/tests/e2e/Github1735/src/SourceClass.php
+++ b/tests/e2e/Github1735/src/SourceClass.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Github1735;
+
+class SourceClass
+{
+    public function hello(): array
+    {
+        $var = [
+            'class' => new class() {
+                public function foo(): bool
+                {
+                    return true;
+                }
+            },
+        ];
+
+        return $var;
+    }
+}
--- a/tests/e2e/Github1735/tests/SourceClassTest.php
+++ b/tests/e2e/Github1735/tests/SourceClassTest.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Github1735\Test;
+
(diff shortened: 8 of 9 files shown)


Slamdunk:
Man, `Generator`s are awesome, except when you have to debug an infinite chain of them :(

image

sanmai:
We can see where it happens in the stack trace:
```
Exception trace:
  at vendor/webmozart/assert/src/Assert.php:2042
 Webmozart\Assert\Assert::reportInvalidArgument() at vendor/webmozart/assert/src/Assert.php:762
 Webmozart\Assert\Assert::eq() at vendor/webmozart/assert/src/Assert.php:1727
 Webmozart\Assert\Assert::count() at src/TestFramework/Coverage/XmlReport/TestLocator.php:86
 Infection\TestFramework\Coverage\XmlReport\TestLocator->getTestsForFunctionSignature() at n/a:n/a
```

sanmai:
This is where we initialize the range:

https://github.com/infection/infection/blob/c770354f156043bf25cfd434fe226f52e476428e/src/TestFramework/Coverage/LineRangeCalculator.php#L47-L56

It looks like `ReflectionVisitor::IS_ON_FUNCTION_SIGNATURE` detection is failing here.

sanmai:
There's something very confusing because if `IS_ON_FUNCTION_SIGNATURE` is set we should be getting a single element array, and what's more confusing is that `NodeMutationGenerator` refers to the same flag and it isn't the same there.

https://github.com/infection/infection/blob/804227770afbb876b4cbd24552535ddc0cdd331a/src/Mutator/NodeMutationGenerator.php#L177-L180

I don't think I have the mental capacity to understand this right now.

sanmai:
It is interesting that if I restructure the fixture like so, the error goes away:
```diff
--- a/tests/e2e/Github1735/src/SourceClass.php
+++ b/tests/e2e/Github1735/src/SourceClass.php
@@ -6,13 +6,15 @@ class SourceClass
 {
     public function hello(): array
     {
+        $class = new class() {
+            public function foo(): bool
+            {
+                return true;
+            }
+        };
+
         $var = [
-            'class' => new class() {
-                public function foo(): bool
-                {
-                    return true;
-                }
-            },
+            'class' => $class,
         ];
 
         return $var;
```

maks-rafalko:
Thank you @Slamdunk for reporting the issue and creating a reproducer, this helps so much. Thanks @sanmai for the valuable feedback, helped me quicker understand the issue.

So, the bug in because of line 

https://github.com/infection/infection/blob/4053dfd6988ff3ea9fd552550e171e92fdb347fb/src/TestFramework/Coverage/LineRangeCalculator.php#L49

> what's more confusing is that NodeMutationGenerator refers to the same flag and it isn't the same there.

this is because here

https://github.com/infection/infection/blob/804227770afbb876b4cbd24552535ddc0cdd331a/src/Mutator/NodeMutationGenerator.php#L177-L180

we use original node, but here

https://github.com/infection/infection/blob/c770354f156043bf25cfd434fe226f52e476428e/src/TestFramework/Coverage/LineRangeCalculator.php#L47-L56

we use "outermost array node", wich is completely separate node for which `IS_ON_FUNCTION_SIGNATURE` is no longer `true` but `false`.

That being said, the update I pushed is just to use original node (`public function foo()`) to determine the start/end line of the method inside an anonymous class, instead of using outermost array node that contains this anonymous class.

The reported bug is fixed.

---

HOWEVER, I still think Infection can't (or is not able to currently) working with anonymous classes properly, even if this particular test passes.

Why?

This is how the XML coverage report look like:

<details>
```xml
<?xml version="1.0"?>
<phpunit xmlns="https://schema.phpunit.de/coverage/1.0">
  <file name="SourceClass.php" path="/">
    <totals>
      <lines total="21" comments="0" code="21" executable="4" executed="4" percent="100.00"/>
      <methods count="1" tested="1" percent="100.00"/>
      <functions count="0" tested="0" percent="0"/>
      <classes count="1" tested="1" percent="100.00"/>
      <traits count="0" tested="0" percent="0"/>
    </totals>
    <class name="Github1735\SourceClass" start="5" executable="4" executed="4" crap="1">
      <namespace name="Github1735"/>
      <method name="hello" signature="hello(): array" start="7" end="19" crap="1" executable="4" executed="4" coverage="100"/>
    </class>
    <coverage>
      <line nr="9">
        <covered by="Github1735\Test\SourceClassTest::test_hello"/>
      </line>
      <line nr="10">
        <covered by="Github1735\Test\SourceClassTest::test_hello"/>
      </line>
      <line nr="13">
        <covered by="Github1735\Test\SourceClassTest::test_hello"/>
      </line>
      <line nr="18">
        <covered by="Github1735\Test\SourceClassTest::test_hello"/>
      </line>
    </coverage>
    <source>
      <line no="1">
        <token name="T_OPEN_TAG">&lt;?php</token>
      </line>
      <line no="2"/>
      <line no="3">
        <token name="T_NAMESPACE">namespace</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_STRING">Github1735</token>
        <token name="T_SEMICOLON">;</token>
      </line>
      <line no="4"/>
      <line no="5">
        <token name="T_CLASS">class</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_STRING">SourceClass</token>
      </line>
      <line no="6">
        <token name="T_OPEN_CURLY">{</token>
      </line>
      <line no="7">
        <token name="T_WHITESPACE">    </token>
        <token name="T_PUBLIC">public</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_FUNCTION">function</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_STRING">hello</token>
        <token name="T_OPEN_BRACKET">(</token>
        <token name="T_CLOSE_BRACKET">)</token>
        <token name="T_COLON">:</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_ARRAY">array</token>
      </line>
      <line no="8">
        <token name="T_WHITESPACE">    </token>
        <token name="T_OPEN_CURLY">{</token>
      </line>
      <line no="9">
        <token name="T_WHITESPACE">        </token>
        <token name="T_VARIABLE">$var</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_EQUAL">=</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_OPEN_SQUARE">[</token>
      </line>
      <line no="10">
        <token name="T_WHITESPACE">            </token>
        <token name="T_CONSTANT_ENCAPSED_STRING">'class'</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_DOUBLE_ARROW">=&gt;</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_NEW">new</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_CLASS">class</token>
        <token name="T_OPEN_BRACKET">(</token>
        <token name="T_CLOSE_BRACKET">)</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_OPEN_CURLY">{</token>
      </line>
      <line no="11">
        <token name="T_WHITESPACE">                </token>
        <token name="T_PUBLIC">public</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_FUNCTION">function</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_STRING">foo</token>
        <token name="T_OPEN_BRACKET">(</token>
        <token name="T_CLOSE_BRACKET">)</token>
        <token name="T_COLON">:</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_STRING">bool</token>
      </line>
      <line no="12">
        <token name="T_WHITESPACE">                </token>
        <token name="T_OPEN_CURLY">{</token>
      </line>
      <line no="13">
        <token name="T_WHITESPACE">                    </token>
        <token name="T_RETURN">return</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_STRING">true</token>
        <token name="T_SEMICOLON">;</token>
      </line>
      <line no="14">
        <token name="T_WHITESPACE">                </token>
        <token name="T_CLOSE_CURLY">}</token>
      </line>
      <line no="15">
        <token name="T_WHITESPACE">            </token>
        <token name="T_CLOSE_CURLY">}</token>
        <token name="T_COMMA">,</token>
      </line>
      <line no="16">
        <token name="T_WHITESPACE">        </token>
        <token name="T_CLOSE_SQUARE">]</token>
        <token name="T_SEMICOLON">;</token>
      </line>
      <line no="17"/>
      <line no="18">
        <token name="T_WHITESPACE">        </token>
        <token name="T_RETURN">return</token>
        <token name="T_WHITESPACE"> </token>
        <token name="T_VARIABLE">$var</token>
        <token name="T_SEMICOLON">;</token>
      </line>
      <line no="19">
        <token name="T_WHITESPACE">    </token>
        <token name="T_CLOSE_CURLY">}</token>
      </line>
      <line no="20">
        <token name="T_CLOSE_CURLY">}</token>
      </line>
      <line no="21"/>
    </source>
  </file>
</phpunit>
```
</details>

Look at this piece:

```xml
<class name="Github1735\SourceClass" start="5" executable="4" executed="4" crap="1">
      <namespace name="Github1735"/>
      <method name="hello" signature="hello(): array" start="7" end="19" crap="1" executable="4" executed="4" coverage="100"/>
</class>
```

Only 1 method (`hello`) is marged as covered, while we have 2 methods (`hello`, `foo`) inside the class `SourceClass`:

* `hello` is a method of `SourceClass`
* `foo` is a method of `class@anonymous`

What it breaks is that Infection incorrectly determines that the test that covers `hello` method also covers `foo` method. While it is the case in the current state of this PR, I can easily break this assumption by this change in the test:

```diff
    public function test_hello()
    {
        $sourceClass = new SourceClass();
-       self::assertTrue($sourceClass->hello()['class']->foo());
+       self::assertIsArray($sourceClass->hello());
    }
```

In this case, `test_hello` test case does cover `SourceClass::hello()` method, but **does not cover** `class@anonymous::foo()` method.

In practice, this leads to a bug with `--only-covered` option.

With the change above to the test, if I run

```
cd tests/e2e/Github1735/ && php ../../../bin/infection --only-covered
```

then I still get 2 mutations, one of them is escaped. However, `class@anonymous::foo()` is **not** covered and should not be mutated.

What to do with this? I don't know, because PHPUnit doesn't provide a coverage data Infection needs to build `TestLocations` for `foo()` method.

I think we should leave it as is for now, as I don't see a proper solution but at the same time mutating methods of anonymous classes located in other methods of real classes still sounds good and helpful.

maks-rafalko (approved):

src/TestFramework/Coverage/LineRangeCalculator.php
```diff
@@ -44,15 +44,17 @@
...
+            return new NodeLineRangeData($startLine, $startLine);
+        }
 
-        $endLine = $node->getAttribute(ReflectionVisitor::IS_ON_FUNCTION_SIGNATURE, false) === true
-            ? $node->getStartLine() // function signature node should always be 1-line range: (start, start)
-            : $node->getEndLine();
+        $outerMostArrayNode = $this->getOuterMostArrayNode($originalNode);
 
-        return new NodeLineRangeData($node->getStartLine(), $endLine);
+        return new NodeLineRangeData($outerMostArrayNode->getStartLine(), $outerMostArrayNode->getEndLine());
```
Slamdunk, line 57: When I first started a fix a week ago, I came to a conclusion similar to this, but all other tests were failing.
Glad to see I was in the right path. Still, the `Generator` thing makes debugging to painful :weary: 

Putting `iterator_to_array` everywhere has been tedious, I'd like to have a magic php settings to turn every `Generator` into an array immediately for debug purposes.

Slamdunk:
Thank you for your help :)

src/TestFramework/Coverage/LineRangeCalculator.php
maks-rafalko: > Still, the Generator thing makes debugging to painful weary

Same feelings here. I had to debug them a while ago and they just blew my mind, it's very hard. Though I didn't do any investigation on how to improve this process.

theofidry: Same here...

sanmai: It isn't that generators are at the fault but the structure we have here isn't friendly to generator-style programming.

Issue #1752, opened 2022-11-18
If condition negation mutator


manhunto:
**Is your feature request related to a problem? Please describe.**
Hello, I was surprised that infection doesn't support the negation of the whole `if` statement. For example, there is code like this

```php
if ($this->someMethodCall()) {
}
````

And any mutation wasn't mutate this code.

I expected that my code would be mutated into this

```php
if (!($this->someMethodCall())) {
}
````

There is a mutation that negates my condition but only when I write code like this, but I usually skip `=== true`.

```php
if ($this->someMethodCall() === true) {
}
````

**Describe the solution you'd like**
I cloned the infection repository, and I've tried to implement it, and I have POC of some conditions that can be mutated. But I want to know before the final implementation, there were some attempts to implement it, and it failed. Or maybe there are some other cases that I don't know now :)

Thank you, and have a good day!

**Comment by maks-rafalko** [comment]
From what I know/remember, we have never tried to do it and never discussed it, so I'm +1 here.

But I think we should not do the following mutation:

```diff
- if ($this->someMethodCall() === true) {
+ if (!($this->someMethodCall() === true)) {
}
``` 

because it's a duplicate of 

```diff
- if ($this->someMethodCall() === true) {
+ if ($this->someMethodCall() !== true) {
}
``` 
At the same time, let's look into more complex example and intersections with existing mutators:

Original code: `! (a() || b())`

Boolean matrix:

| a   |      b      |  result |
|----------|:-------------:|------:|
| true |  true | false |
| true |    false   |   false |
| false| true |    false |
| false| false |    true |

**Example 1:** New proposed mutation :

```diff
- ! (a() || b()) 
+ a() || b()
```

The boolean matrix of mutated code is completely different - nice:

| a   |      b      |  result |
|----------|:-------------:|------:|
| true |  true | true |
| true |    false   |   true |
| false| true |    true |
| false| false |    false |

And let's compare to the existing `LogicalOr` mutator:

```diff
- ! (a() || b()) 
+ ! (a() && b()) 
```

The boolean matrix is not duplicated, which is a good thing:

| a   |      b      |  result |
|----------|:-------------:|------:|
| true |  true | false |
| true |    false   |   true |
| false| true |    true |
| false| false |    true |

At this point, introducing the new proposed mutator gives us the new boolean matrix which is a great addition.

Now, let's try to mutate farther, inside the sub-conditions:

**Example 2:** negate `a()` call

```diff
- ! (a() || b()) 
[...]

PR #1753, opened 2022-11-27
Add negation mutators

manhunto:
This PR:

- [x] Add mutators that negate whole condition in `if` and `else if` (it only mutates simple expressions like function call, method call etc. to not duplicate with other existing mutators)
- [x] Add mutators that negate whole condition in `LogicalAnd` and `LogicalOr` expressions
- [x] Add mutators that negate every sub expression together in `LogicalAnd` and `LogicalOr` expressions (even in complex conditions)
- [x] Add mutators that negate every sub expression separately in `LogicalAnd` and `LogicalOr` expressions (even in complex conditions, only simple sub expressions like in `if` and `else if`)
- [x] Covered by tests
- [ ] Doc PR: TODO (if everything will be approved)

1. Negation of entire simple expression in if.
```diff
-if ($this->fooBar()) {
+if (!$this->fooBar()) {
```

2. Negation of entire simple expression in elseif.
```diff
if ($this->fooBar() {
- } elseif ($this->fooBaz()) {
+ } elseif (!$this->fooBaz()) {
}
```

3. Negation of entire && and || condition (even in if statements)
```diff
- $var = $a && $b;
+ $var = !($a && $b);
```

```diff
- $var = $a || $b;
+ $var = !($a || $b);
```

4. Negation of every sub expression in && and || condition (even in if statements)
```diff
- $var = $a && $b;
+ $var = !$a && !$b;
```

```diff
- $var = $a || $b;
+ $var = !$a || !$b;
```

5. Negation of every sub expression separately in && and || condition (even in if statements)
```diff
- $var = $a && $b;
# Mutation 1
+ $var = !$a && $b;
# Mutation 2
+ $var = $a && !$b;
```

```diff
- $var = $a || $b;
# Mutation 1
+ $var = !$a || $b;
# Mutation 2
+ $var = $a || !$b;
```

I've decided to create one PR for all mutators because they are related. They share traits and I think it is better to review it together, but I can split it if you think that it will be easier to maintain.

Closes: https://github.com/infection/infection/issues/1752


src/Mutator/Boolean/LogicalOrAllSubExprNegation.php
```diff
@@ -0,0 +1,105 @@
...
+use Infection\Mutator\MutatorCategory;
+use Infection\PhpParser\Visitor\ParentConnector;
+use PhpParser\Node;
+
+/**
+ * @internal
+ *
+ * @implements Mutator<Node\Expr\BinaryOp\BooleanOr>
+ */
+final class LogicalOrAllSubExprNegation implements Mutator
```
theofidry, line 50: Would it make sense to re-use the code from the other?

theofidry: Actually nevermind, it's a problem of duality that's already present with a few other mutators, I don't think it makes sense to try to fix it in this PR

src/Mutator/Operator/ElseifNegation.php
```diff
@@ -0,0 +1,94 @@
...
+use Infection\Mutator\MutatorCategory;
+use Infection\Mutator\SimpleExpression;
+use PhpParser\Node;
+
+/**
+ * @internal
+ *
+ * @implements Mutator<Node\Stmt\ElseIf_>
+ */
+final class ElseifNegation implements Mutator
```
theofidry, line 50:
```suggestion
final class ElseIfNegation implements Mutator
```
?

src/Mutator/Operator/ElseifNegation.php
```diff
@@ -0,0 +1,94 @@
...
+{
+    use GetMutatorName;
+    use SimpleExpression;
+
+    public static function getDefinition(): ?Definition
+    {
+        return new Definition(
+            'Negates an expression in elseif statement.',
```
theofidry, line 58:
```suggestion
            'Negates an elseif expression.',
```
?

src/Mutator/Operator/ElseifNegation.php
theofidry, line 50: Same here: does it really make sense to dissociate `elseif` and `if`? I understand from a node manipulation it's more atomic, but from a user perspective there is no difference between the two.

But I'm thinking this might be more of an issue caused by the current Mutator API, in which case this should be fixed separately.

src/Mutator/SimpleExpression.php
```diff
@@ -0,0 +1,54 @@
...
+namespace Infection\Mutator;
+
+use PhpParser\Node;
+
+/**
+ * @internal
+ */
+trait SimpleExpression
+{
+    private function isSimpleExpression(Node\Expr $expr): bool
```
theofidry, line 45: Regarding the name, from the implementation to I understand correctly that the idea here is to know if an expression contains only one operand as opposed to several ones like `BooleanAnd` for example?

If that's the case I would consider `isSingleOperandExpression`, WDYT?

src/Mutator/SimpleExpression.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        return $expr instanceof Node\Expr\FuncCall
```
theofidry, line 47: looking at the implementation the method could be static

src/Mutator/SimpleExpression.php
```diff
@@ -0,0 +1,54 @@
...
+declare(strict_types=1);
+
+namespace Infection\Mutator;
+
+use PhpParser\Node;
+
+/**
+ * @internal
+ */
+trait SimpleExpression
```
theofidry, line 43: From this alone I am not convinced there is any benefit of it being a trait. It could very well be a regular stateless class with a static method and a private constructor.

But maybe the rest of the team prefers traits... I'll let them voice their opinion on it

src/Mutator/SimpleExpression.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    private function isSimpleExpression(Node\Expr $expr): bool
+    {
+        return $expr instanceof Node\Expr\FuncCall
+            || $expr instanceof Node\Expr\MethodCall
+            || $expr instanceof Node\Expr\StaticCall
+            || $expr instanceof Node\Expr\Variable
+            || $expr instanceof Node\Expr\ArrayDimFetch
+            || $expr instanceof Node\Expr\ClassConstFetch;
```
theofidry, line 52: Note for other reviewers, I did not verify the list completeness

src/Mutator/Util/AbstractSingleSubExprNegation.php
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * @param array<string, mixed> $attributes
+     */
+    abstract protected function create(Node\Expr $left, Node\Expr $right, array $attributes): Node\Expr;
+
+    private function countSubExpressionsToNegate(Node\Expr $node, int &$count = 0): int
+    {
+        if ($this->isInstanceOf($node)) {
+            if (!isset($node->left, $node->right)) {
```
theofidry, line 89: from what I can see only `BinaryOp` nodes can have left & right within the `Expr` nodes, so maybe:

- `isInstanceOf` could be replaced by `isSupportedBinaryOpExpr(Node\Expr\AssignOp): bool`
- the existing `isInstanceOf()` calls could be combined with a `Node\Expr\BinaryOp` `instanceof` check – or this could be abstracted into a `isSupportedNode()` private method

This would also get rid of this `LogicException` thrown

theofidry (approved):
Sorry for the delay in the review, good job! 👏 

It's been too long since I've had a good view on the mutators and I didn't completely follow the conversation with @maks-rafalko  previously neither so I do think it deserve another round of review. As such I wouldn't recommend to address all the points yet just in case another maintainer disagrees with them :)

src/Mutator/Boolean/LogicalAndNegation.php
```diff
@@ -0,0 +1,100 @@
...
+
+        if ($parent instanceof Node\Expr\BinaryOp\BooleanAnd) {
+            return false;
+        }
+
+        if ($parent instanceof Node\Expr\BooleanNot) {
+            return false;
+        }
+
+        return true;
```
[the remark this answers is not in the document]
sanmai: Can we keep it the way it is now? 
Having separate return statements serves several goals:
- Makes it obvious if a certain statement is covered, without having to resort to expensive branch coverage reporting or mutation testing.
- Make it easy to add another return statement without having to obstruct the authorship of existing statements.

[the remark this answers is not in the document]
theofidry: I really don't feel any of those two things should dictate the code style, especially since not easily enforceable by a CS fixer, but as you prefer

src/Mutator/Boolean/LogicalAndAllSubExprNegation.php
```diff
@@ -0,0 +1,105 @@
...
+use Infection\Mutator\MutatorCategory;
+use Infection\PhpParser\Visitor\ParentConnector;
+use PhpParser\Node;
+
+/**
+ * @internal
+ *
+ * @implements Mutator<Node\Expr\BinaryOp\BooleanAnd>
+ */
+final class LogicalAndAllSubExprNegation implements Mutator
```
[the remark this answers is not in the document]
manhunto: I get your point and I agree with you, but keep it as separate mutators is better for maintenance.

tests/phpunit/Mutator/Boolean/LogicalAndAllSubExprNegationTest.php
```diff
@@ -0,0 +1,119 @@
...
+$var = !(!a() && b());
+PHP
+            ],
+        ];
+
+        yield 'It mutates more complex expressions' => [
+            <<<'PHP'
+<?php
+
+$var = $A > 1 && $this->foo() === false && self::bar() >= 10;
```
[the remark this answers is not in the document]
manhunto: Could give me some code example to better understand it?

manhunto:
Thank you for review @theofidry I've introducer some of changes proposed by you and I'm waiting for comments from another reviewers:)

manhunto:
Hi @maks-rafalko, any updates on this PR? :)

maks-rafalko:
Oh, somehow I missed this one, sorry. Could you please rebase so we can move forward?

maks-rafalko:
Ok, I managed to rebase it myself.

Notes for reviewers: we need to analyze if the provided mutators somehow generate duplicate muations (with existing in `master` mutators and between each other).

My idea is to create boolean matrices (as I did here https://github.com/infection/infection/issues/1752#issuecomment-1320848169) and compare all of them, finding duplicate results.

icanhazstring:
Just out of curiosity as I don't know the full functionality behind.

Isn't there a hashmap containing a list of, let's say, md5 hashes of the content of each mutated file? So that there can be no duplicate mutation?

Or is it because that some mutations look different but produce the same logical outcome?

maks-rafalko:
@icanhazstring I'm more about boolean matrices created by new mutators here. The simplest thing to understand it is [De Morgan's law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)

where `!$a && !$b === !($a || $b)`. By introducing many new different mutations, we can have duplicates in boolean matrices, which is not efficient as the same (logically) mutators will consume more and more time.

However, you question about hashes gives some source for thinking. We do have md5 hashing for each mutation, but I don't think we check uniqueness to avoid duplicates on physical, not logical level:

https://github.com/infection/infection/blob/0228b466ff2eb930bbc072b51accc76151e1b2c3/src/Mutation/Mutation.php#L173-L186

icanhazstring:
Maybe we could improve this hash based on the uniqueness of these logical mutations?
Don't know if we could come up with a parser implementation based on de morgan's law to figure out duplicates.

maks-rafalko (approved):
@manhunto this is absolutely awesome work. I've spent this day investigating the code and checking for duplicates. Made and update to not negate `$a === false & $b !== true` because it's duplicate of `TrueValue` / `FalseValue` mutators. Did the same for `Or` analogue.

Also, added more tests and simplified child classes to reduce the number of abstract methods.

I agree that from maintenance POV it's better to keep classes separate, at lease for now, because the logic is quite complex.

Thank you for your great contribution, going to merge it as soon as the build passes.

manhunto:
@maks-rafalko thank you for helping me with this PR -- for the changes you introduced in this PR and docs :)

PR #1754, opened 2022-11-28
Rename Makefile related variables

theofidry:
This PR is a pure renaming.

Working on the Makefiles a bit recently, I wanted to apply a few renamings in order to be more in line with the original Makefile terminoly. A TL:DR, this is a _rule_:

```
targets : pre-requisites
    recipe
```

We do use "target(s)" correctly (technically in Makefile, a rule may designate multiple targets, in our case though it does not matter). We use however "dependencies" instead of "pre-requisites" and "recipe" is unused (we do not consider the body of the rules).

So in this PR, I renamed:

- `targets` to `rules`: probably the most contentious one, it is slightly more correct although in our case the distinction not really existent
- `dependencies` to `pre-requisite`
- `targetSet` to `rule` (also makes the change of targets -> rules more natural)

### DIFF
--- a/tests/phpunit/AutoReview/Makefile/MakefileTest.php
+++ b/tests/phpunit/AutoReview/Makefile/MakefileTest.php
@@ -114,25 +114,25 @@ EOF;
 
     public function test_phony_targets_are_correctly_declared(): void
     {
-        $targets = Parser::parse(file_get_contents(self::MAKEFILE_PATH));
+        $rules = Parser::parse(file_get_contents(self::MAKEFILE_PATH));
 
         $phony = null;
         $targetComment = false;
         $matchedPhony = true;
 
-        foreach ($targets as [$target, $dependencies]) {
+        foreach ($rules as [$target, $prerequisites]) {
             if ($target === '.PHONY') {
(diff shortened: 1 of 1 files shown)


tests/phpunit/AutoReview/Makefile/MakefileTest.php
```diff
@@ -204,16 +204,16 @@ public function test_phony_targets_are_correctly_declared(): void
 
     public function test_no_target_is_being_declared_twice(): void
```
localheinz, line 205: @theofidry 

How do you feel about renaming the test methods as well?

theofidry: Happy to in general, worth second look with fresh eyes, but I would argue _this specific one_ is fine IMO, unless you have a more suited one? (it feels replacing target to rule in this method incorrect since what matters is not the fact that a rule may be declared twice, but that the targets are – the rules pre-requisites and recipes could be different)

sanmai (approved):

theofidry:
Since the failing build is MT which is unrelated to this PR, I'll merge it

PR #1755, opened 2022-11-28
Refactor MakefileTest

theofidry:
This is built upon #1754 so if you want to check a cleaner diff do skip the first two commits.

This PR is also pure refactoring: there is no new elements being tested, but it is more than simple renaming. The goal here is to prepare for further upcoming changes but I think those changes themselves should be good enough on their own.

### DIFF
--- a/tests/phpunit/AutoReview/Makefile/MakefileTest.php
+++ b/tests/phpunit/AutoReview/Makefile/MakefileTest.php
@@ -114,25 +114,25 @@ EOF;
 
     public function test_phony_targets_are_correctly_declared(): void
     {
-        $targets = Parser::parse(file_get_contents(self::MAKEFILE_PATH));
+        $rules = Parser::parse(file_get_contents(self::MAKEFILE_PATH));
 
         $phony = null;
         $targetComment = false;
         $matchedPhony = true;
 
-        foreach ($targets as [$target, $dependencies]) {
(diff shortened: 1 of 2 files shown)


tests/phpunit/AutoReview/Makefile/MakefileTest.php
```diff
@@ -278,84 +254,149 @@ static function (array $targetSet) use ($target, $dashCount): bool {
...
             array_filter(
-                $testTargets,
-                static function (array $targetSet): bool {
-                    return strpos($targetSet[0], 'test-') === 0
-                        && substr_count($targetSet[0], '-') === 2;
+                Parser::parse(file_get_contents(self::MAKEFILE_PATH)),
+                static function (array $rule) use ($filterDockerTarget): bool {
+                    [$target, $prerequisites] = $rule;
+
+                    $isCommentRule = count($prerequisites) !== 0 && str_starts_with($prerequisites[0], '## ');
```
theofidry, line 351: note that I replaced a few occurrences of `strpos()` to `str_starts_with`. I thought we had the polyfills setup but apparently this is not the case so this should be reverted

theofidry: Actually not a problem: the auto-review tests are run on PHP 8.0

theofidry:
Ready for review :)

sanmai (approved):

PR #1757, opened 2022-11-29
Add missing -c flag for touch commands

theofidry:
This flag ensures the `touch` call does not create the file. This is perfect because the only purpose of the touch call here is to update the timestamp, not create a file.

This was already used in a few places, but a few calls were missing it and a few places did not have the touch call although could make use of it.


sanmai (approved):

maks-rafalko:
Thank you @theofidry

PR #1762, opened 2022-12-01
Migrate to fidry/makefile

theofidry:
I created [fidry/makefile](https://github.com/theofidry/makefile) recently because I was in need of a simple Makefile rule parser like we started in Infection for both PHP-Scoper, other OSS projects for which I otherwise skipped the Makefile tests (didn't want to copy/paste) and private projects.

The core idea is identical: the parser is initially copied from here. The key differences are:

- a `Rule` object that makes it a bit nicer
- a few bug fixes
- more coverage

The [release section](https://github.com/theofidry/makefile/releases) provides more details (the first commit was the original parser unchanged).

The package license is also unchanged except the name... Not sure about that... Happy to sort it out if you feel there is something wrong there.


tests/phpunit/AutoReview/Makefile/MakefileTest.php
```diff
@@ -109,139 +67,11 @@ public function test_the_default_goal_is_the_help_command(): void
...
-                continue;
-            }
-
-            $phony = null;
-            $targetComment = false;
-            $matchedPhony = false;
-        }
-    }
-
-    public function test_no_target_is_being_declared_twice(): void
```
theofidry, line 209: this test is now in the parent test case

tests/phpunit/AutoReview/Makefile/MakefileTest.php
```diff
@@ -109,139 +67,11 @@ public function test_the_default_goal_is_the_help_command(): void
...
     }
 
-    public function test_the_makefile_can_be_parsed(): void
-    {
-        Parser::parse(file_get_contents(self::MAKEFILE_PATH));
-
-        $this->assertTrue(true);
-    }
-
-    public function test_phony_targets_are_correctly_declared(): void
```
theofidry, line 119: this test is now in the parent test case

maks-rafalko (approved):

PR #1765, opened 2022-12-02
Bump minimum PHP version required to PHP 8.1

### LINKED ISSUE #1760: Revisit the supported versions
I think Infection reached a good level of stability. Of course we can all find more features, bugs & co, but there is now a good set of versions behind it and a good usage adoption.

PHP 7.4 is EOL, 8.0 is in security fix only and 8.2 around the corner. What do you think of dropping support for PHP 7.4 & 8.0?

I think users that are still on 7.4 / 8.0 can live with the current infection versions and don't need (or cannot anyway) be on the bleeding edge.

Disclaimer: I do use infection in a few PHP 7.4 repo still.

theofidry:
Closes #1760

- Note that this PR does not take care of any code migration.
- A few issues have been identified but will be addressed separately

PR doc: https://github.com/infection/site/pull/250


.github/workflows/ci.yaml
```diff
@@ -17,13 +17,10 @@ jobs:
...
       matrix:
         operating-system: [ ubuntu-latest ]
-        php-version: [ '8.0' ]
+        php-version: [ '8.1' ]
         coverage-driver: [ pcov, xdebug ]
         e2e-runner: [ 'bin/infection' ]
         include:
-          - { operating-system: 'windows-latest', php-version: '8.0', coverage-driver: 'xdebug', e2e-runner: 'bin/infection' }
-          - { operating-system: 'ubuntu-latest', php-version: '8.0', coverage-driver: 'pcov', e2e-runner: 'build/infection.phar' }
-          - { operating-system: 'ubuntu-latest', php-version: '8.0', coverage-driver: 'xdebug', e2e-runner: 'build/infection.phar' }
```
maks-rafalko, line 26: Looks like we need to add to the matrix PHP 8.1 with `xdebug` and `build/infection.phar`?

maks-rafalko (commented):
Could you please create a PR on our docs to add this new requirement?

See https://github.com/infection/site/blob/master/src/guide/installation.md

.github/workflows/ci.yaml
theofidry: Fixed

theofidry:
Done: https://github.com/infection/site/pull/250

maks-rafalko (approved):
Looks good (except the failing build)

.github/workflows/ci.yaml
maks-rafalko: I guess to save time, as we didn't want to check all the combinations

src/FileSystem/DummySymfony5FileSystem.php
```diff
@@ -0,0 +1,121 @@
...
+{
+    public function copy(string $originFile, string $targetFile, bool $overwriteNewerFiles = false): void
+    {
+    }
+
+    public function mkdir($dirs, int $mode = 0777): void
+    {
+    }
+
+    public function exists($files): void
```
sidz, line 54: `exists` doesn't return void in symfony/filesystem 5.4. https://github.com/symfony/filesystem/blob/v5.4.13/Filesystem.php#L107
Return type must be boolean bool.

maks-rafalko:
@theofidry what is now blocking this PR?

theofidry:
@maks-rafalko CI not green and a lot of changes. One thing that sprout from this was that it was running Symfony6 I think, which was not working although we declared Symfony6 compatibility.

So to move forward, I think https://github.com/infection/infection/pull/1824 is required, then I'll be able to rebase this and see what's left

Makefile
```diff
@@ -146,11 +145,8 @@ test-unit-parallel: $(PARATEST) vendor
 
 .PHONY: test-unit-docker
 test-unit-docker:	## Runs the unit tests on the different Docker platforms
-test-unit-docker: test-unit-80-docker
-
-.PHONY: test-unit-80-docker
-test-unit-80-docker: $(DOCKER_FILE_IMAGE) $(PHPUNIT)
-	$(DOCKER_RUN_80) $(PHPUNIT) --group $(PHPUNIT_GROUP)
+test-unit-docker:
+	@echo "No test currently registered"
```
maks-rafalko, line 149: didn't get this, what does it mean? Can't we run just tests with docker for PHP 8.1?

maks-rafalko: looks like you missed this comment :)

theofidry: I was thinking we were doing this for older versions only since I saw there was already PHP 8.1 related constants, but no docker tests for 8.1.

If that's the not the case then it makes sense to restore it and add PHP 8.2 as well

maks-rafalko: reverted https://github.com/infection/infection/pull/1839

PR #1766, opened 2022-12-02
Upgrade to Box 4.5.1

theofidry:
This PR:

- Upgrade to Box 4.5.1
- Reverts #1623 (necessary to upgrade to this Box version)


theofidry:
Quick update: I made some progress, I think the problem is with PHPStan or to be more accurate it's extension installer.

I want to capture this on Box side and once it's done I'll update this.

I however don't think I'll be able to finish this before the end of the month.

theofidry:
I finally got a problematic minimal reproducer working and I now have an idea on how to fix it. So hopefully we'll be able to easily upgrade to 4.3.0 :)

theofidry:
So the issue right now is that we're trying to build the PHAR on the same systems are we're trying to test, which doesn't work because some are on PHP 8.0 and Box does not work on them.

I also think it's quite computationally wasteful to have to rebuild the PHAR all the time; would you be happy with having a pre-build that compiles the (test) PHAR and then we download it to execute the various e2e tests that are using that PHAR?

maks-rafalko:
> I also think it's quite computationally wasteful to have to rebuild the PHAR all the time; would you be happy with having a pre-build that compiles the (test) PHAR and then we download it to execute the various e2e tests that are using that PHAR?

sounds good to me 👍

scoper.inc.php
```diff
@@ -33,58 +33,10 @@
...
-$polyfillsStubs = \array_map(
-    static fn (SplFileInfo $fileInfo) => $fileInfo->getPathname(),
-    \iterator_to_array(
-        Finder::create()
-            ->files()
-            ->in(__DIR__ . '/vendor/symfony/polyfill-*/Resources/stubs')
-            ->name('*.php'),
-        false,
-    ),
-);
```
maks-rafalko, line 59: are polyfills automatically handled in the new Box? As we had a bug with them recently, unfortunately without a reproducer tests as we need to build PHP without some extensions (e.g. without `ext-intl`)

theofidry: It should now be handled yes; I'll double check however

maks-rafalko (approved):

Makefile
```diff
@@ -221,10 +221,12 @@ $(PSALM): Makefile
 
 $(INFECTION): vendor $(shell find bin/ src/ -type f) $(BOX) box.json.dist .git/HEAD
 	composer require infection/codeception-adapter infection/phpspec-adapter
+	composer install --no-dev
```
maks-rafalko, line 224: out of curiosity: I thought it was done automatically by Box. Isn't it the case anymore?

theofidry: It is, unfortunately https://github.com/box-project/box/issues/580 is still a problem; So even if Box removes the dev deps, at the moment, the dev plugins are still enabled hence Composer will try to load them (which will fail since the files are no longer here)

theofidry: Will add a note about this

PR #1767, opened 2022-12-04
Improve CPU core count detection: integrate FidryCpuCoreCounter

theofidry:
See https://github.com/theofidry/cpu-core-counter/issues/11 for the motivations.

The code should be sensibly the same, the notable differences are:

- fixed the deprecated usage of `hw.ncpu`
- add more ways to detect the CPU core counts
- better handling of failures


maks-rafalko (approved):
Fine with it, makes sense to reuse the code rather than copy

sanmai (approved):

PR #1772, opened 2022-12-11
Remove Symfony6 support

theofidry:
I'm having too much of a hard time in #1765. The root of the issue is:

- Symfony6 was introduced in #1606 but without tests
- Symfony6 was actually never executed ever in the Infection CI, so any issue raised was missed
- There is signature incompatibilities between Symfony5 and Symfony6, which is a problem for code such as `DummyFileSystem` (fatal error kind of problem – so any autoloading of the file is breaking things hard)

I think we'll have to either review a few usages, assuming we can make it work, or drop Symfony5 support in the next version.

Meanwhile however I think it's best to (temporarily) disallow Symfony6 to allow to bump the minimum required PHP version to 8.1.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,10 +60,10 @@
         "sanmai/later": "^0.1.1",
         "sanmai/pipeline": "^5.1 || ^6",
         "sebastian/diff": "^3.0.2 || ^4.0",
-        "symfony/console": "^5.4 || ^6.0",
-        "symfony/filesystem": "^5.4 || ^6.0",
-        "symfony/finder": "^5.4 || ^6.0",
-        "symfony/process": "^5.4 || ^6.0",
+        "symfony/console": "^5.4",
+        "symfony/filesystem": "^5.4",
+        "symfony/finder": "^5.4",
+        "symfony/process": "^5.4",
         "thecodingmachine/safe": "^2.1.2",
         "webmozart/assert": "^1.3"
     },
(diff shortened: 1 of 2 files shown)


sanmai:
What's the cost of dropping Symfony 5?

theofidry:
5.4 is an LTS so it will be around for a while, hence I'm not too happy about dropping it. But I also don't really see a way at the moment.

That said, there is this PR and the 8.1 upgrade to do first, maybe someone can find a way to make it work once we are done with that.

sidz:
I don't see a reason to drop symfony 6 support. Instead we need to run testsuites against symfony 6 and fix all possible cases. 
I suppose we won't release new Infection version without symfony 6 and than reintroduce it.

theofidry:
>I don't see a reason to drop symfony 6 support. Instead we need to run testsuites against symfony 6 and fix all possible cases.

Because I can't make it work 😅 

You can check https://github.com/infection/infection/pull/1765 which attempted to do that. There's been several back & forth. So either someone else give it a stab and succeeds, or we accept this PR and move incrementally in another direction, or this gets stuck for a while. What is clear is that bumping to 8.1 as the min-requirement cannot be done in one go due to a lot of broken jobs in the CI.

theofidry:
Just to clarify in case it is misunderstood: when I talk of removing and adding back for Symfony6 I mean for the next release, not in a distant future.

sidz:
Could you please check my comment in https://github.com/infection/infection/pull/1765#discussion_r1046244584

theofidry:
@sidz I did fix it; might need to double check though

theofidry:
Closing this as we have both #1765 and #1775 attempting to fix it directly, and #1786 drafting another way that would make it easier to bump PHP without bumping Symfony, i.e. allowing #1765 to carry one without the Symfony6 issue being a blocker.

PR #1775, opened 2022-12-13
ci: Run CI against Symfony 5 and 6

sidz:
This PR improves our pipeline and running CI with Symfony 5 and 6.

### DIFF
--- a/.github/workflows/autoreview.yaml
+++ b/.github/workflows/autoreview.yaml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        php-version: ['8.0']
+        php-version: ['8.1']
 
     name: Autoreview on PHP ${{ matrix.php-version }}
 
--- a/.github/workflows/autoreview.yaml
+++ b/.github/workflows/autoreview.yaml
@@ -16,7 +16,8 @@ jobs:
 
     strategy:
       matrix:
-        php-version: ['8.1']
+        # Should use the latest supported PHP version
+        php-version: ['8.2']
 
     name: Autoreview on PHP ${{ matrix.php-version }}
 
@@ -28,7 +29,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php-version }}
-          tools: composer:v2.1
+          tools: composer
         env:
           # This is necessary when installing a tool with a specific version
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -51,7 +52,7 @@ jobs:
         run: |
           composer install --no-interaction --prefer-dist --no-progress
 
-      - name: Run tests
-        run: |
-          make autoreview -k
+      - name: Run auto-review tests
+        # Ensures we see all the errors
+        run: make autoreview --keep-going
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,27 +17,32 @@ jobs:
     strategy:
       matrix:
         operating-system: [ ubuntu-latest ]
-        php-version: [ '8.0' ]
+        php-version: [ '8.0', '8.1', '8.2' ]
         coverage-driver: [ pcov, xdebug ]
-        e2e-runner: [ 'bin/infection' ]
+        e2e-runner: [ 'bin/infection', 'build/infection.phar' ]
+        dependencies:
+          - "locked"
+          - "lowest"
+          - "highest"
+        exclude:
+          - php-version: 8.0
+            dependencies: 'highest'
+          - php-version: 8.1
+            coverage-driver: xdebug
+          - php-version: 8.2
+            coverage-driver: xdebug
+          - e2e-runner: 'build/infection.phar'
+            dependencies: 'lowest'
+          - e2e-runner: 'build/infection.phar'
+            dependencies: 'highest'
         include:
-          - { operating-system: 'windows-latest', php-version: '8.0', coverage-driver: 'xdebug', e2e-runner: 'bin/infection' }
-          - { operating-system: 'ubuntu-latest', php-version: '8.0', coverage-driver: 'pcov', e2e-runner: 'build/infection.phar' }
-          - { operating-system: 'ubuntu-latest', php-version: '8.0', coverage-driver: 'xdebug', e2e-runner: 'build/infection.phar' }
-          - { operating-system: 'ubuntu-latest', php-version: '8.1', coverage-driver: 'pcov', e2e-runner: 'bin/infection' }
-          - { operating-system: 'ubuntu-latest', php-version: '8.1', coverage-driver: 'pcov', e2e-runner: 'build/infection.phar' }
-          - { operating-system: 'ubuntu-latest', php-version: '8.2', coverage-driver: 'xdebug', e2e-runner: 'bin/infection' }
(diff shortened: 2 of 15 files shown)


.github/workflows/autoreview.yaml
```diff
@@ -22,20 +22,20 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
```
theofidry, line 25: at the very least those changes look safe enough to be extracted to reduce the noise?

sidz: yeah. Going to revert them

.github/workflows/ci.yaml
```diff
@@ -55,23 +65,32 @@ jobs:
...
             composer-
 
-      - name: Install dependencies
+      - name: Install locked dependencies
+        if: ${{ matrix.dependencies == 'locked' }}
         run: |
-          composer install --no-interaction --prefer-dist --no-progress
+          composer install --no-interaction --prefer-dist --no-progress --optimize-autoloader
+
+      - name: Install lowest dependencies
```
theofidry, line 80: I think we could migrate to the ramsey action which offers the flexibility to set lowest/highest, to do the update or install and also offers caching. This is nothing that we can't do, but that would remove some boiler-template and maybe make this PR a bit easier

.github/workflows/mt.yaml
```diff
@@ -53,7 +53,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          composer update --no-interaction --prefer-dist --no-progress
+          composer install --no-interaction --prefer-dist --no-progress
```
sidz, line 56: I don't remember why do we use "highest" dependencies instead of locked. So lets use locked

.github/workflows/ci.yaml
```diff
@@ -17,28 +17,33 @@ jobs:
...
-        php-version: [ '8.0' ]
+        php-version: [ '8.0', '8.1', '8.2' ]
         coverage-driver: [ pcov, xdebug ]
-        e2e-runner: [ 'bin/infection' ]
+        e2e-runner: [ 'bin/infection', 'build/infection.phar' ]
+        dependencies:
+          - "locked"
+          - "lowest"
+          - "highest"
+        exclude:
```
sidz, line 27: we probably may exclude more cases. Give me nudge if we need to add something else

.github/workflows/ci.yaml
```diff
@@ -55,23 +60,32 @@ jobs:
...
+        run: |
+          composer update --no-interaction --prefer-dist --no-progress --optimize-autoloader
 
       - name: Run tests
         shell: bash
         run: |
-          if [[ "${{ matrix.php-version }}" == '8.2' ]]; then
-            export SYMFONY_DEPRECATIONS_HELPER=max[self]=0
-          fi
+          export SYMFONY_DEPRECATIONS_HELPER=max[self]=0
```
sidz, line 88: We will need to remove this hack and fix all direct deprecations. For instance `strlen`

.github/workflows/mt.yaml
theofidry: probably because then we never test infection installed as a package, since it will only picked the locked with the configured platform

.github/workflows/ci.yaml
theofidry, line 88: Maybe we can sort it out in a separate PR, but I wonder if we really want to enable the deprecations when not running the locked dependencies

psalm.xml
```diff
@@ -4,7 +4,7 @@
     resolveFromConfigFile="true"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
-    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
+    xsi:schemaLocation="https://getpsalm.org/schema/config.xsd vendor/vimeo/psalm/config.xsd"
```
theofidry, line 7:
```suggestion
    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
```

sidz:
Currently I'm waiting when @maks-rafalko adds GITHUB_TOKEN secret in the repository settings to avoid rate limit exception

theofidry:
@sidz actually the `GITHUB_TOKEN` secret should already be configured: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret

sidz:
thanks @theofidry for pointing to the documentation. I assumed that secret needs to be added manually

theofidry:
@sidz I went ahead with my refactors. Just to me clear I'm avoiding adding the Symfony6 support to not overlap with your work. If this hinders this PR by resulting in conflicts I'll do a PR on yours to resolve them.

.github/workflows/mt.yaml
theofidry: Sorted out in #1791

composer.json
```diff
@@ -88,9 +88,6 @@
         "thecodingmachine/phpstan-safe-rule": "^1.2.0"
     },
     "config": {
-        "platform": {
```
theofidry, line 91: I don't think removing the platform completely is the way, since we do want it. More likely we want to unset it when testing non-locked dependencies like done in https://github.com/infection/infection/pull/1786/files#diff-6e608e02c595d53ab6b70822a2bf19abcfc6ddcc976c2f536ad5bfca20f0443fR51-R53

tests/phpunit/Fixtures/Finder/MockRealPathFinder.php
```diff
@@ -1,21 +1,20 @@
 <?php
 
+declare(strict_types=1);
```
theofidry, line 3: Looks like we could extract those changes in a different PR?

tests/phpunit/Fixtures/Finder/MockRelativePathFinder.php
```diff
@@ -1,21 +1,20 @@
 <?php
 
+declare(strict_types=1);
```
theofidry, line 3: same for this one

.github/workflows/ci.yaml
theofidry: For me it's quite hard to tell at this point, because I don't know what is changed here because we were not testing a wide enough set, or because of the Symfony6 compatibility.

Don't you think it would be best to go towards #1786 first, which allows to test a wider range whilst keeping the Symfony6 compatibility issue out, and then add Symfony6 to fix the compatibility?

If the conflicts are a concern I'm more than happy to take care of them

theofidry (commented):
👍 overall. I think there is a bit too much mixed in there for my taste to distinguish what is due to Symfony6 and what is due to test a wider range of combinations.

But if you think it's counter-productive to go with my suggestion I can do my rework later.

theofidry:
@maks-rafalko to me the failure looks legit, but I don't think it's anything new. It just looks like we never ran the unit tests on Windows.

Is it something we care about?
yes -> I would disable it in this PR, create another PR where I enable it and try to fix it separately; since this is not new
no -> we remove it and go on

theofidry:
Notes for myself for when I come back to this:

- Make easier to adapt the protection rules (#1806)
- Split the CI to make it easier (#1787)
- Prepare the CI to test against different dependency versions (#1815)
- Prepare the CI to test against different Symfony versions (#1817)
- Update this PR to add Symfony6 support
- Fixing the incomplete builds:
  - add the missing bits in the range (test against lowest & highest for example)
  - fix the broken builds, e.g. #1805

PR #1781, opened 2022-12-19
refactor(ci): Migrate to ramsey/composer-install

theofidry:
See [ramsey/composer-install](https://github.com/ramsey/composer-install).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -47,24 +47,10 @@ jobs:
           ini-values: memory_limit=512M, xdebug.mode=off
           tools: composer:v2.1
 
-      - name: Get composer cache directory
-        id: composer-cache
-        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
-
-      - name: Cache dependencies
-        uses: actions/cache@v2
-        with:
-          path: ${{ steps.composer-cache.outputs.dir }}
-          key: composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('composer.*') }}
-          restore-keys: |
-            composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('composer.*') }}-
-            composer-${{ runner.os }}-${{ matrix.php-version }}-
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -47,24 +47,10 @@ jobs:
...
-
       - name: Install dependencies
-        run: |
-          composer install --no-interaction --prefer-dist --no-progress
+        uses: ramsey/composer-install@v2
+        with:
+            dependency-versions: locked
 
       - name: Run tests
         shell: bash
```
theofidry, line 56: Didn't change it for bellow: I admit I'm really not comfortable with it yet (the tests executed)

sidz:
@theofidry  To be honest I don't see any benefit of using _ramsey/composer-install_. As for me `composer install/update` looks better and predictable.

Also I'd like to merge https://github.com/infection/infection/pull/1775 before any of your CI improvements PR just to avoid doing same job twice.

theofidry:
@sidz if you don't see the benefits with that diff I don't think I can convince you otherwise. To me it's clearly more readable and not having to manually do the caching which also implies manually handling the different parameters of the cache is a pretty big win

theofidry:
>Also I'd like to merge https://github.com/infection/infection/pull/1775 before any of your CI improvements PR just to avoid doing same job twice.

I was hoping the opposite actually. If you're comfortable with #1775 then fine, but I personally found quite hard to untangle the whole thing. After the refactorings adding Symfony6 to the CI should only be a matter of adding it [here](https://github.com/infection/infection/pull/1786/files#diff-6e608e02c595d53ab6b70822a2bf19abcfc6ddcc976c2f536ad5bfca20f0443fR24) and it would be a build where it's clear what you are testing

theofidry:
Another bonus, the output is nicer https://github.com/infection/infection/pull/1781/checks (see https://github.com/infection/infection/pull/1784/checks) for the current.

Of course can be achieved as well without, but it is yet another thing to worry about.

PR #1787, opened 2022-12-19
ci(refactor): Extract unit & integration tests from the CI workflow

theofidry:
I think the end-to-end workflow is quite complex and hard to work with. There is several concerns brought by the nature of end-to-end tests and they also tend to be inherently slow. This makes it hard to further tweak the configuration and test a wider range of constraints.

I'm proposing to extract the unit & integration tests. This will allow to diversify the build to test:

- A wider range of operating systems
- More different PHP versions
- Highest/lowest or locked versions
- Specific Symfony versions

A PoC of that work can be seen in #1786.

⚠️ Requires to update the project settings afterwards as the build name changes

### DIFF
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -1,0 +1,84 @@
+# yamllint disable rule:line-length
+# yamllint disable rule:braces
+
+name: Unit & Integration Tests
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - master
+
+jobs:
+  tests:
+    runs-on: ${{ matrix.operating-system }}
+
+    strategy:
+      matrix:
+        operating-system: [ ubuntu-latest ]
+        php-version: [ '8.0' ]
+        coverage-driver: [ pcov, xdebug ]
+        include:
+          - { operating-system: 'windows-latest', php-version: '8.0', coverage-driver: 'xdebug' }
+          - { operating-system: 'ubuntu-latest', php-version: '8.0', coverage-driver: 'pcov' }
+          - { operating-system: 'ubuntu-latest', php-version: '8.0', coverage-driver: 'xdebug' }
+          - { operating-system: 'ubuntu-latest', php-version: '8.1', coverage-driver: 'pcov' }
+          - { operating-system: 'ubuntu-latest', php-version: '8.1', coverage-driver: 'pcov' }
+          - { operating-system: 'ubuntu-latest', php-version: '8.2', coverage-driver: 'xdebug' }
+
+    name: CI on ${{ matrix.operating-system }} with PHP ${{ matrix.php-version }}, using ${{ matrix.coverage-driver }}
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Configure for PHP >= 8.2
+        if: "matrix.php-version >= '8.2'"
+        run: |
+          composer config platform.php 8.1.99
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-version }}
+          coverage: ${{ matrix.coverage-driver }}
+          ini-values: memory_limit=512M, xdebug.mode=off
+          tools: composer
+
+      - name: Get composer cache directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+      - name: Cache dependencies
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('composer.*') }}
+          restore-keys: |
+            composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('composer.*') }}-
+            composer-${{ runner.os }}-${{ matrix.php-version }}-
+            composer-${{ runner.os }}-
+            composer-
+
+      - name: Install dependencies
+        run: |
+          composer install --no-interaction --prefer-dist --no-progress
+
+      - name: Run tests
+        shell: bash
+        run: |
+          if [[ "${{ matrix.php-version }}" == '8.2' ]]; then
+            export SYMFONY_DEPRECATIONS_HELPER=max[self]=0
+          fi
+
+          make test-unit
+
+      - name: Run integration tests
+        shell: bash
+        run: |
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -137,7 +137,7 @@ test-autoreview: $(PHPUNIT) vendor
 .PHONY: test-unit
 test-unit:	 	## Runs the unit tests
 test-unit: $(PHPUNIT) vendor
-	$(PHPUNIT) --group $(PHPUNIT_GROUP)
+	$(PHPUNIT) --group $(PHPUNIT_GROUP) --exclude-group e2e
```
maks-rafalko, line 140: where do these unit tests with group `e2e` run now?

theofidry: I believe we have `make test-e2e E2E_PHPUNIT_GROUP=e2e` in `ci.yaml`:

```
      - name: Run the whole set of E2E tests
        if: runner.os != 'Windows' && matrix.e2e-runner == 'bin/infection'
        env:
          TERM: xterm-256color
        run: |
          if [[ "${{ matrix.php-version }}" == '8.2' ]]; then
            export SYMFONY_DEPRECATIONS_HELPER=max[self]=0
          fi

          make test-e2e E2E_PHPUNIT_GROUP=e2e INFECTION=bin/infection
```

theofidry:
@sidz any chance you can review this?

sidz (approved):

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,84 @@
...
+      matrix:
+        operating-system: [ ubuntu-latest ]
+        php-version: [ '8.0' ]
+        coverage-driver: [ pcov, xdebug ]
+        include:
+          - { operating-system: 'windows-latest', php-version: '8.0', coverage-driver: 'xdebug' }
+          - { operating-system: 'ubuntu-latest', php-version: '8.0', coverage-driver: 'pcov' }
+          - { operating-system: 'ubuntu-latest', php-version: '8.0', coverage-driver: 'xdebug' }
+          - { operating-system: 'ubuntu-latest', php-version: '8.1', coverage-driver: 'pcov' }
+          - { operating-system: 'ubuntu-latest', php-version: '8.1', coverage-driver: 'pcov' }
```
maks-rafalko, line 27:
```suggestion
          - { operating-system: 'ubuntu-latest', php-version: '8.1', coverage-driver: 'xdebug' }
```

because this is a duplicate of line 26

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,84 @@
...
+          - { operating-system: 'ubuntu-latest', php-version: '8.2', coverage-driver: 'xdebug' }
```
maks-rafalko, line 28: what about `pcov` for 8.2? any reason not to run it?

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,84 @@
+# yamllint disable rule:line-length
+# yamllint disable rule:braces
+
+name: Unit & Integration Tests
```
maks-rafalko, line 4: the name of the file is `unit-tests.yaml`, but here we are saying this is also for Integration tests. Let's rename the file to `tests.yaml`? Does it make sense? (I'm not sure as we still have e2e tests in `ci.yaml`)

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,84 @@
...
+    name: CI on ${{ matrix.operating-system }} with PHP ${{ matrix.php-version }}, using ${{ matrix.coverage-driver }}
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Configure for PHP >= 8.2
+        if: "matrix.php-version >= '8.2'"
+        run: |
+          composer config platform.php 8.1.99
```
maks-rafalko, line 39: is it still needed? cc @sidz

.github/workflows/unit-tests.yaml
```diff
@@ -0,0 +1,84 @@
...
+            composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('composer.*') }}-
+            composer-${{ runner.os }}-${{ matrix.php-version }}-
+            composer-${{ runner.os }}-
+            composer-
+
+      - name: Install dependencies
+        run: |
+          composer install --no-interaction --prefer-dist --no-progress
+
+      - name: Run tests
```
maks-rafalko, line 68:
```suggestion
      - name: Run unit tests
```

maks-rafalko (changes requested):

.github/workflows/unit-tests.yaml
theofidry: `tests.yaml` sounds good to me.

.github/workflows/unit-tests.yaml
theofidry: indeed; actually updated the matrix this should do this

.github/workflows/unit-tests.yaml
theofidry: I think we could rework this to use `--ignore-platform-req=php+` instead, but I would prefer to do this later. I've tried to re-use as much as possible `ci.yaml` and rework the different steps later

theofidry:
@maks-rafalko to me the failure looks legit, but I don't think it's anything new. It just looks like we never ran the unit tests on Windows.

Is it something we care about?
yes -> I would disable it in this PR, create another PR where I enable it and try to fix it separately; since this is not new
no -> we remove it and go on

.github/workflows/unit-tests.yaml
sidz: @maks-rafalko we don't need it anymore as php 8.2 has been released.

also this section has been removed in https://github.com/infection/infection/pull/1775/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddL37-L41

maks-rafalko:
> It just looks like we never ran the unit tests on Windows.

I think we did 

https://github.com/infection/infection/blob/55902e06057ad0b5f6884eca1669601287421f20/.github/workflows/ci.yaml#L24

and below

https://github.com/infection/infection/blob/55902e06057ad0b5f6884eca1669601287421f20/.github/workflows/ci.yaml#L79

I would for for yes - we need to run unit tests on windows as well. Here we disabled Infection on Infection on windows (https://github.com/infection/infection/pull/1611), but unit tests IMO should be executed there.

theofidry:
@maks-rafalko Merging this then. I'll open a PR to enable back the tests on Windows, since I'm confident that this PR does not actually break any code.

PR #1782, opened 2022-12-19
refactor(ci): Execute Auto-Review tests on PHP 8.2

theofidry:
- Execute the Auto-Review tests on PHP 8.2 instead of 8.0: I think it should be executed on the latest supported version and will help to incrementally go towards #1765.
- Remove the locked Composer version: At the very least for an auto-review test I do not think the Composer version should matter.
- Use the extended option of `-k` for better readabilty

⚠️  Requires to update the project settings afterwards as the build name changes
⚠️  Not a big deal, but if merged before/after #1789 to mind the `env` setting for the php setup step which is no longer necessary (see this change https://github.com/infection/infection/pull/1789/files#diff-87ab0e37e52f2d5a98de1c719e9222197e1f875c57ed290a6b10f0c640e5b001R32)

PR #1788, opened 2022-12-21
Add `allow-plugins` config for for `pestphp/pest-plugin` to avoid failing builds

maks-rafalko:
Should fix build failures like this https://github.com/infection/infection/actions/runs/3734696981/jobs/6337055910#step:10:2511

Same has already been added to another e2e test:

https://github.com/infection/infection/blob/b6ae298e9a2000d3af2eadd7343c5776db6a1c24/tests/e2e/PestTestFramework/composer.json#L15-L19

Should fix https://github.com/infection/infection/pull/1782

### Why it fails randomly?

Seems like the reason why it fails randomly is when `setup php` action fails to setup composer **2.1**, and (I assume) composer **2.2** is installed by default, resulting in failures because of absense of `allowed-plugins`

See https://github.com/infection/infection/actions/runs/3734759922/jobs/6337196890#step:4:18 (fails) and https://github.com/infection/infection/actions/runs/3751882494/jobs/6373403435#step:4:15 (passes)

### DIFF
--- a/tests/e2e/SymfonyFlex/composer.json
+++ b/tests/e2e/SymfonyFlex/composer.json
@@ -13,5 +13,10 @@
         "psr-4": {
             "SymfonyFlex\\Test\\": "tests/"
         }
+    },
+    "config": {
+        "allow-plugins": {
+            "symfony/flex": true
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


maks-rafalko:
Merging as it's easy fix and will unblock other PRs

maks-rafalko:
the root cause (rate limit error) will be additionally fixed by https://github.com/infection/infection/pull/1775/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR55

sidz:
yep. You're right. our CI uses composer v2.1 but it looks like in case of rate limit it setup-php took latest stable

PR #1791, opened 2022-12-21
refactor(ci): Review the mutation testing workflow

theofidry:
My understanding of that build was, from what I understand in #1365, to have a dedicated pipeline mimicking what a typical user pipeline may look like and testing our mutation badge feature. Over time this was slightly tweaked:

- The test for windows was removed because overly slow (#1611)
- Remove testing against different dependencies (#1535) – Likely due to instability in the build

I think the build is still useful, but we now have a more comprehensive and diversified test suite to test the different OS, dependencies, code coverage reports and co in `ci.yaml`. Hence this build can be more scoped and from that simplified.

- Document the above in the build
- Bump the PHP version to 8.2
- Use the locked version of the dependencies
- Use the latest Composer version available

Tried to switch to Xdebug. I was curious if there is a big time difference but in general I also know Xdebug is more accurate than pcov, so we may prefer this to enforce a specific MSI score.

Comparing with [this build](https://github.com/infection/infection/actions/runs/3734647686/jobs/6336945755), we had:

|  | pcov | Xdebug | delta |
|--|--|--| --|
| time taken (unit tests) | 1m7 | 1m34|+40%|
| time taken (infection) | 8m7 | 13m41 | +70%|
| MSI |73%| | 75% | +2% |

At least from those results it doesn't look worth it.

⚠️ Requires to update the project settings afterwards as the build name changes

### DIFF
--- a/.github/workflows/mt.yaml
+++ b/.github/workflows/mt.yaml
@@ -62,7 +62,7 @@ jobs:
 
       - name: Collect coverage report
         run: |
-            SYMFONY_DEPRECATIONS_HELPER=disabled php vendor/phpunit/phpunit/phpunit --stop-on-failure \
+          SYMFONY_DEPRECATIONS_HELPER=disabled php vendor/phpunit/phpunit/phpunit --stop-on-failure \
             --coverage-xml=build/logs/coverage-xml \
             --log-junit=build/logs/junit.xml
 
--- a/.github/workflows/mt.yaml
+++ b/.github/workflows/mt.yaml
@@ -63,7 +63,7 @@ jobs:
 
       - name: Collect coverage report
         run: |
-          php -dxdebug.mode=coverage vendor/phpunit/phpunit/phpunit --stop-on-failure \
+            XDEBUG_MODE=coverage php vendor/phpunit/phpunit/phpunit --stop-on-failure \
             --coverage-xml=build/logs/coverage-xml \
             --log-junit=build/logs/junit.xml
 
--- a/.github/workflows/mt.yaml
+++ b/.github/workflows/mt.yaml
@@ -38,8 +38,7 @@ jobs:
         with:
           # Any supported PHP version is fine
           php-version: '8.2'
-          coverage: xdebug
-          ini-values: xdebug.mode=off
+          coverage: pcov
           tools: composer
 
       - name: Get composer cache directory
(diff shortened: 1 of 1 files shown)


.github/workflows/mt.yaml
```diff
@@ -1,7 +1,16 @@
...
+# This build is to test the whole execute the PHPUnit tests & collect the coverage
+# report and run infection.
+# We are not interested in:
+# - Running it on different systems (e.g. Windows, which was disabled because too slow)
+# - Min/maxing the PHP version used: it can be any supported version
+# - Using different dependencies than the locked ones
+# - Using different code coverages
+# All of those variants may be interesting but are better tested in more scoped
+# tests like in ci.yaml.
+name: Infection on Infection
```
maks-rafalko, line 13: Let's revert to `Mutation Testing` as now the full name is a bit strange:

`Infection on Infection / Infection complete run on Infection`

After reverting, will be:

`Mutation Testing / Infection complete run on Infection`

.github/workflows/mt.yaml
```diff
@@ -52,12 +56,13 @@ jobs:
...
-        run: |
-          composer update --no-interaction --prefer-dist --no-progress
+        # Run the locked dependencies: we are not interested in testing different
+        # variations here so having a stable set is better.
+        run: composer install --no-interaction --prefer-dist --no-progress
 
       - name: Collect coverage report
         run: |
-          php vendor/phpunit/phpunit/phpunit --stop-on-failure \
+          SYMFONY_DEPRECATIONS_HELPER=disabled php vendor/phpunit/phpunit/phpunit --stop-on-failure \
```
maks-rafalko, line 65: why is it needed?

maks-rafalko (approved):

.github/workflows/mt.yaml
theofidry: Because depending on the deprecations it can fail this build. It's not necessarily bad, but we already check for the deprecations in `ci.yaml`. So having it here is redundant, and disabling it guaranteeing we get to the interesting part of that build which is running infection with coverage generated and with the badge API key

theofidry: Lemme know if that's a satisfying answer for you or you prefer to keep it

PR #1794, opened 2022-12-23
Fix coverage failure in case of an invalid file

mabar:
This PR:

- [ ] Adds new feature ...
- [ ] Covered by tests

My lib supports two incompatible versions of a dependency. Code for older version extends class that does not exist in newer one. Currently code coverage fails with `Uncaught Error: Class "Latte\Macros\MacroSet" not found` https://github.com/orisai/localization/actions/runs/3766779916/jobs/6403641877 (shown run uses only temporary commit, I also had to do the same fix in https://github.com/sebastianbergmann/php-code-coverage/pull/973)

After applying the fix, coverage no longer fails. And file is correctly reported as uncovered.
https://github.com/orisai/localization/commit/a0f7a0b766c738d0df99eed52f8f529aceabb4f3
https://github.com/orisai/localization/actions/runs/3766885353

### DIFF
--- a/src/PhpParser/Visitor/ReflectionVisitor.php
+++ b/src/PhpParser/Visitor/ReflectionVisitor.php
@@ -37,6 +37,7 @@ namespace Infection\PhpParser\Visitor;
 
 use function array_pop;
 use function count;
+use Error;
 use Infection\Reflection\AnonymousClassReflection;
 use Infection\Reflection\ClassReflection;
 use Infection\Reflection\CoreClassReflection;
@@ -184,7 +185,11 @@ final class ReflectionVisitor extends NodeVisitorAbstract
         $fqn = FullyQualifiedClassNameManipulator::getFqcn($node);
 
         if ($fqn !== null) {
-            return CoreClassReflection::fromClassName($fqn->toString());
+            try {
+                return CoreClassReflection::fromClassName($fqn->toString());
+            } catch (Error $e) {
+                return new NullReflection();
+            }
         }
 
         // TODO: check against interfaces


mabar:
@maks-rafalko @theofidry Could you check that please? Thank you

maks-rafalko:
Hello, sorry for the late reply. To be honest, I'm not sure I completely get the issue.

Could you please point to the code where "lib supports two incompatible versions of dependency?" I would like o clearly understand what is going on and why php-coverage/infection work with class that don't exist, as well as why it's needed in the lib (probably you should have to create a new major version instead of doing something like that).

Thank you.

mabar:
> Could you please point to the code where "lib supports two incompatible versions of dependency?"

Sure! 
There was a breaking change in [latte/latte](https://github.com/nette/latte) package.
In v2 template extensions were created by extending MacroSet class https://github.com/orisai/localization/blob/7cf546c92320b7b8fffbff86895a65c162bfe616/src/Bridge/Latte/TranslationMacros.php#L13
In v3, MacroSet is removed and instead, Extension is extended https://github.com/orisai/localization/blob/7cf546c92320b7b8fffbff86895a65c162bfe616/src/Bridge/Latte/TranslationExtension.php#L21
When I run Infection, Latte 3 is installed and MacroSet class does not exist. Due to extending non-existent class, Infection breaks.

To verify my patch, run:

```
git clone git@github.com:orisai/localization.git
cd localization
composer update
make mutations
```

I have removed the patch in another branch to demonstrate failure with latest Infection:

```
git checkout drop-infection-patch
composer update
make mutations
```

> probably you should have to create a new major version instead of doing something like that

I would usually agree, but with tens of dependencies, it makes upgrades so painful. Also maintenance, I don't really want to maintain more major releases when I don't have to :)

Issue #1797, opened 2023-01-07
ArrayItemRemoval mutator should not mutate an array when set as an attributes argument


lsv:
| Question    | Answer
| ------------| ---------------
| Versions | as as https://infection-php.dev/r/7283

https://infection-php.dev/r/7283

How can the mutant be uncovered?

sidz:
Good catch @lsv . Fixed in https://github.com/infection/infection/pull/1798

Thanks

PR #1798, opened 2023-01-10
Do not remove array item when it is a part of attribute argument

sidz:
This PR fixes https://github.com/infection/infection/issues/1797

ArrayItemRemoval mutator should be skipped when array is an attributes argument


maks-rafalko (approved):
👍 

thank you, @sidz

Issue #1796, opened 2023-01-06
`LogicalOr` to `&&` must not be applied to same variable test


Neirda24:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.26.16
| Test Framework version | PHPUnit 9.5.27
| PHP version | 8.1.13
| Platform    | MacOS
| Github Repo | https://infection-php.dev/r/yl6e

```php
$numberOfMatches = preg_match_all('something', $string, $matches); 

if (false === $numberOfMatches || 0 === $numberOfMatches) {
    // do something
}
```

cannot be replaced with

```php
$numberOfMatches = preg_match_all('something', $string, $matches); 

if (false === $numberOfMatches && 0 === $numberOfMatches) {
    // do something
}
```

using the `LogicalOr` mutator.

The same variable can't be both `false` and `int` using the `===` strict comparison.

maks-rafalko:
Feel free to submit the PR to fix this issue. The place to start: 

https://github.com/infection/infection/blob/a222e757657f252bce05505ce5910520a9c21fc7/src/Mutator/Boolean/LogicalOr.php#L75

Neirda24:
Sure will try to take a look. Thanks for the heads up.

PR #1801, opened 2023-01-16
Trying to remove false positive on logical or.

Neirda24:
This PR:

- [x] Improve LogicalOr mutatability
- [x] Covered by tests

Fixes https://github.com/infection/infection/issues/1796

It improves the LogicalOr mutatability by detecting if we are trying to mutate to something that doesn't make sense.
For example :

```php
$myVar === 'hello' || $myVar === 'world';
```

cannot be mutated to 

```php
$myVar === 'hello' && $myVar === 'world';
```

I think there a lot more usecase that I don't cover such as

```php
$myVar < 5 || $myVar > 10;
```

can't be mutated but

```php
$myVar < 10 || $myVar > 5;
```

could. So not only should we check the variable name but also to what it compares. Although it would require to handle a lot of cases and Node class doesn't help much at the moment. Help appreciated to improve this further.

---
Edit : also handled some use cases regarding < & >

### DIFF
--- a/src/Mutator/Boolean/LogicalOr.php
+++ b/src/Mutator/Boolean/LogicalOr.php
@@ -35,13 +35,13 @@ declare(strict_types=1);
 
 namespace Infection\Mutator\Boolean;
 
+use function get_class;
+use function in_array;
 use Infection\Mutator\Definition;
 use Infection\Mutator\GetMutatorName;
 use Infection\Mutator\Mutator;
 use Infection\Mutator\MutatorCategory;
 use PhpParser\Node;
-use function get_class;
-use function in_array;
 
 /**
  * @internal
--- a/src/Mutator/Boolean/LogicalOr.php
+++ b/src/Mutator/Boolean/LogicalOr.php
@@ -35,6 +35,8 @@ declare(strict_types=1);
 
 namespace Infection\Mutator\Boolean;
 
+use function get_class;
+use function in_array;
 use Infection\Mutator\Definition;
 use Infection\Mutator\GetMutatorName;
 use Infection\Mutator\Mutator;
@@ -77,6 +79,86 @@ DIFF
 
     public function canMutate(Node $node): bool
     {
-        return $node instanceof Node\Expr\BinaryOp\BooleanOr;
+        if (!$node instanceof Node\Expr\BinaryOp\BooleanOr) {
+            return false;
+        }
+
+        $nodeLeftType = get_class($node->left);
+        $nodeRightType = get_class($node->right);
+
+        $mustCheckForVariableNamesNodeTypes = [
+            Node\Expr\BinaryOp\Identical::class,
+            Node\Expr\BinaryOp\NotIdentical::class,
+            Node\Expr\BinaryOp\Equal::class,
+            Node\Expr\BinaryOp\NotEqual::class,
+            Node\Expr\BinaryOp\Greater::class,
+            Node\Expr\BinaryOp\GreaterOrEqual::class,
+            Node\Expr\BinaryOp\Smaller::class,
+            Node\Expr\BinaryOp\SmallerOrEqual::class,
+        ];
+
+        if (
+            (
+                $node->left instanceof Node\Expr\BinaryOp\Identical
+                && $node->right instanceof Node\Expr\BinaryOp\Identical
+            ) || (
+                $node->left instanceof Node\Expr\BinaryOp\Equal
+                && $node->right instanceof Node\Expr\BinaryOp\Equal
+            )
+        ) {
+            $varName = '';
+
+            if ($node->left->left instanceof Node\Expr\Variable) {
+                $varName = $node->left->left->name;
+            } elseif ($node->left->right instanceof Node\Expr\Variable) {
+                $varName = $node->left->right->name;
+            }
+
+            if ($node->right->left instanceof Node\Expr\Variable) {
+                return $varName !== $node->right->left->name;
+            } elseif ($node->right->right instanceof Node\Expr\Variable) {
+                return $varName !== $node->right->right->name;
+            }
+        }
+
+        $greaterOp = [
+            Node\Expr\BinaryOp\Greater::class,
+            Node\Expr\BinaryOp\GreaterOrEqual::class,
+        ];
+
+        $smallerOp = [
+            Node\Expr\BinaryOp\Smaller::class,
+            Node\Expr\BinaryOp\SmallerOrEqual::class,
+        ];
+
+        if (
+            (
+                in_array(get_class($node->left), $greaterOp, true) === true
+                && in_array(get_class($node->right), $smallerOp, true) === true
+            ) || (
+                in_array(get_class($node->left), $smallerOp, true) === true
+                && in_array(get_class($node->right), $greaterOp, true) === true
+            )
+        ) {
+            $varNameLeft = null;
+
+            if ($node->left->left instanceof Node\Expr\Variable) {
+                $varNameLeft = $node->left->left->name;
+            } elseif ($node->left->right instanceof Node\Expr\Variable) {
+                $varNameLeft = $node->left->right->name;
+            }
+
+            $varNameRight = null;
+
+            if ($node->right->left instanceof Node\Expr\Variable) {
+                $varNameRight = $node->right->left->name;
+            } elseif ($node->right->right instanceof Node\Expr\Variable) {
+                $varNameRight = $node->right->right->name;
+            }
+
+            return $varNameLeft !== $varNameRight;
+        }
+
+        return true;
     }
 }
--- a/src/Mutator/Boolean/LogicalOr.php
+++ b/src/Mutator/Boolean/LogicalOr.php
@@ -83,20 +83,6 @@ DIFF
             return false;
         }
 
-        $nodeLeftType = get_class($node->left);
-        $nodeRightType = get_class($node->right);
-
-        $mustCheckForVariableNamesNodeTypes = [
-            Node\Expr\BinaryOp\Identical::class,
-            Node\Expr\BinaryOp\NotIdentical::class,
-            Node\Expr\BinaryOp\Equal::class,
-            Node\Expr\BinaryOp\NotEqual::class,
-            Node\Expr\BinaryOp\Greater::class,
-            Node\Expr\BinaryOp\GreaterOrEqual::class,
-            Node\Expr\BinaryOp\Smaller::class,
-            Node\Expr\BinaryOp\SmallerOrEqual::class,
-        ];
-
         if (
             (
                 $node->left instanceof Node\Expr\BinaryOp\Identical
--- a/src/Mutator/Boolean/LogicalOr.php
+++ b/src/Mutator/Boolean/LogicalOr.php
@@ -35,6 +35,7 @@ declare(strict_types=1);
 
 namespace Infection\Mutator\Boolean;
 
+use function array_intersect;
 use function get_class;
 use function in_array;
 use Infection\Mutator\Definition;
@@ -42,7 +43,6 @@ use Infection\Mutator\GetMutatorName;
 use Infection\Mutator\Mutator;
 use Infection\Mutator\MutatorCategory;
 use PhpParser\Node;
-use function array_intersect;
 use function property_exists;
 
 /**
@@ -85,7 +85,7 @@ DIFF
             return false;
         }
 
-        $nodeLeft  = $node->left;
+        $nodeLeft = $node->left;
         $nodeRight = $node->right;
 
         if (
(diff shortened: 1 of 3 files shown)


src/Mutator/Boolean/LogicalOr.php
```diff
@@ -77,6 +79,72 @@ public function mutate(Node $node): iterable
...
+                && $node->right instanceof Node\Expr\BinaryOp\Identical
+            ) || (
+                $node->left instanceof Node\Expr\BinaryOp\Equal
+                && $node->right instanceof Node\Expr\BinaryOp\Equal
+            )
+        ) {
+            $varName = '';
+
+            if ($node->left->left instanceof Node\Expr\Variable) {
+                $varName = $node->left->left->name;
```
Neirda24, line 98: for some reasons, phpstan and phpstorm cannot infer the type Variable to `$node->left->left` although it is checked against the instanceof right before...

tests/phpunit/Mutator/Boolean/LogicalOrTest.php
```diff
@@ -66,11 +66,62 @@ public function mutationsProvider(): iterable
...
+$myVar < 5 || $myVar > 10;
+PHP
+            ,
+        ];
+
+        yield 'It does not mutates logical or if same variable is tested against "Greater" and "Smaller" #2.' => [
+            <<<'PHP'
+<?php
+
+$myVar < 10 || $myVar > 5;
```
Neirda24, line 124: This should trigger a mutation because of the value. Not in the current scope of this PR ATM.

src/Mutator/Boolean/LogicalOr.php
```diff
@@ -77,6 +79,72 @@ public function mutate(Node $node): iterable
...
+        ];
+
+        $smallerOp = [
+            Node\Expr\BinaryOp\Smaller::class,
+            Node\Expr\BinaryOp\SmallerOrEqual::class,
+        ];
+
+        if (
+            (
+                in_array(get_class($node->left), $greaterOp, true) === true
```
Neirda24, line 122: Could have applied this logic only once but I figured if we need to compare the values too then the logic might differ

src/Mutator/Boolean/LogicalOr.php
```diff
@@ -77,6 +79,76 @@ public function mutate(Node $node): iterable
...
+            return false;
+        }
+
+        $equalOp = [
+            Node\Expr\BinaryOp\Identical::class,
+            Node\Expr\BinaryOp\Equal::class,
+        ];
+
+        if (
+            in_array(get_class($node->left), $equalOp, true) === true
```
maks-rafalko, line 92: `get_class($node->left)` - does it make sense to move to variable? As it's being called many times. Same for `$node->right`

src/Mutator/Boolean/LogicalOr.php
```diff
@@ -77,6 +79,76 @@ public function mutate(Node $node): iterable
...
+            && in_array(get_class($node->right), $equalOp, true) === true
+        ) {
+            $varNameLeft = null;
+
+            if ($node->left->left instanceof Node\Expr\Variable) {
+                $varNameLeft = $node->left->left->name;
+            } elseif ($node->left->right instanceof Node\Expr\Variable) {
```
maks-rafalko, line 99: could you please explain what source code leads to this checks/AST? I think I don't understand AST structure :)

src/Mutator/Boolean/LogicalOr.php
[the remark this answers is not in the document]
Neirda24: As you wish. I always add a strict comparison because sometimes we want it to be equal to false and the `!` is easy to miss / remove. I'll adapt to your coding style.

Neirda24: Yes we could. When upgrading PHP we could then use `$node->left::class` and maybe we will have to rollback the variable.

src/Mutator/Boolean/LogicalOr.php
Neirda24: 😅 
```php
$myVar === 'hello' || $myVar === 'world';
```

for example (see added tests). Because it must not be mutated to 

```php
$myVar === 'hello' && $myVar === 'world';
```

So I check that on both side of the `||` we compare the same variable name against an equality. but because the variable could be on either the right side of `===` or the left side, I have to check node->left->left, node->left->right + node->right->right, node->left->left  for the variable name.

maks-rafalko: cool, now I fully understand it. Then IMO in tests we need to add a case where on the left side of `||` we have a variable on the right side of `===` (now there is no such case)

Neirda24: yes there is : https://github.com/infection/infection/pull/1801/files#diff-6cb62609cbceaf882b2803a9e26b277dc505d2cae4258e29d23acf9a9d067d37R106

maks-rafalko: I mean on the left side of `||`. So additional test will be

```diff
+ 'hello' === $myVar || 'world' === $myVar;
```

Neirda24: fixed.

src/Mutator/Boolean/LogicalOr.php
maks-rafalko: ok, let's do it

Neirda24: done

src/Mutator/Boolean/LogicalOr.php
```diff
@@ -77,6 +79,79 @@ public function mutate(Node $node): iterable
...
+
+        $smallerOp = [
+            Node\Expr\BinaryOp\Smaller::class,
+            Node\Expr\BinaryOp\SmallerOrEqual::class,
+        ];
+
+        if (
+            (
+                in_array($classNodeLeft, $greaterOp, true) === true
+                && in_array($classNodeRight, $smallerOp, true) === true
```
maks-rafalko, line 130: still present, please check all cases :)

maks-rafalko:
could you also please fix PHPStan issues? https://github.com/infection/infection/actions/runs/3985550828/jobs/6833052618#step:7:90

Neirda24:
@maks-rafalko : sorry for the delay. I've figured that some of what I was trying to fix shouldn't be. I'll update the PR with added tests soon !

Neirda24:
@maks-rafalko : Okay I've pushed new code and fixed coding style using cs. I'll gadly take some help finishing this as it requires I think a lot of tests combinations to reduce the number of mutants. Some of those mutants can't be tested against I think. Let me know.

maks-rafalko:
@Neirda24  could you please rebase on master?

Neirda24:
@maks-rafalko : here are the result of psalm :

```bash
$ docker compose run php81 make psalm
./.tools/psalm --threads=max

Deprecated: Return type of HumbugBox383\KevinGH\RequirementChecker\RequirementCollection::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in phar:///opt/infection/.tools/psalm/.box/src/RequirementCollection.php on line 15

Deprecated: Return type of HumbugBox383\KevinGH\RequirementChecker\RequirementCollection::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in phar:///opt/infection/.tools/psalm/.box/src/RequirementCollection.php on line 19
Target PHP version: 8.0 (inferred from composer.json)
Scanning files...
Analyzing files...

░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░  60 / 411 (14%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 120 / 411 (29%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 180 / 411 (43%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 240 / 411 (58%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 300 / 411 (72%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 360 / 411 (87%)
░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░

ERROR: UndefinedPropertyFetch - src/Mutator/Boolean/LogicalOr.php:100:25 - Instance property PhpParser\Node\Expr::$left is not defined (see https://psalm.dev/039)
        $nodeLeftLeft = $nodeLeft->left;


ERROR: UndefinedPropertyFetch - src/Mutator/Boolean/LogicalOr.php:101:26 - Instance property PhpParser\Node\Expr::$right is not defined (see https://psalm.dev/039)
        $nodeLeftRight = $nodeLeft->right;


ERROR: UndefinedPropertyFetch - src/Mutator/Boolean/LogicalOr.php:103:26 - Instance property PhpParser\Node\Expr::$left is not defined (see https://psalm.dev/039)
        $nodeRightLeft = $nodeRight->left;


ERROR: UndefinedPropertyFetch - src/Mutator/Boolean/LogicalOr.php:104:27 - Instance property PhpParser\Node\Expr::$right is not defined (see https://psalm.dev/039)
        $nodeRightRight = $nodeRight->right;


ERROR: UndefinedMethod - src/Mutator/Boolean/LogicalOr.php:202:41 - Method PhpParser\Node\Expr::getOperatorSigil does not exist (see https://psalm.dev/022)
            return (match ("{$nodeLeft->getOperatorSigil()}::{$nodeRight->getOperatorSigil()}") {


ERROR: UndefinedMethod - src/Mutator/Boolean/LogicalOr.php:202:75 - Method PhpParser\Node\Expr::getOperatorSigil does not exist (see https://psalm.dev/022)
            return (match ("{$nodeLeft->getOperatorSigil()}::{$nodeRight->getOperatorSigil()}") {


------------------------------
6 errors found
------------------------------
416 other issues found.
You can display them with --show-info=true
------------------------------
Psalm can automatically fix 5 of these issues.
Run Psalm again with
--alter --issues=InvalidNullableReturnType,InvalidReturnType --dry-run
to see what it can fix.
------------------------------
```

But regarding `ERROR: UndefinedPropertyFetch - src/Mutator/Boolean/LogicalOr.php:100:25 - Instance property PhpParser\Node\Expr::$left is not defined (see https://psalm.dev/039)
        $nodeLeftLeft = $nodeLeft->left;` for example the issue is that right before here is the code :

```php
        if (
            !property_exists($nodeLeft, 'left')
            || !property_exists($nodeLeft, 'right')
            || !property_exists($nodeRight, 'left')
            || !property_exists($nodeRight, 'right')
        ) {
            return true;
        }

        $nodeLeftLeft = $nodeLeft->left;
        $nodeLeftRight = $nodeLeft->right;

        $nodeRightLeft = $nodeRight->left;
        $nodeRightRight = $nodeRight->right;
```

looks like a psalm issue don't you think ? There is no Interface that ensure there is a left / right property.

Neirda24:
@maks-rafalko : Fixed psalm but it looks ugly.... Let me know what you think

maks-rafalko:
> @maks-rafalko : Fixed psalm but it looks ugly.... Let me know what you think

yeah, no the best look, but I can't advice anything better, so I'm personally ok with that.

That's already a huge job! Could you look into PHPStan error as well and make `Autoreview` build green?

> Some of those mutants can't be tested against I think. Let me know.

If they can't be tested, probably we can ignore them?

Neirda24:
@maks-rafalko : there is an open issue about this (I thought I saw it merged with an article but couldn't find it) https://github.com/phpstan/phpstan/issues/4451 . Basically PHPStan complains that there is no `default` but in this case the default would be an execption same as the default behavior without the `default` key. There is still an option to set teh default to true (In doubt mutate) otherwise. What do you prefer ? It seems to be the only thing left to fix (I'll also look into mutations in CI to see if I can fix more)

maks-rafalko:
Probably exception will be better (if this is the same as default behavior when `default` is absent?

> I'll also look into mutations in CI to see if I can fix more

👍

Neirda24:
Ok I think I fixed everything I could. I added more tests and fixed psalm. Although introducing the default in match statement fixes psalm, it creates a new mutation that we cannot test (unreachable statement). Mutation still failing are Either LogicalOr | LogicalAnd or InstanceOf and I can't seems to be able to test them directly. I think we are dealing with other false positive here. Could you check and let me know what you think please @maks-rafalko  ? Also I re-arramged a bit the tests.

maks-rafalko:
I've fixed CS and added ignore rules.

Thank you for your contribution @Neirda24

maks-rafalko (approved):

Neirda24:
Thank you for the final touch !

PR #1803, opened 2023-01-17
Update webmozart/assert to 1.11

### LINKED ISSUE #1802: PHP 8.2: Deprecated: Use of "static" in callables is deprecated in .../vendor/webmozart/assert/src/Assert.php on line 939
| Question    | Answer
| Infection version | 0.26.16 (`infection.phar --version`)
| Test Framework version | PHPUnit 9.5.26
| PHP version | 8.2.1 (`php -v`)
| Platform    | Ubuntu
| Github Repo | -

When using PHP Infection with PHP 8.2, the following deprecation warning message happens:
```
Deprecated: Use of "static" in callables is deprecated in phar:///usr/local/bin/infection/vendor/webmozart/assert/src/Assert.php on line 939
```

From what I investigated, this is because PHP Infection 0.26.16 is requiring `webmozart/assert:^1.3` and thanks to another dependency, the actual version installed for webmozart/assert is **1.9.1**.   
The deprecation is fixed on webmozart/assert **1.11.0**

Would it be possible to change the composer.json to rely on `webmozart/assert:^1.11` so that PHP Infection would benefit of this fix?
Or do you have other dependencies/supports that prevent you to rely on a more recent version of webmozart/assert package?

Thanks a lot!

sanmai:
Fixes #1802

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,7 +65,7 @@
         "symfony/finder": "^5.4 || ^6.0",
         "symfony/process": "^5.4 || ^6.0",
         "thecodingmachine/safe": "^2.1.2",
-        "webmozart/assert": "^1.3"
+        "webmozart/assert": "^1.11"
     },
     "conflict": {
         "phpunit/php-code-coverage": ">9,<9.1.4 || >9.2.17,<9.2.21",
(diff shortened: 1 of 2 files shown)


localheinz (approved):

maks-rafalko (approved):
👍

PR #1805, opened 2023-01-20
ci: Enable back the integration tests on Windows

theofidry:
Follow up of https://github.com/infection/infection/pull/1787#issuecomment-1381480404.


sanmai:
~I recall they were particularly flaky~

There's an actual test failure!

theofidry:
Yes! I'm quite surprised though as it's a code that was suppose to be executed already

sanmai (approved):
If it works, it works 👍

PR #1806, opened 2023-01-20
ci: Add meta jobs to ease editing the protection rules

### LINKED ISSUE #1800: Improve the CI checks
I think CI changes such as #1775 highlight a quite annoying problem: keeping the GitHub branch protection rules up to date.

There is different ways to deal with this, one which is to use a bot to keep in sync the rule protection based on what is declared in the workflow. However I'm not too fond of relying on external tooling for this, so I looked for another solution and managed to find one that I liked. The idea is very simple, instead of having a matrix generating:

```yaml
Test with PHP 7.4
Test with PHP 8.0
Test with PHP 8.1
Test with PHP 8.2
```

And manually adding each one of those jobs, you can simply use a "meta" job:

```yaml
    validate-tests:
        name: Tests status
        runs-on: ubuntu-latest
        needs: unit-tests
        if: always()
        steps:
            - name: Successful run
              if: ${{ !(contains(needs.*.result, 'failure')) }}
              run: exit 0

            - name: Failing run
              if: ${{ contains(needs.*.result, 'failure') }}
              run: exit 1
```

It will wait on all the job of the matrix, if all passed it will pass, otherwise it will fail. Then in the protection rule you can depend on this job only, which has a static name.

For an actual demo, you can check https://github.com/box-project/box/pull/812/checks where I've extensively reworked the CI to use this to the maximum (as I was quite sick of changing the protection rules...).

If you like the idea I'm happy to try to bring this to infection step by step.

theofidry:
With this the protection rules can depend on static names. Of course they may still change over time as we adapt the CI, but this provides a lot more flexibility as we can now freely edit the matrix without having to constantly edit the project projection rules.

Closes #1800.

### DIFF
--- a/.github/workflows/autoreview.yaml
+++ b/.github/workflows/autoreview.yaml
@@ -56,3 +56,18 @@ jobs:
         # Ensures we see all the errors
         run: make autoreview --keep-going
 
+  # This is a meta job to avoid to have to constantly change the protection rules
+  # whenever we touch the matrix.
+  tests-status:
+    name: Autoreview Status
+    runs-on: ubuntu-latest
+    needs: tests
+    if: always()
+    steps:
+      - name: Successful run
+        if: ${{ !(contains(needs.*.result, 'failure')) }}
+        run: exit 0
+
+      - name: Failing run
+        if: ${{ contains(needs.*.result, 'failure') }}
+        run: exit 1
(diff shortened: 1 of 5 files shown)


sanmai (approved):
:heart:

theofidry:
Rules updated

Issue #857, opened 2019-11-26
Directory resolution inconsistency


Bilge:
Given the following JSON configuration fragment

```json
    "source": {
        "directories": [
            "foo"
        ]
    },
    "phpUnit": {
        "configDir": "bar"
    },
```

* `foo` is resolved relative to the current working directory.
* `bar` is resolved relative to the location of the configuration file.

This is inconsistent and very odd behaviour.

Bilge:
Note there are other directory specifiers in a typical configuration file to which this also applies, e.g. log file location.

maks-rafalko:
> Note there are other directory specifiers in a typical configuration file to which this also applies, e.g. log file location.

this one has been fixed

Bilge:
I waited 6 years for relative directories.

maks-rafalko:
Well, it could have been a day if you had implemented it 😉

Help is always appreciated.

PR #1811, opened 2023-01-27
Treat log paths as relative to config directory

LeoVie:
Partly fixes https://github.com/infection/infection/issues/857 (https://github.com/infection/infection/issues/857#issuecomment-558550587)

Log paths will be treated as relative to config directory with this change.

This is a breaking change for all users, that use relative paths in their config, right now. 

One one hand I'm not 100% sure, if this is worth it. On the other hand infection is still in `0.x`, so breaking changes should be accepted?

### DIFF
--- a/src/Configuration/ConfigurationFactory.php
+++ b/src/Configuration/ConfigurationFactory.php
@@ -153,7 +153,7 @@ class ConfigurationFactory
             ),
             $this->retrieveFilter($filter, $gitDiffFilter, $isForGitDiffLines, $gitDiffBase, $schema->getSource()->getDirectories()),
             $schema->getSource()->getExcludes(),
-            $this->retrieveLogs($schema->getLogs(), $useGitHubLogger, $htmlLogFilePath),
+            $this->retrieveLogs($schema->getLogs(), $configDir, $useGitHubLogger, $htmlLogFilePath),
             $logVerbosity,
             $namespacedTmpDir,
             $this->retrievePhpUnit($schema, $configDir),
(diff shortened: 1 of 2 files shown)


src/Configuration/ConfigurationFactory.php
```diff
@@ -348,4 +358,19 @@ private function detectCiGithubActions(): bool
...
+    ): ?string {
+        if ($path === null) {
+            return null;
+        }
+
+        if (Path::isAbsolute($path)) {
+            return $path;
+        }
+
+        return sprintf('%s/%s', $configDir, $path);
```
maks-rafalko, line 374: Looks good to me. While it's a BC

1. we are still on 0.x
2. I think users mostly place `infection.json` at the root of the project, so I hope not many users will be affected. Anyway, we will note it in the release notes/changelog

maks-rafalko (approved):

maks-rafalko:
Looks good to me. While it's a BC

1. we are still on 0.x
2. I think users mostly place `infection.json` at the root of the project, so I hope not many users will be affected. Anyway, we will note it in the release notes/changelog

Please look at the failed build.

LeoVie:
Thanks for the feedback. Could you give me any hint, which tests exactly fail at the moment? I'm not really getting the CI output... Maybe a re-run of the failed job could be good enough.

maks-rafalko:
there was a bug with `php://out` and `php://stderr` that were prefixed to make them absolute paths, I've fixed it in https://github.com/infection/infection/pull/1851 because I couldn't push to your fork since it's too old

Thank you for your contribution, will merge in https://github.com/infection/infection/pull/1851

PR #1814, opened 2023-02-03
fix(phpunit): pass DiffOutputBuilderInterface instance to Differ

simPod:


maks-rafalko:
Ultimately, we need a `e2e` test for PHPUnit 10. However, currently we haven't merged https://github.com/infection/infection/pull/1765 yet, and unfortunately this would force us to run such `e2e` test by condition (if the build is executed on PHP 8.1+ because PHPUnit 10 requires PHP 8.1).

---

I confirm that with this patch Infection works with PHPUnit 10, checked locally

maks-rafalko (approved):

maks-rafalko:
Thank you @simPod

PR #1815, opened 2023-02-03
ci: Adapt the matrix to allow testing against different dependency ranges

theofidry:
Rework the matrix to make it easier to test against`locked`, `highest` and `lowest` dependencies.
Also allow to update and test the Composer version used easily.

Extracted from #1775.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,7 +30,7 @@ jobs:
           - { operating-system: 'ubuntu-latest', php-version: '8.1', coverage-driver: 'pcov', e2e-runner: 'build/infection.phar' }
           - { operating-system: 'ubuntu-latest', php-version: '8.2', coverage-driver: 'xdebug', e2e-runner: 'bin/infection' }
 
-    name: CI on ${{ matrix.operating-system }} with PHP ${{ matrix.php-version }}, using ${{ matrix.coverage-driver }} with ${{ matrix.e2e-runner }}
+    name: CI on ${{ matrix.operating-system }} with PHP ${{ matrix.php-version }} (${{ matrix.composer }}; ${{ matrix.dependency }}), using ${{ matrix.coverage-driver }} with ${{ matrix.e2e-runner }}
 
     steps:
       - name: Checkout code
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,16 +19,16 @@ jobs:
         operating-system: [ ubuntu-latest ]
         php-version: [ '8.0' ]
         dependency: [ locked ]
-        composer: [ composer:v2.1 ]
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -58,16 +59,24 @@ jobs:
         uses: actions/cache@v2
```
theofidry, line 59: note that this block, L59 to L79, could be a single simple action with `ramsey/install` as shown [here](https://github.com/infection/infection/pull/1781/files#diff-87ab0e37e52f2d5a98de1c719e9222197e1f875c57ed290a6b10f0c640e5b001R33).

.github/workflows/e2e-tests.yaml
```diff
@@ -59,16 +60,24 @@ jobs:
...
-            composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('composer.*') }}-
+            composer-${{ runner.os }}-${{ matrix.php-version }}-${{ matrix.dependencies }} }}-
             composer-${{ runner.os }}-${{ matrix.php-version }}-
             composer-${{ runner.os }}-
             composer-
 
       - name: Install dependencies
-        run: |
-          composer install --no-interaction --prefer-dist --no-progress
+        if: ${{ matrix.dependencies == 'locked' }}
```
maks-rafalko, line 71: it doesn't work in any job (see https://github.com/infection/infection/actions/runs/4091779346/jobs/7056080843) because of typo.

Here - dependencIES
In other case - dependencY (https://github.com/infection/infection/pull/1815/files#diff-deaabb10671a187e9d8bc2a25d16791aa368a5a6e3d970c3d42977712b6fd60fR21)

I think it should be `dependencies` everywhere

.github/workflows/e2e-tests.yaml
```diff
@@ -18,17 +18,19 @@ jobs:
       matrix:
         operating-system: [ ubuntu-latest ]
         php-version: [ '8.0' ]
+        dependency: [ locked ]
```
maks-rafalko, line 21: do we need to include `lowest`, `highest` here? Otherwise, I think I don't get how the are execued then

theofidry: The idea of this PR was just to be a preparation step. Adding `lowest` and `highest` should be done in dedicated PR IMO as they will require additional changes (code-wise).

.github/workflows/e2e-tests.yaml
theofidry: Renamed it

.github/workflows/e2e-tests.yaml
theofidry: Rephrased the PR description, sorry this was not clear enough

maks-rafalko (approved):

PR #1363, opened 2020-10-22
Remove PHPDBG from the build matrix on Travis CI

sanmai:
This PR:

- [ ] Removes PHPDBG from the build matrix on Travis CI

PR #1817, opened 2023-02-03
ci: Enforce the CI to test against Symfony5

theofidry:
Depends on #1815.

Adapt the CI to enforce testing against Symfony5 and allowing to easily add a build against Symfony6, preparing #1775.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -30,7 +30,7 @@ jobs:
             coverage-driver: xdebug
             symfony-version: '5.4.*'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yaml
```diff
@@ -18,20 +18,28 @@ jobs:
       matrix:
         operating-system: [ ubuntu-latest ]
         php-version: [ '8.0', '8.1', '8.2' ]
+        # TODO: include highest & lowest
+        dependency: [ locked ]
+        # TODO: include phpdbg later
```
maks-rafalko, line 23: btw, we don't need phpdbg as PHPUnit dropped support for it and we removed it https://github.com/infection/infection/pull/1363

maks-rafalko (approved):

PR #1824, opened 2023-02-05
ci: Run the tests against Symfony6

theofidry:
This is mostly an extract from #1775.

After this PR what should be left is adding the complementary jobs to the CI (lowest/highest).

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -15,11 +15,11 @@ jobs:
     runs-on: ${{ matrix.operating-system }}
 
     strategy:
+      fail-fast: false
       matrix:
         operating-system: [ ubuntu-latest ]
         php-version: [ '8.0', '8.1', '8.2' ]
-        # TODO: include highest & lowest
-        dependencies: [ locked ]
+        dependencies: [ locked, lowest, highest ]
         coverage-driver: [ pcov, xdebug ]
         symfony-version: [ '5.4.*', '6.*.*' ]
         include:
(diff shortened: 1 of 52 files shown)


theofidry:
I really can't tell if the STTY issue is really related or not

devTools/phpstan-src.neon
```diff
@@ -47,11 +47,15 @@ parameters:
             message: '#^Function shell_exec is unsafe to use#'
             path: '../src/TestFramework/CommandLineBuilder.php'
 
+
+
```
maks-rafalko, line 51: not needed?

maks-rafalko (commented):
what is the issue with STTY?

theofidry:
>what is the issue with STTY?

no longer appearing... Not sure why...

.github/workflows/tests.yaml
```diff
@@ -21,8 +21,7 @@ jobs:
         # TODO: include highest & lowest
         dependencies: [ locked ]
         coverage-driver: [ pcov, xdebug ]
-        # TODO: include Symfony6 later
-        symfony-version: [ '5.4.*' ]
+        symfony-version: [ '5.4.*', '6.*.*' ]
```
maks-rafalko, line 24: not sure I understand how it works now.

All the builds have now `locked` value for `dependencies` key. Each GH job runs `composer install`, ignoring (as I see) SF 5.4.* or 6..* - jobs are just use the versions from the lock file.

- should we add `highest` and `lowest` on line 21?
- should we vary SF versions **only** when `dependencies` is not `locked`?

theofidry: Resolved in #1835

theofidry:
It would be best to exclude the combination `locked & 6.*.*` but not sure how to do that

sidz:
@theofidry see https://github.com/infection/infection/pull/1775/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR27 if you need to exclude smth

theofidry:
@sidz I'll do that in a separate PR I think; it's only optimizations after all

maks-rafalko (approved):

PR #1835, opened 2023-03-15
ci: Run tests against higher and lower versions

theofidry:
Extracted from #1775.

This is actually a pre-requesite for #1824.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -18,8 +18,7 @@ jobs:
       matrix:
         operating-system: [ ubuntu-latest ]
         php-version: [ '8.0', '8.1', '8.2' ]
-        # TODO: include highest & lowest
-        dependencies: [ locked ]
+        dependencies: [ locked, lowest, highest ]
         coverage-driver: [ pcov, xdebug ]
         # TODO: include Symfony6 later
         symfony-version: [ '5.4.*' ]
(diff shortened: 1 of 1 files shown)


sanmai (approved):

composer.json
```diff
@@ -78,6 +78,7 @@
         "fidry/makefile": "^0.2.0",
         "helmich/phpunit-json-assert": "^3.0",
         "phpspec/prophecy-phpunit": "^2.0",
+        "phpspec/prophecy": "^1.15",
```
maks-rafalko, line 81: could you please explain why is it needed?

maks-rafalko (approved):

composer.json
theofidry: with prefer-lowest it could get to 1.3.0 which has some issues: https://github.com/infection/infection/actions/runs/4434955792/jobs/7781506076#step:13:71

maks-rafalko: but is it a direct dependency of Infection? If not, probably it's better to add it to `conflicts: {"phpspec/prophecy: "<1.15"}`? does it make sense?

theofidry: it's a dependency from `phpspec/prophecy-phpunit`.

I think correcting transient or indirect dependencies can be done like so. Using `conflict` would make it incompatible as a package, which is not true (it's only a problem for our tests)

PR #1836, opened 2023-03-20
Log thread count as part of performance

icanhazstring:
When using `--threads=max` it's hard to figure out how many threads were actually used. This will help figure out are more details about the performance of the infection run.

This is just a suggestion for now. If you think it's useful, I can add coverage for tests.

This PR:

- [x] Adds new feature ...
- [x] Covered by tests

### DIFF
--- a/src/Resource/Listener/PerformanceLoggerSubscriber.php
+++ b/src/Resource/Listener/PerformanceLoggerSubscriber.php
@@ -42,7 +42,7 @@ use Infection\Resource\Memory\MemoryFormatter;
 use Infection\Resource\Time\Stopwatch;
 use Infection\Resource\Time\TimeFormatter;
 use function memory_get_peak_usage;
-use function Safe\sprintf;
+use function sprintf;
 use Symfony\Component\Console\Output\OutputInterface;
 
 /**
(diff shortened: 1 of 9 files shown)


src/Resource/Listener/PerformanceLoggerSubscriber.php
```diff
@@ -68,7 +69,8 @@ public function onApplicationExecutionWasFinished(ApplicationExecutionWasFinishe
             sprintf(
                 'Time: %s. Memory: %s',
```
icanhazstring, line 70:
```suggestion
                'Time: %s. Memory: %s'. Threads: %s,
```

maks-rafalko:
I'm personally ok with such addition.

maks-rafalko:
I think we can easily make it as a service in the container, rather than class with static method. Regarding caching the result: it will be already done automatically by underlying service if we instantiate it and inject through constructor (thus - once)

https://github.com/theofidry/cpu-core-counter/blob/b58e5a3933e541dc286cc91fc4f3898bbc6f1623/src/CpuCoreCounter.php#L46-L51

icanhazstring:
Ah I see. Will convert as a service with non static method call then 👍

src/Resource/Memory/MemoryFormatterInterface.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Resource\Memory;
+
+/**
+ * @internal
+ */
+interface MemoryFormatterInterface
```
icanhazstring, line 41: Question: Ok to use interfaces here for `MemoryFormatter` and `TimeFormatter` to be able to test it properly.
Or should we remove `final` from those classes to be able to provide mockability or fake fixtures?

icanhazstring: Got it. Will update 👍

maks-rafalko (approved):
Thank you for this addition, @icanhazstring

icanhazstring:
Your welcome. Thanks for the nice review 🙂
Also nice developer tooling. Was really easy to work with 💃

Issue #1846, opened 2023-04-24
Attributes on parameters of method signature are not detected as part of method signature


niconoe-:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.26.20
| Test Framework version | PHPUnit 10.1.1
| PHP version | 8.2.5
| Platform    | Ubuntu

Hi all!

I discovered a really annoying bug with PHP Attributes applied on parameters being parsed by Infection when they're part of a method signature, in that Infection does not detect them as part of method signature.
In order to better explain, here is a snippet example:

```php
<?php
declare(strict_types=1);

class FooEntity
{
    public function __construct(
        #[\Doctrine\ORM\Mapping\Column(nullable: false)]
        public string $label,
    ) {
    }
}
```

**The important node here is `false` as it will be mutated to `true`.** When you do a unit test that covers everything here, including that the `nullable` arg is actually set to `false` and nothng else, PHP Infection will **wrongly infer** that the `FalseValue` mutant replacing `false` by `true` is uncovered.

Why that? That's the bug.

Here you can see on my snippet that the `false` in the attribute is declared line 7, but the codeCoverage (no matter if xDebug or PCOV, I've tried both) will declare that:
* line coverage is on range [10] (the `}` of the method declaration)
* method coverage is on range [6;10] (from `public function…` to the closing `}` of the method).

Like I said in this case, the `FalseValue ` mutant will wrongly be declared as **uncovered**, because the associated test coverage will be the _line coverage_, even if this Node is declared on an attribute that applies on a parameter of a method signature.

After long hours of debugging, I discovered that the issue comes from `ReflectionVisitor::isPartOfFunctionSignature`, which, when used with the Node `false`, is returning `false` while it's definitly `true`. It does so because only the 1st level parent is analyzed, but with the PHP Attribute feature, only looking at the 1st level parent of the Node is no longer sufficient. For example, my `ConstFetch` Node ("false") has the parent Node\Arg, which has the parent Node\Attribute, which has the parent Node\AttributeGroup, which finally has the parent Node\Param.

I tried a fix that loop over the parents until the parent is a Node\Param or there's no parent and I apply the same filter than previously, and finally it worked: the test coverage applied to the mutant is now the **method coverage**, so the mutant is in the appropriate range, and is finally considered as covered (and in my case, even killed because I wrote my test in a way to kill such mutant).

Here is the fix:
```php
    private function isPartOfFunctionSignature(Node $node): bool
    {
[...]

PR #1847, opened 2023-04-25
Fix node being undetected as part of function signature when too deeply nested.

niconoe-:
This PR:

- [x] Covered by tests
- [x] Fixes https://github.com/infection/infection/issues/1846

To detect if a Node is part of a signature of a function/method, it used to be checked if that Node is a "function-like" Node or if its direct parent is a Node\Param. 
With the usage of PHP Attributes, Nodes to check can be more deeply nested, and looking at the direct parent is no longer sufficient.
This PR ensures that all ancestors are being parsed until a Node\Param is found, or no more parent exists.

### DIFF
--- a/src/PhpParser/Visitor/ReflectionVisitor.php
+++ b/src/PhpParser/Visitor/ReflectionVisitor.php
@@ -129,19 +129,25 @@ final class ReflectionVisitor extends NodeVisitorAbstract
         return null;
     }
 
+    /**
+     * Loop on all parents of the node until one is a Node\Param or a function-like, which means it is part of a
+     * signature.
+     */
     private function isPartOfFunctionSignature(Node $node): bool
     {
         if ($this->isFunctionLikeNode($node)) {
             return true;
         }
 
-        $parent = ParentConnector::findParent($node);
-
-        if ($parent === null) {
-            return false;
+        if ($node instanceof Node\Param) {
+            return true;
         }
 
-        return $parent instanceof Node\Param || $node instanceof Node\Param;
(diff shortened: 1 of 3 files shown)


sanmai (approved):
The change to the visitor makes sense, but I didn't quite get the tests (probably because I'm too tired)

niconoe-:
> The change to the visitor makes sense, but I didn't quite get the tests (probably because I'm too tired)

I basically copy-pasted the integration test "ReflectionVisitorTest::test_it_marks_nodes_which_are_part_of_the_function_signature", but I added an attribute with a "Node\Expr\ConstFetch" element within in the duplicated test. As this node is actually part of the method signature, we expect `true`, but it cannot be detected as part of the method signature without my fix: the parent's node is not a "Node\Param" while it has an ancestor which is a "Node\Param".

In a clearer way, you can try with the Unit Test I added without the source code edition and you'll see that the dataset with the provider about the "Node\Expr\ConstFetch" node won't pass the test, while when applying my fix, it passes. :slightly_smiling_face:

maks-rafalko (approved):
Spent a couple of minutes with tests and yeah, they do make sense, as well as the source code change.

I absolutely appreciate your time debugging Infection and finding the root and fixing it, thank you.

maks-rafalko:
Regarding the differencies on Playground before and after update: I believe this is related to different versions of PHP (8.0 before, 8.1 currently), PHPUnit (9.5 before, 10.1 currently). Because too much has changed related to coverage.

Issue #1845, opened 2023-04-23
Do not mutate `instanceof` in `assert()`


mabar:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.26.20
| Test Framework version | PHPUnit 9.6.7
| PHP version | 8.2.5
| Platform    | Debian
| Github Repo | -

See at https://infection-php.dev/r/ylep

maks-rafalko:
That's true. For this, we specifically has a [feature to ignore mutants](https://infection.github.io/guide/how-to.html#Do-not-mutate-the-source-code-matched-by-regular-expression), see updated playground: https://infection-php.dev/r/p3xk

```json5
{
    "mutators": {
        "@default": true,
        "global-ignoreSourceCodeByRegex": [
            "assert\\(.*\\);"
        ]
    }
}
```

mabar:
@maks-rafalko Wouldn't it be better to ignore `assert(true)` by default? It just does not make sense at all to have it reported

maks-rafalko:
> I don't know if it is possible to do that internally for specific rules (instanceof) or for assert(true) in general. So ignore it if it's not simple to do.

It's absolutely doable, and I agree we can ignore such mutation on Mutator level. If you want to help, please start here:

https://github.com/infection/infection/blob/37dd2f590bec3d0058e11dbeefc8f93fc9132c62/src/Mutator/Boolean/InstanceOf_.php#L83

For now, just use `ignoreSourceCodeByRegex` - this will solve your issue

sanmai:
I doubt it will be easy though. Infection doesn't come with a static analyzer so it can't tell if your `assert()` is a global method or a method imported from somewhere else.

maks-rafalko:
> An alternative solution could be `@infection-ignore instanceof`.

we already have a feature on config level for ignoring globally or per-mutator: https://infection.github.io/guide/how-to.html#Do-not-mutate-the-source-code-matched-by-regular-expression

```json
{
    "mutators": {
        "InstanceOf_": {
            "ignoreSourceCodeByRegex": [
                "assert\\(.*\\);"
            ]
        }
    }
}
```

> I doubt it will be easy though. Infection doesn't come with a static analyzer so it can't tell if your assert() is a global method or a method imported from somewhere else.

don't think it's a problem to be honest. We don't check all the internal functions (`array_map`, et.c) whether they are native or from other package, but still mutate them. IMO it's ok to not mutate `instanceof` in **any** `assert()` function, be it native or not.

PR #1852, opened 2023-05-11
Do not mutate `$var instanceof ClassName` inside `assert()` function as it's impossible or hard to kill

maks-rafalko:
Fixes https://github.com/infection/infection/issues/1845


sidz (approved):

maks-rafalko:
Thank you @sidz

Issue #1825, opened 2023-02-06
Infection's `--only-covering-test-cases` flag does not work correctly on PHPUnit 10


acelaya:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.26.19
| Test Framework version | PHPUnit 10.0.4
| PHP version | 8.2.2
| Platform    | Ubuntu
| Github Repo | https://github.com/shlinkio/shlink-installer

I'm updating a project to PHPUnit 10, and with that version, infection reports 42% of killed mutants, while it reports 90% with PHPUnit 9.6

I have even manually applied some of the mutants reported as "not killed", and then the tests no longer pass, so these are false positives.

* Project: https://github.com/shlinkio/shlink-installer
* Pull request updating to PHPUnit 10: https://github.com/shlinkio/shlink-installer/pull/181
* Build: https://github.com/shlinkio/shlink-installer/actions/runs/4107812419/jobs/7087757162

acelaya:
Same here https://github.com/shlinkio/shlink-config/pull/25.

This one is a farely smaller project,  but the score has dropped from 95% to 70% and they are also false positives.

sanmai:
I could not reproduce the problem. On PHPUnit 9.6.3 I'm looking at:

```
27 mutations were generated:
      23 mutants were killed
       0 mutants were configured to be ignored
       0 mutants were not covered by tests
       4 covered mutants were not detected
```
After running `composer require -W --dev phpunit/phpunit:^10` followed by `php vendor/bin/phpunit --migrate-configuration` I get:

```
27 mutations were generated:
      23 mutants were killed
       0 mutants were configured to be ignored
       0 mutants were not covered by tests
       4 covered mutants were not detected
```
There is no difference whatsoever.

acelaya:
What I'm doing in order to avoid running the tests twice is first run phpunit like this:

`phpunit --order-by=random --testdox --colors=always --coverage-clover=build/clover.xml --coverage-xml=build/coverage-xml --log-junit=build/junit.xml`

And then run infection like this:

`infection --threads=4 --min-msi=85 --log-verbosity=default --only-covered --only-covering-test-cases --coverage=build --skip-initial-tests`

The code coverage is generated with `pcov`.

sanmai:
We'll that's peculiar indeed. If I do this:
```
composer require -W --dev phpunit/phpunit:^9
[...]

PR #1854, opened 2023-05-13
Make PHPUnit 10.1 XML coverage report and test cases names with provider compatible with Infection and old format

maks-rafalko:
In PHPUnit >=10 data providers with keys are stored as `Class\\test_method#some key`

In PHPUnit <10 data providers with keys are stored as `Class\\test_method with data set "some key"`.

We need to translate to the old format because this is what PHPUnit <10 and >=10 understand from CLI `--filter` option.

In other words, PHPUnit 10 has a BC break in XML coverage reports that this PR workarounds.

Fixes https://github.com/infection/infection/issues/1825


sidz (approved):

maks-rafalko:
thank you, @sidz

PR #1856, opened 2023-05-14
Remove `xdebug-filter.php` as it's not used and deprecated

maks-rafalko:
See https://github.com/sebastianbergmann/phpunit/issues/4226

Also, we actually don't use it.


sanmai (approved):

maks-rafalko:
thank you @sanmai

Issue #1498, opened 2021-03-18
It would be useful if you could kill infection in a way that it still writes out some partial log


HenkPoley:
Currently you can have it running for hours on your dev machine, and then you are at a point that you _have_ to edit the code for some reason. So you kill the test run. But nothing useful for all that work gets stored (as far as I know).

HenkPoley:
So.. the default configuration (infection.json.dist) doesn't write any logs?

sanmai:
Probably you are. There are no logs because they're big users of buffer space, and, getting back to the feature request, logs are written after the fact, in a pretty sorted fashion. Using `--show-mutations` is way more efficient.

HenkPoley:
Okay, so maybe have a little thing show up "hmm, your initial test suite is taking longer than 30 seconds, are you sure you want to run without --show-mutations?" 😅

HenkPoley:
In what way is `--show-mutations` more efficient? It's not outputting mutations in-line during runtime. So it still needs to buffer them?

sanmai:
It is, but it only buffers escaped mutations. In your case example it'll be only 8 out of other 1500.

maks-rafalko:
@HenkPoley released in `0.27.0`

HenkPoley:
Ah, thank you, a colleague recently even discovered my CI run from back when I entered this issue 😅

PR #1857, opened 2023-05-14
Calculate results and show metrics if Infection is interrupted with `SIGINT` (ctrl + c)

maks-rafalko:
Implements https://github.com/infection/infection/issues/1498

Before (in `master`), when Infection is interrupted, all the results were lost:

infection-signal-1

With this updated, all the accumulated results (before hitting `CTLR+C`) are used to prepare metrics. Also, file logs are created as well with accumulated results.

If `-s` is used, it additionally prints to console so developer can see mutants that were executed before `CTRL + C`:

infection-signal-2

### DIFF
--- a/src/Process/Runner/ParallelProcessRunner.php
+++ b/src/Process/Runner/ParallelProcessRunner.php
@@ -61,11 +61,18 @@ final class ParallelProcessRunner implements ProcessRunner
      */
     private array $availableThreadIndexes = [];
 
+    private bool $shouldStop = false;
+
     /**
      * @param int $poll Delay (in milliseconds) to wait in-between two polls
      */
     public function __construct(private readonly int $threadCount, private readonly int $poll = 1000)
     {
+    }
+
+    public function stop(): void
+    {
+        $this->shouldStop = true;
     }
 
     public function run(iterable $processes): void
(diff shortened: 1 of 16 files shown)


src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -92,6 +99,10 @@ public function run(iterable $processes): void
 
         // start the initial batch of processes
         while ($process = array_shift($bucket)) {
+            if ($this->shouldStop) {
```
maks-rafalko, line 102: This is exactly how Symfony Messenger handles the same signals. Basically, the idea and implementation are similar to Symfony's one.

See 

- https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Messenger/EventListener/DispatchPcntlSignalListener.php
- https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Messenger/EventListener/StopWorkerOnSignalsListener.php
- https://github.com/symfony/symfony/blob/8c637a5bbfea8b322ec93eb51ac47c0db8948e2a/src/Symfony/Component/Messenger/Worker.php#L111

maks-rafalko:
Not sure I understand Fixer failure: https://github.com/infection/infection/actions/runs/4971604352/jobs/8896243867?pr=1857#step:5:41

locally, with docker I have exaclty the same PHP: 8.1.17 and exactly the same Fixer's version, but it doesn't fail

Issue #1863, opened 2023-05-16
Infection can't run custom Test Framework extensions because TestFrameworkTypes is hardcoded 


thompsnm:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.26.3
| Test Framework version | Custom framework via an Infection extension
| PHP version | 8.0.28
| Platform    | CentOS
| Github Repo | -

## Description

My team has implemented a binary that wraps around PHPUnit and configures / executes tests for us when we simply point it at a particular directory. I want to implement an Infection Test Framework extension that we would only use internally that executes tests via our binary. Unfortunately, that's not possible because specifying a Test Framework whose type is not hardcoded in `TestFrameworkTypes` causes Infection to throw an error.

## Reproduction Steps

1. Create a new Test Framework extension by following [the "How to write an extension for Infection" steps](https://github.com/infection/extension-installer#how-to-write-an-extension-for-infection)
2. Install your Test Framework extension by following [the "How to install extension" steps](https://github.com/infection/extension-installer#how-to-install-extension)
3. Configure Infection to use your Test Framework extension by specifying its Adapter Name as the `testFramework`. For instance:
```
"testFramework": "PhpUnitTestRunner"
```
4. Run Infection

### Actual Result

Infection throws the following error:

```
In Assert.php line 2060:

  Expected one of: "pest", "phpunit", "phpspec", "codeception". Got: "PhpUnitTestRunner"
```

### Expected Result

I would suggest instead of [hardcoding the Test Framework Types](https://github.com/infection/infection/blob/f669c4ebcda0050072fbb69bfb81a57f7709ea3e/src/TestFramework/TestFrameworkTypes.php#L48-L53) that the `TYPES` array should dynamically include [the Adapter Name](https://github.com/infection/abstract-testframework-adapter/blob/master/src/TestFrameworkAdapterFactory.php#L51) of any installed Test Framework extensions.

PR #1871, opened 2023-06-22
Include installed Test Frameworks in TestFrameworkTypes - Fixes 1863

thompsnm:
#### This PR:

Fixes https://github.com/infection/infection/issues/1863

#### Per the issue description:

My team has implemented a binary that wraps around PHPUnit and configures / executes tests for us when we simply point it at a particular directory. I want to implement an Infection Test Framework extension that we would only use internally that executes tests via our binary. Unfortunately, that's not possible because specifying a Test Framework whose type is not hardcoded in TestFrameworkTypes causes Infection to throw an error.

#### Changes mage:

I replaced all references to the constant array `TestFrameworkTypes::TYPES` with a static function `TestFrameworkTypes::getTypes()` that parses `GeneratedExtensionsConfig::EXTENSIONS` to include the names of all installed Adapters.

### DIFF
--- a/src/TestFramework/TestFrameworkTypes.php
+++ b/src/TestFramework/TestFrameworkTypes.php
@@ -35,8 +35,10 @@ declare(strict_types=1);
 
 namespace Infection\TestFramework;
 
-use Infection\ExtensionInstaller\GeneratedExtensionsConfig;
+use function count;
 use Infection\AbstractTestFramework\TestFrameworkAdapterFactory;
+use Infection\ExtensionInstaller\GeneratedExtensionsConfig;
+use function is_a;
 use Webmozart\Assert\Assert;
 
 /**
--- a/src/Command/ConfigureCommand.php
+++ b/src/Command/ConfigureCommand.php
@@ -88,7 +88,7 @@ final class ConfigureCommand extends BaseCommand
                 InputOption::VALUE_REQUIRED,
                 sprintf(
                     'Name of the Test framework to use ("%s")',
-                    implode('", "', TestFrameworkTypes::TYPES)
+                    implode('", "', TestFrameworkTypes::getTypes())
                 ),
                 TestFrameworkTypes::PHPUNIT
             );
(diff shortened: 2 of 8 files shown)


src/TestFramework/TestFrameworkTypes.php
```diff
@@ -45,10 +51,40 @@ final class TestFrameworkTypes
...
+     * @param mixed[] $installedExtensions
+     *
+     * @return string[]
+     */
+    public static function getTypes(
+        array $installedExtensions = GeneratedExtensionsConfig::EXTENSIONS
+    ): array {
+        $types = self::$defaultTypes;
+
+        if (count($installedExtensions) > 0) {
```
thompsnm, line 74: I see there's a warning from "[GitHub Actions / Mutation Testing Code Review Annotations 8.1](https://github.com/infection/infection/actions/runs/5350969967/jobs/9704233444?pr=1871)" on this line that changing `>` to `>=` is an escaped Mutant. I just confirmed that this case would be caught by `phpstan analyse`:

```
 ------ ---------------------------------------------------------------------
  Line   TestFramework/TestFrameworkTypes.php
 ------ ---------------------------------------------------------------------
  74     Comparison operation ">=" between int<0, max> and 0 is always true.
 ------ ---------------------------------------------------------------------
```

src/TestFramework/TestFrameworkTypes.php
```diff
@@ -45,10 +51,40 @@ final class TestFrameworkTypes
...
+            foreach ($installedExtensions as $installedExtension) {
+                $factory = $installedExtension['extra']['class'];
+
+                Assert::classExists($factory);
+
+                if (!is_a($factory, TestFrameworkAdapterFactory::class, true)) {
+                    continue;
```
sanmai, line 81: Should we throw here instead, telling user they made a mistake?

sanmai (approved):
Sounds like a good idea.

maks-rafalko (approved):
Thank you @thompsnm

PR #1880, opened 2023-10-09
Allow missing elements when destructuring arrays in foreach loops

MidnightDesign:
This PR:

- [x] Covered by tests

This fixes a bug where Infection threw an exception when generating mutations for a codebase containing `<?php foreach ($items as [, $value]) {  }`.

The issue was an assertion that checked whether all elements in `Array_#$items` were non-null.

I'm not sure this is the ideal solution for this problem. Maybe we _do_ want to generate mutants in this position. But for now, it will at least no longer throw.

If I _just_ removed the assertion and widened the argument type on `getItemsIndexes`, then it would generate a mutant removing the empty first element, but I'm unsure whether that's an interesting mutation to have.

### DIFF
--- a/src/Mutator/Removal/ArrayItemRemoval.php
+++ b/src/Mutator/Removal/ArrayItemRemoval.php
@@ -115,12 +115,6 @@ DIFF
      */
     public function mutate(Node $node): iterable
     {
-        $parent = $node->getAttribute('parent');
-        // Don't mutate destructured values in foreach loops
-        if ($parent instanceof Node\Stmt\Foreach_ && $parent->valueVar === $node) {
-            return;
-        }
-
         foreach ($this->getItemsIndexes($node->items) as $indexToRemove) {
             $newArrayNode = clone $node;
             unset($newArrayNode->items[$indexToRemove]);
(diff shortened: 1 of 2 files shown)


src/Mutator/Removal/ArrayItemRemoval.php
```diff
@@ -116,7 +115,11 @@ public static function getDefinition(): ?Definition
      */
     public function mutate(Node $node): iterable
     {
-        Assert::allNotNull($node->items);
+        $parent = $node->getAttribute('parent');
+        // Don't mutate destructured values in foreach loops
+        if ($parent instanceof Node\Stmt\Foreach_ && $parent->valueVar === $node) {
```
maks-rafalko, line 120: Shouldn't it be in the `canMutate()` method below? Since there we are deciding whether to mutate `Node` or not

MidnightDesign: Sorry, I totally overlooked that method. An update is on its way.

maks-rafalko (approved):
Thank you @MidnightDesign

maks-rafalko:
Released https://github.com/infection/infection/releases/tag/0.27.4

PR #1883, opened 2023-10-15
Micro optimization to reduce array_merge calls

sidz:
Not a big win but this is a micro optimization will reduce amount of `array_merge` calls in case `excludes` option has more than one entry as Finder is doing array merge all the time `notPath` is called.

see https://github.com/symfony/finder/blob/5.4/Finder.php#L288

### DIFF
--- a/src/FileSystem/SourceFileCollector.php
+++ b/src/FileSystem/SourceFileCollector.php
@@ -59,17 +59,12 @@ class SourceFileCollector
             return [];
         }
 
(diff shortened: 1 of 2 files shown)


sanmai (approved):

maks-rafalko (approved):

maks-rafalko:
Bot blocks auto-merge, merging by hand. Thanks @sidz

PR #1887, opened 2023-10-16
Replace Safe\sprintf with native sprintf

sidz:
This PR removes deprecated `Safe\sprintf` function usage in a favor of native.

Since PHP 8 `sprintf` throws an exception in case of bad format, etc.
See https://www.php.net/sprintf#refsect1-function.sprintf-errors

Also function is deprecated in Safe either: https://github.com/thecodingmachine/safe/blob/master/deprecated/strings.php#L296


maks-rafalko (approved):
Thank you @sidz

PR #1891, opened 2023-10-24
Update many dependencies

sanmai:
This PR:

- [x] Updates a bunch of dependencies


maks-rafalko:
should we close it @sanmai? there were many changes recently in composer.lock, so if for some reason it's needed, it's better to start from scratch anyway :)

PR #1892, opened 2023-10-29
Add Symfony 7 support

sidz:
This PR brings Symfony 7 support


.github/workflows/tests.yaml
```diff
@@ -21,13 +21,18 @@ jobs:
...
         include:
           - operating-system: windows-latest
             php-version: '8.1'
             dependencies: locked
             coverage-driver: xdebug
             symfony-version: '5.4.*'
+        exclude:
+          - php-version: '8.1'
+            symfony-version: '7.*.*'
+          - dependencies: 'locked'
```
sidz, line 34: it seems like we could exclude symfony-version = 6 with locked dependecies either as it doesn't make sense as we locked with symfony 5.4 components so far.

composer.json
```diff
@@ -74,7 +74,6 @@
     },
     "require-dev": {
         "ext-simplexml": "*",
-        "brianium/paratest": "^6.3",
```
sidz, line 77: We have an issue with `paratest`:

On one hand we can't update `paratest` to the newest version as it requires phpunit 10.4 but we use 9.6.
On other hand we can't update to phpunit 10 as we use `phpspec/prophecy-phpunit` which is not compatible with newest phpunit. Also `symfony/phpunit-bridge` is not compatible with newest phpunit either (see https://github.com/symfony/symfony/issues/49069) yet.

I didn't want to update phpunit in this PR and decided to remove `paratest` as it seems like we don't use it in CI at all. I suppose it was used to speed up tests locally only.


Additionally I think we could get rid of `phpspec/prophecy` (in another PR) as we don't use it a lot and it could be easily replace by phpunit itself.

maks-rafalko: I always use `paratest` locally as it speed ups quite a slow test suite.

It also finds dependencies between tests (like this one: https://github.com/infection/infection/pull/1544/files#r683828251) which directly affects on MSI (before this fix, we had incorrectly killed mutants).

So, I would like to keep it TBH.

----

> Additionally I think we could get rid of phpspec/prophecy (in another PR) as we don't use it a lot and it could be easily replace by phpunit itself.

I would better do this (if it unblocks PHPUnit 10)

> Also symfony/phpunit-bridge is not compatible with newest phpunit either (see https://github.com/symfony/symfony/issues/49069) yet.

do you know where we use it? on one of the Symfony project we got rid of it because of PHPUnit 10 incopatibility, but it wasn't used at all.

sidz: > do you know where we use it?
everywhere :) https://github.com/infection/infection/blob/master/phpunit.xml.dist#L39

- https://github.com/infection/infection/blob/master/.github/workflows/tests.yaml#L61
- https://github.com/infection/infection/blob/master/.github/workflows/tests.yaml#L100

sidz: > It also finds dependencies between tests (like this one: https://github.com/infection/infection/pull/1544/files#r683828251) which directly affects on MSI (before this fix, we had incorrectly killed mutants).

This is a fix for paratest as far as I can see. Am I missed smth?

maks-rafalko: > This is a fix for paratest as far as I can see. Am I missed smth?

it's for Infection itself, and for paratest, because both libs expose `TEST_TOKEN`. 

When infection is executed on itself with `--threads=X, where X > 1`, tests are executed in parallel, and without that fix there were failing tests, which led to killed (incorrectly) mutants.

So, in other words, introducing Paratest at that time discovered that bug.

maks-rafalko: discussed. Currently this seems like the most optimal solution. Probably, paratest will be revised later

maks-rafalko (approved):

theofidry:
Is this really required? As you mentioned there is issues with Paratest, PHPUnit10 and actually PHP 8.1 and Symfony7 is not even out yet.

sidz:
There is no issues with symfony7 and php 8.1. Symfony 7 supports php >= 8.2 :) We will have to do smth now or in the next month anyway.

We just need try to update to PHPUnit 10 after this PR will be merged and after that we will be able to bring paratest back

composer.json
Slamdunk: Paratest [`v6.11.0`](https://github.com/paratestphp/paratest/releases/tag/v6.11.0) has just been released with `Symfony:v7` support

sidz: will update PR later today

theofidry:
Do we want to finish #1805 first? (I am totally fine with doing without)

sidz:
@theofidry don't think https://github.com/infection/infection/pull/1805 blocks this one

tests/phpunit/AutoReview/Makefile/MakefileTest.php
```diff
@@ -269,7 +269,6 @@ protected function getExpectedHelpOutput(): string
 [33mtest:[0m		 	 Runs all the tests
 [33mtest-docker:[0m		 Runs all the tests on the different Docker platforms
 [33mtest-unit:[0m	 	 Runs the unit tests
-[33mtest-unit-parallel:[0m	 Runs the unit tests in parallel
```
theofidry, line 272: I guess this should be restored

sidz: yup. Nice that we have a test :)

tests/phpunit/Fixtures/Console/FakeOutput.php
```diff
@@ -4,69 +4,12 @@
...
-        throw UnsupportedMethod::method(__CLASS__, __FUNCTION__);
-    }
+use function class_alias;
+use Composer\InstalledVersions;
+use function version_compare;
+
+if (version_compare((string)InstalledVersions::getPrettyVersion('symfony/console'), 'v6.0', '<')) {
+    class_alias(FakeOutputSymfony5::class, \Infection\Tests\Fixtures\Console\FakeOutput::class);
+} else {
+    class_alias(FakeOutputSymfony6::class, \Infection\Tests\Fixtures\Console\FakeOutput::class);
```
theofidry, line 14: not sure I think my comment got erased with the forced push? Anyway I don't quite understand why this Sf6 stub is needed _now_ when it is about Symfony6? Same with the change for the Sf5 stub

sidz: Otherwise there will be methods compatibility exception as in Symfony 7 native return types have been added https://github.com/symfony/symfony/blob/v7.0.0-BETA2/src/Symfony/Component/Console/Output/OutputInterface.php#L40

Actually this fix is needed for Symfony 7 but I realized that starting from Symfony 6 all argument types are defined in the interface either. See https://github.com/symfony/symfony/blob/6.0/src/Symfony/Component/Console/Output/OutputInterface.php#L39

But PHP doesn't complaint about this type of inconsistency and that's why we didn't get errors with Symfony 6. I basically decided that this will be okay if we add stub starting from Symfony 6.

theofidry (approved):

sidz:
Thanks @maks-rafalko , @theofidry

PR #1907, opened 2023-12-13
Force Mutators to include remedies

ttomdewit:
This PR:

- [x] Adds new feature ...
- [ ] Covered by tests

In order to improve the developer experience this PR forces Mutators to include remedies in their definitions.

### DIFF
--- a/tests/phpunit/Mutator/DefinitionTest.php
+++ b/tests/phpunit/Mutator/DefinitionTest.php
@@ -36,18 +36,10 @@ declare(strict_types=1);
 namespace Infection\Tests\Mutator;
 
 use function array_fill_keys;
-use Infection\Mutator\Boolean\TrueValue;
-use Infection\Mutator\Boolean\TrueValueConfig;
 use Infection\Mutator\Definition;
-use Infection\Mutator\Extensions\BCMath;
-use Infection\Mutator\Extensions\BCMathConfig;
-use Infection\Mutator\Extensions\MBString;
-use Infection\Mutator\Extensions\MBStringConfig;
 use Infection\Mutator\Mutator;
 use Infection\Mutator\MutatorCategory;
 use Infection\Mutator\ProfileList;
-use Infection\Mutator\Removal\ArrayItemRemoval;
-use Infection\Mutator\Removal\ArrayItemRemovalConfig;
 use Infection\Tests\SingletonContainer;
 use PHPUnit\Framework\TestCase;
 use function sprintf;
@@ -91,11 +83,11 @@ final class DefinitionTest extends TestCase
     /**
      * @dataProvider mutatorsProvider
      */
-    public function test_it_must_be_instantiated_with_remedies(
+    public function test_it_must_define_remedies(
         Mutator $mutator
     ): void {
         $this->assertNotNull(
-            $mutator->getDefinition()->getRemedies(),
+            $mutator::getDefinition()->getRemedies(),
             sprintf(
                 'Definition of [%s] must provide remedies.',
                 $mutator->getName(),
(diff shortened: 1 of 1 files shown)


tests/phpunit/Mutator/DefinitionTest.php
```diff
@@ -74,4 +87,60 @@ public function valuesProvider(): iterable
...
+        $mutators = $mutatorFactory->create(array_fill_keys(
+            ProfileList::ALL_MUTATORS,
+            []
+        ), false);
+
+        foreach ($mutators as $name => $mutator) {
+            $this->assertInstanceOf(Mutator::class, $mutator);
+
+            switch ($mutator) {
+                case $mutator instanceof TrueValue:
```
theofidry, line 119: do the mutators really need to be instantiated? The definition is static so maybe passing the mutator class is enough

theofidry (commented):
For the infection team that was not part of the Discord conversation: the goal is to document the remedies which can later be used to point the user how to address some escaped mutations. To avoid pushing all remedies at once out of fear to end up with "low quality ones just for the sake of having it", this PR can introduce the requirement, and the remedies be fixed over time in separate PRs.

tests/phpunit/Mutator/DefinitionTest.php
ttomdewit: Ah so this is what I forgot to enquire about in the Discord chat. I apologise. If it can be done without instantiating I’d prefer it and will give it a spin tomorrow.

tests/phpunit/Mutator/DefinitionTest.php
```diff
@@ -74,4 +87,60 @@ public function valuesProvider(): iterable
             'The diff',
         ];
     }
+
+    /**
+     * @dataProvider mutatorsProvider
+     */
+    public function test_it_must_be_instantiated_with_remedies(
```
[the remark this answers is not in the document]
ttomdewit: Resolved in https://github.com/infection/infection/pull/1907/commits/165b1b0298b73e0c054183df7b8df5949fb975d9

tests/phpunit/Mutator/DefinitionTest.php
ttomdewit: Resolved in https://github.com/infection/infection/pull/1907/commits/165b1b0298b73e0c054183df7b8df5949fb975d9

tests/phpunit/Mutator/DefinitionTest.php
```diff
@@ -74,4 +79,37 @@ public function valuesProvider(): iterable
...
+            sprintf(
+                'Definition of [%s] must provide remedies.',
+                $mutator->getName(),
+            ),
+        );
+    }
+
+    public function mutatorsProvider(): iterable
+    {
+        $mutatorFactory = SingletonContainer::getContainer()->getMutatorFactory();
```
ttomdewit, line 100: The reason for this inline call the the container is because providers run before `setUp` methods, in which case a call to something like `$this->mutatorFactory` would be `null`.

theofidry:
@infection/core are you fine with the idea? If so I would rebase it

maks-rafalko:
If the idea to ultimately have all the remedies, then absolutely yes.

Just today I was discussing with @Wirone this ticket (in context of his question about not always understanding how to kill mutants). Is it a coinsidence? :)

Also, as a followup for this ticket: we can add links (or inline explanations) on how to kill mutant to our GitHub annotations, so contributor will better understand what to fix when mutant is escaped. I like how Psalm adds shor links to the output, and you can open it and read explanations. This is something I would like to have for Infection too (probably integrated with Playground to show "killing tests" there).

Wirone:
> Just today I was discussing with @Wirone this ticket (in context of his question about not always understanding how to kill mutants). Is it a coinsidence? :)

No, it's not a coincidence, after our conversation I [asked](https://discord.com/channels/1090672946370052166/1091081297759309926/1226874628589228042) @ttomdewit what's the state of this, it seems like it sparked some reaction 🙂.

ttomdewit:
> > Just today I was discussing with @Wirone this ticket (in context of his question about not always understanding how to kill mutants). Is it a coinsidence? :)
> 
> No, it's not a coincidence, after our conversation I [asked](https://discord.com/channels/1090672946370052166/1091081297759309926/1226874628589228042) @ttomdewit what's the state of this, it seems like it sparked some reaction 🙂.

It did indeed! @theofidry and I chatted a bit in DM over Discord!

Let me know if I need to do anything to get this in, and probably no sooner than June I will look into documenting mutants one PR at a time (no promises at the moment, not doing very well mentally).

theofidry:
>I like how Psalm adds short links to the output with found issues, and you can open it and read explanations. This is something I would like to have for Infection too (probably integrated with Playground to show "killing tests" there).

From seeing fast evolving docs from Rector, that did that approach, I have mixed feelings about it.

_Ideally_, this would be done with the links bound to the released version, and the doc is versioned too. For example if you had the following code:

```
<<<REMEMDY

Short explanation.

See more at https://docs.infection.com/remedies/<mutant-name>

REMEDY;
```

When building infection, you would have the shipped code updated to point to a specific version:

```
<<<REMEMDY

Short explanation.

See more at https://docs.infection.com/v0.18/remedies/<mutant-name>

REMEDY;
```

And when accessing the docs you could easily jump from a version to another. Also note that the version the code is bound to would be the matching minor version to allow us to update the docs still.

But I think this would be quite a lot of work... So meanwhile I guess even if the remedy description is a bit lengthy it is fine. Or we have to live with the fact that it points to the latest updated doc which will be updated hence if using an old version of infection it may have dead links.

Issue #1451, opened 2020-12-03
Reduce coverage runs when using `forceCoversAnnotation="true"` + `--diff-git-filter`?


Ocramius:
While talking with a team that uses `--diff-git-filter` and `forceCoversAnnotation="true"` on their codebase, we realized that runtime can be further slimmed down by intersecting `--diff-git-filter` results with the respective `@covers` annotations and respective tests.

In doing so, mutation suite runs that require full test suite coverage right now could run against just a couple changed files, at a massive reduction of wasted CPU/Memory/IO time.

This could also probably be done through a plugin system, which (sorry, my bad) is mostly waiting for my feedback.

sanmai:
How does `forceCoversAnnotation="true"` play here?

I see that we can include only changed files while generating the coverage report:
```xml
    <coverage>
        <include includeUncoveredFilesInCodeCoverageReport="false">
            <file>src/File/Changed/InTheCommit.php</file>
            <file>src/File/Changed/InAnotherCommit.php</file>
        </include>
    </coverage>
```
But we might have it pre-generated for us, and then there's no benefit.

Ocramius:
This could play out as following:

 1. we know that `Foo.php` and `Bar.php` were modified
 2. we know that `FooTest.php` has `/** @covers Foo */`
 3. we know that `BarTest.php` does **NOT EXIST** or **DOES NOT HAVE** `/** @covers Bar */`
 4. we can now decide to run only `FooTest.php` when generating coverage

theofidry:
@Ocramius if `BarTest` does not exists then `Bar` is not covered in which case it could be skipped with `--only-covered`. The PHPUnit annotation is only for forcing the end-user to apply them but it might be:

- Wrong to force it on our side: e.g. someone using the PHPUnit bridge with its automatic @covers
- Someone not interested in relying on the coverage report for running the mutations

So whilst I definitely encourage the `forceCoversAnnotation` usage, especially since there is a PHP-CS-Fixer rule to add `@coversNothing` if no `@covers` is found, I think forcing it via Infection might be too disrupting and opinionated

Ocramius:
No enforcement: just upfront filtering when it is already enforced by config

theofidry:
Oh you mean skipping the executing of the tests that don't have coverage information in the initial test run?

Ocramius:
Correcto: we just use the additional information about the test suite config to reduce execution scope as much as possible.

No change for the end-user.

[...]

PR #1919, opened 2024-03-17
Reduce "Initial Tests Run" stage time by running only covering test files

maks-rafalko:
This PR:

- [x] Adds new feature to speed up "Initial Tests Run" stage and as a result - speed up Infection execution by a lot
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/259

Originally requested here: https://github.com/infection/infection/issues/1451

## Idea explanation

If we have X tests files and run Infection with for example `infection --git-diff-lines` and only 1 file is updated/added - it does not make sense to run all the X tests to generate coverage if we know that only 1 file of X covers the source code file.

In practice, PHPUnit even has different methods that help define which test file covers which class:

- [`requireCoverageMetadata`](https://docs.phpunit.de/en/11.0/configuration.html#the-requirecoveragemetadata-attribute) configuration attribute (previously `forceCoversAnnotation="true"`)
- [`CoversClass`](https://docs.phpunit.de/en/11.0/attributes.html#code-coverage) attribute / [`@covers`](https://docs.phpunit.de/en/11.0/annotations.html#covers) annotation and others

With the following code:

```php
use use PHPUnit\Framework\Attributes\CoversClass;

#[CoversClass(Calculator::class)]
class CalculatorTest extends TestCase {}
```

we can say that `Calculator` class is covered by `CalculatorTest` test class, and vice-versa. So, from Infection point of view, when we use `--git-diff-lines` and someone creates PR with updated `Calculator` class, we can run only `CalculatorTest` and skip other X test files (as they anyway don't add any useful coverage), saving a lot of time.

## Things to note

- Implemented feature doesn't change default Infection behavior and can be enabled by adding `--map-source-class-to-test`, so it's opt-in
- implementation is intentionally simple but already gives a huge win, see below. But I designed it in such a way that it can be extended to map source class to test class with more complex strategies

## Example with PHP-CS-Fixer

After maintainers of PHP-CS-Fixer [decided to integrate Infeciton](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/7874) into their pipeline, I was analysing why "Initial Tests Run" stage takes so much time when even only 1 file is changed in PR.

It turned out, that PHP-CS-Fixer has ~24,000 unit tests, that all are executed when we change 1 source file. 

At the same time, PHP-CS-Fixer has `@covers` annotation on _all test classes_, and we can 100% map which test class covers which source class, and run only covering test classes instead of all ~24,000 tests.

Before:

```
FAST_LINT_TEST_CASES=1 infection.phar --git-diff-lines

    ____      ____          __  _

#StandWithUkraine

Infection - PHP Mutation Testing Framework version dev-master

Running initial test suite...

PHPUnit version: 10.5.13

 24768 [============================] 4 mins
```

After:

```
FAST_LINT_TEST_CASES=1 infection.phar --git-diff-lines --map-source-class-to-test

    ____      ____          __  _

#StandWithUkraine

Infection - PHP Mutation Testing Framework version dev-master

Running initial test suite...

PHPUnit version: 10.5.13

   41 [============================] 12 secs
```

Difference:

```diff
- 24768 [============================] 4 mins
+    41 [============================] 12 secs
```

So, for a simple PR with 1 file chaged, now "Initital Tests Run" runs 41 tests by 12 seconds instead of 24,768 tests by 4 minutes. 

## Technical comments

I said the current implementation is intentionally simple, meaning the following: currently, we are not finding which test class covers particular changed file by finding all the `CoversClass` attributes / `@covers` annotations, but just make an assumption that if you have `Calculator` class, then test that covers it is named `CalculatorTest`. 

Actually this is true for Infection, this is how we map Source Class to Test Class:

https://github.com/infection/infection/blob/eb4842d0658b91f5805ef78333a87ed805263ccb/tests/phpunit/AutoReview/SourceTestClassNameScheme.php#L57-L64

and this is also true for PHP-CS-Fixer, they also have such mapping from Source Class to Test Class:

https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/3b3bfe90fe8d5d920328a1f36288d2dfacf5d6b3/tests/AutoReview/ProjectCodeTest.php#L73-L78

Since we now know that `Calculator` class is covered by `CalculatorTest` test class, we can very simply build a `--filter` option for "Initial Tests Run" stage, just by configuring regex:

```
--filter='CalculatorTest|SomeOtherClassTest'
```

I've named this "strategy" of mapping source class to test class as `simple`. Later, if needed, we can implement more complex non-standard strategies, e.g. the one that will phisically parse all files from `tests` folder and find those classes that have `CoversClass(Calculator::class)`, if we need more strict mapping.

But I intentionally didn't do it at this point.


src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
```diff
@@ -57,21 +60,34 @@ final class ArgumentsAndOptionsBuilder implements CommandLineArgumentsAndOptions
...
+            && $this->mapSourceClassToTestStrategy !== null
+            && !in_array('--filter', $options, true)) {
+            $options[] = '--filter';
+
+            $options[] = implode(
+                '|',
+                array_map(
+                    static fn (SplFileInfo $sourceFile): string => sprintf('%sTest', $sourceFile->getBasename('.' . $sourceFile->getExtension())),
+                    $this->filteredSourceFilesToMutate,
+                )
```
maks-rafalko, line 90: These 10 lines actually the real implementation of the feature that needs attention.

as I wrote in the description, this is a `simple` strategy that maps `SourceClass` to `SourceClassTest`

src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
```diff
@@ -57,21 +60,34 @@ final class ArgumentsAndOptionsBuilder implements CommandLineArgumentsAndOptions
...
-        ];
+        $options = $this->prepareArguments($configPath, $extraOptions);
 
-        if ($extraOptions !== '') {
-            $options = array_merge(
-                $options,
-                array_map(static fn ($option): string => '--' . $option, explode(' --', ltrim($extraOptions, '-')))
+        if ($this->filteredSourceFilesToMutate !== []
+            && $this->mapSourceClassToTestStrategy !== null
+            && !in_array('--filter', $options, true)) {
```
theofidry, line 82:
```suggestion
            && !in_array('--filter', $options, true)
          ) {
```

?

src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
```diff
@@ -57,21 +60,34 @@ final class ArgumentsAndOptionsBuilder implements CommandLineArgumentsAndOptions
...
+            $options[] = '--filter';
+
+            $options[] = implode(
+                '|',
+                array_map(
+                    static fn (SplFileInfo $sourceFile): string => sprintf('%sTest', $sourceFile->getBasename('.' . $sourceFile->getExtension())),
```
theofidry, line 88: What do you think about extracting this piece in a dedicated method, e.g. `guessTestFileName()`?

src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
```diff
@@ -123,6 +140,26 @@ public function buildForMutant(string $configPath, string $extraOptions, array $
...
+    {
+        $options = [
+            '--configuration',
+            $configPath,
+        ];
+
+        if ($extraOptions !== '') {
+            $options = array_merge(
+                $options,
+                array_map(static fn ($option): string => '--' . $option, explode(' --', ltrim($extraOptions, '-')))
```
theofidry, line 156: I would put this on two lines for clarity

src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
```diff
@@ -80,10 +96,11 @@ public function buildForInitialTestsRun(string $configPath, string $extraOptions
 
     /**
      * @param TestLocation[] $tests
+     * @return list<string>
      */
     public function buildForMutant(string $configPath, string $extraOptions, array $tests, string $testFrameworkVersion): array
     {
-        $options = $this->buildForInitialTestsRun($configPath, $extraOptions);
+        $options = $this->prepareArguments($configPath, $extraOptions);
```
theofidry, line 103: what is that change about exactly?

theofidry (approved):

src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
maks-rafalko: previously, `buildForMutant` was using public `buildForInitialTestsRun()` method and on top of it added more arguments/options.

Now, after `buildForInitialTestsRun()` has it's own logic for `--filter`, that should be *only* for "Initital Tests Run" stage, we can't reuse it as before.

So I extracted the common reusable part to `prepareArguments()` private helper method, and reused it in both `buildForInitialTestsRun()` and `buildForMutant()`.

src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
maks-rafalko: done with another name, more related to option name and feature itself

src/TestFramework/MapSourceClassToTestStrategy.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Infection\TestFramework;
+
+/**
+ * @internal
+ */
+final class MapSourceClassToTestStrategy
```
sidz, line 41: We could use enums for this purposes.

And to avoid nullable values technically we could use `NONE` as an option instead of NULL

src/Command/RunCommand.php
```diff
@@ -668,4 +681,31 @@ private function getThreadCount(InputInterface $input): int
...
+        // any string: the argument provided, but only `'simple'` is allowed for now
+        if ($inputValue === false) {
+            return null;
+        }
+
+        if ($inputValue === null) {
+            return MapSourceClassToTestStrategy::SIMPLE;
+        }
+
+        if (is_string($inputValue) && !in_array($inputValue, [MapSourceClassToTestStrategy::SIMPLE], true)) {
```
sidz, line 700: Do we need use in_array with just one item in haystack? simple comparison will work either :)

sidz (approved):

src/Command/RunCommand.php
theofidry: In this case you could even do:

```suggestion
        if (MapSourceClassToTestStrategy::SIMPLE === $inputValue) {
```

alternatively you can skip the `is_string()` as the `in_array()` has the strict parameter type. If it was the wrong type to begin with, we are in the failure scenario in which case the extra "unnecessary" looping doesn't matter performance wise

src/TestFramework/Factory.php
```diff
@@ -146,20 +146,17 @@ public function create(string $adapterName, bool $skipCoverage): TestFrameworkAd
...
             return [];
         }
 
-        /** @var list<string> $filteredPaths */
-        $filteredPaths = array_filter(array_map(
-            static fn (SplFileInfo $file) => $file->getRealPath(),
-            iterator_to_array($this->sourceFileFilter->filter($this->infectionConfig->getSourceFiles()))
-        ));
+        /** @var list<SplFileInfo> $files */
+        $files = iterator_to_array($this->sourceFileFilter->filter($this->infectionConfig->getSourceFiles()));
```
theofidry, line 158: could this change (that is a bit noisy as it is affecting a few method signatures / phpstan / mutations) be extracted?

src/TestFramework/Factory.php
```diff
@@ -146,20 +146,17 @@ public function create(string $adapterName, bool $skipCoverage): TestFrameworkAd
...
         if ($this->sourceFileFilter->getFilters() === []) {
             return [];
         }
 
-        /** @var list<string> $filteredPaths */
-        $filteredPaths = array_filter(array_map(
-            static fn (SplFileInfo $file) => $file->getRealPath(),
-            iterator_to_array($this->sourceFileFilter->filter($this->infectionConfig->getSourceFiles()))
-        ));
+        /** @var list<SplFileInfo> $files */
```
theofidry, line 157: is that phpdoc really necessary?

src/TestFramework/PhpUnit/Adapter/PestAdapterFactory.php
```diff
@@ -70,7 +72,8 @@ public static function create(
         array $sourceDirectories,
         bool $skipCoverage,
         bool $executeOnlyCoveringTestCases = false,
-        array $filteredSourceFilesToMutate = []
```
theofidry, line 73: unrelated, but we could update the CS fixer to append the trailing commas all the time to avoid this sort of diff

src/TestFramework/PhpUnit/Adapter/PestAdapterFactory.php
```diff
@@ -104,7 +110,7 @@ public static function create(
                 $projectDir,
                 new JUnitTestCaseSorter()
             ),
-            new ArgumentsAndOptionsBuilder($executeOnlyCoveringTestCases),
+            new ArgumentsAndOptionsBuilder($executeOnlyCoveringTestCases, $filteredSourceFilesToMutate, $mapSourceClassToTestStrategy),
```
theofidry, line 113: might be worth a line returns to avoid super long lines?

```suggestion
            new ArgumentsAndOptionsBuilder(
                $executeOnlyCoveringTestCases,
                $filteredSourceFilesToMutate,
                $mapSourceClassToTestStrategy,
            ),
```

(same for the other adapters)

src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
```diff
@@ -123,6 +141,34 @@ public function buildForMutant(string $configPath, string $extraOptions, array $
         return $options;
     }
 
+    private function mapSourceClassToTestClass(SplFileInfo $sourceFile): string
```
theofidry, line 144: could be static

src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
```diff
@@ -123,6 +141,34 @@ public function buildForMutant(string $configPath, string $extraOptions, array $
...
+    {
+        return sprintf('%sTest', $sourceFile->getBasename('.' . $sourceFile->getExtension()));
+    }
+
+    /**
+     * @return list<string>
+     */
+    private function prepareArgumentsAndOptions(string $configPath, string $extraOptions): array
```
theofidry, line 152: could be static

tests/phpunit/Configuration/ConfigurationFactoryTest.php
```diff
@@ -143,7 +144,8 @@ public function test_it_can_create_a_configuration(
         bool $expectedShowMutations,
         ?float $expectedMinCoveredMsi,
         array $expectedIgnoreSourceCodeMutatorsMap,
-        bool $inputExecuteOnlyCoveringTestCases
+        bool $inputExecuteOnlyCoveringTestCases,
+        ?string $mapSourceClassToTest,
```
theofidry, line 148: I'm happy with the current change, just mentioning this in case you didn't think of it and you would prefer it:

```suggestion
        ?string $mapSourceClassToTest = MapSourceClassToTestStrategy::SIMPLE,
```

It would allow you to skip the modification of all the provide data sets

theofidry (approved):

src/TestFramework/Factory.php
maks-rafalko: short answer: yes, because deeper in the stack we have return type `iterable<SplFileInfo|Trace>` and here we need to specify it.

In other word - this type was here, and removing of this `@var` produces a phpstan error.

I tried quickly to replace `iterable<SplFileInfo|Trace>`  with `iterable<T>` in the stack - but that produces too many errors that need to be analyzed separately. Something wrong with types across those classes.

src/TestFramework/MapSourceClassToTestStrategy.php
maks-rafalko: Yeah, I thought about it. But intentionally didn't use Enumb because we have all the "enum-like" classes as usual classes with constants. So it would be strange to have somewhere enums and somewhere classes with constants - I decided to be consistent here.

If we decide to use enums - we can replace all the classes at once as a separate PR. Currently, 0 enums are in the code base.

src/Command/RunCommand.php
maks-rafalko: > Do we need use in_array with just one item in haystack? simple comparison will work either :)

I did it on purpose to show that there will be more strategies. Rewrote those to use a method that returns an array.

src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
maks-rafalko: it's not a common practice in Infection's codebase, so I decided to leave it as is. 

If we decide to use static methods every time there is no `$this` in it - let's discuss it and probably integrate a fixer? But I wonder what's the benefit and whether it worth doing it.

src/TestFramework/PhpUnit/Adapter/PestAdapterFactory.php
maks-rafalko: agree, extracted to https://github.com/infection/infection/issues/1920

maks-rafalko:
Wanted to share:



https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/7874#issuecomment-2010894591

PR #1921, opened 2024-03-18
Upgrade PHPUnit from 9.6 to 10.5+

maks-rafalko:
Why not PHPUnit 11? Because it requires PHP ^8.2, while we support ^8.1 yet.

The final goal is to get rid of `symfony/phpunit-bridge` and use native settings to require `#[CoversClass]` instead of bridge's  [`CoverageListener`](https://symfony.com/doc/current/components/phpunit_bridge.html#code-coverage-listener)

- [`requireCoverageMetadata`](https://docs.phpunit.de/en/11.0/configuration.html#the-requirecoveragemetadata-attribute) configuration attribute (previously `forceCoversAnnotation="true"`)
- [`CoversClass`](https://docs.phpunit.de/en/11.0/attributes.html#code-coverage) attribute 

See

- https://github.com/symfony/symfony/issues/49069#issuecomment-1966560507
- https://github.com/symfony/symfony/issues/49069#issuecomment-1969518232


maks-rafalko:
Ok, I think now I'm lost.

I have two things I don't understand really:

1. why the heck `symfony/phpunit-bridge` is even [installed](https://github.com/infection/infection/actions/runs/8335187607/job/22810194116#step:14:55) when I removed it 

image

2. Why should I disable deprecation helper explicitly when I don't even want to enable it (it self-registers automatically), I just want to use native `phpunit/phpunit`

Without the latest commits, we have the following error:

```
1) Infection\Tests\ContainerTest::test_it_can_build_lazy_source_file_data_factory_that_fails_on_use
TypeError: call_user_func(): Argument #1 ($callback) must be a valid callback, class "PHPUnit\Util\ErrorHandler" not found

/home/runner/work/infection/infection/vendor/symfony/phpunit-bridge/DeprecationErrorHandler.php:126
/home/runner/work/infection/infection/vendor/justinrainbow/json-schema/src/JsonSchema/Constraints/Constraint.php:48
```

The latest (top) line in the stacktrace says `phpunit-bridge` is used :exploding_head: 

@sidz probably you have an idea what's going on?

.gitignore
```diff
@@ -1,6 +1,8 @@
 !/.tools/.gitkeep
 !/build/.gitkeep
 !/build/cache/.gitkeep
+.phpunit.autoreview.cache
+.phpunit.cache
```
sanmai, line 5: Consider you can cover all needs with '.phpunit.*'

tests/phpunit/Event/Subscriber/MutationTestingConsoleLoggerSubscriberTest.php
```diff
@@ -226,20 +227,20 @@ public function test_it_outputs_escaped_mutants_when_mutation_testing_is_finishe
...
-                    ],
-                ],
-                [
-                    '<options=bold>0</options=bold> mutations were generated:',
-                ]
-            );
+                    ], $value),
+                    2 => $this->assertSame('<options=bold>0</options=bold> mutations were generated:', $value),
+                    default => 'noop'
+                };
```
sanmai, line 242: It is kinda sad we have to do this on PHPUnit 10.

sanmai (approved):
Spot-checked this and that, all makes sense.

sanmai:
IIRC we use symfony/phpunit-bridge to limit coverage reports to relevant files, so removing it might increase the testing time

maks-rafalko:
> IIRC we use symfony/phpunit-bridge to limit coverage reports to relevant files, so removing it might increase the testing time

sorry, I had to explain the final goal, just added to PR's description. So the goal is to add `#[CoversClass()]` to all the test classes and force it by PHPUnit's config.

This way we will have native support instead of [`CoverageListener`](https://symfony.com/doc/current/components/phpunit_bridge.html#code-coverage-listener).

Currently, phpuni-bridge is broken for PHPUnit 10. And symfony [core members say](https://github.com/symfony/symfony/issues/49069#issuecomment-1969518232) that we need to use PHPUnit's feature to detect deprecations.

So bridge is not needed. At with all the bugs it has with PHPUnit 10+, it's usage is dangerous as it overwrites error handler and conflicts with phpunit's one.

phpunit.xml.dist
```diff
@@ -6,38 +6,30 @@
...
+      <exclude>tests/phpunit/Fixtures</exclude>
+    </testsuite>
+  </testsuites>
 
-            <exclude>tests/phpunit/AutoReview</exclude>
-            <exclude>tests/phpunit/Fixtures</exclude>
-        </testsuite>
-    </testsuites>
-
-    <coverage cacheDirectory="./reports/.coverage-cache">
```
maks-rafalko, line 26: - [ ] check if we still have cache

sidz:
> @sidz probably you have an idea what's going on?

Sorry @maks-rafalko I've missed that you tagged me. As I can see `symfony/phpunit-bridge` is removed now. What was the issue?

maks-rafalko:
Hey @sidz 

There are 2 issues. The first one is that `symfony/phpunit-bridge` is still being installed as show in this [comment](https://github.com/infection/infection/pull/1921#issuecomment-2005375968). There was a recommendation to check if it's related to some composer's deps cache - I will play with it.

But the second issue is more interesting. When there is a phpunit-bridge in vendor, it is somehow automatically used, even though we don't have any listeners enabled from this bundle. And it breaks the builds.

After I tried to manually disable bridge by this commit https://github.com/infection/infection/pull/1921/commits/bdf26593a79f99274fd69372eb32b0a77ba14f65, build passes. But it's strange that we need to disable something when it wasn't enabled :) So I was wondering probably you have an idea what's going on. If not - please ignore.

maks-rafalko:
Ok, the last commit demonstrates the issue, log is here: https://github.com/infection/infection/actions/runs/8347659643/job/22847769363#step:16:34

phpunit.xml.dist
maks-rafalko: good thing, it's now the only one cache on the top level `<phpunit>` tag: https://docs.phpunit.de/en/10.5/configuration.html#the-cachedirectory-attribute

sidz:
@maks-rafalko Could you please upgrade `fidry/makefile` package to at least 0.2.3 or even to 1.0. This package installs symfony/phpunit-bridge ?

maks-rafalko:
> Could you please upgrade `fidry/makefile` package to at least 0.2.3 or even to 1.0.

thanks, that is the reason :+1:

theofidry:
Just for it to not be lost I left this comment there: https://github.com/infection/infection/pull/1923/files#r1530874907

theofidry (commented):
I know this PR is not finished yet, especially with all the PRs that were extracted from it, but I would consider the migration of the configuration file as a separate step (same for migrating to the PHPUnit attributes).

maks-rafalko:
> but I would consider the migration of the configuration file as a separate step

it complains about old format which looks quite weird



> same for migrating to the PHPUnit attributes

agree, planning to do it separately (and there are rector's rules for that)

tests/phpunit/Event/Subscriber/MutationTestingConsoleLoggerSubscriberTest.php
maks-rafalko: integrated community replacement that helps to avoid it

tests/phpunit/WithConsecutive.php
```diff
@@ -0,0 +1,110 @@
...
+        // build callback
+        for ($index = 0; $index < $parametersCount; ++$index) {
+            $result[$index] = Assert::callback(static function ($value) use ($values, $index) {
+                static $map = null;
+                $map ??= $values[$index];
+
+                $expectedArg = array_shift($map);
+
+                if ($expectedArg !== null) {
+                    $expectedArg->evaluate($value);
```
maks-rafalko, line 101: I changed it here a bit to allow working with our cases where we expect `atLeaseOnce()` and asserting only first N from M calls.

phpunit_autoreview.xml
```diff
@@ -1,25 +1,24 @@
...
-         xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/schema/9.2.xsd"
+         xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/schema/10.4.xsd"
          bootstrap="./vendor/autoload.php"
          colors="true"
          executionOrder="random"
          stopOnFailure="true"
          failOnWarning="true"
          failOnRisky="true"
-         verbose="true"
+         cacheDirectory=".phpunit.autoreview.cache"
```
theofidry, line 10: same comments as for the previous one

tests/phpunit/WithConsecutive.php
```diff
@@ -0,0 +1,110 @@
...
+use PHPUnit\Framework\Constraint\IsEqual;
+use RuntimeException;
+
+/**
+ * Usage: ->with(...WithConsecutive::create(...$withCodes))
+ *
+ * Originally copied from https://gist.github.com/oleg-andreyev/85c74dbf022237b03825c7e9f4439303 and modified
+ * See also https://github.com/sebastianbergmann/phpunit/issues/4026#issuecomment-1644072411
+ */
+final class WithConsecutive
```
theofidry, line 52: might be worth extracting in a dedicated PR?

theofidry (approved):

tests/phpunit/WithConsecutive.php
sanmai: Doing it in #1921

sanmai:
Here are stats from #1940:

```
5269 mutations were generated:
    3918 mutants were killed
     147 mutants were configured to be ignored
     561 mutants were not covered by tests
     556 covered mutants were not detected
       1 errors were encountered
       1 syntax errors were encountered
       9 time outs were encountered
      76 mutants required more time than configured
```

And here are from this PR:
```
5269 mutations were generated:
    3877 mutants were killed
     147 mutants were configured to be ignored
     561 mutants were not covered by tests
     557 covered mutants were not detected
      27 errors were encountered
       1 syntax errors were encountered
       6 time outs were encountered
      93 mutants required more time than configured
```
MSI are not affected, however the difference is curious nevertheless.

maks-rafalko:
> might be worth extracting in a dedicated PR?

to be honest I have hard time to understand why.

I do understand why we are extracting unrelated changes, this is absolutely fine. But the new `WithConsequtive` class is **directly** related to this PR. Without upgrading from PHPUnit 9.6 to 10.5 we would not need it?

I think that related changes should be in the same PR/commit, later it's easier to understand why they are done.

---

Thanks for your help and review @sanmai @theofidry @sidz, great improvements!

PR #1923, opened 2024-03-19
Additional PHPUnit 10 compatibility fixes

sanmai:
This PR:

- [x] Introduces more PHPUnit 10 related changes
- [x] Extracted from #1921

I'm looking to reduce the diff from #1921 to the bare minimum so we can investigate it further.

### DIFF
--- a/tests/phpunit/PHPUnitConvecutiveReplacement.php
+++ b/tests/phpunit/PHPUnitConvecutiveReplacement.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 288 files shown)


maks-rafalko (approved):

tests/phpunit/PHPUnitConvecutiveReplacement.php
```diff
@@ -0,0 +1,88 @@
...
+declare(strict_types=1);
+
+namespace Infection\Tests;
+
+use function array_column;
+use function count;
+use PHPUnit\Framework\Constraint\Callback;
+use PHPUnit\Framework\Constraint\Constraint;
+
+trait PHPUnitConvecutiveReplacement
```
theofidry, line 43: - typo: `PHPUnitConvecutiveReplacement` -> `PHPUnitConsecutiveReplacement`?
- why does hthis need to be a trait if the only thing it contains is a static method? This could very well be a class with a private constructor (there is `self::assert*` but that could be fixed by using PHPUnit Assertion directly)
- might warrant some docs? This is a very confusing piece of code IMO

Also might warrant some docs?

PR #1930, opened 2024-03-19
Remove php 8.3 and symfony 7.0 hacks in CI as no longer needed

sidz:
This PR removes PHP 8.3 and Symfony 7 hacks which we setup right after each of them were released to be able to install dependencies which wasn't compatible with latest versions.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -54,14 +54,6 @@ jobs:
         if: matrix.dependencies != 'locked'
         run: composer config --unset platform.php
 
-      - name: Configure for PHP >= 8.3
-        if: "matrix.php-version >= '8.3'"
-        run: composer config platform.php 8.2.99
-
-      - name: Configure minimum-stability
-        if: matrix.symfony-version == '7.*.*'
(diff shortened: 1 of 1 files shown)


maks-rafalko (approved):

.github/workflows/tests.yaml
```diff
@@ -54,14 +54,6 @@ jobs:
         if: matrix.dependencies != 'locked'
         run: composer config --unset platform.php
 
-      - name: Configure for PHP >= 8.3
-        if: "matrix.php-version >= '8.3'"
-        run: composer config platform.php 8.2.99
-
-      - name: Configure minimum-stability
```
theofidry, line 61: maybe we should instead configure it for the future version (PHP 8.4 and sf 8.0)?

sidz: We could revert this change once we introduce support of new versions IMO

theofidry: If we remember it :P

sidz: I did it a couple of times already ;)

PR #1937, opened 2024-03-20
Replace PHPUnitConvecutiveReplacement with WithConsecutive

sanmai:
Another chunk from #1921


src/Logger/GitHubAnnotationsLogger.php
```diff
@@ -65,7 +65,7 @@ public function getLogLines(): array
 
 {$escapedExecutionResult->getMutantDiff()}
 TEXT
-            ,
+                ,
```
sanmai, line 68: These should not be here.

PR #1943, opened 2024-03-22
Upgrade phpstan to the latest version

maks-rafalko:
I need to create a custom Rector rule, but rector is out of date.

I can't update rector as it requires new phpstan. So here is an update of phpstan.

There are about ~180 new errors which I'm not ready to fix atm, added to a baseline.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -2872,16 +2872,16 @@
         },
         {
             "name": "phpstan/phpstan",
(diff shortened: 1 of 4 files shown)


sanmai (approved):

maks-rafalko:
thank you @sanmai

PR #1944, opened 2024-03-22
Upgrade rector to the latest version

maks-rafalko:
I need to create a custom Rector rule, but rector is out of date.

So here is an update of rector first.


maks-rafalko:
> Could you please add tests to make sure this change works as expected?

unrelated, but glad you are not sleeping

sanmai (approved):

maks-rafalko:
thank you @sanmai

PR #1945, opened 2024-03-22
Require coverage metadata in phpunit.xml.dist

maks-rafalko:
Require coverage metadata in `phpunit.xml.dist`.

If any test class will not have `CoversClass` attribute, tests will fail (test becomes risky, [see proof](https://github.com/infection/infection/actions/runs/8397593208/job/23001150322?pr=1945#step:12:88))

----

**Why is it needed?**

- because it's a good practice (e.g. read https://localheinz.com/articles/2023/02/22/documenting-the-system-under-test-in-phpunit/)
- to introduce `--map-source-class-to-test` into Infection

**Does build fail if attribute isn't added to a test class?**

Yes, phpunit will return non-zero exit code

In the next PR, I will introduce rector rule that I've used to automatically add `CoversClass` in this PR, and which will be used on CI here at Infection.


sanmai (approved):
:100:

maks-rafalko:
Thank you @sanmai

PR #1947, opened 2024-03-23
Install `pcntl` in docker image as `make autoreview` fails with absent `SIGINT` const

maks-rafalko:
```bash
docker compose run php81 make autoreview
```

didn't work before, because Psalm was failing, complaining `\SIGINT` doesn't exist.

Now it works.


sanmai (approved):

maks-rafalko:
thank you @sanmai

Issue #1948, opened 2024-03-24
Gitlab logger errors when run outside git repository


darthf1:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.27.11
| Test Framework version | PHPUnit 10.5.14
| PHP version | 8.3.4
| Platform    | Linux
| Github Repo | -

Hi!

I would like to try out the Gitlab logger, by adding the following in my `infection.json`:
```json
{
  "$schema": "vendor/infection/infection/resources/schema.json",
  "logs": {
      "gitlab": "var/cache/infection/infection-gitlab.json"
  }
}
```

When the run is finished, and the logs are being generated, I get the following error: 
```
Generated Reports:
         - /home/www/app/var/cache/infection/infection-gitlab.json

Please note that some mutants will inevitably be harmless (i.e. false positives).
fatal: not a git repository (or any parent up to mount point /home/www)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).

In ExecException.php line 9:
                    
  An error occured  
                    

run [--test-framework TEST-FRAMEWORK] [--test-framework-options TEST-FRAMEWORK-OPTIONS] [-j|--threads THREADS] [--only-covered] [-s|--show-mutations] [--no-progress] [--force-progress] [-c|--configuration CONFIGURATION] [--coverage COVERAGE] [--mutators MUTATORS] [--filter FILTER] [--formatter FORMATTER] [--git-diff-filter GIT-DIFF-FILTER] [--git-diff-lines] [--git-diff-base GIT-DIFF-BASE] [--logger-github [LOGGER-GITHUB]] [--map-source-class-to-test [MAP-SOURCE-CLASS-TO-TEST]] [--logger-gitlab [LOGGER-GITLAB]] [--logger-html LOGGER-HTML] [--noop] [--only-covering-test-cases] [--min-msi MIN-MSI] [--min-covered-msi MIN-COVERED-MSI] [--log-verbosity LOG-VERBOSITY] [--initial-tests-php-options INITIAL-TESTS-PHP-OPTIONS] [--skip-initial-tests] [--ignore-msi-with-no-mutations] [--debug] [--dry-run]

Script XDEBUG_MODE=off INFECTION_ALLOW_XDEBUG=0 vendor/bin/infection --threads=$(nproc) --no-progress handling the infection event returned with error code 1
```

[...]

PR #1949, opened 2024-03-24
feat: use CI variables to detect project path

darthf1:
Fixes https://github.com/infection/infection/issues/1948.

This fixes the following issue on Gitlab CI:

```
Generated Reports:
         - /builds/xxx/xxx/var/cache/infection/infection-gitlab.json
fatal: detected dubious ownership in repository at '/builds/xxx/xxx'
To add an exception for this directory, call:
	git config --global --add safe.directory /builds/xxx/xxx
In ExecException.php line 9:
                    
  An error occured  
                    
run [--test-framework TEST-FRAMEWORK] [--test-framework-options TEST-FRAMEWORK-OPTIONS] [-j|--threads THREADS] [--only-covered] [-s|--show-mutations] [--no-progress] [--force-progress] [-c|--configuration CONFIGURATION] [--coverage COVERAGE] [--mutators MUTATORS] [--filter FILTER] [--formatter FORMATTER] [--git-diff-filter GIT-DIFF-FILTER] [--git-diff-lines] [--git-diff-base GIT-DIFF-BASE] [--logger-github [LOGGER-GITHUB]] [--map-source-class-to-test [MAP-SOURCE-CLASS-TO-TEST]] [--logger-gitlab [LOGGER-GITLAB]] [--logger-html LOGGER-HTML] [--noop] [--only-covering-test-cases] [--min-msi MIN-MSI] [--min-covered-msi MIN-COVERED-MSI] [--log-verbosity LOG-VERBOSITY] [--initial-tests-php-options INITIAL-TESTS-PHP-OPTIONS] [--skip-initial-tests] [--ignore-msi-with-no-mutations] [--debug] [--dry-run]
```

By utilizing the `CI_PROJECT_DIR` ([Gitlab](https://docs.gitlab.com/ee/ci/variables/predefined_variables.html)) and the `GITHUB_WORKSPACE` ([Github](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)) ENV variables to get the project root dir, the `git` command does not have to be invoked.

This does not resolve the issue with running Infection with one of these loggers both outside  the CI and outside a GIT repository, but since my project is in a mono-repo and the pathmapping is different inside my Docker container, and the value of these loggers is inside a CI environment; I'm happy with this change.

I tested this change in a CI runner in my private Gitlab project. I have not tested this in CI runner in a Github project, as I don't have any. But the change seems pretty straight forward.


darthf1:
All green @maks-rafalko  :D

tests/phpunit/Logger/GitHubAnnotationsLoggerTest.php
```diff
@@ -71,4 +93,44 @@ public static function metricsProvider(): iterable
...
+                        @@ @@
+
+                        - echo 'original';
+                        + echo 'escaped#0';
+
+                        DIFF,
+                )),
+                'a1b2c3',
+                MutatorName::getName(For_::class),
+                '/my/project/dir/foo/bar',
```
sanmai, line 120: If you extend `CreateMetricsCalculator` to include a static variable for a path prefix, you can avoid adding your own ResultsCollector.

The test will look like this:
```php
    public function test_it_logs_correctly_using_github_workspace(): void {
        \Safe\putenv('GITHUB_WORKSPACE=/my/project/dir');

        self::setOriginalFilePrefix('/my/project/dir/');

        $resultsCollector = self::createCompleteResultsCollector();

        $logger = new GitHubAnnotationsLogger($resultsCollector);

        $this->assertStringContainsString('warning file=foo/bar', $logger->getLogLines()[0]);
    }
```

tests/phpunit/Logger/GitLabCodeQualityLoggerTest.php
```diff
@@ -128,6 +148,60 @@ public static function metricsProvider(): iterable
         ];
     }
 
+    public function test_it_logs_correctly_with_ci_project_dir(): void
+    {
+        \Safe\putenv('CI_PROJECT_DIR=/my/project/dir');
+
+        $resultsCollector = new ResultsCollector();
```
sanmai, line 155: Here you can pull a similar trick as well.

tests/phpunit/Logger/GitHubAnnotationsLoggerTest.php
darthf1: Yes, much better (Y)

darthf1:
Ready @sanmai :)

maks-rafalko (approved):
I wonder if we can break the fix introduced here: https://github.com/infection/infection/pull/1576

I think no, as `GITHUB_WORKSPACE` should return the root folder (exactly what we need), but just in case..

darthf1:
> I wonder if we can break the fix introduced here: https://github.com/infection/infection/pull/1576
> 
> I think no, as `GITHUB_WORKSPACE` should return the root folder (exactly what we need), but just in case..
> 
> ----
> 
> Let's wait for @sanmai's review yet

Yeah, when I read the Github docs, they should both resolve to the same directory.  But I have no practical experience.

tests/phpunit/Logger/GitHubAnnotationsLoggerTest.php
```diff
@@ -71,4 +87,16 @@ public static function metricsProvider(): iterable
             ],
         ];
     }
+
+    public function test_it_logs_correctly_with_ci_github_workspace(): void
+    {
+        \Safe\putenv('GITHUB_WORKSPACE=/my/project/dir');
+        self::$pathPrefix = '/my/project/dir/';
```
sanmai, line 94: What do you think about resetting the path prefix during the tear down?

I couldn't find if it breaks anything now, but I can imagine it could get into way when adding another test.

And please don't force-push things: we are going to squash-merge everything anyway.

darthf1: Since tests are executed in a random order, shouldn't it already break then?

I can do the change later today, not behind a computer right now.

maks-rafalko:
Could you please also merge `master` into this branch? Should fix MT build check

darthf1:
Comments addressed; all green.

maks-rafalko (approved):
Sill :+1:

sanmai (approved):
Nice! 💯

maks-rafalko:
Thank you @darthf1

maks-rafalko:
Released https://github.com/infection/infection/releases/tag/0.28.1

PR #1954, opened 2024-04-11
test: Force mutators to include remedies

theofidry:
This PR is a rebased version of #1907.

It currently skips all but one mutator to set a baseline. Existing mutators should no longer be skipped in the future and this check ensures that new mutators will require a remedy to be documented.

There is more to do with remedies to, for example some ideas of https://github.com/infection/infection/pull/1907#issuecomment-2043212338, but I think it is vastly out of scope of this PR and should be considered at a later time.

Closes #1907.


maks-rafalko (approved):

sidz (approved):
:+1:

theofidry:
Thank you @ttomdewit!

ttomdewit:
> Thank you @ttomdewit!

Thank you! 🥰

PR #1961, opened 2024-05-12
Upgrade Rector and fix new issues

maks-rafalko:


src/TestFramework/Coverage/CoveredTraceProvider.php
```diff
@@ -46,10 +45,6 @@
  */
 final class CoveredTraceProvider implements TraceProvider
 {
-    /**
-     * @param PhpUnitXmlCoverageTraceProvider|TraceProvider $primaryTraceProvider
-     * @param BufferedSourceFileFilter|FileFilter $bufferedFilter
```
sanmai, line 51: This is unfortunate because without this doc comment one can no longer jump into the implementation with a single click.

maks-rafalko: ok, reverted. Reported https://github.com/rectorphp/rector/issues/8633

let's see if this is a bug or if there are any arguments supporting this behavior

maks-rafalko:
Ok, this is a bug from Rector, I will merge with the exclusion and later we will remove the code added in this PR.

However, please read the issue, these phpdocs will be removed with another set in Rector - `deadCode`: https://getrector.com/demo/3c272224-96a4-4e16-8414-6d01bff60d65

PR #1962, opened 2024-05-12
Enable Rector's `AddCoversClassAttributeRector` rule

maks-rafalko:
- Rule: https://github.com/rectorphp/rector-phpunit/blob/main/docs/rector_rules_overview.md#addcoversclassattributerector
- see my explanation why it's needed https://github.com/rectorphp/rector-phpunit/pull/319#issue-2204576592

Actually, I have already applied this rule previously for Infection's codebase when created it locally. Now it's enabled and will fix all future tests after new Rector 1.0.5 is released, so we can use it from the Rector's core.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3517,16 +3517,16 @@
         },
         {
             "name": "rector/rector",
-            "version": "1.0.3",
+            "version": "1.0.5",
             "source": {
                 "type": "git",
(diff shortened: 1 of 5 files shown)


sanmai (approved):
:100:

sidz (approved):

PR #1963, opened 2024-05-14
Move `Infection\Mutator\Mutator` to a separate package

maks-rafalko:
This is a follow up PR to support custom mutators, see #1686 

In order to provide our users (developers who are going to write custom mutators) a convenient interface, we need to extract it out of Infection core.

Thus, here is a new package: https://github.com/infection/mutator

Now, `Mutator` interface is there, and if someone decides to create a package with custom mutators, they will rely on this highly abstract and stable `infection/mutator` package, rather than relying on instable (in terms of Package Design) `infection/infection`.

On the other side, `infection/infection` now depends on `infection/mutator` as well.

test-framework-adapter(1)

If you need to understand package design principles and why I'm doing it, see

- EN https://github.com/infection/infection/issues/922
- EN https://link.springer.com/book/10.1007/978-1-4842-4119-6
- RU https://www.youtube.com/watch?v=JKt7hF_nxBM


theofidry (approved):

sanmai (commented):
Nice!

PR #1966, opened 2024-05-18
Attempt to improve test time adding

### LINKED ISSUE #1965: Mutations should be skipped base on individual time, not total time

Mutations for a given line of code seem to be skipped based on the _total_ amount of time taken for all tests that cover that line. In my case, I have a few lines of code that are covered by nearly every test case; so while each test takes 100 ms, mutations are skipped entirely for those lines because 100 ms x 300 tests = 30 s, which is greater than the timeout.

Of course I can just increase the timeout as a workaround, but this is still not ideal behavior.

Mutations should only be skipped base on the time required (maybe maximum time required?) to run a single test case per line.

**Additional context**

As discussed in Discord: https://discord.com/channels/767914934016802818/771020567343136831/1241042229473316934

CC @sanmai who authored this code

sanmai:
This PR:

- [ ] Attempts to fix #1965

It didn't work, but I figured I'll leave some breadcrumbs for the future explorer.

### DIFF
--- a/src/TestFramework/Coverage/JUnit/JUnitTestExecutionInfoAdder.php
+++ b/src/TestFramework/Coverage/JUnit/JUnitTestExecutionInfoAdder.php
@@ -103,7 +103,7 @@ class JUnitTestExecutionInfoAdder
         return new TestLocation(
             $test->getMethod(),
             $testFileData->path,
-            $testFileData->time,
+            $test->getExecutionTime() ?? $testFileData->time,
         );
     }
 }
--- a/src/TestFramework/Coverage/JUnit/JUnitTestCaseTimeAdder.php
+++ b/src/TestFramework/Coverage/JUnit/JUnitTestCaseTimeAdder.php
@@ -71,21 +71,12 @@ final class JUnitTestCaseTimeAdder
 
         foreach ($this->tests as $testLocation) {
             $methodName = $testLocation->getMethod();
-            $methodSeparatorPos = strpos($methodName, '::');
 
-            if ($methodSeparatorPos === false) {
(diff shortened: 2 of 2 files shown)


src/TestFramework/Coverage/JUnit/JUnitTestExecutionInfoAdder.php
```diff
@@ -103,7 +103,7 @@ private function createCompleteTestLocation(TestLocation $test): TestLocation
         return new TestLocation(
             $test->getMethod(),
             $testFileData->path,
-            $testFileData->time,
+            $test->getExecutionTime() ?? $testFileData->time,
```
sanmai, line 106: There are per-test timings in the JUnit report I had lying around, so in theory it should be possible to get a more accurate numbers.

maks-rafalko: could you please explain the idea of this PR? Sorry, without debugging I think I don't get it

sanmai: The idea: https://github.com/infection/infection/issues/1965#issuecomment-2119078924

staabm: Does this mean we know timing from previous runs based on this junit report?

sanmai: Previous runs in a sense of unit test runs, yes. Infection runs, no.

PR #1969, opened 2024-05-20
Custom mutator generator

maks-rafalko:
This is a follow up for https://github.com/infection/infection/pull/1686, docs: https://github.com/infection/site/pull/262

### Short description

This PR adds new CLI command to generate custom mutator in a target project, as well as a test for convenient development / testing of brand new mutators.

### Details

A lot of things needs to be explained here. After hours of thinking / experimenting how it should be, I came up with this solution that helps developers to generate a "skeleton" of new mutator and _a test to for convenient development_.

Things to note:

- without this PR, it's already possible to create custom mutators - everything is ready on Infection side and [e2e test](https://github.com/infection/infection/tree/80a15797ad47ad54658a24bdfa2d509964269988/tests/e2e/Custom_Mutator) confirms it and there is even a draft version of the docs: https://github.com/infection/site/blob/18ecec970795d0f701513a29f8765429ff925331/src/guide/custom-mutators.md
- this PR significantly improves **DX**. The real goal of this change is to increase the speed of custom mutator creation, improve convenience and to allow to test custom mutator in an "Infection runtime" (see below)

This PR adds a couple of things:

1. adds `bin/infection custom-mutator` CLI command to generate skeleton for new custom mutator - mutator itself and test file
2. moves `BaseMutatorTestCase.php` (and other related classes) to `src` from `tests`, so that developers can test their mutators in "Infection runtime". By this, I mean that developers will not blindly create new mutator class, but they will mutate the code in tests and see which mutants (new versions of the source code) are produced. Exactly the same as we do it in out unit tests

### `bin/infection custom-mutator`

This idea is taken from Rector's `custom-rule` command, see https://getrector.com/documentation/custom-rule ([implementation](https://github.com/rectorphp/rector-src/blob/b8847be9d485ec1b3cde5ee5e099ab3626479eb7/src/Console/Command/CustomRuleCommand.php)).

What it does is basically generates 2 files - mutator and test for it. See `__Name__.php` and `__Name__Test.php` files in the diff to understand how it works. `__Name__` is replaced with the real name of the new mutator during creation of the files.

### Moving  `BaseMutatorTestCase.php` from `tests` to `src` - why?

This is needed to allow users to write the following tests:

https://github.com/infection/infection/blob/80a15797ad47ad54658a24bdfa2d509964269988/tests/phpunit/Mutator/Arithmetic/PlusTest.php#L58-L73

But to achieve it, we need to do really a lot of things:

- parse the source code into AST
- configure our visitors ([parent connector, reflection visitor, name resolver, ignorers, etc.](https://github.com/infection/site/blob/18ecec970795d0f701513a29f8765429ff925331/src/guide/custom-mutators.md#understanding-node-attributes))
- generate N mutants from this code
- assert

Without our base class provided for the end developers, it will be very difficult to write all of this logic themselves. So we basically say for them: "look here, we have Infection runtime which you can use to test your new custom mutator - if you want - you can use it."

In order to use "Infection runtime" in custom mutator's tests, developers will need to require `infection/infection`. You can ask: "wait, but we [extracted](https://github.com/infection/infection/pull/1963) `infection/mutator` exactly to **not** depend on `infection/infection`?". And this is still the case!

The project that will implement a custom mutator still needs to depend on `infection/mutator` in `require` section. But if they want to use our `custom-mutator` CLI generator and our base test class, they can require `infection/infection` to the `require-dev` section. From dependencies POV, they still depend only on `infection/mutator` and `infection/infection` is a dev dependency only for test. So no problem here.

Actually, Rector does the same thing. They have their base test class in `src` and package it with the source code: 

- https://github.com/rectorphp/rector-src/blob/b8847be9d485ec1b3cde5ee5e099ab3626479eb7/templates/custom-rule/utils/rector/tests/Rector/__Name__/__Name__Test.php#L10
- https://github.com/rectorphp/rector-src/blob/b8847be9d485ec1b3cde5ee5e099ab3626479eb7/src/Testing/PHPUnit/AbstractRectorTestCase.php#L35

I've already created documentation for writing Custom Mutators - https://github.com/infection/site/pull/261 - and after this PR it will be greatly simplified from developer's point of view, improving DX. Docs PR supporting this change: https://github.com/infection/site/pull/262

Example of usage:

image

Generated files:

image


src/Testing/SimpleMutation.php
maks-rafalko, line 1: this file was just moved from `tests` to `src`, no changes (despite github's diff)

src/Testing/SimpleMutationsCollectorVisitor.php
maks-rafalko, line 1: this file was just moved from `tests` to `src`, no changes (despite github's diff)

theofidry:
spend the day in some docker shit, will wait tomorrow / after tomorrow for a review 😅

maks-rafalko:
ping @theofidry

src/Command/CustomMutatorCommand.php
```diff
@@ -0,0 +1,120 @@
...
+                }
+
+                return $answer;
+            });
+            $mutator = $io->askQuestion(
+                $question,
+            );
+        }
+
+        $mutator = ucfirst((string) $mutator);
```
theofidry, line 81: provided you move the question to `::interact()`, you could then move `$io->getInput()->getArgument('Mutator name')` and this `ucfirst()` in a dedicated method `getMutatorName(IO $io): string`, where if there is further validation to do we can do there too

src/Command/CustomMutatorCommand.php
```diff
@@ -0,0 +1,120 @@
...
+
+        // find all files in templates directory
+        $finder = Finder::create()
```
theofidry, line 84: is there going to be more templates? It otherwise feels like it could be constants

src/Command/CustomMutatorCommand.php
```diff
@@ -0,0 +1,120 @@
...
+            ->in(__DIR__ . '/../CustomMutator/templates');
+
+        $currentDirectory = getcwd();
+
+        $generatedFilePaths = [];
+
+        $fileInfos = iterator_to_array($finder->getIterator());
+        $fileSystem = $this->getApplication()->getContainer()->getFileSystem();
+
+        foreach ($fileInfos as $fileInfo) {
```
theofidry, line 95: if you only iterate over them you don't need to do the `iterator_to_array()`, a `Finder` is `Traversable`

src/Command/CustomMutatorCommand.php
```diff
@@ -0,0 +1,120 @@
...
+            $generatedFilePaths[] = $newFilePath;
+        }
+
+        $io->title('Generated files');
+        $io->listing($generatedFilePaths);
+        $io->success(
+            sprintf('Base classes for the "%s" mutator were created. Complee the missing parts inside them.', $mutator),
+        );
+
+        return true;
```
theofidry, line 113:
```suggestion
        return self::SUCCESS;
```

src/Command/CustomMutatorCommand.php
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    private function replaceNameVariable(string $rectorName, string $contents): string
```
theofidry, line 116:
```suggestion
    private static function replaceNameVariable(string $rectorName, string $contents): string
```
?

src/Command/CustomMutatorCommand.php
```diff
@@ -0,0 +1,120 @@
...
+        foreach ($fileInfos as $fileInfo) {
+            // replace __Name__ with $mutator
+            $newContent = $this->replaceNameVariable($mutator, $fileInfo->getContents());
+            $replacedNamePath = $this->replaceNameVariable($mutator, $fileInfo->getRelativePathname());
+
+            $newFilePath = $currentDirectory . '/src/Mutator/' . $replacedNamePath;
+
+            $fileSystem->dumpFile($newFilePath, $newContent);
+
+            $generatedFilePaths[] = $newFilePath;
```
theofidry, line 104: You could move this block in a dedicated method for readability. Since those are templates, you could call it `renderTemplate(string $mutatorName, SplFileInfo $template, string $currentDirectory): string`

src/Command/CustomMutatorCommand.php
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        $io->title('Generated files');
+        $io->listing($generatedFilePaths);
+        $io->success(
+            sprintf('Base classes for the "%s" mutator were created. Complee the missing parts inside them.', $mutator),
```
theofidry, line 110: typo:

```suggestion
            sprintf('Base classes for the "%s" mutator were created. Complete the missing parts inside them.', $mutator),
```

theofidry, line 110:
```suggestion
            sprintf('Base classes for the mutator "%s" were created. Complete the missing parts inside them.', $mutator),
```

src/Command/CustomMutatorCommand.php
theofidry, line 81: On top of my head a class name in PHP must pass the following regex:

```
/^[\p{L}_][\p{L}_\d]*$/
```

(If you think it is worth adding a minimal validation of the name)

maks-rafalko:
It would be also nice if you can go through the docs and try to do it yourself :) just to check if it's clear for you what to do

src/CustomMutator/templates/__Name__.php
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * @return iterable<TODO>
+     */
+    public function mutate(Node $node): iterable
+    {
+        // TODO: update the logic to return mutated nodes
+        yield $node;
+    }
+
+    public static function getDefinition(): Definition
```
theofidry, line 61: shouldn't this be at the top?

src/Testing/BaseMutatorTestCase.php
```diff
@@ -33,21 +33,19 @@
 
 declare(strict_types=1);
 
-namespace Infection\Tests\Mutator;
+namespace Infection\Testing;
 
+use function array_flip;
```
theofidry, line 38: there is more changes that I would do to this class since we make it public (could be done before?). Since I can't comment the non-changed lines I'll add them here:

- inline the type for `#mutator`
- what was the TODO about in the `setUp()`?

src/CustomMutator/templates/__Name__Test.php
```diff
@@ -0,0 +1,111 @@
...
+use PHPUnit\Framework\Attributes\DataProvider;
+
+#[CoversClass(__Name__::class)]
+final class __Name__Test extends BaseMutatorTestCase
+{
+    /**
+     * @param string|string[] $expected
+     */
+    #[DataProvider('mutationsProvider')]
+    public function test_it_can_mutate(string $input, $expected = []): void
```
theofidry, line 50:
```suggestion
    public function test_it_can_mutate(string $input, string|array $expected = []): void
```

theofidry, line 50: maybe this method could be part of the base test case and the provider be declared as abstract?

src/CustomMutator/templates/__Name__Test.php
```diff
@@ -0,0 +1,111 @@
...
+    {
+        $this->doTest($input, $expected);
```
theofidry, line 52: Maybe time to find a more fitting name?

```suggestion
        $this->assertMutatesInput($input, $expected);
```

src/Testing/BaseMutatorTestCase.php
```diff
@@ -118,17 +116,25 @@ final public function doTest(string $inputCode, $expectedCode = [], array $setti
 
     final protected function createMutator(array $settings = []): Mutator
     {
-        $mutatorClassName = SourceTestClassNameScheme::getSourceClassName(static::class);
+        $mutatorClassName = $this->getTestedMutatorClassName();
+
+        $isBuiltinMutator = array_key_exists($mutatorClassName, array_flip(ProfileList::ALL_MUTATORS));
+        $mutatorName = $isBuiltinMutator ? MutatorName::getName($mutatorClassName) : $mutatorClassName;
 
         // TODO: this is a bit ridicule...
```
theofidry, line 124: is it still ridicule? 😄

src/Testing/BaseMutatorTestCase.php
```diff
@@ -189,7 +195,7 @@ private function assertSyntaxIsValid(string $realMutatedCode): void
             $returnCode,
             sprintf(
                 'Mutator %s produces invalid code',
```
theofidry, line 197:
```suggestion
                'The mutator "%s" produced invalid PHP code.',
```

Might be worth tweaking it to print for which output this is the case?

src/Testing/SimpleMutation.php
```diff
@@ -0,0 +1,87 @@
...
+use Infection\PhpParser\MutatedNode;
+use PhpParser\Node;
+
+/**
+ * @internal
+ */
+final class SimpleMutation extends Mutation
+{
+    /**
+     * @param MutatedNode $mutatedNode
```
theofidry, line 49: inline the type?

src/Testing/SimpleMutationsCollectorVisitor.php
```diff
@@ -0,0 +1,93 @@
...
+    public function __construct(
+        private readonly Mutator $mutator,
+        /**
+         * @var Node[]
+         */
+        private readonly array $fileAst,
+    ) {
+    }
+
+    public function leaveNode(Node $node)
```
theofidry, line 65:
```suggestion
    public function leaveNode(Node $node): void
```
?

theofidry (commented):
I had a brief look, couldn't get to try it or check it out locally. It's actually a lot less changes than I expected (from the diff)

src/Command/CustomMutatorCommand.php
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * @internal
+ */
+final class CustomMutatorCommand extends BaseCommand
+{
+    protected function configure(): void
+    {
+        $this
+            ->setName('custom-mutator')
```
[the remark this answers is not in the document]
maks-rafalko: I'm not against. But we have other commands, like

- `describe` 

which doesn't have `verb:noun` structure. Would it be ok from consistency POV to use `make:custom-mutator`?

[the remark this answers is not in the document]
theofidry: Personally I think it's fine, I had a bit of the same dilemma in Box

src/Command/CustomMutatorCommand.php
maks-rafalko: it's our own implementation that works with `bool`, not `int`

src/Testing/SimpleMutationsCollectorVisitor.php
maks-rafalko:
```
  65     Return type (void) of method Infection\Testing\SimpleMutationsCollectorVisitor::leaveNode() should be  
         compatible with return type (array<PhpParser\Node>|int|PhpParser\Node|null) of method                  
         PhpParser\NodeVisitor::leaveNode() 
```

src/Command/CustomMutatorCommand.php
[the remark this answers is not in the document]
maks-rafalko: done

src/Command/CustomMutatorCommand.php
```diff
@@ -0,0 +1,120 @@
...
+            ->setDescription('Creates a custom mutator')
+            ->addArgument('Mutator name', InputArgument::OPTIONAL);
+    }
+
+    protected function executeCommand(IO $io): bool
+    {
+        $mutator = $io->getInput()->getArgument('Mutator name');
+
+        if ($mutator === null) {
+            $question = new Question('What mutator do you wish to create (e.g. `AnyStringToInfectedMutator`)?');
```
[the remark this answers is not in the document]
maks-rafalko: moved to interact(). However it's not that easy.

We have 2 ways how `$mutatorName` is initialized - provided as argument, and as answer to Question Helper.

When it's probided directly as argument (`make:custom-mutator MyMutator`) - we still need to `trim()` and `ucfirst()` and to validate. I've added tests to cover these cases

src/Testing/BaseMutatorTestCase.php
maks-rafalko: I don't know, it was created by you.

We shouldn't create TODOs without the ticket anymore in the future

src/CustomMutator/templates/__Name__Test.php
maks-rafalko: I decided to not do it, since IMO now it's more readable for the end user

maks-rafalko:
@theofidry please review again

maks-rafalko:
New thing: https://infection-php.dev/ast - this will help custom mutator developers to better/easier understand what to use in `Mutator::canMutate()`. The code is clickable after submitting

theofidry (approved):

src/Command/MakeCustomMutatorCommand.php
```diff
@@ -0,0 +1,150 @@
...
+use Symfony\Component\Console\Question\Question;
+use function trim;
+use function ucfirst;
+
+/**
+ * @internal
+ */
+final class MakeCustomMutatorCommand extends BaseCommand
+{
+    private const MUTATOT_NAME_ARGUMENT = 'Mutator name';
```
sidz, line 57:
```suggestion
    private const MUTATOR_NAME_ARGUMENT = 'Mutator name';
```

src/Command/CustomMutatorCommand.php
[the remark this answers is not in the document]
sidz: what about `make:mutator` ? ^_^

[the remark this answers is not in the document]
maks-rafalko: Like it! Renamed.

src/Command/MakeCustomMutatorCommand.php
maks-rafalko: done, thanks

sidz (approved):

maks-rafalko:
thank you @theofidry and @sidz 

@sanmai feel free to add comments, will address in separate PR

PR #1970, opened 2024-05-21
Remove our own `ParentConnectorVisitor` and use `nikic-phpparser`'s one

maks-rafalko:
Follow up for https://github.com/infection/infection/pull/1967

During revision of our visitors, I found that php-parser already have [native parent connector visitor](https://github.com/nikic/PHP-Parser/blob/d327cf2acf22ad3d1fac5a382b7420bb6fd73bb8/lib/PhpParser/NodeVisitor/ParentConnectingVisitor.php), so it doesn't make sense to support our own anymore.


sidz (approved):
:+1:

PR #1973, opened 2024-06-01
Fix usage of custom mutator with bootstrap file

maks-rafalko:
On Infection Playground, for each "playground run" we don't use composer to load files in PSR-4 format, we use [custom very simple PSR-4 compliant autoloader](https://github.com/infection/playground/blob/b88e85d96a0f3b262579c4c3e3c4b307ff508dae/app/infection-builds/Psr4AutoloaderClass.php), which is [configured](https://github.com/infection/playground/blob/b88e85d96a0f3b262579c4c3e3c4b307ff508dae/app/src/Infection/ConfigBuilder.php#L49) to be used in `infection.json5` in `bootstrap` setting.

In `master` branch, Infection tries to `resolveMutators()` **before** loading bootstrap files, thus failing finding custom mutator.

In order to prepare configuration, where custom mutators are resolved, we need to load (`require_once`) custom bootstrap file before mutators are resolved.

This fixes the bug and allows Playground to use custom mutators right in the browser.


tests/phpunit/Configuration/ConfigurationFactoryTest.php
```diff
@@ -936,7 +937,113 @@ public static function valueProvider(): iterable
...
+                ),
+                'config/tmp',
+                new PhpUnit(
+                    'config/phpunit-dir',
+                    'config/phpunit',
+                ),
+                null,
+                null,
+                null,
+                ['@default' => false, 'CustomMutator' => true],
```
maks-rafalko, line 985: here is a trick: we use `CustomMutator` alias, that is only available if `bootstrap` file is executed, making this alias available for class autoloader.

This confirms that custom mutator works with bootstrap

Issue #1810, opened 2023-01-27
[Logs] Add colorful text log for CLI output


Slamdunk:
Hi, I'd like to have a colorful text log for CLI.
I am aware of the precious logs that are already present, like `HTML`, `JSON`, `Summary` and `Stryker`, but they are not very TDD friendly.
This is particularly necessary since Infection removes all indentation, so murators like `ArrayItemRemoval` are very hard to read when hundreds of LoC are compressed in a single line in `text` output.

I'm currently workarounding this with the `text` log, making it a compatible `diff --unified` format and redirecting it to [`diff-so-fancy`](https://github.com/so-fancy/diff-so-fancy), with this little bash script:

<details>
<summary><code>text-to-colorful-diff.sh</code></summary>
<pre>
cat - \
  | sed \
    -e ':a' \
    -e 'N' \
    -e '$!ba' \
    -e 's#) \(\S\+\):\(\S\+\)    \[M\] \(\S\+\)\n\n--- Original\n+++ New\n@@ @@#) Mutant failed: \x1B[1;33m\3\x1B[0m\n\ndiff --git a/\1 b/\1\n--- a/\1\n+++ b/\1\n@@ -\2,1 +\2,1 @@#g' \
  | diff-so-fancy \
  | sed \
    -e ':a' \
    -e 'N' \
    -e '$!ba' \
    -e 's#─\+\x1B\[0m\n##g' \
    -e 's#modified: \S\+\n##g'
</pre>
</details>

### Before

![image](https://user-images.githubusercontent.com/152236/215028845-17c2dd3c-e5c7-402e-bc18-628f5f017af6.png)

### After

![image](https://user-images.githubusercontent.com/152236/215028965-513c4587-513f-4bd8-9cad-9420c15dcc4d.png)

maks-rafalko:
Hello, let me ask a couple of questions to understand how it's going to work

1. Are you talking about text file, generated by `text` logger?
2. Are you talking about stdout output by `text` logger, like this `"logs": {"text": "php://stdout"}`
3. Is it supposed to be done by ANSI sequences?

If using ANSI sequenses, I guess the readability of the file will be reduced in the "raw" editor mode, because of ANSI chars (like `[1;32m Some string [0m`). So do we need a flag/option to enable this ANSI mode?

Slamdunk:
[...]

PR #1974, opened 2024-06-03
[Logs] Highlight inline differences in CLI

Slamdunk:
Fix https://github.com/infection/infection/issues/1810

> Why not to use [`jfcherng/php-diff`](https://github.com/jfcherng/php-diff) or [`diff-so-fancy`](https://github.com/so-fancy/diff-so-fancy)?

Infection's use case is much more limited: diffs are one-line only, and within the same line only one difference can happen.
Yes the algorithm in this PR needs some time to get understood, but less than tracking and maintaining another dependency.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <files psalm-version="5.11.0@c9b192ab8400fdaf04b2b13d110575adc879aa90">
+  <file src="src/Differ/DiffColorizer.php">
+    <InvalidArrayOffset>
+      <code>$lines[$prevIndex]</code>
+    </InvalidArrayOffset>
+  </file>
   <file src="src/Mutator/Extensions/BCMath.php">
     <ImpureFunctionCall>
       <code><![CDATA[$this->converters[$name->toLowerString()]($node)]]></code>
--- a/src/Container.php
+++ b/src/Container.php
@@ -285,7 +285,7 @@ final class Container
                 $container->getPrinter(),
                 $container->getMutantCodeFactory(),
             ),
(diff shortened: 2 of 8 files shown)


psalm-baseline.xml
```diff
@@ -1,5 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <files psalm-version="5.11.0@c9b192ab8400fdaf04b2b13d110575adc879aa90">
+  <file src="src/Differ/DiffColorizer.php">
+    <InvalidArrayOffset>
+      <code>$lines[$prevIndex]</code>
```
Slamdunk, line 5: I don't know why PSalm doesn't get the [`Assert::greaterThan($index, 0);`](https://github.com/infection/infection/pull/1974/files#diff-959b4198e0999129d1b54c6c33f931e33184201166012bb135dce6515ba78196R65)

src/Container.php
```diff
@@ -285,7 +285,7 @@ public static function create(): self
                 $container->getPrinter(),
                 $container->getMutantCodeFactory(),
             ),
-            Differ::class => static fn (): Differ => new Differ(new BaseDiffer(new UnifiedDiffOutputBuilder())),
+            Differ::class => static fn (): Differ => new Differ(new BaseDiffer(new UnifiedDiffOutputBuilder(''))),
```
Slamdunk, line 288: I hope you are ok removing the `"--- Original\n+++ New\n"` header, which only adds visual clutter to the output

maks-rafalko (approved):
It works just awesome. Thank you very much @Slamdunk 

A couple of real-world examples that became very easy to understand:

image

image

or this one

image

theofidry:
I'll miss that everyday in PHPUnit now 🤣

PR #1976, opened 2024-06-05
Add support for `colinodell/json5` v3

Slamdunk:
The `composer.lock` has been updated with the following command:

```console
$ docker compose run php81 composer update
```


maks-rafalko:
> `docker compose run php81 composer update`

shouldn't it be just `docker compose run php81 composer update colinodell/json5`? 

I mean, this updated all the deps, besides `colinodell/json5`, which broke the build.

Slamdunk:
Indeed :see_no_evil:

maks-rafalko (approved):

maks-rafalko:
Thank you @Slamdunk

maks-rafalko:
Released: https://github.com/infection/infection/releases/tag/0.29.3

maks-rafalko:
By the way, added info about target default branch, https://github.com/infection/infection/commit/99e03a7740c9204218f8e0495439374a9da72033

this is why GH actions didn't work properly, not finishing with status reports.

Slamdunk:
That's weird to me: I expected `master` to hold the upcoming MAJOR release with BC-Breaking changes, and `0.29` the upcoming MINOR release.

If that's not the case, which is the branch I'll have to push my PR against, when proposing BC-Breaking changes?

maks-rafalko:
In general, you are right and I'm sure after 1.0 release we will revise the current branching model and it will be something like you expect.

However right now we are still on `0.x` and historically we use "master-based development", we just merge everything in `master` and it becomes the next release. We are _trying_ to not break BC, but that's kind of ok on `0.x` (even composer [treats](https://getcomposer.org/doc/articles/versions.md#caret-version-range-) `0.3.0 -> 0.4.0` as a "major" release with potential BC breaks.


> If that's not the case, which is the branch I'll have to push my PR against, when proposing BC-Breaking changes?

so, answering to your question, now it will be just `master`

PR #1978, opened 2024-06-06
feat: introduce `--logger-project-root-directory` command option

vincentchalamon:
This PR:

- [x] Adds new feature: introduce `--logger-project-root-directory` command option to customize the logger project root directory
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/263

Fixes https://github.com/infection/infection/issues/1977

The detection of the project root dir, in GitLab and GitHub loggers, imposes the directory used and prevents the usage of a custom directory (for instance: using a custom Docker image in GitLab CI).

This new approach keeps the detection of GitLab and GitHub environment variables, but let the developer define a custom project root directory while using the command:

```shell
vendor/bin/infection --logger-gitlab=$CI_PROJECT_DIR/infection-gitlab.json --logger-project-root-directory=$WORKDIR
```

PR #1979, opened 2024-06-06
chore: update GitHub actions used

vincentchalamon:
This PR updates GitHub actions used in the CI, and introduces [GitHub Actions concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency)

### DIFF
--- a/.github/workflows/autoreview.yaml
+++ b/.github/workflows/autoreview.yaml
@@ -23,7 +23,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
@@ -39,7 +39,7 @@ jobs:
         run: echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
 
       - name: Cache dependencies
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: ${{ steps.composer-cache.outputs.dir }}
           key: composer-${{ runner.os }}-${{ matrix.php-version }}-${{ hashFiles('composer.*') }}
--- a/.github/workflows/autoreview.yaml
+++ b/.github/workflows/autoreview.yaml
@@ -10,6 +10,10 @@ on:
       - main
       - master
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   tests:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


.github/workflows/autoreview.yaml
```diff
@@ -10,6 +10,10 @@ on:
       - main
       - master
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
theofidry, line 14: in Box I'm using `group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}`, not too sure what's the diff tbh but LGTM

.github/workflows/cs.yaml
```diff
@@ -7,14 +7,18 @@ on:
       - main
       - master
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
```
sanmai, line 12: It makes total sense for a proprietary project as you pay for every minute, but it can hurt visibility for open-source projects. What are your arguments pro and contra?

vincentchalamon: I recently contributed to this project (#1978), and each commit triggered the CI pipeline. The CI can take a bit long (up to 30mn for E2E tests).

GitHub allows Open Source projects to use its GitHub-hosted runners for free, but with a limit of runners and CPU. That means while the E2E tests are still running from a previous commit, if I push a new commit (and the number of GitHub runners allowed is reached), I'll have to wait longer for the new CI to start.

Enabling concurrency automatically cancels previous CI runs for the same PR/branch (with this configuration) thanks to `cancel-in-progress: true`.

.github/workflows/autoreview.yaml
vincentchalamon: IMHO it's the same, as the [GitHub Contexts documentation](https://docs.github.com/en/actions/learn-github-actions/contexts) describes `github.head_ref`:
> The head_ref or source branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is either pull_request or pull_request_target.

maks-rafalko (approved):
LGTM, but let's wait for @sanmai's review

.github/workflows/cs.yaml
theofidry: I would also not care if the tests were super fasts, but we are hurting for bandwidth, 1 or 2 PR and we block the whole pipeline for 30min - 1h.

I would also say it's good to save needless computation time, I'd wager it's very rare that you need the CI of past commits to carry on

sanmai (approved):
I consulted with LLM and sounds like my concerns are unwarranted. Let's do it :+1:

maks-rafalko:
Thank you @vincentchalamon

Issue #1984, opened 2024-06-19
0.29.* PHAR version throws Fatal Error: Undefined constant when running tests with PHP-Parser 4


kristos80:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.29.0
| Test Framework version | PHPUnit
| PHP version | 8.3.8 
| Platform    | MacOS
| Github Repo | -

After installing from PHAR ```wget https://github.com/infection/infection/releases/download/0.29.0/infection.phar```, when running mutation tests I get the error 
```text
Fatal error: Uncaught Error: Undefined constant PhpParser\NodeVisitor::DONT_TRAVERSE_CURRENT_AND_CHILDREN in phar:///usr/local/bin/infection/src/PhpParser/Visitor/NonMutableNodesIgnorerVisitor.php:20
Stack trace:
...
```

<details>
 <summary>infection.json5</summary>

 ```json5
  {
  "$schema": "https://raw.githubusercontent.com/infection/infection/0.28.1/resources/schema.json",
  "source": {
    "directories": [
      "src"
    ],
  },
  "mutators": {
    "@default": true
  },
  "logs": {
    "text": "infection.log"
  }
}
 ```
</details>

<details>
 <summary>phpunit.xml</summary>

 ```xml
<?xml version="1.0"?>
<phpunit
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.5/phpunit.xsd"
        bootstrap="tests/bootstrap.php"
        colors="true"
        beStrictAboutOutputDuringTests="false"
        displayDetailsOnTestsThatTriggerWarnings="true">
    <testsuites>
        <testsuite name="Kiko">
            <directory>tests</directory>
        </testsuite>
    </testsuites>
    <source>
        <include>
            <directory suffix=".php">src/</directory>
        </include>
    </source>
    <extensions>
        <bootstrap class="DG\BypassFinals\PHPUnitExtension"/>
    </extensions>
</phpunit>
 ```
</details>

<details>
 <summary>Output with issue</summary>

 ```
Fatal error: Uncaught Error: Undefined constant PhpParser\NodeVisitor::DONT_TRAVERSE_CURRENT_AND_CHILDREN in phar:///usr/local/bin/infection/src/PhpParser/Visitor/NonMutableNodesIgnorerVisitor.php:20
Stack trace:
[...]

PR #1985, opened 2024-06-19
Draft: fix: Detect invalid PHP-Parser versions

theofidry:
Related to #1984.

The goal of the PR is to bail out more gracefully when an incompatible version of PHP-Parser is used.

### DIFF
--- a/tests/e2e/IncompatiblePhpParser/expected-output.txt
+++ b/tests/e2e/IncompatiblePhpParser/expected-output.txt
@@ -1,0 +1,10 @@
+Total: 1
+
+Killed: 1
+Errored: 0
+Syntax Errors: 0
+Escaped: 0
+Timed Out: 0
+Skipped: 0
+Ignored: 0
+Not Covered: 0
--- a/tests/phpunit/Console/E2ETest.php
+++ b/tests/phpunit/Console/E2ETest.php
@@ -168,7 +168,6 @@ final class E2ETest extends TestCase
     }
 
     #[DataProvider('e2eTestSuiteDataProvider')]
-    #[RunInSeparateProcess]
     public function test_it_runs_an_e2e_test_with_success(string $fullPath): void
     {
         $this->runOnE2EFixture($fullPath);
@@ -182,6 +181,10 @@ final class E2ETest extends TestCase
             ->directories();
 
         foreach ($directories as $dirName) {
+            if (basename((string) $dirName) !== 'IncompatiblePhpParser') {
+                continue;
+            }
+
             if (file_exists($dirName . '/run_tests.bash')) {
                 // skipping non-standard tests
                 // specifically Memory_Limit - it is very slow to fail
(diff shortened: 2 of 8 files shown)


tests/e2e/IncompatiblePhpParser/expected-output.txt
```diff
@@ -0,0 +1,10 @@
+Total: 1
```
theofidry, line 1: I am a bit confused because at the moment, if I `cd` to this directory and run `../../../build/infection.phar`, it works just fine...

Am I missing something?

tests/phpunit/Console/E2ETest.php
```diff
@@ -168,7 +168,6 @@ public function test_it_runs_configure_command_if_no_configuration(): void
     }
 
     #[DataProvider('e2eTestSuiteDataProvider')]
-    #[RunInSeparateProcess]
```
theofidry, line 171: Changes to this file should be reverted

tests/e2e/IncompatiblePhpParser/expected-output.txt
maks-rafalko: add this and the issue will be reproduced:

```diff
+/**
+ * @infection-ignore-all
+ */
+ class AstSampleProvide
```

maks-rafalko: the key thing is to trigger execution of this line

https://github.com/infection/infection/blob/a663d6cecee222e9b33f11608073c134eb5e6916/src/PhpParser/Visitor/NonMutableNodesIgnorerVisitor.php#L59

theofidry: That did it, I was using the wrong Infection PHAR for some reason

theofidry:
@maks-rafalko @sanmai I've updated the test with the solution, but we now need to:

- decide what to do with it: just write an error to the stderr or stop the process
- decide if we tighten or loosen the check: should it fail at different versions? Just major or minor versions?
- update the test

maks-rafalko:
> decide what to do with it: just write an error to the stderr or stop the process

IMO stop the process

> decide if we tighten or loosen the check: should it fail at different versions? Just major or minor versions?

we can start with major first? at least the problem we are talking about reproduces only between major versions

theofidry:
>at least the problem we are talking about reproduces only between major versions

[Not really](https://github.com/nikic/PHP-Parser/blob/master/CHANGELOG.md#version-410-2018-10-10), this is more about using in Infection a symbol not existing in older version, can very well happen in minor and technically even patch versions

maks-rafalko:
> [Not really](https://github.com/nikic/PHP-Parser/blob/master/CHANGELOG.md#version-410-2018-10-10), this is more about using in Infection a symbol not existing in older version, can very well happen in minor and technically even patch versions

In 4.1.0 it was initially added to PHP-Parser (btw, by me, specifically for Infection needs, lol). But it was moved to another class and deprecated in previous class [only in 5.0](https://github.com/nikic/PHP-Parser/blob/master/UPGRADE-5.0.md#changes-to-the-node-traverser).

but yeah, in theory BC break can be in any branch.

Issue #819, opened 2019-11-08
False positive uncovered mutant with `switch(true)` and `switch(false)`


luispabon:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.14.2
| Test Framework version | PHPUnit 8.4.3
| PHP version | 7.2.24/7.3.11/7.4.0-RC4
| Platform    | Ubuntu/docker
| Github Repo | https://github.com/luispabon/favicon-finder
| Example | https://infection-php.dev/r/emew

Infection is reporting an uncovered mutant on the repository above:

```diff
Not Covered mutants:
====================


1) /home/luis/Projects/favicon/src/Favicon.php:155    [M] TrueValue

--- Original
+++ New
@@ @@
             $links = $dom->getElementsByTagName('link');
             foreach ($links as $link) {
                 /** @var DOMNode $link */
-                switch (true) {
+                switch (false) {
                     case $link->hasAttribute('rel') && strtolower($link->getAttribute('rel')) === 'shortcut icon':
                     case $link->hasAttribute('rel') && strtolower($link->getAttribute('rel')) === 'icon':
                     case $link->hasAttribute('href') && strpos($link->getAttribute('href'), 'favicon') !== false:
```
This is however a false positive. That line is indeed covered, and the switch from true to false does break tests.

You can check by cloning, `composer -o install; make`. This runs phpunit and infection in sequence, infection recycling test results from phpunit. Letting infection generate its own reports makes no difference.

Build: https://ci.auronconsulting.co.uk/teams/main/pipelines/favicon-finder-master/jobs/analyze-master/builds/1

maks-rafalko:
Duplicate of https://github.com/infection/infection/issues/34

It boils down to how PHP works internally, please read exactly the same issue we created earlier https://github.com/infection/infection/issues/34

See Derick's explanation here https://bugs.xdebug.org/view.php?id=1479#c4450

I would say it won't be fixed on Infection side unless someone wants to do some magic to make it work (similar to how @BackEndTea [did it for array coverage](https://github.com/infection/infection/pull/655))

Closing for now, feel free to reopen, fix, and discuss farther though.

maks-rafalko:
Probably, it makes sense to not mutate `switch(false)` to `switch(true)` since we know it is always marked as `Not Covered`. What do you think @luispabon and @infection/core ?

BackEndTea:
Is the switch **never** covered? Or only when its switching over true/a constant?

[...]

PR #1986, opened 2024-06-20
Ignore switch(bool) statements in TrueValue and FalseValue mutators.

shanept:
This PR:

- [x] Modifies TrueValue and FalseValue mutators to ignore switch(boolval) statements.
- [x] Covered by tests

Fixes https://github.com/infection/infection/issues/819

This will stop false positives related to boolean values in switch conditions, where no code coverage has been detected.

### DIFF
--- a/tests/phpunit/Mutator/Boolean/FalseValueTest.php
+++ b/tests/phpunit/Mutator/Boolean/FalseValueTest.php
@@ -80,6 +80,15 @@ final class FalseValueTest extends BaseMutatorTestCase
             ,
         ];
 
+        yield 'It does not mutate switch false to true' => [
+            <<<'PHP'
+                <?php
+
+                switch (false) {}
+                PHP
+            ,
+        ];
+
         yield 'It mutates all caps false to true' => [
             <<<'PHP'
                 <?php
--- a/src/Mutator/Boolean/FalseValue.php
+++ b/src/Mutator/Boolean/FalseValue.php
@@ -39,6 +39,7 @@ use Infection\Mutator\Definition;
 use Infection\Mutator\GetMutatorName;
 use Infection\Mutator\Mutator;
 use Infection\Mutator\MutatorCategory;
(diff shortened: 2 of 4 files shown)


shanept:
**NOTE**: The test cases could probably use particular attention. My test cases are asserting that switch statements are **NOT** mutated. I am not familiar with the Infection codebase, and could not find equivalent test cases or assertions to base mine upon, so it may or may not need massaging. All test cases behave as I would expect, otherwise.

Here, specifically, is how I have accomplished this assertion:
https://github.com/infection/infection/blob/e7aa2fdeefd6cc9701d13b3adb29a0ac86aace3d/src/Testing/BaseMutatorTestCase.php#L116-L123

tests/phpunit/Mutator/Boolean/FalseValueTest.php
```diff
@@ -109,4 +109,15 @@ public function test_does_not_mutate_true_value(): void
...
+    public function test_does_not_mutate_switch_stmt(): void
+    {
+        $inputCode = <<<'PHP'
+            <?php
+
+            switch (false) {}
+            PHP;
+
+        $this->assertInputNotMutated($inputCode);
+    }
```
maks-rafalko, line 122: actually, you don't need to create separate method to test the case when no mutation is happened.

Please, look here to get an example:

https://github.com/infection/infection/blob/a663d6cecee222e9b33f11608073c134eb5e6916/tests/phpunit/Mutator/Boolean/FalseValueTest.php#L74-L81

so you basically need to create the same `yield` with swith. Thus, new created method `assertInputNotMutated` can also be removed

maks-rafalko (commented):
Looks good, thanks. After tests are fixed, happy to merge it

shanept:
Thanks for the feedback @maks-rafalko. I had a feeling there was a better way...
Tests modified as per your instructions, and all the extra test code was removed.

maks-rafalko (approved):
Thank you @shanept

PR #1990, opened 2024-06-29
Run e2e tests in parallel

maks-rafalko:


tests/e2e_tests
```diff
@@ -36,14 +37,33 @@ do
...
-    exit 1
-fi
+    if [ "$failure" = "true" ]
+    then
+        exit 255
+    fi
+}
+
+exitOnError() {
+  exit_code=$1
```
theofidry, line 48: I think the CS was 4 spaces

tests/e2e_tests
```diff
@@ -36,14 +37,33 @@ do
...
 
-if [ $failure = "true" ]
-then
-    echo "One of the e2e tests failed, check the logs";
-    cat error.log
-    exit 1
-fi
+    if [ "$failure" = "true" ]
+    then
+        exit 255
```
theofidry, line 43: why 255 rather than 1?

.github/workflows/e2e-tests.yaml
```diff
@@ -98,11 +98,11 @@ jobs:
         shell: bash
         run: |
           if [[ "${{ matrix.php-version }}" == '8.3' ]]; then
-            ls tests/e2e/*/composer.json | xargs dirname | xargs -I{} composer --working-dir={} config platform.php 8.2.99
-            ls tests/e2e/*/composer.json | xargs dirname | xargs -I{} composer --working-dir={} install --no-interaction --prefer-dist --no-progress
+            ls tests/e2e/*/composer.json | xargs dirname | xargs -I{} -P 4 composer --working-dir={} config platform.php 8.2.99
```
theofidry, line 101: what is this `-P 4` for?

tests/e2e_tests
```diff
@@ -11,8 +11,9 @@ fi
 
 touch error.log
```
theofidry, line 12: unrelated to the MR, just noticed it when trying to understand the script. The following doesn't make much sense to me:

if `error.log` exists, then create+truncate it, otherwise touch it.

I think it could be simplified into:
```suggestion
touch error.log
echo '' > error.log
```

or if you do prefer the shorthand version (personally I had to look it up):

```suggestion
:> error.log
```

The suggestion can't capture it, didn't think I would prefer GitLab diff but here we are... I meant to capture the whole block with the if statement + the touch

tests/e2e_tests
```diff
@@ -23,9 +24,9 @@ do
 
     if [ -f "run_tests.bash" ]
     then
-        output="$(INFECTION_E2E_TESTS_ENV=1 bash run_tests.bash ${1:-bin/infection})"
```
theofidry, line 26: might be worth extracting `$2` into a variable like `dir`?

tests/e2e_tests
```diff
@@ -23,9 +24,9 @@ do
...
+        output="$(INFECTION_E2E_TESTS_ENV=1 bash run_tests.bash ${2:-bin/infection})"
     else
-        output="$(INFECTION_E2E_TESTS_ENV=1 bash ../standard_script.bash ${1:-bin/infection})"
+        output="$(INFECTION_E2E_TESTS_ENV=1 bash ../standard_script.bash ${2:-bin/infection})"
```
theofidry, line 29: hmm just noticed it too... we are actually not capturing the stderr at all. In Box I had to start testing both outputs in several cases so I ended up having to output files, but if it doesn't matter we could just combine the stdout and stderr here?

tests/e2e_tests
```diff
@@ -36,14 +37,33 @@ do
         echo "" >> ../error.log
     fi
     cd ..
```
theofidry, line 39: maybe it would be better to capture the current working dir before the cd and restore it, that way the test could make use of a cd without breaking this script:

```suggestion
pwd=$(pwd)
cd $dir

# ...

cd $pwd
```

tests/e2e_tests
```diff
@@ -36,14 +37,33 @@ do
...
+  msg=$2
+  if [ "$exit_code" -ne 0 ]; then
+    echo -e "\033[31mEXIT $exit_code $msg\033[0m" >&2
+    # If any invocation of the command exits with a status of 255, xargs will stop immediately without reading any further input.
+    # An error message is issued on stderr when this happens.
+    exit 255
+  fi
+}
+
+dirs=$(find . -maxdepth 1 -mindepth 1 -type d)
```
theofidry, line 58:
```suggestion
test_cases_dirs=$(find . -maxdepth 1 -mindepth 1 -type d)
```
?

tests/e2e_tests
```diff
@@ -36,14 +37,33 @@ do
...
+
+# Export the function so xargs can use it
+export -f run_infection_for_dir
+
+# Use xargs to run the run_infection_for_dir function in parallel
+echo "$dirs" | xargs -I {} -P 4 bash -c "run_infection_for_dir {} ${1:-bin/infection}"
```
theofidry, line 64: TIL, I thought you had to use GNU parallel, turns out xargs does it just fine, cool!

I do wonder if we could use the makefile instead at some point (also makes it easier to scale up the number of processes), but that's an entirely different set of issue

theofidry (approved):
Only a superficial review on my end unfortunately, I'm not deep enough into this yet.

A number of comments stem more from trying to have an overview of the script, so IMO they should not be addressed in this PR as out of scope.

Unrelated to the PR, I find the following installations during the e2e tests: 

- https://github.com/infection/infection/actions/runs/9722831274/job/26837206598?pr=1990#step:14:11
- https://github.com/infection/infection/actions/runs/9722831274/job/26837206598?pr=1990#step:14:293

Are they expected?

.github/workflows/e2e-tests.yaml
sanmai: `-P maxprocs`

tests/e2e_tests
maks-rafalko: 255 tells to xargs to stop immediately if one of the process fails. If you use return status 1, it continues execution

tests/e2e_tests
maks-rafalko: fixed

tests/e2e_tests
maks-rafalko: agree, done

tests/e2e_tests
maks-rafalko: done

tests/e2e_tests
maks-rafalko: actually, our tests don't do `cd` and I pretty sure they shouldn't. Left as is for now, but probably I miss something

tests/e2e_tests
maks-rafalko: done

maks-rafalko:
> Unrelated to the PR, I find the following installations during the e2e tests:

it looks wierd, from the quick look it's something from makefile and dependencies of `make test-e2e` target... but I didn't figure out it quickly

tests/e2e_tests
maks-rafalko: done in a separate PR https://github.com/infection/infection/pull/1991

PR #1994, opened 2024-07-14
Fix PHP-Parser 5.0 and 5.1 compatibility in tests

sidz:
This PR fixes PHP-Parser 5.0 and 5.1 compatibility in tests.

see https://github.com/nikic/PHP-Parser/issues/1009

Currently our master branch is broken as highest dependencies install PHP-Parser 5.1 and leads to https://github.com/infection/infection/actions/runs/9928704984/job/27425338269?pr=1993

Fix is a bit ugly but I think it's more than fine to go with this patch as PHP-Parser incompatibility affects only tests.

p.s. Otherwise we could write our own pretty printer :)


maks-rafalko (approved):

sidz:
thanks @maks-rafalko

PR #1995, opened 2024-07-14
Use Null Coalescing Assignment Operator

sidz:
This PR doesn't bring a big value it just replacing  `$variable ?? $variable = $this->dosmth()` with Null Coalescing Assignment Operator `$variable ??= $this->dosmth()`

Unfortunately there is no any rule nor in Rector or PHP-CS-Fixer which may cover such cases and fail builds in the future.

Existing in both projects doesn't take into account properties.
PHP-CS-Fixer: https://cs.symfony.com/doc/rules/operator/assign_null_coalescing_to_coalesce_equal.html
Rector: https://github.com/rectorphp/rector/blob/main/rules/Php74/Rector/Assign/NullCoalescingOperatorRector.php

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -718,7 +718,7 @@ final class Container
 
     public function getDefaultJUnitFilePath(): string
     {
-        return $this->defaultJUnitPath ?? $this->defaultJUnitPath = sprintf(
+        return $this->defaultJUnitPath ??= sprintf(
             '%s/%s',
             Path::canonicalize(
                 $this->getConfiguration()->getCoveragePath(),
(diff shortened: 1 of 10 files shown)


maks-rafalko (approved):
up to you: does it make sense to create an issue in rector/php-cs-fixer to support properties? or is there any issue, probably you already found?

theofidry (approved):

sidz:
> or is there any issue, probably you already found?

I didn't find any solution yet.

> does it make sense to create an issue in rector/php-cs-fixer to support properties?

That was my plan to create an issue in Rector.

sidz:
thanks @maks-rafalko, @theofidry

Issue #1999, opened 2024-10-04
InvalidArgumentException in DiffColorizer


sanmai:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.29.6 
| Test Framework version | PHPUnit 9.6.21
| PHP version |8.2.20
| Platform    | Linux
| Github Repo | https://github.com/kalimatas/php-liquid/pull/219

Downgrading to 0.28.1 fixes the problem.

I'm looking at the following stack trace:

```
In Assert.php line 2074:
                                               
  [Webmozart\Assert\InvalidArgumentException]  
  Expected a value identical to "-". Got: "<"  
                                               

Exception trace:
  at vendor/webmozart/assert/src/Assert.php:2074
 Webmozart\Assert\Assert::reportInvalidArgument() at vendor/webmozart/assert/src/Assert.php:814
 Webmozart\Assert\Assert::same() at vendor/infection/infection/src/Differ/DiffColorizer.php:70
 Infection\Differ\DiffColorizer->colorize() at vendor/infection/infection/src/Event/Subscriber/MutationTestingConsoleLoggerSubscriber.php:145
 Infection\Event\Subscriber\MutationTestingConsoleLoggerSubscriber->showMutations() at vendor/infection/infection/src/Event/Subscriber/MutationTestingConsoleLoggerSubscriber.php:102
 Infection\Event\Subscriber\MutationTestingConsoleLoggerSubscriber->onMutationTestingWasFinished() at vendor/infection/infection/src/Event/EventDispatcher/SyncEventDispatcher.php:59
 Infection\Event\EventDispatcher\SyncEventDispatcher->dispatch() at vendor/infection/infection/src/Process/Runner/MutationTestingRunner.php:123
 Infection\Process\Runner\MutationTestingRunner->run() at vendor/infection/infection/src/Engine.php:134
 Infection\Engine->runMutationAnalysis() at vendor/infection/infection/src/Engine.php:74
 Infection\Engine->execute() at vendor/infection/infection/src/Command/RunCommand.php:403
 Infection\Command\RunCommand->executeCommand() at vendor/infection/infection/src/Command/BaseCommand.php:75
 Infection\Command\BaseCommand->execute() at vendor/symfony/console/Command/Command.php:279
 Symfony\Component\Console\Command\Command->run() at vendor/symfony/console/Application.php:1029
 Symfony\Component\Console\Application->doRunCommand() at vendor/symfony/console/Application.php:316
[...]

PR #2000, opened 2024-10-06
Fallback to the simple diff colorizer (previous implementation) for multiline diffs

maks-rafalko:
Fixes https://github.com/infection/infection/issues/1999

We introduced inline highlighter for diffs in https://github.com/infection/infection/pull/1974 (thanks to @Slamdunk). However, there was an incorrect assumption that we only have one-line diffs. This is not true.

The case which is currently fails is the following:

```diff
@@ @@
      */
     protected function name()
     {
-        return strtolower(get_class($this));
+        strtolower(get_class($this));
+        return null;
     }
 }
```

I checked how `diff-so-fancy` works with multiline diffs, and it turned out that it just skips inline diffs, using simple algorithm, see images below.

`diff-so-fancy` for multiline case (now we have the same result after this MR):

image

`diff-so-fancy` for one-line case (we have the same result):

image

So, for the cases when we have multiline diffs, I just added fallback - our old algorithm of colorizing diffs.

----

I've tested this patch with @sanmai's reproducer, it works fine:

```
vendor/bin/infection --min-msi=80 --min-covered-msi=80 --show-mutations -jmax -vvv --mutators=FunctionCall --filter=AbstractTag
```

image

### DIFF
--- a/src/Differ/DiffColorizer.php
+++ b/src/Differ/DiffColorizer.php
@@ -35,12 +35,16 @@ declare(strict_types=1);
 
 namespace Infection\Differ;
 
+use function array_map;
+use function count;
 use function explode;
 use function implode;
+use function is_array;
 use function mb_strlen;
 use function mb_strpos;
 use function mb_strrpos;
 use function mb_substr;
+use function Safe\preg_match_all;
 use function sprintf;
 use function str_starts_with;
 use function substr;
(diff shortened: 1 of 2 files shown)


sanmai (approved):
Amazing!

Slamdunk:
I'm so sorry I missed this case. I'm glad you found a quick solution 👍

maks-rafalko:
no worries, thanks to this feature, love it ;)

PR #2002, opened 2024-10-25
use $GLOBALS['_composer_autoload_path'] if possible

Kanti:
I cloned infection into a local directory to install it via composer from there. 
So I can work on it in my normal project along side it. ( https://getcomposer.org/doc/05-repositories.md#path)

But than the autoloading is not found.
With the `$GLOBALS['_composer_autoload_path']` it is now possible to install it from anythere.
This is possible since composer 2.2:
https://getcomposer.org/doc/articles/vendor-binaries.md#finding-the-composer-autoloader-from-a-binary

### DIFF
--- a/bin/infection
+++ b/bin/infection
@@ -49,6 +49,7 @@ if (\in_array(\PHP_SAPI, ['cli', 'phpdbg', 'embed'], true) === false) {
 // Infection autoloading
 (static function (): void {
     $autoload = $GLOBALS['_composer_autoload_path'] ?? '';
+
     if ($autoload && \file_exists($autoload)) {
         // Is installed via Composer
         include_once $autoload;
--- a/bin/infection
+++ b/bin/infection
@@ -48,6 +48,13 @@ if (\in_array(\PHP_SAPI, ['cli', 'phpdbg', 'embed'], true) === false) {
 
 // Infection autoloading
 (static function (): void {
+    if (\file_exists($autoload = ($GLOBALS['_composer_autoload_path'] ?? ''))) {
+        // Is installed via Composer
+        include_once $autoload;
+
+        return;
+    }
+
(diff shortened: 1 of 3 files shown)


bin/infection
```diff
@@ -48,6 +48,13 @@ if (\in_array(\PHP_SAPI, ['cli', 'phpdbg', 'embed'], true) === false) {
 
 // Infection autoloading
 (static function (): void {
+    if (\file_exists($autoload = ($GLOBALS['_composer_autoload_path'] ?? ''))) {
```
sanmai, line 51: What do you think about checking if the global exists before checking if the file exists?

So we don't call file_exists on an empty string.

bin/infection
```diff
@@ -48,6 +48,14 @@ if (\in_array(\PHP_SAPI, ['cli', 'phpdbg', 'embed'], true) === false) {
 
 // Infection autoloading
 (static function (): void {
+    $autoload = $GLOBALS['_composer_autoload_path'] ?? '';
```
Kanti, line 51:
```suggestion
    $autoload = $GLOBALS['_composer_autoload_path'] ?? '';

```

maks-rafalko:
> If you want to rely on this in your package you should however make sure to also require "composer-runtime-api": "^2.2" to ensure that the package gets installed with a Composer version supporting the feature.

I guess we should update package.json/lock to require ^2.2 then?

Kanti:
We could, but we don't need to. If the user uses composer 2.2 or higher he can profit from this change. if he dose not, it still works, as we have the other 2 includes.

maks-rafalko:
ok, there is a failing test unrelated to your changes. I will merge this PR as soon as I fix the tests.

maks-rafalko (approved):

maks-rafalko:
Thank you @Kanti

PR #2001, opened 2024-10-07
add failing test for cacheResultFile

Kanti:
This PR:

- [x] adds a new test for PHPUnit 11 and the testdoxSummary attribute
- [ ] find out how we can skip the `PHPUnit11` e2e folder in Github Actions using PHP <8.2
- [ ] fix the code so the failing test is gone

If I set the `testdoxSummary` in my `phpunit.xml` all mutations are magically killed.
It should show me that my `phpunit.xml` is not valid for infection or it should behave normally.

It has something to do with `handleResultCacheAndExecutionOrder` of the `XmlConfigurationManipulator`. because if I remove the lines:
````diff
        if (version_compare($version, '7.3', '>=')) {
-            $this->setAttributeValue($xPath, 'cacheResult', 'true');
-            $this->setAttributeValue($xPath, 'cacheResultFile', sprintf('.phpunit.result.cache.%s', $mutationHash));
-            $this->setAttributeValue($xPath, 'executionOrder', 'defects');

            return;
        }
````

it is working normally again :/

I could not find out why this really happens. Hope this Test helps

PR #2003, opened 2024-10-25
fix phpunit 11 config

Kanti:
As described in #2001 with phpunit 11 infection is broken.
I added the test from #2001 and fixed the configuration by:
- replacing `cacheResultFile`with `cacheDirectory`for every phpunit installation >=11
- replacing `stopOnFailure`with `stopOnDefect` to speed up the infection time by 4x (on my project)

TODOs:
- [x]  find out how we can skip the PHPUnit11 e2e folder in Github Actions using PHP <8.2 (Maybe someone can help me here as I'm not familiar with this test setup.)

### DIFF
--- a/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
+++ b/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
@@ -126,7 +126,7 @@ final class XmlConfigurationManipulator
         $this->setAttributeValue($xPath, 'stderr', 'false');
     }
 
-    public function setStopOnFailure(string $version, SafeDOMXPath $xPath): void
+    public function setStopOnFailureOrDefect(string $version, SafeDOMXPath $xPath): void
     {
         // with phpunit 10.0  we need to use stopOnDefect instead: https://github.com/sebastianbergmann/phpunit/blob/10.0.0/ChangeLog-10.0.md
         if (version_compare($version, '10.0', '>=')) {
--- a/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
+++ b/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
@@ -95,8 +95,17 @@ final class XmlConfigurationManipulator
         $this->setAttributeValue($xPath, 'cacheResult', 'false');
     }
 
-    public function handleResultCacheAndExecutionOrder(string $version, SafeDOMXPath $xPath, string $mutationHash): void
+    public function handleResultCacheAndExecutionOrder(string $version, SafeDOMXPath $xPath, string $mutationHash, string $tmpDir): void
     {
+        // starting from PHPUnit 11.0 the cacheResultFile was removed, we now set cacheDirectory instead https://github.com/sebastianbergmann/phpunit/blob/11.0.0/phpunit.xsd
+        if (version_compare($version, '11.0', '>=')) {
+            $this->setAttributeValue($xPath, 'cacheResult', 'true');
+            $this->setAttributeValue($xPath, 'cacheDirectory', sprintf('%s/.phpunit.result.cache.%s', $tmpDir, $mutationHash));
+            $this->setAttributeValue($xPath, 'executionOrder', 'defects');
+
+            return;
+        }
+
         // starting from PHPUnit 7.3 we can set cache result and "defects" execution order https://github.com/sebastianbergmann/phpunit/blob/7.3.0/phpunit.xsd
         if (version_compare($version, '7.3', '>=')) {
             $this->setAttributeValue($xPath, 'cacheResult', 'true');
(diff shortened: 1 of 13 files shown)


src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
```diff
@@ -117,13 +126,16 @@ public function deactivateStderrRedirection(SafeDOMXPath $xPath): void
...
 
-    public function setStopOnFailure(SafeDOMXPath $xPath): void
+    public function setStopOnFailure(string $version, SafeDOMXPath $xPath): void
     {
-        $this->setAttributeValue(
-            $xPath,
-            'stopOnFailure',
-            'true',
-        );
+        # with phpunit 10.0  we need to use stopOnDefect instead: https://github.com/sebastianbergmann/phpunit/blob/10.0.0/ChangeLog-10.0.md
```
Kanti, line 131:
```suggestion
        // with phpunit 10.0  we need to use stopOnDefect instead: https://github.com/sebastianbergmann/phpunit/blob/10.0.0/ChangeLog-10.0.md
```

maks-rafalko:
> * Maybe someone can help me here as I'm not familiar with this test setup.

one possible way is to add `run_tests.bash` to your e2e test folder and there you can check (pseudocode)

```
if %php version < 8.2%
   exit 0
endif
```

Regarding `cacheResultFilewith` - I will need to dig into it, will try to allocate time in the nearest days, but it won't be fast as I quite limited these days.

maks-rafalko:
> replacing `stopOnFailure`with `stopOnDefect` to speed up the infection time by 4x (on my project)

that's crazy that PHPUnit doesn't fail with incorrect flag usage :( Infection is so much faster when tests execution stops at the first failure rather than executing all the tests. Thanks for fixing it, this is a major bugfix

One question though: how `cacheDirectory` and `stopOnDefect` is related to `testdox` issue you are referencing?

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
```diff
@@ -117,13 +126,16 @@ public function deactivateStderrRedirection(SafeDOMXPath $xPath): void
         $this->setAttributeValue($xPath, 'stderr', 'false');
     }
 
-    public function setStopOnFailure(SafeDOMXPath $xPath): void
+    public function setStopOnFailure(string $version, SafeDOMXPath $xPath): void
```
maks-rafalko, line 129: minor: what do you think about renaming to `setStopOnFailureOrDefect`?

tests/e2e/PHPUnit11/README.md
```diff
@@ -0,0 +1,9 @@
+# Title
+
+* Link to github ticket if relevant
```
maks-rafalko, line 3: Please add link to the GH issue/pr and update summary please

maks-rafalko:
> speed up the infection time by 4x (on my project)

did you analyze your project's Infection log file to understand why there is such a huge difference? I've tested `master`  branch to run Infection on itself and your branch, there is no difference, at least from the first look.

Kanti:
The 4x performance gain is because of my testing structure. It will not effect all projects by that much.

I looked into it and found out that I have mutations that are covered by hundreds of test Cases.
And some mutations trigger an error, that error will not trigger phpunit to stop running the tests. So phpunit will run all the hundreds of tests even if the first one has errored.

With the `stopOnDefect` it will stop after the first error. As well as the first failure.

If you don't have that many test cases that are covering a mutation or the mutation does not trigger an error, you won't have that much of a performance gain.

I thought that the `testdox` was the problem, but it was not. It was the `cacheResultFile` all along.
As I worked on that I also found that `setStopOnFailureOrDefect` needs a change, it was more a coincidence.

maks-rafalko:
ok, thanks for the explanation, now it's much clear. 

> And some mutations trigger an error, that error will not trigger phpunit to stop running the tests.

could you please provide an example of the original code and mutated code that triggers error and stop only with `stopOnDefect` but not with `stopOnFailure`? 

I would like to play with it and see how it affects other users.

Kanti:
I looked into my project and found one problem:
If the type dose not match it is an error and not failure.
I recreated a fictitious test case here: https://infection-php.dev/r/1l4r

As you can see, it runs all 2000 test cases instead of stopping after the first:
grafik

maks-rafalko:
Ok, regarding `cacheResultFile` being removed from PHPUnit 11, just for history and search purposes, here is what was the issue:

```
  1) Test results may not be as expected because the XML configuration file did not pass validation:
  
    Line 2:
    - Element 'phpunit', attribute 'cacheResultFile': The attribute 'cacheResultFile' is not allowed.
  
  
  No tests executed!
```

so this PR solves correctly.

Thank you so much @Kanti for your work on fixing and adding improvements (`stopOnDefect`), this is very helpful and useful.

maks-rafalko:
Released https://github.com/infection/infection/releases/tag/0.29.8

maks-rafalko:
Updated Playground so it clearly visible how this PR improved Infection execution:

- before https://infection-php.dev/r/1l4r
- after https://infection-php.dev/r/751k

PR #2012, opened 2024-12-08
Upgrade libraries to avoid PHP Warnings on PHP 8.4

maks-rafalko:
Running Infection on PHP 8.4 is currently produces a lot of warnings:

```
Implicitly marking parameter $var as nullable is deprecated
```

This PR updates libs to avoid it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,7 +65,7 @@
         "symfony/filesystem": "^5.4 || ^6.0 || ^7.0",
         "symfony/finder": "^5.4 || ^6.0 || ^7.0",
         "symfony/process": "^5.4 || ^6.0 || ^7.0",
-        "thecodingmachine/safe": "^2.1.2",
+        "thecodingmachine/safe": "dev-master as 2.5.0",
         "webmozart/assert": "^1.11"
     },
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -65,7 +65,7 @@
         "symfony/filesystem": "^5.4 || ^6.0 || ^7.0",
         "symfony/finder": "^5.4 || ^6.0 || ^7.0",
         "symfony/process": "^5.4 || ^6.0 || ^7.0",
-        "thecodingmachine/safe": "^2.1.2",
```
maks-rafalko, line 68: `thecodingmachine/safe` has fixes in master, but didn't create a new tag since 2023

sanmai (approved):
Bummer that we can't get a new version of `thecodingmachine/safe`

maks-rafalko:
> Bummer that we can't get a new version of `thecodingmachine/safe`

seems like they have ongoing discussion on how to proceed: https://github.com/thecodingmachine/safe/issues/500#issuecomment-2518796063

but yeah, currently no new release unfortunately

Issue #2015, opened 2024-12-09
Dependency "thecodingmachine/safe" outdated generate Deprecated logs


naspy971:
I'm using php 8.4 and I just installed Infection version ^0.29.8 and I have this type of logs (multiple lines) :

`Deprecated: Safe\mktime(): Implicitly marking parameter $day as nullable is deprecated, the explicit nullable type must be used instead in /var/www/html/app/vendor/thecodingmachine/safe/deprecated/datetime.php on line 78`

Seems like Infection needs to upgrade its `thecodingmachine/safe` dependency.

Thanks.

darthf1:
There is no new stable version of `thecodingmachine/safe` to update to...

Also, yesterday, in https://github.com/infection/infection/pull/2012 the dependency was updated to `dev-master as 2.5.0`, but now my composer gives errors; is there a workaround for this whilst keeping my `minimum-stability`?

```
Loading composer repositories with package information
Restricting packages listed in "symfony/symfony" to "7.2.*"
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires infection/infection 0.29.9 -> satisfiable by infection/infection[0.29.9].
    - infection/infection 0.29.9 requires thecodingmachine/safe dev-master as 2.5.0 -> found thecodingmachine/safe[dev-master] but it does not match your minimum-stability.

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.
```

maks-rafalko:
> I'm using php 8.4 and I just installed Infection version ^0.29.8

Please use yesterday's release 0.29.9

> but now my composer gives errors; is there a workaround for this whilst keeping my minimum-stability?

Use Infection PHAR distribution? Without deep investigating, I'm not aware of any other ways of how we can get the latest changes of `thecodingmachine/safe` without pointing to `dev-master`.

PHAR distribution should be a good workaround for you for now.

maks-rafalko:
please use recommendation from https://github.com/infection/infection/issues/2015#issuecomment-2527373043

If you anyone has other ideas, feel free to suggest.

maks-rafalko:
> I'm not sure I get you ? I actually tried to install the specific version you mentionned

to avoid "minimum-stability" issue with 0.29.9 you can use PHAR distribution instead of a composer package, see https://infection.github.io/guide/installation.html#Phar (but change the version to 0.29.9)

another idea to try on your side is to use

[...]

PR #2016, opened 2024-12-09
Remove thecodingmachine/safe temporarily

sanmai:
This PR:

- [ ] Fixes  https://github.com/infection/infection/issues/2015
- [ ] Covered by tests


sanmai:
Tough luck: we actually depend on some functions from that library to throw exceptions.

PR #2029, opened 2025-02-13
Get rid of Symfony 5.4 support as it's EOLed and on security fixes only

maks-rafalko:
https://symfony.com/releases

2 reasons for this:

- Symfony 5.4 is dead, security fixes only
- We don't have quotes for so many GH Actions, so removing Symfony 5.4 from the matrix will greatly speed up our builds. Currently, when quotes for 1 PR are used, we have to wait a coupls of minutes for the next GH Actions to be started, they are hanging in a pending state

image


sidz:
Just a thought:
what about drop support for Symfony < 6.4 too @maks-rafalko ?

maks-rafalko:
> what about drop support for Symfony < 6.4 too @maks-rafalko ?

like it ;) I will do later in this PR, thanks

sanmai (approved):
Say no more

sidz:
we also could get rid of this kind of hack https://github.com/infection/infection/blob/master/src/FileSystem/DummyFileSystem.php#L42 too

sidz (approved):

PR #2034, opened 2025-02-13
Add PHP 8.2 to docker-compose and remove PHP 8.1

maks-rafalko:
- PHP 8.1 has been dropped
- it's convenient to have 8.2 docker image to be built and used to run particular commands when you are on 8.3/8.4 locally

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,6 @@ services:
       context: devTools
       args:
         PHP_VERSION: '8.2'
-        XDEBUG_VERSION: '3.4.0'
+        XDEBUG_VERSION: '3.4.1'
     volumes:
       - .:/opt/infection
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -1,10 +1,9 @@
...
 services:
-  php81:
+  php82:
     build:
       context: devTools
       args:
-        PHP_VERSION: '8.1'
-        XDEBUG_VERSION: '3.1.5'
+        PHP_VERSION: '8.2'
+        XDEBUG_VERSION: '3.4.0'
```
sidz, line 7: 3.4.1 is latest version

maks-rafalko: updated

sidz (approved):

PR #2037, opened 2025-02-17
Feature/safe revert back

maks-rafalko:
- Replace `shish/safe` back with `thecodingmachine/safe` v3
- `thecofingmachine/safe-phpstan-rule` can not be used yet, because it requires PHPStan v2. So some errors are still in phpstan baseline file

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,11 +61,11 @@
         "sanmai/later": "^0.1.1",
         "sanmai/pipeline": "^5.1 || ^6",
         "sebastian/diff": "^3.0.2 || ^4.0 || ^5.0 || ^6.0 || ^7.0",
-        "shish/safe": "^2.6",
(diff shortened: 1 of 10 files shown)


sidz:
> thecofingmachine/safe-phpstan-rule can not be used yet, because it requires PHPStan v2. So some errors are still in phpstan baseline file

I've started upgrade PHPStan to 2 but didn't finish it yet. So I'll bring `thecofingmachine/safe-phpstan-rule` back in my PR

maks-rafalko:
nice

> I've started upgrade PHPStan to 2 but didn't finish it yet. So I'll bring `thecofingmachine/safe-phpstan-rule` back in my PR

cool, thanks. Could you please review this PR since you are here? :) I will do a release then, as it blocks someone from the community

PR #2038, opened 2025-02-17
Update PHPStan to version 2

sidz:
This PR updates PHPStan to version 2 and also brings `thecodingmachine/phpstan-safe-rule` back

### DIFF
--- a/devTools/phpstan-src.neon
+++ b/devTools/phpstan-src.neon
@@ -9,7 +9,6 @@ parameters:
     sidzIgnoreNumericStrings: true
     ignoreErrors:
         - '#generic class ReflectionClass (but )?does not specify its types#'
-        - '#^Call to static method Webmozart\\Assert\\[a-zA-Z]+::[a-zA-Z]+\(\) with .* will always evaluate to true\.$#'
         - '#^Parameter \#\d+ \$.+ \(.+\) of method .+ should be contravariant with parameter \$.+ \(.+\) of method .+$#'
         - '#^Short ternary operator is not allowed\. Use null coalesce operator if applicable or consider using long ternary\.$#'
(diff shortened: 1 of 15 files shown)


rector.php
```diff
@@ -58,4 +59,6 @@
             // property can't be readonly as it's returned by reference and may be updated
             __DIR__ . '/src/TestFramework/Coverage/TestLocations.php',
         ],
+        ReadOnlyClassRector::class,
+        ReadOnlyPropertyRector::class,
```
maks-rafalko, line 63: is it because it produces too many changes for now? Or is it because we don't these rules at all?

devTools/phpstan-src.neon
```diff
@@ -62,14 +61,14 @@ parameters:
             message: "#Do not use magic number (.*)#"
             count: 8
             path: ../src/Mutator/Extensions/BCMath.php
+        -
+            identifier: arrayFilter.strict
```
maks-rafalko, line 65: same question: is it because it produces too many errors or is it because we don't want this rule at all?

rector.php
sidz: It's out of scope of this ticket so I decided to exclude these rules for now

devTools/phpstan-src.neon
sidz: It produces 8 new errors for cases like this https://github.com/infection/infection/blob/dfe9cf6e65545881c7d21343c494cc8a1fdbfb80/src/Configuration/Schema/InvalidSchema.php#L58

Do we want to define callback explicitly?

maks-rafalko: I would vote no ;)

PR #2039, opened 2025-02-17
PHPUnit 11

sidz:
This PR updates PHPUnit to the 11.5 and fixes some deprecation notes.


composer.json
```diff
@@ -76,13 +76,12 @@
     "require-dev": {
         "ext-simplexml": "*",
         "fidry/makefile": "^1.0",
-        "helmich/phpunit-json-assert": "^3.0",
```
sidz, line 79: seems to be unused

phpunit.xml.dist
```diff
@@ -6,7 +6,11 @@
          executionOrder="random"
          failOnWarning="true"
          failOnRisky="true"
+         failOnDeprecation="true"
```
sidz, line 9: I think that's a good idea to fail builds on deprecation so we won't introduce more deprecation

phpunit.xml.dist
```diff
@@ -6,7 +6,11 @@
...
          requireCoverageMetadata="true"
+         displayDetailsOnPhpunitDeprecations="true"
```
sidz, line 11: Just show details in CI so I as developer could see the reason why build is failed

tests/phpunit/Mutator/Util/AbstractValueToNullReturnValueTest.php
```diff
@@ -50,13 +50,11 @@ final class AbstractValueToNullReturnValueTest extends TestCase
...
      * @var AbstractValueToNullReturnValue&MockObject
      */
-    private $testSubject;
+    private MockObject $testSubject;
 
     protected function setUp(): void
     {
-        $this->testSubject = $this->getMockBuilder(AbstractValueToNullReturnValue::class)
-            ->disableOriginalConstructor()
-            ->getMockForAbstractClass();
```
sidz, line 59: see https://github.com/sebastianbergmann/phpunit/issues/5241

maks-rafalko (approved):

PR #2052, opened 2025-02-26
Add `cs-check` to `make autoreview`

maks-rafalko:
I always run `make autoreview` before pushing to double check everything is good.

But for some reason, we didn't check code syles (by PHP-CS-Fixer) during `make autoreview`.

This PR adds it.

Also, ignore some folders from whitespaces check script as it's currently broken.


sanmai (approved):
Looking good

sidz (approved):

PR #2072, opened 2025-03-29
Get back `devTools/Dockerfile.json` to pin to it on Makefile

maks-rafalko:
# Problem

When I run `make compile-docker` 2 times in a row, docker builds the image 2 times while only the first run should trigger the build.

# Solution

If found that we used `devTools/Dockerfile***.json` files back in time to pin to them, here: https://github.com/infection/infection/pull/304/files

Now, it seems we depend on `devTools/Dockerfile.json` but it's never generated.

So this PR generates `devTools/Dockerfile.json` after image is built and now the second run doesn't trigger image build anymore.

----

@sanmai I would like your review here as my knowledge of Makefile is limited, and you were the author of this `Dockerfile.json` feature :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -247,7 +247,8 @@ phpunit.xml.dist:
 	touch -c $@
 
 $(DOCKER_FILE_IMAGE): devTools/Dockerfile
-	docker-compose build
+	docker compose build php82
+	docker image inspect infection-php82 >> devTools/Dockerfile.json
 	touch -c $@
 
 tests/benchmark/MutationGenerator/sources: tests/benchmark/MutationGenerator/sources.tar.gz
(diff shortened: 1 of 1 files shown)


sanmai (approved):
In Makefile parlance, there's a target (`DOCKER_FILE_IMAGE` in this case) and a source (`devTools/Dockerfile`). And if the target is newer than the source, then make will assume the target is up to date, and do not attempt to rebuild it. But if you change the source, then make will rebuild the target. So this change makes sense.

Also, you can have multiple sources. For example, to trigger the rebuild if some other file changes.

maks-rafalko:
thank you @sanmai

PR #2074, opened 2025-03-29
Add targets to Makefile to allow running commands in docker in addition to a host machine

maks-rafalko:
My goal is to update `CONTRIBUTION.md` guide to explain how Infection should be developed and checked/tested for contributors, because this file is outdated.

During analyzing it, I found that our makefile docker commands are broken, so I want to make them work, at least the most useful ones:

- `make autoreview-docker`
- `make test-unit-docker`
- `make cs-docker`

### DIFF
--- a/devTools/check_trailing_whitespaces.sh
+++ b/devTools/check_trailing_whitespaces.sh
@@ -28,6 +28,7 @@ files_with_trailing_whitespaces=$(
         -type f \
         -not -name "*.cache" \
         -not -name "*.log" \
+        -not -path "./.DS_Store" \
         -not -path "./.composer/*" \
         -not -path "./.cache/*" \
         -not -path "./.box_dump/*" \
--- a/tests/phpunit/AutoReview/Makefile/MakefileTest.php
+++ b/tests/phpunit/AutoReview/Makefile/MakefileTest.php
@@ -38,17 +38,14 @@ namespace Infection\Tests\AutoReview\Makefile;
 use function array_filter;
 use function array_map;
 use function array_shift;
-use function array_unshift;
 use function array_values;
 use function count;
 use Fidry\Makefile\Rule;
 use Fidry\Makefile\Test\BaseMakefileTestCase;
-use function implode;
 use PHPUnit\Framework\Attributes\CoversNothing;
 use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\Attributes\Group;
 use function Safe\array_replace;
-use function sprintf;
 use function str_ends_with;
 use function str_starts_with;
 use function substr;
@@ -187,37 +184,6 @@ final class MakefileTest extends BaseMakefileTestCase
         ];
     }
 
-    public function test_all_docker_test_targets_are_properly_declared(): void
-    {
-        $testRules = self::getTestRules(true);
-
-        foreach ($testRules as $rule) {
-            $target = $rule->getTarget();
-            $prerequisites = $rule->getPrerequisites();
-
-            $subTestTargets = self::getDockerSubTestTargets($target, $testRules);
-
-            if ($subTestTargets === []) {
-                continue;
-            }
-
-            if ($target === 'test-docker') {
-                array_unshift($subTestTargets, 'autoreview');
-            }
-
-            $this->assertSame(
-                $subTestTargets,
-                $prerequisites,
-                sprintf(
-                    'Expected the pre-requisite of the "%s" target to be "%s". Found "%s" instead',
-                    $target,
-                    implode(' ', $subTestTargets),
-                    implode(' ', $prerequisites),
-                ),
-            );
-        }
-    }
-
     public function test_the_test_target_runs_all_the_tests(): void
     {
         $testRules = self::getTestRules(false);
@@ -231,19 +197,6 @@ final class MakefileTest extends BaseMakefileTestCase
         $this->assertEqualsCanonicalizing($rootTestTargets, $testPrerequisites);
     }
 
-    public function test_the_docker_test_target_runs_all_the_tests(): void
-    {
-        $testRules = self::getTestRules(true);
-
-        // Exclude itself
-        $testPrerequisites = array_shift($testRules)->getPrerequisites();
-
-        $rootTestTargets = self::getRootTestTargets($testRules, 2);
-        array_unshift($rootTestTargets, 'autoreview');
-
-        $this->assertEqualsCanonicalizing($rootTestTargets, $testPrerequisites);
-    }
-
     protected static function getMakefilePath(): string
     {
         return self::MAKEFILE_PATH;
@@ -251,6 +204,7 @@ final class MakefileTest extends BaseMakefileTestCase
 
     protected function getExpectedHelpOutput(): string
     {
+        // generate it with `make | tee output.txt`
         return <<<'EOF'
             [33mUsage:[0m
               make TARGET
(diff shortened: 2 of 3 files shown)


tests/phpunit/AutoReview/Makefile/MakefileTest.php
```diff
@@ -187,37 +184,6 @@ public static function rootTestTargetProvider(): iterable
         ];
     }
 
-    public function test_all_docker_test_targets_are_properly_declared(): void
```
maks-rafalko, line 190: Looking for ~10 minutes to this test and the second one that is removed, I didn't get it and it looks like it works incorrectly making incorrect assumption regarding targets naming.

@theofidry do you have any idea what does it do and why it fails with the current Makefile?

devTools/check_trailing_whitespaces.sh
```diff
@@ -28,6 +28,7 @@ files_with_trailing_whitespaces=$(
         -type f \
         -not -name "*.cache" \
         -not -name "*.log" \
+        -not -path "./.DS_Store" \
```
theofidry, line 31: is it really related?

tests/phpunit/AutoReview/Makefile/MakefileTest.php
theofidry: I'll comment in each tests

tests/phpunit/AutoReview/Makefile/MakefileTest.php
```diff
@@ -231,26 +197,14 @@ public function test_the_test_target_runs_all_the_tests(): void
         $this->assertEqualsCanonicalizing($rootTestTargets, $testPrerequisites);
     }
 
-    public function test_the_docker_test_target_runs_all_the_tests(): void
```
theofidry, line 234: This test was to ensure the (phony) `test-docker` executes all (phony) tests, i.e. that`test-docker` executes:

- `autoreview`
- `test-unit-docker`
- `test-e2e-docker`
- `test-infection-docker`
- and if added any `test-x-docker`

tests/phpunit/AutoReview/Makefile/MakefileTest.php
```diff
@@ -231,26 +197,14 @@ public function test_the_test_target_runs_all_the_tests(): void
...
-    {
-        $testRules = self::getTestRules(true);
-
-        // Exclude itself
-        $testPrerequisites = array_shift($testRules)->getPrerequisites();
```
theofidry, line 239: this is getting all the pre-requesites of `test-docker`

tests/phpunit/AutoReview/Makefile/MakefileTest.php
```diff
@@ -231,26 +197,14 @@ public function test_the_test_target_runs_all_the_tests(): void
...
-
-        $rootTestTargets = self::getRootTestTargets($testRules, 2);
```
theofidry, line 241: this is getting all the `test-*-*` (phony) targets

tests/phpunit/AutoReview/Makefile/MakefileTest.php
theofidry: This test ensures that we follow the pattern that for each phony target `test-*-docker`, it contains all `test-*-*-docker`.

For instance:

- `test-docker` contains `test-unit-docker`, `test-e2e-docker` and `test-infection-docker`.
- `test-unit-docker` contains `test-unit-82-docker`, `test-unit-83-docker`, etc.

theofidry (commented):
@maks-rafalko I couldn't check which ones were failing, but IIRC there was some hardcoding to deal with `autoreview` that was not really following the conventions at that time

devTools/check_trailing_whitespaces.sh
maks-rafalko: well, implicitly, without this `make autoreview-docker` doesn't pass

theofidry:
If I may, I think part of the confusion is that this PR does two things:

- move some targets that were not previously ran in docker in docker (the title matches this)
- remove some docker targets*

*: previously we had for instance `test-unit-docker` that was executing `test-unit-82-docker` & `test-unit-83-docker`. At some point one of those was removed and we didn't add 8.4, I can't remember if this was a conscious decision to simply things or if it just happened as we dropped PHP 8.2 support.

In any case, removing `test-unit-82-docker` in favour of `test-unit-docker` (and similarly for others) likely requires some changes in that test.

tests/phpunit/AutoReview/Makefile/MakefileTest.php
maks-rafalko: here we have a problem I think, because we have `autoreview` (cs, phpstan, etc.) and it's pair `autoreview-docker`. 

At the same time we have `test-autoreview` (which runs PHPUnit with autoreview group) and its par `test-autoreview-docker`.

So this convention is not followed. We should either rename `autoreview` command to something else or update/remove this test?

maks-rafalko:
for some reason I can't comment under you comment, so will leave here:

> This test ensures that we follow the pattern that for each phony target test-*-docker, it contains all test-*-*-docker.

here is the problem I think, because we have `autoreview` (cs, phpstan, etc.) and it's pair `autoreview-docker`. 

At the same time we have `test-autoreview` (which runs PHPUnit with autoreview group) and its par `test-autoreview-docker`.

So this convention is not followed. We should either rename `autoreview` command to something else or update/remove this test?

maks-rafalko:
> At some point one of those was removed and we didn't add 8.4, I can't remember if this was a conscious decision to simply things or if it just happened as we dropped PHP 8.2 support.

I don't think we need all the supported versions to be tested locally, we have CI for that.

My goal is simple - to have dockerized targets only for "base" supported version, which is 8.2 atm.

This is needed e.g. for those who don't have 8.2 installed locally (like me)

theofidry:
I am just pointing out it didn't use to be like that hence the conventions, because we didn't want that if we add `test-unit-84-docker`, that we do not forget to add it to `test-unit-docker`.

I agree with the simplification though, IMO if you hit an issue with a specific version, it shouldn't be hard to adjust the PHP docker image used locally to run the same command as the makefile, or even change the makefile temporarily

theofidry:
>So this convention is not followed. We should either rename autoreview command to something else or update/remove this test?

I think it's fine to update the test, the test already had a special case handling for autoreview anyway

theofidry:
Revisiting this: I wonder if it still makes sense.

In todays age I think most things should be executed within a sandbox. It is not incompatible with using the docker commands, but it is already a controlled environment, so to me it feels a bit computationally wasteful and slower to execute those docker commands within the sandbox.

theofidry:
Ported the actual fixes in https://github.com/infection/infection/pull/3532.

maks-rafalko:
since you revisited it, what is left after porting? Do we need this PR?

theofidry:
@maks-rafalko the part left out is the one of the title: make everything run in docker and in docker only.

I ported to _allow_ to run everything in docker, but I think being able to run things outside of it is still valuable (better performance), especially if you already have a containerisation or sorts in place (like a docker sandbox).

I did not close it because I'm not sure how strongly you feel about it

maks-rafalko:
> but I think being able to run things outside of it is still valuable

oh, I think this is an ambiguous title, I didn't mean we should run everything in docker only, I meant we should give an ability for contributors to run everything (like all needed targets to check the code) in Docker *if they want to*

If that works now, we can close this one

maks-rafalko:
ok, I just ran `make autoreview-docker` on master branch and it works. This is what I wanted. Closing

PR #2084, opened 2025-04-26
Generate a prettified JSON5

arokettu:
This PR:

- [x] Adds new feature ...
- [ ] Covered by tests
- [ ] Doc PR: doesn't need to be documented

A small contribution should you choose to accept it. 

Current scope: quoteless keys. Possible future scope: default comments like.

```php
        /*
         * Explicitly add the default profile to the list of mutators, as even if it is
         * empty by default, it is not. If it would actually contain the default profile
         * by default, it would make the profiles feature more obvious to configure.
         */
        $configObject->mutators = [
            // this commentAfter will appear in JSON5
            '@default' => new CommentDecorator(true, commentAfter: 'Explicitly add the default profile'),
        ];
```

the generated config would look like:

```json5
{
    $schema: "https://raw.githubusercontent.com/infection/infection/master/resources/schema.json",
    source: {
        directories: [
            "src",
        ],
    },
    logs: {
        text: "reports/infection.log",
    },
    mutators: {
        '@default': true, // Explicitly add the default profile
    },
}
```

P.S. I can't find if your config generation is tested currently

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,6 +48,7 @@
         "ext-libxml": "*",
         "ext-mbstring": "*",
         "composer-runtime-api": "^2.0",
+        "arokettu/json5-builder": "^1.0",
         "colinodell/json5": "^2.2 || ^3.0",
         "composer/xdebug-handler": "^2.0 || ^3.0",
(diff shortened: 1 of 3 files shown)


arokettu:
> I can't find if your config generation is tested currently

I see now

PR #2087, opened 2025-05-08
Do not fail Infection when project has 0 tests

maks-rafalko:
Infection should not fail if a project has no tests at all.

Case: when the project is being bootstrapped, it can have 0 tests. Imagine a case when a project template is created.

PHPUnit returns 0 exit code, saying "No tests executed!".

Infection should not fail in this case as well, so that CI passes.

### DIFF
--- a/src/TestFramework/Coverage/XmlReport/IndexXmlCoverageParser.php
+++ b/src/TestFramework/Coverage/XmlReport/IndexXmlCoverageParser.php
@@ -97,6 +97,12 @@ class IndexXmlCoverageParser
      */
     private static function assertHasExecutedLines(SafeDOMXPath $xPath, bool $isForGitDiffLines): void
     {
+        $testsNode = $xPath->query('/phpunit/project/tests')->item(0);
+
+        if ($testsNode !== null && !$testsNode->hasChildNodes()) {
+            throw NoTestsInProject::create();
+        }
+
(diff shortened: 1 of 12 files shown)


src/TestFramework/Coverage/XmlReport/IndexXmlCoverageParser.php
```diff
@@ -97,6 +97,12 @@ private function parseNodes(
      */
     private static function assertHasExecutedLines(SafeDOMXPath $xPath, bool $isForGitDiffLines): void
     {
+        $testsNode = $xPath->query('/phpunit/project/tests')->item(0);
+
+        if ($testsNode !== null && !$testsNode->hasChildNodes()) {
+            throw NoTestsInProject::create();
```
maks-rafalko, line 103: Unfortunately, this is not an ideal solution, because it conflicts with the code below, when no tests executed by some mistake in `@covers()` or `--group` or `--filter`.

The only one correct way of determining if PHPUnit "sees" test files in the project is to use `phpunit --list-tests`, but I don't like to run one more process each time to cover this rare case, so I think it doesn't worth it.

Bootstrap project can have an example test and then run Infection, and that's it.

maks-rafalko:
closing due to https://github.com/infection/infection/pull/2087#discussion_r2080108776

theofidry:
@maks-rafalko do we have an issue where we could collect what we need from a PHPUnit report? This way when we need to talk with Sebastian we know what constraints we have

PR #2090, opened 2025-05-17
Introduce dead code detector

maks-rafalko:
See https://github.com/shipmonk-rnd/dead-code-detector

Result: no real dead code is found. A bunch of false-positives (reported to the plugin repository) and 2 edge cases - added comments to `phpstan.neon`.

----

- Related to https://github.com/infection/infection/pull/335
- [x] For now in draft as it has false-positives that [I want to ask about](https://github.com/shipmonk-rnd/dead-code-detector/issues/209)
- [x] blocked by https://github.com/infection/infection/pull/2091
- [x] Reported bug/feature request https://github.com/shipmonk-rnd/dead-code-detector/issues/211

### DIFF
--- a/devTools/phpstan-baseline.neon
+++ b/devTools/phpstan-baseline.neon
@@ -37,6 +37,12 @@ parameters:
 			path: ../src/Logger/Html/StrykerHtmlReportBuilder.php
 
 		-
+			message: '#^Unused Infection\\Mutant\\MutantExecutionResult\:\:getOriginalStartingColumn$#'
+			identifier: shipmonk.deadMethod
+			count: 1
+			path: ../src/Mutant/MutantExecutionResult.php
+
+		-
 			message: '#^Parameter \#1 \$array of function array_intersect expects an array of values castable to string, list\<PhpParser\\Node\\Expr\|string\> given\.$#'
 			identifier: argument.type
 			count: 1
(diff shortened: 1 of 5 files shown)


sanmai:
I think Psalm also had dead code detection.

devTools/phpstan-baseline.neon
```diff
@@ -264,12 +264,24 @@ parameters:
 			count: 2
 			path: ../tests/phpunit/AutoReview/EnvVariableManipulation/EnvTestCasesProvider.php
 
+		-
+			message: '#^Unused Infection\\Tests\\AutoReview\\EnvVariableManipulation\\EnvTestCasesProvider\:\:envTestCaseTupleProvider$#'
+			identifier: shipmonk.deadMethod
+			count: 1
+			path: ../tests/phpunit/AutoReview/EnvVariableManipulation/EnvTestCasesProvider.php
+
```
maks-rafalko, line 272: all issues added to baseline are now false-positived, reported https://github.com/shipmonk-rnd/dead-code-detector/issues/211

maks-rafalko:
> I think Psalm also had dead code detection.

- personally, I found this one is more user-friendly and flexible. For example, if you make change `findUnusedCode` to `true` in https://github.com/infection/infection/blob/8ba127848d901c76b72e5719cb157decfd4a2559/psalm.xml#L10, then it gives many false-positives like unused private constructors, unused methods (when created for tests purposes), etc. This is probably we are ignoring many folders in `psalm.xml` but anyway
- there is a comparison with othe alternatives, where we can see that this plugin is quite powerful: https://github.com/shipmonk-rnd/dead-code-detector?tab=readme-ov-file#comparison-with-tomasvotrubaunused-public
- this plugin is used by PHPStan itself, so I consider it as a good choice because of this fact

devTools/phpstan-baseline.neon
maks-rafalko: this has been fixed, not relevant anymore

janedbal:
You can deny those usages even on level 8 with [forbidMethodCallOnMixed](https://github.com/shipmonk-rnd/phpstan-rules?tab=readme-ov-file#forbidmethodcallonmixed)

maks-rafalko:
Thank you, will have a look at it.

PR #2091, opened 2025-05-19
Use one phpstan config for `src` and `tests` folders

maks-rafalko:
Back in 2018, configs were split to src and tests folder because we wanted to use `level=max` for `src`: https://github.com/infection/infection/pull/335

Now, split configs for `src` and `tests` folders cause an issue with Dead Code detector: https://github.com/shipmonk-rnd/dead-code-detector/issues/209#issuecomment-2889900927

Moreover, using level=max for tests will also be better. If there will be some checks that are redundant for `tests` folder, we will ignore them.

This unblocks https://github.com/infection/infection/pull/2090


sidz (approved):

maks-rafalko:
thank you @sidz

Issue #157, opened 2018-02-05
Support multiple test frameworks


BackEndTea:
Currently infection support PHPUnit, PhpSpec, and in the future Codeception #143. There is also interest in adding Behat #147.

A lot of projects use a combination of test frameworks. They may use PHPUnit for the unit tests, and Behat for feature/e2e tests. So if you ran infection on those projects, you'd like infection to use all the test frameworks possible.
<details>
 <summary>configuration</summary>
 
Currently `infection.json` has the following option

```json
"phpUnit": {
    "configDir": "app",
    "customPath": "\/path\/to\/phpunit-6.1.phar"
}
```

I would propose to put this config within a `testframework` option. Not having that set would default it to standard phpunit. It could look something like this:

```json
"testFrameworks": {
  "phpUnit": {
    "configDir": "app",
     "customPath": "\/path\/to\/phpunit-6.1.phar"
  },
"phpSpec":  true,
"behat": true
}
```

So just setting it with true would enable the test-framework with default option, and additional configs for the framework can be supplied the same way as phpunit.
</details>

<details> 
 <summary>implementation</summary>
This change is going to have a few challenges.

The simpelest way is to simply run all possible tests in sequence, so first try the PhpUnit tests, if those all succeed try the phpSpec tests, and so on.

A nice 'features' for this would be to be able to set different time outs for different test frameworks. ( I expect my behat tests to take longer than my PhpUnit tests). 
</details>

Implementing this will probably be a lot of work, so any ideas etc are welcome.

maks-rafalko:
for those who upvoted it - what are your use-cases? what testframeworks (or other tools) would you like to run in a chain?

theofidry:
I already talked about this with you @maks-rafalko but adding it here for visibility.

Personally I often use both PHPUnit and Behat.

Also while working on the PoC in #2272, I think we could benefit from having the test framework API flexible enough to integrate PHPStan there as a regular test framework rather than the very specific and dedicated coupling we introduced so far.

In the end I think the current integration of PHPStan is fundamentally not very PHPStan specific (you could integrate Psalm in a very similar fashion) and having the ability to combine testing frameworks would allow to integrate it seamlessly*.

*: It does warrant some changes in the test framework API however.

theofidry:
Closing in favour of #2863.

PR #2096, opened 2025-05-29
Introduce MutantProcessContainer to be able to run N processes for each Mutant (like PHPUnit, PHPStan, Psalm, any other "killer")

maks-rafalko:
Introduce MutantProcessContainer to be able to run N processes for each Mutant (like PHPUnit, PHPStan, Psalm, any other "killer")

Previous implementation is on the left on the screenshot, new implementation is on the right.

image

Why is it needed?

Currently, we have the only one "mutant killer" process - PHPUnit. We run it for every single Mutant. Now, with integration of static analysis (PHPStan in particular), we need to run _more mutant killer(s)_ if PHPUnit doesn't kill a Mutant (so it's _Escaped_).

But our `ParallelProcessRunner` is designed to only run "PHPUnit processes" in parallel (which correspond to `MutantProcess` class objects.

So now, we pass not just one "PHPUnit process", but we create a "container" of processes which can have:

- PHPUnit process to kill mutant
- then if it's escaped - PHPStan process
- the if it's escaped - _any other process_

The final idea is if one process inside a `MutantProcessContainer` doesn't kill a Mutant, we run the next _killer process_ (in our case this will be opt-in PHPStan) to try to kill this Mutant. The currently introduced design allows to create as many "killers" as possible, so that the implementation is not limited to PHPStan at all.

Implicitly related and implements https://github.com/infection/infection/issues/157


src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -164,31 +167,39 @@ private function tryToFreeNotRunningProcess(): ?MutantProcess
             }
 
             if (!$process->isRunning()) {
+                $isMutantEscaped = $mutantProcessContainer->getCurrentMutantProcessDetectionStatus() === DetectionStatus::ESCAPED;
```
theofidry, line 170: isEscapedMutant?

src/Process/MutantProcessContainer.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        return $this->getCurrentMutantProcess()->getMutantExecutionResult()->getDetectionStatus();
+    }
+
+    public function hasNextProcessToKillMutant(): bool
+    {
+        return array_key_exists($this->currentProcessIndex, $this->lazyMutantProcessCreators);
+    }
+
+    public function buildNextProcessToKillMutant(): MutantProcess
```
theofidry, line 74: "create" rather than "build"?

src/Process/MutantProcessContainer.php
```diff
@@ -0,0 +1,91 @@
...
+        MutantProcess $phpUnitMutantProcess,
+        /**
+         * @var list<LazyMutantProcessCreator>
+         */
+        private readonly array $lazyMutantProcessCreators,
+    ) {
+        $this->processes[] = $phpUnitMutantProcess;
+    }
+
+    public function getCurrentMutantProcessDetectionStatus(): string
```
theofidry, line 64: when reading this class it is clear to me that it is a holder of processes, hence I feel like those methods could be simplified:

- `getCurrentDetectionStatus()`
- `hasNext()`
- `createNext()` or even `next()`
- `getCurrent()`

This would make the API a bit lighter to read, and would also get closer to the PHP Iterator API which is kind of what this class is, an iterator for mutant processes

src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -164,31 +167,39 @@ private function tryToFreeNotRunningProcess(): ?MutantProcess
...
+
+                if ($isMutantEscaped && $mutantProcessContainer->hasNextProcessToKillMutant()) {
+                    $newMutantProcess = $mutantProcessContainer->buildNextProcessToKillMutant();
+
+                    // TODO add it to the $bucket and run new process; -> in the next PR
+                }
+
```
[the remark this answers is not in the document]
theofidry: Maybe instead of exposing this, you could have the result being `DetectionStatus|Process`, and if it is a process then you need to add it to the bucket again?

In other words you don't need to expose to `ParallelProcessRunner` the fact that it is escaped or not: this is up to the container to decide if that was enough or not.

theofidry (approved):
👍 nice to see progress there!

maks-rafalko:
@sanmai I see you added yourself as a reviewer. I need to merge it as I don't want to rebase 3 levels of PRs, it's getting too time-consuming.

But please, feel free to leave any comment after your review, I will answer and fix in separate PRs.

sanmai:
Thanks, please go ahead! I'm keeping an eye on this, but yet to find the time for a deep dive :crossed_fingers:

PR #2097, opened 2025-05-30
Add new processes to kill mutant to the queue if PHPUnit doesn't kill a Mutant

maks-rafalko:
Follow-up for https://github.com/infection/infection/pull/2096

If in the `MutantProcessContainer`, the first and required PHPUnit process doesn't kill a mutant, build the next "killer process" and run it in parallel to try to kill the same Mutator.

Notes:

- `ParallelProcessRunner` is reused - we need to queue processes as soon as they are created so they are executed in parallel

----

How I've tested it:

in `Container.php`, I've added:

```diff
MutantProcessContainerFactory::class => static fn (self $container): MutantProcessContainerFactory => new MutantProcessContainerFactory(
    $container->getTestFrameworkAdapter(),
    $container->getConfiguration()->getProcessTimeout(),
    $container->getMutantExecutionResultFactory(),
    [
+        new class($container->getMutantExecutionResultFactory()) implements LazyMutantProcessCreator {
+            public function __construct(private MutantExecutionResultFactory $mutantExecutionResultFactory)
+            {
+            }
+
+            public function createMutantProcess(Mutant $mutant): MutantProcess
+            {
+                return new MutantProcess(
+                    new Process(['echo', '1']),
+                    $mutant,
+                    $this->mutantExecutionResultFactory,
+                );
+            }
+        },
    ],
),
```

and this process is executed now for any Escaped mutant.

### DIFF
--- a/devTools/phpstan-baseline.neon
+++ b/devTools/phpstan-baseline.neon
@@ -121,6 +121,12 @@ parameters:
 			path: ../src/PhpParser/Visitor/ReflectionVisitor.php
 
 		-
+			message: '#^Parameter &\$bucket by\-ref type of method Infection\\Process\\Runner\\ParallelProcessRunner\:\:fillBucketOnce\(\) expects array\<Infection\\Process\\MutantProcessContainer\>, array\<Infection\\Process\\MutantProcessContainer\|null\> given\.$#'
+			identifier: parameterByRef.type
+			count: 1
+			path: ../src/Process/Runner/ParallelProcessRunner.php
+
+		-
 			message: '#^Parameter \#2 \$paths of method Composer\\Autoload\\ClassLoader\:\:setPsr4\(\) expects list\<string\>\|string, array\<string\> given\.$#'
 			identifier: argument.type
 			count: 1
--- a/src/Process/Runner/ParallelProcessRunner.php
+++ b/src/Process/Runner/ParallelProcessRunner.php
@@ -59,9 +59,18 @@ final class ParallelProcessRunner implements ProcessRunner
     private const NANO_SECONDS_IN_MILLI_SECOND = 1_000_000;
 
     /**
+     * Here we store "next" killer processes, created if a PHPUnit process doesn't kill a Mutant
+     * For example: static analysis process to try to kill a Mutant
+     *
+     * @var array<int, MutantProcessContainer>
+     */
+    private array $nextMutantProcessKillerContainer = [];
+
+    /**
      * @var array<int, IndexedMutantProcessContainer>
      */
-    private array $runningProcesses = [];
+    private array $runningProcessContainers = [];
+
     /**
      * @var array<int, int>
      */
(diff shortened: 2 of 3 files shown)


devTools/phpstan-baseline.neon
```diff
@@ -120,6 +120,12 @@ parameters:
 			count: 1
 			path: ../src/PhpParser/Visitor/ReflectionVisitor.php
 
+		-
+			message: '#^Parameter &\$bucket by\-ref type of method Infection\\Process\\Runner\\ParallelProcessRunner\:\:fillBucketOnce\(\) expects array\<Infection\\Process\\MutantProcessContainer\>, array\<Infection\\Process\\MutantProcessContainer\|null\> given\.$#'
```
maks-rafalko, line 124: not sure why PHPStan doesn't understand that the array is not empty..

sidz: I believe that's because `array_shift` may return null on line https://github.com/infection/infection/pull/2097/files#diff-b6afb6bb710fec614bf0d83522b6c12084149181b26b2551de2713e3d74019d6R224

Have you tried to explicitly defined `@param-out` ? 

FYI: https://phpstan.org/blog/enhancements-in-handling-parameters-passed-by-reference#enforcing-assigned-type

staabm: minimal reproduced in https://phpstan.org/r/f41f03a5-778a-41fb-aa65-325d2aaa906e

and created a upstream issue: https://github.com/phpstan/phpstan/issues/13093

staabm: will be fixed in the upcoming phpstan release with https://github.com/phpstan/phpstan-src/pull/4036

src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -115,13 +124,13 @@ public function run(iterable $processContainers): iterable
 
             Assert::integer($threadIndex, 'Thread index can not be null.');
 
-            $this->startProcess($process, $threadIndex);
+            $this->startProcess($mutantProcessContainer, $threadIndex);
```
staabm: I think the infection main process is running with `XDEBUG_MODE=coverage` and it could be useful to pass `XDEBUG_MODE=off` into subprocess (like PHPStan etc), to reduce overhead in the subprocesses

maks-rafalko: we reload Infection process to disable Xdebug here thanks to `XdebugHandler`: 

https://github.com/infection/infection/blob/77863c38b661a2aca2d12113be694f3412bef0ba/src/Command/RunCommand.php#L586

so the main process of Infection is executed without Xdebug.

When "initial tests run" PHPUnit process needs Xdebug, we enable it back just for this sub-process:

https://github.com/infection/infection/blob/38991a32f17608df97848403f23921431c9a3842/src/Process/Factory/InitialTestsRunProcessFactory.php#L64

So, I assume when we just run PHPStan as a subprocess, it should be executed without Xdebug. If you know some easy way on how to check whether Xdebug is enabled on PHPStan (like enabling debug info or something like that), pleaase let me know, I will double check.

staabm: When phpstan is invoked with --xdebug it will send out a warning in case xdebug is not loaded

maks-rafalko:
@sanmai I would be happy to discuss it. The main struggling for me here is it's very hard to  write reproducible tests, but I had many issues with ParallelProcessRunner in https://github.com/infection/infection/pull/2098

For example:

- when you run Infection with PHPStan enabled with --threads=1 - everything works
- when you run it with `--threads=max` - it didn't work unless I rewrote `while` / `do-while` cycles

And I wasn't able to reproduce the issue in unit tests yet, because in reality we have more factors: time that processes take, random order, etc.

Anyway, with this final implementation, seems like it works with both cases above. I'm pretty sure that the algorithm itself can be optimized, but I decided to postpone (at least myself) as I wasted too much time on this optimizations without luck, decided to move further.

src/Process/Runner/ParallelProcessRunner.php
maks-rafalko: so, I've tried to add `--xdebug` to the place [where we programmatically run PHPStan in a sub-process](https://github.com/infection/infection/pull/2098/files#diff-6e1a926e017848b9689bf314bfab37024a717cf4582311f659d29be04fd80bf9R83), it outputs:

```
   ! [NOTE] You are running with "--xdebug" enabled, but the Xdebug PHP extension 
   !        is not active. The process will not halt at breakpoints. 
```

so Xdebug is not enabled there, which is good.

staabm: Yep pefect. Thanks for checking

PR #2098, opened 2025-06-01
Kill mutants by Static Analysis (PHPStan) if they are escaped and not killed by PHPUnit tests

maks-rafalko:
This is the first **working** implementation of Infection + PHPStan integration.

What it does for now:

- adds `--static-analysis-tool=phpstan` opt-in option. Later, we will have `phpstan|psalm|xxx` values allowed for this option. For now - just PHPStan
- runs "initial PHPStan run" the same way we run "initial tests run" to check if PHPStan passes, and to **warm up the cache**
- it executes PHPStan as a sub-process for a Mutant **only if PHPUnit doesn't kill it** 
- it runs PHPStan processes **in parallel** inside `ParallelProcessRunner`, leveraging `MutantProcessContainer` from https://github.com/infection/infection/pull/2096
- it has e2e test to ensure PHPStan catches errors (see below)
- it works with the very simple setup for now - PHPStan should be installed in the `vendor/bin/phpstan` (no ability to set custom paths, will be done in other PRs), and phpstan config path is not configurable yet
- it mimics folder/files structure from our Test Frameworks - we also have here interfaces to later support Psalm (or anything else from Static Analysis tools)

# How to play with it

```
# use our docker container
docker compose run --rm --remove-orphans php82 bash 

# go to e2e test
cd tests/e2e/PHPStan_Integration

composer install

# run infection with phpstan
../../../bin/infection --static-analysis-tool=phpstan -s --log-verbosity=all --debug -vvv --threads=max
```

after that, analyze `infection.text.log` in the same folder.

Examples of killed mutant by PHPStan:

```shell
5) /opt/infection/tests/e2e/PHPStan_Integration/src/SourceClass.php:38    [M] ArrayItemRemoval

@@ @@
    /**
     * @return non-empty-array<int, string>
     */
     private function returnNonEmptyArray(): array
     {
-        return ['test'];
+        return [];
     }
 }

$ '/opt/infection/tests/e2e/PHPStan_Integration/vendor/bin/phpstan' '--tmp-file=/opt/infection/tests/e2e/PHPStan_Integration/./infection/mutant.2babe6391fe7eecf9541b58d780d33cb.infection.php' '--instead-of=/opt/infection/tests/e2e/PHPStan_Integration/src/SourceClass.php' '--error-format=json' '--no-progress' '-vv'

Errors: Method PHPStan_Integration\\SourceClass::returnNonEmptyArray() should return 
            non-empty-array<int, string> but returns array{}.
  
  Note: Using configuration file /opt/infection/tests/e2e/PHPStan_Integration/phpstan.neon.
  Result cache restored. 1 file will be reanalysed.
  Result cache was not saved because of --tmp-file and --instead-of CLI options passed (editor mode).
  Elapsed time: 1 second
  Used memory: 74 MB
```

Notes:

- there are many TODOs but I would like to have an early review and I can use this PR as a reference for PHPStan team to start some analysis and to run it with PHPStan integrated with Infection **natively**.

### DIFF
--- a/src/FileSystem/Finder/Exception/FinderException.php
+++ b/src/FileSystem/Finder/Exception/FinderException.php
@@ -72,8 +72,7 @@ final class FinderException extends RuntimeException
     {
         return new self(
             sprintf(
-                'Unable to locate a %s static analysis executable on local system. Ensure that %s is installed and available.',
-                $testFrameworkName,
+                'Unable to locate a %1$s static analysis executable on local system. Ensure that %1$s is installed and available.',
                 $testFrameworkName,
             ),
         );
--- a/src/Process/Runner/InitialStaticAnalysisRunner.php
+++ b/src/Process/Runner/InitialStaticAnalysisRunner.php
@@ -60,9 +60,7 @@ readonly class InitialStaticAnalysisRunner
 
         $this->eventDispatcher->dispatch(new InitialStaticAnalysisRunWasStarted());
 
-        $process->run(function (): void {
-            $this->eventDispatcher->dispatch(new InitialStaticAnalysisSubStepWasCompleted());
-        });
+        $process->run(fn () => $this->eventDispatcher->dispatch(new InitialStaticAnalysisSubStepWasCompleted()));
 
         $this->eventDispatcher->dispatch(new InitialStaticAnalysisRunWasFinished($process->getOutput()));
 


src/Command/RunCommand.php
```diff
@@ -177,6 +179,16 @@ protected function configure(): void
                 ),
                 Container::DEFAULT_TEST_FRAMEWORK,
             )
+            ->addOption(
+                self::OPTION_STATIC_ANALYSIS_TOOL,
+                null,
+                InputOption::VALUE_REQUIRED,
+                sprintf(
+                    'Name of the Static Analysis tool to use ("%s")',
+                    implode('", "', TestFrameworkTypes::getTypes()),
```
staabm:
```suggestion
                    implode('", "', StaticAnalysisToolTypes::getTypes()),
```

src/Event/Subscriber/CiInitialStaticAnalysisRunConsoleLoggerSubscriber.php
```diff
@@ -0,0 +1,74 @@
...
+use Infection\Event\InitialStaticAnalysisRunWasStarted;
+use Infection\StaticAnalysis\StaticAnalysisToolAdapter;
+use InvalidArgumentException;
+use function sprintf;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @internal
+ */
+final readonly class CiInitialStaticAnalysisRunConsoleLoggerSubscriber implements EventSubscriber
```
theofidry, line 47: what is CI specific here?

src/Event/Subscriber/InitialStaticAnalysisRunConsoleLoggerSubscriberFactory.php
```diff
@@ -0,0 +1,63 @@
...
+        private bool $debug,
+        private StaticAnalysisToolAdapter $staticAnalysisToolAdapter,
+    ) {
+    }
+
+    public function create(OutputInterface $output): EventSubscriber
+    {
+        return $this->skipProgressBar
+            ? new CiInitialStaticAnalysisRunConsoleLoggerSubscriber($this->staticAnalysisToolAdapter, $output)
+            : new InitialStaticAnalysisRunConsoleLoggerSubscriber(
```
theofidry, line 57: maybe to distinguish the two, rather than CI & non-CI, this could be interactive vs non-interactive, as you usually do not want a progress bar in a non-interactive environment

src/FileSystem/Finder/Exception/FinderException.php
```diff
@@ -68,6 +68,17 @@ public static function testFrameworkNotFound(string $testFrameworkName): self
         );
     }
 
+    public static function staticAnalysisToolNotFound(string $testFrameworkName): self
+    {
+        return new self(
+            sprintf(
+                'Unable to locate a %s static analysis executable on local system. Ensure that %s is installed and available.',
```
theofidry:
```suggestion
                'Unable to locate a %1$s static analysis executable on local system. Ensure that %1$s is installed and available.',
               $testFrameworkName,
```

src/FileSystem/Finder/StaticAnalysisToolExecutableFinder.php
```diff
@@ -0,0 +1,195 @@
...
+
+        if ($vendorPath !== null) {
+            $pathName = getenv('PATH') !== false ? 'PATH' : 'Path';
+            putenv($pathName . '=' . $vendorPath . PATH_SEPARATOR . getenv($pathName));
+        }
+    }
+
+    private function findComposer(): string
+    {
+        return (new ComposerExecutableFinder())->find();
```
theofidry, line 132: don't we have this executable elsewhere too though? Doesn't that mean its cached result will not be used?

src/Process/Factory/InitialStaticAnalysisProcessFactory.php
```diff
@@ -0,0 +1,59 @@
...
+    public function __construct(
+        private StaticAnalysisToolAdapter $staticAnalysisToolAdapter,
+    ) {
+    }
+
+    public function createProcess(): Process
+    {
+        return new Process(
+            command: $this->staticAnalysisToolAdapter->getInitialRunCommandLine(),
+            timeout: null, // Ignore the default timeout of 60 seconds
```
theofidry, line 56: don't we want a timeout?

src/Process/Runner/InitialStaticAnalysisRunner.php
```diff
@@ -0,0 +1,71 @@
...
+    ) {
+    }
+
+    public function run(): Process
+    {
+        $process = $this->initialStaticAnalysisProcessFactory->createProcess();
+
+        $this->eventDispatcher->dispatch(new InitialStaticAnalysisRunWasStarted());
+
+        $process->run(function (): void {
```
theofidry:
```suggestion
        $process->run(
            fn() => $this->eventDispatcher->dispatch(new InitialStaticAnalysisSubStepWasCompleted())
        );
```

?

theofidry (commented):
First diagonal review, looking forward to try it!

src/FileSystem/Finder/StaticAnalysisToolExecutableFinder.php
maks-rafalko: agree. Extracted to https://github.com/infection/infection/issues/2102

src/Event/Subscriber/InitialStaticAnalysisRunConsoleLoggerSubscriberFactory.php
maks-rafalko: Makes sense. However, I did it the same way as we already did it for test grameworks, with the same naming.

I've extracted your suggestion to https://github.com/infection/infection/issues/2103

src/Event/Subscriber/CiInitialStaticAnalysisRunConsoleLoggerSubscriber.php
maks-rafalko: discussed here: https://github.com/infection/infection/pull/2098#discussion_r2123315786

tl;dr: to remove progress bars

src/Process/Factory/InitialStaticAnalysisProcessFactory.php
maks-rafalko: I guess no, because we don't know how much time PHPStan takes.

We have the same `null` value for initial PHPUnit process for ages :)

staabm: I think I have seen in the past that mutants can timeout. when PHPUnit and PHPStan both don't have a timeout, how can a mutant get into a timed'out state?
sorry for this newbie questions :-)

maks-rafalko: you are very welcome with these questions, let me explain.

We have 2 types of processes:

- **"initial run"** (1 process) - this process is executed _before_ we do a Mutation Testing. For PHPUnit we do it to understand that tests pass, otherwise it doesn't make sense to use MT if tests fail. For PHPStan, the same - we run PHPStan before MT to ensure it works as expected, _but we also run it to warm-up the cache!_
  - initial run doesn't need a timeout because tests or PHPStan can last for tens of minutes
- **"mutant process"** (N processes) - here we create PHPUnit and PHPStan sub-processes for each Mutant (for each changed file). 
  - there is a timeout for such processes! It goes from `infection.json5` from `timeout` setting, and by default is 10s. PHPStan also uses this timeout for "mutant processes" and will be marked as "TIMEOUTED" if it lasts longer

So, in this particular line of code we are speaking about "initial run", that's why timeout from config is not applicable here

maks-rafalko:
Merging as I need to move forward. Anyone who wants to review - feel free to leave a comment.

PR #2099, opened 2025-06-01
Rename `MutantExecutionResultFactory` to `TestFrameworkMutantExecutionResultFactory` and add interface

maks-rafalko:
Extracted from https://github.com/infection/infection/pull/2098 to simplify a big PR

Why?

Because for PHPUnit process and for Static Analysis process (PHPStan in particular) we can not use the same ExecutionResult factory, because

- processes exit with different status codes
- PHPStan uses `stderr` for a lot of "internal" things, we can't rely on it
- because we don't need to parse `stdout` for "passed tests" in case of PHPStan
- and so on

So, each MutantProcess inside MutantProcessContainer (see https://github.com/infection/infection/pull/2096) will have different execution result parsers.

### DIFF
--- a/src/Mutant/TestFrameworkMutantExecutionResultFactory.php
+++ b/src/Mutant/TestFrameworkMutantExecutionResultFactory.php
@@ -1,0 +1,125 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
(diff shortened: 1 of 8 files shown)


src/Mutant/TestFrameworkMutantExecutionResultFactory.php
maks-rafalko, line 1: not sure why git shows it as a new file, it's the same file just renamed

PR #2104, opened 2025-06-03
Allow setting PHPStan executable custom path

maks-rafalko:
This is needed when phpstan executable is not located in a common vendor/bin/phpstan

For example, in https://github.com/Roave/BetterReflection they have it installed in `./tools`, and Infection can not find it.

----

This implementation is exactly the same as we have for PHPUnit custom path.

Usage:

```json
{
    "$schema": "../../../resources/schema.json",
    "source": {
        "directories": [
            "src"
        ]
    },
    "phpStan": {
        "customPath": "tools/vendor/bin/phpstan"
    }
}
```


maks-rafalko:
it works: https://github.com/maks-rafalko/BetterReflection/pull/2

UPD: added to the main repo: https://github.com/Roave/BetterReflection/pull/1510

PR #2106, opened 2025-06-04
Support null-safe method calls in MethodCallRemoval mutator

staabm:
This PR:

- [ ] Adds new feature ...
- [x] Covered by tests

While getting used to the codebase I had a look into existing mutators.
If I am not mistaken I think the `MethodCallRemoval` mutator currently misses to work on [null-safe method calls](https://php.watch/versions/8.0/null-safe-operator)

### DIFF
--- a/src/Mutator/Removal/MethodCallRemoval.php
+++ b/src/Mutator/Removal/MethodCallRemoval.php
@@ -78,6 +78,9 @@ final class MethodCallRemoval implements Mutator
             return false;
         }
 
-        return $node->expr instanceof Node\Expr\MethodCall || $node->expr instanceof Node\Expr\StaticCall;
+        return $node->expr instanceof Node\Expr\MethodCall
+            || $node->expr instanceof Node\Expr\NullsafeMethodCall
+            || $node->expr instanceof Node\Expr\StaticCall
+        ;
     }
 }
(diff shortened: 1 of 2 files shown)


staabm:
I just found https://github.com/infection/infection/blob/440b6d88935b3a8ab16510a2331478e1da7ef279/src/Mutator/Operator/NullSafeMethodCall.php which might mean that this PR is not necessary as is

staabm:
Thanks for this great explanation. 

That also gave me the idea that this mutator should only run on calls to functions/methods which have side-effects. 

Maybe infection can built up some more reflection-knowledge and prevent this mutation for methods which are side-effect-free (or utilize phpstan knowledge for it)

maks-rafalko:
> or utilize phpstan knowledge for it

yeah that is a new big topic and feature that we might try in the future, already discussed with Ondrej. After implementing PHPStan processes as "mutant killers" and kill mutant **after** they are generated and escaped, we can try to use PHPStan `Scope`'s knowledge to determine whether particular mutation is needed **before** it's generated. And yeah, this will require to integrate phpstan inside Infection core to make it smarter.

There are many other cases/mutators that will benefit from it, so it's not only for this one.

maks-rafalko:
thank you @staabm

PR #2107, opened 2025-06-04
BaseMutatorTestCase: lint mulitple files at once

staabm:
This PR:

- [ ] Adds new feature ...
- [ ] Covered by tests
- [ ] Doc PR: https://github.com/infection/site/pull/XXX

Support multi-file lint in `BaseMutatorTestCase` to speedup running the testsuite of infection itself.
Before this PR the suite takes ~35 seconds on my machine, while it runs in 31-32 seconds after this PR on PHP 8.3.21

Utilizes https://github.com/php/php-src/pull/10710

PR #2108, opened 2025-06-04
BaseMutatorTestCase: lint source in-process

staabm:
simpler alternative to https://github.com/infection/infection/pull/2107

before this PR running the test-suite took ~35 seconds on my machine, while it now takes ~10-11 seconds

### DIFF
--- a/src/Testing/BaseMutatorTestCase.php
+++ b/src/Testing/BaseMutatorTestCase.php
@@ -39,7 +39,6 @@ use function array_flip;
 use function array_key_exists;
 use function array_shift;
 use function count;
-use function escapeshellarg;
 use function implode;
 use Infection\Mutator\Mutator;
 use Infection\Mutator\ProfileList;
@@ -49,8 +48,10 @@ use Infection\PhpParser\Visitor\MutatorVisitor;
 use const PHP_EOL;
 use PhpParser\NodeTraverser;
 use PHPUnit\Framework\TestCase;
-use function Safe\exec;
 use function sprintf;
+use Throwable;
+use function token_get_all;
+use const TOKEN_PARSE;
 use Webmozart\Assert\Assert;
 
 abstract class BaseMutatorTestCase extends TestCase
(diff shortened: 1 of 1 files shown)


src/Testing/BaseMutatorTestCase.php
```diff
@@ -179,19 +180,25 @@ private function getMutationsFromCode(string $code, array $settings): array
...
+            $tokens = token_get_all($realMutatedCode, TOKEN_PARSE);
+
+            $this->assertTrue(
+                $tokens !== [],
+                sprintf(
+                    'Mutator %s produces invalid code: %s',
+                    $this->mutator->getName(),
+                    $realMutatedCode,
+                ),
+            );
```
staabm, line 193: this assertion is actually always true, but only invoking `token_get_all($realMutatedCode, TOKEN_PARSE)` without checking the result makes PHPStorm error as it thinks the `token_get_all` call is useless.

src/Testing/BaseMutatorTestCase.php
```diff
@@ -179,19 +180,25 @@ private function getMutationsFromCode(string $code, array $settings): array
 
     private function assertSyntaxIsValid(string $realMutatedCode): void
     {
-        exec(
-            sprintf('echo %s | php -l', escapeshellarg($realMutatedCode)),
-            $output,
-            $returnCode,
-        );
```
staabm, line 186: I think this case pretty much demonstrates why infection itself is so slow when running on real projects - it needs to spawn so many processes.

a few months ago I did a [experiment in phpunit](https://github.com/sebastianbergmann/phpunit/pull/5751#issuecomment-2002582219) to use process forking instead of sub-process creation, which showed a 7-8x performance improvement. maybe thats something which can also be applied to infection as a fast-path in case `pcntl_*` is available.

theofidry (approved):
That's neat!

maks-rafalko:
GH action failure is unrelated

PR #2112, opened 2025-06-04
ProtectedVisibility: Don't mutate method in final class

staabm:
In final classes it does not make a difference whether methods are protected of private.


staabm:
I think we could do similar things for anonymous classes, but I was not able to fix it.

```
<?php

namespace ProtectedVisibilityFinal;

$class = new class
{
    protected function fooAnon(int $param, $test = 1) : bool
    {
        echo 1;
        return false;
    }
};
```

staabm:
the anonymous class case is not worth to complicate if further. will leave the PR as is for now.

its ready to be merged from my POV

maks-rafalko:
thank you @staabm

PR #2115, opened 2025-06-05
Reduce the boilerplate with autowiring

sanmai:
This PR:

- [ ] Adds new feature ...
- [ ] Covered by tests

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -152,6 +152,7 @@ use PhpParser\PrettyPrinter\Standard;
 use PhpParser\PrettyPrinterAbstract;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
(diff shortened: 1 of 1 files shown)


maks-rafalko:
I would really love to see some kind of autowiring 👍 As currently it's very verbose to construct classes

staabm:
> I would really love to see some kind of autowiring

we did something similar in PHPStan recently and utilized https://github.com/ondrejmirtes/composer-attribute-collector for it. maybe infection can also take inspiration.

the advantage of the above package is, that you don't need runtime reflection, but built a static map at `composer dump` time once ahead of time.

staabm:
> addign attributes above the class

yep, it requires attributes. good thing about attributes is, that you have the information at the class and not in a config file somewhere else.

PR #2118, opened 2025-06-05
Reduce more boilerplate with autowiring (stage 1)

sanmai:
This PR:

- [x] Reduces more boilerplate after #2115
- [x] Covered by tests

I intend to clean up the list thoroughly, with a unit test, in a follow-up.

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -1294,6 +1294,7 @@ final class Container
         throw new InvalidArgumentException(sprintf('Unknown service "%s"', $id));
     }
 
+    /** @return iterable<object> */
     private function resolveParameter(ReflectionParameter $parameter): iterable
     {
         $paramType = $parameter->getType();
--- a/src/Container.php
+++ b/src/Container.php
@@ -38,6 +38,7 @@ namespace Infection;
 use function array_filter;
 use function array_key_exists;
 use Closure;
+use function count;
 use Infection\AbstractTestFramework\TestFrameworkAdapter;
 use Infection\CI\MemoizedCiDetector;
 use Infection\CI\NullCiDetector;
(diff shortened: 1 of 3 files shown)


src/Container.php
```diff
@@ -298,6 +302,7 @@ public static function create(): self
                 $container->getMutantCodeFactory(),
             ),
             Differ::class => static fn (): Differ => new Differ(new BaseDiffer(new UnifiedDiffOutputBuilder(''))),
+            SyncEventDispatcher::class => static fn (): SyncEventDispatcher => new SyncEventDispatcher(),
```
sanmai, line 305: I had to return this because our classes depend on an interface, not the concrete implementation. So we have to explicitly specify the implementation in this list.

src/Container.php
```diff
@@ -1285,6 +1279,51 @@ private function get(string $id): object
             return $this->setValueOrThrow($id, $value);
         }
 
+        $resolvedArguments = take($constructor->getParameters())
+            ->map($this->resolveParameter(...))
+            ->toList();
```
maks-rafalko, line 1284: accidently removed my comment, so writing again:

do we need `take()` here just as a syntax sugar to have a`map()` in "functional pipe way"? Am I right that we don't use generators here?

sanmai: Look, arguments can have any type, including null. We need to create a list of arguments that can be incomplete. So we need a method that will either resolve a parameter or return none.

One way to do that is add [an option type](https://en.wikipedia.org/wiki/Option_type). Another class basically, still whole a lot of plumbing. Another way, which we use here, is to use a generator that returns nothing. 

```php
    private function optionGenerator(bool $nothing): iterable
    {
        if ($nothing) {
            return;
        }
        
        yield null;
    }
```

So we use the pipeline here to combine outputs of several “optional” generators into a single array.

```php
// $values is [null] here
$values = take([true, false])
    ->map($this->optionGenerator(...))
    ->toList();
```

src/Container.php
sanmai: We have dependencies configured in two primary ways: via a concrete instance, and via an abstract interface. The first one is easy-peasy. The second is more challenging: the auto-wiring can’t know exactly which concrete class it should use looking at the interface name. So it has to rely on the list to find the possible concrete type. And I think I need to be more thorough here. BRB.

src/Container.php
maks-rafalko: got it, thanks! probably worth adding a comment that we do it because one of the value can be `null`, so that in the future it will be easier to read/understand it

sanmai:
MSI went down with no mutations in the changed code :thinking:

maks-rafalko:
I think this is because MSI is flaky on our codebase (it's not stable). Let's lower it a bit to have a "room"?

src/Container.php
sanmai: Did that in fe52589e04c0d325800b9c9bcd5aba82495fce0b. Let me know if it still not clear enough.

sanmai:
I have a feeling that what I really need is to merge with the main branch. Can you enable the "update branch" feature to make it a push-button easy?

In https://github.com/infection/infection/settings look for "update branch" option.

maks-rafalko:
> Can you enable the "update branch" feature to make it a push-button easy?

Done

PR #2117, opened 2025-06-05
Don't produce mutations for equal type comparisons in IdenticalEqual

staabm:
This PR:

- [x] Adds new feature ...
- [x] Covered by tests
- [ ] Doc PR: https://github.com/infection/site/pull/XXX

Refs https://github.com/infection/infection/issues/2111

running `./build/infection.phar --mutators=IdenticalEqual --debug -s` on its own codebase 7a1244fa8e6fc37525da3bf0fe7fa1202820e08b with perMutator logger enabled:

before this PR
```
# Effects per Mutator

| Mutator        | Mutations | Killed | Escaped | Errors | Syntax Errors | Timed Out | Skipped | Ignored | MSI (%s) | Covered MSI (%s) |
| -------------- | --------- | ------ | ------- | ------ | ------------- | --------- | ------- | ------- | -------- | ---------------- |
| IdenticalEqual |       258 |      7 |     227 |      0 |             0 |         0 |       0 |       0 |     2.71 |             2.99 |

258 mutations were generated:
       7 mutants were killed
       0 mutants were configured to be ignored
      24 mutants were not covered by tests
     227 covered mutants were not detected
       0 errors were encountered
       0 syntax errors were encountered
       0 time outs were encountered
       0 mutants required more time than configured

Metrics:
         Mutation Score Indicator (MSI): 2%
         Mutation Code Coverage: 90%
         Covered Code MSI: 2%

Generated Reports:
         - /Users/m.staab/Documents/GitHub/infection/per-mutator.md

Please note that some mutants will inevitably be harmless (i.e. false positives).
[warning] Dashboard report has not been sent: The current process is not executed in a CI build

Time: 1m 10s. Memory: 0.09GB. Threads: 1

```

after this PR
```
# Effects per Mutator

| Mutator        | Mutations | Killed | Escaped | Errors | Syntax Errors | Timed Out | Skipped | Ignored | MSI (%s) | Covered MSI (%s) |
| -------------- | --------- | ------ | ------- | ------ | ------------- | --------- | ------- | ------- | -------- | ---------------- |
| IdenticalEqual |       247 |      7 |     219 |      0 |             0 |         0 |       0 |       0 |     2.83 |             3.10 |


247 mutations were generated:
       7 mutants were killed
       0 mutants were configured to be ignored
      21 mutants were not covered by tests
     219 covered mutants were not detected
       0 errors were encountered
       0 syntax errors were encountered
       0 time outs were encountered
       0 mutants required more time than configured

Metrics:
[...]

PR #2119, opened 2025-06-05
Don't produce mutations for identical type comparisons in EqualIdentical

staabm:


staabm:
drive-by question - as I was not able to figure it out by looking into the codebase:

does infection run the whole phpunit test-suite for every mutant, or is there some kind of "as soon as phpunit detects a single error we consider the mutant killed and stop running more tests for this mutant"?

maks-rafalko:
Oh, that's a very good question. We have a lot of optimizations! Once again I should have documented all of this, but we are where we are.

> does infection run the whole phpunit test-suite for every mutant,

No, Infection does it's best (read below) to stop as soon as possible and to run as few tests as possible. 

> or is there some kind of "as soon as phpunit detects a single error we consider the mutant killed and stop running more tests for this mutant"

exactly. So by default, when you just run `infection` with no options, for each Mutant we run PHPUnit process with `stopOnDefect`

To better understand what's going on with PHPUnit process, look into this code:

https://github.com/infection/infection/blob/38991a32f17608df97848403f23921431c9a3842/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php#L91-L98

So Infection creates a new `phpunit.xml` for each process and sets a set of attributes there, including `stopOnDefect`:

https://github.com/infection/infection/blob/38991a32f17608df97848403f23921431c9a3842/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php#L135

But not only that. As you can see, we also set `failOnRisky`, `failOnWarning`. 

Also, we have a couple of "advanced" optimizations to kill faster / to optimize the whole process. In particular:

- ("mutant process" stage) We also enable cache, so that the second run will be much faster than the first one. Regarding this, I highly recommend to read this small blog post about this cache feature: https://infection.github.io/2021/09/06/whats-new-in-0.25.0/#Speed-up-Infection-runs-by-remembering-which-test-killed-a-mutant

- ("mutant process" stage) For slow (e.g. functional) tests we have a special option to speed up the process - `--only-covering-test-cases`. There is a dedicated blog post about it: https://infection.github.io/2021/07/27/whats-new-in-0.24.0/#Major-performance-improvement-for-the-projects-with-slow-test-suites-using-only-covering-test-cases-option

- ("initial tests run" stage) We generate on "initil phpunit run" coverage only for those files that you mention in `--filter` (or `--git-diff-filter` or `--git-diff-lines`). There is a dedicated blog post about it: https://infection.github.io/2021/09/06/whats-new-in-0.25.0/#Generate-coverage-for-filtered-files-only-when-filter-or-git-diff-filter-are-used

- ("initial tests run" stage) also, we have a feature to map source classes to tests that is based on `#[CoversClass]` attribute. `--map-source-class-to-test` option can dramatically decrease the time needed for “Initial Tests Run” stage. Please read a detailed explanation here: https://infection.github.io/guide/command-line-options.html#map-source-class-to-test

----

That being said, MT analysis can be speed up a lot, if configured properly. But the main step where to start for any project - is to mutate only touched (changed, added) code in a PR instead of running MT for the whole source code. This way, MT will take seconds (or minutes) instead of tens of minutes (or hours).

Hope this helps!

maks-rafalko:
thank you @staabm

Issue #2111, opened 2025-06-04
Smarter `IdenticalEqual` mutator


staabm:
I am currently playing with the tool and look into how it behaves.

Using the `perMutator` logger I had a look which mutators are applied most when running infection on its own codebase.

One interessting Mutator which came to my eyes is `IdenticalEqual` for which I had a look into the mutations with `./build/infection.phar --mutators=IdenticalEqual --debug -s`. Here a few interessting examples:

```diff
179) /Users/staabm/workspace/infection/src/Str.php:82    [M] IdenticalEqual

@@ @@
         // Trim trailing empty lines
         for ($i = $linesCount - 1; $i >= 0; --$i) {
             $line = $lines[$i];
-            if (trim($line) === '') {
+            if (trim($line) == '') {
                 unset($lines[$i]);
                 continue;
             }
...
182) /Users/staabm/workspace/infection/src/TestFramework/CommandLineBuilder.php:128    [M] IdenticalEqual

@@ @@
     }
     private function isBatchFile(string $path): bool
     {
-        return substr($path, -self::BAT_EXTENSION_LENGTH) === '.bat';
+        return substr($path, -self::BAT_EXTENSION_LENGTH) == '.bat';
     }
 }
...
183) /Users/staabm/workspace/infection/src/TestFramework/Coverage/CoverageChecker.php:132    [M] IdenticalEqual

@@ @@
                 $errors[] = sprintf('- The JUnit file could not be found: %s', $exception->getMessage());
             }
         }
-        if (count($errors) === 0) {
+        if (count($errors) == 0) {
             return;
         }
         $message = sprintf(<<<TXT

...
202) /Users/staabm/workspace/infection/src/TestFramework/Coverage/XmlReport/XmlCoverageParser.php:156    [M] IdenticalEqual

@@ @@
     {
         $methodsCoverage = [];
         foreach ($methodsCoverageNodes as $methodsCoverageNode) {
-            if ((int) $methodsCoverageNode->getAttribute('coverage') === 0) {
+            if ((int) $methodsCoverageNode->getAttribute('coverage') == 0) {
                 continue;
             }
             $methodName = $methodsCoverageNode->getAttribute('name');
...

213) /Users/staabm/workspace/infection/src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php:138    [M] IdenticalEqual

@@ @@
     }
     public function isSyntaxError(string $output): bool
     {
[...]

PR #2121, opened 2025-06-05
Support narrowing a union type containing `false` with a non-falsy value

staabm:
Implemented super basic type narrowing for unions containing `false`, because there are super many builtin functions in php-src which return either a certain value or `false`.

When after type narrowing only a single type is remaining which is compatible with second operand, we skip creating mutations (similar to https://github.com/infection/infection/pull/2117)

### DIFF
--- a/tests/phpunit/Mutator/Boolean/EqualIdenticalTest.php
+++ b/tests/phpunit/Mutator/Boolean/EqualIdenticalTest.php
@@ -336,6 +336,19 @@ final class EqualIdenticalTest extends BaseMutatorTestCase
                 PHP,
         ];
 
+        yield 'It mutates equal operator into identical operator for comparison of different return-typed operands (union vs. named-type)' => [
+            <<<'PHP'
+                <?php
+
+                preg_match() == count();
+                PHP,
+            <<<'PHP'
+                <?php
+
+                preg_match() === count();
+                PHP,
+        ];
+
         yield 'It mutates equal operator of non-reflectable functions' => [
             <<<'PHP'
                 <?php


tests/phpunit/Mutator/Boolean/IdenticalEqualTest.php
```diff
@@ -250,5 +250,60 @@ public static function mutationsProvider(): iterable
...
+                PHP,
+        ];
+
+        yield 'It not mutates identical operator into equal operator for union type operations with non-falsy operand (inverse)' => [
+            <<<'PHP'
+                <?php
+
+                1 === preg_match();
+                PHP,
+        ];
```
staabm, line 311: these 2 are the cases which represent the actual value of the PR.

no more mutations like

```diff
     public function isSyntaxError(string $output): bool
     {
-        return preg_match('/ParseError: syntax error/i', $output) === 1;
+        return preg_match('/ParseError: syntax error/i', $output) == 1;
     }
```

src/Mutator/Util/AbstractIdenticalComparison.php
```diff
@@ -109,12 +121,58 @@ private function getReturnType(Expr\FuncCall $call): ?string
...
 
-        if (!$returnType instanceof ReflectionNamedType) {
-            return $this->reflectionCache[$name] = null;
+        return $this->reflectionCache[$name] = $reflection->getReturnType();
+    }
+
+    /**
+     * @param Node\Scalar|Expr\ConstFetch|Expr\FuncCall $expr
+     */
+    private function narrowReturnType(ReflectionType $returnType, Expr $expr): ?string
```
maks-rafalko, line 131: just for more clarity, what do you think about renaming `$returnType` to something like `$functionCallReturnType`?

Just to minimize the information needed to be kept in the head reading the code and to ease the understanding

maks-rafalko (approved):
Like the idea, totally makes sense.

maks-rafalko:
btw, related to performance discussion from other thread: you can try `bin/infection --git-diff-lines --threads=max -s` to see mutants related to this PR to improve MSI and fix the build

maks-rafalko:
Since build is failing with low MSI, It's interesting to get your feedback about:

- whether it was useful (from tests quality perspective) to "kill" those Mutants
- did you find it hard to do
- did you simplify the code as a result
- did you find any false-positives / equivalent mutants that are impossible to kill
- etc.

Please share your thoughts!

Anyway, if you find it impossible, [we can always ignore some mutants](https://infection.github.io/guide/how-to.html#Disable-in-particular-class-or-method-or-line).

I've just checked your branch though, the following mutant:

```
1) /opt/infection/src/Mutator/Util/AbstractIdenticalComparison.php:88    [M] ArrayItemRemoval

@@ @@
             return $narrowed === 'float';
         }
         if ($expr instanceof Expr\ConstFetch) {
-            return $narrowed === 'bool' && in_array($expr->name->toString(), ['true', 'false'], true);
+            return $narrowed === 'bool' && in_array($expr->name->toString(), ['false'], true);
         }
         if ($expr instanceof Expr\FuncCall) {
             $exprReturnType = $this->getReturnType($expr);
```

can be killed by the following _new_ tests:

```php
yield 'It not mutates identical operator into equal operator for same type operations (true)' => [
    <<<'PHP'
        <?php

        true === is_array();
        PHP,
];

yield 'It not mutates identical operator into equal operator for inverse same type operations (true)' => [
    <<<'PHP'
        <?php

        is_array() === true;
        PHP,
];
```

staabm:
first, let me thank you for your support and time you put into the feedback - much appreciated.

> Since build is failing with low MSI, It's interesting to get your feedback about:
> 
>     * whether it was useful (from tests quality perspective) to "kill" those Mutants
> 
>     * did you find it hard to do
> 
>     * did you simplify the code as a result
> 
>     * did you find any false-positives / equivalent mutants that are impossible to kill
> 
>     * etc.

these are great questions and I need to reflect on them a bit more while working on the PR :-).
I will come back to you after a bit more time invested.

----

what currently is a bit frustrating: running the mutation testing has a pretty slow feedback loop (looking at mutation results; think about a possible fix to kill the mutant; add a test; re-run the suite to verify the mutant was killed as assumed).

In the state like it is right now, where the mutation testing results in 15+ mutants I think I loose a lot of time by manually scanning the result list and verifying whether the single mutant I tried to kill was actually killed. it would be great to get some kind of "difference view" (which mutants were killed/are no longer killed in comparison to a certain state before / I think thats similar to the PHPStan baseline concept).

even using `time bin/infection --git-diff-lines --threads=max -s` the feedback cycle is already 25 seconds.

I can't imagine how super frustrating it must be in case you are working on a project in which the test-suite takes multiple minutes to run (which translates into super super slow mutation testing)

----

> I've just checked your branch though, the following mutant:

added the test, thank you very much

maks-rafalko:
> In the state like it is right now, where the mutation testing results in 15+ mutants I think I loose a lot of time by manually scanning the result list and verifying whether the mutant I tried to kill was actually killed. it would be great to get some kind of "difference view" (which mutants were killed/are no longer killed in comparison to a certain state before / I think thats similar to the PHPStan baseline concept).

this is the feedback I really wanted to hear, with issues, problems, frustrating, confusion and so on 🤣 

I will think about it or even translate to an issue so we all (including core team) can think about it. I do agree that if something has a big feedback loop, it can quickly start to be annoying.. So it's a very serious topic to think about. Thank you

staabm:
> something has a big feedback loop, it can quickly start to be annoying

having said that, I think it would be great to have a CI job which runs infection on a certain project (frozen to a certain commit) and assert..
- how many mutations infection generates on this project -> so we see over time how infection evolves
- how long infection run takes -> so we have kind of a performance baseline

running infection on infection itself has the problem that numbers generated are super volatile, as we change the implementation and the codebase itself at the same time

maks-rafalko:
please merge master, there is a slightly decreased MSI that should unblock you if needed

staabm:
Marking as ready for review, as I am not yet experienced enough to work thru the new mutants this PR reports.
Hopefully after spending more time with mutating testing I get a better mental model to make use of the mutations reported.

thank you

maks-rafalko:
thank you @staabm

PR #2128, opened 2025-06-06
Introduce collision detector

staabm:
see https://github.com/shipmonk-rnd/name-collision-detector

result: rename a single namespace as we had 2 files in the repo which contains same named classes in same named namespace

----

this check is necessary/usefull, as we recently added more runtime reflection dependencies and you easily get confusing tooling errors in case fully qualified class names are not unqiue in the codebase

the new tool runs in 1-2 seconds, so not a big deal perf wise.
I have added it to the "autoreview" makefile target, as looks to be the right step in the dev workflow


build/collision-detector.json
maks-rafalko, line 1: should it be located in `./devTools` instead? I don't even get how it's commited to `build` folder when it's git-ignored

staabm: good catch, fixed.

PR #2131, opened 2025-06-06
Share reflection-cache across Mutators

staabm:
make this cache more efficient by sharing its contents between `EqualIdentical`, and `IdenticalEqual`


maks-rafalko:
thank you @staabm

PR #2132, opened 2025-06-06
Don't produce mutations for same type comparisons of static method calls

staabm:

---

before this PR we only did this optimization for function calls


staabm:
I can do the same for `StaticPropertyFetch` but doing more is not worth the effort.
We would need a full scope-aware analysis to produce even more/better results.
At this point I would re-implement PHPStan Scope in infection, which does not make sense.
Instead we need todo the big task of integrating PHPStan with Infection and leverage PHPStan typesystem.

PR #2125, opened 2025-06-06
Validate autowiring and reduce the boilerplate (stage 2)

sanmai:
See: #2118 #2115

This PR:

- [ ] Adds new feature ...
- [ ] Covered by tests

PR #2133, opened 2025-06-06
Container: more explicit handling of abstract types

sanmai:
In #2125 I tried to go all-in on the auto-wiring, but hit a snag with a segmentation fault in PHP 8.2. Let's take a more balanced approach with laying the groundwork, but leaving the actual migration for later.

Another hurdle is that this class mentions Symfony's Filesystem. So the test has to be tagged as "integration" so every line added to Container will _reduce_ the MSI.

- [x] Covered by tests
- [x] Adds a memoizing `ProjectDirProvider` to clear a TODO


src/Container.php
```diff
@@ -1269,9 +1287,15 @@ private function get(string $id): object
      */
     private function resolveParameter(ReflectionParameter $parameter): iterable
     {
+        if ($parameter->isVariadic()) {
+            return;
+        }
```
sanmai, line 1292: A check against `MutationTestingResultsCollectorSubscriberFactory`.

devTools/phpstan.neon
```diff
@@ -23,6 +23,10 @@ parameters:
             path: '../src/Container.php'
             message: '#^Method Infection\\Container::(get|set).*\(\) should return T of .* but returns object\.$#'
             count: 2
+        -
+            path: '../tests/phpunit/Reflection/ContainerReflection.php'
+            message: '#Method .*Service\(\) should return .* but returns .*#'
+            count: 2
```
sanmai: I spend too much time trying to fix this one. All LLMs are certain this error should not be, yet it is.

tests/phpunit/Reflection/ContainerReflection.php
```diff
@@ -0,0 +1,153 @@
...
+
+    public function __construct(private readonly Container $container)
+    {
+        $this->reflection = new ReflectionClass($container);
+
+        $this->factories = $this->reflection->getProperty('factories');
+        $this->values = $this->reflection->getProperty('values');
+
+        $this->createServiceClosure = $this->reflection->getMethod('createService')->getClosure($container);
+        $this->getServiceClosure = $this->reflection->getMethod('get')->getClosure($container);
```
sanmai, line 69: I suppose we can make these methods public, and add more methods to the container itself.

src/Container.php
```diff
@@ -1238,27 +1238,44 @@ private function get(string $id): object
             return $this->setValueOrThrow($id, $value);
         }
 
+        $value = $this->createService($id);
+
+        if ($value === null) {
+            throw new InvalidArgumentException(sprintf('Unknown service "%s"', $id));
+        }
+
+        return $this->setValueOrThrow($id, $value);
```
sanmai, line 1247: Extracted `createService` so we would need to call `setValueOrThrow` only once.

maks-rafalko:
> Another hurdle is that this class mentions Symfony's Filesystem. So the test has to be tagged as "integration" so every line added to Container will _reduce_ the MSI.

my question is not related to this PR, but: do you know/remember why we skip integration tests from being running under Infection run? The are still quite fast or did I miss something? This would improve MSI.

sanmai:
I think there were several things:
- Tests, that involve writing files are often slower and disk-bound.
- Tests, that involve writing files are often not supposed to be run in parallel (write to fixed file name).

So they either slow, or broken (and broken - means mutations will be marked as "caught" even though it wasn't).

I guess we can mark this tests as non-integration explicitly.

.github/workflows/mt.yaml
```diff
@@ -25,7 +25,7 @@ concurrency:
 
 env:
   MIN_MSI: 77.0
-  MIN_COVERED_MSI: 86.48
+  MIN_COVERED_MSI: 86.02
```
sanmai:
```suggestion
  MIN_COVERED_MSI: 86.00
```

maks-rafalko (approved):
Did I get it right that we are going to remove explicit service declarations in the following PRs?

devTools/phpstan.neon
staabm: these errors can be fixed by putting

```
    /**
     * @template T of object
     *
     * @param callable(class-string<T>):?T $callable
     * @param class-string<T> $id
     * @phpstan-return ?T
     */
```

above `private static function handleCommonErrors(callable $callable, string $id): ?object`

tests/phpunit/Reflection/ContainerReflection.php
```diff
@@ -0,0 +1,153 @@
...
+
+            if ($typeReflection->isInterface()) {
+                continue;
+            }
+
+            yield $method->getName() => $returnTypeClassName;
+        }
+    }
+
+    private static function handleCommonErrors(callable $callable, string $id): ?object
```
staabm, line 151:
```suggestion
    /**
     * @template T of object
     *
     * @param callable(class-string<T>):?T $callable
     * @param class-string<T> $id
     * @phpstan-return ?T
     */
    private static function handleCommonErrors(callable $callable, string $id): ?object
```

devTools/phpstan.neon
sanmai: Thanks! I had this (and several variants of thereof, in one of the commits) but it did not help for some reason.

```
    /**
     * @template T of object
     *
     * @param callable(string): T $callable
     * @param class-string<T> $id
     * @phpstan-return ?T
     */
```
I guess there was something else that set this alarm off.

staabm: I can totally relate that doing this generics stuff sometimes is not that easy and PHPStan error-reports about it are sometimes not helpful.

PR #2134, opened 2025-06-06
Don't produce mutations for same type comparisons of class constants

staabm:


staabm:
ok - found one more case worth adding ;)

PR #2137, opened 2025-06-07
Repro redundat mutations

staabm:


staabm:
I think we could discuss at least these ones - which look like low hanging fruit

```diff
1) /Users/staabm/workspace/infection/src/Foo.php:40    [M] TrueValue

@@ @@
 {
     function doFoo($x)
     {
-        echo $x == true ? 1 : 0;
+        echo $x == false ? 1 : 0; // swaps IF/ELSE case, in the same way ternary-mutator does
     }
 }

3) /Users/staabm/workspace/infection/src/Foo.php:40    [M] Equal

@@ @@
 {
     function doFoo($x)
     {
-        echo $x == true ? 1 : 0;
+        echo $x != true ? 1 : 0; // swaps IF/ELSE case, in the same way ternary-mutator does
     }
 }

4) /Users/staabm/workspace/infection/src/Foo.php:40    [M] DecrementInteger

@@ @@
 {
     function doFoo($x)
     {
-        echo $x == true ? 1 : 0;
+        echo $x == true ? 0 : 0; // IF/ELSE case identical after mutation
     }
 }


7) /Users/staabm/workspace/infection/src/Foo.php:40    [M] IncrementInteger

@@ @@
 {
     function doFoo($x)
     {
-        echo $x == true ? 1 : 0;
+        echo $x == true ? 1 : 1; // IF/ELSE case identical after mutation
     }
 }



8) /Users/staabm/workspace/infection/src/Foo.php:40    [M] Ternary

@@ @@
 {
     function doFoo($x)
     {
-        echo $x == true ? 1 : 0;
+        echo $x == true ? 0 : 1;
     }
 }

```

maks-rafalko:
I would like @infection/core to participate here, as you are touching a very complex topics. 

My 2 cents, to have something to look at for myself in the future:

Questions that you raise here are often being discussed in white papers or scientific articles, so I'm sure there will be no simple answer.

Your post reminded me a great article, that, surprisingly, was shared by you back in 2021 here: https://github.com/infection/infection/discussions/1526 :) 

Article: :https://testing.googleblog.com/2021/04/mutation-testing.html 

Anyone who wants to participate in this discussion, I highly recommend reading it because it's a completely new world of Mutation Testing (that simply is absent here, at Infection).

Now, let's divide your questions by different categories

> whether I am right that we have too many mutations here

The number of mutations generated for each line/file, or even more - grouping mutants into High Order mutants or skipping mutants based on some "historical data" (as explained in the article above) - for me this sounds like a very complex solutions that we can only *try* and only then declare  an answer, is it good or is it bad. Looking at that article - yes, probably we generate many mutations for this particular examples. But I can't generalize this in my head without a deep diving into the topic (=spending month**s** with experiments with real teams dev teams).

> how we can be smarter in mutation generation

this is a much "simpler" question than the first one, and as you showed us (e.g. here https://github.com/infection/infection/pull/2138) - yes, we can :) 

So, by the above, I just wanted to highlight that here and right now we can solve questions like N2, but for N1 we would need months of experiments if we ever have a time for it. Just read that article to understand complexity of the question.

PS: My understanding of the questions above is very limited and I didn't dig into white papers a lot, so please don't treat my words seriously here, I think the real sitation is even worse than I imagine.

staabm:
> I would like @infection/core to participate here, as you are touching a very complex topics.

absolutely. I am just dumping ideas for discussions and I am also fine if we come to the conclusion that some of them are not good/good enough/whatever.

PR #2138, opened 2025-06-07
Don't mutate false/true in ternary

staabm:
I think we should not mutate true/false in ternary expression, since a separate TernaryMutator exists for this exact case.

prevents one redundant mutation per ternary which uses a `true` or `false` in its condition.
this was noticed in https://github.com/infection/infection/pull/2137

refs https://github.com/infection/infection/issues/2111

### DIFF
--- a/src/Mutator/Boolean/FalseValue.php
+++ b/src/Mutator/Boolean/FalseValue.php
@@ -76,12 +76,25 @@ final class FalseValue implements Mutator
 
     public function canMutate(Node $node): bool
     {
-        $parentNode = ParentConnector::findParent($node);
-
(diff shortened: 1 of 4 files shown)


src/Mutator/Boolean/FalseValue.php
```diff
@@ -76,12 +76,25 @@ public function mutate(Node $node): iterable
 
     public function canMutate(Node $node): bool
```
staabm, line 77: in this method I did the same change as in `TrueValue`. the diff is more convoluted because I adjusted the implementation to be more similar to `TrueValue` (refactoring)

PR #2141, opened 2025-06-07
Don't mutate greater/smaller-than in ternary

staabm:
similar to https://github.com/infection/infection/pull/2138

prevents one redundant mutation per ternary which uses a `<` or `<=`, `>`, `>=` in its condition.
this was noticed in https://github.com/infection/infection/pull/2137

refs https://github.com/infection/infection/issues/2111

### DIFF
--- a/tests/phpunit/Mutator/ConditionalNegotiation/GreaterThanNegotiationTest.php
+++ b/tests/phpunit/Mutator/ConditionalNegotiation/GreaterThanNegotiationTest.php
@@ -68,5 +68,13 @@ final class GreaterThanNegotiationTest extends BaseMutatorTestCase
                 PHP
             ,
         ];
+
+        yield 'It does not mutate inside ternary to prevent overlap with TernaryMutator' => [
+            <<<'PHP'
+                <?php
+                $x > 6 ? 'yes' : 'no';
+                PHP
+            ,
+        ];
     }
 }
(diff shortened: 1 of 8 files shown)


tests/phpunit/Mutator/ConditionalNegotiation/GreaterThanNegotiationTest.php
```diff
@@ -68,5 +68,13 @@ public static function mutationsProvider(): iterable
                 PHP
             ,
         ];
+
+        yield 'It does not mutate inside ternary to prevent overlap with TernaryMutator' => [
+            <<<'PHP'
+                <?php
+                $x > 6 ? 'yes' : 'no';
```
maks-rafalko, line 75: here I'm not sure it is the same as e.g. #2140

To not overlap with ternary, we would need to negate `$x > 6`, isn't it?

But negation of `$x > 6` is a `$x <= 6` which isn't happening here. What happens here is `$x >= 6`

So from my POV this doesn't overlap. If you don't agree, then let's find a more conrete example, probably I miss something

staabm: I had a similar thinking at first. but I think you (like me previously) mixed up this mutator with the ones in `src/Mutator/ConditionalBoundary`

> But negation of $x > 6 is a $x <= 6 which isn't happening here.

I think thats exactly whats happening here :) - see test case a few lines above

maks-rafalko: hah, you are right. Thanks for clarifying!

maks-rafalko (approved):

PR #2145, opened 2025-06-07
Don't mutate cast in return of typed function

staabm:
I think having a cast right in front of a return statement of a typed method/function is fine.
whether the cast itself is really necessary is better detected by PHPStan or other SA tools


staabm:
if we agree on this one, I can add support for other cast mutators

maks-rafalko:
> removing the casts turns the code into a fatal error (because strict-types)

am I right that it is always-killed mutator? (because of fatal error)

If so, totally makes sense

staabm:
> am I right that it is always-killed mutator? (because of fatal error)

Yes. 

it is a fatal error when the file beeing analyzed has `declare(strict_types=1);` - added a check against strict-types=1

staabm:
same is required for other Cast classes. same PR or each cast in a separate one?

maks-rafalko (approved):

maks-rafalko:
> same is required for other Cast classes. same PR or each cast in a separate one?

if it's not too time-consuming, better separate PRs. But if it's more convenient to do everything related together, let's do in one :) let me know please if I need to  merge this one

staabm:
I am fine with doing multiple small ones - please merge it as-is.

maks-rafalko:
thank you @staabm

PR #2151, opened 2025-06-08
Don't mutate array-cast in return of array-typed function

staabm:
analog https://github.com/infection/infection/pull/2145


maks-rafalko (approved):
👍

PR #2154, opened 2025-06-08
Don't object cast in arguments when strict_types=1

staabm:
when `strict_types=1` casting arguments is required to prevent a fatal runtime error like https://3v4l.org/DRZPm

therefore we prevent mutations when expression are used in arguments of function/method calls


src/Mutator/Cast/CastArray.php
```diff
@@ -65,24 +63,6 @@ public function canMutate(Node $node): bool
...
-
-                $returnType = $functionScope->getReturnType();
-
-                if ($returnType instanceof Node\Identifier && $returnType->name === 'array') {
-                    return false;
-                }
-            }
-        }
-
-        return true;
```
staabm, line 86: in all cast classes I factored out this common logic into the base-class to prevent duplication and ease maintenance

PR #2155, opened 2025-06-08
Reduce method visibility in Container

staabm:
each unnecessary public method creates a mutant on every run.


maks-rafalko (approved):
this is because they haven't been deleted yet during @sanmai's work on autoconfigure.

Makes sense, thank you.

PR #2162, opened 2025-06-09
Check minimum PHPStan version before starting MT logic (fail fast) if `--static-analysis-tool=phpstan` is used

maks-rafalko:
Also, considered https://getcomposer.org/doc/07-runtime.md#knowing-whether-package-x-is-installed-in-version-y

```php
\Composer\InstalledVersions::satisfies(new VersionParser, 'vendor/package', '2.0.*');
\Composer\InstalledVersions::satisfies(new VersionParser, 'psr/log-implementation', '^1.0');
```

but decided not to use it, because

- it requires one more package
- we had many issues when `Composer\InstalledVersions` is not present on user's installation. We need to add `if (!class_exists())` and then it's impossible to get the version. So implemented a very basic custom logic

### DIFF
--- a/src/StaticAnalysis/PHPStan/Adapter/PHPStanAdapter.php
+++ b/src/StaticAnalysis/PHPStan/Adapter/PHPStanAdapter.php
@@ -35,19 +35,27 @@ declare(strict_types=1);
 
 namespace Infection\StaticAnalysis\PHPStan\Adapter;
 
+use function explode;
 use Infection\Process\Factory\LazyMutantProcessFactory;
 use Infection\StaticAnalysis\PHPStan\Mutant\PHPStanMutantExecutionResultFactory;
 use Infection\StaticAnalysis\PHPStan\Process\PHPStanMutantProcessFactory;
 use Infection\StaticAnalysis\StaticAnalysisToolAdapter;
 use Infection\TestFramework\CommandLineBuilder;
 use Infection\TestFramework\VersionParser;
+use RuntimeException;
+use function sprintf;
+use function str_starts_with;
 use Symfony\Component\Process\Process;
+use function version_compare;
 
 /**
  * @internal
  */
 final class PHPStanAdapter implements StaticAnalysisToolAdapter
 {
+    private const VERSION_1 = 1;
+    private const VERSION_2 = 2;
+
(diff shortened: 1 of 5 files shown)


sanmai:
`Composer\InstalledVersions` is a feature of Composer V2, and only users of Composer V1 don't have it IIRC. Further, today [you can't literally install anything new using Composer V1](https://blog.packagist.com/shutting-down-packagist-org-support-for-composer-1-x/), so if a user installs a new version of Infection, they are definitely using Composer V2, and they have `InstalledVersions`.

maks-rafalko:
actually, I forgot to write another reason - `Composer\InstalledVersions` works with composer-based installations, while phpstan can be used as phar (not installed via composer). That's why this PR parses output of `phpstan --version` rather than checking which version is installed via composer.

src/StaticAnalysis/PHPStan/Adapter/PHPStanAdapter.php
```diff
@@ -93,6 +101,42 @@ public function getVersion(): string
...
+        // we assume all versions greater than 2.1.17 have needed functionality
+        if ($majorVersion > self::VERSION_2) {
+            return;
+        }
+
+        if (
+            $majorVersion === self::VERSION_2
+            && (
+                version_compare($version, '2.1.17', '>=')
+                || str_starts_with($version, '2.1.x-dev') // allow dev versions for development
```
staabm, line 118: not sure I like this line, as you might easily forget to add 2.2.x etc support as soon as PHPStan creates a new minor version.

maybe instead allow -dev versions in generall

maks-rafalko: > as you might easily forget to add 2.2.x etc support as soon as PHPStan creates a new minor version.

we don't need it, because 

- `2.2.x` is [handled by first condition](https://3v4l.org/dQYI8): `version_compare($version, '2.1.17', '>=')`
- this condition (`str_starts_with($version, '2.1.x-dev'`) is only needed to support `2.1.x` because `version_compare(`2.1.x`, '2.1.17', '>=') === false`

so we don't need to add anything when new versions are out

Issue #2156, opened 2025-06-08
Smarter `PublicVisibility` mutator


staabm:
as can be seen in https://github.com/infection/infection/issues/2111#issuecomment-2941502902 the `PublicVisibility` mutator is one most-mutation-producing mutators.

I think we could easily kill a lot of the mutations with a previous AST visitor, by remembering which method-calls we find when traversing.
that way we could already filter out a lot of `PublicVisibility` mutations, without even starting the killing process using PHPUnit.

e.g. when we traverse `SummaryFileLogger`

https://github.com/infection/infection/blob/e5d7621488b1d3da631ffb5d0c7142889d2f08b2/src/Logger/SummaryFileLogger.php#L53-L69

we can easily see by looking at the `PropertyFetch` based `MethodCalls` of typed properties, that a lot of `MetricsCalculator` methods are called from the outside (so they need to be `public`) -> we don't need to run phpunit to tell us.

maks-rafalko:
This sounds cool, to kill mutants even without executing 3rd-party tools (PHPUnit, PHPStan, etc.).

Just to fully understand the idea:

1. will it be the real Symfony's `Process` instance that under the hood runs `proc_open()`?
2. or should we abstract `Process` and add "fake" one to just run reflection-based kill process in the same thread? (since it's just PHP)
3. if this will be a real process: 
  3.1 why do you think it will be faster than PHPUnit?
  3.2 how will we pass accumulated data from the main process to reflection process?

staabm:
At best we do it in-process - when possible

staabm:
In the end this might be the same as adding phpstan type inference before invoking phpunit, ad already discussed elsewhere

maks-rafalko:
But I think you are talking about 2 different behaviors:

- the first is when we are smart enough and do not generate mutations (by using reflection (as you did in many PRs) or by using PHPStan scope)
- the second (which is described in this ticket) is to have a "reflection/type process killer" that kills mutants **after** they are generated

The difference between these 2 approaches is that in the 1st one we don't have a mutant at all, while in the second - we created mutant and kill it. Stats will be different.

staabm:
will implement a POC

PR #2163, opened 2025-06-10
POC: Implement AstPreFilter for PublicVisibility

staabm:
Implements the idea described in https://github.com/infection/infection/issues/2156

Based on Ast information and PHP Reflection knowledge, we try to cover mutations without running external process with the goal of beeing faster.

I have created a new ~~`AstKiller`~~`AstPreFilter` interface for that purpose.
~~AstKiller~~AstPreFilter `visit` methods are invoked per node at the pre-existing node-traversal process.
They are invoked again when Mutations have been generated and they judge via ~~`killMutation`~~`coversMutation` whether a mutation is ~~killed~~covered.

The current POC is not faster as before, but I am not yet sure why this is.
I will share the current implementation to have more eyes on the code and someone can see a obvious mistaken I am currently missing

### DIFF
--- a/src/Event/Subscriber/MutationTestingConsoleLoggerSubscriber.php
+++ b/src/Event/Subscriber/MutationTestingConsoleLoggerSubscriber.php
@@ -191,6 +191,7 @@ final class MutationTestingConsoleLoggerSubscriber implements EventSubscriber
         $this->output->writeln('<options=bold>' . $this->getPadded($this->metricsCalculator->getSyntaxErrorCount()) . '</options=bold> syntax errors were encountered');
         $this->output->writeln('<options=bold>' . $this->getPadded($this->metricsCalculator->getTimedOutCount()) . '</options=bold> time outs were encountered');
         $this->output->writeln('<options=bold>' . $this->getPadded($this->metricsCalculator->getSkippedCount()) . '</options=bold> mutants required more time than configured');
+        $this->output->writeln('<options=bold>' . $this->getPadded($this->metricsCalculator->getCoveredByAstPrefilter()) . '</options=bold> mutants were covered by AST prefilter');
 
         $mutationScoreIndicator = floor($this->metricsCalculator->getMutationScoreIndicator());
         $msiTag = $this->getPercentageTag($mutationScoreIndicator);
--- a/src/Logger/SummaryFileLogger.php
+++ b/src/Logger/SummaryFileLogger.php
@@ -61,6 +61,7 @@ final readonly class SummaryFileLogger implements LineMutationTestingResultsLogg
             'Syntax Errors: ' . $this->metricsCalculator->getSyntaxErrorCount(),
             'Escaped: ' . $this->metricsCalculator->getEscapedCount(),
             'Timed Out: ' . $this->metricsCalculator->getTimedOutCount(),
+            'Covered by AST prefilter: ' . $this->metricsCalculator->getCoveredByAstPrefilter(),
             'Skipped: ' . $this->metricsCalculator->getSkippedCount(),
             'Ignored: ' . $this->metricsCalculator->getIgnoredCount(),
             'Not Covered: ' . $this->metricsCalculator->getNotTestedCount(),
--- a/src/Metrics/MetricsCalculator.php
+++ b/src/Metrics/MetricsCalculator.php
@@ -117,6 +117,11 @@ class MetricsCalculator implements Collector
         return $this->countByStatus[DetectionStatus::KILLED_BY_STATIC_ANALYSIS];
     }
 
+    public function getCoveredByAstPrefilter(): int
+    {
+        return $this->countByStatus[DetectionStatus::COVERED_BY_AST_PREFILTER];
+    }
+
     public function getErrorCount(): int
     {
         return $this->countByStatus[DetectionStatus::ERROR];
--- a/src/Mutant/DetectionStatus.php
+++ b/src/Mutant/DetectionStatus.php
@@ -46,6 +46,7 @@ final class DetectionStatus
 
     public const KILLED_BY_TESTS = 'killed by tests';
     public const KILLED_BY_STATIC_ANALYSIS = 'killed by SA';
+    public const COVERED_BY_AST_PREFILTER = 'covered by AST prefilter';
     public const ESCAPED = 'escaped';
     public const ERROR = 'error';
     public const TIMED_OUT = 'timed out';
@@ -60,6 +61,7 @@ final class DetectionStatus
         self::ESCAPED,
         self::ERROR,
         self::TIMED_OUT,
+        self::COVERED_BY_AST_PREFILTER,
         self::SKIPPED,
         self::SYNTAX_ERROR,
         self::NOT_COVERED,
--- a/tests/phpunit/Logger/SummaryFileLoggerTest.php
+++ b/tests/phpunit/Logger/SummaryFileLoggerTest.php
@@ -70,6 +70,7 @@ final class SummaryFileLoggerTest extends TestCase
                 Syntax Errors: 0
                 Escaped: 0
                 Timed Out: 0
+                Covered by AST prefilter: 0
                 Skipped: 0
                 Ignored: 0
                 Not Covered: 0
@@ -88,6 +89,7 @@ final class SummaryFileLoggerTest extends TestCase
                 Syntax Errors: 2
                 Escaped: 2
                 Timed Out: 2
+                Covered by AST prefilter: 0
                 Skipped: 2
                 Ignored: 2
                 Not Covered: 2


maks-rafalko:
> I will share the current implementation to have more eyes on the code and someone can see a obvious mistaken I am currently missing

I don't see anything critical from the first look, so hard to say without debugging why it's not faster :| did you measure the number of generated mutants - is it less than it was before (on `master`)?

----

The only one note, just to think about, I still have a feeling that it's not a killer, but a pre-filter (or any other more suitable term).

As I described here (https://github.com/infection/infection/issues/2156#issuecomment-2954249472)

1. we **kill** Mutant when it's already generated and is counted in the statistics (affects number of generated mutants and killed mutants)
2. we filter-out mutations when we **prevent** Mutant to be generated (so it's physically doesn't exist on FS and thus not killed). In this case number of generated mutants and killed mutants are not affected - they aren't increased)

And I believe here we have the 2nd case.

staabm:
looking at a counter, I can see the PR is removing 131 PublicVisibility mutations.
how many time this saves depends on how slow the test-suite is and how fast/effectively phpunit is able to kill this very same mutants.

but I found some more cases which we currently miss.. so I think we can remove more mutations... stay tuned

staabm:
@maks-rafalko I think the PR works now as expected - its still not faster though.

it seems in the infection/infection repo the unit test-suite is optimized for infection using CoversClass etc. so that phpunit is really fast in killing mutants.

we are killing 149 PublicVisibility mutations at the current state

could you (or someone else) run this PR on a different application/repo so we can compare before/after numbers?
The PR currently contains a debug out message, so you can easily see how many mutations have been prevented:

```
5740 mutations were generated:
    4250 mutants were killed by Test Framework
       0 mutants were killed by Static Analysis
     158 mutants were configured to be ignored
     670 mutants were not covered by tests
     630 covered mutants were not detected
      10 errors were encountered
       1 syntax errors were encountered
      21 time outs were encountered

Metrics:
         Mutation Score Indicator (MSI): 76%
         Mutation Code Coverage: 88%
         Covered Code MSI: 87%

Generated Reports:
         - /Users/m.staab/Documents/GitHub/infection/infection.log
         - /Users/m.staab/Documents/GitHub/infection/per-mutator.md

[warning] Dashboard report has not been sent: The current process is not executed in a CI build
string("killed with AST: 149") <<--- debug out

Time: 4m 36s. Memory: 0.21GB. Threads: 6
```

staabm:
interessting that on my local machine infection reports

> "killed with AST: 149"

while in [CI I see](https://github.com/infection/infection/actions/runs/15559978698/job/43809604107?pr=2163)

> "killed with AST: 74"

maks-rafalko:
will do later this evening. I [have a PR](https://github.com/Roave/BetterReflection/pull/1510) on quite a big project - BetterReflection - so we can play there (5000+ mutations)

maks-rafalko:
> interessting that on my local machine infection reports
> 
> > "killed with AST: 149"
> 
> while in [CI I see](https://github.com/infection/infection/actions/runs/15559978698/job/43809604107?pr=2163)
> 
> > "killed with AST: 74"

is it the same PHP version as you use locally?

```
 ==> Setup PHP
✓ PHP Installed PHP 8.2.28

==> Setup Tools
✓ composer Added composer 2.8.9

==> Setup Coverage
✓ pcov.enabled=1 Added to php.ini
✓ pcov PCOV 1.0.12 enabled as coverage driver
```

maks-rafalko:
> could you (or someone else) run this PR on a different application/repo so we can compare before/after numbers?

For BetterReflection

With `dev-master`:

```
4409 mutations were generated:
    3660 mutants were killed by Test Framework
     212 mutants were killed by Static Analysis
      11 mutants were not covered by tests
      47 covered mutants were not detected
       7 errors were encountered
       9 time outs were encountered
     463 mutants required more time than configured

Metrics:
         Mutation Score Indicator (MSI): 98%
         Mutation Code Coverage: 99%
         Covered Code MSI: 98%

Note: to see escaped mutants run Infection with "--show-mutations" or configure file loggers.

[warning] Dashboard report has not been sent: The current process is not executed in a CI build

Time: 15m 25s. Memory: 0.40GB. Threads: 4

 [ERROR] The minimum required MSI percentage should be 99%, but actual is 98.53%. Improve your tests!
```

-------------

With this PR:

```
4228 mutations were generated:
    3479 mutants were killed by Test Framework
     211 mutants were killed by Static Analysis
      11 mutants were not covered by tests
      47 covered mutants were not detected
       7 errors were encountered
      10 time outs were encountered
     463 mutants required more time than configured

Metrics:
         Mutation Score Indicator (MSI): 98%
         Mutation Code Coverage: 99%
         Covered Code MSI: 98%

Note: to see escaped mutants run Infection with "--show-mutations" or configure file loggers.

[warning] Dashboard report has not been sent: The current process is not executed in a CI build
string(20) "killed with AST: 181"

Time: 15m 23s. Memory: 0.39GB. Threads: 4

 [ERROR] The minimum required MSI percentage should be 99%, but actual is 98.46%. Improve your tests!
```

---

string(20) "killed with AST: 181"

staabm:
> is it the same PHP version as you use locally?

yes tried on 8.2.x and 8.4.x and both report the same thing locally

> For BetterReflection

thanks for testing. its weird that there is no difference in runtime

maks-rafalko:
I think I have something that affects the execution a lot on my machine (even though I tried not to touch my PC). The second try:

`master`

```
Time: 16m 4s. Memory: 0.40GB. Threads: 4

 [ERROR] The minimum required MSI percentage should be 99%, but actual is 98.53%. Improve your tests!
```
----

this PR

```
string(20) "killed with AST: 181"

Time: 14m 34s. Memory: 0.39GB. Threads: 4

 [ERROR] The minimum required MSI percentage should be 99%, but actual is 98.47%. Improve your tests!
```

probably we need a smaller project to test with, so it doesn't take 15m when any other process may affect the execution. But I have no options right now, will need to find some.

src/AstKiller/AstKiller.php
```diff
@@ -0,0 +1,52 @@
...
+interface AstKiller
+{
+    /**
+     * @return class-string<Mutator<Node>>
+     */
+    public function getMutatorClass(): string;
+
+    public function visit(Node $node): void;
+
+    public function killsMutation(Mutation $mutation): bool;
```
theofidry: note that the trend is to move towards "covering mutations" rather than "killing" it, at least from mutant & a bunch of other mutation tools (scientific papers would be a different story). So maybe we could go with `coversMutation()`? But then again maybe it makes more sense to rename this at the entire project level for consistency if it's the nomenclature we want to follow...

src/AstKiller/PublicVisibility.php
```diff
@@ -0,0 +1,183 @@
...
+use ReflectionMethod;
+use ReflectionNamedType;
+use ReflectionProperty;
+
+class PublicVisibility implements AstKiller
+{
+    /**
+     * @var array<string, array<string, bool>>
+     */
+    private array $seenMethods = [];
```
theofidry:
```suggestion
    private array $methodsSeen/Visited = [];
```
?

src/AstKiller/PublicVisibility.php
```diff
@@ -0,0 +1,183 @@
...
+            $node instanceof Node\Expr\MethodCall
+            || $node instanceof Node\Expr\NullsafeMethodCall
+            || $node instanceof Node\Expr\PropertyFetch
+            || $node instanceof Node\Expr\NullsafePropertyFetch
+        ) {
+            $astChain[] = $node;
+            $node = $node->var;
+        }
+
+        if ($class === null || $astChain === [] || !$node instanceof Node\Expr\Variable) {
```
theofidry: class is not re-assigned past L69 so this early return could already be done for this condition there

src/AstKiller/PublicVisibility.php
```diff
@@ -0,0 +1,183 @@
...
+
+    public function visit(Node $node): void
+    {
+        if (!$node instanceof Node\Expr\MethodCall) {
+            return;
+        }
+        /** @var ClassReflection $class */
+        $class = $node->getAttribute(ReflectionVisitor::REFLECTION_CLASS_KEY);
+
+        $astChain = [];
```
theofidry: why the term "chain"? (as opposed to `$nodes` as it is a list of nodes too)

src/AstKiller/AstKillerManager.php
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Infection\AstKiller;
+
+use Infection\Mutation\Mutation;
+use PhpParser\Node;
+
+class AstKillerManager
```
theofidry:
```suggestion
class AstKillerRegistry
```
?

"manager" is one of those terms that I feel can just be attributed to anything and creeps up too easily

src/AstKiller/AstKillerManager.php
```diff
@@ -0,0 +1,74 @@
...
+    public function visitNode(Node $node): void
+    {
+        foreach ($this->astKillers as $astKiller) {
+            $astKiller->visit($node);
+        }
+    }
+
+    public function killsMutation(Mutation $mutation): bool
+    {
+        foreach ($this->astKillers as $astKiller) {
```
theofidry: I don't recall if we are using `iter` but if we are this could be a filter + any instead to have something a bit more imperative

src/Process/Runner/MutationTestingRunner.php
```diff
@@ -121,6 +125,15 @@ public function run(iterable $mutations, string $testFrameworkExtraOptions): voi
 
                 return false;
             })
+            ->filter(function (Mutant $mutant) use (&$counter): bool {
+                if ($this->astKillerManager->killsMutation($mutant->getMutation())) {
```
theofidry: noting this here as this is the core of this MR, at least what highlights its usage.

My question is what is really the intent here? My understanding is "reducing dumb mutations".

The reason I am asking is because there is a fundamental difference depending of where this is addressed. It has an impact on two things:

- how we design mutators
- how we report this

It is not rare in mutation testing to have a "mutation selector", that can do various optimizations such as detecting redundant mutations to remove them, prioritize the mutants by their chances of having a higher impact or likeliness of the speed at which it will be detected etc. In other words a process of some kind that reduces the initial number of mutations that goes through.

In the current implementation, this is where your "AST killer" would fit.

What I am questioning here is where do we want to draw the line to fit that in this "mutation selector" process, and what should rather be handled at the mutator level. Because mutators could also be made smarter to not generate dumb mutations in the first place.

<hr/>

My opinion on this is that ideally we have both. I think it makes sense to have some visitors to collect and enrich the visited AST for the mutators to be able to leverage that information, be smarter and avoid yielding nonsensical mutations.

I think having the "mutation selector" process will be welcomed, but maybe it just don't have enough to work with for that yet.

theofidry: /cc @maks-rafalko

theofidry (commented):
I didn't really finish the review as it's still a draft and there is some design questions from my side

src/AstKiller/AstKillerManager.php
staabm: I was totally not happy with the name :) - thanks for suggestions

src/AstKiller/PublicVisibility.php
staabm: I call it chain, because it relates to a call like `$this->doFoo()->doBar()->doFooBar()`

theofidry: `$methodCallChain[]` maybe then? I guess it doesn't matter to much tbh, I don't have a strong opinion about it by any means

src/Process/Runner/MutationTestingRunner.php
staabm: > What I am questioning here is where do we want to draw the line to fit that in this "mutation selector" process, and what should rather be handled at the mutator level. Because mutators could also be made smarter to not generate dumb mutations in the first place.

the filter - as implemented in this PR - make use of knowledge it collects on the first whole-project ast traversal. so it has a more holisitic view on all the ast nodes in the codebase beeing tested. mutators in contrast only have a pretty locale view on the current ast-node and its sorroundings.
so with AstPreFilters we can take decision which current mutators can't because they have less context.

A "mutation selector" can still live after the AstPreFilter and order mutations etc. I think thats a separate problem/phase.

src/AstKiller/AstKillerManager.php
staabm: we don't have `iter` in the project atm

staabm:
Addressed the feedback and added rudimentary result logging.

I did not yet look into all different kind of loggers, as I guess we need to discuss when/where we want the `Covered by AST prefilter` metric to show up/be included.

feedback welcome.

---

whether this PR as a whole makes sense depends on whether we find more use-cases which we can implement with it, I guess. might be worth doing some [brainstorming/bringing a few more examples](https://github.com/infection/infection/issues/2286) to the table.

src/Process/Runner/MutationTestingRunner.php
theofidry: Reviewing this with a fresh head, I'm happy with it.

I think there is a lot of improvements that could done to be able to have a higher level and clearer view, as well as a clearer cut between responsibilities, but it goes way out of scope of this MR hence I do not think it's worth holding things back because of that.

PR #2165, opened 2025-06-10
Update PHPStan to 2.1.17

staabm:
lets use a version which is compatible with `--static-analysis-tool=phpstan`

PR #2169, opened 2025-06-10
Generate PHPStan custom mutant config file to set parallel processes to 1

maks-rafalko:
For each Mutant process, we need to set `parallel.maximumNumberOfProcesses: 1` so that PHPStan doesn't create N subprocesses when it itself is executed from a subprocess, because Mutant Processes are run in parallel by Infection.

Also, this is needed to try ideas from https://github.com/infection/infection/pull/2136#issuecomment-2952440312


src/Container.php
```diff
@@ -297,6 +299,11 @@ public static function create(): self
             TestFrameworkConfigLocator::class => static fn (self $container): TestFrameworkConfigLocator => new TestFrameworkConfigLocator(
                 (string) $container->getConfiguration()->getPhpUnit()->getConfigDir(),
             ),
+            StaticAnalysisConfigLocator::class => static function (self $container): StaticAnalysisConfigLocator {
+                return new StaticAnalysisConfigLocator(
+                    '/opt/infection/tests/e2e/PHPStan_Integration',// (string) $container->getConfiguration()->getPhpStan()->getConfigDir(),
```
maks-rafalko: - [x] fix it

infection.json5
```diff
@@ -10,14 +10,17 @@
...
         ]
     },
     "logs": {
         "stryker": {
             "report": "master"
         }
     },
+    phpStan: {
+        configDir: "devTools"
+    },
```
maks-rafalko, line 23: @staabm this is the fix for https://github.com/infection/infection/pull/2165#issuecomment-2960429463

src/Configuration/ConfigurationFactory.php
```diff
@@ -254,6 +255,23 @@ private function retrievePhpUnit(SchemaConfiguration $schema, string $configDir)
...
+    private function retrievePhpStan(SchemaConfiguration $schema, string $configDir): PhpStan
+    {
+        $phpStan = clone $schema->getPhpStan();
+
+        $phpStanConfigDir = $phpStan->getConfigDir();
+
+        if ($phpStanConfigDir === null) {
+            $phpStan->setConfigDir($configDir);
+        } elseif (!Path::isAbsolute($phpStanConfigDir)) {
+            $phpStan->setConfigDir(sprintf(
```
theofidry: maybe instead of using a setter here a factory would be sufficient?

```
$phpstan = $phpstan->withConfigDir(...);
```

This way you don't need to clone it either and if the config dir does not change we do not create a new object reference

src/StaticAnalysis/PHPStan/Adapter/PHPStanAdapterFactory.php
```diff
@@ -40,22 +40,28 @@
...
 final class PHPStanAdapterFactory implements StaticAnalysisToolAdapterFactory
 {
     public static function create(
+        string $staticAnalysisConfigPath,
         string $staticAnalysisToolExecutable,
         float $timeout,
+        string $tmpDir,
     ): StaticAnalysisToolAdapter {
         return new PHPStanAdapter(
+            new Filesystem(),
```
theofidry, line 57: don't we have a filesystem to inject from the container?

src/StaticAnalysis/PHPStan/Process/PHPStanMutantProcessFactory.php
```diff
@@ -78,18 +86,42 @@ public function create(Mutant $mutant): MutantProcess
...
+
+    private function buildMutationConfigFile(string $mutationHash): string
+    {
+        $mutantConfigPath = sprintf(
+            '%s/phpstan.%s.infection.neon',
+            $this->tmpDir,
+            $mutationHash,
+        );
+
+        $this->fileSystem->dumpFile(
```
theofidry, line 114: Not a blocker for here by any means as this is a perf improvement suggestion, whereas your current PR is a fix. It would be great if this could be passed via the CLI instead as it otherwise means we need to dump one more file for every mutation, making it yet more IO intensive

infection.json5
staabm: confirmed, works now for me - thanks

src/StaticAnalysis/PHPStan/Process/PHPStanMutantProcessFactory.php
theofidry: I realise it is not something we'll be able to do every time neither, but it would be idea if we could have one main config dumped (if necessary) for the whole mutation run and then having the mutator specific config configured via CLI parameters

theofidry (approved):

src/StaticAnalysis/PHPStan/Process/PHPStanMutantProcessFactory.php
staabm: its worth trying. but you need to make sure PHPStan can still utlize the result cache.
depending on which parameters you pass or which settings you change in the config, PHPStan might decide it need to re-build its result cache

staabm: btw: there is a option you can pass to make PHPStan fail it it was not able to use the result cache.
https://github.com/phpstan/phpstan-src/blob/6095601023ab1fe4bc9371c09679aaf979e02152/src/Command/AnalyseCommand.php#L111

this might be useful that we don't accidentally do a change and make the result cache useless

maks-rafalko: > It would be great if this could be passed via the CLI instead as it otherwise means we need to dump one more file for every mutation, making it yet more IO intensive

I was told by Ondrej that the only one possible way to override parallelization is through a config file. PHPStan doesn't have CLI option for it.

theofidry: I suspected this was a current limitation rather than an oversight from your part, I just wanted to clarify the implication and that maybe it is an improvement we should request/pursue; Not a blocker by any means.

src/Configuration/ConfigurationFactory.php
maks-rafalko: well, I just copied this from the PHPUnit implementation as is.

But renamed, thanks

src/StaticAnalysis/PHPStan/Adapter/PHPStanAdapterFactory.php
maks-rafalko: I left it as is (as other couple of classes created here via `new ...` because this can be moved later to a separate package as we did with test frameworks, and there we don't have a container.

PR #2172, opened 2025-06-10
Set ignore config when `--mutants` is used and config has `global-ignoreSourceCodeByRegex`

maks-rafalko:
Fixes a bug with with `--mutators=MethodCallRemoval` usage, `global-ignoreSourceCodeByRegex` was lost resulting in not needed mutants to be created.

### DIFF
--- a/src/Configuration/ConfigurationFactory.php
+++ b/src/Configuration/ConfigurationFactory.php
@@ -217,6 +217,10 @@ class ConfigurationFactory
             $mutatorsList = $schemaMutators;
         } else {
             $mutatorsList = array_fill_keys($parsedMutatorsInput, true);
+
+            if (array_key_exists('global-ignoreSourceCodeByRegex', $schemaMutators)) {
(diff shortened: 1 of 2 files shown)


theofidry (approved):
I guess this is an annoying one to test... Maybe via an e2e test?

maks-rafalko:
> I guess this is an annoying one to test... Maybe via an e2e test?

initially wrote and then removed :) The problems with e2e tests:

- they are getting slower and slower
- if report is changed, we have a lot of manual work to change `expected-output.txt` to a new one

So I try to write e2e tests when it's really needed and it's hard to test smth with unit tests. Seems like it's not the case here, as it was quite easy to cover this bug by unit tests. So if you don't mind, I will skip e2e here

theofidry:
Yes... It's a case where you need more infra to be able to write more expressive e2e tests... Not sure if we're there yet though and they are definitely quite slow

Issue #2174, opened 2025-06-11
perMutator log suggestions


staabm:
while sleeping over yesterdays work on the mutator log, I had the following ideas of things which might be useful to have in said logfile

- [x] the column header of "timed out" column could indicate what the limit is (as I thinkt thats something which can be configured somewhere and might be different per project?)
- [ ] timing wise it might be useful to have a overall sum per mutator per "by Test / by SA / all over all": so how much time in sum spent infection on a single mutator across the whole run (160x 2 seconds is worse then 3x 15 seconds)

- [ ] looking at https://github.com/infection/infection/pull/2164#issuecomment-2960663519 we see `ArrayItemRemoval`  took a lot of time in SA and tests for the 162 mutants but still 34 survived. a survived mutant after we tried PHPUnit and PHPStan is a repetitive very expensive run, as it likely needs the ~~whole~~ test-suite to run and also PHPStan to run every time
  - [ ] same for `ConcatOperandRemoval` which we have a lot of mutants and a lot of escaped ones
  - [ ] `MethodCallRemoval` seems also expensive

-> at least for the infection codebase itself it seems these are the ones we should optimize to get a faster overall feedback loop

maks-rafalko:
sounds great 👍

staabm:
btw: what does `%s` mean here?

<img width="357" alt="Image" src="https://github.com/user-attachments/assets/e0adf23c-1718-4006-a5b2-c54c66707451" />

is this a leftover `%s` of `sprintf` or should this mean something like "in percent"?

theofidry:
I think it makes sense to have a log tailored towards debugging/analysing rather than, what is traditionally for users, "what is covered".

To push the idea further, I think what would be nice to have is:
- a detailed breakdown for each mutation issued
- summaries per category of mutation (e.g. all `MethodCallRemoval`)

This would make it easier to review for various purposes: fish for overly expensive mutations, compare the mutations generated per projects etc.

staabm:
@theofidry that make sense. I think we can still put the timeout into the column header of the perMutator log but doing all the other in-detail stuff in a new "perMutation" logger sounds like a good idea.

DebugLog might be a good name, to not mix up the term "mutation" and "mutator" too much

maks-rafalko:
> DebugLog might be a good name, to not mix up the term "mutation" and "mutator" too much

Unfortunately we already have Debug log and it doesn't mean what you think :) it's used for tests only. But yeah, we can rename that one for sure

[...]

PR #2175, opened 2025-06-11
Render process-limit into perMutation logger table header

staabm:
refs https://github.com/infection/infection/issues/2174#issue-3135797022

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -446,6 +446,7 @@ final class Container
                     $container->getLogger(),
                     $container->getStrykerHtmlReportBuilder(),
                     $config->getLoggerProjectRootDirectory(),
+                    $config->getProcessTimeout(),
                 );
             },
             MutationTestingResultsLogger::class => static fn (self $container): MutationTestingResultsLogger => new FederatedLogger(...array_filter([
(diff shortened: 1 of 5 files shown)


src/Container.php
```diff
@@ -446,6 +446,7 @@ public static function create(): self
                     $container->getLogger(),
                     $container->getStrykerHtmlReportBuilder(),
                     $config->getLoggerProjectRootDirectory(),
+                    $config->getProcessTimeout(),
```
maks-rafalko, line 449: just to double check we are on the same page - this is a static value for all the mutators / processes, it's not different or dynamic

maks-rafalko (approved):

src/Container.php
staabm: yes, I did some research and got to the conclusion that this very same process limit is used for MutatorProcess no matter whether its PHPUnit based or Static Analysis based

maks-rafalko: so do we still want this to be merged? (didn't get it from your reply :))

I'm ok with that, just want to be sure that it's not a new information for you after creating PR

staabm: I think it makes sense to have the limit at hand.
it gives a bit more pressure on how many time a timed out mutant costs you for your feedback loop.

these are very expensive.

maks-rafalko:
thank you @staabm

Issue #2146, opened 2025-06-08
`InstanceOf` mutations


staabm:
code like

```php
use PhpParser\Node;

class Foo {
    private function doFoo($x) {
        if ($x instanceof Node\Stmt\Declare_) {
            foreach($x->declares as $declare) { }
        }
    }
}
```

leads to the following 2 mutations:
```diff
1) /Users/staabm/workspace/infection/src/Foo.php:42    [M] InstanceOf_

@@ @@
 {
     function doFoo($x)
     {
-        if ($x instanceof Node\Stmt\Declare_) {
+        if (false) {
             foreach ($x->declares as $declare) {
             }
         }
     }
 }

2) /Users/staabm/workspace/infection/src/Foo.php:42    [M] InstanceOf_

@@ @@
 {
     function doFoo($x)
     {
-        if ($x instanceof Node\Stmt\Declare_) {
+        if (true) {
             foreach ($x->declares as $declare) {
             }
         }
     }
 }
```

I wonder why we emit 2 mutations in this case...? 

why do we emit a mutation which reflects what the code before mutation testing looked like (the `true`)?
is the goal to check whether the "instance of" is redundant? should we only emit mutations with the goal to find bugs, instead of trying to find dead code?

the `if(true)` case is easily caught by IDE or static analysis tooling, as you usually only use `instance of` to narrow a type for the if-body, so you can work on properties/methods which only exist on said type.

---

in contrast, for code like

```php
class Foo {
    function doFoo($x) {
        if($x == true) {}
    }
}
```

we only emit `$x == false`

staabm:
//cc @Ocramius do you have an opinion on this one?

staabm:
I just found a example in the test-suite where it makes sense to emit `true` and `false`:

```php
        yield 'It mutates an instanceof comparison inside other than `assert()` functions' => [
            <<<'PHP'
                <?php

                return someFunc($example instanceof $foo);
                PHP,
            [
                <<<'PHP'
                    <?php

                    return someFunc(false);
                    PHP,
            ],
        ];
```

from my experience the use of `instance of` inside a IF condition is way more common though.
maybe we can/could emit only 2 mutations when not inside a condition? in the infection codebase alone this would remove several hundred mutations.

Ocramius:
> I wonder why we emit 2 mutations in this case...?

Each boolean expression can be turned into "always true" or "always false" to see whether it's a dead branch, or an always-true branch.

[...]

PR #2176, opened 2025-06-11
Don't mutate instanceof into pre-existing case

staabm:
closes https://github.com/infection/infection/issues/2146


staabm:
lot less mutations.. still same runtime.. I am kind of suprised about these infection characteristics.

I don't expect a fast merge of this PR, as the [discussion was a bit controversal](https://github.com/infection/infection/issues/2146). Please don't feel pressured into merging it.

maks-rafalko:
> I am kind of suprised about these infection characteristics.

yeah, that's annoying, eventually want to debug it. First thoughts:

- Infection execution is affected by other processes being run on your OS. So it's better to either not touch your computer in 2 different runs or compare on GH actions
- we need to compare how the very same mutant (for line X column Y node Z) is killed between 2 different runs. Do we run the same tests? Is "killing" test is the same? Was it the first one or not? Because, do you remember, I were explaining that we use phpunit's cache and we try to run "failed" tests first? We also have sorting and try to run the fastest tests first if there is no cache, so all of this may affect the result time

if you would want to play with it, here is how I do it:

```
infection --debug --log-verbosity=all --threads=X
```

with enabled `text` logger inside `infection.json5`.

`--debug` adds output of PHPUnit & PHPStan to the log file. `--log-verbosity=all` adds killed/errored/timouted mutants to the log file.

Then we can just compare (see a diff) between files if `--threads=1` when they are not in a random order, or compare manually when `--threads=N>1`.

Here is what I mean by cache and comparing diffs https://infection.github.io/2021/09/06/whats-new-in-0.25.0/#Speed-up-Infection-runs-by-remembering-which-test-killed-a-mutant

Please look at this image:

image

it clearly shows how "the second" run can be faster thanks to phpunit's cache for each Mutant

sanmai (changes requested):
This is a different mutation: negation, as opposed to constant replacement. It should be treated as such.

src/Mutator/Boolean/InstanceOf_.php
```diff
@@ -54,33 +56,53 @@ final class InstanceOf_ implements Mutator
...
+        if ($node instanceof Node\Expr\BooleanNot) {
+            yield $node->expr;
 
-        yield new Node\Expr\ConstFetch(new Node\Name('false'));
+            return;
+        }
+
+        if (!$node instanceof Node\Expr\Instanceof_) {
+            throw new LogicException(sprintf('Node must be InstanceOf_ but "%s" given.', $node::class));
+        }
```
sanmai:
```suggestion
        Assert::isInstanceOf($node, Node\Expr\Instanceof_::class);
```

sanmai (approved):
I guess I'm more upset that we could not unify all constant-replacing mutators later, but if y'all so keen to optimize this mutation, I guess it makes sense.

staabm:
> This is a different mutation: negation, as opposed to constant replacement. It should be treated as such.


What do I need to change for this comment? Not sure what you are requesting

staabm:
> I guess I'm more upset that we could not unify all constant-replacing mutators later

Could you clarify what this means?

My point of this PR is, that replacing the initial code with true and false always (as it was before) creates 1 redundant mutation which only reflects what the initial code already represented.

This also means with this PR we nearly cut the number of instanceof mutations in half:

```
| InstanceOf_    (this PR)        |       366 | 
| InstanceOf_    (before)         |       720 | 
```

staabm:
In addition it seems a change required for the SA tooling phase, see https://github.com/infection/infection/pull/2136

sanmai:
I understood this, and I can't find there the main discussion for the SA tooling integration. So far it sounds like it only makes Infection slower, not faster. I'm not sure the slow down is worth it: what are the chances it can get faster?

Don't get me wrong: we tried it, and it did not blow our minds. So what is the reason this idea is still on the table?

maks-rafalko:
(everything below is not related to PR, but an answer to this quote)

> So far it sounds like it only makes Infection slower, not faster.

I don't get it, to be honest. Running SA for escaped mutants is _completely opt-in_ (`--static-analysis-tool=phpstan`). 

It doesn't make Infection slower. It's the opposite, during all of this work - we started the whole epic removing useless mutations like this https://github.com/infection/infection/issues/2184 - so as a result Infection will be faster and smarter in its original usage (without SA).

There are developers who don't use Infection without SA, that's why https://github.com/Roave/infection-static-analysis-plugin has been created years ago and [has millions of installs](https://packagist.org/packages/roave/infection-static-analysis-plugin/stats).

What we do here with SA integration (again - optional) is just make it native, with proper parallelization for those who need it.

sanmai:
What if, just what if, the users of [infection-static-analysis-plugin](https://github.com/Roave/infection-static-analysis-plugin) install it not because they need it, but because they do not know better? What if the reason people use it is because the language Infection uses sounds offensive? What if we are solving the wrong problem?

Because if people are installing a plugin that fundamentally breaks the value proposition ([in my own words](https://discord.com/channels/767914934016802818/771020567343136831)) of mutation testing just because they don't like being told their tests have gaps... that says the real issue is how Infection communicates what it's doing, not what it's actually doing.

The removal of useless mutations is great since it does not preclude making Infection intentionally slower by a factor of 10-100. I would seriously dislike the situation when I recommend Infection to someone else, and they install it, toggle this _optional_ switch to enable SA (`--static-analysis-tool`), and later come back to me claiming it is unusable and slow.

### The Fix

What if we normalize the 80% MSI to be 100%? 

So anyone with 80% MSI will feel better about themselves looking at a nice 100% number in the report. It will have the same effect as running every mutation through SA, with no slowbacks.

Like, hey folks, we learned that industry average MSI is 80%, so we declare it as the norm and the goal. But if you are really an extreme programming enthusiast, there's a handy `--extreme` switch that makes it all back to very hard land.

The beauty is that anyone using the `--extreme` flag would be the people who actually understand what mutation testing is for. They want the uncomfortable truths. They want to find their blind spots. They're not looking for participation trophies. Meanwhile, people who don't want to deal with escaped mutants can have their nice round numbers and feel good about themselves, without breaking the tool for everyone else.

sanmai:
Actually, the normalization coefficient is only 5%, not 20%:

https://github.com/phpstan/phpstan/discussions/10966#discussioncomment-13438435

maks-rafalko:
I'm sure that people who use SA on top of MT know what they are doing :) See [background explanation here](https://github.com/Roave/infection-static-analysis-plugin?tab=readme-ov-file#background).

Regarding "slower by a factor of 10-100", well, you make it to an extreme, because

- we developed a PoC, and even Ondrej (maintainer of PHPStan) don't know yet why PHPStan takes at the minimum 1s to be executed for 1 changed file. All of this (I'm sure) will lead to an investigation and as a result - we will have faster PHPStan, faster Infection+PHPStan integration. We already have positive outcomes and PHPStan became faster - see https://github.com/Roave/BetterReflection/pull/1505#issuecomment-2895439621
- developers who use SA on top of MT don't expect it to be fast. It wasn't fast before as well. See https://github.com/phpstan/phpstan/discussions/10966#discussioncomment-13438621

> So anyone with 80% MSI will feel better about themselves looking at a nice 100% number in the report. It will have the same effect as running every mutation through SA, with no slowbacks.

I just completely disagree with the last sentence. I treat SA as a "compiler" in PHP. If we mutate something that leads to uncompilable code, it's already killed. And SA can (optionally) help you to find it. Not more, not less.

But I don't know if we need to debate on it - probably SA with MT is just not for you? That's why it's optional. 

> so anyone with 80% MSI will feel better about themselves looking at a nice 100% number in the report.

I don't care about those who will use it incorrectly (it's like saying that we need to remove `assertTrue(true)` from the core of PHPUnit because people can use it and have 100% line coverage - well they can, but it's their problem).

sanmai:
SA is a "compiler," but it does not protect the code from breaking changes, unlike unit tests. With SA and lacking tests, you can introduce the change that breaks the code in some meaningful ways that subpar tests do not notice. The tests are the final arbiter of correctness, not the compiler. You can have perfectly compilable, SA-clean code that's completely wrong.

What does not make sense in what I say?

staabm:
@sanmai thanks for beeing honest and sharing your feelings.

how would you feel if we replace this PR here, with a "post-filter" which ["removes whole code-blocks based on constant-fetch"](https://github.com/infection/infection/issues/2146#issuecomment-2958859090)

I think this would be a change which makes sense no matter how we decide about SA (and it would also have the same result as this one here, I think)

would this address your point in

> I guess I'm more upset that we could not unify all constant-replacing mutators later, but if y'all so keen to optimize this mutation, I guess it makes sense.

?

sanmai:
@staabm just FYI I approved your PR above, there's no reason to hang on this my comment

There could be a mutation that replaces all parts of a complex boolean statement with true or false. I guess we could make an argument that adding negations instead will introduce a smaller number of mutations at the expense of a more complex code all around.

staabm:
Just want to make sure I will not upset anyone of the core team. thank you.

sanmai:
You are definitely not upsetting anyone; you even got two approvals, so please feel free to undraft this PR whenever. My apologies to too harsh of a language, I will try to get a hand of myself.

maks-rafalko:
I feel like I'm more for a solution explained in this comment https://github.com/infection/infection/issues/2146#issuecomment-2958859090 instead of this PR, but since I'm not sure (it's hard for me to analyze all the pros/cons, because different statements make sense), I will leave it without a decision from my side.

what I like is that the number of mutations 2x times less with this change, and SA doesn't produce false-positives anymore.

Let's merge?

maks-rafalko:
thank you @staabm

staabm:
I will play with https://github.com/infection/infection/issues/2146#issuecomment-2958859090 and maybe it leads to something which allows us to revert this PR

Thanks for all the feedback

PR #2181, opened 2025-06-11
Do not increment/decrement integers inside sleep functions

maks-rafalko:
Related to https://github.com/infection/infection/issues/2183

This is a continuation of our goal to decrease the number of "Unproductive Mutants"

See how it's done in Google https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=9524503

Paragraph 3: "Suppressing Unproductive Mutants".

Simple example: it doesn't make sense to do

```diff
- usleep(1000);
+ usleep(1001);
```

cc @staabm

### DIFF
--- a/src/Mutator/Number/AbstractNumberMutator.php
+++ b/src/Mutator/Number/AbstractNumberMutator.php
@@ -53,7 +53,6 @@ abstract class AbstractNumberMutator implements Mutator
         'usleep',
         'time_nanosleep',
         'set_time_limit',
-        'memory_limit',
         'ini_set',
     ];
 
--- a/tests/phpunit/Mutator/Number/DecrementIntegerTest.php
+++ b/tests/phpunit/Mutator/Number/DecrementIntegerTest.php
@@ -485,9 +485,7 @@ final class DecrementIntegerTest extends BaseMutatorTestCase
                 time_nanosleep(1_000, 0);
                 set_time_limit(1_000);
 
-                memory_limit(123); // in bytes
-                memory_limit(-1);  // no memory limit
-
+                ini_set('memory_limit', 500); // in bytes
                 ini_set('memory_limit', -1);
                 PHP,
         ];
--- a/tests/phpunit/Mutator/Number/IncrementIntegerTest.php
+++ b/tests/phpunit/Mutator/Number/IncrementIntegerTest.php
@@ -256,9 +256,7 @@ final class IncrementIntegerTest extends BaseMutatorTestCase
                 time_nanosleep(1_000, 0);
                 set_time_limit(1_000);
 
-                memory_limit(123); // in bytes
-                memory_limit(-1);  // no memory limit
-
+                ini_set('memory_limit', 500); // in bytes
                 ini_set('memory_limit', -1);
                 PHP,
         ];


src/Mutator/Number/AbstractNumberMutator.php
```diff
@@ -47,6 +48,13 @@
  */
 abstract class AbstractNumberMutator implements Mutator
 {
+    private const SLEEP_FUNCTION_NAMES = [
```
staabm: use a more generic name and also add memory limiting functions ?

staabm (commented):
Like the idea

src/Mutator/Number/AbstractNumberMutator.php
staabm: Or maybe even a whitelist of often used ini_set calls

staabm:
I thought about ignoring calls  to Psr3\Logger calls

maks-rafalko:
> I thought about ignoring calls to Psr3\Logger calls

yeah, after reading Google's document, I'm very excited and impressed on the work they have done regarding removing unwanted mutations. Logger is one of those possible improvements. I'm all for this!

They even have an explanation that any call to `log***()` or `logger->***()` should not be mutated, which also does make sense to me.

maks-rafalko:
Analyzed `DecrementInteger` on the whole Infection run, found many other possible improvements. Extracted to https://github.com/infection/infection/issues/2183

src/Mutator/Number/AbstractNumberMutator.php
```diff
@@ -48,11 +48,13 @@
  */
 abstract class AbstractNumberMutator implements Mutator
 {
-    private const SLEEP_FUNCTION_NAMES = [
+    private const UNWANTED_TO_MUTATE_FUNCTIONS = [
         'sleep',
         'usleep',
         'time_nanosleep',
         'set_time_limit',
+        'memory_limit',
```
staabm: this is also done by ini_set: `ini_set('memory_limit', '2048M');`.

I don't think a `memory_limit()` function exists

maks-rafalko: I've added a case for `ini_set`, or what do you mean? could you please add an example of what should be ignored in mutations?

staabm: `ini_set('memory_limit', -1);` should not be mutated.

tests/phpunit/Mutator/Number/DecrementIntegerTest.php
```diff
@@ -476,6 +476,22 @@ public static function mutationsProvider(): iterable
...
+        yield 'It does not decrement when it is used in sleep functions' => [
+            <<<'PHP'
+                <?php
+                usleep(1_000);
+                sleep(1_000);
+                time_nanosleep(1_000, 0);
+                set_time_limit(1_000);
+
+                memory_limit(123); // in bytes
+                memory_limit(-1);  // no memory limit
```
staabm: there is no `memory_limit` function in php

src/Mutator/Number/AbstractNumberMutator.php
staabm:
```
ini_set('max_execution_time', 0);
ini_set('display_errors', 1);
ini_set('log_errors', 1);
ini_set('mbstring.func_overload', 0);
```
 should not be mutated

maks-rafalko: ah, indeed. Removed and added more examples for `ini_set()`

src/Mutator/Number/AbstractNumberMutator.php
```diff
@@ -47,6 +48,14 @@
  */
 abstract class AbstractNumberMutator implements Mutator
 {
+    private const UNWANTED_TO_MUTATE_FUNCTIONS = [
```
theofidry, line 51:
```suggestion
    private const IGNORED_FUNCTIONS = [
```
?

src/Mutator/Number/AbstractNumberMutator.php
```diff
@@ -47,6 +48,14 @@
...
+        'sleep',
+        'usleep',
+        'time_nanosleep',
+        'set_time_limit',
+        'ini_set',
```
theofidry, line 56: I would order them alphabetically

src/Mutator/Number/AbstractNumberMutator.php
```diff
@@ -61,6 +70,33 @@ protected function isPartOfComparison(Node $node): bool
         return $this->isComparison($parent);
     }
 
+    protected function isInsideUnwantedToMutateFunctions(Node\Scalar\LNumber $node): bool
```
theofidry, line 73: this could be private for now

theofidry, line 73: `isParameterOfIgnoredFunction`?

tests/phpunit/Mutator/Number/DecrementIntegerTest.php
```diff
@@ -476,6 +476,20 @@ public static function mutationsProvider(): iterable
                 PHP,
         ];
 
+        // see https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=9524503 "Suppressing Unproductive Mutants"
```
theofidry, line 479: I'm not sure the link to that paper brings much here tbh

tests/phpunit/Mutator/Number/DecrementIntegerTest.php
```diff
@@ -476,6 +476,20 @@ public static function mutationsProvider(): iterable
...
+        yield 'It does not decrement when it is used in sleep functions' => [
+            <<<'PHP'
+                <?php
+                usleep(1_000);
+                sleep(1_000);
+                time_nanosleep(1_000, 0);
+                set_time_limit(1_000);
+
+                ini_set('memory_limit', 500); // in bytes
+                ini_set('memory_limit', -1);
```
theofidry, line 489: I would add more cases, besides valid escaped mutations here an obvious case missing IMO is if you have a legit numeric value in there that is not passed as a parameter of one of those functions be instead an element of an expression.

For instance:

```php
sleep((static fn () => $time === 1000 ? 10 : 100)());
```

quite an artificial case granted, but it illustrates that you have no guarantee that a numeric found within those functions are necessarily passed as a parameter to that function, in which case they should be mutated as usual.

tests/phpunit/Mutator/Number/IncrementIntegerTest.php
```diff
@@ -247,6 +247,20 @@ public static function mutationsProvider(): iterable
...
+        // see https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=9524503 "Suppressing Unproductive Mutants"
+        yield 'It does not increment when it is used in sleep functions' => [
+            <<<'PHP'
+                <?php
+                usleep(1_000);
+                sleep(1_000);
+                time_nanosleep(1_000, 0);
+                set_time_limit(1_000);
+
+                ini_set('memory_limit', 500); // in bytes
```
sanmai, line 259: This mutation is easy to test, unlike with sleep(). 

If something can be tested, then mutating it is perfectly legit.

sanmai (approved):
Good idea, not sure about ini_set tho. At my previous job we had a lot of odd ini_set, and it was very helpful to know what ini_set isn't actually tested. If you modify this mutator like this, it will be no longer possible.

src/Mutator/Number/AbstractNumberMutator.php
sanmai, line 56: I suppose we want to mutate this method only if it gets called with a list of specific common arguments?
```suggestion
```

staabm: Fine by me 👍

tests/phpunit/Mutator/Number/DecrementIntegerTest.php
```diff
@@ -476,6 +476,20 @@ public static function mutationsProvider(): iterable
...
+                PHP,
+        ];
+
         yield 'It does not decrement limit argument of preg_split function when it equals to 0' => [
```
staabm, line 493: I think we should also add a test for a method called `sleep` and takes an `int` but will not be mutated because its not a `FuncCall` (should kill one of the reported escaped mutants)

staabm: another test could be something like

```
$fn = 'doFoo';
$fn(100);
```

sanmai:
I'd also question the need to not-mutate `sleep()` because using `sleep()` in your code directly, and not [one of the excellent abstractions out there](https://github.com/tumblr/chorus-timekeeper), is a bad idea in the first place.

staabm:
> I'd also question the need to not-mutate `sleep()` because using `sleep()` in your code directly, and not [one of the excellent abstractions out there](https://github.com/tumblr/chorus-timekeeper?rgh-link-date=2025-06-13T20%3A55%3A55Z), is a bad idea in the first place.

while you might be right regarding abstractions from a theoretical point of view, I think infection should support reality, which means there is lots of code out there using `sleep` directly:

- https://github.com/search?q=sleep+language%3Aphp&type=code

thats what DX is all about in the end.

sanmai:
If other people do things that lack taste or intelligence, it does not mean that you have to lower your standards; quite the contrary: I would prefer Infection to mutate `sleep()` and `usleep()` for me because if I'm using them directly, I need a refreshing _slap_ on my face. 

Mutation testing in the core of it is not about _supporting the reality,_ but about a quite a _contrarian practice of extreme programming_: each AST symbol used in sources must exist for a reason, or should be whacked out (a [quote from](https://github.com/infection/infection/issues/2146#issuecomment-2955331510) @Ocramius that I can't help but mention).

sanmai (changes requested):
Do we actually think it a great idea? I would prefer Infection to mutate `sleep()` and `usleep()` for me because if I'm using them directly, I need a refreshing _slap_ on my face.

maks-rafalko:
> Do we actually think it a great idea? I would prefer Infection to mutate `sleep()` and `usleep()` (with increment) for me because if I'm using them directly, I need a refreshing _slap_ on my face.

yeas, I think it's a great idea. You are an expert user of Infection, but think about the usual users, especially those who try it _for the first time._

Thousands of mutants found, including

```diff
- usleep(1000);
+ usleep(1001);
```

they will likely close and throw away it rather than trying to fix it.

I'm thinking that such mutations can only exists in some more brutal levels/profiles, but not by default. But levels - is a topic of another issue. So we can conditionally-enable them (and btw we can ask user about the desired "level" during configuration and write it to a config, so you don't even pass any options).

sanmai:
I suppose we have to really revisit the idea of profiles, because there's no other tool that would remind me that I'm using `sleep()` the wrong way. That said, looking at 192 dependent packages, I'm not completely sure that this new user pain point is really universal. How can we validate this hypothesis?

staabm:
> I suppose we have to really revisit the idea of profiles, because there's no other tool that would remind me that I'm using `sleep()` the wrong way

IMO infection can make better suggestions than "you use sleep() wrong".
and IMO thats the reason why we should defer mutations like `sleep(1000)` and provide more meaningfull ones first.

sanmai:
This PR merged will definitely make Infection less useful for me as a poweruser. Yet we have to see any substantial evidence beyond "IMO" that this is a definite pain point.

sanmai:
As a middle ground, I'd propose that we  make one of the mutations (decrement) to walk around these functions, leaving the other (increment) as it is. It will reduce the number of these mutations by half, while keeping the power users such as myself happy. But only if this optimization is gated behind a flag: https://github.com/infection/infection/issues/2312

maks-rafalko:
Or we can block this PR for some time and really create "levels" & "profiles" - seems like it sounds in every ticket we are discussing last weeks :)

I don't want anyone to be limited by our decisions, I want new users to be engaged/interested/excited, and experience users to be satisfied as well. And the only one way of how to achieve it seems like levels/profiles... There is no hurry to merge this one, let's sleep on it. I will bump other issue with levels after some thinking.

staabm:
> Or we can block this PR for some time and really create "levels" & "profiles" - seems like it sounds in every ticket we are discussing last weeks :)

agree

PR #2190, opened 2025-06-13
ReflectionVisitorTest is more type safe

staabm:
cleaning up phpstan baseline errors by using proper return intersection types.

at the same time makes PHPStorm aware of the returned objects and there public properties so it now started to provide autocompletion

### DIFF
--- a/tests/phpunit/PhpParser/Visitor/ReflectionVisitorTest.php
+++ b/tests/phpunit/PhpParser/Visitor/ReflectionVisitorTest.php
@@ -56,7 +56,10 @@ use PHPUnit\Framework\Attributes\Group;
 #[CoversClass(ReflectionVisitor::class)]
 final class ReflectionVisitorTest extends BaseVisitorTestCase
 {
-    private $spyVisitor;
+    /**
+     * @var NodeVisitor&object{isInsideFunction: bool}
+     */
+    private NodeVisitor $spyVisitor;
 
     protected function setUp(): void
     {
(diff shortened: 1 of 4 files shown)


tests/phpunit/PhpParser/Visitor/ReflectionVisitorTest.php
```diff
@@ -239,10 +238,13 @@ public static function isPartOfSignatureFlagWithAttributesProvider(): iterable
         yield [Node\Expr\Array_::class, false];             // []
     }
 
-    private function getPartOfSignatureSpyVisitor(string $nodeClass)
+    /**
+     * @return NodeVisitor&object{isPartOfSignature: bool}
```
staabm, line 245: using a object-shape in intersection with the return-ed class-types makes the calling context aware of the public property name and type

tests/phpunit/PhpParser/Visitor/ReflectionVisitorTest.php
```diff
@@ -300,13 +306,13 @@ public function enterNode(Node $node): ?int
...
 
-    private function getReflectionClassSpyVisitor()
+    /**
+     * @return NodeVisitor&object{reflectionClass: ClassReflection|null}
+     */
+    private function getReflectionClassSpyVisitor(): NodeVisitor
     {
         return new class extends NodeVisitorAbstract {
-            /**
-             * @var ReflectionClass
```
staabm, line 307: using natives types now helped to identify this phpdoc is wrong

maks-rafalko:
thank you @staabm

PR #2200, opened 2025-06-15
fix: break legend into 2 lines

staabm:
before this PR the legend was a pretty long line

grafik

in contrast to most output which seems to be aligned in ~60 chars columns, it stood out and felt overly long.
after this PR the line is wrapped so it does feel more in-line with the sorrounding text

grafik

----

in the end its just cosmetics, but maybe you prefer the newer look in the same way as I do.
I am also fine with closing the PR - no strong feelings on my end


maks-rafalko (approved):
absolutely support this change, thank you 👍

Issue #2185, opened 2025-06-11
Smarter `IncrementInteger`


maks-rafalko:
- [ ] Do not increment `preg_match() === 1` value because `preg_match()` returns `0`, `1`, `false`

```diff
-        if ($nodes->length === 0) {
+        if ($nodes->length === -1) {
```

PR #2204, opened 2025-06-15
Smarter `DecrementInteger` mutator

staabm:
refs https://github.com/infection/infection/issues/2185

### DIFF
--- a/tests/phpunit/Mutator/Number/DecrementIntegerTest.php
+++ b/tests/phpunit/Mutator/Number/DecrementIntegerTest.php
@@ -41,6 +41,7 @@ use const PHP_INT_MAX;
 use const PHP_INT_MIN;
 use PHPUnit\Framework\Attributes\CoversClass;
 use PHPUnit\Framework\Attributes\DataProvider;
+use function sprintf;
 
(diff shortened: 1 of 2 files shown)


tests/phpunit/Mutator/Number/DecrementIntegerTest.php
```diff
@@ -469,6 +469,56 @@ public static function mutationsProvider(): iterable
...
+
+        yield 'It does not decrement zero when it is being compared as identical with result of stripos()' => [
+            <<<'PHP'
+                <?php
+
+                if (stripos($a, $n) === 0) {
+                    echo 'bar';
+                }
+                PHP,
+        ];
```
sanmai: It almost feels like we can shorten this to:
```
foreach (DecrementInteger::NON_NEGATIVE_INT_RETURNING_FUNCTIONS as $name) {
    yield "It does not decrement zero .... $name" => [sprintf('<?php if (%s() === 0) {}', $name)];
}
```

staabm: sure, why not.

maks-rafalko:
thank you @staabm

PR #2207, opened 2025-06-15
Update MutationTestingRunner to stream-filter mutations in buffered mode

### LINKED ISSUE #2196: ParallelProcessRunner is (potentially) blocking & waiting
copying from private messages with @staabm 

> another observation: when running with parallel runner likephp bin/infection --threads=6 - I wonder that the progress summary sometimes completely stalls.
> 
> I have the feeling that it can happen that when a few/one worker(s) is/are blocked, e.g. because they are running in a infinite loop or a slow mutation, they also prevent other worker processes from picking up new jobs and progress.
>
> It feels like that a single worker can block other workers and we are loosing time because of that 

> in case the progress printer 1:1 reflects reality I think it should progress most of the time (as its pretty unlikely that all workers are doing heavy work and none of them is actually progressing).
>
> it feels it happens too often that the summary is standing still and no progress happens for multiple seconds, even though 95% of the mutations process report a max runtime of 0.x seconds 

----

I personally didn't dig into it, but this is the hardest class that I worked with here in Infection because it's not easy to reproduce such issues with tests and/or debug it.

So @sanmai if you will have some time/energy - your help would be very useful as you are the implementor of "buckets" algorithm. 

I can assume that it worked great before and that I broke it with the latest changes somehow, so no pressure, just looking for another pair of eyes.

sanmai:
Fixes #2196

This PR:

- [x] Updates MutationTestingRunner to stream-filter mutations
- [x] Covered by tests

This should eliminate the initial scheduling-related pause, although we no longer use slightly more efficient array-based filtering. I doubt that any slowdown will be noticeable, though: the whole process should take around the same time.

https://github.com/user-attachments/assets/611e5ac1-7eeb-4b49-ac5e-8cf1deb38211

## Testing Steps

```
XDEBUG_MODE=coverage php vendor/phpunit/phpunit/phpunit --coverage-xml=build/logs/coverage-xml --log-junit=build/logs/junit.xml
```
Followed by:
```
php bin/infection --threads=max --skip-initial-tests --coverage=build/logs
```

## Benchmarks

```
$ time chronic php bin/infection --threads=max --skip-initial-tests --coverage=build/logs
real	1m8,915s
user	20m46,386s
sys	6m58,180s

$ time chronic php bin/infection --threads=max --skip-initial-tests --coverage=build/logs --no-progress
real	1m5,900s
user	20m44,020s
sys	6m49,263s
```
Master branch build:
```
(re-collect the coverage)
$ time chronic php bin/infection --threads=max --skip-initial-tests --coverage=build/logs
real	1m9,429s
user	18m38,953s
sys	5m10,544s
```
It might have even became faster.

### DIFF
--- a/src/Mutator/Number/DecrementInteger.php
+++ b/src/Mutator/Number/DecrementInteger.php
@@ -52,13 +52,22 @@ final class DecrementInteger extends AbstractNumberMutator
 {
     use GetMutatorName;
 
-    private const COUNT_NAMES = [
+    public const NON_NEGATIVE_INT_RETURNING_FUNCTIONS = [
         'count',
+        'iterator_count',
         'grapheme_strlen',
         'iconv_strlen',
         'mb_strlen',
         'sizeof',
         'strlen',
+        'strpos',
+        'stripos',
+        'strrpos',
+        'mb_strpos',
+        'mb_stripos',
+        'mb_strrpos',
+        'preg_match',
+        'preg_match_all',
     ];
 
     public static function getDefinition(): Definition
@@ -151,7 +160,7 @@ final class DecrementInteger extends AbstractNumberMutator
         if ($parentNode->left instanceof Node\Expr\FuncCall && $parentNode->left->name instanceof Node\Name
             && in_array(
                 $parentNode->left->name->toLowerString(),
-                self::COUNT_NAMES,
+                self::NON_NEGATIVE_INT_RETURNING_FUNCTIONS,
                 true,
             )
         ) {
@@ -161,7 +170,7 @@ final class DecrementInteger extends AbstractNumberMutator
         if ($parentNode->right instanceof Node\Expr\FuncCall && $parentNode->right->name instanceof Node\Name
             && in_array(
                 $parentNode->right->name->toLowerString(),
-                self::COUNT_NAMES,
+                self::NON_NEGATIVE_INT_RETURNING_FUNCTIONS,
                 true,
             )
         ) {
--- a/tests/phpunit/Mutator/Number/DecrementIntegerTest.php
+++ b/tests/phpunit/Mutator/Number/DecrementIntegerTest.php
@@ -41,6 +41,7 @@ use const PHP_INT_MAX;
 use const PHP_INT_MIN;
 use PHPUnit\Framework\Attributes\CoversClass;
 use PHPUnit\Framework\Attributes\DataProvider;
+use function sprintf;
 
 #[CoversClass(DecrementInteger::class)]
 final class DecrementIntegerTest extends BaseMutatorTestCase
@@ -419,55 +420,9 @@ final class DecrementIntegerTest extends BaseMutatorTestCase
                 PHP,
         ];
 
-        yield 'It does not decrement zero when it is being compared as identical with result of grapheme_strlen()' => [
-            <<<'PHP'
-                <?php
-
-                if (grapheme_strlen($a) === 0) {
-                    echo 'bar';
-                }
-                PHP,
-        ];
-
-        yield 'It does not decrement zero when it is being compared as identical with result of iconv_strlen()' => [
-            <<<'PHP'
-                <?php
-
-                if (iconv_strlen($a) === 0) {
-                    echo 'bar';
-                }
-                PHP,
-        ];
-
-        yield 'It does not decrement zero when it is being compared as identical with result of mb_strlen()' => [
-            <<<'PHP'
-                <?php
-
-                if (mb_strlen($a) === 0) {
-                    echo 'bar';
-                }
-                PHP,
-        ];
-
-        yield 'It does not decrement zero when it is being compared as identical with result of sizeof()' => [
-            <<<'PHP'
-                <?php
-
-                if (sizeof($a) === 0) {
-                    echo 'bar';
-                }
-                PHP,
-        ];
-
-        yield 'It does not decrement zero when it is being compared as identical with result of strlen()' => [
-            <<<'PHP'
-                <?php
-
-                if (strlen($a) === 0) {
-                    echo 'bar';
-                }
-                PHP,
-        ];
+        foreach (DecrementInteger::NON_NEGATIVE_INT_RETURNING_FUNCTIONS as $name) {
+            yield "It does not decrement zero when it is being compared as identical with result of $name" => [sprintf('<?php if (%s() === 0) {}', $name)];
+        }
 
         yield 'It does not decrement when it is accessed zero index of an array' => [
             <<<'PHP'


src/Process/Runner/MutationTestingRunner.php
```diff
@@ -82,6 +82,7 @@ public function run(iterable $mutations, string $testFrameworkExtraOptions): voi
         $this->eventDispatcher->dispatch(new MutationTestingWasStarted($numberOfMutants, $this->processRunner));
 
         $processContainers = take($mutations)
+            ->stream()
```
maks-rafalko, line 85: > Ensures subsequent operations use lazy, non-array paths

could you please add more explanation about what it changes exactly, I'm not sure I understand.

https://github.com/sanmai/pipeline/blob/a293c031a999e0fa954507ff9c06a68aa13974c1/src/Standard.php#L665-L683

it just forces it to be Generator? And how does it change anything?

Without `--no-progress` (by default) we already wait on line 81:

```
$numberOfMutants = IterableCounter::bufferAndCountIfNeeded($mutations, $this->runConcurrently);
```

here, no?

staabm (approved):
confirmed, it fixes my reported problem.

so from my understanding infection worked even before this PR in the background as expected but did not properly report the progress.

thank you.

src/Process/Runner/MutationTestingRunner.php
sanmai: The hold-up happens after this line, when we actually start processing the mutations as in: run tests and see what tests fail.

So before we had one mode that was fully array-based (non-concurrent), and we had a lazy concurrent generator-based mode. Now we have a full generator-based mode, an a partial generator-based mode that collects mutations at one point to count them, and then resumes the generator-based approach. What does not make sense here?

sanmai:
Before this PR, Infection would filter out uncovered mutation first, and only after filtering and preparing all mutations, it will start processing them. After this PR, Infection only does some part of the preparation to count the mutation, and for the remainder uses a concurrent mode.

src/Process/Runner/MutationTestingRunner.php
maks-rafalko: > What does not make sense here?

I feel like I just don't understand some concepts here yet.

On @staabm's original recording, there were 3 freezes:

> 1.    second 42-52 no progress (I guess mutations get generated at this point; interessting, but not the problem this issue is about)
> 2.    1:08 - 1:17 no progress (these are the stalls my initial report is about; I have 6 workers but at this second 3 run into timeout. why do the other 3 don't create progress in this meanwhile)
> 3.    2:11 - 2:35 no progress - similar behaviour as above

from what I understood, this PR fixes point 1?

What about 2 and 3? I don't get the reason of 2 and 3 freezes, that's why I don't get how this PR can fix them.

At the time of freeze 2 and 3, on `master` branch we already have all the Mutants prepared, don't we? So my understanding that 2 and 3 shouldn't be fixed.

maks-rafalko:
> although we no longer use slightly more efficient array-based filtering

could you also please explain why array-based filtering is "slightly more efficient"? Is it something particularily related to `sanmai/pipeline` or a general "generator vs arrays" theory?

sanmai:
`array_filter()` operates on an existing array in memory using internal C code optimized for performance, where Generator-based filters add function call overhead and whole a lot more stuff in plain PHP instead of optimized C. That's why the pipeline library tries to use arrays if the lazy evaluation is not needed. 

That said, when lazy evaluation is needed (such as with large or infinite streams), the benefits of generators outweigh the costs with reduced memory usage, stepwise execution etc. so inefficiencies of generators become negligible compared to the actual work the pipeline is doing. That's why I used "slightly."

@maks-rafalko What leaves you concerned here?

src/Process/Runner/MutationTestingRunner.php
sanmai: https://github.com/user-attachments/assets/6d336bf7-5b5b-4d02-8a3e-110696c2a05f

Let's review. I'm uploading it now, editing the comment later.
- second 42-52 no progress - should be gone with this fix
- 1:08 - 1:17 no progress - is due to timeouts, that's expected
- 2:11 - 2:35 no progress - also due to timeouts

So we are only fixing the first freeze, as it can be fixed. The other two happen because we have a limited pool of processes that we can launch, so if several processes spin in an endless loop, we can't expect any other process to start. That's my best understanding.

If you think that timeout-related freezes are investigation worthy (I recall they existed even before the bucketed process, for years), we can look at them too, and to be clear - I'm not fixing them here.

maks-rafalko: > 1:08 - 1:17 no progress - is due to timeouts, that's expected

3 timeouts but 6 threads - there shouldn't be a delay IMO.

It doesn't block this PR since it fixes p.1, but I have a feeling that we still, potentially, have an issue with p.2. / p.3

sanmai: I updated the comment, have a look. I recall this was a thing way back in time.

maks-rafalko:
> @maks-rafalko What leaves you concerned here?

Nothing, what I'm trying to achieve by asking questions:

- to understand every single character in this diff because I will have to maintain it. As I wrote somewhere, generators in Infection is one of the hardest thing, probably because I don't use them a lot
- to have as much details as possible for the future. I find it very useful to re-read in years what we are discussing here

----

So, now it's clear that only p.1 from @staabm's original video is fixed here, that was one of my main question, because Markus wrote "confirmed, it fixes my reported problem.", which is only partially true. Points 2-3 are still not clear.

maks-rafalko (approved):

src/Process/Runner/MutationTestingRunner.php
maks-rafalko: > so if several processes spin in an endless loop, we can't expect any other process to start. That's my best understanding.

what does block them? OS? our code? 

I posted a branch where I had 5 timeouts (endless loops) with `--threads=6` and that 1 thread continued to working, so wasn't blocked.

sanmai: Here's how you can think about it: when you see two `TT` appears, for every `T` it is not something that happened right now, but rather a point when a process got timed-out that was started full 25 seconds ago. So for the last 25 seconds we were going with only 4 threads instead of 6, with two threads taken by looping processes; and when the freeze occurred, we have even more "spinning" processes occupying the quota, but we will only see their `T`s way later, once they actually timeout. 

So I think the idea is that `T` is something that has been ongoing for full 25 seconds already. Make sense?

We can certainly investigate further, but it looks like a sensible thing to me.

sanmai:
@maks-rafalko Let me know if you need anything else explained; I have no interest in leaving you holding the bag, even by accident.

src/Process/Runner/MutationTestingRunner.php
maks-rafalko: yeah it makes sense, yet suspicious :) Let's keep an eye on it, I don't think we need to investigate it right now cause it's very hard to reproduce (at least for me)

PR #2208, opened 2025-06-16
Smarter `IncrementInteger` mutator

staabm:
closes https://github.com/infection/infection/issues/2185

counterpart of https://github.com/infection/infection/pull/2204 which prevented decrementing below the lower bound.

---

there are a few more functions which could also use the same logic.. since those are used rarely I think we should not take them into consideration

grafik


maks-rafalko (approved):
👍

maks-rafalko:
thank you @staabm

PR #2215, opened 2025-06-16
ResultsCollector returns list

staabm:
refine return types to ease implementation of POC


staabm:
autoreview failed while it works locally. seems a bogus CI run

maks-rafalko:
thank you @staabm

PR #2219, opened 2025-06-16
Killed timed-out mutant in ExcludeDirsProvider

staabm:
`bin/infection --threads=8 --filter=src/Config/ValueProvider/ExcludeDirsProvider.php`

before this PR:
> Time: 44s. Memory: 18.00MB. Threads: 8

after this PR:
> Time: 19s. Memory: 18.00MB. Threads: 8

### DIFF
--- a/src/Config/ValueProvider/ExcludeDirsProvider.php
+++ b/src/Config/ValueProvider/ExcludeDirsProvider.php
@@ -113,7 +113,7 @@ final readonly class ExcludeDirsProvider
         $question->setAutocompleterValues($autocompleteValues);
         $question->setValidator($this->getValidator(new RootsFileOrDirectoryLocator($sourceDirs, $this->filesystem)));
 
-        while (($dir = $this->questionHelper->ask($io->getInput(), $io->getOutput(), $question)) !== '') {
+        while ($dir = $this->questionHelper->ask($io->getInput(), $io->getOutput(), $question)) {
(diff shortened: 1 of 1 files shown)


maks-rafalko:
please ignore tests-checker, it's dead again :| I will investigate.

Could you please explain the nature of this fix: why now we don't have a timeout?

staabm:
> Could you please explain the nature of this fix: why now we don't have a timeout?

the `NotIdentical` mutators changes the while-condition into

`while (($dir = $this->questionHelper->ask($io->getInput(), $io->getOutput(), $question)) === '') {`

which is a endless loop, when `ask` returns the default-answer which is defined a few lines above:

`$question = new Question($questionText, '');`

staabm:
> please ignore tests-checker, it's dead again :| I will investigate

maybe we should implement a PHPStan rule instead with the same logic as the test-checker.
that way we would not need a separate job for this case and it might be doable in a few lines of code - without an additional project to maintain

maks-rafalko:
sounds interesting. Why I wanted it to be as a GH action is because it's language agnostic, and can be used by any language (and it's used by other repositories). But PHPStan rule will be limited to PHP only. I will continue investigation but will keep this idea in mind if I won't get a luck, thanks :)

maks-rafalko (approved):

maks-rafalko:
thank you @staabm

PR #2220, opened 2025-06-16
debug tests-checker failures

maks-rafalko:
-


maks-rafalko:
hm... seems like it works, but not from forks. Will reenable integration as suggested here https://github.com/orgs/community/discussions/60820#discussioncomment-6445767

PR #2216, opened 2025-06-16
Support `--show-mutations=int|max` to support non-bool values

staabm:
POC for https://github.com/infection/infection/issues/2209

output with this PR:

```
➜  infection git:(cli-mut) ✗ php bin/infection --threads=max --skip-initial-tests --coverage=build/logs --show-mutations=10 --filter=src/Command
Xdebug: [Step Debug] Could not connect to debugging client. Tried: localhost:9003 (through xdebug.client_host/xdebug.client_port).

    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/

#StandWithUkraine

Infection - PHP Mutation Testing Framework version dev-master

[warning] Skipping the initial test run can be very dangerous.
It is your responsibility to ensure the tests are in a passing state to begin.
If this is not done then mutations may report as caught when they are not.


Generate mutants...

Processing source code files: 5/5
.: killed by tests, A: killed by SA, M: escaped, U: uncovered
E: fatal error, X: syntax error, T: timed out, S: skipped, I: ignored

IIIIIUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUIUU....M..M   ( 50 / 241)
.M.............U.MM..MMM..............M.UUUUUUUUUU   (100 / 241)
UUUUUU....MUUUUUUU....M...UUUUUUUUUUUUUUUUUUUUUU..   (150 / 241)
...MM.MMMMMUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUU   (200 / 241)
UUUUUUUUUUIUUUUUUUUUIUUUUMMMMMMMMMMMMMMTT            (241 / 241)
Escaped mutants:
================


1) /Users/m.staab/Documents/GitHub/infection/src/Command/BaseCommand.php:68    [M] MethodCallRemoval

@@ @@
     }
     protected function initialize(InputInterface $input, OutputInterface $output): void
     {
-        parent::initialize($input, $output);
+        
         $this->io = new IO($input, $output);
     }
     protected function execute(InputInterface $input, OutputInterface $output): int


2) /Users/m.staab/Documents/GitHub/infection/src/Command/DescribeCommand.php:69    [M] MethodCallRemoval

@@ @@
[...]

PR #2225, opened 2025-06-17
Re-implement ContainerTest as a PHPStan rule

staabm:
creates a new PHPStan rule which errors when code tries to invoke `new Container()` directly, instead of using `SingletonContainer->getContainer()` instead.

the old test was pretty rough, as it [false-positives on e.g. public class references](https://github.com/infection/infection/pull/2216#discussion_r2151709538). The new rule is working on AST, which means it only hits the `new`-keyword

the previous unit test traversed a big part of the filesystem tree and loaded every class found via reflection.
instead we now use the 'standard' PHPStan way of things, meaning static reflection and re-analyzing only when necessary etc.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -107,7 +107,8 @@
     "autoload-dev": {
         "psr-4": {
             "Infection\\Benchmark\\": "tests/benchmark",
-            "Infection\\Tests\\": "tests/phpunit"
+            "Infection\\Tests\\": "tests/phpunit",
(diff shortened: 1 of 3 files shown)


staabm:
I could add tests, but did not as the previous code was also not tested. 

Waiting for directions :-)

maks-rafalko:
thank you @staabm

Issue #2191, opened 2025-06-13
Add alias for `--test-framework-options="--filter='tests/MyExampleText.php'"` or rename options


maks-rafalko:
Extracted from #2126

Desire: to run Infection for 1 file running 1 test file.

Currently, it can be done like

```
infection --filter=SourceFile.php --test-framework-options="--filter='SourceFileTest.php'" 
```

but it's too verbose. See
 
> ... and a command like `--test-framework-options="--filter='tests/MyExampleText.php'"` looks hard to remember and todo right without having to lookup the docs :) 

 _Originally posted by @staabm in [#2126](https://github.com/infection/infection/issues/2126#issuecomment-2969339020)_

**Problem 1:** 

We need to have something simpler than `--test-framework-options="--filter='SourceFileTest.php'"`. Probably a new option `--tests` and an alias for it `-t=SourceFileTest.php`?

**Problem 2:**

> I agree that I was also confused about infection `--filter` vs. PHPUnit `--filter`.

@staabm what was a confusing with `--filter`, do you suggest to rename it? If so, what are the suggestions? 

- `--source` to mimic infection.json5's `src`, and a pair `--tests` as an alias for `--test-framework-options="--filter='tests/MyExampleText.php'"`

cc @sanmai

maks-rafalko:
copy @theofidry's comment here to not have an offtopic in #2126

> I wish it would be a lot more straightforward:
> 
>
> ```
> infection path/to/my/file
> ```
> 
> From its mapping, Infection can know if it's a test file or a source file.
> 
> 1. if it's a file but outside the config: would yield an error.
> 2. if it's a test file, then the test framework filter can be employed to limit the executed tests, which produces a small coverage which should be sufficient to speed things up
> 3. if it's a source file we currently are limited by our tracing options:
>     - coverage, i.e. we need to work with existing pre-produced coverage (useless in this scenario as too inconvenient to employ) or we need to run all tests to get that coverage
>     - I believe we have a `--map-source-class-to-test` somehow that we can leverage
> 
> I think improving the tracing in general would be key for the 3) workflow. Ideally the testing framework should be able to provide a coverage map of some sorts (or if the testing framework is unwilling to, this could be done in the bridge on our side). > With the expectation that this process is quick, you would then know in advance the select few test files you want to execute.
> 
> Then on top of this you could have additional filters to limit what mutators you want to all. E.g.:
> 
> ```
> infection path/to/my/file --mutator=FalseValue,LogicalAnd
[...]

Issue #2126, opened 2025-06-06
Slow feedback loop working with Infection and fixing Mutants for PR


maks-rafalko:
Extracting to a separate issue to gather ideas and discuss.

Based on @staabm feedback from https://github.com/infection/infection/pull/2121

> what currently is a bit frustrating: running the mutation testing has a _pretty slow feedback loop_ (looking at mutation results; think about a possible fix to kill the mutant; add a test; re-run the suite to verify the mutant was killed as assumed).
> 
> In the state like it is right now, where the mutation testing results in 15+ mutants I think I loose a lot of time by manually scanning the result list and verifying whether the single mutant I tried to kill was actually killed. it would be great to get some kind of "difference view" (which mutants were killed/are no longer killed in comparison to a certain state before / I think thats similar to the PHPStan baseline concept).
>
> even using `time bin/infection --git-diff-lines --threads=max -s` the feedback cycle is already 25 seconds.
> 
> I can't imagine how super frustrating it must be in case you are working on a project in which the test-suite takes multiple minutes to run (which translates into super super slow mutation testing)

_Originally posted by @staabm in https://github.com/infection/infection/issues/2121#issuecomment-2948384649_
            
----

Also, when we run `--threads > 1`, results are always re-ordered, which can make it even more hard to find "that killed mutant" in the output.

I would really like to gather any ideas we can apply to improve it.

For me personally:

- [ ] investigate if there is any technique in other MT libraries from other languages
  - [ ] ask on Discord in MT server

Ideas from the top of my head:

- to show killed mutants comparing to "the last run" to easier understand what was (not) killed
- [x] to have some IDs for mutants so we can get escaped mutant from the log with ID=MUTANT_HASH, and run `infection --filter=MUTANT_HASH` or smth like this to get an instant feedback about its new status (killed or still escaped). So ideas is the same as you work with test: you change the code and re-run the test. Here you change the code&test and re-run the Mutant (implemented https://github.com/infection/infection/pull/2226)
- to reuse all our optimizations like `--map-source-class-to-test` and similar to run only those tests that cover Mutant
- to think about what can be improved on PHPUnit side so that if we change the code in SourceFileA.php, it only runs tests for this file but not others. This is needed to reduce "initial tests run" stage. @staabm told me that he is working on a feature that can tell PHPUnit about the changes in the dependencies tree
[...]

PR #2226, opened 2025-06-17
Introduce `--id=<mutant-hash>` to run the only one Mutant for killing it

maks-rafalko:
Related to:

- https://github.com/infection/infection/issues/2126
- https://github.com/infection/infection/issues/2191

# Problem

It's not convenient if you want to kill a particular Mutant to run Infection and see the output with N escaped Mutants - it is a noise and you are just wasting the time looking for a needed one.

You need an instant feedback: whether it was killed after updating the tests or whether it's still escaped.

# Solution

Now, we can run

```shell
infection --id=<mutant-hash>
```

and Infection will only run just only produce just only that Mutant.

- when it's killed - you immediately see it
- when it's still escaped - you immediately see a diff in the terminal

Pros:

- we don't wait for `N-1` other mutants tried to be killed

Known issues:

- hash can be invalidated if source code before this Mutant is changed
- hash is non-decodable (well - because it's hash), so we can not extract the file of the Mutant by its hash, to speed up "initial tests suite"

I had an idea to make decodable strings, so that we can encode e.g.

```
src/SomeSourceFile.php:9:16;InstanceOf_
```

so that after decoding we know where (in what file) this Mutant was generated. This would allow us to significantly reduce "initial tests run", because we could under the hood add `--filter=src/SourceFile.php` (related to https://github.com/infection/infection/issues/2191#issuecomment-2973746796). But since this is a completely standalone, very hard problem, for now we can leave without it.

For now, it can, and should, be used together with `--filter` and if you want to speed up "initial tests run" - with `--map-source-class-to-test` or with more verbose `--test-framework-options='--filter=xxx'` (separate issue is created to make it more developer-friendly https://github.com/infection/infection/issues/2191)

```
infection --filter=src/SourceClass.php --id=<hash> --map-source-class-to-test
```

and this will give you super fast feedback.

----

how it looks like:

```shell
11) /opt/infection/tests/e2e/Example_Test/src/SourceClass.php:9    [M] Minus [ID] 6fa7eadad5b9fd1a72e4d80c83a61cc3

@@ @@
 {
     public function hello(): string
     {
-        $x = 1 + 2 - 3 * 4;
+        $x = 1 + 2 + 3 * 4;
         return 'hello';
     }
 }
```

This PR:

- [x] Adds new feature `--id` to run only one mutant at a time
- [x] Add more tests to kill all new mutants
- [x] Doc PR: https://github.com/infection/site/pull/272
- [x] implement in `text` file logger as well (if needed)


tests/phpunit/Configuration/ConfigurationFactoryTest.php
```diff
@@ -1342,6 +1350,7 @@ private static function createValueForCoveragePath(
             MapSourceClassToTestStrategy::SIMPLE,
             null,
             StaticAnalysisToolTypes::PHPSTAN,
+            null, // $mutantId
```
sanmai, line 1354: I wonder if a default value can reduce the boilerplate.

src/Mutator/NodeMutationGenerator.php
```diff
@@ -106,7 +107,14 @@ public function generate(Node $node): iterable
...
 
         foreach ($this->mutators as $mutator) {
-            yield from $this->generateForMutator($node, $mutator);
+            foreach ($this->generateForMutator($node, $mutator) as $mutation) {
+                if ($this->mutantId !== null && $mutation->getHash() !== $this->mutantId) {
+                    continue;
+                }
+
+                // todo check that prepending code with same mutator doesn't invalidate ID
+                yield $mutation;
```
sanmai: A potentially better place for this is where we do all other filtering; also it feels like we are breaking the SRP here 🤔

sanmai:
I love this! 

I wonder what do you think about using a different, less sensitive hash string? 

I don't know exactly what string, but imagine the this: sometimes to eliminate a mutation you not only need to improve tests, but also work on the code, and if you add just a single line, all mutation ids will go to the trash.

staabm:
Thanks for working on this important problem.

> * hash can be invalidated if source code before this Mutant is changed

I think this side-effect will ruin the party.

Lets look again at the problem we are trying to solve.
I think the important point is, that we want to find some kind of "scope" the analysis should run on, so the number of mutants is reduced to a small list (not necessarily a single mutant).

what do you think instead of `infection --id=<mutant-hash>` we use a filter which is similar to the ones used in PHPUnit namely `phpunit --filter {TestMethodName} {FilePath}`. I think this would already bring us a big step forward and it would also prevent the "if-source-changes" problem.

In infection we could use `infection --filter {methodName} {FilePath}`

Pros
- the identifier can be guessed by the user, even without running infection beforehand
- the identifer will not change as long as the methods' name not change (and even then the user can predict how a new identifier needs to look like)
- the identifier format is known to people already using PHPUnit

Cons
- its not a single mutant, but a handful or something -> this could also be a advantage as you see how your change affects the code arround the location you are testing (nearby mutants) -> so its not too focused on a single problem

tests/phpunit/Configuration/ConfigurationFactoryTest.php
staabm: I think our data-providers with dozens of array-elements are very unreadable.

just did a small test which verifies, that we could also utilze named arguments:

```php
<?php

class MyTest extends \PHPUnit\Framework\TestCase {
    #[\PHPUnit\Framework\Attributes\DataProvider('provideX')]
    public function testX(string $name, int $id)
    {
        $this->assertTrue(true);
    }

    static public function provideX(): iterable {
        yield [
            'name' => 'test1',
            'id' => 1,
        ];

        yield [ // demonstrates the order of elements does not matter since phpunit passes the elements based on its name
            'id' => 1,
            'name' => 'test1',
        ];

        yield [ // errors because wrong type: TypeError: MyTest::testX(): Argument #1 ($name) must be of type string, int given
            'id' => '1',
            'name' => 123,
        ];
    }
}
```

maks-rafalko:
> I wonder what do you think about using a different, less sensitive hash string?
> 
> I don't know exactly what string, but imagine the this: sometimes to eliminate a mutation you not only need to improve tests, but also work on the code, and if you add just a single line, all mutation ids will go to the trash.

Yeah I understand this, and added it to the original description in "Known issues" - it's kind of ok. Yes, this is a limitation, but I have no idea how we can make an encoded string without ruining it if some of the lines before or on the same line of a mutated has changed. 

On the other side - this is just a half of the cases. Other mutants can be killed by changing the tests, which doesn't break this feature.

If someone has an idea regarding such an algorithm of encoding - let me know please.

> I think this side-effect will ruin the party.

Well, not exactly. As I wrote above, this is only 1 of 2 cases, where source code has to be changed to kill a Mutant. When you change only tests - it's still useful.

> what do you think instead of infection --id=<mutant-hash> we use a filter which is similar to the ones used in PHPUnit namely phpunit --filter {TestMethodName} {FilePath}. I think this would already bring us a big step forward and it would also prevent the "if-source-changes" problem.

I like it. But why *instead of*? It can live together with `--id`, so these features are not exclusive at all. I mean, if you want - use `--id`. If you want more extended scope - use your idea. If you agree, we can extract it to a separate issue and implement independently of this one.

Please note that even 1 line can have tens (!) of mutants, not saying about the whole method. So suggested feature is in no way defeats the purpose of this PR, IMO.

Another feature that we can add to your idea is to run mutations on particular line ([like we have for `ignore` config](https://infection.github.io/guide/how-to.html#Disable-in-particular-class-or-method-or-line)), e.g.:

- `SourceClass::someMethod` - mutate the whole method `someMethod` method in class `SourceClass`
- `SourceClass::someMethod:32` or just `SourceClass::32` - mutate line 32  on class `SourceClass` on line 32

but again, I would suggest to extract it to a separate issue if you don't mind.

sanmai:
Also, `SourceClass::someMethod::SomeMutator`

staabm:
> I like it. But why _instead of_? It can live together with `--id`, so these features are not exclusive at all. I mean, if you want - use `--id`. If you want more extended scope - use your idea. If you agree, we can extract it to a separate issue and implement independently of this one.

my current feeling is that the `--id` variant is not practical as I need to need a full run beforehand and the above mentioned side-effects of changening sources render the mutant identifier useless (so lots of constraints/pre-requisites -> not good DX wise, or you need to explain it a lot).

the method variant seem not affected by this and easy to grasp.

but thats of course my personal point of view

maks-rafalko:
> my current feeling is that the `--id` variant is not practical as I need to need a full run beforehand

I would not agree here. My personal way of using Infection is the following:

- I wrote code, then tests
- then I run `infection --git-diff-lines --map-source-class-to-tests`

and only if there were some escaped mutatns, I start killing them. So, 2 important points here:

- I run infection in any way
- and this is not "the whole run", it's a very limited in scope (to changed lines only) and fast (`--map-source-class-to-tests`) run

And then, my assumption that I will just copy an ID and will start killing it. (I will press arrow `up` in the terminal to repeat previous command and will just type `--id=<ctrl+v>`.

On the other side, with your idea, we can copy `/opt/infection/src/Mutator/NodeMutationGenerator.php:111` to `--filter` (or whatever we will implement) and limit execution to one line. But anyway, here we could also have a problem - if line has many mutations, we are back to the original problem - it's noisy.

I still think we can implement both, and either later decide what is better, or just leave both. Cause all of the above is an assumption, I have never worked with such a workflow yet.

maks-rafalko:
Ok, I've fixed comments, extracted Markus' idea to https://github.com/infection/infection/issues/2237, created docs.

Let's try it..

tests/phpunit/Configuration/ConfigurationFactoryTest.php
theofidry: I had that issue in a lot of my projects, one thing that worked very well is introducing either a static method to create the data set (which provides named parameters and/or the IDE shows the parameter names) and in other cases a `Scenario` object that is a tiny builder used to build the data set data.

Issue #2183, opened 2025-06-11
Smarter `DecrementInteger`


maks-rafalko:
- [x] Do not increment/decrement in sleep and memory functions  https://github.com/infection/infection/pull/2181

----

- [x] Do not decrement more complex cases with count/length comparison  (it doesn't make sense) (https://github.com/infection/infection/pull/2238)

```diff
-        if ($nodes->length === 0) {
+        if ($nodes->length === -1) {
```

or

```diff
-        if ($coveredLineNodes->length === 0) {
+        if ($coveredLineNodes->length === -1) {
```

or

```diff
-        if ($totalCount !== 0) {
+        if ($totalCount !== -1) {
```

or 

```diff
-        if ($testedTotal !== 0) {
+        if ($testedTotal !== -1) {
```

----

- [x] Do not decrement when call contains `**numberOf**()` in comparison with 0 (it doesn't make sense): (https://github.com/infection/infection/pull/2238)

```diff
-        if ($constructor->getNumberOfParameters() === 0) {
+        if ($constructor->getNumberOfParameters() === -1) {
```

----

- [x] Do not decrement assigning 0 to "count" variable: (https://github.com/infection/infection/pull/2238)

```diff
-        $this->callsCount = 0;
+        $this->callsCount = -1;
```

----

- [x] Do not decrement comparison mb_strpos() with 0: (https://github.com/infection/infection/pull/2204)

```diff
-        while ($start !== 0 && mb_strpos($nextLine, mb_substr($previousLine, 0, $start)) !== 0) {
+        while ($start !== 0 && mb_strpos($nextLine, mb_substr($previousLine, 0, $start)) !== -1) {
```

because position can not be negative

----

- [x] Do not decrement comparison of preg_match() with 0 as it can't return -1: (https://github.com/infection/infection/pull/2204)

```diff
-        preg_match('/^(\+|-|index)/', $line) === 0);
+        preg_match('/^(\+|-|index)/', $line) === -1);
```

----

Read https://github.com/infection/infection/issues/2183#issuecomment-2972201036 before - probably should not be done

- [ ] Do not decrement exit codes from 0 to -1:

```diff
-        if ($process->getExitCode() === 0 && $this->testFrameworkAdapter->testsPass($output)) {
+        if ($process->getExitCode() === -1 && $this->testFrameworkAdapter->testsPass($output)) {
```

theofidry:
[...]

PR #2238, opened 2025-06-19
Smarter `DecrementInteger` mutator

staabm:
refs https://github.com/infection/infection/issues/2183

### DIFF
--- a/tests/phpunit/Mutator/Number/DecrementIntegerTest.php
+++ b/tests/phpunit/Mutator/Number/DecrementIntegerTest.php
@@ -551,6 +551,38 @@ final class DecrementIntegerTest extends BaseMutatorTestCase
                 PHP,
         ];
 
+        yield 'It does not decrement with numberOf methodCall comparison left' => [
+            <<<'PHP'
+                <?php
+
+                if ($constructor->getNumberOfParameters() === 0) {}
+                PHP,
+        ];
+
+        yield 'It does not decrement with numberOf methodCall comparison right' => [
+            <<<'PHP'
+                <?php
+
+                if (0 === $constructor->getNumberOfParameters()) {}
+                PHP,
+        ];
+
+        yield 'It does not decrement with numberOf nullsafe methodCall comparison left' => [
+            <<<'PHP'
+                <?php
+
+                if ($constructor?->getNumberOfParameters() === 0) {}
+                PHP,
+        ];
+
+        yield 'It does not decrement with numberOf nullsafe methodCall comparison right' => [
+            <<<'PHP'
+                <?php
+
+                if (0 === $constructor?->getNumberOfParameters()) {}
+                PHP,
+        ];
+
         yield 'It does not decrement with count nullsafe property comparison left' => [
             <<<'PHP'
                 <?php
@@ -599,7 +631,7 @@ final class DecrementIntegerTest extends BaseMutatorTestCase
                 PHP,
         ];
 
-        yield 'It does not decrement with *count* property assignment left' => [
+        yield 'It does not decrement with *count* property assignment' => [
             <<<'PHP'
                 <?php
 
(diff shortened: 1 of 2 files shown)


tests/phpunit/Mutator/Number/DecrementIntegerTest.php
```diff
@@ -490,5 +490,145 @@ public static function mutationsProvider(): iterable
...
+
+        yield 'It does not decrement with *length* variable comparison right' => [
+            <<<'PHP'
+                <?php
+
+                if (0 !== $xyzLength) {}
+                PHP,
+        ];
+
+        yield 'It does not decrement with *count* property comparison left' => [
```
maks-rafalko:
```suggestion
        yield 'It does not decrement with *count* property assigning left' => [
```

same below

staabm:
did leave out `exit` for now, as [we still discuss](https://github.com/infection/infection/issues/2183#issuecomment-2972201036) it without a agreement on all sides

```diff
-        if ($process->getExitCode() === 0 && $this->testFrameworkAdapter->testsPass($output)) {
+        if ($process->getExitCode() === -1 && $this->testFrameworkAdapter->testsPass($output)) {
```

maks-rafalko:
thank you @staabm

tests/phpunit/Mutator/Number/DecrementIntegerTest.php
```diff
@@ -490,5 +490,189 @@ public static function mutationsProvider(): iterable
...
+
+        yield 'It does not decrement with *length* property comparison left' => [
+            <<<'PHP'
+                <?php
+
+                if ($nodes->length === 0) {}
+                PHP,
+        ];
+
+        yield 'It does not decrement with *length* property comparison right' => [
```
theofidry, line 558: do we really want to make assumptions that the consumer code is correct or does what it says?

theofidry: I see where this comes from, but I wonder if it wouldn't be more suited to allow on a project basis via the ignore configuration*, rather than doing it here

*: I do believe our ability to ignore things need to be improved, at least so far I've always had issues with those regexes, but that's a separate problem

maks-rafalko: this eliminates a lot of completely useless mutations, so why not? later, when we have levels, I think we can make it configurable if anyone wants it, but by default I'm pretty sure that anything containing "length" shouldn't be below 0

theofidry: >this eliminates a lot of completely useless mutations, so why not? [...] but by default I'm pretty sure that anything containing "length" 

If the assumption is you're working on good code, then yes... And we are inspecting OSS code which I tend to think is of higher quality than your average private or closed source code. So benchmarking that against projects like infection itself, phpstan or valinor is quite biased.

I don't know if it's right for Infection to really do those assumptions. You don't know the quality of the consumer code neither its nature. Typical example: maybe length is non-sensical but constrained to an `int` in which case instead of `null` you would have `-1`. Or maybe it allows `NAN` too... If we work with the inferred types from a SA then we could do smarter decisions there, but without it it seems risky to me.

So while Infection needs to work out how to be efficient, there is also the question of _trust_ from using Infection.

>when we have levels, I think we can make it configurable if anyone wants it

Totally reasonable. I assume once we add some weight to mutations we could adjust the level based on those weights.

<hr>

I must say in the end though, the only risk is Infection missing things it could otherwise highlight, it's not the end of the world either

staabm: If I get the point of this discussion right, the point is that because of guesswork we might miss important mutations because people might name things with different assumptions. 

And if we get the level concept going this might mean on higher levels we do less guesswork but shift the responsibility to the infection user. 

This means easier use of infection for people on lower level and more work/precise results on higher levels - which makes sense to me (less hurdles on lower level and more involved mutations on higher levels for users with more experience)

PR #2242, opened 2025-06-21
Cleanup MemoizedCiDetector, NullCiDetector

staabm:
stumbled over this class because of a PHPStorm warning

grafik

----

see https://www.php.net/manual/en/reflectionproperty.setaccessible.php

> As of PHP 8.1.0, calling this method has no effect; all properties are accessible by default. 

----

looking at https://github.com/OndraM/ci-detector/pull/65 it seems we no longer need this whole method


maks-rafalko (approved):
👍

Issue #2232, opened 2025-06-19
Smarter LogicalAndAllSubExprNegation with `instanceof`


staabm:
in we see

```diff
1) /Users/staabm/workspace/infection/src/Mutator/Arithmetic/Decrement.php:93    [M] LogicalAndAllSubExprNegation

@@ @@
     }
     public function canMutate(Node $node): bool
     {
-        if (!$node instanceof Node\Expr\PreDec && !$node instanceof Node\Expr\PostDec) {
+        if ($node instanceof Node\Expr\PreDec && $node instanceof Node\Expr\PostDec) {
             return false;
         }
         $parentNode = ParentConnector::findParent($node);
```

which does not make sense, as a single variable cannot be `instanceof` 2 different concrete classes.
this would only work for e.g. a class and a interface, or 2 interfaces.

staabm:
we can do the same for

```diff
35) /Users/staabm/workspace/infection/src/AstFilter/PublicVisibility.php:96    [M] LogicalOr [ID] 9cd21ffae5add3fb650c4447396bbbdb

@@ @@
         }
         $typeName = $class->getName();
         foreach (array_reverse($methodCallChain) as $chainElement) {
-            if ($chainElement instanceof Node\Expr\MethodCall || $chainElement instanceof Node\Expr\NullsafeMethodCall) {
+            if ($chainElement instanceof Node\Expr\MethodCall && $chainElement instanceof Node\Expr\NullsafeMethodCall) {
                 $methodName = $chainElement->name;
                 if (!$methodName instanceof Node\Identifier) {
                     return;

$ 
```

which is impossible for similar reasons

PR #2243, opened 2025-06-23
Smarter `LogicalOr` Mutator

staabm:
closes https://github.com/infection/infection/issues/2232

### DIFF
--- a/src/Mutator/Boolean/LogicalOr.php
+++ b/src/Mutator/Boolean/LogicalOr.php
@@ -40,8 +40,12 @@ use Infection\Mutator\Definition;
 use Infection\Mutator\GetMutatorName;
 use Infection\Mutator\Mutator;
 use Infection\Mutator\MutatorCategory;
+use Infection\Mutator\Util\NameResolver;
+use function is_string;
 use LogicException;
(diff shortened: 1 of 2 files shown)


src/Mutator/Boolean/LogicalOr.php
```diff
@@ -51,6 +55,7 @@
 final class LogicalOr implements Mutator
 {
     use GetMutatorName;
+    private ?string $seenVariabeName = null;
```
theofidry, line 59:
```suggestion

    private ?string $seenVariabeName = null;
```

I wish PHP-CS-Fixer would pick that up but I can't see the rule that does that

theofidry (approved):
I guess there's still one test case missing? LGTM otherwise

maks-rafalko (approved):
👍

staabm:
> I guess there's still one test case missing?

which one? could you give me a hint?

src/Mutator/Boolean/LogicalOr.php
sidz: `class_attributes_separation`

PR #2244, opened 2025-06-24
ci: Declare timeout for GitHub Actions job execution

staabm:
the default of 6h is way to long and can produce a overlong queue in case of endless loop bugs

---

looked into it, because there is [currently a E2E test endless running](https://github.com/infection/infection/actions/runs/15845415150/job/44666330784) which does not end and burns unnecessary resources


maks-rafalko (approved):
👍

maks-rafalko:
thank you @staabm

PR #2248, opened 2025-06-25
Remove IdenticalEqual, EqualIdentical

### LINKED ISSUE #2111: Smarter `IdenticalEqual` mutator

sanmai:
Fixes #2111


src/Mutator/Boolean/EqualIdentical.php
```diff
@@ -1,89 +0,0 @@
...
-
-    public static function getDefinition(): Definition
-    {
-        return new Definition(
-            <<<'TXT'
-                Replaces a loose comparison (using the equal operator (`==`)) with a strict comparison (using the
-                identical operator (`===`)).
-                TXT
-            ,
-            MutatorCategory::SEMANTIC_REDUCTION,
```
theofidry: I know the other one is disabled because it is a semantic addition which is non-sense to have, but what was it about this one?

sanmai: That's a good question. For some reason we had it disabled too. What do you think is the best path?

I guess we could enable it back.

theofidry: Ah I think it was simply redundant with the not equal ones in the end

staabm:
> This feels like a breaking change to me

agree, I also feel its a breaking change.

[on github.com I found a single reference](https://github.com/search?q=%27IdenticalEqual%27++NOT+is%3Afork&type=code) (outside of infection/* repos)

maks-rafalko:
> agree, I also feel its a breaking change.

but we are on 0.x, and it was disabled by default - not a big deal.

> [on github.com I found a single reference](https://github.com/search?q=%27IdenticalEqual%27++NOT+is%3Afork&type=code) (outside of infection/* repos)

https://github.com/kubawerlos/php-cs-fixer-custom-fixers/blob/0cf921cf9aee50ac3091839325f5cf38ab189b97/.dev-tools/src/InfectionConfigBuilder.php#L24

it's disabled and located in a const named `UNWANTED_MUTATORS` :) so it just confirms we shouldn't care about it IMO

Ocramius:
Just ran into this while doing upgrades: I used this mutator, as it is important to check for identity vs equality in some domains (example: ORMs, serializers).

Should not have been dropped, tbh.

theofidry:
@Ocramius you can now create a custom one where you add it back.

I admit I don't understand what it would allow you to check though. Note that it is `=== -> ==`, not `== -> ===`

sanmai:
I would really love to understand the use-case :pray:

Ocramius:
In an identity-map-backed repository, the test **must** verify that object-hash-level identity is used.

For example, written naively:

```php
public function remove(object $toRemove) {
    foreach ($this->identityMap as $key => $item) {
        if ($toRemove === $item) {
            unset($this->identityMap[$key]);
        }
    }
}
```

This is a case where `===` -> `==` can lead to a mutation that hides equality-based object manipulation.

In such a test, the author would need to add 2 equal, non-same objects to the `identityMap`, then remove one: the mutation highlights how such an identity could've been skipped.

theofidry:
@Ocramius this should still be tested I think?

To step back for a moment: what we want here is to prove that this piece of code is bug free.

Now let's say a developer accidentally changes this strict equality for a loose equality and the tests are green:


```php
public function remove(object $toRemove) {
    foreach ($this->identityMap as $key => $item) {
        if ($toRemove == $item) {
            unset($this->identityMap[$key]);
        }
    }
}
```

Then you will have the equal to identity mutation that will kick in:


```diff
public function remove(object $toRemove) {
    foreach ($this->identityMap as $key => $item) {
-        if ($toRemove == $item) {
+        if ($toRemove === $item) {
            unset($this->identityMap[$key]);
        }
    }
}
```

Then you have two scenarios:

1) **tests fails**: would be that the claimed contract is untrue. So either the documentation of the contract is incorrect, or there is no test testing that contract (this can happen when you generate tests & harden them based on the current code, without any verification that it does what you really wanted in the first place).
2) **tests passes**: The mutation is uncovered, and in this case shows the bug you wanted to guard against.

Ocramius:
@theofidry didn't you just say this is `=== -> ==`? :thinking:

PR #2249, opened 2025-06-25
Drop use of inline-`@var`. Use Webmozart\Assert\Assert instead

staabm:
phpdocs might be lying and and cannot be verified by toolng

inline-`@var` should be prevented whenever possible


src/Mutator/FunctionSignature/ProtectedVisibility.php
```diff
@@ -94,8 +95,8 @@ public function canMutate(Node $node): bool
             return false;
         }
 
-        /** @var ClassReflection $class */
```
staabm, line 97: phpdoc was missing `null`.

src/Mutator/ReturnValue/ArrayOneItem.php
```diff
@@ -126,8 +128,8 @@ public function canMutate(Node $node): bool
 
     private function returnTypeIsArray(Node $node): bool
     {
-        /** @var Node\Stmt\Function_|null $functionScope */
```
staabm, line 129: this type did not work for ClassMethod

src/Mutator/Util/AbstractValueToNullReturnValue.php
```diff
@@ -53,8 +54,8 @@ abstract class AbstractValueToNullReturnValue implements Mutator
 
     protected function isNullReturnValueAllowed(Node $node): bool
     {
-        /** @var Node\Stmt\Function_|null $functionScope */
```
staabm, line 56: this type did not work for ClassMethod

tests/phpunit/Mutator/Util/NameResolverTest.php
```diff
@@ -50,7 +50,7 @@ public function test_it_resolves_self_name_to_fully_qualified_name(): void
         $name = new Node\Name('self');
         $name->setAttribute(
             ReflectionVisitor::REFLECTION_CLASS_KEY,
-            new ReflectionClass(self::class),
```
staabm: the test was using php-src `ReflectionClass` which was a bug
(revealed by the assertion)

maks-rafalko (approved):
👍

src/Mutator/FunctionSignature/ProtectedVisibility.php
```diff
@@ -114,8 +115,8 @@ public function canMutate(Node $node): bool
 
     private function hasSameProtectedParentMethod(Node\Stmt\ClassMethod $node): bool
     {
-        /** @var ClassReflection $reflection */
         $reflection = $node->getAttribute(ReflectionVisitor::REFLECTION_CLASS_KEY);
+        Assert::isInstanceOf($reflection, ClassReflection::class);
```
theofidry: I think it would make more sense for `ReflectionVisitor` to provide a utility instead:

```suggestion
        $reflection = ReflectionVisitor::getReflectionClass($node);
```

```php
function getReflectionClass(Node $node): ClassReflection {
  $reflection = $node->getAttribute(self::REFLECTION_CLASS_KEY);
  Assert::isInstanceOf($reflection, ClassReflection::class);

  return $reflection;
}
```

I see that sometimes we want a nullable variant, to which we could have `::tryToGetReflectionClass()`

theofidry (commented):
I think this MR could be split in two: one is an actual fix (using the proper reflection class) and the types used downstream.

The other is switching from `@var` to assertions.

The title otherwise does not capture the diff, and one could work it out from the diff but that requires a deeper review already.

sanmai:
These assert calls are not without the cost. How did you verify they are out of any hot paths?

staabm:
> These assert calls are not without the cost. How did you verify they are out of any hot paths?

I did not do any perf tests, but `isInstanceOf` is a method call which is doing 1 `instanceof` call.
I don't think this calls can slow-down a hot path

src/Mutator/FunctionSignature/ProtectedVisibility.php
staabm: added `ReflectionVisitor::getReflectionClass` - thx

src/Mutator/FunctionSignature/PublicVisibility.php
```diff
@@ -106,8 +106,8 @@ public function canMutate(Node $node): bool
 
     private function hasSamePublicParentMethod(Node\Stmt\ClassMethod $node): bool
     {
-        /** @var ClassReflection $reflection */
-        $reflection = $node->getAttribute(ReflectionVisitor::REFLECTION_CLASS_KEY);
+        $reflection = ReflectionVisitor::getReflectionClass($node);
+        Assert::notNull($reflection);
```
theofidry, line 110: why not go with the two vairants `::tryToGetReflection()` `::getReflection()` to avoid this repetition of the assert?

staabm: from my POV having too many too similar api methods is not useful.

I can adjust in case you have a different taste

maks-rafalko:
thank you @staabm

PR #2250, opened 2025-06-25
Fix wrong ReflectionClass in NameResolverTest

staabm:
refs https://github.com/infection/infection/pull/2249#pullrequestreview-2957764850

the test was using php-src `ReflectionClass` while the implementation expects `Infection\Reflection\CoreClassReflection`


staabm:
I guess the build error is not related to the fix and re-starting CI will fix it

theofidry (approved):

maks-rafalko (approved):

PR #2258, opened 2025-06-27
Use fully-qualified-constant-reference instead of a dynamic lookup

staabm:


src/Process/Runner/InitialTestsRunner.php
```diff
@@ -71,7 +71,7 @@ public function run(
         $this->eventDispatcher->dispatch(new InitialTestSuiteWasStarted());
 
         $process->run(function (string $type) use ($process): void {
-            if ($process::ERR === $type) {
```
staabm, line 74: no need to make this unnecessary dynamic when accessing a global constant

maks-rafalko (approved):

maks-rafalko:
thank you @staabm

PR #2265, opened 2025-06-28
cleanup check_trailing_whitespaces.sh

staabm:
- use `git grep` instead of `find`, so we don't need to list all already `.gitignore`'d files again
- 10x faster. before this PR it took 1,8-2seconds, now it takes 0,04-0,2 seconds

inspired by the [latest version of this file on the php-cs-fixer repo](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/a32defda77a0d406b35a662cef1ada63a5947c3b/dev-tools/check_trailing_spaces.sh)

### DIFF
--- a/devTools/check_trailing_whitespaces.sh
+++ b/devTools/check_trailing_whitespaces.sh
@@ -21,42 +21,27 @@
 # DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-set -Eeuo pipefail
+set -eu
 
-files_with_trailing_whitespaces=$(
-    find . \
-        -type f \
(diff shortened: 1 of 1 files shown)


devTools/check_trailing_whitespaces.sh
```diff
@@ -21,42 +21,27 @@
...
-        -not -path "./vendor/*" \
-        -not -path "./tests/autoloaded/*" \
-        -not -path "./tests/benchmark/*/coverage/*" \
-        -not -path "./tests/benchmark/*/sources/*" \
-        -not -path "./tests/e2e/*" \
-        -not -path "./tests/phpunit/Fixtures/Files/phpunit/format-whitespace/original-phpunit.xml" \
-        -not -path "./tests/phpunit/StringNormalizerTest.php" \
-        -not -path "./tests/phpunit/StrTest.php" \
-        -not -path "./tests/benchmark/Tracing/coverage.tar.gz" \
-        -not -path "./tests/benchmark/MutationGenerator/._sources" \
```
staabm, line 48: this file did not exist. changed it to `sources.tar.gz`

PR #2266, opened 2025-06-28
Improve LogicalOrTest / drop mutator ignores

staabm:
after mutators got smarter in recent PRs, I think we should remove the config based ignores and instead accept the remaining escaped mutants.

the config ignores which get deleted made mutations invisible in https://github.com/infection/infection/pull/2243 which would have been useful to have.


maks-rafalko:
thank you @staabm

PR #2267, opened 2025-06-29
Use/Suggest `executionOrder="defects,random"` for phpunit.xml

staabm:
when re-running a test-suite which failed before it makes a lot of sense to run the previously failed tests first.
that way we validate whether the already known problems are gone and in addition we also shorten the feedback loop.

on my machine re-running a previously failed test-suite, in the scenario when it will fail again after this PR

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -3,7 +3,7 @@
          xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/phpunit.xsd"
          bootstrap="./vendor/autoload.php"
          colors="true"
-         executionOrder="random"
+         executionOrder="defects,random"
          failOnWarning="true"
          failOnRisky="true"
          requireCoverageMetadata="true"
--- a/src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
+++ b/src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
@@ -160,7 +160,7 @@ class PhpUnitAdapter extends AbstractTestFrameworkAdapter implements MemoryUsage
             $recommendations = sprintf(
                 "%s\n\n%s\n\n%s",
                 "Infection runs the test suite in a RANDOM order. Make sure your tests do not have hidden dependencies.\n\n"
-                . 'You can add these attributes to `phpunit.xml` to check it: <phpunit executionOrder="random" resolveDependencies="true" ...',
+                . 'You can add these attributes to `phpunit.xml` to check it: <phpunit executionOrder="defects,random" resolveDependencies="true" ...',
                 'If you don\'t want to let Infection run tests in a random order, set the `executionOrder` to some value, for example <phpunit executionOrder="default"',
                 parent::getInitialTestsFailRecommendations($commandLine),
             );
--- a/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
+++ b/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
@@ -143,7 +143,7 @@ class InitialConfigBuilder implements ConfigBuilder
             return;
         }
 
-        if ($this->addAttributeIfNotSet('executionOrder', 'random', $xPath)) {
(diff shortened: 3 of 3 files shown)


phpunit.xml.dist
```diff
@@ -3,7 +3,7 @@
          xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/phpunit.xsd"
          bootstrap="./vendor/autoload.php"
          colors="true"
-         executionOrder="random"
+         executionOrder="defects,random"
```
staabm, line 6: this feature works in PHPUnit 10.5+, but only recently is allowed in phpunit.xml schema via https://github.com/sebastianbergmann/phpunit/pull/6258 which will be released with 10.5.48+

so while the feature works with older PHPUnit 10.5.x, tools which validate the xml - e.g. PHPStorm - would report that `defects,random` is not valid per schema.

maks-rafalko: if I understand it correctly, Infection's config validation will not be broken as we validate a bit earlier, right?

src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
```diff
@@ -143,7 +143,7 @@ private function addRandomTestsOrderAttributesIfNotSet(string $version, SafeDOMX
             return;
         }
 
-        if ($this->addAttributeIfNotSet('executionOrder', 'random', $xPath)) {
+        if ($this->addAttributeIfNotSet('executionOrder', 'defects,random', $xPath)) {
```
maks-rafalko: if I understand it correctly, Infection's config validation for project's config will not be broken as we validate a bit earlier, right?

theofidry:
Since we check the version of PHPUnit, maybe we can recommend and enable this feature only if the required PHPUnit version is met?

src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
staabm: e2e tests fail with 

```
 There was 1 PHPUnit test runner warning:

1) Test results may not be as expected because the XML configuration file did not pass validation:

  Line 14:
  - Element 'phpunit', attribute 'executionOrder': [facet 'enumeration'] The value 'defects,random' is not an element of the set {'default', 'defects', 'depends', 'depends,defects', 'depends,duration', 'depends,random', 'depends,reverse', 'depends,size', 'duration', 'no-depends', 'no-depends,defects', 'no-depends,duration', 'no-depends,random', 'no-depends,reverse', 'no-depends,size', 'random', 'reverse', 'size'}.
```

therefore I think this might not be true ;)

staabm: lets see how it goes when the next PHPUnit release was published

staabm:
I have adjusted the PR to use a min-phpunit version check for the `executionOrder="defects,random"` case.

is this change fine without new tests?

src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
```diff
@@ -156,7 +156,15 @@ public function getInitialTestsFailRecommendations(string $commandLine): string
     {
         $recommendations = parent::getInitialTestsFailRecommendations($commandLine);
 
-        if (version_compare($this->getVersion(), '7.2', '>=')) {
+        if (version_compare($this->getVersion(), '10.5.48', '>=')) {
```
staabm: In case we want to be super-safe, we would need to check for phpunit 11.5.27+ and 12.2.7+ separately (because there are PHPUnit 11.x and 12.x versions > 10.5.48 which do not support this feature)

maks-rafalko: I would do that, as the current implementation can confuse/mislead users. What do you think?

staabm: adjusted and added tests

maks-rafalko (approved):
thank you @staabm

Issue #2268, opened 2025-06-29
Return removal mutator


sanmai:
**Is your feature request related to a problem? Please describe.**

I'm looking at many instances of this memoization pattern:

```php
private ?object $config = null;

private function getConfig(): object
{
    if (null !== $this->config) {
        return $this->config;
    }
    // ... load and cache
    $this->config = $config;
    return $this->config;
}
```
And there's no way to see this memoization is not tested [by looking at mutation testing results only](https://infection-php.dev/r/m6nn).

**Describe the solution you'd like**

I'd love to see a mutation that would remove one of the return statements if there are several.

```diff
    if (null !== $this->config) {
-        return $this->config;
    }
    // ...
    return $this->config;
```

Alternatively, we can remove the whole if-return branch:

```diff
-   if (null !== $this->config) {
-        return $this->config;
-   }
    // ...
    return $this->config;
```

**Additional context**

Demo: https://github.com/sanmai/infected-memoization

The missing test should install a [spy on `loadConfig()`](https://github.com/sanmai/infected-memoization/blob/0062ee90a90613aaf01c6a3e2ed21cc0d9aeea5c/src/SourceClass.php#L11) to confirm it only called once.

theofidry:
It's the typical instance of code described as "arid code" in the Google papers.

I think this with the logging statements will be the main targets when we get to it.

PR #2269, opened 2025-06-30
Add a new E2E test to show "arid code" aspect of memoization

sanmai:
This PR:

- [ ] Add a demo for https://github.com/infection/infection/issues/2268


tests/e2e/InfectedMemoization/phpunit.xml
```diff
@@ -0,0 +1,18 @@
...
+         bootstrap="vendor/autoload.php"
+         colors="true"
+>
+    <testsuites>
+        <testsuite name="default">
+            <directory>tests</directory>
+        </testsuite>
+    </testsuites>
+
+    <source ignoreIndirectDeprecations="true" restrictNotices="true" restrictWarnings="true">
```
sanmai:
```suggestion
    <source>
```

theofidry:
Note that in general I don't think this sort of scenario needs tests. What you are doing is strictly performance related, and unless you have tools to track each PHP function/methods calls, you are bound to hit a wall.

sanmai:
We can close it later without merging once the mutation is here.

Issue #2123, opened 2025-06-05
Enable MSI for `--git-diff-lines` on PRs, probably for 100%


maks-rafalko:
Thoughts:

- we all know that MT is by design slow. And I know that many developers (especially who didn't try it, or used it incorrectly) think that MT can't be used on CI because it's slow - that's only partially true
- I always suggest people to use `--git-diff-lines` or similar instead of running MT for the whole codebase on PRs. But it's funny - we don't use `--git-diff-lines` ourselves here at Infection
  - well, we [do use it](https://github.com/infection/infection/blob/0d238ba89379da151b589ea3b8107036b62fbb42/.github/workflows/mt-annotations.yaml#L59), but only to generate GH annotations, not for checking MSI 
  - at the same time we run MT for the whole codebase each time PR is created, and in this run we check MSI

Downsides of running MT for the whole codebase:

- it's slow. On GH it now takes ~8m (`Time: 8m 7s. Memory: 0.24GB. Threads: 3`) and generates ~5k mutants
- we can't set MSI=100%, because we don't have a baseline feature and we started with far less than 100%

But, I'm not saying here that we need to disable this GH action step. I'm saying that we need to 

- remove MSI  from that GH action which runs MT for the whole codebase
- and use MSI requirement inside `--git-diff-lines` GH action, [here](https://github.com/infection/infection/blob/0d238ba89379da151b589ea3b8107036b62fbb42/.github/workflows/mt-annotations.yaml#L59)

Benefits of using MSI together with `--git-diff-lines`:

- we can set it to 100% for a diff inside PR, while MSI for the whole project will be (for now) less. Basically, all new code/tests will be stricter. I know it can be very difficult, time-consuming and annoying to get 100% MSI for a PR, but if needed, we can think about some kind of "pass through" hack if we need to merge with MSI < 100%
- we can enable PHPStan mutant killer (`--static-analysis-tool=phpstan`) if we run MT for a diff! it will help to kill more mutants. On the other side, I doubt that we can enable it when we run MT for the whole codebase, because it's very slow (for now)

Why MSI=100% is better for us, maintainers of Infection:

- besides obvious thing that with MSI=100% we will have increased quality of tests, it gives us a cool benefit - you either kill a mutant or you understand that this is impossible. And when it's impossible, usually it means the following:
  - we generate equivalent mutant (so there is a bug in Infection) and we will find it and fix it (so there will be higher chance for us to find equivalent mutants)
[...]

PR #2270, opened 2025-07-01
Check MSI=100 per pull request changed files only

staabm:
closes https://github.com/infection/infection/issues/2123

background and idea described in the above ticket

### DIFF
--- a/.github/workflows/mt-annotations.yaml
+++ b/.github/workflows/mt-annotations.yaml
@@ -64,7 +64,6 @@ jobs:
                     --ignore-msi-with-no-mutations \
                     --min-msi=100 \
                     --min-covered-msi=100 \
-                    --static-analysis-tool=phpstan \
                     --only-covered
 
     # This is a meta job to avoid to have to constantly change the protection rules
--- a/.github/workflows/mt-annotations.yaml
+++ b/.github/workflows/mt-annotations.yaml
@@ -54,10 +54,18 @@ jobs:
               run: |
                   composer install --no-interaction --prefer-dist --no-progress
 
-            - name: Run Infection for added files only
+            - name: Run Infection for changed lines only
               run: |
                   git fetch origin $GITHUB_BASE_REF
-                  php bin/infection --threads=max --git-diff-lines --git-diff-base=origin/$GITHUB_BASE_REF --ignore-msi-with-no-mutations --only-covered
+                  php bin/infection \
+                    --threads=max \
+                    --git-diff-lines \
+                    --git-diff-base=origin/$GITHUB_BASE_REF \
+                    --ignore-msi-with-no-mutations \
+                    --min-msi=100 \
+                    --min-covered-msi=100 \
+                    --static-analysis-tool=phpstan \
+                    --only-covered
 
     # This is a meta job to avoid to have to constantly change the protection rules
     # whenever we touch the matrix.
--- a/.github/workflows/mt.yaml
+++ b/.github/workflows/mt.yaml
@@ -13,7 +13,6 @@
 name: Mutation Testing
 
 on:
-  pull_request:
   push:
     branches:
       - main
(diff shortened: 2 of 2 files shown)


.github/workflows/mt-annotations.yaml
```diff
@@ -54,10 +54,18 @@ jobs:
...
                   git fetch origin $GITHUB_BASE_REF
-                  php bin/infection --threads=max --git-diff-lines --git-diff-base=origin/$GITHUB_BASE_REF --ignore-msi-with-no-mutations --only-covered
+                  php bin/infection \
+                    --threads=max \
+                    --git-diff-lines \
+                    --git-diff-base=origin/$GITHUB_BASE_REF \
+                    --ignore-msi-with-no-mutations \
+                    --min-msi=100 \
+                    --min-covered-msi=100 \
+                    --static-analysis-tool=phpstan \
```
sanmai:
```suggestion
```
Can we discuss this separately? I think this static analysis thing goes against the purpose of mutation testing, and if someone wants to use it, it should not be Infection. 

For example, any mutators from #2268 make no sense from the point of view of a static analysis (as `if($foo) {}` with an empty branch is clearly an error in the actual code), but from the MT position they are perfectly valid way to validate that tests test what they should.

maks-rafalko: I'm ok with extracting it to a separate discussion

maks-rafalko (approved):
While I am created an initial request for it, I suspect that this will be extremely hard to work with, especially for newcomers (but for an experienced user as well).

We can try and see how it works, even a bad result is a result :) Anyway, this can be reverted later with no doubts if it will be too difficult for now.

sanmai (changes requested):
`--static-analysis-tool=phpstan` should not be here

.github/workflows/mt-annotations.yaml
staabm: I am fine with dropping the SA part. I added it because it was mentioned in the referenced issue.

staabm:
> While I am created an initial request for it, I suspect that this will be extremely hard to work with, especially for newcomers (but for an experienced user as well).
> 
> We can try and see how it works, even a bad result is a result :) Anyway, this can be reverted later with no doubts if it will be too difficult.

agree. even if we enforce only 90 or 95-MSI per pull request it will lead to a increase of codebase overall MSI over time

.github/workflows/mt.yaml
```diff
@@ -13,7 +13,6 @@
 name: Mutation Testing
 
 on:
-  pull_request:
```
staabm, line 16: since we no longer run this job per pull request, the mandatory check for "Mutation Testing Status" needs to be disabled in the repo settings.

this means we can also drop the "meta-job" in this file starting from line 90, if you want.

maks-rafalko:
Good luck everyone with this merge! 🤣 

another idea if it will be difficult to get 100% MSI in a particular PR - we can bypass this and pass a build if commit has some postfix/message, I remember PHP-CS-Fixer has something like that

They self-approve PR based on commit message:

https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/35aa86fa1107f7bf34611a2575c5eb589ceb2a72/.github/workflows/self-approval.yml

.github/workflows/mt.yaml
maks-rafalko: hm, I didn't notice it, thanks for highlighting

do we agree to not run the full MT on PRs? I thought @sanmai still wanted to do it

https://github.com/infection/infection/issues/2123#issuecomment-2947350368

staabm: we still run it after merge on the target branch.

thats what I understood from https://github.com/infection/infection/issues/2123#issuecomment-2987146527 and following

sanmai: If we still run it on the main/master branch, that's fine by me :+1:

sanmai:
Right, so we need to disable that requirement.

maks-rafalko:
Ok, I disabled "Mutation Testing Status" requirements, seems like now everything is good

maks-rafalko:
Thank you @staabm

PR #2273, opened 2025-07-02
Fix note after run with --show-mutations=0

staabm:
since https://github.com/infection/infection/pull/2216 ~~the note did not make sense without a additional value~~ promoting the more useful way to pass a option-value makes more sense

`bin/infection --show-mutations` is equivalent to `bin/infection --show-mutations=20`

### DIFF
--- a/tests/phpunit/Event/Subscriber/MutationTestingConsoleLoggerSubscriberTest.php
+++ b/tests/phpunit/Event/Subscriber/MutationTestingConsoleLoggerSubscriberTest.php
@@ -362,7 +362,7 @@ final class MutationTestingConsoleLoggerSubscriberTest extends TestCase
         $dispatcher->dispatch(new MutationTestingWasFinished());
 
         $this->assertStringNotContainsString(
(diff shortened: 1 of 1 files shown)


src/Event/Subscriber/MutationTestingConsoleLoggerSubscriber.php
```diff
@@ -236,7 +236,7 @@ private function showGeneratedLogFiles(): void
 
         // for the case when no file loggers are configured and `--show-mutations` is not used
         if ($this->numberOfShownMutations === 0) {
-            $this->output->writeln(['', 'Note: to see escaped mutants run Infection with "--show-mutations" or configure file loggers.']);
+            $this->output->writeln(['', 'Note: to see escaped mutants run Infection with "--show-mutations=20" or configure file loggers.']);
```
staabm, line 239: to fix the `ArrayItemRemoval` mutation of `''`, we now use a stricter expectation - as [we need 100% MSI per pull request](https://github.com/infection/infection/pull/2270) ;)

PR #2274, opened 2025-07-02
Cleanup MutationTestingConsoleLoggerSubscriberTest

staabm:
- MSI of `MutationTestingConsoleLoggerSubscriber` is now 100
- unify handling of how we assert output across tests
- drop hard to read `WithConsecutive` usages and replace them with `assertStringContainsString` assertions


maks-rafalko (approved):
thank you @staabm

Issue #2218, opened 2025-06-16
DX: git-diff-lines references branch


staabm:
per docs at https://infection.github.io/guide/command-line-options.html#git-diff-lines

we assume the default reference branch to be `master`.
since someone told the industry `master` is a bad word, more and more repositoires use `main` as a default.

it would be great if one could use `--git-diff-lines` without a need to pass additional options or do configuration, when the reference branch is `main` or `master`. both are pretty common and it would be great we would support both per default, without manual fiddling

----

the name of the default branch can be determined using:
```
git remote show origin | sed -n '/HEAD branch/s/.*: //p'
```

[per stackoverflow](https://superuser.com/a/1834800)

PR #2289, opened 2025-07-04
DX: Automatically determine git reference branch

staabm:
closes https://github.com/infection/infection/issues/2218

### DIFF
--- a/src/Configuration/ConfigurationFactory.php
+++ b/src/Configuration/ConfigurationFactory.php
@@ -354,7 +354,7 @@ class ConfigurationFactory
             return $filter;
         }
 
-        $baseBranch = $gitDiffBase ?? GitDiffFileProvider::DEFAULT_BASE;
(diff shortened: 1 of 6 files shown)


staabm:
regarding escaped mutants:

> Escaped Mutant for Mutator "Coalesce": in ConfigurationFactory

-> I did not really change the logic. I am not convinced this mutations needs to be killed with this PR

> Ecaped Mutant for Mutator "ArrayItemRemoval in GitDiffFileProvider

-> to catch the mutant I would need to make assertions on the `RuntimeException` error message.
-> semantically its pretty hard to do assertions on `RuntimeException` as they can contain so many different things
(maybe someone else has a idea how this can be tested in a useful way)

maks-rafalko (approved):
👍

maks-rafalko:
discussed to bypass msi=100 here, merging

PR #2292, opened 2025-07-04
Use colored diffs in GitHub Actions annotations

staabm:
Don't know whether thats supported at all.. just trying..


staabm:
seems not supported by github actions

Issue #2247, opened 2025-06-24
Remove `ConcatOperandRemoval`?


maks-rafalko:
This is a continuation of our goal to reduce redundant / unproductive mutators, see

- https://github.com/infection/infection/issues/2184

I feel like the tests that kill Mutants created by `Concat` mutators must kill mutants generated by `ConcatOperandRemoval` mutator, for example

Consider this code

```php
$result = 'a' . 'b' . 'c';
```

and this bad test for it:

```php
assertContains($result, 'a');
assertContains($result, 'c');
```

Both Mutants created by `Concat` are escaped:

```diff
- $result = 'a' . 'b' . 'c';
+ $result = 'b' . 'a' . 'c';
```

and

```diff
- $result = 'a' . 'b' . 'c';
+ $result = 'a' . 'c' . 'b';
```

Also, we have 1 escaped Mutant for `ConcatOperandRemoval`:

```diff
- $result = 'a' . 'b' . 'c';
+ $result = 'a' . 'c';
```

Now: how to kill `Concat` Mutants? We not only need all the concat parts to be present, but also need to check that the order is correct. So it's easy to kill it by just comparing with the whole result string:

```diff
- assertContains($result, 'a');
- assertContains($result, 'c');
+ assertContains($result, 'abc');
```

and now we have 0 escaped Mutants (so we killed both `Concat` and `ConcatOperandRemoval`).

If we would try to kill only `ConcatOperandRemoval`, we could write these tests (still bad, but for the sake of an example):

```diff
assertContains($result, 'a');
+ assertContains($result, 'b');
assertContains($result, 'c');
```

With these tests, `ConcatOperandRemoval` Mutants are killed, while `Concat` are still not - the order is not checked.

----

What I want to say is it seems like `ConcatOperandRemoval` mutator is not needed - we can already improve our tests by `Concat`.

Speaking about numbers: these are 2 of the most noisy mutators that generate hundreds of Mutants for Infection itself:

| Mutator                         | Mutations | Killed by Test Framework | Test Timings min/avg/max | Killed by Static Analysis | Static Analysis Timings min/avg/max | Escaped | Errors | Syntax Errors | Timed Out | Skipped | Ignored | MSI (%s) | Covered MSI (%s) |
| ------------------------------- | --------- | ------------------------ | ------------------------ | ------------------------- | ----------------------------------- | ------- | ------ | ------------- | --------- | ------- | ------- | -------- | ---------------- |
[...]

Issue #2221, opened 2025-06-16
timed-out mutants limit


staabm:
I think we need a way to set a maximum number of timed-out mutants on a per project basis.
when the limit is reached a non-zero exit code should be forced.

that way it would be possible to track the timed out mutations, as these can easily be accidentally introduced when working on features.

sanmai:
I'd really want a flag such as `--with-timeouts` that would make these mutations count as escaped and logged with `-s` so I can squash them all 😍

staabm:
agree, `--with-timeouts` would be awesome.

we could add a hint for this new option in case the summary contains at least 1 timed-out mutant

maks-rafalko:
could you please explain how `--with-timeouts` would work exactly?

1. is it a proposal **instead of** initial feature request about timeouts limit from this ticket?
2. does it mean that with `--with-timeouts` we would treat timeouts not as killed, but as escaped mutants? If so, it looks like it doesn't prevent developers from adding as many timeouts as possible in oppose to the initial request in this ticket

What I like with the initial idea is that we can say e.g. `maxTimeouts: 10` and with 11-th timeout the build will fails (btw, we already had the same feature request here https://github.com/infection/infection/issues/1133#issuecomment-1962107410)

Also, see

- #1133

sanmai:
Yup, it sounds like we have a duplicate. The issue #1133 proposes the very same idea among others.

With our 100% MSI for PRs I'd prefer that we have no timeouts, so `maxTimeouts: 0`.

maks-rafalko:
> With our 100% MSI for PRs I'd prefer that we have no timeouts, so `maxTimeouts: 0`.

I just realized that we must set `maxTimeouts: 0` for PR with `--git-diff-lines` - so that no timeouts are introduced, but at the same time we can **not** have `maxTimeouts: 0` for the whole MT run (on `master`) because we have ~10 timeouts.

So we would need to use `--max-timeouts=0` option for PRs (not `maxTimeouts: 0` from a config)

sanmai:
`--with-timeouts` should make all timeouts escaped mutations

PR #2296, opened 2025-07-05
Add `ReturnRemoval` mutator

### LINKED ISSUE #2268: Return removal mutator

sanmai:
This PR:

- [x] Adds new `ReturnRemoval` mutator to fix #2268
- [x] Implements `NextConnectingVisitor` to track statement execution order
- [x] Covered by tests
- [x] Doc PR: https://github.com/infection/site/pull/277

### `NextConnectingVisitor`

Added a new AST visitor that connects sequential statement nodes by adding a "next" attribute. This visitor:

- Tracks execution order between statements only (ignores expressions)
- Skips comment nodes (`Nop`)
- Resets the connection chain at function boundaries
- Enables the `ReturnRemoval` mutator to detect if there are statements after a return

### `ReturnRemoval` Mutator

The mutator now intelligently removes return statements without causing language errors:
- Removes any return in functions without return types (void or no type declaration)
- Removes returns that have subsequent statements (detected via `NextConnectingVisitor`)
- Always preserves the last return in functions with return types, preventing type errors
- Does not mutate `return;` or `return null;` at the end of functions without return types, as these are semantically equivalent to implicit null returns

So we avoid both un-killable mutations and fatal errors.

## Benefits

This implementation successfully avoids language errors while still exposing untested code paths. By preserving the last return statement in typed functions, we ensure the code remains syntactically valid while still being able to detect:

- Unreachable code after return statements
- Early returns that may not be properly tested
- Redundant return statements in functions

The approach provides maximum mutation coverage without breaking PHP's type system. Just what we wanted :slightly_smiling_face: 

----

Per-mutator report:

| Mutator       | Mutations | Killed by Test Framework | Test Timings min/avg/max | Killed by Static Analysis | Static Analysis Timings min/avg/max | Escaped | Errors | Syntax Errors | Timed Out (Limit: 25 secs) | Skipped | Ignored | MSI (%s) | Covered MSI (%s) |
| ------------- | --------- | ------------------------ | ------------------------ | ------------------------- | ----------------------------------- | ------- | ------ | ------------- | -------------------------- | ------- | ------- | -------- | ---------------- |
| ReturnRemoval |       569 |                      407 |  0.30 / 0.49 / 5.57 secs |                         0 |             0.00 / 0.00 / 0.00 secs |      70 |      3 |             0 |                          1 |       0 |       0 |    72.23 |            85.45 |


tests/phpunit/Mutator/Removal/ReturnRemovalTest.php
```diff
@@ -0,0 +1,77 @@
...
+
+    public static function mutationsProvider(): iterable
+    {
+        yield 'It mutates a return statement' => [
+            <<<'PHP'
+                <?php
+
+                function foo(): void
+                {
+                    return 'bar';
```
sidz: IMO this is a bad example. This code leads to the Fatal Error.

sanmai: Well, it is a draft PR after all.

sidz: I know :) Just left a note in order to not to miss :)

sanmai:
```suggestion
                    return;
```

tests/phpunit/Mutator/Removal/ReturnRemovalTest.php
```diff
@@ -0,0 +1,77 @@
...
+     */
+    #[DataProvider('mutationsProvider')]
+    public function test_it_can_mutate(string $input, $expected = [], array $settings = []): void
+    {
+        $this->assertMutatesInput($input, $expected, $settings);
+    }
+
+    public static function mutationsProvider(): iterable
+    {
+        yield 'It mutates a return statement' => [
```
sanmai:
```suggestion
        yield 'It mutates a redundant and unnecessary return statement' => [
```

tests/phpunit/Mutator/Removal/ReturnRemovalTest.php
sanmai: Thanks, this one should be better now. I still want to add more tests.

staabm:
looking at `ReturnLastTaggingVisitor`, I think we can do it simpler and without a nested `NodeTraverser` by taking inspiration from [PHPParsers builtin NodeConnectingVisitor](https://github.com/nikic/PHP-Parser/blob/dc47765670927fa0af38f7f95ea5c894099c780a/lib/PhpParser/NodeVisitor/NodeConnectingVisitor.php).

we need to find/tag `Return_` nodes, which do not have a `'next'` node, right?

sanmai:
Sadly [NodeConnectingVisitor](https://github.com/nikic/PHP-Parser/blob/dc47765670927fa0af38f7f95ea5c894099c780a/lib/PhpParser/NodeVisitor/NodeConnectingVisitor.php) only sets the next attribute on nodes that share the same parent. I.e. these do not:

```php
if ($foo) {
    return $bar;
}
return $baz;
```
But it is a good starting point for our own connecting visitor.

staabm:
Yes, posted it for inspiration to implement a smaller/specific one for our use-case

src/Mutator/Removal/ReturnRemoval.php
```diff
@@ -0,0 +1,126 @@
...
+
+        // Complex types are specific return types
+        if ($returnType instanceof ComplexType) {
+            return true;
+        }
+
+        // A void return type is the same as no return type for this mutator.
+        if (
+            $returnType === null
+            || $returnType->toLowerString() === self::VOID
```
staabm: we could also skip `:never` return type

src/PhpParser/Visitor/NextConnectingVisitor.php
```diff
@@ -0,0 +1,73 @@
...
+            return;
+        }
+
+        // We do not account comment nodes for next connections
+        if ($node instanceof Node\Stmt\Nop) {
+            return;
+        }
+
+        $this->previous?->setAttribute(self::NEXT_ATTRIBUTE, $node);
+        $this->previous = $node;
```
staabm, line 86: I think we should set references between AstNodes very carefully.
at best this visitor would be very specific for the use-case and set a more specific attribute, like "isLastReturn" and if-possible only on `Return_` nodes.

references between ast nodes can be very costly, as they hinder the garbage collector.
see e.g. https://phpstan.org/blog/preprocessing-ast-for-custom-rules which describes the cost of a parent/next/previous attribute the `NodeConnectingVisitor` incurs.

sanmai: If PHPStan has to keep the whole codebase in memory to do its thing, then it makes total sense for it. For us, it is a single file at a time. So it should not be a problem, and we already use the parent connecting visitor.

src/Mutator/Removal/ReturnRemoval.php
sanmai: When are we supposed to see a return statement in a function with `:never` return type? I'm looking at:

> Fatal error: A never-returning function must not return

src/PhpParser/Visitor/ReflectionVisitor.php
```diff
@@ -232,7 +233,11 @@ private function getClassReflectionForNode(Node\Stmt\ClassLike $node): ClassRefl
         $fqn = FullyQualifiedClassNameManipulator::getFqcn($node);
 
         if ($fqn !== null) {
-            return CoreClassReflection::fromClassName($fqn->toString());
+            try {
+                return CoreClassReflection::fromClassName($fqn->toString());
+            } catch (ReflectionException) {
+                // Fallback to the workaround
```
sanmai, line 247: Two considerations here:

- Classes in tests are not in the autoloader, so we can't get a reflection for them.
- Our `ReflectionVisitor` does not traverse anything outside of classes.

So to use classes in test fixtures we have to do this.

src/PhpParser/Visitor/ReflectionVisitor.php
```diff
@@ -150,22 +151,30 @@ public function leaveNode(Node $node): ?Node
...
         return $functionScope;
     }
 
+    public static function getFunctionScope(Node $node): Node\FunctionLike
+    {
+        $functionScope = $node->getAttribute(self::FUNCTION_SCOPE_KEY);
+        Assert::isInstanceOf($functionScope, Node\FunctionLike::class);
+
+        return $functionScope;
+    }
```
sanmai, line 176: We use the same get/find convention elsewhere.

maks-rafalko (commented):
Started reviewing but I definitely need more time with this one. Will leave a list of items to do:

- [x] Generate per-mutator log (added to the description and here)
- [ ] analyse `infection.log` generated by `text` logger with `--log-verbocity=all --debug` to see if we have Unproductive Mutants
- [ ] read PHPStan article about connected nodes
- [ ] check `nikic/php-parser` releases and get familiar with weak references in parent connector visitor (suggested by @staabm)
- [ ] analyze if it overlaps with negating `if` (sub)conditions (mutator: `IfNegation` and similar)

`infection.md`:

| Mutator       | Mutations | Killed by Test Framework | Test Timings min/avg/max | Killed by Static Analysis | Static Analysis Timings min/avg/max | Escaped | Errors | Syntax Errors | Timed Out (Limit: 25 secs) | Skipped | Ignored | MSI (%s) | Covered MSI (%s) |
| ------------- | --------- | ------------------------ | ------------------------ | ------------------------- | ----------------------------------- | ------- | ------ | ------------- | -------------------------- | ------- | ------- | -------- | ---------------- |
| ReturnRemoval |       569 |                      407 |  0.30 / 0.49 / 5.57 secs |                         0 |             0.00 / 0.00 / 0.00 secs |      70 |      3 |             0 |                          1 |       0 |       0 |    72.23 |            85.45 |

@sanmai I see there is a timeout, could you please try to fix it so that we don't have increased number of timeouts? Cause they affects the execution time a lot.

----

Ran it locally with 

```shell
bin/infection --mutators=ReturnRemoval --log-verbosity=all --debug --threads=max
```

results:

```
UUUUUU.U.....UU.U....M.......UU.UUEUU.M.......U...   ( 50 / 569)
UU.UUUUU............M..M.MU...............MM..UU..   (100 / 569)
MM.MM...U..MMM.....MMMM.....M...M.................   (150 / 569)
....M................MU.U..U..................M...   (200 / 569)
......M..UU.U.U........U......U.U.................   (250 / 569)
..M.UM..U...U...U......M.M..UU..U.M..MU...........   (300 / 569)
.....U.................U..........................   (350 / 569)
.UUU..U..U.U.....U..UUUU..U..M......M...........M.   (400 / 569)
......M.UUUUUUU..M...UM.MM...M...MM.....U..U....M.   (450 / 569)
MU.UUU........U..M......MM........MM...MMMM......M   (500 / 569)
.......UU...UUUMMM.MM.M.M.MUU..UUU.....M.M.UMUU...   (550 / 569)
.M.M.M.E..E..MT..MM                                  (569 / 569)


569 mutations were generated:
     407 mutants were killed by Test Framework
       0 mutants were caught by Static Analysis
      88 mutants were not covered by tests
      70 covered mutants were not detected
       3 errors were encountered
       1 time outs were encountered

Metrics:
         Mutation Score Indicator (MSI): 72%
         Mutation Code Coverage: 84%
         Covered Code MSI: 85%
```

and `infection.log` text file is attached

[infection.log](https://github.com/user-attachments/files/21132156/infection.log)

sanmai:
Looking into it. 

And this sounds like a great opportunity to re-introduce this feature request: https://github.com/infection/infection/issues/2221#issuecomment-3050384797 (I guess it can be a separate issue?)

maks-rafalko:
> I guess it can be a separate issue?

yeah, sounds like it should be a separate one

maks-rafalko:
I've updated the to do list and included "analyze if it overlaps with negating if (sub)conditions (mutator: `IfNegation` and similar)" - probably AI can help here, I would like to check if there are logical duplicates and if so - get rid of them.

Why? Because this Mutator produces a lot of mutans (569 from ~6k), like 10% of the whole number of generated mutants, so we need to carefully exclude overlaps if there are any.

staabm:
Since the idea of this PR is covering memoization patterns, maybe we can take inspiration from https://github.com/phpstan/phpstan-src/pull/4084 which tries to detect such code in PHPStan - to remove unnecessary mutations

staabm:
Maybe also rename it to MemoizationReturnRemoval

sanmai:
Memorization is just one example. Currently it removes all return statements that can be tested, fully in the spirit of other mutations, while trying not to cause a language error outright.

sanmai:
> I've updated the to do list and included "analyze if it overlaps with negating if (sub)conditions (mutator: IfNegation and similar)" - probably AI can help here, I would like to check if there are logical duplicates and if so - get rid of them.

I'm not sure I understand this request. It takes milliseconds to process a single mutant. It could probably take days of my time to run this very extensive analysis. How can we explain that it is really worth it?

maks-rafalko:
> I'm not sure I understand this request. It takes milliseconds to process a single mutant. It could probably take days of my time to run this very extensive analysis. How can we explain that it is really worth it?

If I'm not mistaken: we have 569 Mutants from this mutator, avg time for 1 mutants is 0.49s. So it's `0.49 * 569=278`. So 278s to run all mutants from this mutator with 1 thread.

And if we have let's say 30% (1%? 40%? 90%? yet to be discovered) of overlapping mutants, we could save a lot of time by not generating them. That was the goal of all those "Smarter XXX" PRs that we did recently.

In other words - if we look into this mutator alone - no problem with such a high number of generated mutants. But if we look into this mutator in combination of all other mutators - we need to carefully exclude duplicates, which is - agree - not a trivial exercise. I will try to do it.

sanmai:
Gotcha. I guess we can try this some other day.

maks-rafalko:
not sure why it's closed though, a lot of work has been done here, I absolutely didn't want it to be closed. 

Just a couple of more steps to get it merged (and all of those steps are really for me as I created a to do for myself 🤣 ).

sanmai:
I don't have the time to consider any more changes to this mutator if these changes _remove_ mutations that can be killed by a good unit test, and looking at the above discussion — that's where we are heading. Correct me if I'm wrong, and if I misunderstood the consensus, we can certainly restart this work.

sanmai:
I guess my point is that this mutator should work well if someone chooses it to be their sole mutator. Therefore, this discussion whenever this mutator overlaps with other mutators does not make sense to me. Every mutator should be great on its end own.

staabm:
IMO if we scope this mutator to the memoization use-case we get 90% of its value and I feel this will reduce a bit part of the unproductive mutations.

in case you don't want to pursue this road, I can try when I have my the next free timeslot.

all "regular return statements" are already well covered by a lot of other mutators, therefore I agree with maks that most of the mutations the PR - as is - generates are not very valuable.
if there are more specific useful examples which are not yet covered by other mutators, we should look at concrete examples.

sanmai:
This memoization point makes no sense because a switch to `sanmai/later` will immediately eliminate all memorization mutations. So it can't be a single purpose of a mutator like this.

staabm:
> This memoization point makes no sense because a switch to `sanmai/later` will immediately eliminate all memorization mutations.

atm `sanmai/later` is consumed by 6 projects based on packagist.org statistics.
bugs in memoization code is something related to all php projects on the planet.

if you have another concrete example where this mutator can catch return related problems which are not yet reported by other mutators, please show them.

maks-rafalko:
just in case, let me sum up my thoughts:

I didn't say this mutator is useless or it produces unproductive mutants, I've created a todo (for me) to check it and I have no answer to this question yet. And nobody has yet, because nobody did an analysis. 

The great examples of what I wanted to check is these issues: 

- https://github.com/infection/infection/issues/2247
- https://github.com/infection/infection/issues/1752#issuecomment-1320848169
- and many cases from "Smarter X mutator" PRs

I've spent hours just to check if 2 mutators have overlapping cases between each other. Exactly the same I wanted to do here. 

That's why I want to very carefully look into all new mutators we want to add - to not slow down MT without a reason (where we have overlapping).

> Every mutator should be great on its end own.

If it would be the only one mutator - no problem. But we have a ~hundred of them, so we must consider overlapping, otherwise Infection will be extremely slow (unproductive). (I understand that this can be treated differently by many of us, but anyway - that's only my opinion).

sanmai:
I use singular mutations to validate tests after reviewing the code for weak points (i.e. to see if this particular kind of construct is really tested), and these "smarter" mutations, that sometimes work and sometimes don't, make Infection useless to me. I really upsets me.

With many proprietary projects it is simply impossible to use all mutators, but using just one, like this one, can be very valuable. And if it becomes "smarter" this value evaporates into thin air.

PR #2299, opened 2025-07-06
Make ArrayItem mutator less evil

### LINKED ISSUE #2293: Smarter ArrayItem mutator
the `ArrayItem` mutator is working on function calls/method calls which can have side-effects.
today I have seen a codebase which creates 3811 ArrayItem mutations.

in the past I have created a [small tool which can detect side effects on code](https://github.com/staabm/side-effects-detector) (which is used by PHPUnit). this tool contains a [long list of functions/methods which are known to be side-effect free](https://github.com/staabm/side-effects-detector/blob/bb1dce7bdb1a033e2846dd82f44db99252478c85/lib/functionMetadata.php) (initially copied from PHPStan).

I don't think we will need this huge list, but WDYT about a extract of e.g. the "top 50" used php-src functions which are side-effect free, which we can utilize as a "skip-list" in ArrayItem mutator.

TLDR; what about adding a short list of known to be side-effect free functions into ArrayItem to reduce unnecessary mutations.

e.g. no need to mutate `[substr($key, $x) => $value]` as `substr` has no side-effects

sanmai:
This PR:

- [ ] Fixes #2293 by making this mutator less obnoxious for no good reason

I guess I wanted to see if mutating out function calls make sense, and, well, today these mutations are largely a waste of time and better covered by unwrapping mutations.

### DIFF
--- a/src/Mutator/Boolean/ArrayItem.php
+++ b/src/Mutator/Boolean/ArrayItem.php
@@ -56,7 +56,7 @@ final class ArrayItem implements Mutator
             <<<'TXT'
                 Replaces a key-value pair (`[$key => $value]`) array declaration with a value array declaration
                 (`[$key > $value]`) where the key or the value are potentially impure (i.e. have a side-effect);
-                For example `[foo() => $b->bar]`.
+                For example `[$this->foo() => $b->bar]`.
                 TXT
             ,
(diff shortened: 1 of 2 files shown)


staabm (approved):

maks-rafalko (approved):

staabm:
needs conflict resolving

PR #2180, opened 2025-06-11
Run PHPUnit worker process with Xdebug protection against endless loops

staabm:
POC!

running `INFECTION_ALLOW_XDEBUG=1 php bin/infection --debug --skip-initial-tests --coverage=build/logs --filter=src/FileSystem/Locator/RootsFileLocator.php --threads=8` with this PR ends the endless running test and infection returns after **1 second**.

running it without `INFECTION_ALLOW_XDEBUG=1` takes 25 seconds (the process timeout kills the process).

using `INFECTION_ALLOW_XDEBUG=1` for that purpose is non-sense of course, buts its the small hack to demonstrate how we could utilize `-d xdebug.max_nesting_level=1000` to end endless running mutations earlier.

---

the mutation which triggers the problem is:

```diff
9) /Users/m.staab/Documents/GitHub/infection/src/FileSystem/Locator/RootsFileLocator.php:115    [M] FunctionCallRemoval

@@ @@
         try {
             return $this->locate(current($fileNames));
         } catch (FileNotFound) {
-            array_shift($fileNames);
+            
             return $this->innerLocateOneOf($fileNames);
         }
     }
 }
```

since we likely have xdebug in the system (e.g. for coverage information), we might also make use of it for this feature. wdyt?

Issue #2192, opened 2025-06-13
Smarter `FunctionCallRemoval` mutator


staabm:
running infection on its own codebase yields

```
9) /Users/m.staab/Documents/GitHub/infection/src/FileSystem/Locator/RootsFileLocator.php:115    [M] FunctionCallRemoval

@@ @@
         try {
             return $this->locate(current($fileNames));
         } catch (FileNotFound) {
-            array_shift($fileNames);
+            
             return $this->innerLocateOneOf($fileNames);
         }
     }
 }

10) /Users/m.staab/Documents/GitHub/infection/src/FileSystem/Locator/RootsFileOrDirectoryLocator.php:112    [M] FunctionCallRemoval

@@ @@
         try {
             return $this->locate(current($fileNames));
         } catch (FileOrDirectoryNotFound) {
-            array_shift($fileNames);
+            
             return $this->innerLocateOneOf($fileNames);
         }
     }
 }
```

as a timed out mutation, because it triggers a endless loop. see e.g [RootsFileLocator](https://github.com/infection/infection/blob/a562599f5c5629a2a3e469fd5545868be35bb6a9/src/FileSystem/Locator/RootsFileLocator.php#L106-L119)

since this is a timed out mutant it costs 25 seconds on a infection run every time.

I thought about making `FunctionCallRemoval` smarter so it
- detects whether a function/method contains recursive calls
- when recursive, don't mutate calls which have side-effects on variables which are passed to recursive calls

I am not 100% sure whether this might kill mutants which are useful, therefore open the issue for discussion first.

Alternatively: instead of trying to prevent the endless recursion by inspecting the ast and not emitting the mutation we could utilize [xdebug endless loop detection as shown in POC](https://github.com/infection/infection/pull/2180)

looking for feedback/opinions

sanmai:
This is not an invalid mutation that either does not make sense or that you can't possibly counter with a test. For this particular mutation, we only need to verify that `array_shift()` is getting called, or eliminate the call.

I think the attitude with mutations must be not "_how this mutation triggers bugs in my code_" but "_what Infection is trying to tell be about the robustness of my code/tests_" and "_what if my code is hot garbage because I did not sleep well that night I wrote it_" (often my situation).

sanmai:
I'm making #2193 to close this issue, unless there are more examples (let's limit these to, say, five), which I will be happy to counter with test/code updates, while I still have the time 🙂

staabm:
[...]

PR #2300, opened 2025-07-07
MutationProcess timeout is based on test-execution time

staabm:
`MutationTestRunner::takingTooLong` is about skipping mutations from which we forsee they will take longer then 30s timeout because we know on average on how long it takes to run the tests in the test-suite (because tests are slow).

thats a independent feature which I don't want to touch.

---

with this PR we use individual timeouts per mutation (instead of a 30s timeout per se) when we actually run the mutation process based on timings we took on initial test run. 

goal is: on "endless running" mutations we will not loose the whole 30 seconds but way less time, 
as we know that the test-suite is usually able to kill the mutation in X seconds and if we take more then e.g. `min('5-times X', 30)` seconds we can stop trying, as we are likely running a endless-loop.

so I try to use shorter timeouts to speedup the overall process.

when we know that tests for `classX::methodY` usually run in 0.0012 seconds, there is no point in waiting 30seconds (atm we wait ~30_000x longer then it usually takes for this method to run a test)

---

PR

```
5649 mutations were generated:
    4344 mutants were killed by Test Framework
       0 mutants were caught by Static Analysis
     165 mutants were configured to be ignored
     571 mutants were not covered by tests
     545 covered mutants were not detected
       8 errors were encountered
      16 time outs were encountered

Metrics:
         Mutation Score Indicator (MSI): 79%
         Mutation Code Coverage: 89%
         Covered Code MSI: 88%

Time: 2m 36s. Memory: 0.19GB. Threads: 8
```

---

Before PR:

```
5645 mutations were generated:
    4341 mutants were killed by Test Framework
       0 mutants were caught by Static Analysis
     165 mutants were configured to be ignored
     571 mutants were not covered by tests
     544 covered mutants were not detected
       8 errors were encountered
      16 time outs were encountered

Time: 3m 9s. Memory: 0.18GB. Threads: 8
```

closes https://github.com/infection/infection/pull/2180

relates to https://github.com/infection/infection/issues/2192

### DIFF
--- a/src/Process/Factory/MutantProcessContainerFactory.php
+++ b/src/Process/Factory/MutantProcessContainerFactory.php
@@ -40,7 +40,6 @@ use Infection\Mutant\Mutant;
 use Infection\Mutant\TestFrameworkMutantExecutionResultFactory;
 use Infection\Process\MutantProcess;
 use Infection\Process\MutantProcessContainer;
-use function max;
 use function min;
 use Symfony\Component\Process\Process;
 
@@ -52,7 +51,7 @@ class MutantProcessContainerFactory
 {
     private const TIMEOUT_FACTOR = 10;
 
-    private const MIN_TIMEOUT = 5;
+    private const TEST_FRAMEWORK_BOOTSTRAP_THRESHOLD = 5;
 
     public function __construct(
         private readonly TestFrameworkAdapter $testFrameworkAdapter,
@@ -67,7 +66,8 @@ class MutantProcessContainerFactory
 
     public function create(Mutant $mutant, string $testFrameworkExtraOptions = ''): MutantProcessContainer
     {
-        $timeout = max(self::MIN_TIMEOUT, min(self::TIMEOUT_FACTOR * $mutant->getMutation()->getNominalTestExecutionTime(), $this->timeout));
+        // getNominalTestExecutionTime() returns the time the test-suite requires to run the test, excluding process creation and test-framework bootstrapping.
+        $timeout = min(self::TEST_FRAMEWORK_BOOTSTRAP_THRESHOLD + (self::TIMEOUT_FACTOR * $mutant->getMutation()->getNominalTestExecutionTime()), $this->timeout);
 
         $process = new Process(
             command: $this->testFrameworkAdapter->getMutantCommandLine(
--- a/src/Process/Factory/MutantProcessContainerFactory.php
+++ b/src/Process/Factory/MutantProcessContainerFactory.php
@@ -40,6 +40,8 @@ use Infection\Mutant\Mutant;
 use Infection\Mutant\TestFrameworkMutantExecutionResultFactory;
 use Infection\Process\MutantProcess;
 use Infection\Process\MutantProcessContainer;
+use function max;
+use function min;
 use Symfony\Component\Process\Process;
 
 /**
@@ -48,6 +50,10 @@ use Symfony\Component\Process\Process;
  */
 class MutantProcessContainerFactory
 {
+    private const TIMEOUT_FACTOR = 10;
+
+    private const MIN_TIMEOUT = 5;
+
     public function __construct(
         private readonly TestFrameworkAdapter $testFrameworkAdapter,
         private readonly float $timeout,
@@ -61,6 +67,8 @@ class MutantProcessContainerFactory
 
     public function create(Mutant $mutant, string $testFrameworkExtraOptions = ''): MutantProcessContainer
     {
+        $timeout = max(self::MIN_TIMEOUT, min(self::TIMEOUT_FACTOR * $mutant->getMutation()->getNominalTestExecutionTime(), $this->timeout));
+
         $process = new Process(
             command: $this->testFrameworkAdapter->getMutantCommandLine(
                 $mutant->getTests(),
@@ -69,7 +77,7 @@ class MutantProcessContainerFactory
                 $mutant->getMutation()->getOriginalFilePath(),
                 $testFrameworkExtraOptions,
             ),
-            timeout: $this->timeout,
+            timeout: $timeout,
         );
 
         return new MutantProcessContainer(
--- a/src/Process/Factory/MutantProcessContainerFactory.php
+++ b/src/Process/Factory/MutantProcessContainerFactory.php
@@ -49,7 +49,7 @@ use Symfony\Component\Process\Process;
  */
 class MutantProcessContainerFactory
 {
-    private const TIMEOUT_FACTOR = 10;
+    private const TIMEOUT_FACTOR = 5;
 
     private const TEST_FRAMEWORK_BOOTSTRAP_THRESHOLD = 5;
 
--- a/tests/e2e/TimeoutSkipped/expected-output.txt
+++ b/tests/e2e/TimeoutSkipped/expected-output.txt
@@ -4,8 +4,8 @@ Killed by Test Framework: 2
 Killed by Static Analysis: 0
 Errored: 0
 Syntax Errors: 0
-Escaped: 2
-Timed Out: 1
+Escaped: 3
+Timed Out: 0
 Skipped: 0
 Ignored: 0
 Not Covered: 0
--- a/tests/e2e/TimeoutSkipped/expected-output.txt
+++ b/tests/e2e/TimeoutSkipped/expected-output.txt
@@ -4,8 +4,8 @@ Killed by Test Framework: 2
 Killed by Static Analysis: 0
 Errored: 0
 Syntax Errors: 0
-Escaped: 3
-Timed Out: 0
+Escaped: 2
+Timed Out: 1
 Skipped: 0
 Ignored: 0
 Not Covered: 0
--- a/tests/phpunit/Process/Factory/MutantProcessContainerFactoryTest.php
+++ b/tests/phpunit/Process/Factory/MutantProcessContainerFactoryTest.php
@@ -155,9 +155,9 @@ final class MutantProcessContainerFactoryTest extends TestCase
     public static function timeoutDataProvider(): iterable
     {
         return [
-            'minimum timeout on a fast test' => [5.1, 0.01, 90],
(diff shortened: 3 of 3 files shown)


src/Process/Factory/MutantProcessContainerFactory.php
```diff
@@ -61,6 +67,8 @@ public function __construct(
 
     public function create(Mutant $mutant, string $testFrameworkExtraOptions = ''): MutantProcessContainer
     {
+        $timeout = max(self::MIN_TIMEOUT, min(self::TIMEOUT_FACTOR * $mutant->getMutation()->getNominalTestExecutionTime(), $this->timeout));
```
maks-rafalko: just for all of us to make it clear: this function currently returns the sum of all covering tests, not individual test time

we have another issue for `S`kipped mutants here: https://github.com/infection/infection/issues/1965

This doesn't block this PR, but wanted to highlight here that we still have some confusing logic for timeouts for `S`kipped mutants

maks-rafalko: hm.. re-reading PR: why do we need `MIN_TIMEOUT`?

tests/e2e/TimeoutSkipped/expected-output.txt
```diff
@@ -4,8 +4,8 @@ Killed by Test Framework: 2
 Killed by Static Analysis: 0
 Errored: 0
 Syntax Errors: 0
-Escaped: 2
-Timed Out: 1
```
maks-rafalko: note for myself once this is ready for reaview (out from draft): this e2e test specifically was created to test `S`kipped mutants, so this change is incorrect IMO - we need to test `S`kipped timeout in this e2e test

src/Process/Factory/MutantProcessContainerFactory.php
staabm: `$mutant->getMutation()->getNominalTestExecutionTime()` is the time PHPUnit requires to run this test after beeing fully bootstrapped (excluding process/bootstrap overhead).

staabm: just pushed a reworked timeout calculation which should be easier to follow

maks-rafalko: yeah, now it's much better, thanks!

tests/e2e/TimeoutSkipped/expected-output.txt
staabm: agree. actually we also discussed whether `sleep()` should be mutated at all which would bring another flavour to this test.

src/Process/Factory/MutantProcessContainerFactory.php
```diff
@@ -48,6 +49,10 @@
  */
 class MutantProcessContainerFactory
 {
+    private const TIMEOUT_FACTOR = 5;
```
staabm, line 52: 5 seconds to spawn a process and boot the test-framework is plenty of time.

but as some devs use windows which is pretty slow at process creation stuff (and maybe even anti-virus software slowing it down even more) let's be generous.

for most cases we still get way more aggressive timeouts then we had before, which is what makes the overall process run faster.
how much time we save depends on the number of timed-out mutants the project has and the number of threads we can utilize.

maks-rafalko (approved):

src/Process/Factory/MutantProcessContainerFactory.php
```diff
@@ -61,6 +66,9 @@ public function __construct(
 
     public function create(Mutant $mutant, string $testFrameworkExtraOptions = ''): MutantProcessContainer
     {
+        // getNominalTestExecutionTime() returns the time the test-suite requires to run the test, excluding process creation and test-framework bootstrapping.
+        $timeout = min(self::TEST_FRAMEWORK_BOOTSTRAP_THRESHOLD + (self::TIMEOUT_FACTOR * $mutant->getMutation()->getNominalTestExecutionTime()), $this->timeout);
```
sanmai, line 70: If a user sets a timeout of 1 second, the effective timeout will be more than 5 seconds still, correct?

I'd imagine we want to use a realistic bootstrap time, or use the user supplied timeout as the upper limit at all times.

staabm: > If a user sets a timeout of 1 second, the effective timeout will be more than 5 seconds still, correct?

No, it will be 1.

> I'd imagine we want to use a realistic bootstrap time, or use the user supplied timeout as the upper limit at all times.

as implemted in the PR the user-timeout is the max upper limit. the calculated timeout, will not be used if it gets bigger than the user configured timeout.

> I'd imagine we want to use a realistic bootstrap time

This highly depends on the machine running infection. Its already very different between the workstations of every person in the dev team and its e.g. also very different between the different kind of CI servers we use (windows vs linux, xdebug vs pcov). A realistic bootstrap time which is accurate for all machines involved is not really possible

staabm: We could lower the bootstrap time and give more weight on the multiplication factor, but atm I don't see a reason why we should do that. 

Do you have a example which the current setting does not work with well enough?

Do you think the calculated timeout with this PR is atm too optimistic or to pessimistic?

src/Process/Factory/MutantProcessContainerFactory.php
staabm: > how much time we save depends on the number of timed-out mutants the project has and the number of threads we can utilize.

example of running infection on its own codebase with 2 threads (comparable to stock GitHub Action which has 2 vCPUs):

tested on [e5f5ff9](https://github.com/infection/infection/pull/2300/commits/e5f5ff9b3565444c51ebc5fb2467172739cec9c8)

before this PR:

```
5661 mutations were generated:
    4362 mutants were killed by Test Framework
       0 mutants were caught by Static Analysis
     165 mutants were configured to be ignored
     568 mutants were not covered by tests
     542 covered mutants were not detected
       8 errors were encountered
      16 time outs were encountered

Metrics:
         Mutation Score Indicator (MSI): 79%
         Mutation Code Coverage: 89%
         Covered Code MSI: 89%

Time: 10m 35s. Memory: 0.20GB. Threads: 2
```

after this PR:
```
5663 mutations were generated:
    4364 mutants were killed by Test Framework
       0 mutants were caught by Static Analysis
     165 mutants were configured to be ignored
     568 mutants were not covered by tests
     542 covered mutants were not detected
       8 errors were encountered
      16 time outs were encountered

Metrics:
         Mutation Score Indicator (MSI): 79%
         Mutation Code Coverage: 89%
         Covered Code MSI: 89%

Time: 8m 19s. Memory: 0.20GB. Threads: 2

```

=> Note: `Time: 10m 35s` vs. `Time: 8m 19s` -> ~21% faster 🚀

sanmai: I had no reason to doubt this.

src/Process/Factory/MutantProcessContainerFactory.php
sanmai: Force-pushes really do not help: I was hoping to review only the changes since my last review, and now I have to start all over again. I would really appreciate plain merges.

sanmai: Thanks for fixing the formula, it makes sense now: `min(our estimation, user's timeout)`

src/Process/Factory/MutantProcessContainerFactory.php
sanmai: Five seconds could be too much for many, and too little for some testers who like to set up the _whole_ world in their data providers. And PHPUnit has to compute the whole data provider set to make filtering that we do - effective, IIRC.

What if we simply lower the timeout significantly? How's that going to change things?

maks-rafalko (approved):
still 👌

src/Process/Factory/MutantProcessContainerFactory.php
staabm: > Force-pushes really do not help: I was hoping to review only the changes since my last review, and now I have to start all over again. I would really appreciate plain merges.

sorry, did not realize its a problem. I will be more careful in the future.

> Thanks for fixing the formula, it makes sense now: `min(our estimation, user's timeout)`

ahh perfect. the initial formula was indeed plain wrong.

src/Process/Factory/MutantProcessContainerFactory.php
staabm: > Five seconds could be too much for many, and too little for some testers who like to set up the _whole_ world in their data providers. And PHPUnit has to compute the whole data provider set to make filtering that we do - effective, IIRC.

the PR timeout calculation takes the timings of the initial test-runs into account. I think this means if a test is slow, we calculate a bigger timeout. and we give the tests a bit of room (e.g. in case the machine running infection has other workloads at the same time) so results are more stable.

In addition I don't think we can save much by running a timeout even lower than 5 seconds (but highly increase the chance of non-stable results).

the timeout is about not wasting time for super slow mutations. fast mutations will exit a process faster then 5 seconds, as soon as they finish the testing.

> What if we simply lower the timeout significantly? How's that going to change things?

my goal was to make infection faster with as little impact on results as possible.

we could of course lower the process timeout, but this might create Timed-Out mutations for existing project which hadn't those before (kind of a BC break).
I assumed there is a reason why the timeout is 30s.

we could also do both - lower the default timeout and stick with the per mutant timeout

tests/e2e/TimeoutSkipped/expected-output.txt
```diff
@@ -4,8 +4,8 @@ Killed by Test Framework: 2
 Killed by Static Analysis: 0
 Errored: 0
 Syntax Errors: 0
-Escaped: 3
-Timed Out: 0
+Escaped: 2
+Timed Out: 1
```
sanmai: This stuff was flapping for me too.

maks-rafalko:
thank you @staabm

PR #2301, opened 2025-07-07
Allow setting static analysis tool in config file

maks-rafalko:
Closes https://github.com/infection/infection/issues/2281

This PR:

- [x] Adds new feature ...
- [x] Covered by tests
- [ ] Doc PR: TODO

PR #2304, opened 2025-07-08
Finally, fix the worst and the most unmaintainable test ever

maks-rafalko:
I'm fed up with this test `tests/phpunit/Configuration/ConfigurationFactoryTest.php`. 

So much time wasted [for all those](https://github.com/infection/infection/pull/2301/files#diff-f33b8a70fc38f5ac2171c5a792363356d123fbeccf47128c209259eb38bbdbaa) `null` / `false` values that need to be added in tens of test cases.

Now, this will not be needed anymore. Got suggestions from [here](https://github.com/infection/infection/pull/2226#discussion_r2153885831) and from other issues and applied a combination of them

- parameters with default values allow to **not** duplicate not needed for a test case parameters
- named parameters in `yield` data providers in combination with default values in parameters greatly improve readability of the test cases and allow to provide only those that are needed for a test case, and nothing more

you can see how much code is removed

image


sanmai (approved):
Let's gooo! 

(I can ask my Claude to fix similar tests elsewhere in the project, based on this example.)

maks-rafalko:
> (I can ask my Claude to fix similar tests elsewhere in the project, based on this example.)

we can try ;) not sure though what the result will be as this is an extreme example, but why not to try..

sanmai:
Just let me know what tests need to be fixed this way, and I'll get it busy.

maks-rafalko:
> Just let me know what tests need to be fixed this way, and I'll get it busy.

this one is a good candidate: 

https://github.com/infection/infection/blob/a569236d59705979a337c6e7a9a03bbae2e66d9d/tests/phpunit/Configuration/Schema/SchemaConfigurationTest.php#L49

it also has many parameters, and small enough to try it with Claude

theofidry:
FYI – not that you may want to rework that once more, I find useful in this kind of test to use a VO object, e.g. `Scenario`, which can hold of those values and have some utility methods.

e.g.

```php
function scenarioProvider(): iterable {
  $baseScenario = Scenario::create(minMsi: 92);

  yield 'with threads' => Scenario->withThreads(100)->build();
}
```

Personally I like that you can use named parameters there instead of associative arrays, and you can also hide a bit of complexity in the VO to allow more expressiveness and better readability in the test case itself

maks-rafalko:
yeah I saw your similar comment in one of the issues, but decided not to do it as the current state completely solved my issue.

Probably as the next step it would be better with Scenario, but I need to play with it. The first issue I see with Scenario is that I we still need `input` and `expected` values to pass to assertions here:

https://github.com/infection/infection/blob/ae0a25acc8d7ad154aa39fa44ce30e33d051a250/tests/phpunit/Configuration/ConfigurationFactoryTest.php#L232

and that would mean that it will be rewritten like:

```php
$this->assertConfigurationStateIs(
+         $config(),
-          $scenario->getConfig()
-          $expectedTimeout,
+          $scenario->getExpectedTimeout(),
// ...
```

etc.

And we will have the same method `create()` with tens of parameters with default values as it was done here with a test case method, so I don't see a value yet (if I didn't miss something).

PR #2310, opened 2025-07-09
Improve ParallelProcessRunner to guard against ReturnRemoval mutator

sanmai:
Ref: https://github.com/infection/infection/pull/2296

### DIFF
--- a/src/Process/Runner/ParallelProcessRunner.php
+++ b/src/Process/Runner/ParallelProcessRunner.php
@@ -40,6 +40,7 @@ use Composer\XdebugHandler\Process;
 use function count;
 use Generator;
 use Infection\Process\MutantProcessContainer;
+use Iterator;
 use function max;
 use function microtime;
 use function range;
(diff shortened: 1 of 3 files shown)


src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -213,10 +215,10 @@ private function startProcess(MutantProcessContainer $mutantProcessContainer, in
      *  - from the input stream of processes containers (original mutant processes)
      *  - from the "next" killer processes, created if a PHPUnit process doesn't kill a Mutant
      *
-     * @param MutantProcessContainer[] $bucket
-     * @param Generator<MutantProcessContainer> $input
+     * @param array<MutantProcessContainer|null> $bucket
```
sanmai:
```suggestion
     * @param array<MutantProcessContainer> $bucket
```

maks-rafalko (approved):

src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -213,22 +215,27 @@ private function startProcess(MutantProcessContainer $mutantProcessContainer, in
...
             }
 
             return 0;
         }
 
         $start = microtime(true);
 
-        $bucket[] = $input->current();
+        $current = $input->current();
+        Assert::notNull($current);
```
sanmai, line 236: `$input->current()` will return `null` if the generator is not yet `valid()`  - exactly what happens if we remove the `return 0` above; Really we should switch this whole thing to SplQueue or some other more appropriate tool.

PR #2316, opened 2025-07-09
Refactor ParallelProcessRunner to simplify [WIP, do not review yet]

### LINKED ISSUE #2314: Replace TEST_TOKEN management with much more simple computed TEST_TOKEN

In https://github.com/infection/infection/pull/1504 we added a `TEST_TOKEN`, and although the idea is sound, the implementation is too complex: just trying to understand the number of things `ParallelProcessRunner` has to wrangle (in https://github.com/infection/infection/pull/2310) made my head hurt yesterday.

I propose that we add a new `SEQ_TEST_TOKEN`, a sequential serial number for each produced mutation emitted as an environment variable. We increment it **right before starting a new mutant process** (it can call an injected dependency to get the next number before starting up). 

Only the `MutantProcess` will be aware of it, no need for `ParallelProcessRunner` even care about it.

If the user only has three servers, they only need to do:

```
$serverId = $SEQ_TEST_TOKEN % 3;
```
To get the server ID, and since there's only three processes in flight (`-j 3`), this token will always map to a free server.

**Additional context**

With this simplification, we can remove TEST_TOKEN and reduce the scope of `ParallelProcessRunner` by one thing.

sanmai:
This PR:

- [ ] Fixes #2314
- [ ] Covered by tests

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,6 +66,7 @@
         "symfony/finder": "^6.4 || ^7.0",
         "symfony/process": "^6.4 || ^7.0",
         "thecodingmachine/safe": "^v3.0",
+        "tumblr/chorus-timekeeper": "^0.1.1",
         "webmozart/assert": "^1.11"
     },
     "conflict": {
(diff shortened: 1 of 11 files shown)


sanmai:
Have to really use Fibers here, return later.

PR #2317, opened 2025-07-10
Add TestTokenHandler to simplify TEST_TOKEN generation

### LINKED ISSUE #2314: Replace TEST_TOKEN management with much more simple computed TEST_TOKEN

sanmai:
This PR:

- [x] Adds a `TestTokenHandler` to compute `TEST_TOKEN` where they needed 
- [x] There are no BC breaks, existing tests should work as before
- [x] Covered by tests

Fixes #2314

In a follow up PR we can simplify the parallel process runner further by extracting the `nextMutantProcessKillerContainer` follow-back into a dedicated class following SRP.


sanmai:
@maks-rafalko What are the chances you can test this still works for you?

src/Process/TestTokenHandler.php
```diff
@@ -33,18 +33,37 @@
...
+     * @return non-negative-int
+     */
+    public function getNextToken(int $base = 1): int
+    {
+        if ($this->threadCount === 0) {
+            return $base;
+        }
+
+        return $this->processCount++ % $this->threadCount + $base;
+    }
```
sanmai, line 68: Git glitch: this wasn't a rename.

sanmai:
OK, I found a problem. Let me fix it and I'll reopen it.

src/Configuration/Configuration.php
```diff
@@ -291,6 +291,7 @@ public function getMsiPrecision(): int
         return $this->msiPrecision;
     }
 
+    /** @return non-negative-int */
```
staabm, line 294: there is also a more readable variant:

```suggestion
    /** @return 0|positive-int */
```

sanmai: I don't think PHPStorm supports these `0|1` types :cry:

maks-rafalko:
> @maks-rafalko What are the chances you can test this still works for you?

let me know please when I can test it

PR #2322, opened 2025-07-10
Directly enqueue follow-up processes

sanmai:
This PR:

- [ ] Makes it slightly more simple

### DIFF
--- a/src/Process/Runner/ParallelProcessRunner.php
+++ b/src/Process/Runner/ParallelProcessRunner.php
@@ -44,6 +44,7 @@ use Iterator;
 use function max;
 use function microtime;
 use function range;
+use SplQueue;
 use Symfony\Component\Process\Exception\ProcessTimedOutException;
 use function usleep;
 use Webmozart\Assert\Assert;
@@ -107,8 +108,8 @@ final class ParallelProcessRunner implements ProcessRunner
         $generator = self::toGenerator($processContainers);
 
         // Bucket for processes to be executed
-        /** @var array<MutantProcessContainer> $bucket */
-        $bucket = [];
+        /** @var SplQueue<MutantProcessContainer> $bucket */
+        $bucket = new SplQueue();
 
         // Load the first process from the queue to buy us some time.
         $this->fillBucketOnce($bucket, $generator, 1);
@@ -122,9 +123,8 @@ final class ParallelProcessRunner implements ProcessRunner
                 break;
             }
 
-            $mutantProcessContainer = array_shift($bucket);
-
-            if ($mutantProcessContainer !== null) {
+            if (!$bucket->isEmpty()) {
+                $mutantProcessContainer = $bucket->dequeue();
                 $threadIndex = array_shift($this->availableThreadIndexes);
 
                 Assert::integer($threadIndex, 'Thread index can not be null.');
@@ -159,7 +159,7 @@ final class ParallelProcessRunner implements ProcessRunner
 
             // In any case try to load at least one process to the bucket
             $this->fillBucketOnce($bucket, $generator, 1);
-        } while ($bucket !== [] || $this->runningProcessContainers !== [] || $this->nextMutantProcessKillerContainer !== []);
+        } while (!$bucket->isEmpty() || $this->runningProcessContainers !== [] || $this->nextMutantProcessKillerContainer !== []);
     }
 
     private function tryToFreeNotRunningProcess(): ?MutantProcessContainer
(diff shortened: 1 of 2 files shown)


src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -133,36 +129,34 @@ public function run(iterable $processContainers): iterable
...
-        } while (!$bucket->isEmpty() || $this->runningProcessContainers !== [] || $this->nextMutantProcessKillerContainer !== []);
+        } while (!$bucket->isEmpty() || $this->runningProcessContainers !== []);
     }
 
-    private function tryToFreeNotRunningProcess(): ?MutantProcessContainer
+    /**
+     * @param SplQueue<MutantProcessContainer> $bucket
+     * @return iterable<MutantProcessContainer>
+     */
+    private function tryToFreeNotRunningProcess(SplQueue $bucket, bool &$terminated = false): iterable
```
sanmai:
```suggestion
    private function tryToFreeNotRunningProcess(SplQueue $bucket, bool &$terminated): iterable
```

composer.json
```diff
@@ -74,6 +75,7 @@
     },
     "require-dev": {
         "ext-simplexml": "*",
+        "dg/bypass-finals": "^1.9",
```
staabm: just a suggestion: maybe use `@final` phpdoc instead of `final` php keyword, so we don't need this ugly dependency

sanmai: Yep, already doing this in the next commit.

sanmai: We'll also remove the 77 tests some very dumb-smart AI added.

src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -176,28 +212,28 @@ private function tryToFreeNotRunningProcess(): ?MutantProcessContainer
...
+            $mutantProcess->markAsFinished();
 
-                unset($this->runningProcessContainers[$index]->mutantProcessContainer);
-                unset($this->runningProcessContainers[$index]);
+            $this->availableThreadIndexes[] = $indexedMutantProcess->threadIndex;
 
-                if ($mutantProcessContainer->hasNext()) {
-                    $mutantProcessContainer->createNext();
+            unset($this->runningProcessContainers[$index]->mutantProcessContainer);
+            unset($this->runningProcessContainers[$index]);
```
sanmai, line 224: I'm not sure this is really necessary, but I left it alone.

src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -210,37 +246,6 @@ private function startProcess(MutantProcessContainer $mutantProcessContainer, in
...
-        $start = microtime(true);
-
-        $current = $input->current();
-        Assert::notNull($current);
-
-        $bucket->enqueue($current);
-        $input->next();
-
-        return (int) (microtime(true) - $start) * self::NANO_SECONDS_IN_MILLI_SECOND; // ns to ms
-    }
```
sanmai, line 242: I had to make it protected to install spies.

tests/phpunit/Process/Runner/ParallelProcessRunnerTest.php
```diff
@@ -155,6 +357,414 @@ public static function threadCountProvider(): iterable
...
+        $method = $reflection->getMethod('wait');
+
+        // Test scenario where poll - timeSpentDoingWork would be exactly 0
+        $timeSpentDoingWork = 10000; // Exactly poll time
+
+        // With original code: max(0, 10000 - 10000) = max(0, 0) = 0
+        // With mutated code: max(1, 10000 - 10000) = max(1, 0) = 1
+        $timeKeeperMock->expects($this->once())
+            ->method('usleep')
+            ->with($this->identicalTo(0)); // Should be 0, not 1
```
sanmai, line 503: Very questionable variable lifespans all over the place here and there.

maks-rafalko (approved):
I noticed that `tumblr/chorus-timekeeper` uses GPLv2 and chatgpt with Claude gives completely different clarifications about possibility of using this inside Infection licensed under BSD-3 Clause.

Without continue spending too much time learning about it, are you aware of this? can we use it here? If not, will continue investigating

tests/phpunit/Process/Runner/ParallelProcessRunnerTest.php
```diff
@@ -142,6 +152,195 @@ public function test_fill_bucket_once_with_exhausted_generator_does_not_continue
         $this->assertSame(0, $result);
     }
 
+    public function test_initial_fill_bucket_once_called_with_one_using_mock(): void
```
staabm, line 154: you did a great job in adding lots of new tests.

I wonder whether we could land some of the tests in a separate PR beforehand, to ease reviewing and separation of regression tests and new features/behaviour?

I might have overlooked something and this might not be doable easily. in such case, just leave it like that.

sanmai: In different circumstances this will be a great idea, and I would do that myself to reduce the review time of each PR, but here I plan to dismantle the runner even further and these tests will be no more.

tests/phpunit/Process/Runner/ParallelProcessRunnerTest.php
```diff
@@ -142,6 +151,195 @@ public function test_fill_bucket_once_with_exhausted_generator_does_not_continue
...
+    {
+        // This test uses a partial mock to verify fillBucketOnce is called with 1
+        // This will kill the IncrementInteger mutation on line 115
```
maks-rafalko: minor: I think these comments were generated by AI, and will be stale as soon as we add a new line to the source file. What do you think about removing line numbers?

maks-rafalko (approved):

src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -176,28 +212,28 @@ private function tryToFreeNotRunningProcess(): ?MutantProcessContainer
...
+            if ($mutantProcessContainer->hasNext()) {
+                $mutantProcessContainer->createNext();
 
-                    return null;
-                }
+                // Enqueue the needed static analysis run
+                $bucket->enqueue($mutantProcessContainer);
 
-                return $mutantProcessContainer;
+                return;
```
sanmai, line 232: I've seen `continue`/`break` flapping here.

PR #2328, opened 2025-07-12
Use executionOrder="defects,random" for phpunit.xml

staabm:
when re-running a test-suite which failed before it makes a lot of sense to run the previously failed tests first.
that way we validate whether the already known problems are gone and in addition we also shorten the feedback loop.

extracted from https://github.com/infection/infection/pull/2267


sanmai (approved):
Cool!

Issue #2318, opened 2025-07-10
Infection should produce only covered mutations by default


sanmai:
@maks-rafalko argues that [Infection should be developer-friendly](https://github.com/infection/infection/issues/2312#issuecomment-3054266357) for newcomers, and here's what I repeatedly observed about uncovered mutations that are the standard default thing for Infection.

When you explain the mutation testing to a peer, it usually goes likes this: mutation testing is about tweaking the code this way or another way to see if the tests do not notice the change, and if tests do not notice the change it the code, means the tests have holes and can be improved.

And I had few folks trying Infection on projects of their own, only to be utterly confused by an avalanche of uncovered mutations: didn't we just talk about changing the _covered_ code? The explanation from there usually goes to propose to use `--only-covered` at all times, and for the very insistent we can explain that **power users** use the uncovered-MSI as a replacement for the code coverage percentile. 

See what I did there? If you ever need to mention that a certain default feature is for power users, you product is **not** new user friendly.

So here comes my proposal to make `--only-covered` the default behavior.

Previously:
- https://github.com/infection/infection/issues/1064
- https://github.com/infection/infection/issues/2313

If we do this, https://github.com/infection/infection/issues/2313 won't be necessary for me anymore.

(I'd go as far as following through on https://github.com/infection/infection/issues/1064, but that's going to be a huge BC break for some hardcore people.)

sanmai:
Here's another piece of evidence, @staabm [said loud and clear](https://github.com/infection/infection/issues/1064#issuecomment-3056140483):

> I do not yet have an idea when uncovered mutations are useful.

A person who's been _contributing_ to Infection has no idea how uncovered mutations are useful? That's saying.

staabm:
> I'd go as far as following through on [#1064](https://github.com/infection/infection/issues/1064), but that's going to be a huge BC break for some hardcore people.

wdyt about a softer transition, aka. turn `--only-covered` to be opt-out instead of the currently used opt-in.
that way we could e.g. ask users to comment on this issue here, so we can see how many people really need the non-covered mutations...?

or are we sure enough that we can just drop it and don't look back?

sanmai:
I actually understood that uncovered mutations are very useful to guide LLMs. They can see it loud and clear that they do not have the test coverage for this line, without having to dig into the coverage report.

This is not to say it should be the default: opt-out `--only-covered` (`--with-uncovered`) will work great for me.

theofidry:
maybe it should be `--allow-uncovered`?

Issue #2313, opened 2025-07-09
Allow setting only-covered in config file


sanmai:
**Is your feature request related to a problem? Please describe.**
When reviewing https://github.com/infection/infection/issues/1064 I understood that I only run Infection in prod with `--only-covered` because I only care about actionable testing defects, not anything that tests do not cover (it is already obvious from the coverage report).

**Describe the solution you'd like**

Similar to https://github.com/infection/infection/issues/2281, I'd like to be able to set `--only-covered` from the config file, and forget about ever needing to type it.

PR #2336, opened 2025-07-16
Remove `--only-covered`; Introduce `--with-uncovered` instead

staabm:
⚠️ Breaking change: we drop support for CLI option `--only-covered`

---

- [x] Docs PR: https://github.com/infection/site/pull/278

Summarizing the actual implementation details
- we change the infection default from "mutate uncovered code" (aka. `--only-covered`) to "don't mutate uncovered code" (aka. `--with-uncovered`)
- `--only-covered` seems to be a bad default, as usually you only want covered lines to be mutated
- running uncovered code is still useful for LLMs, therefore we don't drop the switch

outcome
- we get more useful mutations per default
- we get faster infection runs as we produce less mutations per default

---

in-detail comments/discussions are contained in the referenced issues:
- closes https://github.com/infection/infection/issues/2318
- closes https://github.com/infection/infection/issues/1064
- closes https://github.com/infection/infection/issues/2313


src/Configuration/Configuration.php
```diff
@@ -258,7 +258,7 @@ public function isDebugEnabled(): bool
 
     public function mutateOnlyCoveredCode(): bool
     {
-        return $this->onlyCovered;
+        return !$this->allowUncovered;
```
staabm: here is the actual change in default.

everything else is just naming.

src/Command/RunCommand.php
```diff
@@ -90,7 +90,7 @@ final class RunCommand extends BaseCommand
     private const OPTION_TEST_FRAMEWORK_OPTIONS = 'test-framework-options';
 
     /** @var string */
-    private const OPTION_ONLY_COVERED = 'only-covered';
+    private const OPTION_ALLOW_UNCOVERED = 'allow-uncovered';
```
maks-rafalko: just a question (no need to update for now): we had another proposal to name it `--with-uncovered`, and I like it because we have another proposal to have `--with-timeouts`, so in case both are implemented, we would have

- `--with-uncovered`
- `--with-timeouts`

which looks pretty and consistent.

Looking to this PR from this perspective, do you think it's better to rename it?

maks-rafalko (commented):
could you also please prepare a PR for docs? I've updated initial message to include a checkbox

maks-rafalko:
Another idea (definitely should not be done as part of this PR, but just throwing for a discussion, probably to extract):

now we have the following output at the end of mutation testing:

```
Metrics:
         Mutation Score Indicator (MSI): 100%
         Mutation Code Coverage: 100%
         Covered Code MSI: 100%
```

and now, when we by default mutate only covered code, MSI and Covered MSI will always be the same. 

So the idea is to remove `Mutation Score Indicator (MSI): 100%` from the output unless we use `--allow-uncovered` (`--with-uncovered`)

Related to 

- https://github.com/infection/infection/issues/222

sanmai:
--allow-uncovered leaves the impression we are taking a passive role and this option has no costs, as there things already out there

In reality, with this option we will make a special effort to produce these mutations and the user will spend more time waiting.

--with-uncovered does not have this problem and plays well with --with-timeouts and any other inclusion directives we may add later

I'd also consider --include-... but arguably --with-... reads easier

staabm:
I have no strong feelings. renamed it to `--with-uncovered` which total makes sense for me, too.

staabm:
> and now, when we by default mutate only covered code, MSI and Covered MSI will always be the same.
> 
> So the idea is to remove `Mutation Score Indicator (MSI): 100%` from the output unless we use `--allow-uncovered` (`--with-uncovered`)

agree that we can simplify the summary now, as its redundant.
we should do so in a separate PR, as I think it might get a separate focus and needs some bikeshedding ;).

it might also affect some other result formats/reports?

theofidry:
Maybe instead of a hard break we could add a warning comment when the `--only-covered` option is used?

maks-rafalko:
> Maybe instead of a hard break we could add a warning comment when the `--only-covered` option is used?

do you think it's really needed? From one side I agree it's more dev-friendly. But we are on 0.x, so I'm not sure we need to spend much efforts with BC for this particular case. We can (and always do) write a good changelog for this BC break.

staabm:
> > Maybe instead of a hard break we could add a warning comment when the `--only-covered` option is used?
> 
> do you think it's really needed? From one side I agree it's more dev-friendly. But we are on 0.x, so I'm not sure we need to spend much efforts with BC for this particular case.

I have no strong opinion on this one either. can do whatever you guys think is necessary.

sanmai:
We can reduce the amount of breakage if we assume that if `--min-msi=NN` provided then a user wanted to consider uncovered code, implying `--with-uncovered`.

I'm not sure that `--min-covered-msi=100` is going to trigger the same implicit rule - I suppose not.

My CI on many open-source projects that I maintain will really appreciate this heuristic :slightly_smiling_face: 

i.e. if you start the process like this:

```
infection --min-msi=100
```

it should imply

```
infection --min-msi=100 --with-uncovered
```

theofidry:
I think `min-msi` should applied for covered only, and we should have `--min-uncovered-msi` instead of `--min-covered-msi`

maks-rafalko:
Let's merge it and I will handle `--min-msi` renaming separately.

Thank you @staabm!

staabm:
> I will handle `--min-msi` renaming separately.

thanks!

PR #2343, opened 2025-07-24
Switch to sanmai/di-container

sanmai:
Let's try to see if the package I extracted still works the same.

https://packagist.org/packages/sanmai/di-container


src/Container.php
```diff
@@ -169,7 +160,7 @@
 /**
  * @internal
  */
-final class Container
+final class Container extends DIContainer
```
staabm, line 163: nice to see this custom DI implementation moves out of the infection repo.

maks-rafalko (approved):

PR #2344, opened 2025-07-27
Rename `--min-msi` to `--min-uncovered-msi`

maks-rafalko:
Follow up for https://github.com/infection/infection/pull/2336

Now we have explicit, self-explanatory options

- `--min-covered-msi`
- `--min-uncovered-msi`

- [x] Docs PR: https://github.com/infection/site/pull/279

### DIFF
--- a/tests/phpunit/Metrics/MinMsiCheckerTest.php
+++ b/tests/phpunit/Metrics/MinMsiCheckerTest.php
@@ -146,7 +146,7 @@ final class MinMsiCheckerTest extends TestCase
 
         $msiChecker->checkMetrics(2, 12., 10., $this->consoleOutput);
 
-        $this->assertSame('', normalize_trailing_spaces($this->output->fetch()),);
+        $this->assertSame('', normalize_trailing_spaces($this->output->fetch()));
     }
 
     public function test_does_not_suggest_to_increase_the_min_msi_if_limit_is_zero(): void
--- a/src/Metrics/MinMsiChecker.php
+++ b/src/Metrics/MinMsiChecker.php
@@ -36,7 +36,6 @@ declare(strict_types=1);
 namespace Infection\Metrics;
 
 use Infection\Console\ConsoleOutput;
-use function var_dump;
 
 /**
  * @internal


sanmai:
What are the reasons we have to introduce a breaking change and can't live with `--min-msi` that implicitly adds `--with-uncovered`?

I.e. if you start the process like this:

```
infection --min-msi=100
```

it will imply

```
infection --min-msi=100 --with-uncovered
```

And so everyone who uses Infection on CI (many, many people) would not need to move a finger to adopt this new version.

And we can emit a "deprecated" warning.

tests/phpunit/Metrics/MinMsiCheckerTest.php
```diff
@@ -86,14 +86,23 @@ public function test_it_fails_the_check_if_the_msi_is_lower_than_the_min_msi():
...
     }
 
+    public function test_it_does_not_fails_the_check_if_min_msis_are_zero(): void
+    {
+        $msiChecker = new MinMsiChecker(false, 0., 0.);
+
+        $msiChecker->checkMetrics(2, 0., 0., $this->consoleOutput);
+
+        $this->assertSame('', $this->output->fetch());
+    }
```
maks-rafalko, line 104: the only one escaped mutant left is 

```diff
1) /opt/infection/src/Metrics/MinMsiChecker.php:111    [M] GreaterThan [ID] 8a9730728ec29bfa8cec7d775bbc0d25

@@ @@
     }
     private function isUncoveredMsiInsufficient(float $uncoveredMsi): bool
     {
-        return $this->minUncoveredMsi > 0 && $uncoveredMsi < $this->minUncoveredMsi;
+        return $this->minUncoveredMsi >= 0 && $uncoveredMsi < $this->minUncoveredMsi;
     }
     private function isCoveredCodeMsiInsufficient(float $coveredCodeMsi): bool
     {
```

and this is a good example of unkillable mutants.

Mathematically and logically, actual uncovered MSI can not be <0, that's why I have no idea how to kill it.

maks-rafalko:
> What are the reasons we have to introduce a breaking change and can't live with `--min-msi` that implicitly adds `--with-uncovered`?

I don't like anything that is "implicit". And we are on 0.x so BC is ok (at least for me).

I would like to have as descriptive options as possible, without people thinking that `--min-msi` implicitly implies `--with-uncovered`.

It's hard to think about, hard to explain, hard to maintain.

> And so everyone who uses Infection on CI (many, many people) would not need to move a finger to adopt this new version.

This is what I want to avoid - I don't want to overcomplicate the code just to save user's time. 

Someone who uses 0.x of Infection will better read the upgrade guide, update 1 option, and no implicit logic will be maintained here that will be a reason for questions / discussions / explanations in the future.

tests/phpunit/Metrics/MinMsiCheckerTest.php
sanmai: It sounds like you should be able to call this method with the reflection and validate the boundary.

Also, why `$this->minUncoveredMsi >= 0 && $uncoveredMsi < $this->minUncoveredMsi` is incorrect?

maks-rafalko: I think it's an equivalent, so both `>` and `>=` are correct. How would you write the test that passes for `>` and fails for `>=`?

sanmai:
My biggest concern with this change is with... LLMs. They are very slow to pick up breaking changes like this, so they will be trying to run Infection with `--min-msi` for entirety. So I would really, really want us to to support an implicit pair `--min-msi=100 --with-uncovered`  in addition to `--min-covered-msi` & `--min-uncovered-msi`. 

And I guess I can add this workaround later myself, what do you think @maks-rafalko ?

sanmai (approved):
Looks like a straightforward rename.

tests/phpunit/Metrics/MinMsiCheckerTest.php
sanmai:
```
$this->assertFalse($sut->isUncoveredMsiInsufficient(0));
```
+ apart from calling this with reflection

(To be fair, these are one of the few mutations [that I turn off completely](https://github.com/sanmai/pipeline/blob/7e8e5032320b87593eb531b2923fc5d688cd6642/infection.json.dist#L16-L25).)

sanmai: If `$uncoveredMsi` cannot be less than zero, then we should not check for that:

```diff
- return $this->minUncoveredMsi > 0 && $uncoveredMsi < $this->minUncoveredMsi;
+ return $uncoveredMsi < $this->minUncoveredMsi;
```

sanmai:
Pulling from site#279:

> It reads odd. The option says `UncoveredMsi` but the explanation says "for all the code". What if we are making a mistake here?

So `--min-uncovered-msi` reads like *MSI for uncovered code* which is not what it is.

What if we don't need to rename anything? How can we explain that we _must_ make this rename happen?

maks-rafalko:
> So `--min-uncovered-msi` reads like _MSI for uncovered code_ which is not what it is.
> 
> What if we don't need to rename anything? How can we explain that we _must_ make this rename happen?

actually, I'm ok with that so we can close this PR. Then what we have in `master` right now is already a good solution, and we don't need anything to do after https://github.com/infection/infection/pull/2336

sanmai:
In this case, I suppose I can ask: what stops us from closing this PR and make a new release?

maks-rafalko:
Closing as per discussion

PR #2348, opened 2025-07-30
CI: Added PHP 8.5 setup

staabm:


sidz:
Hey @staabm 

We usually add one soft of hack when we add support to the next PHP version.

Add the following step after "Remove the configured PHP platform".
```
      - name: Configure for PHP >= 8.4
        if: "matrix.php-version >= '8.4"
        run: |
          composer config platform.php 8.4.99
```

See https://github.com/infection/infection/commit/92384c3823d44d0a7df5dd68b7aad5ce38b322cb#diff-95557d53bf91069e59d82b9d8fcfaf52ec84a762858983edd5ef3c9b3e4c8191R61-R65 for reference

staabm:
thanks for the hint. 

I just started sending downstream PRs to our dependencies so we can hopefully get the ecosystem onto 8.5.
in case I get stuck I will come back here and try the suggested workaround

- https://github.com/nette/tester/pull/460

staabm:
locally PHP8.5 works for me. the relevant dependencies need a release but the current dev-versions look good already

maks-rafalko (approved):

.github/workflows/e2e-tests.yaml
```diff
@@ -58,6 +58,11 @@ jobs:
         if: matrix.dependencies != 'locked'
         run: composer config --unset platform.php
 
+      - name: Configure for PHP >= 8.4
+        if: "matrix.php-version >= '8.4"
```
maks-rafalko: I see that something is wrong with GH actions, probably we have extra `'` here before `8.4`?

the same is below

maks-rafalko (changes requested):
many of the GH actions disappeared, I left a comment with a possible issue

image

.github/workflows/e2e-tests.yaml
staabm: good catch, thanks

maks-rafalko (approved):
thank you @staabm

PR #2354, opened 2025-08-02
test: Add multi byte tests to DiffColorizerTest

staabm:
adding more test coverage to prevent regressions when killing timed out mutants

### DIFF
--- a/tests/phpunit/Differ/DiffColorizerTest.php
+++ b/tests/phpunit/Differ/DiffColorizerTest.php
@@ -218,5 +218,38 @@ final class DiffColorizerTest extends TestCase
                 <diff-add>+        $changedLinesRangeXX = <diff-add-inline>2</diff-add-inline>;</diff-add></code>
                 CODE,
         ];
+
+        yield 'multibyte characters full-deletion' => [
+            <<<'CODE'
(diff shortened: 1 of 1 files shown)


tests/phpunit/Differ/DiffColorizerTest.php
```diff
@@ -218,5 +218,37 @@ protected function name()
...
+            <<<'CODE'
+                     function ($a) {
+                -        return 'déjà_vu' . 'bar';
+                +        return 'déjà_vu';
+                     }
+                CODE,
+            <<<'CODE'
+                <code>
+                     function ($a) {
+                <diff-del>-        return 'déjà_v<diff-del-inline>u' . 'ba</diff-del-inline>r';</diff-del>
```
staabm, line 248: tbh I don't know whether this results are correct, but at least they reflect what the current implementation does

maks-rafalko: Looks suspicious, agree. I think there is a bug in the algorithm and utf-8 characters, but since we are not adding/changing business logic here, let's merge and solve/debug it in the future

maks-rafalko (approved):
thank you @staabm

PR #2359, opened 2025-08-05
Fixed mutator list for E2E fixtures

### LINKED ISSUE #2306: Updating E2E test fixtures is an unnecessary chore

When I was adding a new mutator #2296, I spent about 1/3 of time to try to fix all E2E tests because for some there's GOLDEN=1 that magically fixes all issues, and for many there is no standard way to update the expected results. Furthermore, there are often several fixtures you have to update. So it feels like a major timesink.

Update the configuration files in these fixtures to use only specific mutators that they require.

Alternatively, we can device a way to automatically update all fixtures. Say, it can be a Makefile rule. But it feels like a wasted effort because these fixtures are suppposed to test one specific thing at a time.

**Additional context**

Can we also rename the `DX` label into `UX`, because it is about the developer-user experience, and `DX` to actual developer experience?

sanmai:
This PR:

- [x] Fixes #2306

> We successfully addressed GitHub issue #2359 "Updating E2E test fixtures is an unnecessary chore" by configuring specific mutator lists for 24 E2E tests that run actual mutation testing. This prevents these tests from being affected when new mutators are added to Infection's default profile, eliminating the need to update their expected output fixtures every time the mutator set changes.
>
>The solution involved analyzing all 52 E2E tests to distinguish between those that run mutation testing versus configuration-only tests, then adding carefully selected mutator configurations positioned as the second-to-last JSON key to minimize future diff sizes. Tests with custom scripts that only validate configuration, setup, or error handling were intentionally excluded since they don't run mutations and wouldn't benefit from this optimization.

### DIFF
--- a/tests/e2e/Config_Bootstrap/infection.json
+++ b/tests/e2e/Config_Bootstrap/infection.json
@@ -9,5 +9,8 @@
         "summary": "infection.log"
     },
     "bootstrap": "src/bootstrap.php",
+    "mutators": {
+        "PublicVisibility": true
+    },
     "tmpDir": "."
 }
--- a/tests/e2e/Custom_tmp_dir/infection.json
+++ b/tests/e2e/Custom_tmp_dir/infection.json
@@ -8,5 +8,8 @@
     "logs": {
         "summary": "infection.log"
     },
-    "tmpDir": "."
+    "tmpDir": ".",
+    "mutators": {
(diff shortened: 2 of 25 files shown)


maks-rafalko (approved):

maks-rafalko:
thank you @sanmai

Issue #2341, opened 2025-07-19
ForeachKeyRemovalMutator: Foreach key variable removal


sanmai:
I'd love to have a new mutator that removes the key variable from `foreach` loops while preserving the value binding. 

```
- foreach ($foo as $i => $arg) {
+ foreach ($foo as $arg) {
```

It will help me identify:
- Dead or redundant variables,
- Potentially misleading key-value iteration when only the value matters,
- Hidden assumptions about keys that are never validated,
- And a variety of LLM-generated gunk

It complements existing mutators like `Foreach_` and `RemoveForeach_`, but operates in a minimal way by only removing the key variable, keeping the loop functional.

Implementation:

- Target node type: `PhpParser\Node\Stmt\Foreach_`
- Mutation: clone the node, set `$keyVar = null`.

sanmai:
It looks like the consensus is this is a bad idea 💀

maks-rafalko:
1. Context: https://github.com/infection/infection/pull/2361#issuecomment-3168325154
2. Interesting comment/idea that needs to be analyzed, same way as p.1: https://github.com/infection/infection/pull/2361#issuecomment-3263838102
3. also, we can merge it disabled by default for those who don't care about the fact that mutants are killed in ~100% (if p.2 will not lead to anything)

thus, too early to close :)

PR #2361, opened 2025-08-07
Add `ForeachKeyRemoval` mutator

mttsch:
This PR:

- [x] Adds `ForeachKeyRemoval` mutator that removes the key variable from foreach loops
- [x] Covered by tests

Fixes https://github.com/infection/infection/issues/2341


mttsch:
I will create a pull request for the documentation once it is clear that this PR can/will be merged.

maks-rafalko:
I would appreciate if you also can do the following analysis [as I did here](https://github.com/infection/infection/pull/2296#pullrequestreview-2999286532) (I will add these requirements to all new mutators):

-  Generate [per-mutator](https://infection.github.io/guide/usage.html) report for Infetion itself for this new mutator and post it to the initial PR description (`bin/infection --mutators=ForeachKeyRemoval --log-verbosity=all --debug --threads=max`)
- Generate `text` report for the same run as above and attach it here, so we can analyze how this new mutator behaves and what tests kill it (or probably we have invalid cases) (`bin/infection --mutators=ForeachKeyRemoval --log-verbosity=all --debug --threads=max`)
- we will see overlappings with other mutators, if any

If you need any help or have any questions, or if you don't have time for doing that - let me know, I will help. Thanks!

tests/phpunit/Mutator/Loop/ForeachKeyRemovalTest.php
```diff
@@ -0,0 +1,84 @@
...
+                PHP
+            ,
+            <<<'PHP'
+                <?php
+
+                foreach ($array as $value) {
+                }
+                PHP
+            ,
+        ];
```
maks-rafalko, line 72: in case it's got accepted, I would also add more tests, such as:

- `foreach ($array as [$a, $b]) {`
- `foreach ($array as [, , $c]) {`
- `foreach ($arr as &$value) {`

and combinations

https://www.php.net/manual/en/control-structures.foreach.php

mttsch: I already added some more.

mttsch:
@maks-rafalko  Here is the desired information:

# Effects per Mutator

| Mutator           | Mutations | Killed by Test Framework | Test Timings min/avg/max | Killed by Static Analysis | Static Analysis Timings min/avg/max | Escaped | Errors | Syntax Errors | Timed Out (Limit: 25 secs) | Skipped | Ignored | MSI (%s) | Covered MSI (%s) |
| ----------------- | --------- | ------------------------ | ------------------------ | ------------------------- | ----------------------------------- | ------- | ------ | ------------- | -------------------------- | ------- | ------- | -------- | ---------------- |
| ForeachKeyRemoval |        30 |                       29 |  0.27 / 0.39 / 0.70 secs |                         0 |             0.00 / 0.00 / 0.00 secs |       1 |      0 |             0 |                          0 |       0 |       0 |    96.67 |            96.67 |

# Infection Log
[infection.log](https://github.com/user-attachments/files/21670220/infection.log)

maks-rafalko:
Thanks, IMO results are very arguable. 

In almost all the cases we got:

- `Argument #1 ($haystack) must be of type string, null given` because variable doesn't exist (=it's `null`)
- or `Undefined variable $namespace` because we use undefined (removed) variable

all of this fails the tests, so kills mutants.

The only 1 more or less relevant case for the whole Infection source code base is 1 escaped mutant:

```diff
1) /opt/infection/src/Mutator/MutatorParser.php:68    [M] ForeachKeyRemoval [ID] 6fafcc864c0118a3e84ac47ca1b7b53c

@@ @@
             return [];
         }
         $parsedMutators = array_filter(array_map('trim', explode(',', $unparsedMutators)));
-        foreach ($parsedMutators as $index => $profileOrMutator) {
+        foreach ($parsedMutators as $profileOrMutator) {
             if (array_key_exists($profileOrMutator, ProfileList::ALL_PROFILES)) {
                 continue;
             }

$ '/opt/infection/vendor/bin/phpunit' '--configuration' '/tmp/infection/phpunitConfiguration.6fafcc864c0118a3e84ac47ca1b7b53c.infection.xml'
  PHPUnit 11.5.28 by Sebastian Bergmann and contributors.
  
  Runtime:       PHP 8.2.29
  Configuration: /tmp/infection/phpunitConfiguration.6fafcc864c0118a3e84ac47ca1b7b53c.infection.xml
  
  .......                                                             7 / 7 (100%)
  
  Time: 00:00.026, Memory: 10.00 MB
  
  OK (7 tests, 7 assertions)
```

this mutant shows that we can remove foreach key and it is redundant, even when is used.

I have a feeling that in the most cases it will be just killed same way it's killed here, in Infection's source code.

I tried to run it for the whole BetterReflection codebase:

| Mutator           | Mutations | Killed by Test Framework | Test Timings min/avg/max | Killed by Static Analysis | Static Analysis Timings min/avg/max | Escaped | Errors | Syntax Errors | Timed Out (Limit: 10 secs) | Skipped | Ignored | MSI (%s) | Covered MSI (%s) |
| ----------------- | --------- | ------------------------ | ------------------------ | ------------------------- | ----------------------------------- | ------- | ------ | ------------- | -------------------------- | ------- | ------- | -------- | ---------------- |
| ForeachKeyRemoval |        21 |                       18 |  0.27 / 0.57 / 1.13 secs |                         0 |             0.00 / 0.00 / 0.00 secs |       0 |      0 |             0 |                          0 |       3 |       0 |   100.00 |           100.00 |

all 100% of non-skipped mutants are killed with similar "undefined variable" or "must be of type string, null given" errors.

See the whole log attached.

[infection-better-relfection.log](https://github.com/user-attachments/files/21687490/infection-better-relfection.log)

So, my main concern that this mutator generates mutants that in ~100% cases will be killed _even without tests_, thus wasting MT time. 

Probably we can make it opt-in if you need it? 

Could someone, who is interested in this mutator, post the **real** cases where this mutator detects issues from real projects? With MSI / per-mutator logger results as without them it doesn't make sense to analyze it.

mttsch:
@maks-rafalko I have no real opinion on whether the mutator makes sense, I am just trying to contribute to more open source projects at the moment and saw a maintainer suggesting such a mutator in https://github.com/infection/infection/issues/2341 and though it to be a good entrypoint for a first contribution to this project.

Regarding the `MutatorParser` case: I just have a superficial understanding of mutation testing but I did some digging, and isn't the real problem here that `MutatorParserTest` does not actually have a test case that covers the code branch where `$index` would be used in https://github.com/infection/infection/blob/e8897edbed6fd662e33ce025e86d136533d48e29/src/Mutator/MutatorParser.php#L84

If no test covers this branch (easily verifable by adding some output or `exit()` in the condition), mutation testing can not properly "handle" this scenario?

maks-rafalko:
> Regarding the `MutatorParser` case: I just have a superficial understanding of mutation testing but I did some digging, and isn't the real problem here that `MutatorParserTest` does not actually have a test case that covers the code branch where `$index` would be used in

you are right. Let me highlight my thoughts: this mutator **found** the real lack of the tests. But, in majority of the cases mutants generated by this mutator are killed even when we don't have tests (by the code itself, because we remove a variable that is used / expected to be used).

> I have no real opinion on whether the mutator makes sense, I am just trying to contribute to more open source projects at the moment and saw a maintainer suggesting such a mutator in #2341

right, let's wait for other opinions then. Thank you anyway.

mk-mxp:
Would it be possible to use a constant `$key = array_key_first($array);`? That would actually run the code under test instead of erroring out.

maks-rafalko:
> Would it be possible to use a constant `$key = array_key_first($array);`? That would actually run the code under test instead of erroring out.

but how does it correspond to the fact that the idea is to remove the `$key`? Didn't get it. Could you please elaborate?

mk-mxp:
> > Would it be possible to use a constant `$key = array_key_first($array);`? That would actually run the code under test instead of erroring out.
> 
> but how does it correspond to the fact that the idea is to remove the `$key`? Didn't get it. Could you please elaborate?

In the issue #2341, the purpose is given as:

> It will help me identify:
>
> *    Dead or redundant variables,
> *    Potentially misleading key-value iteration when only the value matters,
> *    Hidden assumptions about keys that are never validated,
> *    And a variety of LLM-generated gunk

In the current version, complete removal of the `$key` helps with `Dead or redundant variables` by making the code error out of testing (not passing tests caused by unset variables). You objected this solution with:

> that this mutator generates mutants that in ~100% cases will be killed even without tests, thus wasting MT time.

So I propose to not completely remove the `$key` from the code, but replace it with a constant `$key = array_key_first($array);`, so that the loop now will be executed and actually executed - but with a key that never changes. This would cover the remaining points of the OP, validating that not only the value matters, assumptions about keys are tested and LLM produced errors are surfaced more likely.

So, instead of mutating to:

```diff
         $parsedMutators = array_filter(array_map('trim', explode(',', $unparsedMutators)));
-        foreach ($parsedMutators as $index => $profileOrMutator) {
+        foreach ($parsedMutators as $profileOrMutator) {
             if (array_key_exists($profileOrMutator, ProfileList::ALL_PROFILES)) {
```

the mutation could be:

```diff
         $parsedMutators = array_filter(array_map('trim', explode(',', $unparsedMutators)));
-        foreach ($parsedMutators as $index => $profileOrMutator) {
+        $index = array_key_first($parsedMutators);
+        foreach ($parsedMutators as $profileOrMutator) {
             if (array_key_exists($profileOrMutator, ProfileList::ALL_PROFILES)) {
```

Does that make my proposal understandable?

src/Mutator/Loop/ForeachKeyRemoval.php
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * @psalm-mutation-free
+     *
+     * @return iterable<Node\Stmt\Foreach_>
+     */
+    public function mutate(Node $node): iterable
+    {
+        $newForeachNode = clone $node;
+        $newForeachNode->keyVar = null;
```
sanmai, line 93: Do we actually need to clone it here?

sanmai (approved):
Outside of one escaping mutation in the implementation of this mutator, it looks like just what I'd want. Also, it is pretty fast - all mutations are killed under 0.7 second, so in other way - it is very light on the user.

staabm:
I agree with @maks-rafalko that I don't see the value yet, which this mutation provides (independent from the implemtation in this PR). Detection of a unused variable seems like a case which can be more efficiently be detected with other tools.

in projects with a slow test suite this might lead to a significant slow-down for no real gain

sanmai:
Well, it sounds like Psalm and Phan can detect these, neither of which I'm using these days.

PR #2366, opened 2025-08-12
Remove lower-bound composer constraints

staabm:
since infection requires PHP8.2+ we don't need these lower bounds which support PHP < 8.2


sidz (approved):
:+1:

maks-rafalko:
thank you @staabm

composer.json
```diff
@@ -48,21 +48,21 @@
         "ext-libxml": "*",
         "ext-mbstring": "*",
         "composer-runtime-api": "^2.0",
-        "colinodell/json5": "^2.2 || ^3.0",
-        "composer/xdebug-handler": "^2.0 || ^3.0",
-        "fidry/cpu-core-counter": "^0.4.0 || ^0.5.0 || ^1.0",
+        "colinodell/json5": "^3.0",
+        "composer/xdebug-handler": "^3.0",
```
stof, line 52: This makes inflection incompatible with projects that are stuck with version 2 of `composer/xdebug-handler` due to another dependency, while supporting both v2 and v3 of that package is straightforward.
such project might end up getting an old release of inflection instead of the latest one.

Reducing dependency ranges in `require` (vs `require-dev`) has an impact on project using this package.
PHP is not the only constraint to resolve dependencies. Composer needs to find a set of packages where all constraints are satisfied (and unlike in the npm ecosystem, it cannot install multiple versions of a package, as PHP does not have a module system based on importing symbols. All classes and functions are defined globally).

maks-rafalko: from the PR description I thought that all the removed lower versions were for PHP <=8.2, while it's not true for xdebug-handler - both v2 and v3 [works with PHP ^8.0](https://github.com/composer/xdebug-handler/blob/31d57697eb1971712a08031cfaff5a846d10bdf5/composer.json#L21)

@stof did you find it because you couldn't upgrade Infection to the latest version because of this change?

Would reverting this to `"composer/xdebug-handler": "^2.0 || ^3.0",` resolve your issue?

stof: No. I found this PR because of the discussion in https://github.com/composer/composer/issues/12510 asking to provide a feature in composer to automate this kind of changes (which used this PR as example) and I thought about the impact of the feature (where doing that for require-dev might be OK, but doing that for normal requirements is as risky as any other `composer bump` on them for libraries)

PR #2368, opened 2025-08-13
Fix PHPUnit 9.x compatibility

staabm:
the tested started failling because with https://github.com/infection/infection/pull/2366 we no longer allow old sebastian/diff releases which in turn are required for old phpunit (unintentional side-effect)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,7 +62,7 @@
         "sanmai/duoclock": "^0.1.0",
         "sanmai/later": "^0.1.7",
         "sanmai/pipeline": "^7.0",
-        "sebastian/diff": "^6.0 || ^7.0",
+        "sebastian/diff": "^5.0 || ^6.0 || ^7.0",
         "symfony/console": "^6.4 || ^7.0",
         "symfony/filesystem": "^6.4 || ^7.0",
         "symfony/finder": "^6.4 || ^7.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -62,7 +62,7 @@
         "sanmai/duoclock": "^0.1.0",
         "sanmai/later": "^0.1.7",
         "sanmai/pipeline": "^7.0",
-        "sebastian/diff": "^6.0 || ^7.0",
+        "sebastian/diff": "^5.0 || ^6.0 || ^7.0",
```
staabm: if we still want to support phpunit 9.x we would have to re-add also `^4.0`... are we fine with dropping 9.x which is "End of Bugfix Support" since February 2, 2024?

see https://phpunit.de/supported-versions.html

staabm: Just checked installation stats on packagist.org. 

It seems 9.x is the most widely used version and we should still support it

maks-rafalko: I'm personally ok with either ways. If we decide to not support 9.x, developers can use older Infection versions.

At least this logic is applied when we got read of non-supported PHP versions each year

maks-rafalko (approved):

composer.json
staabm: Will add back 9.x support later today. There is no benefit in dropping it atm

maks-rafalko: cool, thanks. This should also fix https://github.com/infection/infection/pull/2367

staabm:
should be good to go.

failling tests seem to be the ones which fail from time to time

maks-rafalko (approved):

PR #2382, opened 2025-08-26
test: Add more tests for the SourceFileCollector

theofidry:


maks-rafalko (approved):
seems like this PR contains changes from a couple of others

tests/phpunit/FileSystem/SourceFileCollector/SourceFileCollectorTest.php
```diff
@@ -107,7 +110,27 @@ public static function sourceFilesProvider(): iterable
...
         ];
 
-        yield 'one directory, no filter, absolute path excludes' => [
+        yield 'one directory with a child directory excluded via its full path' => [
+            [self::FIXTURES . '/case0'],
+            [self::FIXTURES . '/case0/sub-dir'],
+            [
+                'case0/a.php',
+                'case0/outside-symlink.php',
+                'case0/sub-dir/b.php',  // Does not work
```
maks-rafalko: so is it a bug?

theofidry: yes, unless I got something wrong

theofidry: or it is expected given the setup, I'm not sure tbh, I'm always a bit confused with the finder

maks-rafalko: I would investigate it if you have some more time.

- If it's a correct Finder's behavior, than we shouldn't have these comments?
- If it's Infection's bug - I would rather plan to fix it rather than keep incorrect tests

it looks a bit strange that we confirm incorrect behavior by passing tests :) It's like a TODO that will never be fixed. But feel free to disagree and merge

theofidry: I created https://github.com/infection/infection/issues/2594 to this effect.

I came across a thread (documented in the linked issue) that hints there may be more to this issue. I do not really want to get distracted by this at the moment so I created the issue and hopefully we can come back to it later or maybe someone will grab it.

PR #2392, opened 2025-09-01
fix: Make the source file collector yield a list

theofidry:
In `Infection\TestFramework\Factory`, we have:

```php
/**
 * Get only those source files that will be mutated to use them in coverage whitelist
 *
 * @return list<SplFileInfo>
 */
private function getFilteredSourceFilesToMutate(): array
{
    if ($this->sourceFileFilter->getFilters() === []) {
        return [];
    }

    /** @var list<SplFileInfo> $files */
    $files = iterator_to_array($this->sourceFileFilter->filter($this->infectionConfig->getSourceFiles()));

    return $files;
}
```

And the test in `SourceFileCollectorTest` was hitting at least that it was desired that the yielded values of `SourceFileCollector` would give a list.

However, the unit test was incorrect, as `take($files)->toList()`, as you can imagine, gives a list, regardless of what the yielded keys are.

I am not 100% sure we should update the downstream code or make explicit that `SourceFileCollector` yields a `iterable<string, SplFileInfo>` instead. Neither do I think that this is an error that PHPStan can capture (I could not find anything about PHPStan checking iterables for _lists_, i.e. yielding an iterable with guarantee that no value is lost due to a duplicate key).

Either way, I think this is a (minor) problem, I'm happy to adjust the PR to fix it the other way around if you prefer.

### DIFF
--- a/src/FileSystem/SourceFileCollector.php
+++ b/src/FileSystem/SourceFileCollector.php
@@ -59,12 +59,16 @@ class SourceFileCollector
             return [];
         }
 
-        return Finder::create()
+        $files = Finder::create()
             ->in($sourceDirectories)
             ->exclude($excludedFilesOrDirectories)
             ->notPath($excludedFilesOrDirectories)
             ->files()
             ->name('*.php')
         ;
+
+        foreach ($files as $file) {
+            yield $file;
+        }
     }
 }
--- a/tests/phpunit/FileSystem/SourceFileCollector/SourceFileCollectorTest.php
+++ b/tests/phpunit/FileSystem/SourceFileCollector/SourceFileCollectorTest.php
@@ -63,7 +63,7 @@ final class SourceFileCollectorTest extends TestCase
 
         $files = (new SourceFileCollector())->collectFiles($sourceDirectories, $excludedFilesOrDirectories);
 
-        $files = take($files)->toList();
+        $files = take($files)->toAssoc();
 
         $this->assertSame(
             $expected,


sanmai (approved):
It makes sense in a way that Finder's `getIterator()` returns Iterator<string, SplFileInfo>`, on the other hand consider that we can remove the `assertIsList` assertion as it does not assert anything about the tested code.

So instead it could be:
- We can keep returning the finder as entities are guaranteed to be unique by the filesystem (i.e. we can't have the same key twice).
- Then, if we replace `toList()`  with `toAssoc()` and remove `assertIsList()` - the test will be passing.

theofidry:
Yes, essentially either we want `"list" iterable<SplFileInfo>` or `iterable<string, SplFileInfo>`, both are file, but we need to pick one

maks-rafalko:
> So instead it could be:
> 
>     * We can keep returning the finder as entities are guaranteed to be unique by the filesystem (i.e. we can't have the same key twice).
> 
>     * Then, if we replace `toList()`  with `toAssoc()` and remove `assertIsList()` - the test will be passing.

I think this is better because, personally, I don't like the fact that we need to add 

```diff
+foreach ($files as $file) {
+    yield $file;
+}
```

it makes the code harder to understand. But, if you think the current approach is ok - I'm happy too.

theofidry:
Would could also change the API to make it `iterable<string, SplFileInfo>`, I think that could be useful too tbh (and in this case we don't have to add the piece of code you are mentioning)

maks-rafalko (approved):
I've rerun failed jobs, let's see if they pass (some strange e2e tests failures)

theofidry:
Thanks

PR #2394, opened 2025-09-04
fix: Improve the source filtering messages

theofidry:
Executing infection like so:

```
infection --git-diff-base=upstream/main
```

Results in:

```
Cannot pass "--git-diff-base" without "--git-diff-filter"
```

It is a bit incorrect as you can also pass `--git-diff-lines` instead of `--git-diff-filter`. I updated the message to mention it and while I was at it I also proposed other changes for the other git/source related error messages.


theofidry:
The failures look unrelated

Issue #2380, opened 2025-08-25
RFC: Source Collection


theofidry:
# How source collection currently works in Infection

- a basic configuration is defined at the `infection.json5` level which is used for creating the `SourceFileCollector`
- a **filter** can be provided via the console command option
- a git diff can be provided via the console command
- git diff & filter are used to create `SourceFileFilter` which returns a command separated list of the files captured by the filter which should already belong to the sources configured (note that it does not take care of the excluded directories yet at this level)

Random note: we seem to specify "exclude directories" in some places like in `SourceFilter` but in practice we also pass it files.

In the `TestFramework\Factory` we then apply the filter to the source files:

```php
$this->sourceFileFilter->filter($this->infectionConfig->getSourceFiles())
```

which is used for building the test framework configuration.

Otherwise, the `TraceProvider` provides covered/uncovered (note that in the case of uncovered I think it emits redundant values) to which the `FileFilter` is applied:

```php
$filteredTraces = $this->bufferedFilter->filter(
        $this->primaryTraceProvider->provideTraces(),
);
```

# The problem

I think right now we are dealing with quite a lot of indirection and the sources of not quite clear. It took me a while to understand the flow from above and I'm not too confident I didn't miss anything. Some pieces of code are in the Schema, the configuration, the configuration factory, the filesystem, the tracer...

Overall I do not question the code: it makes sense and seems correct. But it feels like it is a bit all over the place with some out of place.

Another point, but maybe it is more my bias kicking in, is that I think exposing _traces_ as the starting point is a bit confusing. I had to give this one a lot of more thoughts. It is a point that I genuinely do not know if I'm just overthinking this or it is just me intuitively approaching the problem a different way and then being bugged at coming across the opposite way as a solution. Ultimately though, I do not think it matters too much, but I'll try to explain my thoughts.

Either you:

1. start with the source code, and filter by coverage/trace
2. start with the coverage and filter by the source code

Depending of the scenario, one makes more sense than the other. For instance, if you think of analysing a whole codebase, you're more likely going to find _less_ files starting from the coverage (1.). However, I think in most cases the other way around (2.) makes more sense:

- If you start from a diff, you're likely working with a coverage report applying to the whole codebase or at least a bigger piece than just the modified source code.
- If you filter by a source file or symbol, chances are you want to do so for a given source file/symbol rather than a test.

In other words, I feel like if one of the two needs to be a star, it should be the source code.

# The solution

[...]

PR #2395, opened 2025-09-04
refactor: Simplify the source filtering API

theofidry:
Currently, we have 4 separate options to manipulate the filtering/git information. The values of those is simply transferred in various places, which is not ideal when those values can be mutually exclusive.

This PR tries to bring a bit more cohesion whilst making the code easier to read. I'll break down the different kind of changes here:

- Introduce a `GitOptions` value object that is meant to hold the values passed from the options of `RunCommand`.
- Since `filter` and the git options are mutually exclusive, and that we now have unions in PHP, I decided to combine them into a "source filter" parameter.
- For the `Configuration`, we want the filter itself (normal or git-based) and pieces of the git information that is used here and there elsewhere. In other words, for the git options, there is two distinct purposes: the filter, and the rest. To clarify this, the 4 options are combined into two:
  - the resolved filter (whether it was the normal one or a git based one)
  - `GitSource` which contains git related information necessary for other places (unrelated to the filter itself)

TODOs:

- [ ] Depends on #2394 
- [ ] Adjust the tests

### DIFF
--- a/src/Command/RunCommand.php
+++ b/src/Command/RunCommand.php
@@ -490,18 +490,37 @@ final class RunCommand extends BaseCommand
         $gitDiffBase = $input->getOption(self::OPTION_GIT_DIFF_BASE);
 
         if ($isForGitDiffLines && $gitDiffFilter !== Container::DEFAULT_GIT_DIFF_FILTER) {
-            throw new InvalidArgumentException(sprintf('Cannot pass both "--%s" and "--%s" options: use none or only one of them', self::OPTION_GIT_DIFF_LINES, self::OPTION_GIT_DIFF_FILTER));
+            throw new InvalidArgumentException(
+                sprintf(
+                    'The options "--%s" and "--%s" are mutually exclusive. Please use only one of them.',
+                    self::OPTION_GIT_DIFF_LINES,
+                    self::OPTION_GIT_DIFF_FILTER,
+                ),
+            );
         }
 
         if ($gitDiffBase !== Container::DEFAULT_GIT_DIFF_BASE && $gitDiffFilter === Container::DEFAULT_GIT_DIFF_FILTER && $isForGitDiffLines === Container::DEFAULT_GIT_DIFF_LINES) {
-            throw new InvalidArgumentException(sprintf('Cannot pass "--%s" without "--%s"', self::OPTION_GIT_DIFF_BASE, self::OPTION_GIT_DIFF_FILTER));
+            throw new InvalidArgumentException(
+                sprintf(
+                    'The option "--%s" cannot be used without the option "--%s" or "--%s".',
+                    self::OPTION_GIT_DIFF_BASE,
+                    self::OPTION_GIT_DIFF_LINES,
+                    self::OPTION_GIT_DIFF_FILTER,
+                ),
+            );
         }
 
         $filter = trim((string) $input->getOption(self::OPTION_FILTER));
 
         if ($filter !== '' && $gitDiffFilter !== Container::DEFAULT_GIT_DIFF_BASE) {
             throw new InvalidArgumentException(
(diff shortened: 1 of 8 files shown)


src/Command/RunCommand.php
```diff
@@ -699,4 +658,113 @@ private static function getFormatterName(InputInterface $input): FormatterName
...
+
+    private static function assertGitBaseHasRequiredFilter(
+        ?string $gitDiffFilter,
+        bool $isForGitDiffLines,
+        ?string $gitDiffBase,
+    ): void {
+        // TODO: previously was $gitDiffBase !== Container::DEFAULT_GIT_DIFF_BASE
+        //  I do not understand the point of those default values if the default is not using git
+        //  If we provide a default there, then the requirement is still valid, as a base requires a filter.
+        if ($gitDiffBase === null) {
```
theofidry: extract this change

src/Command/RunCommand.php
```diff
@@ -699,4 +658,113 @@ private static function getFormatterName(InputInterface $input): FormatterName
...
+
+        $hasFilter = $gitDiffFilter !== null || $isForGitDiffLines;
+
+        if ($hasFilter) {
+            return;
+        }
+
+        // TODO: seems like you can pass DEFAULT_GIT_DIFF_LINES instead of DEFAULT_GIT_DIFF_FILTER
+        //  too thought?
+        // TODO: port the improve message
```
theofidry: todo

src/Command/RunCommand.php
```diff
@@ -699,4 +658,113 @@ private static function getFormatterName(InputInterface $input): FormatterName
...
+        self::assertOnlyOneTypeOfFiltering($filter, $gitOptions);
+
+        return $filter ?? $gitOptions;
+    }
+
+    private static function getFilter(InputInterface $input): ?string
+    {
+        $value = trim((string) $input->getOption(self::OPTION_FILTER));
+
+        return $value === '' ? null : $value;
```
theofidry: that's a behaviour change that, I think, could already be extracted

src/Command/RunCommand.php
```diff
@@ -699,4 +658,113 @@ private static function getFormatterName(InputInterface $input): FormatterName
...
+        ?string $gitDiffFilter,
+        bool $isForGitDiffLines,
+    ): void {
+        if (
+            $gitDiffFilter !== null
+            && $isForGitDiffLines
+        ) {
+            throw new InvalidArgumentException(
+                sprintf(
+                    'Cannot pass both "--%s" and "--%s" options: only one type of filtering is allowed.',
```
theofidry: to review the various exceptions see if there is something to change or that changed to extract them

src/Command/RunCommand.php
```diff
@@ -802,8 +787,8 @@ private static function assertGitBaseHasRequiredFilter(
         ?string $gitDiffFilter,
         ?string $gitDiffBase,
     ): void {
-        if ($gitDiffBase !== Container::DEFAULT_GIT_DIFF_BASE
-            && $gitDiffFilter === Container::DEFAULT_GIT_DIFF_FILTER
```
theofidry: to extract

src/Configuration/ConfigurationFactory.php
```diff
@@ -194,9 +188,13 @@ public function create(
             isDryRun: $dryRun,
             ignoreSourceCodeMutatorsMap: $ignoreSourceCodeMutatorsMap,
             executeOnlyCoveringTestCases: $executeOnlyCoveringTestCases,
-            isForGitDiffLines: $useGitDiff,
-            gitDiffBase: $refinedGitBase,
-            gitDiffFilter: $gitDiffFilter,
+            isForGitDiffLines: $sourceFilter instanceof GitDiffFilter,
```
theofidry, line 191: To clarify: this is temporary, but this is only going to be sorted via #2380. In the end, those 3 parameters will be removed and instead we will only pass `?SourceFilter` to the configuration, in which `GitDiffFilter` contains all the information help by those 3 values.

PR #2407, opened 2025-09-20
fix filter for PHPUnit

### LINKED ISSUE #1545: Infection does not run any tests if `--filter` is too big

theofidry:
This PR fixes #1545.

This is a draft for myself, I will extract reviewable PRs from it.

### DIFF
--- a/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
+++ b/src/TestFramework/PhpUnit/CommandLine/ArgumentsAndOptionsBuilder.php
@@ -35,34 +35,23 @@ declare(strict_types=1);
 
 namespace Infection\TestFramework\PhpUnit\CommandLine;
 
-use function array_key_exists;
 use function array_map;
 use function array_merge;
 use function count;
-use function end;
 use function explode;
 use function implode;
 use function in_array;
 use Infection\AbstractTestFramework\Coverage\TestLocation;
 use Infection\TestFramework\CommandLineArgumentsAndOptionsBuilder;
-use function is_numeric;
 use function ltrim;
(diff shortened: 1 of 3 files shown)


maks-rafalko (approved):
👍

PR #2409, opened 2025-09-22
perf: Fix non-cached usage of version_compare in FilterBuilder

theofidry:
I came across this while implementing #2407.

I fixed it in this specific place, but I wonder if we should have a more generic utility like a `MemoizedVersionCompare`, but that would be more suited in a separate PR.


maks-rafalko:
do you have any numbers? I just never saw it's a problem, so wonder how you have noticed this is an issue

theofidry:
Not really. I just know `version_compare()` is not cached and its performance varies based on the comparison and version format. Since we call it to just know if it's PHPUnit 10+, it seems silly to be potentially doing that thousand of times.

But I doubt I could see it in a Blackfire profile given how long the actual mutation processes take.

PR #2410, opened 2025-09-22
refactor: Minor tweaks to FilterBuidler

theofidry:
In #2408 I was careful to introduce as few changes as possible to make the review easier.

This PR introduces the remaining refactor only changes for #2407.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1162,16 +1162,16 @@
         },
         {
             "name": "sanmai/pipeline",
-            "version": "7.1",
+            "version": "7.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/sanmai/pipeline.git",
-                "reference": "d01565ef9f5cd7d1019c5f8bee09497067511f36"
+                "reference": "9ead38671b86a2c689484a871c72ce645326e9dd"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/sanmai/pipeline/zipball/d01565ef9f5cd7d1019c5f8bee09497067511f36",
-                "reference": "d01565ef9f5cd7d1019c5f8bee09497067511f36",
+                "url": "https://api.github.com/repos/sanmai/pipeline/zipball/9ead38671b86a2c689484a871c72ce645326e9dd",
+                "reference": "9ead38671b86a2c689484a871c72ce645326e9dd",
                 "shasum": ""
             },
             "require": {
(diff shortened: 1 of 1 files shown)


src/TestFramework/PhpUnit/CommandLine/FilterBuilder.php
```diff
@@ -65,60 +65,78 @@ public static function createFilters(
         $filters = [];
 
         foreach ($tests as $testLocation) {
-            $testCaseString = $testLocation->getMethod();
-
-            $partsDelimitedByColons = explode('::', $testCaseString, self::MAX_EXPLODE_PARTS);
+            $test = $testLocation->getMethod();
```
maks-rafalko, line 68: not critical, but I'm not sure why it was renamed. It's not test, it's exactly test case (method).

and on line 72 below we have `$testCaseClassName` extracted from this `$test` variable, so the change is a bit unclear.

`$testCase` and `$testCaseClassName` - would it work?

theofidry: Apologies I sorta expected that question and forgot to add a mention.

I am wondering if `TestLocation::getMethod()` is the correct name here, because essentially this is the string representation of the executed test:

```
Infection\Tests\EngineTest::test_initial_test_run_succeeds
Infection\Tests\TestFramework\PhpUnit\CommandLine\ArgumentsAndOptionsBuilderTest::test_it_can_build_the_command_with_filter_option_for_covering_tests_for_mutant#1 simple test case, 1 with data set and special character &gt;
FeatureA:Scenario A1
```

So there is cases were it's the simple `__METHOD__` of the test executed, but there is cases where it does not match (data providers, non-PHPUnit frameworks).

This is for `TestLocation`, but this piece of code we are in is PHPUnit specific, so in practice the only two forms we have (here) is:

```
Infection\Tests\EngineTest::test_initial_test_run_succeeds
Infection\Tests\TestFramework\PhpUnit\CommandLine\ArgumentsAndOptionsBuilderTest::test_it_can_build_the_command_with_filter_option_for_covering_tests_for_mutant#1 simple test case, 1 with data set and special character &gt;
```

I wouldn't call it `testCase` because (at least for me) the test case is the PHPUnit test class, i.e. it would be `Infection\Tests\EngineTest` or `Infection\Tests\TestFramework\PhpUnit\CommandLine\ArgumentsAndOptionsBuilderTest`.

It is not the method either because if you have a data provider it is more than that.

Nonetheless, `$testMethod` is a bit confusing too because it potentially has the provider key so I renamed those.

maks-rafalko: > wouldn't call it `testCase` because (at least for me) the test case is the PHPUnit test class

funny, I think I write too much TS these days, so I mixed up PHPUnit test case (class) with jest test cases (separate "methods").

Makes sense, thanks for the explanation

theofidry: To be transparent I'm still not fully satisfied with the terminology, but I find hard to name the different (PHPUnit) forms. This is the best I've found so far but maybe we'll figure out something better another time.

PR #2428, opened 2025-10-07
Fix escaping mutation in FilterBuilder

sanmai:
I noticed there's a singular escaping mutation in #2407; let's get it killed.


staabm (approved):

theofidry:
Makes me wonder why that works as opposed to ignore that mutation: we don't increment/decrement values from constants?

sanmai:
I don't think they show up in the coverage reports, and even if they did - testing constants is going to be silly.

PR #2429, opened 2025-10-07
PHPStan integration: drop `--fail-without-result-cache`

staabm:
drop `--fail-without-result-cache` which does not work when a project contains PHPStan extension classes and those get mutated. these will not create a cache hit and emit the following warning:

```
Result cache not used because extension file /Users/m.staab/dvl/phpstan-doctrine/src/Rules/Doctrine/ORM/QueryBuilderDqlRule.php hash does not match.
```

background story: see https://github.com/phpstan/phpstan-doctrine/pull/686#issuecomment-3376112789 and following


maks-rafalko (approved):

maks-rafalko:
thank you @staabm 

Released https://github.com/infection/infection/releases/tag/0.31.4

PR #2433, opened 2025-10-09
Fix --dry-run option fatal error

sanmai:
## Summary
Fixes #2432 - `--dry-run` option was causing fatal errors when trying to retrieve output from non-started processes.

An example:
```
> In Process.php line 569:
                                                                 
  [Infected\Symfony\Component\Process\Exception\LogicException]  
  Start time is only available after process start.              
                                                                 

Exception trace:
  at vendor/symfony/process/Process.php:569
 Infected\Symfony\Component\Process\Process->getStartTime() at src/Mutant/TestFrameworkMutantExecutionResultFactory.php:25
 Infected\Infection\Mutant\TestFrameworkMutantExecutionResultFactory->createFromProcess() at src/Process/MutantProcess.php:45
 Infected\Infection\Process\MutantProcess->getMutantExecutionResult() at src/Process/Runner/MutationTestingRunner.php:84
 Infected\Infection\Process\Runner\MutationTestingRunner::containerToFinishedEvent() at vendor/sanmai/pipeline/src/Standard.php:284
 Infected\Pipeline\Standard::applyOnce() at vendor/sanmai/pipeline/src/Standard.php:890
 Infected\Pipeline\Standard->eachInternal() at vendor/sanmai/pipeline/src/Standard.php:878
 Infected\Pipeline\Standard->each() at src/Process/Runner/MutationTestingRunner.php:32
 Infected\Infection\Process\Runner\MutationTestingRunner->run() at src/Engine.php:82
 Infected\Infection\Engine->runMutationAnalysis() at src/Engine.php:42
 Infected\Infection\Engine->execute() at src/Command/RunCommand.php:84
 Infected\Infection\Command\RunCommand->executeCommand() at src/Command/BaseCommand.php:29
 Infected\Infection\Command\BaseCommand->execute() at vendor/symfony/console/Command/Command.php:152
 Infected\Symfony\Component\Console\Command\Command->run() at vendor/symfony/console/Application.php:723
 Infected\Symfony\Component\Console\Application->doRunCommand() at vendor/symfony/console/Application.php:244
 Infected\Symfony\Component\Console\Application->doRun() at vendor/symfony/console/Application.php:137
 Infected\Symfony\Component\Console\Application->run() at bin/infection:78
 require() at build/infection.phar:14 
```

## Changes
Modified `TestFrameworkMutantExecutionResultFactory` to handle dry-run mode with three guards:
1. `retrieveProcessOutput()` - returns empty string instead of asserting process termination
[...]

PR #2435, opened 2025-10-09
Fix --dry-run option using DryRunProcess wrapper (alternative approach)

### LINKED ISSUE #2432: `--dry-run` doesn't work and produces fatal errors
Run `infection --dry-run` - it doesn't work.

This should be fixed.

sanmai:
## Summary
Fixes #2432 - `--dry-run` option was causing fatal errors when trying to retrieve output from non-started processes.

This is an alternative implementation to #2433, using a decorator pattern instead of guard clauses.

## Changes
Created `DryRunProcess` class that:
- Wraps a real `Process` using `fromShellCommandline()` factory method
- Presents it as already terminated with passing test output
- Returns `'OK (0 tests, 0 assertions)'` via `PASSING_TEST_OUTPUT` constant to trigger `ESCAPED` status
- Makes coupling between process layer and test framework semantics explicit

Modified `MutantProcessContainerFactory` to wrap processes in dry-run mode using `DryRunProcess::fromProcess()`.

## Implementation details
Uses Symfony's `Process::fromShellCommandline()` to create a process from the real command line string. This is cleaner than manually storing the command line in a property and overriding `getCommandLine()`.

## Why this approach
- **Exercises real process creation code** - catches bugs in command construction
- **Cleaner separation** - no guards in `TestFrameworkMutantExecutionResultFactory`
- **Decorator pattern** - wraps existing process with additional behavior
- **Single point of change** - only the factory needs to know about dry-run
- **Works universally** - compatible with both PHPUnit and PHPStan adapters
- **More test coverage** - exercises actual detection logic including `testsPass()` implementation

## Abstraction leaks (both approaches)
**Both approaches involve coupling, just at different layers:**

**This PR (#2435)**: `DryRunProcess` (process layer) knows about `TestFrameworkAdapter::testsPass()` (test framework layer)
- Leak: Process knows the pattern `"OK ("` triggers passing test detection
- Made explicit via `PASSING_TEST_OUTPUT` public constant
- Tested in both `DryRunProcessTest` and `TestFrameworkMutantExecutionResultFactoryTest`
- **Benefit**: Exercises full detection logic for better smoke testing

**Alternative (#2433)**: `TestFrameworkMutantExecutionResultFactory` (result interpretation layer) knows about dry-run mode (execution layer)
- Leak: Result factory understands execution context (dry-run)
- Guard clauses check configuration before interpreting results
- **Downside**: Skips detection logic with early returns

Neither approach is "pure" - we're choosing which coupling is more acceptable.

## Comparison with #2433
- **#2433**: Guard clause approach
  - Result factory coupled to execution mode
  - Skips detection logic (early return for dry-run)
- **This PR**: Decorator pattern approach
  - Process layer coupled to test framework semantics
  - Exercises full detection logic

Both approaches work correctly. This PR prioritizes exercising real code paths (following the [Sculpture Principle](https://sanmai.github.io/SculpturePrinciple/)) over minimizing cross-layer coupling.

## Testing
- ✅ All unit tests pass
- ✅ `./build/infection.phar --dry-run --no-progress` works without errors
- ✅ Mutants correctly marked as ESCAPED
- ✅ Test uses `PASSING_TEST_OUTPUT` constant to verify the coupling

### DIFF
--- a/src/Process/DryRunProcess.php
+++ b/src/Process/DryRunProcess.php
@@ -42,9 +42,9 @@ use Symfony\Component\Process\Process;
  *
  * Wraps a real Process to simulate a terminated process for dry-run mode.
  *
- * The real process is constructed normally (ensuring command-line construction
- * is exercised) but never started. This wrapper presents it as already terminated
- * with passing test output, causing all mutants to be marked as ESCAPED.
+ * The real process is constructed normally but never started. This wrapper
+ * presents it as already terminated with passing test output, causing
+ * all mutants to be marked as ESCAPED.
  */
 final class DryRunProcess extends Process
 {
@@ -54,21 +54,9 @@ final class DryRunProcess extends Process
      */
     public const PASSING_TEST_OUTPUT = 'OK (0 tests, 0 assertions)';
 
-    private readonly string $commandLine;
-
-    /**
-     * Creates a minimal Process without a command (never started) and stores the real command line.
-     */
-    public function __construct(Process $realProcess)
+    public static function fromProcess(Process $process): self
     {
-        parent::__construct([]);
-
-        $this->commandLine = $realProcess->getCommandLine();
-    }
-
-    public function getCommandLine(): string
-    {
-        return $this->commandLine;
+        return self::fromShellCommandline($process->getCommandLine());
     }
 
     public function isTerminated(): bool
--- a/src/Process/Factory/MutantProcessContainerFactory.php
+++ b/src/Process/Factory/MutantProcessContainerFactory.php
@@ -84,7 +84,7 @@ class MutantProcessContainerFactory
         );
 
         if ($this->configuration->isDryRun()) {
-            $process = new DryRunProcess($process);
+            $process = DryRunProcess::fromProcess($process);
         }
 
         return new MutantProcessContainer(
--- a/tests/phpunit/Process/DryRunProcessTest.php
+++ b/tests/phpunit/Process/DryRunProcessTest.php
@@ -43,20 +43,10 @@ use Symfony\Component\Process\Process;
 #[CoversClass(DryRunProcess::class)]
 final class DryRunProcessTest extends TestCase
 {
-    public function test_it_returns_real_command_line_not_empty_parent_command(): void
-    {
-        $realProcess = new Process(['php', 'vendor/bin/phpunit', '--filter', 'SomeTest']);
-        $dryRunProcess = new DryRunProcess($realProcess);
-
-        // Verify we return the real command line, not an empty string from parent's []
-        $this->assertSame($realProcess->getCommandLine(), $dryRunProcess->getCommandLine());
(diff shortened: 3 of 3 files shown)


src/Process/DryRunProcess.php
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    public function getOutput(): string
+    {
+        // Returns output that TestFrameworkAdapter::testsPass() recognizes as passing tests.
+        // The pattern "OK (" triggers testsPass() to return true. Combined with exit code 0,
+        // this causes mutants to be marked as ESCAPED, which is correct for dry-run mode where
+        // tests are not actually executed.
+        return 'OK (0 tests, 0 assertions)';
+    }
```
sanmai, line 75: A classic abstraction leak here as well.

We could add a `Configuration` check to `TestFrameworkMutantExecutionResultFactory`'s `retrieveDetectionStatus()` instead but it will increase coupling. And it will be against the spirit of the current approach: we need dry-run to be a good smoke test, and here it exercises the actual detection logic, not just process creation.

theofidry: this could instead be passed as part of the constructor. We have access to the test framework where `DryProcess` is being instantiated. If the output needs to change for a test framework (not ideal in itself: that is quite brittle IMO), this could be moved to the test framework adapter.

tests/phpunit/Mutant/TestFrameworkMutantExecutionResultFactoryTest.php
```diff
@@ -321,7 +322,7 @@ public function test_it_can_crate_a_result_from_an_escaped_mutant_process(): voi
         ;
         $processMock
             ->method('getOutput')
-            ->willReturn('Tests passed!')
+            ->willReturn(DryRunProcess::PASSING_TEST_OUTPUT)
```
theofidry, line 325: this would also remove the need to make that constant public and could be an arbitrary string instead (so the previous `Test passed!` value could be preserved)

sanmai: We can move this constant to somewhere inside the test framework adapter; did you envision something else?

theofidry: does it need to be put anywhere yet?

```php
DryRunProcess:fromProcess(
  $process,
  // 'OK (0 tests, 0 assertions)', as the default value – can be overridden in the tests
);
```

the constant can then be `private`

sanmai: My concern here is that the our test passing detection may deviate from what the dry process produces and stop detecting it as "mutant not caught" which is the reason I made sure the test uses this constant; if we "untie" them, at some later day the dry run may stop working and we may not even know it is the case.

With the amount of stuff written by LLMs these day, through tests are as important as ever.

PR #2436, opened 2025-10-10
Update description of `--dry-run`

maks-rafalko:
Related to:

- https://github.com/infection/infection/pull/2435
- https://github.com/infection/site/pull/284

### DIFF
--- a/src/Command/RunCommand.php
+++ b/src/Command/RunCommand.php
@@ -132,6 +132,8 @@ final class RunCommand extends BaseCommand
 
     private const OPTION_LOGGER_HTML = 'logger-html';
 
+    private const OPTION_LOGGER_TEXT = 'logger-text';
+
     private const OPTION_USE_NOOP_MUTATORS = 'noop';
 
     private const OPTION_EXECUTE_ONLY_COVERING_TEST_CASES = 'only-covering-test-cases';
@@ -332,6 +334,12 @@ final class RunCommand extends BaseCommand
                 'Path to HTML report file, similar to PHPUnit HTML report.',
             )
             ->addOption(
+                self::OPTION_LOGGER_TEXT,
+                null,
+                InputOption::VALUE_REQUIRED,
+                'Path to text report file.',
+            )
+            ->addOption(
                 self::OPTION_USE_NOOP_MUTATORS,
                 null,
                 InputOption::VALUE_NONE,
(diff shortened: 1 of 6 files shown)


src/Command/RunCommand.php
```diff
@@ -396,7 +396,7 @@ protected function configure(): void
                 self::OPTION_DRY_RUN,
                 null,
                 InputOption::VALUE_NONE,
-                'Will not apply the mutations',
+                'Applies mutations in-memory and runs fake processes with 0 exit code.',
```
theofidry:
```suggestion
                'Do not evaluate the mutants; The processes are faked.',
```
?

I find "applies mutations in-memory" a bit confusing.

maks-rafalko: > I find "applies mutations in-memory" a bit confusing.

but this is what really happens

- "Will not apply the mutations"
- "Do not evaluate the mutants"

sounds for me like no mutations will be done. But the code is mutated (just not dumped), and tests are not executed.

staabm: tbh I do not have an idea what --dry-run is doing, with neither of your descriptions (and I really don't know what it does, so its hard for me to suggest a different text) :).

what are the implications of --dry-run ?

maks-rafalko: @staabm does this docs PR help with understanding? https://github.com/infection/site/pull/284/files

It's hard to explain it in one-line here. Or we can add a more verbose text similar to what docs PR has.

staabm: thanks.

I try to summarize my understanding.

```suggestion
                'Prints all mutations without running killer processes.',
```

maks-rafalko: ok, I like it more that previous options :)

But it doesn't print **all** (it depends on `--show-mutations=x`).

What about 

- `Applies all mutations and does not run killer processes.`

Is it clear enough?

staabm: > But it doesn't print all (it depends on `--show-mutations=x`.

ok, this means when not passing `--show-mutations=x` we still see the default list (20) of mutations.

so let me propose

- `Runs mutation testing without running killer processes.`

or

- `Runs mutation testing and does not run killer processes.`

I feel the word "applies" is hard to grasp as it does not really tell what 'apply' means

maks-rafalko:
```suggestion
                'Runs mutation testing and does not run killer processes..',
```

maks-rafalko: updated to `Runs mutation testing and does not run killer processes.` as suggested by Markus (for whom it's now clear what it does)

maks-rafalko: applied your suggestion, thanks

src/Command/RunCommand.php
```diff
@@ -396,7 +396,7 @@ protected function configure(): void
                 self::OPTION_DRY_RUN,
                 null,
                 InputOption::VALUE_NONE,
-                'Will not apply the mutations',
+                'Runs mutation testing and does not run killer processes..',
```
maks-rafalko:
```suggestion
                'Runs mutation testing and does not run killer processes.',
```

maks-rafalko:
thank you @theofidry and @staabm

src/Command/RunCommand.php
staabm: > I feel the word "applies" is hard to grasp as it does not really tell what 'apply' means

...

> applied your suggestion

just realized you were trolling my suggestion =)

maks-rafalko: hah, not really. But it would be funny

PR #2438, opened 2025-10-10
Introduce `--logger-text` option to enforce text logger file path

romm:
Allows the usage of `--logger-text` option when running Infection, to override the "logs.text" option from the configuration file.

This PR:

- [X] Adds new feature ...
- [X] Covered by tests
- [X] Doc PR: https://github.com/infection/site/pull/285

I'm coming from https://github.com/phpstan/phpstan-doctrine/pull/686/files#r2416912180 — I like to work with a text log file when running Infection locally, however it makes sense to use `php://stdout` when running it in the pipeline. Adding this console option allows to override the option set in the configuration file.

I did not find an existing test for the `--logger-html` option, to create a similar test for this newly added option. What should I do?

If the PR is ok for you, I'll add a PR to update the documentation.

### DIFF
--- a/src/Command/RunCommand.php
+++ b/src/Command/RunCommand.php
@@ -132,6 +132,8 @@ final class RunCommand extends BaseCommand
 
     private const OPTION_LOGGER_HTML = 'logger-html';
 
+    private const OPTION_LOGGER_TEXT = 'logger-text';
+
     private const OPTION_USE_NOOP_MUTATORS = 'noop';
 
     private const OPTION_EXECUTE_ONLY_COVERING_TEST_CASES = 'only-covering-test-cases';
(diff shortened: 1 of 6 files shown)


maks-rafalko:
> I did not find an existing test for the `--logger-html` option, to create a similar test for this newly added option. What should I do?

It's tested here I guess:

https://github.com/infection/infection/blob/242785d48ac2dc00a1d3a77b2048b289dc82cbc9/tests/phpunit/Configuration/ConfigurationFactoryTest.php#L355-L389

You can create something similar. It's one of the hardest tests in Infection so sorry in advance 🤣 

> If the PR is ok for you, I'll add a PR to update the documentation.

absolutely makes sense

Regarding 100% MSI required here for PRs, feel free to skip this if it's too hard, we will merge anyway (I don't want to force you writing tests for the places covered badly previously).

romm:
Thanks for the hint! I shamefully copy-paster these tests, but that should be enough. 😊

maks-rafalko:
could you please fix https://github.com/infection/infection/actions/runs/18403064509/job/52436622828?pr=2438#step:7:48? And we are good to go

maks-rafalko:
Thank you @romm 👍

maks-rafalko:
Released https://github.com/infection/infection/releases/tag/0.31.6

romm:
That was fast! Thank you 😊

Issue #2442, opened 2025-10-10
show uncovered mutants in the output


staabm:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.31.6 (`infection.phar --version`)
| Test Framework version | PHPUnit
| PHP version | 8.3.26 (cli) (built: Sep 23 2025 17:57:26) (NTS)
| Platform    | MacOS
| Github Repo | https://github.com/staabm/infection-example/tree/124b3e3667eac026ea61267c44294030aca03480

- checkout repo
- run `composer install`

- apply diff
```diff
diff --git a/src/User.php b/src/User.php
index e1b1541..f5550c4 100644
--- a/src/User.php
+++ b/src/User.php
@@ -6,4 +6,8 @@ class User {
     public function __construct(int $age) {
         $this->age = $age;
     }
+
+    public function getName(): string {
+        return "John Doe";
+    }
 }
```

- run `vendor/bin/infection --log-verbosity=all --debug --with-uncovered`

I would expect in the diff output of the cli command, that uncovered mutants would show up.
only escaped ones are showing up though.

looking into the infection.log I can see a uncovered one though:

```

Not Covered mutants:
====================

1) /Users/staabm/workspace/infection-example/src/User.php:10    [M] PublicVisibility [ID] 7979b0a4066b5159a508637988338a07

@@ @@
     {
         $this->age = $age;
     }
-    public function getName(): string
+    protected function getName(): string
     {
         return "John Doe";
     }
 }
```

PR #2443, opened 2025-10-11
Show uncovered mutants to output when `--with-uncovered` is used

maks-rafalko:
Fixes https://github.com/infection/infection/issues/2442

There are many issues with our loggers and collectors that need to be addressed separately, but the main confusion was: we have many different things affecting what mutants needs to be shown in logs/console:

- `--logger-*` options.  (also affect collectors)
- `--log-verbosity=*`  (also affect collectors)
- `-v|-vv`
- `--debug`
- `--show-mutations=x` (also affect collectors)
- `--with-uncovered` (also affect collectors)

That is very confusing.

Issue number one with  #2442 that decision to show or not to show uncovered mutants was made by `-v` - this has been changed and now controller by `--with-uncovered` option. Additionally, if `--show-mutations=0` (disabled) - no uncovered mutants will be shown, so it's clear now.

Issue number two is that we have a separate class that controls what statuses are being collected: `TargetDetectionStatusesProvider`, and it's very confusing that it affects different parts of the system. We do have a TODO (related to https://github.com/infection/infection/pull/1430) to improve it, but this is out of the scope of this bugfix:

https://github.com/infection/infection/blob/9965d1f6e372bdca7992d7b2e9526e77765b27ff/src/Metrics/TargetDetectionStatusesProvider.php#L73

So, now it works this way:

- `bin/infection` - uncovered are not shown
- `bin/infection --with-uncovered` - uncovered are shown
- `bin/infection --with-uncovered --show-mutations=0` - uncovered are not shown

### DIFF
--- a/src/Event/Subscriber/MutationTestingConsoleLoggerSubscriberFactory.php
+++ b/src/Event/Subscriber/MutationTestingConsoleLoggerSubscriberFactory.php
@@ -54,7 +54,7 @@ final readonly class MutationTestingConsoleLoggerSubscriberFactory implements Su
         private FederatedLogger $mutationTestingResultsLogger,
         private ?int $numberOfShownMutations,
         private OutputFormatter $formatter,
-        private bool $showMutationScoreIndicator,
+        private bool $withUncovered,
     ) {
     }
 
@@ -68,7 +68,7 @@ final readonly class MutationTestingConsoleLoggerSubscriberFactory implements Su
             $this->diffColorizer,
             $this->mutationTestingResultsLogger,
             $this->numberOfShownMutations,
-            $this->showMutationScoreIndicator,
+            $this->withUncovered,
         );
     }
 }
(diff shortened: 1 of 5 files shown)


src/Event/Subscriber/MutationTestingConsoleLoggerSubscriberFactory.php
```diff
@@ -54,7 +54,7 @@ public function __construct(
         private FederatedLogger $mutationTestingResultsLogger,
         private ?int $numberOfShownMutations,
         private OutputFormatter $formatter,
-        private bool $showMutationScoreIndicator,
+        private bool $withUncovered,
```
sanmai, line 57: Constructed here: 

https://github.com/infection/infection/blob/24011fa795e12f843a72f586081b26340f0a4fe3/src/Container.php#L422-L430

staabm:
not to blame anyone, just as a note: it seems this bug was introduced by a KI based PR in https://github.com/infection/infection/pull/2378

maks-rafalko:
> not to blame anyone, just as a note: it seems this bug was introduced by a KI based PR in #2378

no, this is actually "an old" bug that we discovered just now. This doesn't remove the fact that AI also introduced a bug in other place recently :) but it's ok

thanks all for the review

PR #2444, opened 2025-10-11
test: Add integration test for coverage trace provider

theofidry:
## Goal

The goal of this PR is to bring light about what the final result is, i.e. for a given coverage report, what traces are we getting and what those traces actually are.

This will bring the following benefits:
- Have a clear picture of what `Trace` objects we are consuming look like.
- Ensure we do not have any regressions from reworking any of the underlying code.

## Context

To generate a `Trace`, a lot of works needs to be done. We use about 11 services, wired a special way, in order to navigate through the various coverage reports and construct the `Trace` objects in an efficient way. I believe there is room for improvements there, I have a plan for it, but this is not the scope of this PR.

While doing that work, I was struggling a lot with figuring out how a `Trace` actually looks like in its final form, for a given coverage report. We have a lot of unit tests for the different steps, but none with a clear big picture.

Anyway, this is what brought me here, but I believe this change on its own makes sense even if we do not move forward with [my upstream plan](https://github.com/theofidry/infection/pull/1).

## How it works

To provide real results, a mini-project is shipped for which we can add source code and tests and generate a coverage report. Since coverage reports include absolute path, we cannot commit it. Instead, I went with the solution of copying it and replace the absolute paths by a placeholder and commit this instead. Before executing the tests, we copy the report from the template if it does not exist.

From there we create our service and generate the traces and compare the traces we want.

This test is heavy:
- The expectation list is long: I think it is fine, we can't do much to avoid this and this is meant as a smoke integration test, i.e. not a test we want to update all the time.
- It's a lot of services to instantiate: it is just how it is design-wise at the moment.

### DIFF
--- a/tests/phpunit/TestFramework/Tracing/PHPUnitCoverageTracerTest.php
+++ b/tests/phpunit/TestFramework/Tracing/PHPUnitCoverageTracerTest.php
@@ -240,7 +240,7 @@ final class PHPUnitCoverageTracerTest extends TestCase
             $splFileInfo,
             new SyntheticTrace(
                 sourceFileInfo: $splFileInfo,
-                realPath: realpath($canonicalDemoCounterServicePathnam),
+                realPath: realpath($canonicalDemoCounterServicePathname),
                 relativePathname: $canonicalDemoCounterServicePathname,
                 hasTest: true,
                 tests: new TestLocations(


tests/phpunit/TestFramework/Tracing/SyntheticTrace.php
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Infection\TestFramework\Tracing;
+
+use DomainException;
+use Infection\TestFramework\Coverage\NodeLineRangeData;
+use Infection\TestFramework\Coverage\TestLocations;
+use Infection\TestFramework\Coverage\Trace;
+use Symfony\Component\Finder\SplFileInfo;
+
+final readonly class SyntheticTrace implements Trace
```
theofidry, line 49: This trace is so far purely for testing purposes. It is used to declare a trace state, i.e. the expectations of the generate trace state, hence there is no need for any laziness and no value is calculated/inferred.

theofidry: May be worth en explanatory comment and maybe move it to the test namespace

tests/phpunit/TestFramework/Tracing/Fixtures/README.md
```diff
@@ -0,0 +1,3 @@
+To keep things simple but true, a dummy service was implemented and an AI
```
theofidry: review that once the test generation is figured out (problems are listed bellow)

tests/phpunit/TestFramework/Tracing/Fixtures/phpunit/junit.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<testsuites>
+  <testsuite name="/Users/tfidry/Project/Humbug/infection/tests/phpunit/TestFramework/Tracing/Fixtures/phpunit.xml" tests="12" assertions="25" errors="0" failures="0" skipped="0" time="0.021691">
```
theofidry: the JUnit report contains absolute paths which is not ideal, but we cannot regenerate it either since the execution time is important. Maybe we need to play around with placeholders, i.e. having:

```suggestion
  <testsuite name="/path/to/project/tests/phpunit/TestFramework/Tracing/Fixtures/phpunit.xml" tests="12" assertions="25" errors="0" failures="0" skipped="0" time="0.021691">
```

And for the test execution replace `/path/to/project` with the actual value.

tests/phpunit/TestFramework/Tracing/Fixtures/phpunit.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
theofidry, line 2: the idea is to have a "nano project" for which we could easily generate the real code coverage at any time:

```
XDEBUG_MODE=coverage vendor/bin/phpunit --configuration=/path/to/this/phpunit.xml
```

Having this mini project has several benefits IMO. Currently most of our services work with hand crafted samples of XML reports. This is great to capture various edge cases and keep things readable. However, this also means we can easily miss changes in code coverage report (i.e. PHPUnit updates something, and we don't pick up on this change).

I am not challenging our current approach, but I also believe having _some_ integration tests would be helpful. In order to do that, this small project will allow us to create real code, real test, and generate an up-to-date coverage report from them.

That said, maybe the current location is too specific or too hidden and we may want to move that somewhere else.

tests/phpunit/TestFramework/Tracing/TraceAssertion.php
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Infection\Tests\TestFramework\Tracing;
+
+use Infection\TestFramework\Coverage\Trace;
+use PHPUnit\Framework\Assert;
+
+final class TraceAssertion
```
theofidry, line 41: Due to the laziness of some traces, and the fact that `Trace` is an interface, we cannot simply compare objects between each other, hence this custom assertion.

tests/phpunit/TestFramework/Tracing/PHPUnitCoverageTracerTest.php
```diff
@@ -0,0 +1,437 @@
...
+use PHPUnit\Framework\Attributes\CoversNothing;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\TestCase;
+use function sprintf;
+use Symfony\Component\Filesystem\Path;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\SplFileInfo;
+
+#[CoversNothing]
+final class PHPUnitCoverageTracerTest extends TestCase
```
theofidry, line 70: this is the star of the PR. Don't miss it, GitHub likes to hide it

tests/phpunit/TestFramework/Tracing/PHPUnitCoverageTracerTest.php
```diff
@@ -0,0 +1,437 @@
...
+    protected function setUp(): void
+    {
+        $coveragePath = self::FIXTURE_DIR . '/phpunit';
+
+        $testFrameworkAdapterStub = $this->createStub(TestFrameworkAdapter::class);
+        $testFrameworkAdapterStub
+            ->method('hasJUnitReport')
+            ->willReturn(true);
+
+        $this->provider = new CoveredTraceProvider(
```
theofidry, line 87: FWIW, I know this looks awful, we are wiring 11 services, but I believe this can be greatly improved. But I think even without those improvements this test still makes sense and provides value

tests/phpunit/TestFramework/Tracing/PHPUnitCoverageTracerTest.php
```diff
@@ -0,0 +1,437 @@
...
+            ),
+        );
+    }
+
+    #[DataProvider('traceProvider')]
+    public function test_it_can_create_a_trace(
+        SplFileInfo $fileInfo,
+        Trace $expected,
+    ): void {
+        $actual = 'No Trace found.';
```
theofidry: let's just remove this and properly fail if the `$actual` is not defined.

tests/phpunit/TestFramework/Tracing/PHPUnitCoverageTracerTest.php
```diff
@@ -0,0 +1,437 @@
...
+
+        foreach ($this->provider->provideTraces() as $trace) {
```
theofidry, line 129: this is a bout of a roundabout way to do things, this is because this test focuses on what a trace should look like for a given file. So if we add more covered code, we _don't have to_ necessarily update this test.

tests/phpunit/TestFramework/Tracing/SyntheticTrace.php
maks-rafalko: yeah, comment and/or moving would improve the understanding in the futere

tests/phpunit/TestFramework/Tracing/Fixtures/phpunit/junit.xml
maks-rafalko: voting for "placeholder" with replacing, otherwise it will bring more noise in diffs during re-running on different machine, if I understood it correctly

Issue #2425, opened 2025-10-06
Configurable Output Format: files or stdout / stderr


xepozz:
Infection provides `logger-gitlab` CLI flag which specifies output file for one of json formats (gitlab)
But there's no flags to get the json directly after calling the command.

It would be super convenient to have any of these formats as an stdout result.
I'm playing with plugin for Intellij IDEA for Infection and now I'll put anything into a temporary file which is passed to the `logger-gitlab` flag.

maks-rafalko:
Hey, you can use

- `infection --logger-gitlab=php://stdout` 
- `infection --logger-gitlab=php://stderr`

It's documented here: https://infection.github.io/guide/usage.html

Would it work for you?



> I'm playing with plugin for Intellij IDEA for Infection

would be happy to see/test the results if you will end up with a working plugin eventually

maks-rafalko:
closing, feel free to reopen if it doesn't work for you

xepozz:
> * `infection --logger-gitlab=php://stdout`
> * `infection --logger-gitlab=php://stderr`

Nice. Thought it could work, but preferred to ask.

> would be happy to see/test the results if you will end up with a working plugin eventually

First beta is here: https://plugins.jetbrains.com/plugin/28650-infection

It works as phpstan/psalm/mago/etc: when you open a file it runs tests, collects results and annotate editor elements. Currently, only diff preview is available.
Later I'll rework the plugin and make it work as phpunit/codeception plugins does:
- Running and actions as "... with Coverage"
- Filling code coverage panel
- Annotating elements as now

xepozz:
I'd ask you to add in the output line ending / char ending position of the transforming AST node.
I understand that it doesn't mean any whitespaces, but it would be convenient to highlight only changed nodes instead of a full line of even line with partial content.

By the last I mean this snippet

<img width="389" height="234" alt="Image" src="https://github.com/user-attachments/assets/f65f7854-ad9d-4d47-81c7-5d5326e9a7e1" />

In the results I have only line `64`
<img width="1035" height="209" alt="Image" src="https://github.com/user-attachments/assets/24f59dfa-2eeb-45d4-99ed-bf7836ca8311" />

maks-rafalko:
We are using php-parser lib and creating diff between **already pretty printed file** and **mutated file**, like this

https://github.com/infection/infection/blob/38991a32f17608df97848403f23921431c9a3842/src/Mutant/MutantFactory.php#L80

[...]

PR #2448, opened 2025-10-14
Format-preserving pretty-printing for Mutants

maks-rafalko:
Format-preserving pretty printing is hard :| After a week of evenings of debugging, here we go.

https://github.com/nikic/PHP-Parser/blob/master/doc/component/Pretty_printing.markdown#formatting-preserving-pretty-printing

Why is this needed?

- To correctly highlight mutated code in PhpStorm Plugin (requested here: https://github.com/infection/infection/issues/2425#issuecomment-3377140398)
- To have original code in code diffs for each Mutant. Previously, the code was reformatted by PrettyPrinter from `php-parser`

Instead of 1000 words, examples:

Before:

```diff
@@ @@
     }
     public function logVerbosityDeprecationNotice(string $valueToUse): void
     {
-        $this->logger->notice('Numeric versions of log-verbosity have been deprecated, please use, ' . $valueToUse . ' to keep the same result', ['block' => true]);
+        $this->logger->notice('Numeric versions of log-verbosity have been deprecated, please use, ' . $valueToUse . ' to keep the same result', []);
     }
     public function logUnknownVerbosityOption(string $default): void
     {
```

After:

```diff
@@ @@
     {
         $this->logger->notice(
             'Numeric versions of log-verbosity have been deprecated, please use, ' . $valueToUse . ' to keep the same result',
-            ['block' => true],
+            [],
         );
     }
```

As you can see, previously Infection **modified** the source code formatting, then created a Mutant, and then compared modified original code with mutant.

Now, original code is compared with a mutant as is - we ONLY change the mutated node. No more reformatting.

This gives us much more readable and clear diffs. Long multiline arrays are no longer changed to lengthy one-liners.

----

During implementation, I found several bugs in `php-parser`:

- https://github.com/nikic/PHP-Parser/pull/1115
- https://github.com/nikic/PHP-Parser/pull/1116
- https://github.com/nikic/PHP-Parser/issues/1117
- https://github.com/nikic/PHP-Parser/issues/1119

These are not blockers, but interesting to see the answers from Nikita. I have a feeling that either these are bugs or we don't use nodes replacements correctly (probably we should do them in-place) - let's see.

----

- [x] check how HTML report works after these changes

### DIFF
--- a/tests/phpunit/Mutation/FileMutationGeneratorTest.php
+++ b/tests/phpunit/Mutation/FileMutationGeneratorTest.php
@@ -395,15 +395,7 @@ final class FileMutationGeneratorTest extends TestCase
         string $relativePathname,
         ?bool $hasTests = null,
     ): Trace {
-        $splFileInfoMock = $this->createMock(SplFileInfo::class);
-        $splFileInfoMock->method('__toString')->willReturn($file);
-        $splFileInfoMock->method('getFilename')->willReturn($file);
-        $splFileInfoMock->method('getRealPath')->willReturn($file);
-        $splFileInfoMock->method('getContents')->willReturn(
-            file_exists($file) ? file_get_contents($file) : 'content',
-        );
-        $splFileInfoMock->method('getRelativePath')->willReturn($relativePath);
-        $splFileInfoMock->method('getRelativePathname')->willReturn($relativePathname);
+        $splFileInfoMock = $this->createSplFileInfoMock($file, $relativePath, $relativePathname);
 
         $proxyTraceMock = $this->createMock(Trace::class);
         $proxyTraceMock
@@ -419,5 +411,20 @@ final class FileMutationGeneratorTest extends TestCase
         }
 
         return $proxyTraceMock;
+    }
+
+    private function createSplFileInfoMock(string $file, string $relativePath, string $relativePathname): SplFileInfo&MockObject
+    {
+        $splFileInfoMock = $this->createMock(SplFileInfo::class);
+        $splFileInfoMock->method('__toString')->willReturn($file);
+        $splFileInfoMock->method('getFilename')->willReturn($file);
+        $splFileInfoMock->method('getRealPath')->willReturn($file);
+        $splFileInfoMock->method('getContents')->willReturn(
+            file_exists($file) ? file_get_contents($file) : 'content',
+        );
+        $splFileInfoMock->method('getRelativePath')->willReturn($relativePath);
+        $splFileInfoMock->method('getRelativePathname')->willReturn($relativePathname);
+
+        return $splFileInfoMock;
     }
 }


src/Mutator/Arithmetic/Assignment.php
```diff
@@ -74,7 +76,11 @@ public static function getDefinition(): Definition
      */
     public function mutate(Node $node): iterable
     {
-        yield new Node\Expr\Assign($node->var, $node->expr, $node->getAttributes());
+        yield new Node\Expr\Assign(
+            $node->var,
+            $node->expr,
+            array_diff_key($node->getAttributes(), array_flip(['origNode'])),
```
maks-rafalko: for me: simplify and move to some static method call? This is repeated almost for every mutator, except `PublicVisibility` / `ProtectedVisibility` that don't require it

src/Mutant/MutantFactory.php
```diff
@@ -71,7 +63,7 @@ public function create(Mutation $mutation): Mutant
         );
 
         $mutatedCode = lazy($this->createMutatedCode($mutation));
-        $originalPrettyPrintedFile = lazy($this->getOriginalPrettyPrintedFile($mutation->getOriginalFilePath(), $mutation->getOriginalFileAst()));
+        $originalPrettyPrintedFile = lazy((static fn () => yield $mutation->getOriginalFileContent())());
```
sanmai:
```suggestion
        $originalPrettyPrintedFile = later(static fn () => yield $mutation->getOriginalFileContent());
```

maks-rafalko: thanks, fixed

src/Mutator/Arithmetic/Assignment.php
theofidry: Given how it is used everywhere, maybe we can use a utility there? `NodeAttributes::getAllExceptOriginalNode()`?

tests/phpunit/Mutation/FileMutationGeneratorTest.php
```diff
@@ -391,10 +395,20 @@ private function createTraceMock(
         string $relativePathname,
         ?bool $hasTests = null,
     ): Trace {
+        $splFileInfoMock = $this->createMock(SplFileInfo::class);
```
theofidry: I don't get why this is necessary?

maks-rafalko:
> So how do we test the result for a specific mutator? When replacing `-` with `+` there is no change of character/position so I guess it's an easy scenario, but if there is bigger transformations then how does it plays out and how can we see that?

that's why I've created https://github.com/infection/infection/pull/2450 - to be able to test exactly that formatting that happens for Mutants. Currently, different prinetrs are used so it's impossible.

But I will do it in separate PR what will target this one, as it's quite big

src/Mutator/Arithmetic/Assignment.php
maks-rafalko: that's what I've already planned here: https://github.com/infection/infection/pull/2448#discussion_r2430331567. Yes, I will create a static class/method for it

tests/phpunit/Mutation/FileMutationGeneratorTest.php
maks-rafalko: because now `FileMutationGenerator` calls `$sourceFile->getContents()` and for all the fake paths like `/path/to/file` is started failing.

On the other side, there are other cases which relies on the real file's content from the fixtures, so this mock supports both

theofidry: Could be worth to have a dedicated `DummySplFileInfo` instead of a mock?

maks-rafalko: could you please explain how would it work?

maks-rafalko:
> So how do we test the result for a specific mutator? When replacing `-` with `+` there is no change of character/position so I guess it's an easy scenario, but if there is bigger transformations then how does it plays out and how can we see that?

I've spent much more time on it and updated the tests to use format-preserving pretty printer. 

Many of the tests have been updated (as they obviously failed), unfortunately it's impossible to move it to a separate PR, so please take a look.

I've found already ~4 potential bugs in `php-parser` (mentioned them in the PR description), but none of them are blockers, yet annoying. You will find workarounds for them in the diff.

src/Mutation/Mutation.php
```diff
@@ -79,6 +81,8 @@ public function __construct(
         private readonly MutatedNode $mutatedNode,
         private readonly int $mutationByMutatorIndex,
         private readonly array $tests,
+        private readonly array $originalFileTokens,
```
theofidry, line 82: since those are readonly maybe they could be public?

src/Mutator/Boolean/LogicalAndNegation.php
```diff
@@ -76,7 +76,14 @@ public static function getDefinition(): Definition
      */
     public function mutate(Node $node): iterable
     {
-        yield new Node\Expr\BooleanNot($node);
+        // Clone the node to remove the origNode from the wrapped expression
+        // see bug https://github.com/nikic/PHP-Parser/issues/1119
+        $wrappedNode = clone $node;
+        $wrappedAttrs = $wrappedNode->getAttributes();
+        unset($wrappedAttrs['origNode']);
+        $wrappedNode->setAttributes($wrappedAttrs);
```
theofidry:
```suggestion
        $wrappedNode->setAttributes(NodeAttributes::getAllExceptOriginalNode($node));
```

src/Mutator/Boolean/LogicalOrNegation.php
```diff
@@ -76,7 +77,17 @@ public static function getDefinition(): Definition
      */
     public function mutate(Node $node): iterable
     {
-        yield new Node\Expr\BooleanNot($node);
+        // Clone the node to remove the origNode from the wrapped expression
+        // see bug https://github.com/nikic/PHP-Parser/issues/1119
+        $wrappedNode = clone $node;
+        $wrappedAttrs = $wrappedNode->getAttributes();
+        unset($wrappedAttrs['origNode']);
+        $wrappedNode->setAttributes($wrappedAttrs);
```
theofidry: likewise here (maybe a few other places too)

tests/phpunit/Mutator/NodeAttributesTest.php
```diff
@@ -0,0 +1,102 @@
...
+
+use Infection\Mutator\NodeAttributes;
+use PhpParser\Node\Scalar\Int_;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\TestCase;
+
+#[CoversClass(NodeAttributes::class)]
+final class NodeAttributesTest extends TestCase
+{
+    public function test_it_returns_all_attributes_except_original_node(): void
```
theofidry: I find this test a bit boilertemplaty, this could be made simpler with a data provider:

```php
/**
     * @param array<string, mixed> $expected
     */
    #[DataProvider('nodeProvider')]
    public function test_it_gets_all_attributes_except_the_original_node(
        Node $node,
        array $expected,
    ): void
    {
        $actual = NodeAttributes::getAllExceptOriginalNode($node);

        self::assertSame($expected, $actual);
    }

    public static function nodeProvider(): iterable
    {
        yield 'nominal' => (static function () {
            $attributes = [
                'startLine' => 1,
                'endLine' => 2,
                'startFilePos' => 3,
                'endFilePos' => 4,
                'customAttribute' => 'value',
            ];

            $node = new Int_(42);
            $node->setAttributes($attributes);
            $node->setAttribute('origNode', $node);

            return [$node, $attributes];
        })();

        yield 'no attributes' => [
            new Int_(42),
            [],
        ];

        yield 'all attributes when original node is not present' => [
            new Int_(
                42,
                [
                    'startLine' => 1,
                    'endLine' => 2,
                ],
            ),
            [
                'startLine' => 1,
                'endLine' => 2,
            ],
        ];
    }
```

src/PhpParser/FileParser.php
```diff
@@ -69,4 +70,14 @@ public function parse(SplFileInfo $fileInfo): array
             throw UnparsableFile::fromInvalidFile($filePath, $throwable);
         }
     }
+
+    /**
+     * Return tokens for the last parse.
+     *
+     * @return Token[]
+     */
+    public function getTokens(): array
```
theofidry, line 79: I'm not a fan of this API. It fits in the current usage but it lacks any cohesiveness. Can't we change `FileParser` to return a tuple instead?

```suggestion
    /**
     * @throws UnparsableFile
     *
     * @return array{Stmt[], Token[]}
     */
    public function parse(SplFileInfo $fileInfo): array
    {
        try {
            $statements = $this->parser->parse($fileInfo->getContents()) ?? [];
            $tokens = $this->parser->getTokens();

            return [$statements, $tokens];
        } catch (Throwable $throwable) {
            $filePath = $fileInfo->getRealPath() === false
                ? $fileInfo->getPathname()
                : $fileInfo->getRealPath()
            ;

            throw UnparsableFile::fromInvalidFile($filePath, $throwable);
        }
    }
```

I appreciate it has a greater impact though

tests/phpunit/Mutation/FileMutationGeneratorTest.php
theofidry: Something like this:

<details>
<summary>DummySplFileInfo.php</summary>

```php
<?php

declare(strict_types=1);

namespace Infection\Tests\FileSystem;

use Symfony\Component\Finder\SplFileInfo as SymfonyFinderSplFileInfo;

final class DummySplFileInfo extends SymfonyFinderSplFileInfo
{
    public function __construct(
        private readonly string $filename,
        private readonly string $realPath,
        private readonly string $relativePath,
        private readonly string $relativePathname,
        private readonly string $contents,
        private readonly string $toString,
    )
    {
        parent::__construct(__FILE__, $relativePath, $relativePathname);
    }

    public static function forFile(
        string $file,
        string $relativePath,
        string $relativePathname,
        string $contents,
    ): self {
        return new self(
            filename: $file,
            realPath: $file,
            relativePath: $relativePath,
            relativePathname: $relativePathname,
            contents: $contents,
            toString: $file,
        );
    }

    public function getFilename(): string
    {
        return $this->filename;
    }

    public function getRealPath(): string
    {
        return $this->realPath;
    }

    public function getRelativePath(): string
    {
        return $this->relativePath;
    }

    public function getRelativePathname(): string
    {
        return $this->relativePathname;
    }

    public function getContents(): string
    {
        return $this->contents;
    }

    public function __toString(): string
    {
        return $this->toString;
    }
}

```
</details>

Then your current code becomes:

```php
$splFileInfo = DummySplFileInfo::forFile(
            $file,
            $relativePath,
            $relativePathname,
            file_exists($file) ? file_get_contents($file) : 'content',
        );
```

And it could also be used in other places; I don't have any atm, but maybe it is a case we'll come across and a scenario we avoided in the past because it was too annying.

theofidry: Alternatively I would at least move that piece of code into a factory `createSplFileInfoMock()`

theofidry (approved):

src/Mutation/Mutation.php
maks-rafalko: I decided not to do it, because it won't be consistent. We can translate private readonly properties to public as a separate PR using Rector I guess on the project level

src/PhpParser/FileParser.php
maks-rafalko: Decided not to do it because `->parse($file)` and `->getTokens()` this is exactly the API that `php-parser` exposes. Yes, it's arguable, but we only have a wrapper.

Such change will affect many files and IMO if we really want it (I have doubts) it shouldn't be done here.

tests/phpunit/Mutation/FileMutationGeneratorTest.php
maks-rafalko: moved to a separate factory method

> $splFileInfo = DummySplFileInfo::forFile(

tbh, I don't see a benefit of extracting it to another class at the current stage. If there are - please outline them

theofidry: I personally find it more expressive and removes the reliance on the mocks, but at this stage it's more a matter of preference than practicality.

src/PhpParser/FileParser.php
theofidry: It's already questionable in `PHP-Parser` and the advantage of having a wrapper is to be able to correct such API issues.

>Such change will affect many files

Does it? From what I saw it only affects 3 files which you are already touching. That said this could be done in a separate PR.

In the end I don't think it's a fundamental change so I won't insist more than that, but it would be nicer to not introduce loose APIs.

PR #2449, opened 2025-10-16
test: Cosmetic changes to SourceFileCollectorTest

theofidry:


tests/phpunit/FileSystem/SourceFileCollector/SourceFileCollectorTest.php
```diff
@@ -36,32 +36,38 @@
...
+    ): void {
+        $files = (new SourceFileCollector())->collectFiles(
+            $sourceDirectories,
+            $excludedFilesOrDirectories,
+        );
 
         self::assertIsEqualCanonicalizing(
             $expectedList,
-            take($files)->toAssoc(),
+            iterator_to_array($files),
```
sanmai: When I see `iterator_to_array()` I have to recall or look up what the the default behavior, so it is almost always a waste of a reader's time; with `take($files)->toAssoc()` there's no such problem - it does what it says. No second guessing, no extra time spent looking up what is what.

Further, calling `iterator_to_array` the wrong way could lead to very confusing bugs, so it is very important that the second argument of  `iterator_to_array` has the correct value. Once again, with `take($files)->toAssoc()` or `take($files)->toList()` there's not confusion whatsoever.

theofidry: Tbh it's a case where I really prefer iter's `toArrayWithKeys()` which is both more describe and direct. But I also don't really to introduce yet another lib for this sort of things

PR #2450, opened 2025-10-16
Use the same pretty-printer for source code and tests

maks-rafalko:
This is a preparing step for https://github.com/infection/infection/pull/2448

We must be sure that the printer we use is exactly the same in the source code (for generating Mutants) **and in the tests**.

Otherwise, we can't test what we really generate as a pretty-printed code during MT.

Format-preserving pretty-printing is done in https://github.com/infection/infection/pull/2448 and more tests for preserving format will be done in further PRs.

For now - just the use of the same printer (that under the hood will use format-preserving strategy in https://github.com/infection/infection/pull/2448)

### DIFF
--- a/tests/phpunit/Mutant/MutantCodePrinterTest.php
+++ b/tests/phpunit/Mutant/MutantCodePrinterTest.php
@@ -90,9 +90,9 @@ final class MutantCodePrinterTest extends TestCase
         $mutation = new Mutation(
             '/path/to/acme/Foo.php',
             $statements,
-            Plus::class,
-            MutatorName::getName(Plus::class),
-            [
+            mutatorClass: Plus::class,
+            mutatorName: MutatorName::getName(Plus::class),
+            attributes: [
                 'startLine' => 5,
                 'startTokenPos' => 9,
                 'startFilePos' => 29,
@@ -101,8 +101,8 @@ final class MutantCodePrinterTest extends TestCase
                 'endFilePos' => 30,
                 'kind' => 10,
             ],
-            Node\Scalar\Int_::class,
-            MutatedNode::wrap(
+            mutatedNodeClass: Node\Scalar\Int_::class,
+            mutatedNode: MutatedNode::wrap(
                 new Node\Scalar\Int_(
                     15,
                     [
@@ -116,8 +116,8 @@ final class MutantCodePrinterTest extends TestCase
                     ],
                 ),
             ),
-            0,
-            [],
+            mutationByMutatorIndex: 0,
+            tests: [],
         );
 
         yield 'basic namespace with echo' => [


src/Mutant/MutantCodeFactory.php
```diff
@@ -61,6 +60,6 @@ public function createCode(Mutation $mutation): string
 
         $mutatedStatements = $traverser->traverse($mutation->getOriginalFileAst());
 
-        return $this->printer->prettyPrintFile($mutatedStatements);
+        return $this->mutatedCodePrinter->print($mutatedStatements, $mutation);
```
maks-rafalko, line 63: purpose of the PR: we use our own printer wrapper **here** and **in the tests**.

In the next PRs it will be converted to format-preserving pretty-printer under the hood, which will affect both tests and the source code (this is the goal)

src/Mutant/MutantCodePrinter.php
```diff
@@ -0,0 +1,60 @@
...
+{
+    public function __construct(
+        private readonly PrettyPrinterAbstract $printer,
+    ) {
+    }
+
+    /**
+     * @param Node[] $mutatedStatements
+     */
+    public function print(array $mutatedStatements, Mutation $mutation): string
```
theofidry, line 56: what will it need the mutation object for?

src/Mutant/MutatedCodePrinter.php
```diff
@@ -0,0 +1,60 @@
...
+
+use Infection\Mutation\Mutation;
+use PhpParser\Node;
+use PhpParser\PrettyPrinterAbstract;
+
+/**
+ * @internal
+ * @final
+ */
+class MutatedCodePrinter
```
theofidry:
```suggestion
class MutantCodePrinter
```
? we have `MutantCodeFactory` so it seems a bit more consistent to me. FWIW "mutant" also signals "changed code" although I don't think that subtlety makes much of a difference

tests/phpunit/Mutant/MutatedCodePrinterTest.php
```diff
@@ -0,0 +1,129 @@
...
+                        'startFilePos' => 29,
+                        'endLine' => 5,
+                        'endTokenPos' => 9,
+                        'endFilePos' => 30,
+                        'kind' => 10,
+                    ],
+                ),
+            ),
+            0,
+            [],
```
theofidry: maybe named parameters would be nice here?

theofidry (approved):

src/Mutant/MutantCodePrinter.php
maks-rafalko: this will be needed for building format-preserving pretty printing in #2448. I've added this in advance to reduce to diff later ;)

PR #2452, opened 2025-10-18
perf: Fix the trace generation profiling

theofidry:
Currently `make profile` does not work. Eventually I got it to generate profiles, but it gave very... empty results. Turns out that no trace was actually generated at all.

There is still much to do here, I'll split in several PRs and add explanations. But this is a draft where I got it working with the following as an example of profile: https://blackfire.io/profiles/9bfe5382-9d80-465e-9b62-31d0f7496c5f/graph

TODOs:

- [x] Extract the splitting of the profile command (https://github.com/infection/infection/pull/2454)
- [x] Extract the source collection fix (https://github.com/infection/infection/pull/2455)
- [x] Extract an instrumentor refactoring (#2460)
- [x] ~~Extract the simplification of the profile script (removing the mutation/trace count)~~
- [x] Extract the fix of the Blackfire deprecated option `--samples` (https://github.com/infection/infection/pull/2517)
- [x] Fix the mutation generation profiling
- [x] Fix the general profiling command
- [x] Review what can be split
- [x] Find a way to add a test to ensure we do not break them in the future
- [x] Add a safeguard to ensure we are actually profiling something
- [ ] Check the trace lazy state

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -150,6 +150,8 @@ return (new Config())
         'phpdoc_align' => [
             'align' => 'left',
         ],
+        // Allow inline Psalm suppress statements
+        'phpdoc_to_comment' => false,
         'php_unit_dedicate_assert' => true,
         'php_unit_method_casing' => [
             'case' => 'snake_case',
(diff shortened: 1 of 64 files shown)


theofidry:
Last item closed by https://github.com/infection/infection/pull/2544.

Issue #1329, opened 2020-09-13
Relative path from command should be based on cwd


mabar:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.17.5
| Test Framework version | PHPUnit 9.2.6
| PHP version | 7.4.4
| Platform    | Debian (via Windows WSL)
| Github Repo | -

I'm running infection from project root with following command:
```
vendor/bin/infection \
	--configuration=build/infection.json \
	--coverage=var/tmp
```

Problem is that relative path of `--coverage` option is computed from `infection.json` location, not the current working directory. Running the command results in following error:

```
In CoverageChecker.php line 223:

  Could not find the "index.xml" file. Please ensure that the XML coverage report has been properly generated at the right place. The PHPUnit option for the path given is "--coverage-xml=/path/to/project/build/var/tmp/coverage-xml"


In IndexXmlCoverageLocator.php line 87:

  Could not find any "index.xml" file in "/path/to/project/build/var/tmp"
```

The statement `The PHPUnit option for the path given is "--coverage-xml=/path/to/project/build/var/tmp/coverage-xml"` is not true, because it was `--coverage-xml=/path/to/project/var/tmp/coverage-xml` (without build)

Once I noticed what is the problem, it's easy to fix. Just change `--coverage=var/tmp` to `--coverage=../var/tmp`, but I feel like this behavior is incorrect.

mabar:
Meanwhile paths in `build/infection.json` are relative to project root (cwd) and not the file itself. Following config works correctly:

```json
{
    "source": {
        "directories": [
            "src"
        ]
    },
    "phpUnit": {
        "configDir": ""
    },
    "logs": {
        "text": "var/tmp/infection.txt"
    },
    "mutators": {
        "@default": true
    }
}

```

It should ideally be:

```json
{
    "source": {
        "directories": [
            "../src"
        ]
    },
    "phpUnit": {
        "configDir": ""
    },
    "logs": {
        "text": "../var/tmp/infection.txt"
    },
    "mutators": {
        "@default": true
    }
}

```

e.g. phpcs and phpunit work with the base directory of config file for relative paths contained in that file.

`build/phpcs.xml`
```xml
<ruleset
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:noNamespaceSchemaLocation="./../vendor/squizlabs/php_codesniffer/phpcs.xsd">

	<!-- Show only relevant part of path -->
[...]

PR #2455, opened 2025-10-18
fix!: Load source files relative to the configuration file location instead of the working directory

### LINKED ISSUE #857: Directory resolution inconsistency

theofidry:
Fixes #857, #1329.

Documentation PR: https://github.com/infection/site/pull/286.

Currently the paths configured in `source.directories` in `infection.json5` are relative to the working directory.

On one hand this is awkward because it's been the current behaviour for so long, some users may rely on it. On the other hand, I think it is fundamentally fragile, unexpected and last but not least inconsistent with most (if not all – I may have missed some) other relative paths that are relative to the configuration file.

Besides the described scenarios of the linked issues, I think the following scenario (which is the initial one I encountered while working on #2452) shows how incredibly error-prone this behaviour is. Let's consider the following structure:

```
infection5.json
src/
  ...
tests/
  benchmark/
    test.php
    infection5.json
    src/
       ...
  
```

Both infection files have the following source:

```json
{
    "source": {
        "directories": [
            "src"
        ]
    },
}

```

And `test.php` looks something like that:

```php
<?php

$container = Container::create()->withValues(
  configFile: __DIR__.'/infection5.json,
  // ...
);
```

Then, if you execute the test from the root directory, the source files collected will NOT be `/path/to/project/tests/benchmark/src` as you may expect but `/path/to/project/src`.

This is because in the `SourceFileCollector`, we have:

```php
Finder::create()->in($sourceDirectories)
```

And `$sourceDirectories` is a relative path (technically could be absolute I guess, but I don't think anyone ever use an absolute one).

In practice this directly affects us for our benchmarks. Whilst I was trying to get them working (see #2452), I couldn't understand why no trace was generated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 **Fixed:**
 
-* [BR BREAK!] The relative paths of the sources (`source.directories` in the infection configuration file) are no longer relative to the current working directory but instead to the location of the configuration file.
+* [BC BREAK!] The relative paths of the sources (`source.directories` in the Infection configuration file) are no longer relative to the current working directory but instead to the location of the configuration file.
 
 
 ## [0.31.0](https://github.com/infection/infection/tree/0.31.0) (2025-07-28)


theofidry:
I need to update `ConfigurationFactory` but I'm having a bit of an issue understanding how it's working if I'm honest...

maks-rafalko (approved):
If I understand it correctly, this is a breaking change.

Please document it properly in the changelog, as I'm pretty sure we will have issues after the next release is published. During the release, I will copy changelog to release itself.

Also, I would suggest to update https://infection.github.io/guide/usage.html#Configuration-settings to mention that `src.directories` are relative to a config file.

Related to 

- https://github.com/infection/infection/issues/1329
- https://github.com/infection/infection/issues/857

Could you please check if those 2 issues are fixed by this change? If so, we can close them. At least #857
 seemed to be fixed.

theofidry:
@maks-rafalko for that mention, I was so engrossed with the benchmark scenario I did not realise at all this was having an impact outside of it. I'll adjust the PR accordingly

theofidry:
@maks-rafalko I did a PR for the doc, added the changelog entry (but not the date) and updated the description to better describe the situation.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Change Log
 
+## [0.32.0]
+
+**Fixed:**
+
+* [BR BREAK!] The relative paths of the sources (`source.directories` in the infection configuration file) are no longer relative to the current working directory but instead to the location of the configuration file.
```
maks-rafalko:
```suggestion
* [BC BREAK!] The relative paths of the sources (`source.directories` in the Infection configuration file) are no longer relative to the current working directory but instead to the location of the configuration file.
```

maks-rafalko (approved):

PR #2456, opened 2025-10-18
perf: Ensure the profiling script fails upon invalid output

theofidry:
Currently if no mutation or trace is generated, the profiling scenario does not fail. One might notice something wrong from inspecting the Blackfire profile, but it would be nicer to have a straight up failure.

Note that this extra code has no impact on the actual profiled code.

### DIFF
--- a/tests/benchmark/BlackfireInstrumentor.php
+++ b/tests/benchmark/BlackfireInstrumentor.php
@@ -54,9 +54,11 @@ final class BlackfireInstrumentor
     }
 
     /**
-     * @param Closure(): void $main
+     * @template T
+     *
+     * @param Closure(): T $main
      */
-    public static function profile(Closure $main, SymfonyStyle $io): void
+    public static function profile(Closure $main, SymfonyStyle $io): mixed
     {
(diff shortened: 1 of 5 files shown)


theofidry:
Oups... Accidentally introduced the submodule here... will revert that

tests/benchmark/BlackfireInstrumentor.php
```diff
@@ -54,9 +54,11 @@ private function __construct()
     }
 
     /**
-     * @param Closure(): void $main
+     * @template T
+     *
+     * @param Closure(): T $main
      */
-    public static function profile(Closure $main, SymfonyStyle $io): void
+    public static function profile(Closure $main, SymfonyStyle $io): mixed
```
staabm, line 61: should this method be typed to return `T`?

theofidry: 🤦

theofidry: Fixed in https://github.com/infection/infection/pull/2464

PR #2461, opened 2025-10-19
test: Add missing cases to DotFormatterTest

theofidry:
Add the missing cases for the `DetectionStatus` for `DotFormatterTest` and leverage the assertion from `EnumBucket` to prevent missing this scenario in the future.

### DIFF
--- a/tests/phpunit/Console/OutputFormatter/DotFormatterTest.php
+++ b/tests/phpunit/Console/OutputFormatter/DotFormatterTest.php
@@ -57,34 +57,6 @@ final class DotFormatterTest extends TestCase
         $output = new BufferedOutput();
         $formatter = new DotFormatter($output);
 
-        $expected = implode(
-            "\n",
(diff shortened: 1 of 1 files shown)


src/Mutant/MutantExecutionResult.php
```diff
@@ -50,11 +50,10 @@
  */
 class MutantExecutionResult
 {
-    private readonly string $detectionStatus;
-
     private readonly string $mutatorClass;
 
     /**
+     * @param DetectionStatus::* $detectionStatus
```
theofidry: maybe worth to extract those but it currently make the CI fail so I was keen to get it fixed... 

If you prefer, I can do a PR fixing the baseline first and extract this change to reduce the noise here.

theofidry: erratum, it is better to do so indeeed

PR #2465, opened 2025-10-19
chore: Fix the AutoReview PHPUnit configuration

theofidry:
- Copy the settings from the PHPUnit configuration file of the regular test suite over to the AutoReview one. Among other things, this displays the following PHPUnit deprecation:

```
There was 1 PHPUnit test runner deprecation:

1) Metadata found in doc-comment for method Infection\Tests\AutoReview\ProjectCode\ProjectCodeTest::test_infection_bin_is_executable(). Metadata in doc-comments is deprecated and will no longer be supported in PHPUnit 12. Update your test code to use attributes instead.
```

- Fix the PHPUnit deprecration by using the corresponding Attribute instead.


theofidry:
Unfortunately it still exists :) Even Composer is actively developed on Windows 😅

PR #2468, opened 2025-10-20
refactor: Replace pseudo-enums by enums

theofidry:
This PR replaces the pre-PHP 8.1 enums that we had by PHP enums. Along the way to make things easier a few enum-related utilities have been introduced.

Some of the enum related utilities were extracted from internal Webmozarts libraries that I wrote, I plan to move forward to make this one public, but meanwhile a simplified (the library supports a few more cases) version. Since it is annoying with the license for PHP-CS-Fixer, the I copied the license in a place where PHP-CS-Fixer would not replace it. It's not ideal, but I also don't really want to spend time on this.

I tried to keep a small clean scope but I may have gotten distracted, if you feel something would be better extracted to reduce the noise I'll be happy to oblige.

Also the impact of the replacement of each enum vary wildly, if you prefer I can also do them in separate PRs.

### DIFF
--- a/tests/phpunit/AutoReview/ProjectCode/ProjectCodeProvider.php
+++ b/tests/phpunit/AutoReview/ProjectCode/ProjectCodeProvider.php
@@ -59,6 +59,7 @@ use Infection\FileSystem\Finder\ConcreteComposerExecutableFinder;
 use Infection\FileSystem\Finder\NonExecutableFinder;
 use Infection\FileSystem\Finder\TestFrameworkFinder;
 use Infection\FileSystem\SourceFileCollector;
+use Infection\Framework\Enum\EnumBucket;
 use Infection\Logger\Http\StrykerCurlClient;
 use Infection\Logger\Http\StrykerDashboardClient;
 use Infection\Metrics\MetricsCalculator;
@@ -142,6 +143,7 @@ final class ProjectCodeProvider
     public const CONCRETE_CLASSES_WITH_TESTS_IN_DIFFERENT_LOCATION = [
         FilterBuilder::class,
         SourceFileCollector::class,
+        EnumBucket::class,
     ];
 
     /**


src/Mutation/MutationAttributeKeys.php
```diff
@@ -35,33 +35,80 @@
...
-    public const END_FILE_POS = 'endFilePos';
+    /**
+     * @param array<value-of<self>, mixed> $values
+     * @param array<value-of<self>, mixed> $keysAsIndex
+     */
+    private static function assertAllAttributesExist(array $values, array $keysAsIndex): void
+    {
+        if (count($values) === count($keysAsIndex)) {
+            return;
+        }
```
maks-rafalko, line 97: is it the right condition? what if `[a]` and `[b]` provided?

src/Metrics/TargetDetectionStatusesProvider.php
```diff
@@ -104,77 +103,77 @@ private function findRequired(): Generator
...
         // Follows the logic in JsonLogger
         if ($this->logConfig->getJsonLogFilePath() !== null) {
-            yield DetectionStatus::KILLED_BY_TESTS;
+            yield DetectionStatus::KILLED_BY_TESTS->name => DetectionStatus::KILLED_BY_TESTS;
 
-            yield DetectionStatus::KILLED_BY_STATIC_ANALYSIS;
+            yield DetectionStatus::KILLED_BY_STATIC_ANALYSIS->name => DetectionStatus::KILLED_BY_STATIC_ANALYSIS;
 
-            yield DetectionStatus::ESCAPED;
+            yield DetectionStatus::ESCAPED->name => DetectionStatus::ESCAPED;
```
sanmai, line 139: I'm curious to whenever we really need to specify `->name` here

theofidry: in this provider yes, but maybe not for the reason you expect.

Previously we were doing an `array_flip` and since we preserve the keys, we get rid of duplicate entries this way. Also noteworthy, we were using the keys, because it was simpler to do so.

Now however, we no longer use the keys directly, but we still need the de-duplication to happen. To this effect there is 3 ways:

- `array_unique()` on the generated array, unfortunately this does not work with enums.
- still passing a key.
- re-indexing the array by key afterwards.

I opted with the 2nd option because it was the most straightforward, but I don't have a strong opinion on the subject. Also note it's a piece of code we want to get rid of (see the `@deprecrated`), so I didn't think too hard about it.

src/Mutation/MutationAttributeKeys.php
theofidry: That would indeed be a problem, but this is a private static method and `keysAsIndex` holds all values. Since we are manipulating indexed arrays, we can have two entries with the same key either, so I don't see a scenario where this would be a problem.

That said I appreciate the code may be less straightforward than initially thought. It's a minor internal part so I do not think it's a problem, but I can rework it if you wish

theofidry:
I forgot to check the mutations, I'll have a look if any of the newly introduced code is affected. Otherwise, this is most likely pre-existing escaped mutations which addressing is not the scope of the PR

PR #2475, opened 2025-10-21
Change API of PhpParser to return tuple

maks-rafalko:
This is an extracted PR to fix review comment https://github.com/infection/infection/pull/2448#discussion_r2443249752

### DIFF
--- a/src/PhpParser/FileParser.php
+++ b/src/PhpParser/FileParser.php
@@ -55,12 +55,15 @@ class FileParser
     /**
      * @throws UnparsableFile
      *
-     * @return Stmt[]
+     * @return array{0:Stmt[], 1: Token[]}
      */
     public function parse(SplFileInfo $fileInfo): array
     {
         try {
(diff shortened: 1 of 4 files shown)


src/PhpParser/FileParser.php
```diff
@@ -55,12 +55,15 @@ public function __construct(
     /**
      * @throws UnparsableFile
      *
-     * @return Stmt[]
+     * @return array{0:Stmt[], 1: Token[]}
```
theofidry, line 58: I think it can be made into:

```suggestion
     * @return array{Stmt[], Token[]}
```

theofidry (approved):
👏

src/PhpParser/FileParser.php
maks-rafalko: thanks, fixed https://github.com/infection/infection/commit/715f6b1f7de0aef5e0c8e0211517dda59fc385f6

PR #2476, opened 2025-10-21
perf(Tracing): Provide a better source for the Tracing benchmarks

theofidry:
The current sources shipped are not very practical. It is hard to update and results in a ton of console warnings.

The issue with this sort of thing is that a live project is much better, but then you need to deal with all sort of dependencies. A bigger project is better, but a bigger project often also has higher infrastructure (DB/extensions, etc.) requirements.

As discussed together, I created https://github.com/infection/benchmark-source which should provide a sizeable codebase we can use for benchmark purposes.

With this PR, the tracing profiling now works. There is more changes/refinement to do, but we now get a Blackfire profile which looks like this:

Screenshot 2025-10-26 at 18 16 04

### DIFF
--- a/tests/benchmark/Tracing/infection.json5
+++ b/tests/benchmark/Tracing/infection.json5
@@ -8,7 +8,7 @@
         ],
     },
     "phpUnit": {
-        configDir: "benchmark-source",
+        "configDir": "benchmark-source",
         "customPath": "./../../../vendor/bin/phpunit"
     }
 }


maks-rafalko:
I wonder what are the downsides with working with this submodule folder? I worked with them ~10 years ago and only have bad flashbacks, but probably it's not an issue anymore.

- Do we need any special commands while working with Infection? 
- If I don't need traces benchmarks, would it just work without installing any submodules?

My main concern that it should not add any complexity for our contributors. It should just work as before, ideally. 

----

Btw, failing e2e test is suspicious :)

theofidry:
>I wonder what are the downsides with working with this submodule folder? I worked with them ~10 years ago and only have bad flashbacks, but probably it's not an issue anymore.

There is annoyances, you manage another project. Don't think it's more annoying than having sources to tar/untar.

>Do we need any special commands while working with Infection?

No.

>If I don't need traces benchmarks, would it just work without installing any submodules?

Indeed, unless the benchmark are somewhat part of the e2e tests which they may be... I'm checking that. Found it, extracted it to https://github.com/infection/infection/pull/2478

theofidry:
OK scratch that I'm not happy with it. If you want to skip the wall of text, read the last paragraph.

There is two issues when benchmarking and profiling:

1. the impact of the (benchmarked/profiled) code
2. the reliability of the execution

1. In our case the coverage generated is too small. We can still see some stuff in the profile, but we might be misled about the cost of certain actions. For instance (I know it's not true it is for the sake of the example), loading the index file will have a flat cost regardless of the coverage size, but parsing it will be exponential. If it is too small you may be mistakenly thing the two costs are comparable, when in a real-world scenario the latter will have a much greater impact.

2. Due to the nature of I/O, network etc, executing a piece of code once is rarely if ever reliable for a profile/benchmark. Traditionally, Blackfire offers the `--samples` options and PHPBench has "revolutions" to execute the piece of code several time to aggregate the metrics to offer more reliable data.

When working on #2452, I could get around both with Blackfire. In pseudo code, this is what I could do:

```
loop {
  setUp // initialize the containers & all the faff
  
  enableProfiling
  executeCodeToProfile
  disableProfiling
}
```

Blackfire no longer provides the `--samples` so unfortunately we have to do that by hand. It is not ideal because maybe the code we execute corrupts sufficiently the executing process that (e.g. there is big data leaks) that it provides inconsistent results, but maybe it doesn't in which case it's a neat way to solve both 1. and 2. problems.

However, I cannot do this for PHPBench (see https://github.com/phpbench/phpbench/issues/896 about related issues).

So this is a dead end.

<hr/>

Back to the drawing board, to recap the problem is having a comprehensive codebase that is sufficiently large to generate a good size coverage data that is not utter garbage (e.g. if all source code is a one liner and has a single assertion, this will likely be too poor data).

I think in the age of AI, this could easily be achievable with a mixture of AI prompts and some scripts to generate large pieces of data. For instance, as a naive approach, I could copy x500 times the source code and tests of CPUCoreFinder, each instance in a different namespace et voilà. If we need updates/tweaks, since it's code for which the quality has no importance, we can easily update/refactor it using AI if needed (e.g. when there is a PHP or PHPUnit upgrade).

One issue we've had in the past was about how to store this data to use it for our benchmark purposes. Back then (see https://github.com/infection/infection/pull/1178), we opted for storing it in a tar and untaring it locally. It was a pragmatic approach, but hardly practical. It is impractical to update and also bloats the project git size.

Having the entire code within the project is also not ideal. In #2444, I did do that, but this meant to stay very small. Here we are talking of having hundreds of source files and tests + maybe some utilities to generate/update/maintain that code.

I do think that submodules are a good compromise, so **what do you think of creating a repository under the infection organisation for benchmark purposes**? Then I can update this PR to use this repo as a submodule instead. As a bonus, we can actually commit the coverage data to avoid having to handle that process here.

theofidry:
ping @infection/core

sanmai:
> what do you think of creating a repository under the infection organisation for benchmark purposes?

That sounds like a great idea :+1:

maks-rafalko:
sounds good 👌

theofidry:
WIP in https://github.com/infection/benchmark-source

src/Configuration/ConfigurationFactory.php
```diff
@@ -247,38 +248,54 @@ private function retrieveTmpDir(
...
 
-    private function retrievePhpUnit(SchemaConfiguration $schema, string $configDir): PhpUnit
+    private function retrievePhpUnit(SchemaConfiguration $schema, string $infectionConfigDir): PhpUnit
     {
         $phpUnit = clone $schema->getPhpUnit();
 
         $phpUnitConfigDir = $phpUnit->getConfigDir();
+        $customPath = $phpUnit->getCustomPath();
+
+        $newConfigDir = null === $phpUnitConfigDir
```
theofidry: to extract this fix

theofidry: extracted to https://github.com/infection/infection/pull/2490

theofidry:
Updated; there is one pending fix that prevents the merge (and is the cause of the CI failure) but it otherwise can be reviewed. Profiles now works, although as already mentioned, there is further refining to do still (which I will do in separate PRs).

Makefile
```diff
@@ -268,18 +268,24 @@ $(BENCHMARK_MUTATION_GENERATOR_SOURCES): tests/benchmark/MutationGenerator/sourc
...
 
-$(BENCHMARK_TRACING_COVERAGE): tests/benchmark/Tracing/coverage.tar.gz
-	@echo "Untarring the coverage, this might take a while"
-	cd tests/benchmark/Tracing; tar -xzf coverage.tar.gz
-	touch -c $@
-
-$(BENCHMARK_TRACING_SOURCES): tests/benchmark/Tracing/sources.tar.gz
-	@echo "Untarring the sources, this might take a while"
-	cd tests/benchmark/Tracing; tar -xzf sources.tar.gz
+$(BENCHMARK_TRACING_COVERAGE_DIR):
```
theofidry: to double check if it is enough, I think I'm missing a rule for the git submodule init

tests/benchmark/Tracing/infection.json5
```diff
@@ -0,0 +1,14 @@
+{
```
theofidry, line 1: the config is now here in the benchmark rather than the donwstream package. Maybe that will need to change, but for now I think it's fine as it is

Makefile
```diff
@@ -268,18 +269,27 @@ $(BENCHMARK_MUTATION_GENERATOR_SOURCES): tests/benchmark/MutationGenerator/sourc
 	cd tests/benchmark/MutationGenerator; tar -xzf sources.tar.gz
 	touch -c $@
 
-$(BENCHMARK_TRACING_COVERAGE): tests/benchmark/Tracing/coverage.tar.gz
-	@echo "Untarring the coverage, this might take a while"
-	cd tests/benchmark/Tracing; tar -xzf coverage.tar.gz
+$(BENCHMARK_TRACING_COVERAGE_DIR): $(BENCHMARK_TRACING_COVERAGE_SOURCE_DIR)
```
theofidry, line 272: to clarify what this is doing: in `benchmark-source` we store the coverage under `dist`, but because the coverage contains absolute path we "anonymise" it by replacing the absolute path by `/path/to/project`. Here we are simply inversing this operation.

Thinking about it, maybe benchmark source could provide a copy utility to do that instead, but that can be done later

theofidry:
This PR is blocking quite a long chain of changes.

Unless I misunderstood, given the initial intention of using a submodule was fine and that the suggested idea of using a codebase we have control over was approved, I'll assume this change is ok so I'll merge this.

If there is things to address I believe they can be addressed afterwards.

PR #2482, opened 2025-10-24
refactor(test): Introduce a ConfigurationBuilder

theofidry:
This PR introduces `ConfigurationBuilder`, a utility class to easily create a `Configuration` object.

The initial motivator for this change is the desired to make `Configuration` a readonly final value object that can use promoted properties. This is not possible at the moment because we rely on PHPUnit mocks to create a mock for which we mock a handful of specific getters.

Instead, with this builder, we can now create a specific `Configuration` and updates the existing tests to leverage this new builder instead of the PHPUnit mocks.

### Usages & Pros

Examples:

```php
// Signals we just want a configuration object; we do not care about its values
$config = ConfigurationBuilder::withMinimalTestData()->build();

// Create a configuration object with a specific state; but we only need to care
// about the parts of the state we are interested in
$config = ConfigurationBuilder::withMinimalTestData()
    ->withSkipInitialTests(true)
    ->withWithUncovered(true)
    ->build();

// Create a configuration object from an existing instance.
$config = ConfigurationBuilder::from($previous)
    ->withSkipInitialTests(true)
    ->build();
```

Although a bit verbose, it is expressive and readable. This removes the need of using any mocks, which also has the constraint of preventing us to use those sort of objects in data providers, as mocking requires a non-static factory method (`$this->createMock()`).

Being able to create such objects instead of mocks is also quite powerful now that we can compare two objects easily:

```php
// Won't work with mocks!
$this->assertEquals($expectedConfig, $actualConfig);

// And this is a lot less verbose than using `$this->assertConfigStateIs(...)`.
```

Another benefit of this approach is its resilience to the test of time. If tomorrow we add/remove an attribute, the existing tests should not be affected. I say "should" instead of "won't" because it is possible a test mistakenly relied on an implicit state. For example:

```php
$config = ConfigurationBuilder::withMinimalTestData()
    ->withSkipInitialTests(true)
    // ->withWithUncovered(true) – one could have forgot to call this and not notice the issue as `true` was the default value
    ->build();
```

But in my experience, this has never been a big deal. When noticed, it's easily fixed in a dedicated PR to avoid any noise and that is just adding one missing call nothing more.

### Cons

IMO the first biggest con is that it requires boilerplate. On the other hand, it's the sort of code that an automatic code generator or any AI can easily create without trouble and this boilerplate does not creep up in the rest of the code: it is only one class that is quite verbose, the actual API exposed is fine.

The 2nd biggest disadvantage is highlighted by a diff when @sanmai commented:

```diff
-- $config = $this->createMock(Configuration::class);
-- $config
--     ->expects($this->once())
--     ->method('shouldSkipInitialTests')
--     ->willReturn(true);
++ $config = ConfigurationBuilder::withMinimalTestData()
++    ->withSkipInitialTests(true)
++    ->withWithUncovered(true)
++    ->build();
```

As you can see, we loose the check of `$this->once()`. More generally, since we do not use a mock, we can no longer spy on various calls.

However, I think this is fine. Here's my thinking:

- In the first place, I think this sort of testing derives from being a bit of a control freak (if you feel targeted, take no offense, I include myself in that category in regards to tests). It too often leads to tests that breaks at the slightest change and the value of testing this is fairly limited.
- If you really like those sort of tests, you will be very frustrated in PHP because you cannot spy on properties, which is fairly limiting.
- The above is truer as of PHP8 since you may want to leverage public readonly properties to reduce boilerplate.
- This also requires any object that you have to not be `final` as otherwise you cannot mock them.

So technically, yes something is lost there, but I think spying on calls is fine to be limited to services rather than extended to all entities/value objects.


.php-cs-fixer.php
```diff
@@ -157,7 +157,7 @@
             'case' => 'snake_case',
         ],
         'php_unit_set_up_tear_down_visibility' => true,
-        'php_unit_strict' => true,
+        'php_unit_strict' => false,
```
theofidry, line 161: unfortunately this setting forces to compare objects by reference which we do not want

tests/phpunit/Configuration/ConfigurationBuilderTest.php
```diff
@@ -0,0 +1,68 @@
...
+
+#[CoversClass(ConfigurationBuilder::class)]
+final class ConfigurationBuilderTest extends TestCase
+{
+    #[DataProvider('configurationProvider')]
+    public function test_it_can_create_a_builder_from_a_built_instance(Configuration $configuration): void
+    {
+        $actual = ConfigurationBuilder::from($configuration)->build();
+
+        $this->assertSame($configuration, $actual);
```
sanmai:
```suggestion
        $this->assertEquals($configuration, $actual);
```

tests/phpunit/EngineTest.php
```diff
@@ -436,12 +426,10 @@ public function test_memory_limiter_is_applied_after_static_analysis_when_enable
...
-        $config = $this->createMock(Configuration::class);
-        $config
-            ->expects($this->once())
-            ->method('shouldSkipInitialTests')
-            ->willReturn(true)
-        ;
+        $config = ConfigurationBuilder::withMinimalTestData()
+            ->withSkipInitialTests(true)
+            ->withWithUncovered(true)
+            ->build();
```
sanmai, line 432: We might have remove a tested assertion somewhere in this PR. Since we no longer make a full MT run on PRs (IIRC), so there's a chance this updated test will miss some mutations that were otherwise caught.

.php-cs-fixer.php
maks-rafalko: is it possible to ignore it in 1 place instead of disable at all?

theofidry: Unfortunately no, see https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/3871

tests/phpunit/EngineTest.php
theofidry: I think if anything the builder will _introduce_ assertions, the ones from `Configuration::__construct()`. The mock on the other hand does not have any

.php-cs-fixer.php
theofidry: Tbh I think this rule is a not very productive: it encourages a lot of boilerplate instead of allowing to use `assertEquals()` for objects. I'll look into introducing an alternative solution to ensure `assertEquals()` does strictly compare object properties

theofidry: Extracted this to https://github.com/infection/infection/pull/2486

tests/phpunit/EngineTest.php
sanmai: The mock had `->expects($this->once())`.

theofidry: Ah indeed I missed the `once` there. I updated the PR to reflect on that

tests/phpunit/Configuration/ConfigurationBuilder.php
```diff
@@ -0,0 +1,604 @@
...
+        private string $testFramework,
+        private ?string $bootstrap,
+        private ?string $initialTestsPhpOptions,
+        private string $testFrameworkExtraOptions,
+        private ?string $staticAnalysisToolOptions,
+        private string $coveragePath,
+        private bool $skipCoverage,
+        private bool $skipInitialTests,
+        private bool $debug,
+        private bool $withUncovered,
```
theofidry: maybe better to name it `$uncovered` here otherwise the wither will be `::withWithUncovered()`

tests/phpunit/Configuration/ConfigurationBuilder.php
```diff
@@ -0,0 +1,604 @@
...
+            timeout: 5.0,
+            sourceDirectories: ['src', 'lib'],
+            sourceFiles: [
+                new SplFileInfo('src/Foo.php', 'src/Foo.php', 'src/Foo.php'),
+                new SplFileInfo('src/Bar.php', 'src/Bar.php', 'src/Bar.php'),
+            ],
+            sourceFilesFilter: 'src/Foo.php,src/Bar.php',
+            sourceFilesExcludes: ['vendor', 'tests'],
+            logs: new Logs(
+                'text.log',
```
theofidry: hmm using named parameters there too would be better

PR #2483, opened 2025-10-24
refactor: Leverage public readonly values

theofidry:
This PR leverages PHP8 readonly classes and promoted properties to reduce boilerplate and move from private properties with getters to public readonly properties when possible.

This was done with a bit of help of rector by configuring `ClassPropertyAssignToConstructorPromotionRector` and removing the skipped rules `ReadOnlyClassRector` and `ReadOnlyPropertyRector`.

### DIFF
--- a/src/Engine.php
+++ b/src/Engine.php
@@ -220,8 +220,6 @@ final readonly class Engine
             );
         }
 
-        $configuration = $this->config;
-
-        return $configuration->testFrameworkExtraOptions;
+        return $this->config->testFrameworkExtraOptions;
     }
 }
--- a/src/Process/Factory/MutantProcessContainerFactory.php
+++ b/src/Process/Factory/MutantProcessContainerFactory.php
@@ -83,9 +83,7 @@ class MutantProcessContainerFactory
             timeout: $timeout,
         );
 
-        $configuration = $this->configuration;
-
-        if ($configuration->isDryRun) {
+        if ($this->configuration->isDryRun) {
             $process = DryRunProcess::fromProcess($process);
         }
 


rector.php
```diff
@@ -47,11 +47,17 @@
...
+            'allow_model_based_classes' => true,
+        ],
+    )
     ->withSkip([
         ReadOnlyPropertyRector::class => [
             // property can't be readonly as it's returned by reference and may be updated
             __DIR__ . '/src/TestFramework/Coverage/TestLocations.php',
         ],
-        ReadOnlyClassRector::class,
-        ReadOnlyPropertyRector::class,
```
maks-rafalko, line 56: why was it removed? are these implied in `ClassPropertyAssignToConstructorPromotionRector`?

theofidry: they were part of the _skipped_ rules, not the configured ones. IIRC it was from the Rector2 upgrade to not introduce unintended changes during the upgrade

src/Process/Factory/MutantProcessContainerFactory.php
```diff
@@ -83,7 +83,9 @@ public function create(Mutant $mutant, string $testFrameworkExtraOptions = ''):
             timeout: $timeout,
         );
 
-        if ($this->configuration->isDryRun()) {
+        $configuration = $this->configuration;
+
+        if ($configuration->isDryRun) {
```
sanmai:
```suggestion
        if ($this->configuration->isDryRun) {
```

src/Engine.php
```diff
@@ -215,11 +215,13 @@ private function getFilteredExtraOptionsForMutant(): string
...
-                $this->config->getTestFrameworkExtraOptions(),
+                $this->config->testFrameworkExtraOptions,
                 $this->adapter->getInitialRunOnlyOptions(),
             );
         }
 
-        return $this->config->getTestFrameworkExtraOptions();
+        $configuration = $this->config;
+
+        return $configuration->testFrameworkExtraOptions;
```
sanmai:
```suggestion
        return $this->config->testFrameworkExtraOptions;
```

sanmai (approved):
spot-checked, makes sense

rector.php
```diff
@@ -47,11 +47,17 @@
     ->withRules([
         AddCoversClassAttributeRector::class,
     ])
+    ->withConfiguredRule(
+        ClassPropertyAssignToConstructorPromotionRector::class,
+        [
+            'inline_public' => false,
+            'rename_property' => true,
```
staabm, line 54: does the infection code-base supports named-arguments everywhere?

my experience with `rename_property=true` is, that it sometimes renames properties, which can lead to subtle BC breaks.

I would either play it save and disable this option, or alternatively it might make sense to use `@no-named-arguments` phpdoc to signal in wich classes we do not support named args (see e.g. PHPUnit, which defines this annotation on nearly all classes - enforced via PHP-CS-Fixer)

theofidry: >does the infection code-base supports named-arguments everywhere?

Not sure what you mean by that.

>my experience with rename_property=true is, that it sometimes renames properties, which can lead to subtle BC breaks.

That was the case here actually, although I changed it by hand. For instance it made more sense to have the property `bool $isDryRun` which was previously exposed by `::isDryRun()` than now exposing `$dryRun`.

So far all the touched code is tagged as `@internal`.

But we can change the rector rules anytime so if that causes problems in the future we can disable it then.

PR #2484, opened 2025-10-24
build: Add support for direnv

theofidry:
[direnv](https://direnv.net/) is a nice utility. This PR proposes a very simple integration where the Composer bins of the project are added to your path. As a result, you can now type:

```shell
$ php-cs-fixer
$ phpunit
```

And they will work. Additionally, it will automatically load a `.envrc.local` (which is not committed) if present, allowing one to further extend their own envrc.

Note that this is not required by anyone to use, but can allow one to customize their own experience with this project.

### DIFF
--- a/tests/phpunit/Configuration/Entry/PhpStanTest.php
+++ b/tests/phpunit/Configuration/Entry/PhpStanTest.php
@@ -36,7 +36,6 @@ declare(strict_types=1);
 namespace Infection\Tests\Configuration\Entry;
 
 use Infection\Configuration\Entry\PhpStan;
-use Infection\Configuration\Entry\PhpUnit;
 use PHPUnit\Framework\Attributes\CoversClass;
 use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,6 +1,8 @@
 !/.tools/.gitkeep
 !/build/.gitkeep
 !/build/cache/.gitkeep
+/.direnv/
+/.envrc.local
```
sanmai, line 5: I have these globally ignored so I do not need to add them in every single project. Consider that you can do that too.

sanmai (approved):

.envrc
```diff
@@ -0,0 +1,4 @@
+export PATH=$PWD/vendor/bin:$PATH
+export PATH=$PWD/.tools:$PATH
```
maks-rafalko, line 2: curious, what does this allow you to do? what does it simplify exactly? (keeping in mind that we have `make *` commands and `docker-compose.yaml`

.gitignore
theofidry: I do but it's not a universal practice. Often IDEs related stuff make it here but there is many cases where it is not for direnv. Since `.gitignore` is cheap, i.e. it has close to no maintenance cost, I think it's fine

.envrc
theofidry: you could type `php-cs-fixer` and it will use the one from `/path/to/infection/.tools/php-cs-fixer`.

>keeping in mind that we have make * commands and docker-compose.yaml

Yes, it's not really intended as a replacement, more of if for whatever reason you wanna use that executable file it will be slightly easier. I hesitated to add it but I'm adding the Composer bins already, I thought adding this would make sense too.

We could totally do without any of this tbh, the only part I really am keen to have is the `.envrc.local` part

PR #2490, opened 2025-10-26
fix!: Make PHPUnit or PHPStan custom path absolute

theofidry:
This is similar to #2455.

Currently the custom path of PHPUnit and PHPStan is left alone. In the documentation it is documented as an absolute path, but it can also be relative. In this scenario, I think it should be relative to the infection configuration file.

This is also necessary for #2476.

Docs PR https://github.com/infection/site/pull/287


theofidry:
@maks-rafalko there is because before it depending on the working directory for relative paths.

I think in practice it is not much of an issue because:

- Not many people are likely to use a custom path.
- Most people execute infection from the same working directory (root project, same in the CI)
- If you didn't, you may have had an issue about the path not being found and made it so that it is executed from the same directory every time (i.e. go to the previous point)

In other words, unless you were aware of this feature and actively using it, which I think is an extreme minority (I hope non-existent) amount of people, you are not affected.

But technically, it is a BC break nonetheless.

PR #2499, opened 2025-10-27
test(benchmark): Update the benchmark scripts

theofidry:
I do not like too much this sort of PR that does a lot of various things, but untangling it is too much effort for what it is, both for me and any reviewer.

This PR:

- Removes the `generate-mutations.php` and `provide-traces.php` scripts. Those were originally meant as alternatives to the respective `profile.php` files. However, I think #2460 is a more elegant solution and thus it has since then be made irrelevant.
- Adjust the handling of the `$maxCount`: for both benchmarks it is now a `positive-int`. This is to have consistency on both side and simplify a bit the downstream code.
- Make the max count argument into an option. The rationale is that otherwise, doing `php profile.php -1` will fail, as there is no option `-1`. Instead you should call `php profile.php "-1"` but I think at this point using an option is better.
- Add a bit more validation & user friendly messages upon failure for the max count option.
- Bumped the maximum count for the tracing benchmark: we generate ~250 traces, which is relatively fast hence I do not think we need to limit it by default (one can still, for debugging purposes or whatever).
- Bumped the maximum number of mutations. I was thinking of removing the limit (like for the tracing), but we generate over 36k mutations... which does take a while.
- Rename `BenchmarkTest` for `BenchmarkSmokeTest` with an explanation. This is to both make it more obvious what this test is about and not confusing future incoming `*Bench` tests (will be detailed in a dedicated PR as they get introduced).
- Update `BenchmarkTest` to use the respective `profile.php` files instead.
- Rename `--max-mutations-count` to `--max-mutation-count` for consistency & correctness.
- Fix `provide-traces-closure.php` max loop check, previously if called with maxCount=2, then it would generate only 1 trace. As a result calling it with maxCount=1 resulted in an error as no trace would be generated.

### DIFF
--- a/tests/phpunit/BenchmarkSmokeTest.php
+++ b/tests/phpunit/BenchmarkSmokeTest.php
@@ -35,9 +35,9 @@ declare(strict_types=1);
 
 namespace Infection\Tests;
 
+use Infection\Framework\OperatingSystem;
 use Infection\Tests\TestingUtility\Process\TestPhpExecutableFinder;
 use function is_dir;
-use const PHP_OS_FAMILY;
 use const PHP_SAPI;
 use PHPUnit\Framework\Attributes\CoversNothing;
 use PHPUnit\Framework\Attributes\DataProvider;
@@ -62,7 +62,7 @@ final class BenchmarkSmokeTest extends TestCase
     #[DataProvider('provideBenchmarks')]
     public function test_all_the_benchmarks_can_be_executed(array $command, string $sourcesLocation): void
     {
-        if (PHP_OS_FAMILY === 'Windows') {
+        if (OperatingSystem::isWindows()) {
             $this->markTestSkipped('Not interested in profiling on Windows.');
         }
 
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,6 @@ BOX_URL="https://github.com/humbug/box/releases/download/4.5.1/box.phar"
 
 PHP_CS_FIXER=./.tools/php-cs-fixer
 PHP_CS_FIXER_URL="https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/download/v3.65.0/php-cs-fixer.phar"
-PHP_CS_FIXER_CACHE=.php_cs.cache
 
 PHPSTAN=./vendor/bin/phpstan
 RECTOR=./vendor/bin/rector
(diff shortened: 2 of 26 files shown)


tests/phpunit/BenchmarkSmokeTest.php
```diff
@@ -91,13 +95,21 @@ public function test_all_the_benchmarks_can_be_executed(string $path, string $so
...
+                '--debug',
+            ],
             self::BENCHMARK_DIR . '/MutationGenerator/sources',
         ];
 
         yield 'Tracing' => [
-            realpath(self::BENCHMARK_DIR . '/Tracing/provide-traces.php'),
-            self::BENCHMARK_DIR . '/Tracing/sources',
+            [
+                Path::canonicalize(self::BENCHMARK_DIR . '/Tracing/profile.php'),
```
theofidry, line 106: the change to `Path` has been extracted in https://github.com/infection/infection/pull/2501

tests/phpunit/BenchmarkSmokeTest.php
```diff
@@ -91,13 +95,21 @@ public function test_all_the_benchmarks_can_be_executed(string $path, string $so
     public static function provideBenchmarks(): iterable
     {
         yield 'MutationGenerator' => [
-            realpath(self::BENCHMARK_DIR . '/MutationGenerator/generate-mutations.php'),
+            [
```
theofidry, line 96: the downside of making the max count an option is that we no longer have a generic API. However, I don't think it's much of a problem. New scripts may not fit that API at all, so having a more flexible solution would be better, and I think passing the command to execute does just that.

tests/phpunit/BenchmarkSmokeTest.php
```diff
@@ -91,13 +95,21 @@ public function test_all_the_benchmarks_can_be_executed(string $path, string $so
...
+                Path::canonicalize(self::BENCHMARK_DIR . '/MutationGenerator/profile.php'),
+                '1',
```
theofidry: oups, should be using the option here!

tests/phpunit/BenchmarkSmokeTest.php
```diff
@@ -91,18 +95,32 @@ public function test_all_the_benchmarks_can_be_executed(string $path, string $so
...
+                '--debug',
+            ],
+            self::BENCHMARK_DIR . '/Tracing/coverage',
         ];
     }
 
     private function getPhpExecutable(): string
     {
-        return $this->phpExecutable ??= (new PhpExecutableFinder())->find();
+        if (!isset($this->phpExecutable)) {
```
theofidry: this change is someone noisier than I'd like and there is a bunch of other places that deserves fixing too, so extracted this to https://github.com/infection/infection/pull/2503

PR #2500, opened 2025-10-27
test: Silence output when dumping the autoloader for the benchmark

theofidry:
Because we have the following in our `composer.json`:

```json
    "autoload-dev": {
        "psr-4": {
            "Infection\\Benchmark\\": "tests/benchmark",
```

The following code:

- `tests/benchmark/MutationGenerator/sources/`
- `tests/benchmark/Tracing/benchmark-source/`

is included. As a result, when we execute `composer dump --classmap-authoritative` as part of the benchmark scripts, we get a wall of warnings as the code in those directories do not follow the classmap (hence is skipped).

I can't find a nice way to address this without trading off convenience (we do not want to have to manually include every single source code file under `tests/benchmark`). Hence, my proposal is to add `--quiet` to silence those warnings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,7 +121,7 @@ profile:
 
 .PHONY: profile_mutation_generator
 profile_mutation_generator: vendor $(BENCHMARK_MUTATION_GENERATOR_SOURCES)
-	composer dump --classmap-authoritative
+	composer dump --classmap-authoritative --quiet
 	blackfire run \
 		--samples=5 \
 		--title="MutationGenerator" \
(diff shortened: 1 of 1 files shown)


sanmai:
Since we do `composer dump` anyway, what do you think about speccing `tests/benchmark` as `classmap` in the config?

theofidry:
@sanmai the "issue" is that I do not think we want it to be part of the root autoload: at least the `benchmark-source` has its own already (not used though, so far).

It is not a problem in practice afterwards because this code is not really executed by our code (except `BenchmarkTest` I guess).

I don't have a strong opinion about it though.

sanmai:
Not sure if I know right away why we need to PSR-4 this hierarchy in the first place.

theofidry:
>Not sure if I know right away why we need to PSR-4 this hierarchy in the first place.

The (benchmarked/profiled) source code we do not, but the utilities around it, e.g. instrumentors, are better off being included in the autoload one way or another.

Feel free to experiment with the classmap, maybe it is better? I have no clue tbh

PR #2502, opened 2025-10-27
ci: Move the benchmark test to its dedicated workflow

theofidry:
`BenchmarkSmokeTest` was configured as an integration test but always incomplete as its sources were never provided in the CI.

To execute the benchmark test, we need the sources, and I do not think it is worth troubling ourselves with the whole matrix we have for our unit & integration tests. So I propose to move it to its own workflow.

Another benefit is if we add actual benchmarks (as opposed to scripts to execute with Blackfire), then we can include them there without impacting the rest of the CI. This is already sketched out in https://github.com/infection/infection/pull/2516.


.github/workflows/benchmarks.yaml
```diff
@@ -0,0 +1,54 @@
...
+        uses: ramsey/composer-install@v3
+
+      - name: Run tests
+        shell: bash
+        run: make test-benchmark
+
+  # This is a meta job to avoid to have to constantly change the protection rules
+  # whenever we touch the matrix.
+  tests-status:
+    name: Unit & Integration Tests Status
```
theofidry:
```suggestion
    name: Benchmarks Status
```

.github/workflows/benchmarks.yaml
```diff
@@ -0,0 +1,54 @@
+# yamllint disable rule:line-length
+# yamllint disable rule:braces
+
+name: Benchmark Tests
```
theofidry:
```suggestion
name: Benchmarks
```

Makefile
```diff
@@ -145,7 +145,7 @@ autoreview: cs-check phpstan psalm validate test-autoreview rector-check detect-
 
 .PHONY: test
 test:		 	## Runs all the tests
-test: autoreview test-unit test-e2e test-infection
+test: autoreview test-unit test-benchmark test-e2e test-infection
```
sidz, line 146: maybe we could add `test-benchmark-docker`  target too? So it will be easy to run benchmarks on different PHP versions inside of docker containers?

maks-rafalko: out of curiosity: do you actually use `***-docker` targets?

I find myself jumping into a container e.g. `docker compose run ...` and just run usual targets there like `make test` etc., but probably that's only me

.github/workflows/benchmarks.yaml
maks-rafalko: I'm not sure this is a good name, as well as `BenchmarkSmokeTest` look misleading.

When I see "Benchmark Tests" I expect that this GH action will do benchmarking, but in fact it runs `BenchmarkSmokeTest` which has a description: 

> Smoke test to ensure the benchmark profile scripts somewhat work.

So we don't actually do a benchmarking, we check that **the script** that does benchmarks works.

Are you going to add other real benchmarks to this GH action so this name will make sense? Or will it just check that benchmark scripts are working?

Makefile
theofidry: >maybe we could add test-benchmark-docker target too?

I've just tried, but that requires more work as the coverage path would no longer match. I'm not too keen on working on that.

.github/workflows/benchmarks.yaml
theofidry: The plan is indeed to execute actual benchmarks: #2516

theofidry:
Since https://github.com/infection/infection/pull/2516 has been approved I'll merge this as it is. If we want to we can always revert the part of extracting the PHPUnit test execution to this workflow.

theofidry:
The `Benchmark Status` workflow has also been added to the mandatory passing checks for a merge.

PR #2503, opened 2025-10-28
test(TestPhpExecutableFinder): Introduce a PHP executable finder for tests

theofidry:
This finder denotes itself from a non-test one by the fact that it has a hard assert statement without any actionable documented by the user and that the result is memoized in a static property.

This helps to remove some boilerplate and fix some PHPStan types.


sanmai (approved):
Boilerplate - gone :+1:

PR #2504, opened 2025-10-28
refactor(Str): Consolidate the string related utilities

theofidry:
We have a lot of string related utilities:

- `Str::trimLineReturns()`
- `StringNormalizer::normalizeString()`
- `normalizeLineReturn()`
- `normalize_trailing_spaces()`
- `LineReturnNormalizer::normalize()`

`Str` and `StringNormalizer` are both part of `src` but `@internal`, whilst the others are part of `tests`. A few of them do very similar things or the same thing and some utilities are now replaceable by the Symfony `Path`.

In this PR, I consolidate this by moving most things to `Str`:

- `normalize_trailing_spaces()` and `StringNormalizer::normalizeString()` were almost identical. The only difference is that the latter was not handling `\r\n` or `\r`  line returns. The implementation of `normalize_trailing_spaces()` has thus been kept and moved as `Str::normalize()` – I didn't keep "trailing spaces" because it does more than that, it also normalises the line returns.
- `normalizeLineReturn()` and `LineReturnNormalizer::normalize()` are very similar. They transform all the possible line returns into respectively `\n` and `PHP_EOL`. Since we need both, I renamed them respectively `Str::toLinuxLineReturn()` and `Str::toSystemLineReturn()`*.

Also rather than having `Str` at the root, I moved it under the `Framework` namespace with the other utilities.

*: note that the implementation has changed to actually work correctly on Windows. Indeed there was an issue with `str_replace` illustrated by https://3v4l.org/RlV8X.

### DIFF
--- a/src/Framework/Str.php
+++ b/src/Framework/Str.php
@@ -55,6 +55,9 @@ final class Str
 
     private const LINE_RETURNS_REGEX = '/\r\n|\r|\n/';
 
+    /**
+     * @psalm-suppress InvalidReturnStatement,InvalidReturnType
+     */
     public static function toSystemLineReturn(string $value): string
     {
         return preg_replace(
@@ -64,6 +67,9 @@ final class Str
         );
     }
 
+    /**
+     * @psalm-suppress InvalidReturnStatement,InvalidReturnType
+     */
     public static function toLinuxLineReturn(string $value): string
     {
         return preg_replace(


tests/phpunit/TestingUtility/Platform.php
```diff
@@ -36,19 +36,24 @@
 namespace Infection\Tests\TestingUtility;
 
 use Infection\CannotBeInstantiated;
-use const PHP_EOL;
-use function str_replace;
+use const PHP_OS_FAMILY;
 
-final class LineReturnNormalizer
+final class Platform
```
theofidry: not sure why it sees this as a renaming, but this should be extracted

tests/phpunit/Helpers.php
```diff
@@ -35,43 +35,16 @@
...
 use function substr;
 use Symfony\Component\Filesystem\Exception\IOException;
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Filesystem\Path;
 use function sys_get_temp_dir;
 
-/**
- * Normalizes path. Replaces backslashes with forward ones
- */
-function normalizePath(string $value): string
```
theofidry: those two are replaced by `Path`, IMO it's is worth extracting too

tests/phpunit/Helpers.php
```diff
@@ -35,43 +35,16 @@
...
-    return implode(
-        "\n",
-        array_map(
-            'rtrim',
-            explode("\n", normalizeLineReturn($value)),
-        ),
-    );
-}
-
 function generator_to_phpunit_data_provider(iterable $source): iterable
```
theofidry: to address this one in a separate PR

tests/phpunit/TestingUtility/Platform.php
theofidry: extracted to https://github.com/infection/infection/pull/2505

tests/phpunit/Helpers.php
theofidry: done in https://github.com/infection/infection/pull/2506

tests/phpunit/StrTest.php
```diff
@@ -137,4 +308,37 @@ public static function stringProvider(): iterable
                 TXT,
         ];
     }
+
+    #[DataProvider('utf8StringConversionProvider')]
+    public function test_it_converts_strings_to_utf8_encoding(string $input, string $expected): void
```
theofidry, line 323: extracted to https://github.com/infection/infection/pull/2507

tests/phpunit/Helpers.php
theofidry: https://github.com/infection/infection/pull/2508

src/Str.php
theofidry, line 52: probably could be moved under `Framework`

src/Framework/Str.php
```diff
@@ -51,51 +53,116 @@ final class Str
...
+            self::LINE_RETURNS_REGEX,
+            "\n",
+            $value,
+        );
+    }
+
+    /**
+     * Removes trailing spaces and normalizes the line return (to the unix/linux one).
+     */
+    public static function normalize(string $value): string
```
sanmai, line 85: normalization could mean many other things (i.e. I immediately thought of unicode normalization)
```suggestion
    public static function trimLines(string $value): string
```

sanmai (approved):
LGTM with one suggestion, while we are at it

src/Framework/Str.php
theofidry: On another hand this does more than trimming lines, and it would be very close name-wise to `::trimLineReturns()` which does something quite different (removing leading and trailing blank lines).

I don't like much `::normalize()` though so unless someone has another suggestion, I'll apply this one.

src/Framework/Str.php
```diff
@@ -51,51 +53,116 @@ final class Str
...
 
-    public static function trimLineReturns(string $string): string
+    private const LINE_RETURNS_REGEX = '/\r\n|\r|\n/';
+
+    /**
+     * @psalm-suppress InvalidReturnStatement,InvalidReturnType
+     */
+    public static function toSystemLineReturn(string $value): string
+    {
+        return preg_replace(
```
theofidry, line 63: note here that I used `preg_replace()` because I found it simpler at that time, but it is also managable with `strtr`:

```php
return strtr(
            $string,
            [
                "\r\n" => "\n",
                "\r"   => "\n",
            ],
        );
```

in case you have a preference...

theofidry:
As there was more unexpected behaviours & confusions I extracted multiple pieces from this PR. See https://github.com/infection/infection/pulls?q=is%3Apr+is%3Amerged+%22%28Str%29%3A%22.

PR #2513, opened 2025-10-29
fix(benchmark): Limit the number of generated mutations to the max count

theofidry:
Previously, executing:

```shell
php tests/benchmark/MutationGenerator/profile.php --debug --max-mutation-count=1
```

was still resulting in 292 mutations generated because we were not breaking the outer loop.


theofidry:
Given the simplicity and the affected code, I'll go ahead and merge this one.

maks-rafalko:
I feel like we need some kind of documentation about what we can do with all this new benchmarks you are working on and how to use it.

I'm already lost, but didn't look into it yet so apologies if it's described or easy to understand

theofidry:
I'd be happy to have a round of review with you about the docs. I tried to keep things self-documented and documented what I can but there is certainly several pieces that could use more docs. I am less sure which ones and how to document them

PR #2512, opened 2025-10-29
test(benchmark): Fix typo in the printed message

theofidry:
There can be 0 or 1 mutation/trace generated too.

PR #2514, opened 2025-10-29
test(benchmark): Test the output of the profile scripts

theofidry:
As shown in #2513, #2499 and #2476, the current profile scripts are somewhat fragile and too little tested. They are not meant to be a critical piece of code we need to babysit but they were probably broken for years without us noticing.

This PR also adds the output to the tested subject, which gives a bit more guarantee without breaking a leg.


tests/benchmark/MutationGenerator/generate-mutations-closure.php
```diff
@@ -91,7 +91,7 @@ static function (SplFileInfo $fileInfo): Trace {
             ++$count;
 
             if ($count >= $maxCount) {
-                break;
+                break 2;
```
theofidry, line 94: extracted to https://github.com/infection/infection/pull/2513

tests/benchmark/Tracing/profile.php
```diff
@@ -122,7 +122,7 @@
 
 $output->writeln(
     sprintf(
-        '%d traces generated.',
+        '%d trace(s) generated.',
```
theofidry, line 125: extracted to https://github.com/infection/infection/pull/2512

tests/benchmark/MutationGenerator/profile.php
```diff
@@ -121,7 +121,7 @@
 
 $output->writeln(
     sprintf(
-        '%d mutations generated.',
+        '%d mutation(s) generated.',
```
theofidry, line 124: extracted to https://github.com/infection/infection/pull/2512

sidz (approved):

PR #2516, opened 2025-10-29
test(benchmark): Introduce PHPBench benchmarks

### LINKED ISSUE #1176: [RFC] Performance monitoring

theofidry:
Closes #1176.

### Why PHPBench instead of the `profile.php`

`profile.php` is meant to leverage the Blackfire instrumentor to precisely profile the piece of code we want. Tools like Blackfire however, tend to have a significant overhead distorting the real performances of the inspected code. It is often a fine tradeoff because it allows to have very detailed information on the impact of each pieces of codes.

[PHPBench](https://phpbench.readthedocs.io/en/latest/) however, will execute a piece of code and collects the general metrics from it. We also take multiple samples (called _iterations_ in PHPBench) to check the aggregated result to have a more reliable output.

As a result, this tools gives more reliable information on the performance of the code, but you do not have any details about what specific piece of code is taking computation time, or how many networks requests are made etc.

As an example, IIRC in PHP-Scoper I had a piece of code taking 3s user time, but that ended up taking 2min for a Blackfire profile. This is because a profiler cannot "release" information about the calling code until it is finished, so the more long nested calls you have, the worse the impact is. It is still useful to know what to improve, but not necessarily reliable in terms of actual time taken. 

**TL:DR; PHPBench = black box with reliable metrics, Blackfire = white box with distorted metrics.**

### How to execute the benchmarks

```shell
make benchmarks
```

The output will look like this:

<details>
<summary>PHPBench results</summary>

```
Subjects: 2, Assertions: 0, Failures: 0, Errors: 0
+------------------------+-------------------------+-----+------+-----+-----------+---------+--------+
| benchmark              | subject                 | set | revs | its | mem_peak  | mode    | rstdev |
+------------------------+-------------------------+-----+------+-----+-----------+---------+--------+
| TracingBench           | benchTracing            |     | 1    | 5   | 23.634mb  | 5.193s  | ±2.02% |
| MutationGeneratorBench | benchMutationGeneration |     | 1    | 5   | 294.591mb | 51.513s | ±1.55% |
+------------------------+-------------------------+-----+------+-----+-----------+---------+--------+

Average iteration times

5.2s      │ █ █ ▅ █ █ 
4.6s      │ █ █ █ █ █ 
3.9s      │ █ █ █ █ █ 
3.3s      │ █ █ █ █ █ 
2.6s      │ █ █ █ █ █ 
2.0s      │ █ █ █ █ █ 
1.3s      │ █ █ █ █ █ 
654.1ms   │ █ █ █ █ █ 
          └───────────
            1 2 3 4 5 

[█ <current>] 

1: 0 2: 1 3: 2 4: 3 
5: 4 

Average iteration times

53.2s     │ ▇ █ ▆ █ ▆ 
46.5s     │ █ █ █ █ █ 
39.9s     │ █ █ █ █ █ 
33.2s     │ █ █ █ █ █ 
26.6s     │ █ █ █ █ █ 
19.9s     │ █ █ █ █ █ 
13.3s     │ █ █ █ █ █ 
6.6s      │ █ █ █ █ █ 
          └───────────
            1 2 3 4 5 

[█ <current>] 

1: 0 2: 1 3: 2 4: 3 
5: 4 
```

</details>

How to read it?

For the table, we have the following interesting metrics:

- `its`: the number of iterations executed. The greater the more likely reliable the aggregated data is.
- `mem_peak`: the peak memory usage across all samples.
- `mode`: the [KDE mode](https://en.wikipedia.org/wiki/Kernel_density_estimation). The lower the better, simplifying it a bit, it is a better "mean" metric. 
- `rstdev`: the [relative standard deviation](https://en.wikipedia.org/wiki/Coefficient_of_variation): the lower the better. A greater value indicates the code is behaviour differently (e.g. one time takes 1s, another 6s – not great). A way to lower it is to increase the number of iterations (if the code is inherently unstable for example). The more stable the code is the less iterations are needed. Typically we want to stay underneath 5%.

### Benchmark Report improvements

I'm not very proficient with the PHPBench configuration, ideally it would be great to:

- Remove the revision number (`revs`) from the table, which is a feature we cannot really leverage in our case. It's a bit of a lengthy on boring explanation as to why, but we almost guaranteed cannot use it.
- Some units could have a better title.
- The charts have a weird Y axis and the title is generic, so you do not know for what benchmark it is.

But I don't really want to spend more time on this, so someone can improve them at a later time if they wish to.

### Workflow limits

In Box, I went a step further storing the report for the main branch and comparing the PR results with the main branch results, but it's quite a bit more work and required some extra code to do a proper diff between those reports, so I am not keen to go that far here (at least myself, maybe someone else wants to do it).

**So currently, it is only executed in the CI, allowing for quick inspection**. It will not fail if there is a performance regression.

If we think it won't be useful and just a waste of CI, then we can ditch #2502 and only add this to the `Makefile` for now.

### DIFF
--- a/tests/benchmark/MutationGenerator/MutationGeneratorBench.php
+++ b/tests/benchmark/MutationGenerator/MutationGeneratorBench.php
@@ -53,9 +53,7 @@ final class MutationGeneratorBench
 
     public function setUp(): void
     {
-        $generateMutations = require __DIR__ . '/generate-mutations-closure.php';
-
-        $this->main = static fn () => $generateMutations(PHP_INT_MAX);
+        $this->main = (require __DIR__ . '/create-main.php')(PHP_INT_MAX);
     }
 
     public function tearDown(): void
--- a/tests/benchmark/Tracing/TracingBench.php
+++ b/tests/benchmark/Tracing/TracingBench.php
@@ -53,9 +53,7 @@ final class TracingBench
 
     public function setUp(): void
     {
-        $provideTraces = require __DIR__ . '/provide-traces-closure.php';
-
-        $this->main = static fn () => $provideTraces(PHP_INT_MAX);
+        $this->main = (require __DIR__ . '/create-main.php')(PHP_INT_MAX);
     }
 
     public function tearDown(): void


composer.json
```diff
@@ -77,6 +77,7 @@
     "require-dev": {
         "ext-simplexml": "*",
         "fidry/makefile": "^1.0",
+        "phpbench/phpbench": "^1.4",
```
theofidry, line 80: there is the open question of whether this should be here or a separate `composer.json`

maks-rafalko:
I know it's in draft yet, but didn't get an idea how it should fail the build (if it should). Could you please clarify?

theofidry:
@maks-rafalko I didn't plan to add a failure. So the build would execute and we can inspect it as we wish. It has limited use as the jobs logs are not preserved indefinitely, but it's still something.

Unfortunately having a better integration to compare the results between master and the PR is too much work, I don't really want to do it.

If you think it's too much computation waste though we could ditch #2502 and then I would remove the CI part here.

.github/workflows/benchmarks.yaml
```diff
@@ -0,0 +1,77 @@
...
+        uses: ramsey/composer-install@v3
+
+      - name: Run tests
+        shell: bash
+        run: make test-benchmark
+
+  benchmarks:
+      runs-on: ubuntu-latest
+      timeout-minutes: 30
+      name: 'Tests'
```
theofidry:
```suggestion
      name: 'Benchmarks'
```

theofidry:
Todo: add a piece of doc.

Makefile
```diff
@@ -128,6 +139,12 @@ profile_mutation_generator: vendor $(BENCHMARK_MUTATION_GENERATOR_SOURCES)
 		php tests/benchmark/MutationGenerator/profile.php
 	composer dump
 
+.PHONY: benchmark_mutation_generator
+benchmark_mutation_generator: vendor $(BENCHMARK_MUTATION_GENERATOR_SOURCES)
+	composer dump --classmap-authoritative --quiet
+	vendor/bin/phpbench run tests/benchmark/MutationGenerator $(PHPBENCH_REPORTS)
+	composer dump
```
maks-rafalko, line 145: could you please explain why do we need composer dump here?

theofidry: Otherwise you're left with the result of `composer dump --classmap-authoritative`. It is not a problem in the CI, as nothing is done after the benchmark, but it will cause issues during development locally (we do the same when profiling with Blackfire in case you did not notice)

composer.json
theofidry: Going with that approach for now.

PR #2517, opened 2025-10-29
chore(benchmark): Fix usage of the deprecrated Blackfire `--samples` option

theofidry:
The option `--samples` is now a noop option. To leverage samples we are now expected to do it ourselves.

In our case it is not that simple, because we need to re-do the setup step for every execution of the profiled script. Indeed, the profiled code is a non rewindable generator.

As a result, in this PR, the closure script no longer does the setup + return the function to profile, but instead returns a closure doing that.

### DIFF
--- a/devTools/phpstan.neon
+++ b/devTools/phpstan.neon
@@ -124,9 +124,14 @@ parameters:
         -
             message: '#Call to function iterator_.* on a separate line has no effect#'
             path: ../tests/*
+
+        # PHPStan fails to detect that we have at least one loop hence the result cannot be null.
         -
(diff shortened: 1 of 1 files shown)


tests/benchmark/MutationGenerator/create-main.php
```diff
@@ -44,57 +45,68 @@
 
 require_once __DIR__ . '/../../../vendor/autoload.php';
 
-$container = Container::create();
+/**
+ * @return iterable<SplFileInfo>
+ */
+function collectSources(): iterable
```
theofidry: extracted to https://github.com/infection/infection/pull/2537

PR #2521, opened 2025-10-29
refactor(Str): Move Str to the Framework namespace

theofidry:
Extracted from #2504 to reduce the noise.

A few utilities have been moved to `Framework` which I think is a better place for re-usable utilities. `Framework` being better (IMO) than `Utility`, although I appreciate it is subjective.


theofidry:
Since it is extracted from https://github.com/infection/infection/pull/2504 which has already been approved I'll be merging this without waiting for approval.

PR #2522, opened 2025-10-29
refactor(Str): Introduce line ending replacement utilities

theofidry:
Extracted from #2504 for better visibility.

Currently we have:

- `normalizeLineReturn()` which replaces the line endings of the value by the unix line endings.
- `LineReturnNormalizer::normalize()` which replaces the line endings of the value by the system line endings. Note that the implementation was flawed in that if it received a non-unix line endings, it would not replace it by the system line ending.

The initial proposition for the replacement was:

- `normalizeLineReturn()` -> `Str::toLinuxLineReturn()`.
- `LineReturnNormalizer::normalize()` -> `Str::toSystemLineReturn()`.

I decided to change "linux" for "unix", which is more universal and "line return" for "line endings" which is more commonly used (and for example is the term used by PHPUnit). So the final replacement is:

- `normalizeLineReturn()` -> `Str::toUnixLineEndings()`.
- `LineReturnNormalizer::normalize()` -> `Str::toSystemLineEndings()`.

Both implementations have been consolidated and tests were added.


src/Framework/Str.php
```diff
@@ -52,6 +53,39 @@ final class Str
...
+
+    private const UNIX_LINE_ENDINGS_REPLACEMENT = [
+        "\r\n" => "\n",
+        "\r" => "\n",
+    ];
+
+    /**
+     * @psalm-suppress InvalidReturnStatement,InvalidReturnType
+     */
+    public static function toSystemLineEndings(string $value): string
```
theofidry, line 70: I changed the implementation to use `strtr()` which seems better, although a bit more verbose. Since the implementation remains straightforward and that verbosity is strictly internal to this class, I paid no mind to it.

theofidry:
Given this is extracted from #2504 which was approved I will not wait for a review for this PR.

sanmai (approved):

PR #2488, opened 2025-10-26
test: Refactor the ConfigurationFactoryTest

theofidry:
# Problem

The `ConfigurationFactoryTest` uses 40+ parameters in its test methods, making it difficult to understand test scenarios
  and maintain the test suite. It is difficult to read, difficult to understand, hard to update an typically we skipped a few parameters/values because it was too much pain. Also, updating any scenario would make the diff unreadable leading to having to trust the change.

# Solution

This PR attempts to improve a bit the DX and readability for the `ConfigurationFactoryTest`. I say attempt, as, although I think it is an improvement, it is certainly not as easy as I would like. Unfortunately, at the end of the day, it is still a complicated test.

This is a massive diff. If you wish to review it in details, be my guest, but I propose to first walk you through some of the key changes. Unfortunately git thinks the test has not been moved + updated but that the new and old file are completely unrelated, so some of the diffs may be hard to see. I'll provide a few sample to showcase the impact.

**There is no test removed or added.** It is a pure refactoring.

## Introduce a `Scenario` object

The scenario object is meant to be use parameter instead of X arguments for the test method. It serves two purposes:

- Provide a typed API.
- Provide a more friendly API to create new scenarios.

What the scenario is (minus the various methods which are here to construct new scenarios):

```php
final class ConfigurationFactoryScenario
{
    public function __construct(
        public bool $ciDetected,
        public bool $githubActionsDetected,
        public SchemaConfigurationBuilder $schemaBuilder,
        public ConfigurationFactoryInputBuilder $inputBuilder,
        public Configuration $expected,
    ) {
    }

    // ...
}
```

I picked that terminology from Behat. I feel it better captures the intent and is more flexible/correct than "dataSet" which we might have used in the past.

## Impact of the scenario for the test method

Usage of the scenario for the test method:

```php
    #[DataProvider('valueProvider')]
    public function test_it_can_create_a_configuration(
        ConfigurationFactoryScenario $scenario,
    ): void {
        $schema = $scenario->schemaBuilder->build();

        $actual = $this
            ->createConfigurationFactory(
                $scenario->ciDetected,
                $scenario->githubActionsDetected,
                $schema,
            )
            ->create(...$scenario->inputBuilder->build($schema))
        ;

        $this->assertEquals(
            $scenario->expected,
            $actual,
        );
    }
```

I will not bother with an unreadable diff, but this is a showcase of how it was before:

<details>

<summary>Previous `ConfigurationFactoryTest::test_it_can_create_a_configuration()`</summary>

[...]

PR #2539, opened 2025-11-01
test(Builder): Add more builders

theofidry:
Originally was #2488 but in there I tackled `ConfigurationFactoryTest`, which is quite a beast and I would like to handle separately.

So this PR is a continuation of #2482. As a reminder, the goal is:

- Remove the need of mocks for `@final` value objects.
- Remove boilerplate to compare value objects.
- Make the value objects `final`.

Unless I missed something, this PR enables #2483.

Note that although this adds quite a bit of code, it is mostly 1) the builders themselves, for which the implementation is indeed verbose 2) assertions + builders tests. The tests code that is updated to use builders is otherwise simpler.


tests/phpunit/Mutant/MutantBuilder.php
```diff
@@ -37,22 +37,184 @@
...
+        return new self(
+            $mutant->getFilePath(),
+            $mutant->getMutation(),
+            $mutant->getMutatedCode(),
+            $mutant->getDiff(),
+            $mutant->getPrettyPrintedOriginalCode(),
+        );
+    }
+
+    public static function materialize(
```
theofidry, line 75: mark it as deprecrated

sanmai (approved):

PR #2542, opened 2025-11-01
refactor: Improve the source filtering readability

theofidry:
This PR essentially moves the whole filter + git filter + git options block to a separate method. Then a bit more stuff is extracted for readability purposes.

I believe on its own it helps a bit to follow what is going on there. My main motivation behind this is to introduce a value object to make a lot simpler to navigate through that piece of code (see #2395).

### DIFF
--- a/src/Command/RunCommand.php
+++ b/src/Command/RunCommand.php
@@ -495,44 +495,12 @@ final class RunCommand extends BaseCommand
             );
         }
 
-        $gitDiffFilter = $input->getOption(self::OPTION_GIT_DIFF_FILTER);
-        $isForGitDiffLines = (bool) $input->getOption(self::OPTION_GIT_DIFF_LINES);
-        $gitDiffBase = $input->getOption(self::OPTION_GIT_DIFF_BASE);
-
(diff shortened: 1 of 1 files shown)


src/Command/RunCommand.php
```diff
@@ -699,4 +667,92 @@ private static function getFormatterName(InputInterface $input): FormatterName
...
+    }
+
+    private static function assertGitBaseHasRequiredFilter(
+        ?string $gitDiffFilter,
+        bool $isForGitDiffLines,
+        ?string $gitDiffBase,
+    ): void {
+        if ($gitDiffBase !== Container::DEFAULT_GIT_DIFF_BASE
+            && $gitDiffFilter === Container::DEFAULT_GIT_DIFF_FILTER
+            && $isForGitDiffLines === Container::DEFAULT_GIT_DIFF_LINES
```
sanmai, line 729: I guess you can fully test these by calling them with reflection.

sanmai (approved):

src/Command/RunCommand.php
theofidry: Probably. I'm not too keen on trying to test this until #2488 at the very least. Until then (and it may still be), any sort of test is bound to be extremely hard to read.

I'm seeking to update this though so I may need to do this, we will see.

PR #2549, opened 2025-11-04
test: Remove useless test

theofidry:
Like in #2538 and #2539, removes useless test that does not test anything.

Note that we do have a test, its factory, which checks the state of the schema for a given file.

### DIFF
--- a/devTools/phpstan-baseline.neon
+++ b/devTools/phpstan-baseline.neon
@@ -583,12 +583,6 @@ parameters:
 			path: ../tests/phpunit/Configuration/Schema/SchemaConfigurationFactoryTest.php
 
 		-
-			message: '#^Method Infection\\Tests\\Configuration\\Schema\\SchemaConfigurationTest\:\:test_it_can_be_instantiated\(\) has parameter \$mutators with no value type specified in iterable type array\.$#'
-			identifier: missingType.iterableValue
-			count: 1
-			path: ../tests/phpunit/Configuration/Schema/SchemaConfigurationTest.php
-
-		-
 			message: '#^Method Infection\\Tests\\Console\\E2ETest\:\:runInfection\(\) has parameter \$argvExtra with no value type specified in iterable type array\.$#'
 			identifier: missingType.iterableValue
 			count: 1


maks-rafalko:
> hat does not test anything.

why it doesn't test anything? Didn't get. What if we remove Factory at some point?

theofidry:
@maks-rafalko removing the factory will not change anything. Currently this test does only two things:

- Indirectly test the the assertion (which is otherwise not use and that we do not need anymore thanks to #2487 and the builders) – note that for the few assertions remaining I added test when adding one. They are still useful for objects that contains lazy or computed data like `Trace` or `Mutant`.
- Test the object getters.

Testing pure getters (i.e. with no extra logic) has always been a bit debatable. With https://github.com/infection/infection/pull/2483 there won't be any getter at all so I kinda fail to see what purpose this test would have.

I didn't wait for approval because it's the exact same work as #2538 which has already been merged.

maks-rafalko:
> Test the object getters.

isn't it what factory test also does (implicitly, under the hood)? What is the difference then?

So the main question is why we say that `SchemaConfiguration`'s test is useless, while we test this class via it's factory which basically also creates an instance of `SchemaConfiguration` and checks it's properties (be it via public getters or public properties IMO doesn't matter)? I miss the point I guess.

theofidry:
>isn't it what factory test also does (implicitly, under the hood)? What is the difference then?

Not quite no, if that was the case we wouldn't need a factory in the first place. The schema factory creates a `SchemaConfiguration` object from a configuration file path and its JSON decoded content. It has some type juggling logic as well as a few normalization steps. So there is definitely actual logic to be tested there.

>So the main question is why we say that SchemaConfiguration's test is useless

If your class looks like this:

```php
final readonly class SimpleValueObject {
  public function __construct(
    public string $value1,
    public string $value2,
  ) {}
}
```

IMO there is nothing to test there. Yes you can write a typo, not have it readonly or final, maybe it should implement a specific interface, but this are pure language features, is it really something you want to test?

Now if there was some validation, or a tiny bit of logic, for example `Infection\Configuration\Entry\PhpUnit` which is a super simple value object too, but has `::withAbsolutePaths()` which does some stuff, then I think it's valuable to test that piece of code.

maks-rafalko:
> Now if there was some validation, or a tiny bit of logic

`SchemaConfiguration` constructor:

```
Assert::nullOrGreaterThanEq($timeout, 0);
Assert::nullOrOneOf($testFramework, TestFrameworkTypes::getTypes());
Assert::nullOrOneOf($staticAnalysisTool, StaticAnalysisToolTypes::getTypes());
```

so it has some validation (=logic). 

Well, I don't say that we need to revert it, I just want to understand every piece of changes that we are doing, and their underlying ideas, because I need to maintain it and in a year be able to answer to the question "why this class is excluded from requiring a test"

theofidry:
>so it has some validation (=logic).

Hmm actually that's true. I admit I'm always conflicted when it's those type safeguards.

That being said, we now have PHPStan, and it's an internal piece (the one validating the user input is its factory), so I'll at the very least update the PHPStan types.

PR #2551, opened 2025-11-04
Add support for Symfony 8

Kocal:
This PR increase version constraints on `symfony/*` packages to allow Symfony 8, which will be released this month.

**EDIT:** opened https://github.com/phpbench/phpbench/pull/1125

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -26,7 +26,7 @@ jobs:
         php-version: [ '8.2', '8.3', '8.4', '8.5' ]
         dependencies: [ locked, lowest, highest ]
         coverage-driver: [ pcov, xdebug ]
-        symfony-version: [ '6.4.*', '7.*.*' ]
+        symfony-version: [ '6.4.*', '7.*.*', '8.0.*' ]
         include:
           - operating-system: windows-latest
             php-version: '8.2'
@@ -36,6 +36,10 @@ jobs:
         exclude:
           - dependencies: 'locked'
             symfony-version: '7.*.*'
+          - php-version: '8.2'
+            symfony-version: '8.0.*'
+          - php-version: '8.3'
+            symfony-version: '8.0.*'
 
     name: Tests on ${{ matrix.operating-system }} with PHP ${{ matrix.php-version }} (${{ matrix.dependencies }}; Symfony ${{ matrix.symfony-version }}), using ${{ matrix.coverage-driver }}
 
--- a/composer.json
+++ b/composer.json
@@ -63,10 +63,10 @@
         "sanmai/later": "^0.1.7",
         "sanmai/pipeline": "^7.0",
         "sebastian/diff": "^4.0 || ^5.0 || ^6.0 || ^7.0",
-        "symfony/console": "^6.4 || ^7.0",
-        "symfony/filesystem": "^6.4 || ^7.0",
-        "symfony/finder": "^6.4 || ^7.0",
-        "symfony/process": "^6.4 || ^7.0",
+        "symfony/console": "^6.4 || ^7.0 || ^8.0",
+        "symfony/filesystem": "^6.4 || ^7.0 || ^8.0",
+        "symfony/finder": "^6.4 || ^7.0 || ^8.0",
+        "symfony/process": "^6.4 || ^7.0 || ^8.0",
         "thecodingmachine/safe": "^v3.0",
         "webmozart/assert": "^1.11"
     },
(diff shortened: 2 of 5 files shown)


.github/workflows/tests.yaml
```diff
@@ -33,6 +33,12 @@ jobs:
             dependencies: locked
             coverage-driver: xdebug
             symfony-version: '6.4.*'
+          - operating-system: ubuntu-latest
+            php-version: '8.4'
```
sidz: Ideally, we would like to test against 8.5 too. I'd suggest to add symfony 8 in the `symfony-version` and add to exclude testing against php < 8.4

Kocal: Indeed, that's a better alternative.

However, I had to modify the `composer config minimum-stability dev` step to run when we try to install a dev version of Symfony. Maybe I missed something with the matrix configuration, but it's not possible to **only** configure the old `matrix.minimum-stability` config **only** when using Symfony 8.0.x-dev, right?

sidz: @Kocal it is possible. GA supports `if` condition.

As an example you could take a look at the PR https://github.com/infection/infection/pull/1892 and  what we did there in order to introduce Symfony 7.0 support

Kocal: I mean, directly with the matrix, like I was able to do with `include` from the 1st commit. 

It should be fine now, I see that some Symfony 8 packages are installed in  https://github.com/infection/infection/actions/runs/19081901778/job/54512486280?pr=2551, but there is still PHPBench that force Symfony 7.x for `symfony/console` or `symfony/filesystem` for example.

.github/workflows/tests.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
         php-version: [ '8.2', '8.3', '8.4', '8.5' ]
         dependencies: [ locked, lowest, highest ]
         coverage-driver: [ pcov, xdebug ]
-        symfony-version: [ '6.4.*', '7.*.*' ]
+        symfony-version: [ '6.4.*', '7.*.*', '8.0.x-dev' ]
```
Jean85:
```suggestion
        symfony-version: [ '6.4.*', '7.*.*', '^8.0.0-BETA-1' ]
```

.github/workflows/tests.yaml
```diff
@@ -36,6 +36,10 @@ jobs:
         exclude:
           - dependencies: 'locked'
             symfony-version: '7.*.*'
+          - php-version: '8.2'
+            symfony-version: '8.0.x-dev'
+          - php-version: '8.3'
+            symfony-version: '8.0.x-dev'
```
Jean85:
```suggestion
          - php-version: '8.2'
            symfony-version: '^8.0.0-BETA-1'
          - php-version: '8.3'
            symfony-version: '^8.0.0-BETA-1'
```

.github/workflows/tests.yaml
```diff
@@ -60,6 +64,10 @@ jobs:
         run: |
           composer config platform.php 8.4.99
 
+      - name: Configure Composer minimum-stability to dev
+        if: "endsWith(matrix.symfony-version, '-dev')"
+        run: composer config minimum-stability dev
+
```
Jean85:
```suggestion
      - name: Configure Composer minimum-stability to beta
        if: "endsWith(matrix.symfony-version, '-BETA-1')"
        run: composer config minimum-stability beta

```

Kocal:
Comments addressed, thanks!

.github/workflows/tests.yaml
```diff
@@ -26,7 +26,7 @@ jobs:
         php-version: [ '8.2', '8.3', '8.4', '8.5' ]
         dependencies: [ locked, lowest, highest ]
         coverage-driver: [ pcov, xdebug ]
-        symfony-version: [ '6.4.*', '7.*.*' ]
+        symfony-version: [ '6.4.*', '7.*.*', '^8.0.0-BETA-1' ]
```
sidz: Personally I wouldn't be so strict. For example BETA-2 has been released 3 days ago and we would like to test against newest beta version :)

Jean85: The constraint has `^` in front, so any newer beta would be considered (and preferred).

sidz: my bad @Jean85, I've missed it.

Kocal:
Waiting for https://github.com/phpbench/phpbench/pull/1128 to be released :)

simPod:
https://github.com/phpbench/phpbench/releases/tag/1.4.3

.github/workflows/tests.yaml
simPod: Can use stable now 🫠

Kocal: Yes, this has been modified

Kocal:
PR rebased, squashed, and now testing with Symfony 8 **stable** version :) 

Waiting for the CI to be green... ⏳

maks-rafalko:
thank you @Kocal

Issue #2493, opened 2025-10-27
DX: Improve log output


staabm:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.31.8 (`infection.phar --version`)
| Test Framework version | PHPUnit/PhpSpec 11.5.23
| PHP version | 8.3.27
| Platform    | Ubuntu in GitHub Actions
| Github Repo | phpstan-src

Running infection in GitHub Actions with

```
      - name: "Run infection"
        run: |
          infection \
            --ignore-msi-with-no-mutations \
            --log-verbosity=all \
            --debug \
            --logger-text=php://stdout
```

can lead to the following logs:

```
   ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
 _/ // / / / __/  __/ /__/ /_/ / /_/ / / / /
/___/_/ /_/_/  \___/\___/\__/_/\____/_/ /_/

#StandWithUkraine

Infection - PHP Mutation Testing Framework version 0.31.7

Notice:  You are running Infection with PCOV enabled.

Running initial test suite...

PHPUnit version: 9.6.29

Running initial Static Analysis...

PHPStan version: 2.1.31

.: killed by tests, A: killed by SA, M: escaped, U: uncovered
E: fatal error, X: syntax error, T: timed out, S: skipped, I: ignored


Generate mutants...

Processing source code files...
MMMMMMM
Escaped mutants:
================


1) /home/runner/work/build-infection/build-infection/src/Rules/Deprecations/DefaultDeprecatedScopeResolver.php:23    [M] PHPStan\Infection\TrinaryLogicMutator [ID] 463811730f65f1e71ad3903441d208d1

@@ @@
             return true;
         }
         $function = $scope->getFunction();
-        if ($function !== null && $function->isDeprecated()->yes()) {
+        if ($function !== null && !$function->isDeprecated()->no()) {
             return true;
         }
         return false;
     }
 }

...

Timed Out mutants:
==================

Skipped mutants:
================

Killed by Test Framework mutants:
=================================

Killed by Static Analysis mutants:
==================================

Errors mutants:
===============

Syntax Errors mutants:
======================
Warning: Escaped Mutant for Mutator "PHPStan\Infection\TrinaryLogicMutator":

@@ @@
             return true;
         }
         $function = $scope->getFunction();
-        if ($function !== null && $function->isDeprecated()->yes()) {
[...]

PR #2552, opened 2025-11-04
feat(Logger): Implement GitHubActionsLogTextFileLogger to improve DX of text file logger with GA

staabm:
Closes https://github.com/infection/infection/issues/2493.

As per #2493, the current `TextFileLogger` output is a bit misleading in GitHub Actions when outputting directly to the stdout.

This PR introduces a new `GitHubActionsLogTextFileLogger` logger which is used instead of `TextFileLogger` when we use `--logger-text=php://stdout` in a GitHub Action context.

This logger outputs separate log sections using the GitHub Action [native formatting](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#grouping-log-lines) which allows to expand/collapse the separate parts.

To achieve this, the base implementation of `TextFileLogger` was extracted into a base class `BaseTextFileLogger` which both text loggers now implements.

### DIFF
--- a/src/Logger/FileLoggerFactory.php
+++ b/src/Logger/FileLoggerFactory.php
@@ -85,7 +85,7 @@ class FileLoggerFactory
         }
 
         if ($logConfig->getTextLogFilePath() !== null) {
-            yield $logConfig->getTextLogFilePath() => $this->createTextLogger();
+            yield $logConfig->getTextLogFilePath() => $this->createTextLogger($logConfig);
         }
 
         if ($logConfig->getHtmlLogFilePath() !== null) {
@@ -131,8 +131,20 @@ class FileLoggerFactory
         );
     }
 
-    private function createTextLogger(): LineMutationTestingResultsLogger
+    private function createTextLogger(Logs $logConfig): LineMutationTestingResultsLogger
     {
+        if (
+            $logConfig->getUseGitHubAnnotationsLogger()
+            && $logConfig->getTextLogFilePath() === 'php://stdout'
+        ) {
+            return new GitHubActionsLogTextFileLogger(
+                $this->resultsCollector,
+                $this->logVerbosity === LogVerbosity::DEBUG,
+                $this->onlyCoveredCode,
+                $this->debugMode,
+            );
+        }
+
         return new TextFileLogger(
             $this->resultsCollector,
             $this->logVerbosity === LogVerbosity::DEBUG,
--- a/src/Logger/GitHubActionsLogTextFileLogger.php
+++ b/src/Logger/GitHubActionsLogTextFileLogger.php
@@ -1,0 +1,73 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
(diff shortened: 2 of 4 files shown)


src/Logger/TextFileLogger.php
```diff
@@ -43,13 +43,11 @@
 use Infection\Mutant\MutantExecutionResult;
 use const PHP_EOL;
 use function sprintf;
-use function str_repeat;
-use function strlen;
 
 /**
  * @internal
  */
-final readonly class TextFileLogger implements LineMutationTestingResultsLogger
```
staabm, line 52: extracted a new `BaseTextFileLogger` out of `TextFileLogger` so I could easily overwrite the markdown headline formatting and insert the github action log section wrappers in `GitHubActionsLogTextFileLogger`

no functional changes to `TextFileLogger`

src/Logger/FileLoggerFactory.php
```diff
@@ -131,9 +131,21 @@ private function wrapWithFileLogger(string $filePath, LineMutationTestingResults
         );
     }
 
-    private function createTextLogger(): LineMutationTestingResultsLogger
+    private function createTextLogger(Logs $logConfig): LineMutationTestingResultsLogger
     {
-        return new TextFileLogger(
+        if (
+            $logConfig->getUseGitHubAnnotationsLogger()
+            && $logConfig->getTextLogFilePath() === 'php://stdout'
```
staabm, line 138: the new `GitHubActionsLogTextFileLogger` will be used when cli options are passed in a way which makes the text-formatter write directly into the github action log-stream.

I implemented it in a way, that when we write into a regular file, we still get  markdown formatted file - as I think the github log specific format does not make much sense in a file to be written

staabm:
I did not yet test this in a real project and did not see it yet in action

staabm:
I assumed its markdown because we have this markdown formatted headlines

```
Skipped mutants:
================
```

maks-rafalko:
nah, it's just a delimeter AFAIK

UPD: well, if you think it should be markdown, then we need other changes (change the extension to `*.md`, etc.). Otherwise this simply doesn't work

staabm:
> So I would leave it as is if you don't know other details that I missed

done

staabm:
> Let's create a docs for this?

we already have this in the docs

> The option supports `php://stdout` value which can, for instance, be useful in a CI environment.

https://github.com/infection/site/blob/8d3dafa740183e5d4974e80b228d0115ca94a7a5/src/guide/command-line-options.md?plain=1#L304

and to enable this logger there are no new switches required.

what do you think needs additional explanation?

maks-rafalko:
> what do you think needs additional explanation?

I thought that adding information about this GH feature folding would be useful, otherwise no one knows about it

Anyway, happy to merge, but the build is failing because of Psalm

staabm:
> but the build is failing because of Psalm

good call. fixed, thank you.

src/Logger/BaseTextFileLogger.php
```diff
@@ -0,0 +1,214 @@
...
+        if (!$this->debugVerbosity) {
+            $notes[] = 'Note: Pass `--log-verbosity=all` to log information about killed and errored mutants.';
+        }
+
+        if (!$this->debugMode) {
+            $notes[] = 'Note: Pass `--debug` to log test-framework output.';
+        }
+
+        if ($notes !== []) {
+            foreach ($notes as $note) {
```
theofidry, line 77: is there a reason why we have `$notes` besides `$logs` here? They are both initialised at the beginning so we could instead just have `$logs`.

src/Logger/GitHubActionsLogTextFileLogger.php
```diff
@@ -0,0 +1,71 @@
...
+        string $headlinePrefix,
+        bool &$separateSections,
+    ): string {
+        $results = trim(parent::getResultsLine($executionResults, $headlinePrefix, $separateSections));
+
+        if ($results === '') {
+            return "0 $headlinePrefix mutants" . PHP_EOL;
+        }
+
+        return
```
theofidry: I would prefer we stick to `sprintf`

src/Logger/BaseTextFileLogger.php
theofidry: nvm it's how it was.

@maks-rafalko here's a good example of why I like to split things. This PR effectively has a one pure refactoring + one feature.

If you have a refactoring PR which is to extract the `BaseTextFileLogger`, it is easier to review and focus on what it is: you check the methods visibility, maybe double check that there is no code change, but overall it's (IMO) a specific way to review the PR.

And for the feature PR, it's a lot easier to see what's new code and what needs more attention.

It does mean that this change wouldn't be atomic, so for sure it's trade-offs.

That being said, the feature is still simple and it's already been reviewed twice and it's an external contribution, so it would be impractical to request changes here, but I hope it illustrates a bit better my reasoning (although sometimes I do get caught up in smth and go overboard!).

theofidry (approved):

src/Logger/GitHubActionsLogTextFileLogger.php
```diff
@@ -0,0 +1,71 @@
...
+namespace Infection\Logger;
+
+use Infection\Mutant\MutantExecutionResult;
+use const PHP_EOL;
+use function trim;
+
+/**
+ * @internal
+ */
+final readonly class GitHubActionsLogTextFileLogger extends BaseTextFileLogger
```
theofidry, line 50: reviewing this once more after reading your comments of the original issue, could you link the GA docs?


```suggestion
/**
 * Uses the GitHub Actions line grouping feature to make the output more digestable and collapsable.
 * 
 * @link https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#grouping-log-lines
 *
 * @internal
 */
final readonly class GitHubActionsLogTextFileLogger extends BaseTextFileLogger
```

src/Logger/BaseTextFileLogger.php
staabm: Sorry for that, you are right. 

When back at the computer later today I can separate the refactoring changes

theofidry: Let's not bother here it's fine really. It was more an illustration to a point @maks-rafalko made the other day

maks-rafalko: I absolutely understand when the split is needed, but there was a case were it was not needed :) 

But, I must admit I'm not always agree with the need of this split, because of 1) time 2) OSS nature 3) other conditions. It's not like "yes or no", it's a more complex question.

Anyway, thanks for sharing

staabm:
separated the PR into 3 commits (2 refactoring, and 1 implementing the new feature).

adressed feedback.

thank you

theofidry:
Thank you @staabm!

@maks-rafalko since this was rebased with multiple clean commits I enabled the merge (as opposed to squash) for this PR

staabm:
just verified the implementation in action:

grafik

---

grafik

---

if you want to test it, see the "Run infection" step in this action: https://github.com/phpstan/build-infection/actions/runs/19402296810/job/55511543333?pr=38
(link will break in the future, when the github action logs got cleaned up)

theofidry:
That's really neat! I wonder if we could make the diff better too (coloured)

staabm:
> if we could make the diff better too (coloured)

[github action logs support ANSI colors](https://github.blog/news-insights/product-news/a-better-logs-experience-with-github-actions/#opening-the-door-to-a-more-colorful-experience) - this should be possible

staabm:
[opened a new issue](https://github.com/infection/infection/issues/2582) so its not forgotten

PR #2553, opened 2025-11-04
Get rid of unnecessary iterator_to_array() call

staabm:
IIRC `iterator_to_array()` creates a unncessary copy of the underlying structure


maks-rafalko (approved):

theofidry:
>creates a unncessary copy of the underlying structure

What do you mean by that?

staabm:
> What do you mean by that?

consuming a iterator directly with foreach is more efficient in comparison to copying the iterator into an array and loop over the array

theofidry (approved):

theofidry:
Yes ok, I thought you meant something else. All good! Thank you @staabm

PR #2554, opened 2025-11-04
[POC] Run InitialTestSuite/InitialStaticAnalysis only when mutations found

staabm:
initial static test run and initial static analysis run can take up to ten's of minutes (in phpstan-src) case.

since phpstan-src uses only very few [self-implemented mutators tailored for real world problems in the PHPStan api](https://github.com/phpstan/build-infection/tree/1.x/src/Infection) we see a few PRs coming in, in which no mutations will be generated at all (which is fine).

but since initial test/static analysis processes still take a lot of time, our CI is working 15 minutes, to finally conclude that no mutations have been found and nothing needs to be killed.

with this PR we are proposing running the initial processes only after we identified that actual mutations could be found.

please take this PR as a inspiration for discussion.

### DIFF
--- a/src/Engine.php
+++ b/src/Engine.php
@@ -44,6 +44,7 @@ use Infection\Event\EventDispatcher\EventDispatcher;
 use Infection\Metrics\MetricsCalculator;
 use Infection\Metrics\MinMsiChecker;
 use Infection\Metrics\MinMsiCheckFailed;
+use Infection\Mutation\Mutation;
 use Infection\Mutation\MutationGenerator;
 use Infection\PhpParser\Visitor\IgnoreNode\NodeIgnorer;
 use Infection\Process\Runner\InitialStaticAnalysisRunFailed;
@@ -89,19 +90,24 @@ final readonly class Engine
      */
     public function execute(): void
     {
-        $initialTestSuiteOutput = $this->runInitialTestSuite();
-        $this->runInitialStaticAnalysis();
+        $mutations = $this->runMutationGeneration();
+        $numberOfMutants = IterableCounter::bufferAndCountIfNeeded($mutations, $this->config->noProgress());
 
-        /*
-         * Limit the memory used for the mutation processes based on the memory
-         * used for the initial test run.
-         * This is done AFTER static analysis to avoid restricting PHPStan's memory.
-         */
-        if ($initialTestSuiteOutput !== null) {
-            $this->memoryLimiter->limitMemory($initialTestSuiteOutput, $this->adapter);
+        if ($numberOfMutants > 0) {
+            $initialTestSuiteOutput = $this->runInitialTestSuite();
+            $this->runInitialStaticAnalysis();
+
+            /*
+             * Limit the memory used for the mutation processes based on the memory
+             * used for the initial test run.
+             * This is done AFTER static analysis to avoid restricting PHPStan's memory.
+             */
+            if ($initialTestSuiteOutput !== null) {
+                $this->memoryLimiter->limitMemory($initialTestSuiteOutput, $this->adapter);
+            }
+
+            $this->runMutationAnalysis($mutations, $numberOfMutants);
         }
-
-        $this->runMutationAnalysis();
 
         try {
             $this->minMsiChecker->checkMetrics(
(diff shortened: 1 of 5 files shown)


maks-rafalko:
Interesting idea. 

Just to give you some thoughts as an alternative - have you considered `--skip-initial-tests` and providing PHPUnit coverage to Infection? 

https://github.com/infection/infection/blob/500e0bc90ae861e73537145c07bb06e311819f5c/.github/workflows/mt.yaml#L81

ondrejmirtes:
That would be nice because we could provide the coverage data from paraunit/paratest which runs much faster.

staabm:
> Just to give you some thoughts as an alternative - have you considered `--skip-initial-tests` and providing PHPUnit coverage to Infection?

I did so with https://github.com/phpstan/phpstan-src/pull/4525
it improved our CI time from 15m 36s => 12m 26s

---

I am not sure yet whether having some logic which can skip the intial-step as a whole is more useful, then this small improvement when running it beforehand in parallel without a chance of skipping it anymore.

---

side-note: what I also noticed [while looking into this is](https://github.com/phpstan/phpstan-src/actions/runs/19078182639/job/54499398288?pr=4525), merging coverage information seems to be a very time intensive task:

grafik

4m40s test suite runtime + 1m coverage merging

maks-rafalko:
It still runs "Running initial Static Analysis...", for which we don't have `--skip-initial-***` analogue.

But I don't see in the output how much time is spent on this step. Perhaps, if you have PHPStan cache beforehand, we can add such an option and skip "Running initial Static Analysis..." step as well

maks-rafalko:
> I am not sure yet whether having some logic which can skip the intial-step as a whole is more useful, then this small improvement when running it beforehand in parallel without a chance of skipping it anymore.

yeah, probably just skipping the whole "initial runs" step is much more effective in phpstan case

staabm:
> It still runs "Running initial Static Analysis...", for which we don't have `--skip-initial-***` analogue.
> 
> But I don't see in the output how much time is spent on this step. Perhaps, if you have PHPStan cache beforehand, we can add such an option and skip "Running initial Static Analysis..." step as well

while such option would make sense, it will not help for this particular case in the phpstan-src repo - as with every commit the phpstan revision changes and therefore the initial run will always be with a invalid result-cache.
(or we need to re-store it from other actions running beforehand)

src/Engine.php
```diff
@@ -89,20 +90,25 @@ public function __construct(
...
-
-        /*
-         * Limit the memory used for the mutation processes based on the memory
-         * used for the initial test run.
-         * This is done AFTER static analysis to avoid restricting PHPStan's memory.
-         */
-        if ($initialTestSuiteOutput !== null) {
-            $this->memoryLimiter->limitMemory($initialTestSuiteOutput, $this->adapter);
+        $mutations = $this->runMutationGeneration();
+        $numberOfMutants = IterableCounter::bufferAndCountIfNeeded($mutations, $this->config->noProgress());
```
sanmai, line 94: We could probably look at the first one `take($mutations->peek(1))->toList()` and discard the entire thing right after.

staabm: do you mean your suggestion as a separate optimization unrelated to this PR?

sanmai: `IterableCounter::bufferAndCountIfNeeded()` should not be needed here as it will load all mutations, conditionally. I understand you are running this against a small subset of changes, but consider that this branch will trigger for anyone running Infection against their whole codebase. 

I think it will be easier to leave the existing mutation process intact, and run the mutation generation here just to fetch one single mutation, and only. If there are no mutations, then we can abort right here. If there's just one mutation, we do not need to generate any more than that.

staabm: This line was 1:1 moved from MutationTestingRunner to this location. 

Shouldn't your idea also work before this PR than?

theofidry:
Whilst I find the idea interesting, it has a few caveats.

Unless you're executing infection against the whole codebase, in an incremental run, ideally, you would only mutate eligible code. To know if the code is eligible a bunch of stuff is necessary:

1. the code changed
2. the code is covered
3. the code is not ignored/excluded
4. maybe more? For example we could at some point leverage richer types coming from SA to enrich the AST to provide more info to the mutators to make it easier to prevent issuing unproductive mutants.

Excluding the 4th point, which is purely theoretical at the moment, for achieving 2. you already need the coverage. A user may already provide the coverage, in which case the initial test suite run is a safety check, but otherwise it cannot be skipped and is necessary in order to generate any mutation.

So this does not meant this idea cannot work, but there is a few more checks to do.

For the SA run, atm we do not have an equivalent to coverage AFAIK hence until 4) is implemented, this won't be an issue either.

Another aspect as well is the UI. Right now it's easy to report things to the stdout since they are done in order, but now it might lead to a bit of a confusing reporting?

staabm:
> It still runs "Running initial Static Analysis...", for which we don't have `--skip-initial-***` analogue.
>
> But I don't see in the output how much time is spent on this step. Perhaps, if you have PHPStan cache beforehand, we can add such an option and skip "Running initial Static Analysis..." step as well

@maks-rafalko thinking about this again... in case the PHPStan cache already is warmed, the initial PHPStan run will likely only take a few seconds.. so I think its not worth bothering for this use-case.

> To know if the code is eligible a bunch of stuff is necessary:
> 
>     1. the code changed
> 
>     2. the code is covered

ohh thats a goos point. to determine which code should be mutated, we already need coverage information. that means this PR does not make sense.

thanks for the feedback to all of you

PR #2543, opened 2025-11-02
PoC XML report rework

theofidry:
**This PR is a PoC/Draft, not ready for review.**

https://github.com/infection/infection/issues/2816

## The problem

There is some architectural changes (see [Architecture PoC](https://github.com/infection/infection/pull/2272/files#diff-9bc3105e506d7103f2b25138650f3e784a373d04b529f12ff34136a2fccc72c6)) that I would like to bring. To achieve them, we need a service that takes care of generating a trace, i.e. something like:

```php 
interface Tracer {
  trace(SplFileInfo $fileInfo): Trace;
}
```

While investigating a way to do this, I re-discovered that the current process to generate a trace is extremely complex. Notably going through the various coverage reports and updating the trace at various stages is quite hard to follow.

**The main issues I have with the current implementation are**:

- The implementation feels leaky and all over the place (relatively subjective).
- There is no overview of what the final result really is (we have a lot of small tests, but none for which we know, for a given coverage report, what are the final traces we get and what values they hold).
- The implementation is very tightly coupled to the current format of coverage report. It makes sense, but equally it is a piece subject to change. The current coverage report is extremely inefficient (heavy, lots of files, inadequate format, missing pieces of data...). We do with what we have, but there is room for drastically improving this piece. However, doing so means rewriting this entire part of the application, and as per 2) it would be hard to assess the final changes.

**The goal of this PR is to provide a PoC to bring some of those changes, see what is missing and what should be extracted. Among other things, another side-objective is to collect a more exhaustive list of requirements of the constraints we want for a new code coverage report.**

## The solution

The idea of this PR is to introduce a more comprehensive abstraction to interpret the PHPUnit XML (+JUnit) report. Each files is represented by its respective `*Report` class, which takes care of providing the consumer information. Essentially, we are merging the report + reader + parser into one. This allows to have a more cohesive API, and makes optimising the process simpler.

## Status

- [ ] Collect the report requirements (partially done; requires formatting and present this information in a dedicated issue)
- [ ] Extract various changes introduced on the way that are required for this but make sense on their own.
  - [ ] Extract the SafeXPath changes
  - [ ] Extract the report locators changes 
- [x] Introduce an integration test in the old architecture to ensure we are not loosing any feature. This test allows to see what a trace looks like for a given report which is also useful on its own without any other changes. See https://github.com/infection/infection/pull/2444.
[...]

PR #2559, opened 2025-11-05
refactor(SafeDOMXPath): Use the XML namespace

theofidry:
While working on #2543, I noticed that `DOMDocument::load(string $filename)` is a bit faster than getting the file content and using `DOMDocument::loadXML(string $xml)`.

The main issue is... namespaces. When loading the string, we can, beforehand, remove the namespaces found. This allows us to do queries without caring about them at all.

However, if you load the document from the file, then this solution is out of the window. You need to register a namespace and then uglify all of your queries with it.

So this PR focuses on switching to namespaces. Switching to `DOMDocument::load()` will be done separately.

<hr/>

You can find the results of the benchmark in #2561.

Personally, I hate either approach: I find ugly to load the file to remove the namespaces, but the namspaces are a b**** to handle and makes queries ugly.

So either we think the performance benefit is worth the tradeoff and we merge this PR, or we don't in which case I'll adapt #2543 to stick to loading the file first and stripping namespaces from the string.

### DIFF
--- a/src/Process/Runner/ParallelProcessRunner.php
+++ b/src/Process/Runner/ParallelProcessRunner.php
@@ -94,8 +94,8 @@ class ParallelProcessRunner implements ProcessRunner
     public function run(iterable $processContainers): iterable
     {
         /*
-         * It takes about 100000 ms for a mutated process to finish, where it takes
-         * about 5000 ms to make it. Therefore instead of just waiting we can produce
+         * It takes about 100,000 ms for a mutated process to finish, where it takes
+         * about 5,000 ms to make it. Therefore, instead of just waiting, we can produce
          * new processes so that when a process or several finish, we would have
          * additional jobs on hand, without a need to wait for them to be created.
          *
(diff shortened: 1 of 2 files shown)


tests/phpunit/TestFramework/Coverage/XmlReport/XPathFactoryTest.php
```diff
@@ -44,8 +44,11 @@ final class XPathFactoryTest extends TestCase
 {
     public function test_it_removes_namespace(): void
```
sanmai:
```suggestion
    public function test_it_retains_namespace(): void
```

src/TestFramework/Coverage/XmlReport/XPathFactory.php
```diff
@@ -48,26 +47,22 @@ final class XPathFactory
...
-    {
-        /** @var string $cleanedXml */
-        $cleanedXml = preg_replace('/xmlns=\".*?\"/', '', $xml);
+        // @phpstan-ignore property.nonObject
+        $namespaceUri = $document->documentElement->namespaceURI;
+        Assert::notNull($namespaceUri, 'Expected the first document element to have a namespace URI. None found.');
 
-        Assert::string($cleanedXml);
+        $xPath = new SafeDOMXPath($document);
+        $xPath->registerNamespace($namespace, $namespaceUri);
```
sanmai, line 64: I wonder why this mutation escapes.

theofidry: I also wonder why the annotation is messed up again too

theofidry: well actually the mutation escaping is simple: we do not have a test checking it. For simplicity I'll move it to the `SafeDOMXPath` in a future PR so it wouldn't be a problem

maks-rafalko:
Just 2 cents: IMO it makes sense to use namespace AND switch to `DOMDocument::load()` in one PR, not sure why we split it.

Imagine looking into a git history 5 years later, in one commit we will see "Use the XML namespace" - but **why** it's done is not clear from neither commit headline nor commit message body. We will have to open this PR and read the explanation here. Or, at least, as a workaround we need to add the whole description of this PR to the commit message body.

I appreciate extracting different things from big PRs, but this one is overextracted IMO 🤣

theofidry:
>not sure why we split it.

A mix of habits and self-defense.

In general I like to split by scope, I find it easier to review or come back later. But yes sometimes it gets a bit too watered down. I'd have included both but got a bit lost in the late hours of the night 😅 . Sometimes it's autopilot kicking in and confusing scope with one type of change.

>but why it's done is not clear from neither commit headline nor commit message body. We will have to open this PR and read the explanation here.

That is very true, on the other hand, I feel like you'll always be missing some amount of why or insight without looking up the repo PR/discussion. That being said, I try to still refer to the stories in the description and if they get merged, the referenced number will be mentioned in that commit.

Another route would be some feature branches, but honestly even keeping up the drafts up to date is pretty demanding, I don't think I'd manage to maintain feature branches without missing a lot personally.

Anyway looking forward to get that XML rabbit hole buried.

PR #2560, opened 2025-11-06
refactor(ParallelProcessRunner): Refactor ParallelProcessRunner for clarity

sanmai:
I refactored this in #2322, but a few weeks later it hard to get through. Let's make it better to understand what is going on by renaming the `wait()` method to `sleepRemaining()` with named parameter usage (`timeSpentDoingWork:`), making the adaptive polling mechanism self-documenting. 

The new name eliminates the ambiguity of the generic "wait" terminology.

No behavioral changes.


src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -187,9 +190,10 @@ protected function fillBucketOnce(SplQueue $bucket, Iterator $input, int $thread
     }
 
     /**
+     * Adaptive polling: sleep for remaining poll interval after accounting for work done.
      * @param int $timeSpentDoingWork Time to subtract from the poll time when we did some work in between polls
      */
-    protected function wait(int $timeSpentDoingWork): void
+    protected function sleepRemaining(int $timeSpentDoingWork): void
```
theofidry, line 197: is there a reason those methods are `protected`?

src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -95,7 +95,7 @@ public function run(iterable $processContainers): iterable
     {
         /*
          * It takes about 100000 ms for a mutated process to finish, where it takes
```
theofidry:
```suggestion
         * It takes about 100,000 ms for a mutated process to finish, where it takes
```

src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -95,7 +95,7 @@ public function run(iterable $processContainers): iterable
...
-         * about 5000 ms to make it. Therefore instead of just waiting we can produce
+         * about 5000 ms to make it. Therefore, instead of just waiting, we can produce
```
theofidry:
```suggestion
         * about 5,000 ms to make it. Therefore, instead of just waiting, we can produce
```

src/Process/Runner/ParallelProcessRunner.php
theofidry, line 103:
```suggestion
         * we use a generator here which is both non-rewindable and will fail loudly if tried.
```

theofidry, line 259: why is it passed by reference?

src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -166,6 +168,7 @@ protected function hasProcessesThatCouldBeFreed(int $threadCount): bool
      *
      * @param SplQueue<MutantProcessContainer> $bucket
      * @param Iterator<MutantProcessContainer> $input
+     * @return int Milliseconds spent doing work
```
theofidry, line 172: it seems strange the CS fixer is not picking up the missing blank line in-between `@param` and `@return` 🤔

src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -166,6 +168,7 @@ protected function hasProcessesThatCouldBeFreed(int $threadCount): bool
      *
      * @param SplQueue<MutantProcessContainer> $bucket
      * @param Iterator<MutantProcessContainer> $input
```
theofidry, line 170:
```suggestion
     * @param Iterator<MutantProcessContainer> $input
     * @param positive-int $threadCount
```

src/Process/Runner/ParallelProcessRunner.php
theofidry, line 176: Isn't the SA picking up on that or is it really necessary?

theofidry, line 76:
```suggestion
     * @param positive-int $threadCount
     * @param int $poll Delay (in milliseconds) to wait in-between two polls
```

theofidry, line 115: reviewing this, I don't quite get why this is necessary. It seems what separates this statement from the next bucket filling is fairly fast stuff, hence for the sake of simplicity this ~~micro~~ nano-optimisation may not be worth

theofidry, line 249:
```suggestion
        $mutantProcessContainer->getCurrent()->getProcess()->start(
            null,
            [
                'INFECTION' => '1',
                'TEST_TOKEN' => $threadIndex,
            ],
        );
```

theofidry, line 71: this can only be `posivite-int` too here no?

theofidry, line 244:
```suggestion
    /**
     * @param positive-int $threadIndex
     */
    private function startProcess(MutantProcessContainer $mutantProcessContainer, int $threadIndex): void
```

theofidry, line 112: another suggestion to try to make this class a bit more digestable would be to have a bucket class decorating `SplQueue`, i.e. an adapter, which could encapsulate a bit more of our usage in its API. For instance `fillBucketOnce()` could very well belong there (I don't think giving a clock to the bucket is a problem).

Maybe there is a bit more state that could be moved there too, it's a bit hard for me to say without trying it a bit. But if we can this would allow this class to focus on the manipulation of the processes, i.e. creating the processes and interpreting the results, rather than having to deal with the bucket itself.

theofidry, line 57: this has an interface, is there a need for it to be `@final` rather than `final`?

src/Process/Runner/ParallelProcessRunner.php
sanmai: Yes, for mocking in tests.

src/Process/Runner/ParallelProcessRunner.php
sanmai: It can be a very large array and we do not want to trigger any copy-on-write semantics. There's also a chance this is a hangover from a previous implementation.

sanmai: MT is adding a negative integer here IIRC.

sanmai: Yep, that's my intention is to add a decorator to SplQueue in a follow-up.

sanmai: In-test mocking, as with `protected` for some methods

sanmai:
@theofidry What other blockers do you see here? I don't want this to spill out beyond the intended goal.

src/Process/Runner/ParallelProcessRunner.php
theofidry: Is that needed? I would think mocking the clock would be sufficient

theofidry:
@sanmai there's no blocker :) Anything I mentioned was just suggestions based on a fresh review, feel free to ignore them or address them separately

theofidry (approved):

PR #2562, opened 2025-11-06
refactor(ParallelProcessRunner): extract a decorator of SplQueue 

sanmai:
This PR:

- [x] Introduces `ProcessQueue` to encapsulate SplQueue operations, [addressing the suggestion](https://github.com/infection/infection/pull/2560#discussion_r2497745899) to make `ParallelProcessRunner` more digestible by separating bucket management from process orchestration. 
- [x] Covered by tests

The new class absorbs `fillBucketOnce()`  (renamed to `enqueueFrom()`), along with its timing measurements via injected `DuoClock`, allowing the runner to focus on process lifecycle (creation, execution, result interpretation) rather than queue mechanics. This eliminates parameter  threading of `$bucket` through methods, improves testability further through dependency injection, and makes the runner's control flow more linear and easier to follow.

### DIFF
--- a/tests/phpunit/Process/Runner/ProcessQueueTest.php
+++ b/tests/phpunit/Process/Runner/ProcessQueueTest.php
@@ -98,8 +98,7 @@ final class ProcessQueueTest extends TestCase
 
     public function test_enqueue_from_with_valid_iterator_enqueues_item(): void
     {
-        $clock = new TimeSpy();
-        $queue = new ProcessQueue($clock);
+        $queue = new ProcessQueue();
 
         $container = $this->createMock(MutantProcessContainer::class);
         $iterator = $this->createMock(Iterator::class);
@@ -160,8 +159,7 @@ final class ProcessQueueTest extends TestCase
 
     public function test_enqueue_from_accepts_items_when_below_capacity(): void
     {
-        $clock = new TimeSpy();
-        $queue = new ProcessQueue($clock);
+        $queue = new ProcessQueue();
 
         // Add one item
         $queue->enqueue($this->createMock(MutantProcessContainer::class));
@@ -242,8 +240,7 @@ final class ProcessQueueTest extends TestCase
 
     public function test_enqueue_from_advances_iterator(): void
     {
-        $clock = new TimeSpy();
-        $queue = new ProcessQueue($clock);
+        $queue = new ProcessQueue();
 
         $container = $this->createMock(MutantProcessContainer::class);
         $iterator = $this->createMock(Iterator::class);
@@ -261,8 +258,7 @@ final class ProcessQueueTest extends TestCase
 
     public function test_enqueue_from_with_default_max_depth_of_one(): void
     {
-        $clock = new TimeSpy();
-        $queue = new ProcessQueue($clock);
+        $queue = new ProcessQueue();
 
         $container = $this->createMock(MutantProcessContainer::class);
         $iterator = $this->createMock(Iterator::class);


src/Process/Runner/ProcessQueue.php
```diff
@@ -0,0 +1,120 @@
...
+ * @internal
+ */
+class ProcessQueue
+{
+    private const MINIMAL_DEPTH = 1;
+
+    private const NANO_SECONDS_IN_MILLI_SECOND = 1_000_000;
+
+    /**
+     * @psalm-suppress InvalidParamDefault
```
sanmai:
```
ERROR: InvalidParamDefault - src/Process/Runner/ProcessQueue.php:57:9 - Default value type SplQueue<mixed> for argument 1 of method Infection\Process\Runner\ProcessQueue::__construct does not match the given type SplQueue<Infection\Process\MutantProcessContainer> (see https://psalm.dev/062)
        /** @var SplQueue<MutantProcessContainer> $bucket */
        private readonly SplQueue $bucket = new SplQueue(),
```
Sounds like we need to upgrade Psalm.

src/Process/Runner/ParallelProcessRunner.php
```diff
@@ -73,12 +69,13 @@ class ParallelProcessRunner implements ProcessRunner
     private bool $shouldStop = false;
 
     /**
-     * @param int $poll Delay (in milliseconds) to wait in-between two polls
+     * @param non-negative-int $poll Delay (in milliseconds) to wait in-between two polls
```
theofidry, line 72:
```suggestion
     * @param int<0,max> $poll Delay (in milliseconds) to wait in-between two polls
```
?

src/Process/Runner/ProcessQueue.php
staabm: > Sounds like we need to upgrade Psalm.

I try upgrading in https://github.com/infection/infection/pull/2565 .. it's a mess.

src/Process/Runner/ProcessQueue.php
```diff
@@ -52,10 +52,10 @@ class ProcessQueue
...
     private const NANO_SECONDS_IN_MILLI_SECOND = 1_000_000;
 
+    /**
+     * @param SplQueue<MutantProcessContainer> $bucket
+     */
     public function __construct(
-        /**
-         * @param SplQueue<MutantProcessContainer> $bucket
-         */
         private readonly SplQueue $bucket = new SplQueue(),
```
theofidry: does it really need to be part of the constructor? I kinda expected it to be strictly interneral

sanmai: Good call. I thought we could DI a mock, but in the end we did not need that.

src/Process/Runner/ParallelProcessRunner.php
sanmai: We use `positive-int` quite a bit, and `non-negative-int` is [a close relative](https://phpstan.org/writing-php-code/phpdoc-types#integer-ranges), so it feels odd to use `int<0,max>` here.

maks-rafalko (approved):

src/Process/Runner/ProcessQueue.php
```diff
@@ -0,0 +1,121 @@
...
+
+    private const NANO_SECONDS_IN_MILLI_SECOND = 1_000_000;
+
+    /**
+     * @var SplQueue<MutantProcessContainer>
+     */
+    private readonly SplQueue $bucket;
+
+    public function __construct(
+        private readonly DuoClock $clock = new DuoClock(),
```
theofidry, line 61: the downside of those default values is that they hide the lack of missing wirings. For instance here, I'd expect this clock to the instance used in `ParallelProcessRunner`, but it is not the case.

tests/phpunit/Process/Runner/ProcessQueueTest.php
```diff
@@ -0,0 +1,292 @@
...
+        $result = $queue->enqueueFrom($iterator);
+
+        $this->assertSame(0, $result);
+        $this->assertTrue($queue->isEmpty());
+    }
+
+    public function test_enqueue_from_with_valid_iterator_enqueues_item(): void
+    {
+        $clock = new TimeSpy();
+        $queue = new ProcessQueue($clock);
```
theofidry: the clock doesn't seem to be used here?

tests/phpunit/Process/Runner/ProcessQueueTest.php
```diff
@@ -0,0 +1,292 @@
...
+        $result = $queue->enqueueFrom($iterator);
+
+        // With original: (1001.0 - 1000.0) * 1_000_000 = 1_000_000
+        // With mutation: (1001.0 + 1000.0) * 1_000_000 = 2_001_000_000
+        $this->assertSame(1_000_000, $result);
+    }
+
+    public function test_enqueue_from_advances_iterator(): void
+    {
+        $clock = new TimeSpy();
```
theofidry: likewise here

tests/phpunit/Process/Runner/ProcessQueueTest.php
```diff
@@ -0,0 +1,292 @@
...
+            ->willReturn($container);
+        $iterator->expects($this->once())
+            ->method('next'); // This is the critical assertion
+
+        $queue->enqueueFrom($iterator);
+    }
+
+    public function test_enqueue_from_with_default_max_depth_of_one(): void
+    {
+        $clock = new TimeSpy();
```
theofidry: likewise here

tests/phpunit/Process/Runner/ParallelProcessRunnerTest.php
```diff
@@ -132,213 +132,46 @@ public function test_it_handles_all_kids_of_processes_with_infinite_threads(int
...
-        // Verify fillBucketOnce was called multiple times with 1
-        $this->assertGreaterThan(1, $callCount, 'fillBucketOnce should be called multiple times');
-    }
-
-    public function test_initial_fill_bucket_once_method_call_is_required(): void
-    {
-        // This test uses a partial mock to verify the initial fillBucketOnce is called
-        // before any other operations in the run() method.
+        // This test verifies that enqueueFrom is called before the do-while loop starts
+        // to pre-load the first process, buying time during setup
```
theofidry, line 138: is this test really useful? Essentially this tests that `ParallelProcessRunner` calls `::enqueueFrom()` before anything... which is that the source code is? In other words, we're not testing a calculation or outcome, we are testing that we declared calling `::enqueueFrom()` at the right time; which would have value if it was dynamically called without any control, but here we explicitly call it first.

I guess without it you could have a mutation or covered lined missing, because what justifies its [calling `::enqueueFrom()`] existence is a performance optimisation that has no impact on the final outcome. But maybe it is better to have this than a test that feels like a 1:1 match the source code?

Maybe it's just me, feel free to ignore this rambling if you disagree. I don't really mean to start on debate on that

src/Process/Runner/ProcessQueue.php
sanmai: We can remove the defaults from here, and let the DI container handle it. That said, the clock instance is simple wrapper over `time()` etc that has no state whatsoever.

tests/phpunit/Process/Runner/ParallelProcessRunnerTest.php
sanmai: It is a carry over from the previous test IIRC, and the reason is escaping mutations. Still, this test is entirely written by Claude, so if it covers all lines and leaves no escaping mutations, it is a win in my book.

tests/phpunit/Process/Runner/ProcessQueueTest.php
sanmai: Nailed all these down.

PR #2565, opened 2025-11-07
Update psalm to latest 5.x version

staabm:


staabm:
I tried at least 10 different psalm 5.x and 6.x versions. even the latest 7.0.0-beta11.. everytime we get 

> Uncaught InvalidArgumentException: Could not get class storage for infection\framework\enum\t in phar:///home/runner/work/infection/infection/.tools/psalm/src/Psalm/Internal/Provider/ClassLikeStorageProvider.php:39

it feels unusable atm.

it seems there are ~14 open issues mentioning this problem: https://github.com/vimeo/psalm/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20%22Could%20not%20get%20class%20storage%22

sanmai:
Yeah, it looks like we might have to abandon it after all. Vimeo was acquired by Bending Spoons, the company known for buying tech companies and firing everyone.

theofidry:
Well Matthew Brown (creator of Psalm) stepped down of the project before that IIRC. But it has been handed over to [members of the community](https://github.com/danog) IIRC. So I don't think the project is affected much by the acquisition.

The latest version seems to be Psalm6 no?

I don't have a strong opinion as to whether we should keep using Psalm or not. But maybe it is worth opening an issue on their repo?

staabm:
see https://github.com/infection/infection/pull/2565#issuecomment-3501126827 

there is a blocking issue, which others also reported already

PR #2566, opened 2025-11-07
test: Fix the CI

theofidry:
Seems like I accidentally broke it in #2559. Somehow I missed the autoreview failure, I thought the failure was only due to the escape mutations which were expected (it was for classes out of scope of the PR).


staabm:
seems we worked at the same time on this problem. IMO my alternative is a bit cleaner and fixes more problems: https://github.com/infection/infection/pull/2568

update: my PR has psalm issues :/

PR #2567, opened 2025-11-07
refactor(SchemaConfiguration): Leverage PHPStan types and move part of the validation to the factory

theofidry:
Follow up from https://github.com/infection/infection/pull/2559#issuecomment-3499603576.

### DIFF
--- a/devTools/phpstan.neon
+++ b/devTools/phpstan.neon
@@ -142,7 +142,7 @@ parameters:
             identifier: shipmonk.deadMethod
             path: ../tests/benchmark/*/*Bench.php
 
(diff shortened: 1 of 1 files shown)


devTools/phpstan.neon
```diff
@@ -141,6 +141,17 @@ parameters:
         -
             identifier: shipmonk.deadMethod
             path: ../tests/benchmark/*/*Bench.php
+
+        # We are testing user inout values hence by design they may not fit the specified type
```
maks-rafalko:
```suggestion
        # We are testing user input values hence by design they may not fit the specified type
```

PR #2568, opened 2025-11-07
Fix PHPStan errors

staabm:


staabm:
hmm I just see this very old psalm release does not play well with the PHPStan fixes.

[Psalm upgrade seems not possible atm](https://github.com/infection/infection/pull/2565)

PR #2570, opened 2025-11-08
perf(SafeDomXPath): Leverage `SafeDomXPath::fromFile()`

theofidry:
This PR leverages the factory introduced in #2558, as discussed in #2559 for which the benchmarks showed that there is a slight time improvement and a good improvement on the memory consumption.

I'm not too happy with how I had to tweak the tests to make it work, but I don't really want to revamp them entirely especially since I plan to remove them in #2543.


theofidry:
Looks like there's more work needed

PR #2572, opened 2025-11-09
refactor: Leverage FileSystem

theofidry:
A long while back, I extracted some filesystem related utilities to `fidry/filesystem`. I had accounted for the need in Infection at that time and I've been using them in Box & PHP-Scoper.

Since then, Infection needs evolved so I had to fit in a few more things, but nothing major.

- Add the filesystem test case utility
- A few additional for `FileSystem`
- A `FileSystem` interface which makes mocking a bit cleaner
- A `FS` utility (that I mostly used for testing) – it can be used as a more consistent alternative to `Safe\*` methods but we do not have to change things either.
- A readonly `FileSystem`
- A fake `FileSystem`
- A `SplFileInfo` builder

- [ ] Get rid of the 2 `MockSplFileInfo`


src/Framework/OperatingSystem.php
```diff
@@ -45,6 +45,11 @@ final class OperatingSystem
 {
     use CannotBeInstantiated;
 
+    public static function isMacOs(): bool
```
theofidry, line 48: Extracted in https://github.com/infection/infection/pull/2573

src/FileSystem/FakeFileSystem.php
```diff
@@ -0,0 +1,191 @@
...
+declare(strict_types=1);
+
+namespace Infection\FileSystem;
+
+use DomainException;
+use Fidry\FileSystem\FileSystem;
+use Symfony\Component\Finder\Finder;
+use Traversable;
+
+// TODO: replace with the one from the library
```
theofidry: TODO

maks-rafalko:
Let me ask one not convenient question: do we really need it? At the **current** state of the diff, I don't see it adds much, but instead we need to learn new API.

What is currently not convenient that requires introducing an abstraction above SF Filesystem? (I can miss something, of course)

theofidry:
>Let me ask one not convenient question: do we really need it?

Absolutely not. There is absolutely no _need_ whatsoever. 

>What is currently not convenient that requires introducing an abstraction above SF Filesystem? (I can miss something, of course)

`Filesystem` lacks a few things, it is extended it to add them. It's just that in a library it makes a bit more sense to expose an interface than 3 classes extending each other as the main API.

I had a few more grievances in #2543 (some I encountered in there but did not address either) but this can totally be done without.

>instead we need to learn new API.

I feel this statement a bit dramatic given how small the API we are talking about is, as it is as much learning from a separate library as our custom `FilesystemTestCase`.

That said, I agree in the current form the impact is a lot lesser. So I don't plan to propose it in its current form.

But it's not without results:

- There is a few changes/things I noticed that I would like to change within Infection. Trying to integrate it showed those points, but they don't require the lib to be directly addressed.
- I'm getting out of it a few QoL that would be nice. It would be nicer if they were in Symfony itself, but that is a separate problem (that I'm tracking in https://github.com/theofidry/filesystem/issues/29) and I have a discussed planned with one of the Symfony core team members at the SymfonyCon about it. Maybe some will be picked up, may none, we will see, I don't really have any expectations and in any case it is things I'll benefit from in at least Box and PHP-Scoper which are the primary consumers of this library.

TL:DR; I plan to continue first without it. Maybe later it will reach a point where there is a sufficient impact to justify proposing this migration, meanwhile it's fairly minor.

theofidry:
I do feel more strongly about #2488 though :P At least I _really_ don't want to touch that factory test in its current form, I'm getting PTSD at the idea of adding more parameters 🙈

sidz:
> Let me ask one not convenient question: do we really need it? At the current state of the diff, I don't see it adds much, but instead we need to learn new API.
> What is currently not convenient that requires introducing an abstraction above SF Filesystem? (I can miss something, of course)

Thought that it was just me who doesn't see any benefits of moving out from symfony/filesystem. :)

theofidry:
>Thought that it was just me who doesn't see any benefits of moving out from symfony/filesystem. :)

I mean, I wouldn't call it "moving away" when _it is_ the `symfony/filesystem` underneath with 90% of the API kept 😅

sidz:
@theofidry Yeah, I saw that's library you introduced is just a wrapper around symfony/filesystem ;)

tests/phpunit/TestFramework/Coverage/XmlReport/IndexXmlCoverageParser/generated_index.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0"?>
```
theofidry: to remove

tests/phpunit/FileSystem/FileSystemTestCase.php
```diff
@@ -35,62 +35,28 @@
...
-use Symfony\Component\Filesystem\Filesystem;
-use Symfony\Component\Filesystem\Path;
-use function sys_get_temp_dir;
+use function is_string;
 
 /**
  * @private
  */
-abstract class FileSystemTestCase extends TestCase
+abstract class FileSystemTestCase extends FidryFileSystemTestCase
```
theofidry: get rid of this boilerplate

tests/phpunit/EnvVariableManipulation/BacksUpEnvironmentVariables.php
```diff
@@ -37,6 +37,7 @@
 
 use Webmozart\Assert\Assert;
 
+// TODO: use an extension instead; or is there a backup globals?
```
theofidry, line 40: TODO (unrelated to the PR)

src/FileSystem/Finder/StaticAnalysisToolExecutableFinder.php
theofidry, line 61: TODO: extract the desired changes from the executable finder

PR #2580, opened 2025-11-15
chore: Update phpstan/phpstan-phpunit to 2.0.8

staabm:
with phpstan-phpunit 2.0.8 data-provider methods are validated against the signatures of the corresponding tests.
this made PHPStan error in a few cases which have been fixed with his PR
(e.g. data-sets not providing values according to types, or wrong number of parameters etc)

See a [related blog post](https://staabm.github.io/2025/11/15/phpstan-validates-phpunit-data-provider.html)


staabm:
the remaining error is fixed by https://github.com/infection/infection/pull/2579 - which I think deserves a separate PR, because its a real bug

theofidry:
Thank you @staabm

PR #2584, opened 2025-11-16
test(FileSystem): Update the make_tmp_dir function

theofidry:
I'm not too happy with the current implementation of `make_tmp_dir`:

- It is a plain function which is a bit annoying with the autoloading. I managed to move all the other functions that was in this `Helpers.php`, this one is the last one.
- The implementation is a bit confusing and while it works it's really tailored to an approach, which is relying on the `TEST_TOKEN` and having one main tmp directory, in `FileSystemTestCase`.
- I do not think the approach of having a "namespace" directory makes much sense:
  - It is is more likely to break things with parallelisation. We have to use `TEST_TOKEN` to actively avoid that, but that's again very specific and if one launches the tests differently it breaks.
  - We do not use consistently this in the project anyway, so that the assumption that all temporary files created by the Infection test suite are in one directory that can be safely deleted is false.

Instead the proposed solution is – I hope, a bit more straightforward:

- The utility function `make_tmp_dir` is moved to `FS::tmpDir()` (the naming is picked after other Symfony Filesystem method names, e.g. `readLink()`, `readFile()`.
- The implementation of `::tmpDir()` is closer to the native implementation of `tempnam()`.
- `FileSystemTestCase` now creates a temporary directory in the system default temporary path based on the used class name. It still has a random suffix (which uses `bin2hex(random_bytes())` rather than `random_int`) so we keep the best of both worlds: the names are still namespaced, making it easy to remove leftover files generated by Infection tests (shouldn't be needed in practice though) but remains sufficiently unique that it is unlikely to collide with the temporary directory created from another thread.
- Just for good measure I also introduced `::tmpFile()`, which is basically `tempnam()` renamed for consistency.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -26,7 +26,7 @@ jobs:
         php-version: [ '8.2', '8.3', '8.4', '8.5' ]
         dependencies: [ locked, lowest, highest ]
         coverage-driver: [ pcov, xdebug ]
-        symfony-version: [ '6.4.*', '7.*.*' ]
+        symfony-version: [ '6.4.*', '7.*.*', '8.0.*' ]
         include:
           - operating-system: windows-latest
             php-version: '8.2'
@@ -36,6 +36,10 @@ jobs:
         exclude:
           - dependencies: 'locked'
             symfony-version: '7.*.*'
+          - php-version: '8.2'
+            symfony-version: '8.0.*'
+          - php-version: '8.3'
+            symfony-version: '8.0.*'
 
     name: Tests on ${{ matrix.operating-system }} with PHP ${{ matrix.php-version }} (${{ matrix.dependencies }}; Symfony ${{ matrix.symfony-version }}), using ${{ matrix.coverage-driver }}
 
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ phpstan-baseline: vendor $(PHPSTAN)
 
 .PHONY: psalm-baseline
 psalm-baseline: vendor
-	$(PSALM) --threads=max --set-baseline=psalm-baseline.xml
+	$(PSALM) --threads=max --set-baseline=devTools/psalm-baseline.xml
 
 .PHONY: detect-collisions
 detect-collisions: vendor $(PHPSTAN)
@@ -100,7 +100,7 @@ detect-collisions: vendor $(PHPSTAN)
 
 .PHONY: psalm
 psalm: vendor $(PSALM)
-	$(PSALM) --threads=max
+	$(PSALM) --threads=max --use-baseline=devTools/psalm-baseline.xml
 
 .PHONY: rector
 rector: vendor $(RECTOR)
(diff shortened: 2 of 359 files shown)


tests/phpunit/FileSystem/FileSystemTestCase.php
```diff
@@ -35,62 +35,42 @@
...
-
     protected string $cwd = '';
 
     protected string $tmp = '';
 
-    public static function tearDownAfterClass(): void
-    {
-        // Cleans up whatever was there before. Indeed upon failure PHPUnit fails to trigger the
-        // `tearDown()` method and as a result some temporary files may still remain.
-        self::removeTmpDir();
```
maks-rafalko, line 64: could you please explain why we removed it? What will happen if PHPUnit fails (as described in the comment)?

tests/phpunit/FileSystem/FileSystemTestCase.php
```diff
@@ -35,62 +35,42 @@
...
         (new Filesystem())->remove($this->tmp);
     }
 
-    final protected static function removeTmpDir(): void
+    /**
+     * If the test case is `App\Tests\MyFilesystemServiceTestCase`, the default prefix will be "App\Tests\MyFilesystemServiceTestCase".
+     */
+    protected function getTmpDirPrefix(): string
     {
-        $testToken = getenv('TEST_TOKEN');
```
maks-rafalko, line 88: I'm pretty sure this was added to work with paratest (its support though has been dropped).

Will it work now with parallel execution?

tests/phpunit/FileSystem/FileSystemTestCase.php
theofidry: the `::tearDown()` method does get called, even if the test fail. I don't know if it was an earlier version of PHP and/or due to the introduction of `Throwable` or smth and it changed at some point. But that comment is not accurate anymore.

tests/phpunit/FileSystem/FileSystemTestCase.php
theofidry: Since every directory created is unique, there is no need to add a `TEST_TOKEN` value. It was only relevant before because for one process, we were using the same root directory which gets cleaned up as well by that process. So if you had 2 processes at the same time, one could have deleted that parent directory while the other was still going on.

As a downside the temporary directories are "less contained", as they are no longer in a `/path/to/sys-tmp-dir/Infection` directory, but they will still look like `/path/to/sys-tmp-dir/InfectionTests...ClassName8d9e0f1a` so you can always easily do `rm -r /path/to/sys-tmp-dir/Infection*`

tests/phpunit/FileSystem/FileSystemTestCase.php
staabm: making sure `tearDown()` is called is test-covered in phpunit-src since https://github.com/sebastianbergmann/phpunit/commit/cdaeb141fefd91c427d548d2ef40be7234ec69d5

you need this fix maybe for very old phpunit versions. maybe it makes sense to formulate a minimum phpunit version in `composer.json` with a conflicts with `phpunit/phpunit < x.y` constraint

theofidry: It's only for infections own tests so even if you install it as a dev dep I doubt it would make sense to care (for this case).

theofidry: Pretty cool that you fond that commit though :D

PR #2585, opened 2025-11-16
refactor(SafeDOMXPath): Introduce a more friendly API to SafeDOMXpath

theofidry:
Extracted from #2543.

From  #2598 it is apparent that PHPStan does not account correctly for `DOMNodeList` array access (`$domNodeList[0] ≠ $domNodeList->item(0)`) and only the later form shows the type issues. As a result of this, making our code stricter makes it unnecessarily ugly.

This PR introduces two key changes:

- We no longer rely on the magic getter in `SafeDOMXPath`. Instead we make `SafeDOMXpath#document` a public readonly property.
- Instead of having untyped `::query()` that may return anything, I introduced a couple of utility methods that makes it a bit easier to read (at the cost of more heavy testing of `SafeDOMXPath` – although IMO they have the merit of showing a bit more the API of this service for those unfamiliar with XPath).

This PR is a strict refactoring, no change of behaviour is intended here.

### DIFF
--- a/src/TestFramework/Coverage/XmlReport/IndexXmlCoverageParser.php
+++ b/src/TestFramework/Coverage/XmlReport/IndexXmlCoverageParser.php
@@ -127,7 +127,7 @@ class IndexXmlCoverageParser
         ];
 
         foreach ($sourceQueries as $sourceQuery) {
-            $source = $xPath->queryAttribute($sourceQuery)?->nodeValue;
+            $source = $xPath->queryList($sourceQuery)->item(0)?->nodeValue;
 
             if ($source !== null) {
                 return $source;
--- a/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
+++ b/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
@@ -197,11 +197,11 @@ final readonly class XmlConfigurationManipulator
             return true;
         }
 
-        $schema = $xPath->queryAttribute('/phpunit/@xsi:noNamespaceSchemaLocation')?->nodeValue;
+        $schema = $xPath->queryList('/phpunit/@xsi:noNamespaceSchemaLocation');
 
         $original = libxml_use_internal_errors(true);
 
-        if ($schema !== null && !$xPath->document->schemaValidate($this->buildSchemaPath($schema))) {
+        if ($schema->length > 0 && !$xPath->document->schemaValidate($this->buildSchemaPath($schema[0]->nodeValue))) {
             throw InvalidPhpUnitConfiguration::byXsdSchema(
                 $configPath,
                 $this->getXmlErrorsString(),
@@ -388,10 +388,10 @@ final readonly class XmlConfigurationManipulator
 
     private function getOrCreateNode(SafeDOMXPath $xPath, DOMDocument $dom, string $nodeName): DOMElement
     {
-        $node = $xPath->queryElement(sprintf('/phpunit/%s', $nodeName));
+        $node = $xPath->queryList(sprintf('/phpunit/%s', $nodeName));
 
-        if ($node !== null) {
-            return $node;
+        if ($node->length > 0) {
+            return $node[0];
         }
 
         return $this->createNode($dom, $nodeName);
--- a/src/TestFramework/PhpUnit/Config/XmlConfigurationVersionProvider.php
+++ b/src/TestFramework/PhpUnit/Config/XmlConfigurationVersionProvider.php
@@ -108,8 +108,12 @@ final class XmlConfigurationVersionProvider
             return null;
         }
 
-        $schema = $xPath->queryAttribute('/phpunit/@xsi:noNamespaceSchemaLocation');
+        $schema = $xPath->queryList('/phpunit/@xsi:noNamespaceSchemaLocation');
 
-        return $schema?->nodeValue;
+        if ($schema->length === 0) {
+            return null;
+        }
+
+        return $schema[0]->nodeValue;
     }
 }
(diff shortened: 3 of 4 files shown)


src/TestFramework/Coverage/XmlReport/IndexXmlCoverageParser.php
```diff
@@ -111,14 +111,19 @@ private static function assertHasExecutedLines(SafeDOMXPath $xPath, bool $isForG
...
+            $source = $xPath->queryAttribute('//p:project/@source')?->nodeValue;
+
+            if ($source !== null) {
+                return $source;
+            }
         }
 
-        // PHPUnit < 6
-        return $xPath->query('//p:project/@name')[0]->nodeValue;
+        throw new RuntimeException('Could not find the source attribute for the project');
```
theofidry: I think it is a fix as previously it meant we ended up with a `TypeError` and now we have a `RuntimeException`. We may want to have a more specific and user friendly exception.

This should be extracted and covered by tests.

src/TestFramework/Coverage/XmlReport/XmlCoverageParser.php
```diff
@@ -74,24 +74,25 @@ private static function createTestLocationsGenerator(SafeDOMXPath $xPath): itera
 
     private static function retrieveTestLocations(SafeDOMXPath $xPath): TestLocations
     {
-        $linesNode = $xPath->query('/p:phpunit/p:file/p:totals/p:lines')[0];
-
-        $percentage = $linesNode->getAttribute('percent');
+        // @phpstan-ignore method.nonObject
```
theofidry: here there is no guarantee that the percentage is found: should we leave it as is (which can cause a `TypeError`) or check for it?

Note that this is directly from reading the coverage report which is a hotpath: we can justify risking an obscure error in favour of performance.

src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
```diff
@@ -149,11 +149,13 @@ private function addRandomTestsOrderAttributesIfNotSet(string $version, SafeDOMX
...
     private function addAttributeIfNotSet(string $attribute, string $value, SafeDOMXPath $xPath): bool
     {
-        $nodeList = $xPath->query(sprintf('/phpunit/@%s', $attribute));
+        $count = $xPath->queryCount(sprintf('/phpunit/@%s', $attribute));
 
-        if ($nodeList->length === 0) {
-            $node = $xPath->query('/phpunit')[0];
-            $node->setAttribute($attribute, $value);
+        if ($count === 0) {
+            // @phpstan-ignore method.nonObject
```
theofidry: Likewise here

src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
```diff
@@ -168,13 +168,15 @@ private function buildPath(string $mutationHash): string
...
+        $bootstrap = $xPath->queryAttribute('/phpunit/@bootstrap');
 
-        if ($bootstrap->length > 0) {
-            $bootstrap[0]->nodeValue = $customAutoloadFilePath;
+        if ($bootstrap !== null) {
+            $bootstrap->nodeValue = $customAutoloadFilePath;
         } else {
-            $node = $xPath->query('/phpunit')[0];
-            $node->setAttribute('bootstrap', $customAutoloadFilePath);
+            // @phpstan-ignore method.nonObject
```
theofidry: Likewise here

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
```diff
@@ -340,16 +340,18 @@ private function removeAttribute(SafeDOMXPath $xPath, string $name): void
...
             '/phpunit/@%s',
             $name,
         ));
 
         if ($nodeList->length > 0) {
             $nodeList[0]->nodeValue = $value;
         } else {
-            $node = $xPath->query('/phpunit')[0];
-            $node->setAttribute($name, $value);
+            // @phpstan-ignore method.nonObject
```
theofidry: Likewise here

src/TestFramework/Coverage/XmlReport/IndexXmlCoverageParser.php
```diff
@@ -111,14 +111,19 @@ private static function assertHasExecutedLines(SafeDOMXPath $xPath, bool $isForG
...
-        $sourceNodes = $xPath->query('//p:project/@source');
+        $sourceQueries = [
+            '//p:project/@source',  // PHPUnit >= 6
+            '//p:project/@name',    // PHPUnit < 6
+        ];
 
-        if ($sourceNodes->length > 0) {
-            return $sourceNodes[0]->nodeValue;
+        foreach ($sourceQueries as $sourceQuery) {
+            $source = $xPath->queryAttribute('//p:project/@source')?->nodeValue;
```
sanmai:
```suggestion
            $source = $xPath->queryAttribute($sourceQuery)?->nodeValue;
```
Sounds like this was the intent?

theofidry: indeed; I remember fixing this but looks like I got distracted and forgot to push it

maks-rafalko (approved):

src/TestFramework/Coverage/XmlReport/IndexXmlCoverageParser.php
theofidry: Done in https://github.com/infection/infection/pull/2597

PR #2587, opened 2025-11-17
check for OPTION_IGNORE_MSI_WITH_NO_MUTATIONS first

### LINKED ISSUE #2589: Config value of ignoreMsiWithNoMutations is ignored
| Question    | Answer
| Infection version | dev
| Test Framework version | PHPUnit/PhpSpec 11.5.44
| PHP version | 8.4
| Platform    | Linux
| Github Repo | -

Let's use our existing E2E test:

```
cd tests/e2e/Ignore_MSI_Zero_Mutations/
```

Running Infection there without `--ignore-msi-with-no-mutations` leads to a failure.
```
../../../bin/infection --filter=notExistentFile.php --min-msi=100
```

But with the switch it no longer fails, as expected:

```
../../../bin/infection --ignore-msi-with-no-mutations --filter=notExistentFile.php --min-msi=100
```

Then if we add the config option:

```diff
--- a/tests/e2e/Ignore_MSI_Zero_Mutations/infection.json
+++ b/tests/e2e/Ignore_MSI_Zero_Mutations/infection.json
@@ -9,7 +9,8 @@
         "summary": "infection.log"
     },
     "mutators": {
         "PublicVisibility": true
     },
+    "ignoreMsiWithNoMutations": true,
     "tmpDir": "."
 }
```

The command without the CLI switch still fails:

```
../../../bin/infection --filter=notExistentFile.php --min-msi=100
```

```
0 mutations were generated:
       0 mutants were killed by Test Framework

Metrics:
         Mutation Code Coverage: 0%
         Covered Code MSI: 0%

 [ERROR] The minimum required MSI percentage should be 100%, but actual is 0%. Improve your tests!                      
```

wickedOne:
this pr fixes an issue where the config value of `ignoreMsiWithNoMutations` is ignored / overwritten due to the `OPTION_IGNORE_MSI_WITH_NO_MUTATIONS` option being casted to `false` when omitted.

Fixes #2589

### DIFF
--- a/src/Command/RunCommand.php
+++ b/src/Command/RunCommand.php
@@ -86,6 +86,11 @@ final class RunCommand extends BaseCommand
     /** @var string */
     public const OPTION_IGNORE_MSI_WITH_NO_MUTATIONS = 'ignore-msi-with-no-mutations';
 
+    /**
+     * Sentinel value for VALUE_OPTIONAL options to distinguish "not provided" from "provided without value"
+     */
+    public const OPTION_VALUE_NOT_PROVIDED = false;
+
     /** @var string */
     private const OPTION_TEST_FRAMEWORK = 'test-framework';
 
@@ -395,7 +400,7 @@ final class RunCommand extends BaseCommand
                 null,
                 InputOption::VALUE_OPTIONAL,
                 'Ignore MSI violations with zero mutations',
-                false,
+                self::OPTION_VALUE_NOT_PROVIDED,
             )
             ->addOption(
                 self::OPTION_DEBUG,
--- a/src/Command/RunCommandHelper.php
+++ b/src/Command/RunCommandHelper.php
@@ -164,10 +164,10 @@ final readonly class RunCommandHelper
     {
         $ignoreMsiWithNoMutations = $this->input->getOption(RunCommand::OPTION_IGNORE_MSI_WITH_NO_MUTATIONS);
 
-        // `false` means the option was not provided at all -> return null to preserve config value
+        // OPTION_VALUE_NOT_PROVIDED means the option was not provided at all -> return null to preserve config value
         // `null` or any other value means the option was provided -> return true to enable it
         return match ($ignoreMsiWithNoMutations) {
-            false => null,
+            RunCommand::OPTION_VALUE_NOT_PROVIDED => null,
             default => true,
         };
     }
--- a/tests/phpunit/Command/RunCommandHelperTest.php
+++ b/tests/phpunit/Command/RunCommandHelperTest.php
@@ -156,7 +156,7 @@ final class RunCommandHelperTest extends TestCase
 
     public static function providesIgnoreMsiWithNoMutations(): iterable
     {
-        yield 'not provided' => [null, false];
+        yield 'not provided' => [null, RunCommand::OPTION_VALUE_NOT_PROVIDED];
 
         yield 'provided without value' => [true, null];
 


src/Command/RunCommand.php
```diff
@@ -529,7 +530,8 @@ private function createContainer(IO $io, LoggerInterface $logger): Container
             // To keep in sync with Container::DEFAULT_SKIP_INITIAL_TESTS
             (bool) $input->getOption(self::OPTION_SKIP_INITIAL_TESTS),
             // To keep in sync with Container::DEFAULT_IGNORE_MSI_WITH_NO_MUTATIONS
-            (bool) $input->getOption(self::OPTION_IGNORE_MSI_WITH_NO_MUTATIONS),
+            // VALUE_OPTIONAL with default false: false = not provided, null/other = provided
+            $input->getOption(self::OPTION_IGNORE_MSI_WITH_NO_MUTATIONS) !== false ? true : null,
```
sanmai: As documented: https://symfony.com/doc/current/console/input.html#options-with-optional-arguments

sanmai:
@wickedOne Thanks for the PR and my sincere apologies for the PR hijack :pray:

src/Command/RunCommandHelper.php
```diff
@@ -159,4 +159,16 @@ public function getNumberOfShownMutations(): ?int
...
+    public function getIgnoreMsiWithNoMutations(): ?bool
+    {
+        $ignoreMsiWithNoMutations = $this->input->getOption(RunCommand::OPTION_IGNORE_MSI_WITH_NO_MUTATIONS);
+
+        // OPTION_VALUE_NOT_PROVIDED means the option was not provided at all -> return null to preserve config value
+        // `null` or any other value means the option was provided -> return true to enable it
+        return match ($ignoreMsiWithNoMutations) {
+            RunCommand::OPTION_VALUE_NOT_PROVIDED => null,
+            default => true,
+        };
```
sanmai, line 172: https://symfony.com/doc/current/console/input.html

```php
$this
    // ...
    ->addOption(
        'yell',
        null,
        InputOption::VALUE_OPTIONAL,
        'Should I yell while greeting?',
        false // this is the new default value, instead of null
    )
;
```

```php
$optionValue = $input->getOption('yell');
if (false === $optionValue) {
    // in this case, the option was not passed when running the command
    $yell = false;
    $yellLouder = false;
} elseif (null === $optionValue) {
    // in this case, the option was passed when running the command
    // but no value was given to it
    $yell = true;
    $yellLouder = false;
} else {
    // in this case, the option was passed when running the command and
    // some specific value was given to it
    $yell = true;
    if ('louder' === $optionValue) {
        $yellLouder = true;
    } else {
        $yellLouder = false;
    }
}
```

sanmai:
I tested this fixes #2589 following the testing steps, should be RFR.

wickedOne:
> @wickedOne Thanks for the PR and my sincere apologies for the PR hijack 🙏

haha, i was intending to write some tests this evening and learned you already took care of that, so thank _you_  😄

src/Command/RunCommand.php
wickedOne: are you sure this is correct?

when it has an optional value you could run the command with `--ignore-msi-with-no-mutations=false` then it sets to `null`. if you have `"ignoreMsiWithNoMutations": true` in your config, wouldn't you get an undesired result?

sanmai: Well, you are not supposed to run `--ignore-msi-with-no-mutations` with values. At least as far as the documentation is concerned.

sanmai: What do you think is the better approach here? We refactored this into a helper method.

maks-rafalko (approved):
seems like works as expected (tried in our e2e tests)

it would be nice to have e2e test with config value in addition to the existing e2e test with cli option though

PR #2588, opened 2025-11-17
refactor: Introduce a ReportLocator interface for the coverage report locators

theofidry:
Extracted from #2543.

As of now, the JUnit & Index locator are 95% the same code and same tests, yet different classes. More annoying as well is that those two classes are needed elsewhere hence we directly mock them.

This PR introduces a `ReportLocator` interface. This has several benefits:

- This makes mocking simpler (no more not final `@final` class)
- It introduces a `FakeLocator` implementation that is otherwise a bit more annoying to mimic with mocks.
- It introduces a `FixedLocator` which is useful when we know exactly where the report is, i.e. there is no lookup to do. Quite useful for some tests but there is nothing test specific in the implementation.
- It introduces a `BaseReportLocator` base implementation*.

*: I had a lot of back and forth on how to proceed there. My conclusion is that we have the same behaviour for both locators. What differs is the exceptions because we want to fine-tune the exception message. So this base implementation offers the hooks to do so.

To have a finer control on the FileSystem, new methods have been introduced. I refrained from letting it creep up to other places for now however. Note that the implementation is quite rudimentary, but I believe `fidry/filesystem` is more adequate for a more proper implementation. We don't really have a need for this yet so I just kept the naïve implementation for now.

I reverted the original chance to move the caching to a dedicated locator (`MemoizedLocator`) as there is currently issues in the container to achieve this. See https://github.com/sanmai/di-container/issues/25


src/Container.php
```diff
@@ -264,8 +262,11 @@ public static function create(): self
                 $container->getIndexXmlCoverageParser(),
                 $container->getXmlCoverageParser(),
             ),
-            IndexXmlCoverageLocator::class => static fn (self $container): IndexXmlCoverageLocator => new IndexXmlCoverageLocator(
-                $container->getConfiguration()->coveragePath,
+            IndexReportLocator::class => static fn (self $container): ReportLocator => new MemoizedLocator(
```
theofidry: @sanmai from what I see we cannot have that with the current container.

I notice we have some code where we have `ServiceFoo` and then `MemoizedServiceFoo`, but if feels quite boilertemplaty. Is there any way we could get rid of that?

Looks like it's a sanity check that is done, and tbh I'd argue it makes sense most of the time, but maybe it could either be disabled for a specific service or extended to allow this case for interfaces (i.e. whilst the output is not the same instance as the ID, it is still of the same parent type).

tests/phpunit/TestFramework/Coverage/Locator/BaseReportLocator/BaseReportLocatorTest.php
```diff
@@ -0,0 +1,190 @@
...
+        $this->filesystem->touch('file1.demo');
+        $this->filesystem->touch('file2.demo');
+
+        $expectedReportsPathnames = [
+            Path::canonicalize($this->tmp . '/file1.demo'),
+            Path::canonicalize($this->tmp . '/file2.demo'),
+        ];
+
+        try {
+            $this->locator->locate();
```
theofidry, line 195: missing `$this->fail()`

src/Container.php
sanmai: Yeah, if you have two services implement the same interface and two different implementations, you will need to assemble one of the injected services yourself. I.e. in this case you will need to assemble `MemoizedLocator` but I also feel like I'm missing something from your argument. What did you want to do?

theofidry: Well in this case you cannot have `MemoizedLocator` as an ID as there would be two services with that ID...

tests/phpunit/TestFramework/Coverage/Locator/FixedLocatorTest.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Infection\DevTools\TestFramework\Coverage\Locator;
+
+use Infection\TestFramework\Coverage\Locator\FixedLocator;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\TestCase;
+
+#[CoversClass(FixedLocator::class)]
+final class FixedLocatorTest extends TestCase
```
theofidry, line 43: won't location...

src/TestFramework/Coverage/Locator/FixedLocator.php
```diff
@@ -0,0 +1,55 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+declare(strict_types=1);
+
+namespace Infection\TestFramework\Coverage\Locator;
+
+final readonly class FixedLocator implements ReportLocator
```
theofidry, line 41: Missing `@internal`; double check other classes

Need a comment: this is for test purposes only.

src/Container.php
sanmai: And you have `IndexReportLocator` and `JUnitReportLocator` so IDs are different. What service requires exactly  `MemoizedLocator`? We have to change it to require one of those.

theofidry: I'm lost 😅 Maybe I didn't correctly understand your previous suggestion. Could you explain differently what needs to be done here?

theofidry: Maybe to expend, I don't see how to manually configure this. At the end of the day I want a `MemoizedLocator` for both `IndexReportLocator` and `JUnitReportLocator`.

I think it would be trivial to do with either the absence of that check of in a traditional container, but I don't really see how to make it work here without further changes.

sanmai: What stops you from manually wiring these dependencies up into where they should go? Like, the old fashioned way before the "magic" I added. It should still work.

theofidry: I don't understand what you mean; as in inlining their definition and not registering as services in the container? They are re-used so we need shared services here

tests/phpunit/TestFramework/Coverage/XmlReport/IndexXmlCoverageLocatorTest.php
```diff
@@ -44,23 +45,28 @@
...
     {
         parent::setUp();
 
-        $this->locator = IndexXmlCoverageLocator::create($this->tmp);
+        $this->filesystem = new FileSystem();
+
+        $this->locator = IndexXmlCoverageLocator::create(
+            $this->filesystem,
+            $this->tmp,
+            $this->tmp . '/coverage-xml/index.xml',
```
theofidry: TODO: assert somewhere that this path needs to be absolute

tests/phpunit/TestFramework/Coverage/Locator/BaseReportLocator/BaseReportLocatorTest.php
```diff
@@ -0,0 +1,180 @@
...
+                    'No report found in "%s".',
+                    $this->tmp,
+                ),
+            ),
+        );
+
+        $this->locator->locate();
+    }
+
+    public function test_it_cannot_find_the_report_not_suitable_file_was_found(): void
```
theofidry:
```suggestion
    public function test_it_cannot_find_the_report_no_suitable_file_was_found(): void
```

PR #2590, opened 2025-11-18
feat: Improve the file locators error reporting

theofidry:
Extracted from #2588.

This PR improves the error reporting when failing to locate a report (JUnit for the JUnit locator and index.xml for the other).

The changes are as follows:

- Besides checking whether `$coveragePath` exists, we also check that it is a readable path. Fundamentally there is no change, if the path is a file or a non-readable directory we already end up with an exception. With this change alone, what would change is which line throws the exception, but the exception class & message would be identical.
- Rather than mentioning an interpretation of the regex used in the message, we state the pattern. While I agree it is less readable (reading `*.junit.xml` is a lot more readable than `/^(.+\.)?junit\.xml$/i`), the current code has the following flaws:
  -  The current interpretation of the regex is incorrect. For example for JUnit, it maches a **case insensitive** `*junit.xml` (also note it is `*junit`, not `*.junit`).
  - It is harder to keep synchronized with the regex.
  - Since we use a regex, we might explicitly let the user understand what is being used rather than a potentially incorrect interpretation of the regex. _If_ you feel strongly about that point, I'm happy to adapt it though._
- New exception classes with new messages: it makes it easier to understand what the problem is.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ phpstan-baseline: vendor $(PHPSTAN)
 
 .PHONY: psalm-baseline
 psalm-baseline: vendor
-	$(PSALM) --threads=max --set-baseline=psalm-baseline.xml
+	$(PSALM) --threads=max --set-baseline=devTools/psalm-baseline.xml
 
 .PHONY: detect-collisions
 detect-collisions: vendor $(PHPSTAN)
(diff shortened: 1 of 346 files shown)


theofidry:
I'd like to move forward here so...

PR #2591, opened 2025-11-21
test: Simplify the expected DX in infection for test cases

theofidry:
As of now we check if a given source code class name has a corresponding test, and if not `CodeProvider` needs to be updated.

Originally the convention was that for `Infection\Console\Application`, we expect the test `Infection\Console\ApplicationTest`.

However, in the recent months, I've introduced another pattern where the test would be `Infection\Console\Application\ApplicationTest` (it is a case where the test requires some fixtures and it is a lot clearer to have them next to the test case using them). Whenever this happens, one need to update `CodeProvider::CONCRETE_CLASSES_WITH_TESTS_IN_DIFFERENT_LOCATION`. It is annoying, and this PR aims at getting rid of this.

In other words, we now allow:

```
src/
  Console/
    Application.php
tests/
  Console/
    ApplicationTest.php
    // or:
    Application/
      ApplicationTest.php
      ...Fixtures or related files here
```
This PR improves the `ClassName` utility introduced in #2592 to deal with this scenario.


maks-rafalko:
hm.. does `--map-source-class-to-test` https://infection.github.io/guide/command-line-options.html#map-source-class-to-test still working?

I think yes, but just in case so you are aware of this functionality, that is based on file names.

theofidry:
>I think yes, but just in case so you are aware of this functionality, that is based on file names.

I don't think it would for this scenario... We could adjust the implementation I suppose, not sure. It's a convention I do follow a lot, but I don't expect it to be widespread either.

If you think it's worth it though I can open an issue to support this.

PR #2592, opened 2025-11-21
refactor(ClassName): Consolidate utilities into a ClassName utility

theofidry:
While working on #2591 I noticed several things:

- We have a need to get a class short name in multiple places. I already noticed it in `FilterBuilder` in the past but at that time it was a bit overkill or I was a bit undecided about introducing a utility class.
- We have `SourceTestClassName` which is internal but now is part of the source code as it may end up being used (indirectly, via the `BaseMutatorTest`) for custom mutators.
- in #2591 I need to extend the beahviour of `SourceTestClassName`.

This PR proposes to consolidate all of thus under a `ClassName` utility. It:

- Introduces `::getShortName()` and leverages it.
- Renames `SourceTestClassName` to `ClassName`.
- Renames `::getSourceClassName()` to `::getCanonicalSourceClassName()` and likewise for test. It may confuse briefly what "canonical" refers to in this context, but the fact is that this is not a super all purpose generic utility: it is quite specific to infection and our naming conventions. Hence, canonical here refers to our "normally expected value following our conventions". I'll also update the documentation in an upcoming PR.

### DIFF
--- a/tests/phpunit/Framework/ClassNameTest.php
+++ b/tests/phpunit/Framework/ClassNameTest.php
@@ -33,20 +33,58 @@
 
 declare(strict_types=1);
 
-namespace Infection\Tests\AutoReview;
+namespace Infection\Tests\Framework;
 
-use Infection\Testing\SourceTestClassNameScheme;
+use Closure;
+use Infection\Framework\ClassName;
 use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
 
-#[CoversClass(SourceTestClassNameScheme::class)]
-final class SourceTestClassNameSchemeTest extends TestCase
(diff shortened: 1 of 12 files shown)


tests/phpunit/Framework/ClassNameTest.php
```diff
@@ -33,44 +33,82 @@
...
+            'ClassName',
+        ];
+
+        yield 'UTF-8' => [
+            'Webmozarts\ClassName\Ütf8',
+            'Ütf8',
+        ];
+
+        yield 'emoji' => [
+            'Webmozarts\ClassName\Special😋Class',
```
theofidry, line 73: technically picked it from the libraries of my workplace which are under a proprietary license. It's fairly boring code, so no need for license approval here. Left the name as a nod.

theofidry:
Merging this as this is fairly boring stuff.

PR #2593, opened 2025-11-21
build: Introduce a var directory

### LINKED ISSUE #2498: Re-organise the project artefacts
This is a follow up of #2496.

In there we configure various things under `build`.

I admit I'm not too fan of it, we tend to mix distribution artefacts (`infection.phar`) with cache & whatever. Additionally, we have:

- `.tools`, which is not tracked by git either
- `.composer` which is another cache

In other words, with have a mixed bag under `build` but we also have extra other directories.

Now let's be clear: this is nitpicks or minor things for newcomers. But what I propose is:

- `dist` for distribution files (the PHAR)
- `var` under which we can have
  - the various caches
  - the downloaded tools

This way we cut the number of non-tracked directories by whilst following a bit better Linux conventions.

That being said, I agree it's not a big deal; If you don't want me to update this feel free to close the issue

theofidry:
As discussed in #2498, moves the artifacts from `build` to `var`.

Technically there is also the download of the dev tools, but I think it's more work and a contentious point. I did change that (it was non-existent in the first place) in the PhpSpec adapter and I found better to be able to give the version instead of the Makefile (which meant that changing the Makefile does not invalidate all dev tools) and being able to auto-update, but those are more changes I'm not really willing to dig into at this point.

As a bonus, I've added the GitHub Actions cache configuration for all of those tools, hoping this helps a bit the CI.

Closes #2498.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 /.tools/
 /build/
 /devTools/Docker*.json
+/dist/
 /infection-cache
 /infection.log
 /per-mutator.md


theofidry:
To add the rector cache dir too...

.php-cs-fixer.php
```diff
@@ -192,5 +193,5 @@
         'blank_line_after_opening_tag' => false,
     ])
     ->setFinder($finder)
-    ->setCacheFile(__DIR__ . '/build/.php_cs.cache')
+    ->setCacheFile(__DIR__ . '/var/cache/php-cs-fixer')
```
sanmai, line 196:
```suggestion
    ->setCacheFile(__DIR__ . '/var/cache/php-cs-fixer.json')
```
nit; since it is a json

.github/workflows/release.yaml
```diff
@@ -48,13 +48,13 @@ jobs:
...
-              build/infection.phar
+              --output var/infection.phar.asc \
+              var/infection.phar
 
       - name: Upload PHAR to release
         uses: svenstaro/upload-release-action@6b7fa9f267e90b50a19fef07b3596790bb941741 # latest
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
-          file: build/infection.phar*
+          file: var/infection.phar*
```
sanmai: I think the original proposal was to use `dist`

.github/workflows/autoreview.yaml
```diff
@@ -39,6 +39,34 @@ jobs:
           # This is necessary when installing a tool with a specific version
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Configure PHP-CS-Fixer GitHub Actions cache
+        uses: actions/cache@9255dc7a253b0ccc959486e2bca901246202afeb # v5.0.1
+        with:
+          path: ./var/cache/php-cs-fixer
+          key: "php-cs-fixer-${{ github.run_id }}"
```
sanmai:
```suggestion
          key: "${{ matrix.php-version }}-php-cs-fixer-${{ github.run_id }}"
```

.gitignore
```diff
@@ -1,6 +1,7 @@
 !/.tools/.gitkeep
 !/build/.gitkeep
```
sanmai: I supposed we want to clean these up, too, while keeping `/build` in the ignore list as a whole.

theofidry: I'm always hesitant about this sort of things. It's nice to have it clean and tidy, but if someone with an older checkout which had those artifact comes, we don't want to have `build` suddenly committed.

theofidry: Ah you probably meant the `gitkeep`, good point

.github/workflows/release.yaml
theofidry: We can go with `dist`, I traditionally used it. I think `var` is a bit more correct because it's any misc things whereas `dist` is usually meant for distribution files, which cache certainly is not. And sure we have the PHAR which should be part of `dist`, but since it's a temporary artefact anyway I also don't feel it being problematic to have it in `var` instead.

So if you feel strongly about it, I'm happy to compromise and use `dist` instead, but I otherwise think `var` is more correct here

.php-cs-fixer.php
theofidry: I thought it was serialized code at some point, guess it changed. I don't mind changing it, but ultimately it makes no difference since PHP-CS-Fixer will do its thing regardless of what extension you specified.

.github/workflows/release.yaml
sanmai: My concern is that nuking `var` is more consequential than nuking just `dist` with `infection.phar.*` in it. So it is not as much what is correct, but rather what requires less decision making in the moment.

theofidry: if you fear the possibility of nuking the PHAR (I don't particularly, I rarely nuke the entirety of `var` and making over the PHAR is not too much of a problem), I can still preserve `dist` for it

PR #2596, opened 2025-11-24
refactor(rector): Enable more rector rules

theofidry:
This PR is not to be squashed.

I wanted to introduce more rector rules but introducing them in one go is too much, and doing a PR for each is too much...

So instead this PR introduces various rules & presets, enabling one rule at a time per commit.

### DIFF
--- a/src/Mutator/Removal/MatchArmRemoval.php
+++ b/src/Mutator/Removal/MatchArmRemoval.php
@@ -125,6 +125,7 @@ final class MatchArmRemoval implements Mutator
                 foreach (array_keys($armConds) as $j) {
                     $conds = $armConds;
 
+                    /** @psalm-suppress InvalidArrayOffset */
                     unset($conds[$j]);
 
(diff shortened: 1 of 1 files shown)


src/Mutator/Boolean/LogicalOrNegation.php
```diff
@@ -79,7 +79,7 @@ public function mutate(Node $node): iterable
         // Clone the node to remove the origNode from the wrapped expression
         // see bug https://github.com/nikic/PHP-Parser/issues/1119
         $wrappedNode = clone $node;
-        $wrappedAttrs = $wrappedNode->getAttributes();
+        $wrappedNode->getAttributes();
```
darthf1, line 82: Maybe this line should be removed altogether?

```suggestion
```

theofidry: It could yes, but I would prefer to limit retouching too much stuff and keep it as much as possible related to the rector rules applied.

theofidry:
It's a bit annoying to have to rebase it, so I'll go ahead and merge it... We can always revert an individual rule disliked afterwards.

PR #2601, opened 2025-11-25
fix(SafeDOMXPath): Harden the code with the new SafeDOMXPath API

theofidry:
In #2585 a new API was introduced. I had to revert some changes there because technically this result in a change of behaviour hence is a fix not a refactor.


theofidry:
Merging this as this was originally part of https://github.com/infection/infection/pull/2585.

PR #2603, opened 2025-11-26
test(report): Consolidate the Index & JUnit report locator tests

theofidry:
This PR is a preparatory step for #2588.

It adds some tests and rename other bits so that the names align with the tests in #2588. The goal is to reduce the noise there and make it easier to identify how the tests have been moved around. It is otherwise too hard to follow if there is a change of behaviour or where did what test go.


theofidry:
This was a PITA to get green but the code itself is quite boring so I'll go ahead and merge this.

PR #2604, opened 2025-11-28
Support `--exclude-source-from-xml-coverage` fast path for PHPUnit 12.5+

staabm:
requires 
- https://github.com/sebastianbergmann/phpunit/issues/6422
- https://github.com/infection/infection/pull/2607

---

when PHPUnit needs to generate the `<source>` element in coverage-xml reports it spents a considerable amount of time generating the this information. infection does not make use of it though.

therefore we utilize `--exclude-source-from-xml-coverage` which will be available as of PHPUnit 12.5+ to prevent this unnecessary work and speedup the process. on phpstan-src we can see a ~15% speedup of the coverage-xml generation process (which takes up to 1 minute in CI)

### DIFF
--- a/src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
+++ b/src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
@@ -187,7 +187,7 @@ class PhpUnitAdapter extends AbstractTestFrameworkAdapter implements MemoryUsage
     }
 
     /**
-     * As of PHPUnit 12.5, the `--exclude-source-from-xml-coverage` is available which removes the `source` element which from the XML report which contained the list of tokens of the source code file.
+     * As of PHPUnit 12.5, the `--exclude-source-from-xml-coverage` is available which removes the `source` element from the XML report which contained the list of tokens of the source code file.
      */
     public static function supportsExcludingSourceFromCoverage(string $version): bool
     {
--- a/tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapterTest.php
+++ b/tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapterTest.php
@@ -378,20 +378,20 @@ final class PhpUnitAdapterTest extends TestCase
     }
 
     #[DataProvider('coverageWithoutSourceProvider')]
-    public function test_supports_coverage_without_source(bool $expected, string $version): void
+    public function test_supports_coverage_without_source(string $version, bool $expected): void
     {
         $this->assertSame($expected, PhpUnitAdapter::supportsExcludingSourceFromCoverage($version));
     }
 
     public static function coverageWithoutSourceProvider(): iterable
     {
-        yield [false, '11.5.599'];
+        yield ['11.5.599', false];
 
-        yield [false, '12.0'];
+        yield ['12.0', false];
 
-        yield [true, '12.5'];
+        yield ['12.5', true];
 
-        yield [true, '13.0'];
+        yield ['13.0', true];
     }
 
     private function getPHPUnitAdapter(string $version = '9.0'): PhpUnitAdapter


tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapterTest.php
```diff
@@ -210,6 +199,60 @@ public function test_it_provides_initial_test_run_command_line_when_coverage_rep
...
+                '--exclude-source-from-xml-coverage',
+                '--coverage-xml=/tmp/coverage-xml',
+                '--log-junit=/tmp/infection/junit.xml',
+            ],
+            $initialTestRunCommandLine,
+        );
+    }
+
     #[Group('integration')]
     public function test_it_provides_initial_test_run_command_line_when_coverage_report_is_requested_and_pcov_is_in_use(): void
```
staabm, line 257: initially I tried to use a data-provider on this test, but because of the exzessiv use of mocks we nearly need to change every line. therefore a data-provider would need a lot of parameters and this would turn this test into a unreadable mess.

therefore I decided to add a separate test to not overcomplicate things.

tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapterTest.php
```diff
@@ -66,18 +66,7 @@ protected function setUp(): void
         $this->cliArgumentsBuilder = $this->createMock(CommandLineArgumentsAndOptionsBuilder::class);
         $this->commandLineBuilder = $this->createMock(CommandLineBuilder::class);
 
-        $this->adapter = new PhpUnitAdapter(
```
staabm: this change will go away after https://github.com/infection/infection/pull/2607 merged and this PR got rebased

staabm:
I think we should wait until PHPUnit 12.5.0 is released before merging. Feedback is welcome before that though :)

~~looking at CI I cannot see whether we have test running for newer PHPUnit versions like 12.x .. all jobs I had a look at are running 11.x ..?~~ I just found `tests/e2e/PHPUnit12` which seems to be good enough

src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
```diff
@@ -177,6 +186,11 @@ public function getInitialTestsFailRecommendations(string $commandLine): string
         return $recommendations;
     }
 
+    public static function supportsCoverageWithoutSourceFastPath(string $version): bool
```
theofidry:
```suggestion
    /**
     * As of PHPUnit 12.5, the `--exclude-source-from-xml-coverage` is available which removes the `source` element which from the XML report which contained the list of tokens of the source code file.
     */
    public static function supportsExcludingSourceFromCoverage(string $version): bool
```
?

tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapterTest.php
```diff
@@ -323,6 +377,23 @@ public static function executionOrderProvider(): iterable
         yield [true, '13.0'];
     }
 
+    #[DataProvider('coverageWithoutSourceProvider')]
+    public function test_supports_coverage_without_source(bool $expected, string $version): void
```
theofidry: for the test itself, I think the reverse order makes more sense. For the assertions themselves I think it's more muscle memory than anything at this point, but for the test input it messes me up too hard 😅

theofidry (approved):
Looks good!

theofidry:
Let's wait for PHPUnit 12.5.0 then!

tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapterTest.php
staabm: reverse order of what? of the parameters?

I tend to use the 1st parameter of a test as the expected values (like PHPUnit assert* methods do)

theofidry: >I tend to use the 1st parameter of a test as the expected values (like PHPUnit assert* methods do)

Yes.

I get where this is coming from, but honestly even the order of the PHPUnit assert* method is natural for everyone. For a test however it's very common to have (with a data provider):

- setup
- act
- assert

So having the expect before the input feels a bit weird.

staabm:
since 12.5.0 was released today, this should be good to go

theofidry:
Thank you @staabm!

I'm kinda curious how drastic the size is gonna be, looking forward to check it out in the Infection benchmark-source.

staabm:
another data-point: while coverage-xml report generation was already super fast (~100ms) when running on the ParaTest codebase, adding `--exclude-source-from-xml-coverage` makes it ~50% faster

see https://github.com/paratestphp/paratest/pull/1056

theofidry:
Before:

```
make generate_coverage  61.94s user 5.23s system 91% cpu 1:13.48 total
du -sh dist/coverage/*
 12M	dist/coverage/junit.xml
 56M	dist/coverage/xml
```

After:

```
make generate_coverage  53.21s user 4.10s system 95% cpu 1:00.02 total
du -sh dist/coverage/*
 12M	dist/coverage/junit.xml
 28M	dist/coverage/xml
```

So it was 14% faster and 50% lighter.

PR #2606, opened 2025-11-28
Enable `failOnDeprecation="true"` in PHPUnit configuration

staabm:
to make sure we see deprecations early.

without a failling test these errors are easily overlooked.

I am adding `ignoreIndirectDeprecations="true"` at the same time, so our CI is does not fail for PHP version related runtime deprecation warnings from within `vendor/` folder

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -31,7 +31,7 @@
         </testsuite>
     </testsuites>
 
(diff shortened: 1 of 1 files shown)


sidz (approved):

theofidry:
Thank you @staabm !

PR #2609, opened 2025-11-28
Build: Allow PHPUnit 12.x

staabm:
Idea is to get some test-coverage for PHPUnit 12.x. AFAICT we don't have test coverage for this major version ATM.

since PHPUnit 12.x requires PHP 8.3+ and PHPUnit 11.x requires PHP 8.2+ we automatically get coverage for both versions, because of our CI matrix with a mix of PHP versions


staabm:
I just found `tests/e2e/PHPUnit12` which seems to be good enough

PR #2618, opened 2025-11-29
refactor(filter): Make it more explicit that isForGitDiffLines is an alias

theofidry:


src/Configuration/ConfigurationFactory.php
```diff
@@ -369,12 +369,9 @@ private function retrieveFilter(string $filter, ?string $gitDiffFilter, bool $is
             return $filter;
         }
 
+        $gitDiffFilter ??= 'AM';
         $baseBranch = $gitDiffBase ?? $this->gitDiffFileProvider->provideDefaultBase();
 
-        if ($isForGitDiffLines) {
-            return $this->gitDiffFileProvider->provide('AM', $baseBranch, $sourceDirectories);
-        }
-
```
maks-rafalko, line 377: I find this less clear tbh. in which way it is more explicit?

in the previous version it's immediately clear that if `isForGitDiffLines` is `true`, then `AM` is used. Now it takes time to understand it, isn't it?

theofidry: Hmm very true.

I was adamant to change it because I found it more confusing before (see the bug fix https://github.com/infection/infection/pull/2617)

I think reverting this however would be wrong because it would introduce back some confusion, and the surrounding code changed a lot in meaning. `$isForGitDiffLines` no longer corresponds to the user passing `--git-diff-lines` but is also true of they pass `--git-diff-filter`.

That said in the caller, the variable passed is `$useGitDiff`, I think we could clear up that confusion with passing forward that rename. There would still be a bit of confusion why we have potentially a null git filter I guess, but it makes clearer that this is "filter vs git filter".

theofidry: https://github.com/infection/infection/pull/2666

PR #2619, opened 2025-11-29
fix(filter): Add the git base branch to the Configuration

theofidry:
Vaguely similar to #2617.

The git base branch can be explicitly passed by the user but we have a fallback, namely `GitDiffFileProvider::provideDefaultBase()`.

However, for some reasons, although we fetch its values in the `ConfigurationFactory`, we did not pass it to `Configuration` unless the user explicitly set it.

Prior to #2614, this was not a problem other than some potential confusion and unnecessary calls made (which should not have any performance impact since we cached the result).

So in a way this is only really a fix because of #2614, but I think it was incorrect before that too (it just didn't cause a hard failure to the user).


tests/phpunit/TestingUtility/Process/GenericProcessException.php
```diff
@@ -0,0 +1,43 @@
...
+declare(strict_types=1);
+
+namespace Infection\Tests\TestingUtility\Process;
+
+use Exception;
+use Symfony\Component\Process\Exception\ExceptionInterface;
+
+final class GenericProcessException extends Exception implements ExceptionInterface
+{
+}
```
sanmai: Looks like something from another PR, #2614?

sanmai (approved):
There's an unrelated file, other than that LGTM

tests/phpunit/TestingUtility/Process/GenericProcessException.php
theofidry: ah yes, good catch

Issue #2622, opened 2025-11-30
Consolidate the Git API


theofidry:
## Context

Infection needs some information from git, namely:

- the git default branch
- the list of files that changed based on the git filter used
- the lines changed within a file to know whether a node should be mutated

So far all of this is provided by `GitDiffFileProvider`.

## Problem

What I don't like about the current implementation:

- `GitDiffFileProvider` also giving the default branch does not make much sense to me (naming wise).
- `Infection\Logger\GitHub\GitDiffFileProvider` makes no sense: this is neither GitHub nor logger related!
- Some parts of the implementation are convoluted:
  - We needlessly modify the commit reference (#2621)
  - `::provideWithLines()` gets a string, explodes + filter, put it back as a string, to be later exploded and filtered again by `DiffChangedLinesParser` which is then memoized.
- The method names no longer make much sense. Originally it was just `GitDiffFileProvider::provide()`, but it now has 2 more methods and the whole group is no longer coherent.

### Simplify the diffLines API

Instead of having `GitDiffFileProvider::provideWithLines()` + `DiffChangedLinesParser::parse()`, I propose a `::getChangedDiffLines()` method.

From an implementation point of view, the output of `::provideWithLines()` is quite unstructured and bound to the underlying implementation. This makes it harder for us to follow what is really the expected output there, and we need to know that here to understand what it outputs, but also when checking `DiffChangedLinesParser::parse()` as it is what it consumes to give a more usable output.

Instead, `::getChangedDiffLines()` would instead give directly the output of the current `DiffChangedLinesParser::parse()` which is easier to explain and document.

From a testing perspective, this also simplifies things. From an implementation perspective, maybe this makes `::provideWithLines()` too complicated, but if that's the case we could still use `DiffChangedLinesParser::parse()`, it would simply be used internally and not exposed.

### A new Git interface

`GitDiffFileProvider` is now pretty much an adapter for the `git` command line. But equally, the fact that we call `git` from the command line with processes is an implementation detail.

Taking into account the previous point, I propose the following API:

```php
namespace Infection\Git;

interface Git {
  function getDefaultBaseBranch(): string; // previously ::provideDefaultBase()
  function getChangedFileRelativePaths(): string; // previously ::provide()
  function getChangedDiffLines(): string; // previously ::provideWithLines() + DiffChangedLinesParser::parse()
}
```

[...]

PR #2623, opened 2025-11-30
refactor(git): Introduce a Git interface

theofidry:
This PR introduces a new `Git` interface as per #2622.

Besides the renaming and moving the classes, no other change was done.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -15,8 +15,44 @@ Contributions are always welcome. Here are a few guidelines to be aware of:
 
 ## Tests
 
-To run the tests locally, you can run `make test`. It however requires [Docker][docker]. For more
+To run the tests locally, you can run `make test`. It, however, requires [Docker][docker]. For more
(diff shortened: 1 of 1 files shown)


src/Git/CommandLineGit.php
```diff
@@ -48,14 +48,12 @@
...
 
 /**
- * @final
- *
  * @internal
+ *
+ * Implementation of the Git contract leveraging the git binary via processes.
  */
-class GitDiffFileProvider
+final class CommandLineGit implements Git
```
sanmai, line 55: Not super fond of this plain `Git` because it is unclear what it does, otoh I can't think of a better name so let's keep it.

sanmai (approved):

src/Git/CommandLineGit.php
theofidry: Yeah, at the end it's pretty much for "git stuff that infection uses". Hopefully the method names give a bit more clarify and if we have something a lot more specific as well, it can be its own API and in the `Git` namespace instead of jamming it here.

PR #2628, opened 2025-12-01
test(git): Move the setup of the service to the testcase setup

theofidry:


tests/phpunit/Git/CommandLineGitTest.php
```diff
@@ -56,17 +57,26 @@
...
+    private ShellCommandLineExecutor&MockObject $commandLineMock;
+
+    private Git $git;
+
+    protected function setUp(): void
+    {
+        $this->commandLineMock = $this->createMock(ShellCommandLineExecutor::class);
+
+        $this->git = new CommandLineGit($this->commandLineMock);
+    }
```
maks-rafalko, line 69: I'm pretty sure that as a best practices it was suggested (in one of the Sebastian's talk) to create such services in a private method and reuse in each test method.

What was the reason to move it to `setUp` and in which way is it better?

theofidry: >What was the reason to move it to setUp and in which way is it better?

Let's boilerplate: you can access the configured service right away and adjust the mock configuration as needed.

PR #2631, opened 2025-12-01
test(git): Add integration tests

theofidry:
Related to #2622.

Mocking excessively for `CommandLineGit` is fine for asserting behaviour, but there is no real guarantee this is how it really behaves in a real world.

I planned to introduce some changes but was a bit unsure if it would just not end up really breaking the integration somehow and I would like a faster feedback loop and stronger guarantee that running the end-to-end test suite.

Whilst having mocks still make sense for clearer and more detailed tests, I think we could also have some smoke integration tests.

Edit: welp, I realise this need the appropriate fetch depth which I don't think is good to rely on. So that means that test would only be useful locally.


theofidry:
I think it's still valuable, but if you disagree I'll remove it.

PR #2633, opened 2025-12-01
docs(git): Document why we try to do a git merge-base <userGitDiffBase> HEAD

theofidry:


src/Git/CommandLineGit.php
```diff
@@ -124,6 +124,24 @@ public function provideWithLines(string $baseBranch): string
...
+    //
+    //  main:     A --- B --- C --- F --- G --- H
+    //                         \
+    //  feature:                D --- E  (user changes)
+    //
+    // Then `git diff main HEAD` will give (D,E,F,G,H). So infection would
+    // touch code the user did not touch.
+    //
+    // To prevent this, we try to find the best common ancestor, here C.
+    // As a result, we would do `git diff C HEAD` which would give (D,E).
```
maks-rafalko, line 144: git diff merge-base C head?

is it a mistake?

theofidry: We would do `git merge-base main HEAD` which would give `C`, hence the new base is `C` and later when computing the diff the command will be `git diff C HEAD`. I guess I need to rephrase a bit

PR #2637, opened 2025-12-01
test(git): Use a custom Git implementation for the ConfigurationFactoryTest

theofidry:
Since the Git interface has been introduced in #2623, it is simpler to introduce a dedicated implementation rather than a mock. Indeed, it is easier to maintain the various methods with it as the parameter types and names are part of the constract.

### DIFF
--- a/tests/phpunit/Configuration/ConfigurationFactory/ConfigurationFactoryTest.php
+++ b/tests/phpunit/Configuration/ConfigurationFactory/ConfigurationFactoryTest.php
@@ -85,7 +85,7 @@ final class ConfigurationFactoryTest extends TestCase
 {
     private const GIT_DEFAULT_BASE_BRANCH = 'test/default';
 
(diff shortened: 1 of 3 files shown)


tests/phpunit/Configuration/ConfigurationFactory/ConfigurationFactoryTest.php
```diff
@@ -1415,7 +1415,7 @@ static function (array $source, array $excludes) use ($schema) {
                     count($sourceDirectories) === 0
                         ? ''
                         : implode(', ', $sourceDirectories),
-                    self::GIT_DIFF_FILTER_PROVIDER_RESULT,
+                    self::GIT_DIFF_MODIFIED_FILES,
```
theofidry, line 1418: ofc forgot to remove the implementation here... _sight_

PR #2638, opened 2025-12-02
refactor(git): Inline `DiffChangedLinesParser` 

theofidry:


tests/phpunit/Differ/DiffChangedLinesParserTest.php
```diff
@@ -1,128 +0,0 @@
...
-    #[DataProvider('provideDiffs')]
-    public function test_it_converts_diff_to_files_and_changed_lines_map(
-        string $diff,
-        array $expectedMap,
-    ): void {
-        $collector = new DiffChangedLinesParser();
-
-        $resultMap = $collector->parse($diff);
-
-        $this->assertSame($this->convertToArray($expectedMap), $this->convertToArray($resultMap));
```
theofidry, line 64: note that this is no longer necessary and can be simplified with `assertEquals()`

tests/phpunit/Differ/DiffChangedLinesParserTest.php
```diff
@@ -1,128 +0,0 @@
...
-    }
-
-    public static function provideDiffs(): Generator
```
theofidry, line 67: double check the provide naming & signature in the project

tests/phpunit/Differ/DiffChangedLinesParserTest.php
```diff
@@ -1,128 +0,0 @@
...
-    {
-        yield 'one file with added lines in different places' => [
```
theofidry, line 69: add more tests first: we do not want to introduce bugs as we refactor

tests/phpunit/Differ/DiffChangedLinesParserTest.php
```diff
@@ -1,128 +0,0 @@
...
-                @@ -37,0 +38 @@ namespace Infection;
-                @@ -533 +534,2 @@ final class Container
-                @@ -535,0 +538,3 @@ final class Container
-                @@ -1207,0 +1213,5 @@ final class Container
-                diff --git a/src/Differ/FilesDiffChangedLines.php b/src/Differ/FilesDiffChangedLines.php
-                new file mode 100644
-                @@ -0,0 +1,18 @@
-                DIFF,
-            [
-                realpath('src/Container.php') => [
```
theofidry, line 99: we could use the FS to mock here instead of real paths... Those are annoying for this test (and I suppose why it's an integration test)

src/Git/CommandLineGit.php
```diff
@@ -105,19 +115,34 @@ public function getChangedFileRelativePaths(string $diffFilter, string $base, ar
...
             'diff',
             $base,
             '--unified=0',
             '--diff-filter=AM',
         ]);
-        $lines = explode(PHP_EOL, $filter);
-        $lines = array_filter($lines, static fn (string $line): bool => preg_match('/^(\\+|-|index)/', $line) === 0);
 
-        return implode(PHP_EOL, $lines);
+        $diffLines = explode(PHP_EOL, $diff);
```
theofidry, line 128: extract the simplification/refactoring in a separate PR

PR #2381, opened 2025-08-25
PoC: SourceCollector

theofidry:
A draft for #2380.

The goal of this PR is to figure out actionables that can be extracted and merged.

PR #2648, opened 2025-12-02
fix: Throw the correct exception then no appropriate source is found

theofidry:
In Infection, in general, we want to throw a helpful exception if we could not find any source code to mutate. This can happen at different levels:

- When collecting the configured source (+ potentially filters).
- When reading the coverage report which indicates no code is covered by the test framework.
- When getting the changed lines (as: there is no modified lines at all for any file).

Now, when the source is _filtered_, i.e. it is a subset of the configured source, we decided to not fail. The rationale being that if there is no code to mutate for a full run, there is a configuration issue, and otherwise it is simply that there is no applying mutations.

Until now we had various exceptions, sometimes not correctly documented, thrown at different places. This PR attempts to unify all of those cases under one exception `NoSourceFound`. In the end, cause is the same: there is no source (files, lines, covered by tests) to mutate.
  - What differs, is the **reason**, which can be elaborated on by the **message**.
  - **How we react**, is based on the context (whether the source is filtered), which can be conveyed by a property (`NoSourceFound#isSourceFiltered`).

Note that the current location may seem a bit weird, but there is soon more stuff that is going to land there (you can have a peak at #2380 and #2381).

### DIFF
--- a/src/Git/CommandLineGit.php
+++ b/src/Git/CommandLineGit.php
@@ -111,8 +111,6 @@ final readonly class CommandLineGit implements Git
 
     public function getChangedLinesRangesByFileRelativePaths(string $diffFilter, string $base): array
     {
-        $filter = 'AM';
-
         $diff = $this->shellCommandLineExecutor->execute([
             'git',
             'diff',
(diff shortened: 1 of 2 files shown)


src/Git/Git.php
```diff
@@ -103,6 +104,8 @@ public function getChangedFileRelativePaths(
      * ]
      * ```
      *
+     * @throws NoSourceFound
+     *
      * @return array<string, list<ChangedLinesRange>>
```
theofidry:
```suggestion
     * @return non-empty-array<string, list<ChangedLinesRange>>
```

PR #2649, opened 2025-12-02
fix(git): Do not include invalid changed lines ranges

theofidry:
Related to #2622.

In theory, the raw git diff lines may include changes for which there is no changed line in the new file. If that happens, currently, we create a `ChangedLinesRange` object nonetheless, which is incorrect.

Whilst it is a fix, it currently has no impact for the users. Indeed, we apply the git filter `AM`. For this (the buggy) scenario to occur, you need to have and old file for which the lines are deleted, without any other change. This cannot happen with that filter.

Nonetheless, this is still worth fixing as the filter may (and is likely) to change in the future.

### DIFF
--- a/tests/phpunit/Git/CommandLineGitTest.php
+++ b/tests/phpunit/Git/CommandLineGitTest.php
@@ -133,10 +133,10 @@ final class CommandLineGitTest extends TestCase
     ): void {
         $this->commandLineMock
             ->method('execute')
-            ->with(['git', 'diff', 'main', '--unified=0', '--diff-filter=AM'])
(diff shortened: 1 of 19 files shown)


tests/phpunit/Git/CommandLineGitTest.php
```diff
@@ -512,6 +503,24 @@ public static function gitChangedLinesRangesProvider(): iterable
                 ],
             ],
         ];
+
+        yield 'only some files have changed lines in their new files' => [
```
theofidry, line 521: to also include the case where the same file has several deletes, but one modification too.

What cannot happen in practice is that a file has only deletes, because the filter `AM` would filter it out, but a specific diff for that file can still have a deletion only.

PR #2650, opened 2025-12-02
fix(git): Respect the git filter specified by the user

theofidry:
Related to #2622.

If the user specifies a filter, e.g. `--git-diff-filter=A`, then we do not apply that filter for the changed git diff lines.

In practice, I cannot see a case where this would be a problem for the user. Indeed, if the user requests `A`, `M` or `D`, the only thing that can happen is that we are getting more lines than necessary. To be a problem, you would need to see different changed lines for the same file. AFAIK this cannot happen, as the `--diff-filter` option of git influences which files are shown, but do not influence the content.

Nonetheless, there is three benefits:

- Make it explicit: one reading the code won't be left wondering why we don't respect the filter (hi me).
- Bail out earlier: if there is no changed lines we throw an exception to indicate that there is no node to mutate.
- Reduce the size the output, which can be a performance improvement depending of its size.

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -150,6 +150,7 @@ use PhpParser\PrettyPrinter\Standard;
 use PhpParser\PrettyPrinterAbstract;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
+use ReflectionClass;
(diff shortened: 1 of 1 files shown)


sanmai (approved):
reviewed the code, makes sense

PR #2651, opened 2025-12-02
fix(git): Pass the option correctly to the git process

theofidry:
Related to https://github.com/infection/infection/issues/2622.

AFAIK the canonical way to pass an option is `--long-name=value`. In our case this is not a problem because `git` supports both, but we might as well use the correct form.

### DIFF
--- a/tests/phpunit/Git/CommandLineGitTest.php
+++ b/tests/phpunit/Git/CommandLineGitTest.php
@@ -133,7 +133,7 @@ final class CommandLineGitTest extends TestCase
     ): void {
         $this->commandLineMock
             ->method('execute')
-            ->with(['git', 'diff', 'main', '--unified=0', '--diff-filter', 'AM'])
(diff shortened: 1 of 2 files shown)


theofidry:
@sanmai could you confirm? I'm actually not certain

tests/phpunit/Git/CommandLineGitTest.php
```diff
@@ -105,7 +105,7 @@ public function test_it_gets_the_relative_paths_of_the_changed_files_as_a_string
         $this->commandLineMock
             ->method('execute')
             ->with(
-                ['git', 'diff', 'main', '--diff-filter', 'AM', '--name-only', '--', 'app/', 'my lib/'],
+                ['git', 'diff', 'main', '--diff-filter=AM', '--name-only', '--', 'app/', 'my lib/'],
```
theofidry, line 108: wait on #2650 which will need to be adjusted too.

PR #2662, opened 2025-12-03
docs(git): Document our decision on not couting a renamed file as having all of its lines changed

theofidry:
I stumbled upon this case in ##2658 and wanted to shed some light on it.

As a reminder, currently if a file is renamed, none of its lines appears in the diff meaning there is no line to mutate. There was a question of whether this is correct.

We decided that we cannot see a case where we would want to change that behaviour. Whilst the test is already there, this PR adds a comment to document that this is a conscious decision.

<hr/>

_Originally post_

Title: fix(git): Count all lines of a renamed file as modified

Currently if a file is renamed, whilst it is listed as part of the changed files, hence parsed, we do not compute any changed lines.

This PR fixes that, but it begs the question: is it what we want?

### DIFF
--- a/tests/phpunit/Git/CommandLineGitTest.php
+++ b/tests/phpunit/Git/CommandLineGitTest.php
@@ -577,6 +577,12 @@ final class CommandLineGitTest extends TestCase
                 --- a/src/Container.php
                 +++ b/src/LegacyContainer.php
                 DIFF,
+            // There is no new or updated code, hence nothing to mutate.
+            // While discussing this, we thought _maybe_ there was a very narrow edge case
+            // where this _may_ not be true, but it is not compelling enough to change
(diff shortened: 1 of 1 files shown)


maks-rafalko:
> This PR fixes that, but it begs the question: is it what we want?

no new or updated code - no need to mutate, that's what I would expect

theofidry:
@maks-rafalko but changing a file name has side effects: it affects autoloading, file location...

Hmm... Thinking about this a bit more, unless the code was using its own location, e.g. `__FILE__.'/../../to/smth'`, this won't be problem. And the code using that moved file will have a line of code changed.

So I guess it's too narrow of a case.

theofidry:
Reminder for myself: I think it would be worth documenting this decision in the test still

PR #2666, opened 2025-12-04
refactor(git): Remove confusing fallback

theofidry:
Addresses the feedback of https://github.com/infection/infection/pull/2618/files#r2586971034.

This is actually a great spot because the original piece of code was confusing and my changes only removed some confusion. Eventually though, I could simplify things (see #2650) where, to keep things simpler, `$isForGitDiffLines` which previously corresponded to the user passing `--git-diff-filter`, is sorted in `RunCommand` and no longer forwarded to `ConfigurationFactory`.

As a result, this means this confusing piece of code is actually dead code.

### DIFF
--- a/src/Configuration/ConfigurationFactory.php
+++ b/src/Configuration/ConfigurationFactory.php
@@ -389,15 +389,15 @@ class ConfigurationFactory
     private function retrieveFilter(
         string $filter,
         ?string $gitDiffFilter,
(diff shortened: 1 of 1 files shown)


theofidry:
Note that the escaped mutation is a false-positive based on `RunCommand`, but would otherwise be legit. I hope to be able to fix that with a clearer API in the future.

PR #2668, opened 2025-12-04
feat(git): Add a logger

theofidry:
Inject a logger to `CommandLineGit` to log whenever the process used failed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -79,7 +79,7 @@
     "require-dev": {
         "ext-simplexml": "*",
         "fidry/makefile": "^1.0",
-        "fig/log-test": "^1.0",
+        "fig/log-test": "^1.2",
         "phpbench/phpbench": "^1.4",
         "phpstan/extension-installer": "^1.4",
         "phpstan/phpstan": "^2.1",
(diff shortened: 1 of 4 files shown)


src/Container.php
```diff
@@ -560,7 +560,10 @@ public static function create(): self
                 );
             },
             MemoizedComposerExecutableFinder::class => static fn (): ComposerExecutableFinder => new MemoizedComposerExecutableFinder(new ConcreteComposerExecutableFinder()),
-            Git::class => static fn (): Git => new CommandLineGit(new ShellCommandLineExecutor()),
+            Git::class => static fn (self $container): Git => new CommandLineGit(
+                new ShellCommandLineExecutor(),
+                $container->getLogger(),
```
maks-rafalko, line 566: for my understanding: we are using `Infection\Logger\Logger` which implements `Psr\Log\LoggerInterface`, right?

maks-rafalko (approved):

src/Container.php
theofidry: correct

PR #2669, opened 2025-12-04
feat(git): Add git commands

theofidry:
This PR adds a bunch of commands essentially exposing our Git API to the CLI.

The intent here is to somewhat ease debugging in some scenarios. You may get away with a `var_dump();die;` statement sometimes, but it is sometimes not possible (e.g. when using the PHAR).

Now, one there is one big concerns here which is maintainability: do we want to main them. It's not that much code, neither is it critical by any means. But maybe we don't want to have that around. Personally I would like more such utility commands for various debugging purposes, but maybe you do not.

### DIFF
--- a/src/Command/Git/GitBaseReferenceCommand.php
+++ b/src/Command/Git/GitBaseReferenceCommand.php
@@ -35,15 +35,10 @@ declare(strict_types=1);
 
 namespace Infection\Command\Git;
 
-use function array_fill_keys;
 use Infection\Command\BaseCommand;
 use Infection\Command\Git\Option\BaseOption;
 use Infection\Console\IO;
-use Psr\Log\LoggerInterface;
-use Psr\Log\LogLevel;
 use function sprintf;
-use Symfony\Component\Console\Logger\ConsoleLogger;
-use Symfony\Component\Console\Output\OutputInterface;
 
 /**
  * @internal
@@ -66,7 +61,7 @@ final class GitBaseReferenceCommand extends BaseCommand
 
     protected function executeCommand(IO $io): bool
     {
-        $logger = self::createLogger($io->getOutput());
+        $logger = LoggerFactory::create($io->getOutput());
         $base = BaseOption::get($io);
 
         $git = $this->getApplication()->getContainer()->getGit();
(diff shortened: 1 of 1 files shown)


theofidry:
@maks-rafalko thoughts?

maks-rafalko:
it's hard to comment here as I feel like I never had a need in such commands in Infection context

what are the use cases?

theofidry:
I guess mostly debugging without having to resort to `var_dump()` calls, which are also more annoying to pull off when using a PHAR.

Not a common scenario for sure though.

src/Source/Collector/GitDiffSourceCollector.php
```diff
@@ -95,7 +95,7 @@ private static function convertToPlainFilter(
         return PlainFilter::tryToCreate(
             $git->getChangedFileRelativePaths(
                 $sourceFilter->value,
-                $sourceFilter->base,
+                $sourceFilter->reference,
```
theofidry: to extract

src/Git/Git.php
```diff
@@ -125,7 +125,7 @@ public function getChangedLinesRangesByFileRelativePaths(
     ): array;
 
     /**
-     * Find as good common ancestors as possible for a merge and falls back to the given base otherwise.
+     * Find as good common ancestors as possible with HEAD for a merge and falls back to the given base otherwise.
```
theofidry, line 128: to extract

src/Configuration/SourceFilter/GitDiffFilter.php
```diff
@@ -43,12 +43,12 @@
...
     /**
-     * @param non-empty-string $value
-     * @param non-empty-string $base
+     * @param non-empty-string $value E.g. 'AMD'
+     * @param non-empty-string $reference The best common ancestor git reference when possible.
      */
     public function __construct(
         public string $value,
-        public string $base,
+        public string $reference,
```
theofidry: to extract

src/Command/ListSourcesCommand.php
```diff
@@ -59,7 +59,7 @@ public function __construct()
     protected function configure(): void
     {
         ConfigurationOption::addOption($this);
-        FilterOptions::addOption($this);
+        SourceFilterOptions::addOption($this);
```
theofidry, line 62: to extract

theofidry:
I'd like to try.

If it turns out they are too problematic or too annoying to maintain we can always remove them in the future, they are not really core of the user-experience so I doubt it will be an issue.

PR #2684, opened 2025-12-09
feat: Make PlainFilter null if contains no values

theofidry:
Transform `PlainFilter::create()` to `::tryToCreate()`. This allows us to harden the type of `PlainFilter#values` to be a non-empty array.

Side note: "tryCreate" would be more idiomatic in C#/.net, Java or JS, but I don't find that omitting the "to" brings that much clarity and makes the code less expressive IMO.

### DIFF
--- a/src/FileSystem/SourceFileCollector.php
+++ b/src/FileSystem/SourceFileCollector.php
@@ -69,7 +69,7 @@ class SourceFileCollector
         private readonly array $excludedFilesOrDirectories,
         private readonly ?PlainFilter $filter,
     ) {
-        $this->filtered = count($this->filter->values ?? []) !== 0;
+        $this->filtered = $this->filter !== null;
     }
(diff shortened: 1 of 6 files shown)


src/FileSystem/SourceFileCollector.php
```diff
@@ -170,10 +170,9 @@ private function createFinder(): Finder
     private function filter(Iterator $iterator): Iterator
     {
         // TODO: could use Finder::setFilter() instead!
-        if ($this->isFiltered()) {
+        if ($this->filter !== null) {
```
theofidry, line 173: note that here the previous code is fine. I prefer this form however as PHPStan understands it better, removing the need of the `@phpstan-ignore` bellow

theofidry:
Thanks for the quick review @sanmai 😄

PR #2688, opened 2025-12-10
refactor: Allow to pass a SourceFilter to the Configuration

theofidry:
In the future we will be able to pass the `GitDiffFilter` directly instead of converting it to a `PlainFilter`.


theofidry:
Hmm looks like this can't be extracted.

PR #2697, opened 2025-12-10
feat(SourceCollector): Introduce a LazyCacheSourceCollector

theofidry:
This PR introduces `LazyCacheSourceCollector` which allows to defer the creation of the collector and takes care of caching the collected entries.

For the sake of simplicity, and my sanity, starting the collection triggers the entire collection (instead of streaming it) and sub-sequent calls get the cached array. Otherwise this easily gets too complicated, for example if two services get two distinct iterables and start consuming at their own pace...

Fair warning, I was tired so I mostly used Claude for that one, need a fresh review.

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -440,16 +440,12 @@ final class Container extends DIContainer
                 $container->getMemoryFormatter(),
                 $container->getConfiguration()->threadCount,
             ),
-            FileMutationGenerator::class => static function (self $container): FileMutationGenerator {
-                $configuration = $container->getConfiguration();
-
-                return new FileMutationGenerator(
-                    $container->getFileParser(),
-                    $container->getNodeTraverserFactory(),
-                    $container->getLineRangeCalculator(),
-                    $container->getSourceLineMatcher(),
-                );
-            },
+            FileMutationGenerator::class => static fn (self $container): FileMutationGenerator => new FileMutationGenerator(
+                $container->getFileParser(),
+                $container->getNodeTraverserFactory(),
+                $container->getLineRangeCalculator(),
+                $container->getSourceLineMatcher(),
+            ),
             FileLoggerFactory::class => static function (self $container): FileLoggerFactory {
                 $config = $container->getConfiguration();
 
@@ -721,6 +717,9 @@ final class Container extends DIContainer
         return $this->get(FileSystem::class);
     }
 
+    /**
+     * @throws NoSourceFound
+     */
     public function getUnionTraceProvider(): UnionTraceProvider
     {
         return $this->get(UnionTraceProvider::class);
@@ -866,6 +865,9 @@ final class Container extends DIContainer
         return $this->get(FilteringResultsCollectorFactory::class);
     }
 
+    /**
+     * @throws NoSourceFound
+     */
     public function getTestFrameworkAdapter(): TestFrameworkAdapter
     {
         return $this->get(TestFrameworkAdapter::class);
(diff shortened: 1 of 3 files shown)


src/Source/Collector/LazyCacheSourceCollector.php
```diff
@@ -0,0 +1,106 @@
...
+    {
+        if ($this->cache === null) {
+            /** @psalm-suppress InvalidArgument */
+            $this->cache = iterator_to_array(
+                $this->getCollector()->collect(),
+                preserve_keys: false,
+            );
+        }
+
+        return new ArrayIterator($this->cache);
```
sanmai, line 88: It sounds like we are at a fork in the road here:

- Either we do the work that we may not need (collect all files before they are actually required)
- Or we limit ourselves to the number of consumers (for generators, there could be only one, unless we use `NoRewindIterator` here)

Surely Claude can whip up a custom buffering aggregate iterator that buffers the output of any internal iterators and reuses the buffered output for any new external iterators. 

How do you explain that this complexity is essential and unavoidable? 

What is the problem that needs to be solved?

theofidry: >How do you explain that this complexity is essential and unavoidable?

I thought the current solution was _simple_ 😅 

Having `SourceCollector->collect()` called in different places is reasonable. So either it needs to be stored in the container somehow (previously was done in `Configuration`) or cached in a Decorating source collector. I opted for the latter, because having a caching decorator seems natural to me and removing stuff from the `Configuration` enabled quite a few simplifications. Even if we find this stuff here complex, it is only one implementation behind an interface, rather than `Configuration` with multiple parameters that leaks in different places.

The fundamental issue otherwise is `iterable`. It is too wide to have much an expectation on what it actually is from the consumer perspective: an array? a non-rewindable traversable? an iterator? a generator?

so if you have:

```php
// in a service
$source1 = $sourceCollector->collect();
// in another service
$source2 = $sourceCollector->collect();

function take1(iterable $iterable): mixed {
  foreach($iterable as $item) {
    // in real code this would be more due to a `yield` triggered by the first service
    // and then service2 takes over at some point, before service1 finishes to yield everything
    return $item;
  }
}

take1($source1);
take1($source2);
```

What would be your expectations? If you respect a streaming nature, then:

```php
take1($source1); // => $file0
take1($source2); // => $file1 <- we continued!
```

I think that would be very error prone. But having something that does streaming + cache for a shared `iterable` is... a 100 times more complex or something. And to test you would need about 5 different other iterable related utilities.

So I went for a much simpler solution: the value collection is done whenever you request to collect the files and this is cached.

<hr>

I realise however that the current implementation is a bit lazy. There is no reason for it to be in one collector, we could have a lazy collector and a cache collector.

Also `Lazy` may not be the best name: we make the instantiation of the source collector lazy, but not the collection, i.e. `$lazyCollector->collect()` starts the computation of `::collect()`, which may not be what you expect from a lazy service... Not unheard of, there is no consensus there, but may be surprising.

sanmai: Right, so we did not have an _outside intention_ to make the original source collector produce an iterator - it was done purely for convenience's sake. In this case, whether we convert the whole thing into an array or not should not matter.

sanmai: If we can make it an array (i.e., it can fit into memory), then the problem of iterating over the same iterator twice is best avoided.

LMK if you want to review this again once you're done.

theofidry: I went with `iterable` because we tended to use generators in practice with the convenience of swapping them for arrays in tests.

But before all the source collection was loaded at once for the Configuration, so I guess it's not a downgrade to enforce an array.

So if you're happy with that I'll update the API separately and rebase this PR afterwards.

theofidry:
@sanmai splitted this PR in separate, simpler ones:

- https://github.com/infection/infection/pull/2699
- https://github.com/infection/infection/pull/2700
- https://github.com/infection/infection/pull/2702

PR #2699, opened 2025-12-11
feat(SourceCollector): Introduce a lazy implementation of the source collector

theofidry:
This introduces a new adapter to defer the instantiation of the source collector.

Note that this could be made redundant by having the container supporting lazy services out of the box, but that is not the case yet.


theofidry:
@sanmai as mentioned on Discord, fell free to raise concerns/issues, I'm happy to address them afterwards even if the PR is merged.

maks-rafalko:
> This introduces a new adapter to defer the instantiation of the source collector.

could you explain why do we need this?

theofidry:
It's not needed per say, but instantiating a collector, for the git one at least, results in some extra work. Without laziness, this work is done whenever  `SourceCollector` is required, e.g. to instantiate a service that uses it.

With laziness, the collection of the files and heavy construction instantiation is deferred to when we actually use the service.

PR #2700, opened 2025-12-11
feat(SourceCollector): Make the SourceCollector return an array instead of an iterable

theofidry:
After the discussion of https://github.com/infection/infection/pull/2697#discussion_r2609295618, it showed that there i just no need for us to use an `iterable` and that doing so causes too many issues, for no added benefit.


maks-rafalko:
could you please explain why was it needed?

theofidry:
@maks-rafalko updated the description

PR #2703, opened 2025-12-11
feat(SourceCollector): Fail earlier when no source file is found

theofidry:
## Problem

Currently, if no source file is found, either because the configured source in `infection.json5` results in no file or because of the `--filter` applied, then Infection still executes the initial test run.

## Solution

This PR changes the contract of `SourceCollector` to document the throw of `NoSourceFound` and make `BasicSourceCollector` throw if no source is found. Now, the above scenario will fail before the initial test run is attempted. Examples of exceptions thrown:

```shell
# no filter applied
In NoSourceFound.php line 98:
No source file found for the configured sources.

# filter applied
In NoSourceFound.php line 98:
No source file found for the filter applied to the configured sources. The filter used was: "src/Unknown1,src/Unknown2".
```

The exception thrown respects whether a filter is applied, i.e. in the 2nd scenario we would would terminate with a success error code.

Note that this behaviour was already present if the user used `--git-diff-lines` or `--git-diff-filter`. `GitDiffSourceCollector::collect()` was failing (actually it is `::create()`, but in practice we lazily instantiate this service hence this only happens when calling `::collect()`).

### Change 1)

One (very minor) issue is due to our simplistic implementation of `GitDiffSourceCollector`, in which we re-use `BasicSourceCollector`, we could have a slightly confusing exception message. Indeed, if the git diff itself returns changed files, but none of them are part of the configured source, then we would end up with the message from above. It could be improved by mentioning information about the git info we are using (like we do if no file is found from git diff).

This could either be fixed by re-implementing `GitDiffSourceCollector` in a less lazy way or or catching the exception and rethrowing it with a corrected message, but it is not the scope of this PR.

### Change 2)

Maybe obvious, but it now means that where previously the test were executed regardless of whether a filter was applied or any source code was configured, this is now no longer true.


theofidry:
Hmm for some reason running the following:

```shell
./tests/e2e_tests bin/infection.phar tests/e2e/Ignore_MSI_Zero_Mutations

Success
```

Doesn't seem to do anything for me... It's instantaneous, as if nothing is tested, and it does fail in the CI.

src/Source/Collector/BasicSourceCollector.php
```diff
@@ -73,17 +74,13 @@ public function isFiltered(): bool
...
-            return [];
-        }
+        $files = $this->doCollect();
 
-        $iterator = $this->filter(
-            $this
-                ->createFinder()
-                ->getIterator(),
-        );
+        if (count($files) === 0) {
```
maks-rafalko, line 79: I'm a bit confused. 

I thought previously we were reading files using generator. Why is it an array now? Or do I miss something?

theofidry: >I thought previously we were reading files using generator.

Correct. Initially the type of the interface was `collect(): iterable`. However, this was changed to `array` in  #2700.

As to why this was elaborated in https://github.com/infection/infection/pull/2697#discussion_r2609295618.

Maybe we'll still need to do that eventually at some point, because I kinda fear we have the same problem elsewhere, but we can revisit that at any time.

Also note that before any of this `SourceCollector` refactoring, we had the files collected in `ConfigurationFactory` and stored as `SplFileInfo[]` in `Configuration`. We still did additional filtering on top of it afterwards, but that initial collection was done eargerly. So compared to the starting point, there should not be any performance degradation.

PR #2272, opened 2025-07-01
PoC of a new architecture

theofidry:
**This PR is not really meant for review yet.**

Over the last years I spent quite a bit of time reading various research papers (the publicly available ones are listed on https://github.com/theofidry/awesome-mutation-testing?tab=readme-ov-file#publications).

After a few ideas being discussed recently, I carried on with an idea that I barely started long ago to review a bit more the mutation testing process and how things should work whilst being easy to use, debug and experiment.

So I went through that process and documented it in `workflow.md`. There is still a lot of unclear pieces, but at the very least it should capture enough items to ensure everything that is already done within Infection fits that model, and that future ideas will fit in just as well.

Translating those ideas to code is always more delicate so wrote some pseudo-code to clarify that.

The short-term goal is to create new RFC issues with a clearer purpose and actionable, but have this PR as a more general background and reference.

- [x] https://github.com/infection/infection/issues/2380
- Tracing
  - [x] Tracer: https://github.com/infection/infection/issues/2751
  - [ ] Remove `TraceProviderAdapterTracer` (depends on the `Coverage` changes)
- Coverage: https://github.com/infection/infection/issues/2816
  - [ ] Consolidate the existing API: https://github.com/infection/infection/pull/2543
  - [ ] Introduce an interface
  - [ ] Leverage the new PHPUnit features
  - [ ] Propose a new format: https://github.com/infection/infection/issues/2600
- Target mode
  - [x] Improve the current base: https://github.com/infection/infection/issues/2595
  - [x] Allow a target mode: https://github.com/infection/infection/issues/2191
  - [ ] Allow to target specific symbols: https://github.com/infection/infection/issues/2237
- ~~AST enrichment (pre-traverse)~~ https://github.com/infection/infection/issues/2797
  - [x] AST Testing framework: https://github.com/infection/infection/pull/2780
  - [x] Move traverse logic to pre-traverse: https://github.com/infection/infection/pull/3026
- MutationAnalyser
  - [ ] Heuristics
  - [ ] MutantProcess
- [ ] https://github.com/infection/infection/issues/2866
  - [x] https://github.com/infection/infection/pull/3095
  - [ ] https://github.com/infection/infection/issues/222
- [ ] Test framework
  - Consolidate the current API:
    - [ ] https://github.com/infection/infection/issues/2863
    - [ ] https://github.com/infection/infection/issues/3082
    - [ ] https://github.com/infection/infection/issues/3424
  - Support more test frameworks
[...]

PR #2707, opened 2025-12-13
feat(Tracer): Introduce the Tracer

theofidry:
Depends on #2708.

This PR completes a 6 years old challenge: https://github.com/infection/infection/pull/927

We introduced back then `Trace`, which is an envelope for a source file which contains all the necessary data about its tests. The `Tracer` is the service that provides such a trace object for a given source file.

## Motivation

Currently we work as follow:

- go through the coverage data (the XML coverage index, in `PhpUnitXmlCoverageTraceProvider`)
- for each tests, find the source the test is about
- filter out the tests for which the source is not part of the collected sources (this is `CoveredTraceProvider` using `BufferedSourceFileFilter`)
- if we allow non-covered code, we create a `ProxyTrace` for each source file that was not found in the tests (this is `UncoveredTraceProvider` using `BufferedSourceFileFilter`)
- for each trace, start the mutation generation process

One one hand this makes sense because we start from the data we have: the coverage. On another, it seems a bit weird to start from the coverage which may contain a lot more than necessary, especially if we request specific source files (via filters).

In [Architecture PoC](https://github.com/infection/infection/pull/2272/files#diff-9bc3105e506d7103f2b25138650f3e784a373d04b529f12ff34136a2fccc72c6), the proposal is to:

- parse the source file
- enrich the AST (mark what code is eligible, dry areas, add types, etc.)
- starting the mutation generation process

So for that purpose, the current architecture is unfit as we start from the wrong end. The `Tracer` allows to inverse the control flow: we request the data (the `Trace`) for a given source file, not the other way around.

## The solution

The API of the Tracer is fairly simple, and in all implementation attempts, it always ended up with the same API.

The implementation idea draws inspiration from some experiments I did when working on the coverage data in https://github.com/infection/infection/pull/2543. The idea is:

- make the trace provider into a generator
- to lookup the information, we traverse the generator, caching the results as we go
- once we found the desired result, we pause the traverse and yield that result
- once we request more information, if it is not in the cache, we resume the traversal

The current implementation may feel somewhat redundant and a bit of a patchwork, and that is because it is. The plan is to "inline" `TraceProvider` entirely (I started some work towards that in #2543). It changes nothing to the general principle however.

Some benchmarks were conducted in https://github.com/infection/infection/pull/2708#issuecomment-3659713580 and I extended the current tracing benchmarks.

The TL:DR; is that this should not result in any performance degradation. The gains are quite unpredictable by their nature, but the idea is that this API will _enable_ gains. See the PR for a more detailed explanation.


src/Framework/IterableCounter.php
```diff
@@ -33,9 +33,10 @@
 
 declare(strict_types=1);
 
-namespace Infection;
+namespace Infection\Framework;
```
theofidry: to extract (and to not forget to move its test too)

tests/phpunit/Mutation/FileMutationGeneratorTest.php
```diff
@@ -377,32 +413,11 @@ public static function provideBoolean(): iterable
...
-        }
-
-        return $proxyTraceMock;
-    }
-
-    private function createSplFileInfoMock(string $file, string $relativePath, string $relativePathname): SplFileInfo&MockObject
-    {
+        string $relativePath = '',
+        string $relativePathname = '',
+    ): SplFileInfo&MockObject {
```
theofidry, line 428: there is some stuff that could be extracted here

src/TestFramework/PhpUnit/PhpUnitTracer.php
```diff
@@ -0,0 +1,138 @@
...
+use Infection\TestFramework\Tracing\Trace\Trace;
+use Infection\TestFramework\Tracing\TraceProvider;
+use Infection\TestFramework\Tracing\Tracer;
+use SplFileInfo;
+use Webmozart\Assert\Assert;
+
+/**
+ * @internal
+ */
+final class PhpUnitTracer implements Tracer
```
theofidry: Actually the caching part could be extracted

src/TestFramework/Tracing/TraceProviderAdapterTracer.php
```diff
@@ -0,0 +1,138 @@
...
+     */
+    private ?Generator $traceGenerator = null;
+
+    /**
+     * @var array<string, Trace|null> Traces indexed by their source pathname.
+     */
+    private array $indexedTraces = [];
+
+    public function __construct(
+        private readonly TraceProvider $traceProvider,
```
theofidry, line 72: in the end the current implementation is not PHPUnit specific

src/TestFramework/PhpUnit/PhpUnitTracer.php
theofidry: I decided against it. A non-cached version is unusable, so whilst it would be technically feasible especially thanks to the interface, it would be tightly coupled to a `CachedTracer`. Rather than making this coupling hidden, I prefer to make it explicit.

theofidry:
The mutation escape is legit: it is not critical to outcome, only performances

devTools/phpstan.neon
```diff
@@ -146,6 +146,12 @@ parameters:
             message: '#::profile\(\) should return int but returns int.*\|null#'
             path: ../tests/benchmark/DummyInstrumentor.php
 
+        # This plugin does not seem to understand the PHPBench data providers
+        -
+            identifier: shipmonk.deadMethod
+            message: '#Unused .+Bench::provide*#'
+            path: ../tests/benchmark/*/*Bench.php
```
staabm, line 153: reported upstream in https://github.com/shipmonk-rnd/dead-code-detector/issues/377

PR #2708, opened 2025-12-14
feat: Introduce the basic plumbing to deal with `iterable`

theofidry:
## Background

_sight_, so where to start.

I've now dealt 3 times with various issues related to `iterable`. Fundamentally, the issue is, they **MAY**:

- be lazy
- have internal caching *1
- be rewindable *2
- be partially shared *3
- be mutable *4

*1: An implementation of `Iterator` maybe doing `$iterator->key()` will lookup the value, cache it, so that if you do `$iterator->current()` (without moving the cursor, without calling`::next()` or `::rewind()`), this will not trigger the lookup again. BUT, this is entirely at the discretion of the implementation, this is not guaranteed.

*2: A `foreach` construct and `iterable_to_array()` both trigger an initial `::rewind()`. So in practice, consuming a non-rewindable `iterable` as such is annoying. Also note that the traditional implementation of `NoRewindIterator` just _ignores_ `::rewind()`, but does not forbid it – I assume because that would make it impractical.

*3: Let's say you have:

```php
function take1(Iterator $iterator): mixed {
  $value = $iterator->current();
  $iterator->next();

  return $value;
}

$iterator1 = $service->get();
$value1 = take1($iterator1);

$iterable2 = $service->get();
$value2 = take1($iterable2);
```

In this case, a basic implementation would be that `$service` returns a fresh `Iterator`. In this case `$value1 === $value2`.

Now let's say the lookup of the values is very expensive. Then with the above, it means we will do that expensive lookup twice for the same value, not nice! Fair enough, we cache it. This is not going to be enough, we also need to _share_ the iterator, or at least the cache that is being used need to be shared. Cool! ...Except: now we are in a dilemma, when we do `$value2 = take1($iterable2)`, does it returned the cached `$value1`, or does it get the next value? How to distinguish the two?

> [!NOTE]
> In theory, our code is affected by this problem. In practice, it has not been an issue because every place where we needed `iterable`, it was consumed in _one_ place, so this "concurrency" usage has not manifested. Or maybe I missed a place in which case it could be the source of some edgy very hard to debug issues.

*4: A generator for instance, can have values injected at any time via `::send()`.

Most of those issues are inherent to having a flexible streamable type. This is not really a PHP issue, and frankly unless we enforce `array`, reducing `iterable` to `Iterator` or `Generator` is not going to solve the problem.

Now, those are not eclectic problems. As mentioned, I've tried to tackle this 3 times now already:

- in #2543 when working with a pausable cachable stream for the XML report
- in #2697 before deciding to switch to `array` for the `SourceCollector` because of *3.
- in #2707 because I simply have interest in having the behaviour tested and nailed down without humongous unreadable boilerplate.

## The solution

The extend of the solution will vary and evolve depending on our usage. I think all of those issues are very solvable, but we need some primitives. PHP gives us some tools already, but they are a bit... primitive or lacking in expressiveness.

In this PR I extracted what I needed for #2707 and I think it will be all (or at least 90%) of what I'll need for #2543. Maybe #2697 could be solved as well, but I think that requires more work and I don't think it's that simple so I prefer to leave this alone for now.

What this PR offers:

- `GeneratorFactory`: a utility to create a `Generator` from `iterable`. Boring API, but the implementation is a bit annoying and it was worth having some tests for it.
- Move `IterableCounter` next to `GeneratorFactory`.

In terms of source code, this is it. But on top of this, this PR introduces a number of utilities:

- `NonRewindableIterator`: decorates an `Iterator` but bails out on _any_ `::rewind()`.
- `TrackableIterator`: decorates an `Iterator` but allows to easily check the current status of the traverse (e.g. to know if we iterated over X items)
- `YieldOnceIterator`: decorates an `Iterator` to make so that `::key()` and `::current()` cannot be called more than once for a given position. This allows to simplify drastically a test where we want to ensure we have a caching of some sorts in place.

## Example of usage

Why are they necessary? Consider the following scenario:

- You have an `iterable` as a source.
- You get requested an item from the source by an identifier, this means you need to iterate of that source to get the item.
- Looking up the source is expensive, hence:
  1. You do not want to look up more than necessary, so you stop iterating over the source as soon as you found your item.
  2. You do not want to do that lookup again so you cache the item.
  3. You already looked up items on the way (e.g. if you were interested in item3, you went through item1 and 2). So you cache them too.

So essentially, this is having an `iterable` as a source and you are implementing a pausable lookup with cache on top of it. This little scenario is what I'm using for some XML reports in #2543 and the algorithm used for the Tracer in #2707.

Now, the implementation of it is actually is relatively straightforward (not super small, but not particularly big either):

```php
final class AlgorithmDemo
{
    private bool $traversed = false;

    private ?Generator $itemGenerator = null;

    /**
     * @var array<string, Item|null> Traces indexed by their source pathname.
     */
    private array $indexedItems = [];

    public function __construct(
        private readonly SourceProvider $sourceProvider,
    ) {
    }

    public function find(string $id): ?Item
    {
        return array_key_exists($id, $this->indexedItems)
            ? $this->indexedItems[$id]
            : $this->lookup($id);
    }

    private function lookup(string $id): ?Item
    {
        if ($this->traversed) {
            $this->indexedItems[$id] = null;

            return null;
        }

        $items = $this->getItemGenerator();

        // Do not use a foreach loop as it does a rewind which we do not want
        // to do.
        while ($items->valid()) {
            $item = $items->current();
            $items->next();

            $this->indexedItems[$item->id] = $item;

            if ($item->id === $id) {
                return $item;
            }

            if ($this->traversed) {
                break;
            }
        }

        $this->indexedItems[$id] = null;

        return null;
    }

    private function getItemGenerator(): Generator
    {
        if ($this->itemGenerator === null) {
            $this->itemGenerator = GeneratorFactory::fromIterable(
                $this->sourceProvider->getAll(), // iterable
            );
        }

        return $this->itemGenerator;
    }
}
```

As you can see, the only real magic here is `GeneratorFactory::fromIterable()`. Otherwise it is just making sure we got a single generator as a source and that we do not rewind it.

To test that however, oh boy, that's a different beast.

You can see the Claude attempt [here](https://github.com/infection/infection/pull/2707/changes/1a60bd261ae9dcf6c482a23351dc6f38b06e2a33#diff-68517967762cad06460aac17d92a520f27d5dbddc3343e6d09a4b4d2ab33f41d). Honestly, it does the job decently, but it's a lot of hard to read code IMO.

This is because there is a lot of edge cases to consider and a lot of piping is necessary. You need to be able to:

- mimic the different cases of `iterable`, that alone is annoying (and with this PR, we can just leverage `GeneratorFactory::fromIterable()` instead – handling the edge cases is its responsibility).
- spy on the generator to:
  - know how much we are progressing over time (i.e. not a "we expect X calls" classic mock, but live tracking of the number of calls made)
  - see if we are caching the results

This is possible with any extra tools, Claude did it. I also did it before, at least partially for #2543. It's not fun, and it is even less fun to review the tests and I suspect even less so to maintain them.

Now, with the utilities introduced above, we can achieve this with a lot more ease by composing iterators. See for yourself, [the diff (I recommend a split view)](https://github.com/infection/infection/pull/2707/changes/584cd195eba7d5c2d54d6617fc85f540b276b9ae#diff-68517967762cad06460aac17d92a520f27d5dbddc3343e6d09a4b4d2ab33f41d).

### DIFF
--- a/src/Framework/Iterable/GeneratorFactory.php
+++ b/src/Framework/Iterable/GeneratorFactory.php
@@ -41,6 +41,9 @@ use Iterator;
 use IteratorAggregate;
 
 /**
+ * The goal of this utility is to easily convert a generic `iterable` into a `Generator`,
+ * which has a much friendlier API.
+ *
  * @internal
  */
 final class GeneratorFactory
@@ -64,13 +67,19 @@ final class GeneratorFactory
 
                 $iterable->next();
             }
-        } elseif ($iterable instanceof IteratorAggregate) {
+
+            return;
+        }
+
+        if ($iterable instanceof IteratorAggregate) {
             yield from self::fromIterable($iterable->getIterator());
-        } else {
-            // For arrays, Generators, and other iterables, foreach is safe
-            foreach ($iterable as $key => $value) {
-                yield $key => $value;
-            }
+
+            return;
+        }
+
+        // For arrays, Generators, and other iterables, foreach is safe
+        foreach ($iterable as $key => $value) {
+            yield $key => $value;
         }
     }
 }
--- a/rector.php
+++ b/rector.php
@@ -125,6 +125,9 @@ return RectorConfig::configure()
             __DIR__ . '/tests/phpunit/FileSystem/FileSystemTestCase.php',
         ],
         AddInstanceofAssertForNullableInstanceRector::class,
+        AddParamArrayDocblockFromDataProviderRector::class => [
+            __DIR__ . '/tests/phpunit/Framework/Iterable/GeneratorFactory/GeneratorFactoryTest.php',
+        ],
         AddReturnTypeDeclarationBasedOnParentClassMethodRector::class => [
             __DIR__ . '/tests/phpunit/Fixtures/Console/FakeOutputSymfony5.php',
         ],
--- a/tests/phpunit/Framework/Iterable/GeneratorFactory/GeneratorFactoryTest.php
+++ b/tests/phpunit/Framework/Iterable/GeneratorFactory/GeneratorFactoryTest.php
@@ -50,14 +50,10 @@ use Traversable;
 final class GeneratorFactoryTest extends TestCase
 {
     /**
-     * @param mixed[] $iterable
-     * @param array<string, string> $expected
-     */
-    #[DataProvider('iterableProvider')]
-    /**
      * @param iterable<mixed> $iterable
      * @param mixed[] $expected
      */
+    #[DataProvider('iterableProvider')]
     public function test_it_can_generate_a_generator_from_an_iterable(
         iterable $iterable,
         array $expected,
--- a/tests/phpunit/Framework/Iterable/GeneratorFactory/SimpleIteratorAggregate.php
+++ b/tests/phpunit/Framework/Iterable/GeneratorFactory/SimpleIteratorAggregate.php
@@ -41,6 +41,7 @@ use Traversable;
 /**
  * @template TKey of array-key
  * @template TValue
+ *
  * @implements IteratorAggregate<TKey, TValue>
  */
 final readonly class SimpleIteratorAggregate implements IteratorAggregate
--- a/tests/phpunit/TestingUtility/Iterable/TrackableIterator.php
+++ b/tests/phpunit/TestingUtility/Iterable/TrackableIterator.php
@@ -40,27 +40,28 @@ use Iterator;
 /**
  * This iterator is a utility to enrich an existing by providing tracking capabilities.
  *
- * It allows to track the last yielded values, without making assumption about the
+ * It allows tracking the last yielded values without making an assumption about the
  * internal logic of the decorated iterator. This is achieved by tracking the yielded
  * values in separate properties.
  *
- * For example, a custom iterator could have a weird logic in which `::current()` can return
+ * For example, a custom iterator could have a unique logic in which `::current()` can return
  * the value only once or trigger various side effects. In which case, we cannot use `::current()`
- * to check what was the last yielded value. This class offers a separate property to achieve this.
+ * to check what the last yielded value was. This class offers a separate property to achieve this.
  *
  * @template TKey of array-key
  * @template TValue
+ *
  * @implements Iterator<TKey, TValue>
  */
 final class TrackableIterator implements Iterator
 {
     // Value representing the absence of a yielded key. We cannot use `null` as it is a result
     // that may be returned by `::key()`.
-    public const EMPTY_KEY = '__ππae5181162f0f0a5daacf223fee61d13d142e276807525867a836d3d6968854a0';
+    public const EMPTY_KEY = '__θθae5181162f0f0a5daacf223fee61d13d142e276807525867a836d3d6968854a0';
 
     // Value representing the absence of a yielded value. We cannot use `null` as it is a result
     // that may be returned by `::value()`.
-    public const EMPTY_VALUE = '__ππfb31e5bcd01897b407311d26f33b78be5b4604f9199fe6e72240c5aee1a2ee44';
+    public const EMPTY_VALUE = '__θθfb31e5bcd01897b407311d26f33b78be5b4604f9199fe6e72240c5aee1a2ee44';
 
     private bool $yieldedAnyValue = false;
 
--- a/tests/phpunit/TestingUtility/Iterable/YieldOnceIterator.php
+++ b/tests/phpunit/TestingUtility/Iterable/YieldOnceIterator.php
@@ -42,7 +42,7 @@ use Iterator;
  * This iterator is a utility to enrich an existing Iterator.
  *
  * It decorates an iterator and ensures the keys and values can be
- * yield once and only once, unless the iterator is rewind.
+ *  yielded once and only once, unless the iterator is rewind.
  *
  * This aims to mimic an iterable for which the fetching item
  * operation is heavy, hence should not be done more than once, but
@@ -50,6 +50,7 @@ use Iterator;
  *
  * @template TKey of array-key
  * @template TValue
+ *
  * @implements Iterator<TKey, TValue>
  */
 final class YieldOnceIterator implements Iterator


tests/phpunit/TestingUtility/PHPUnit/ThrowableAssertions.php
```diff
@@ -0,0 +1,70 @@
...
+use PHPUnit\Framework\TestCase;
+use Throwable;
+
+/**
+ * Small utility for a neater try/catch. This is useful when we cannot use `TestCase::expectException*()`
+ * utilities or when more assertions than the ones provided need to be done against the throwable.
+ *
+ * @require-extends TestCase
+ */
+trait ThrowableAssertions
```
theofidry: this is not a good name for a trait, to change

theofidry:
@sanmai designating you for review as are the iterable lover/master 😄

src/Framework/Iterable/GeneratorFactory.php
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public static function fromIterable(iterable $iterable): Generator
+    {
+        if ($iterable instanceof Iterator) {
+            // For non-Generator Iterators, manually iterate to avoid calling `::rewind()`.
+            while ($iterable->valid()) {
+                yield $iterable->key() => $iterable->current();
+
+                $iterable->next();
+            }
```
sanmai, line 69: What stops us from using `NoRewindIterator` here?

https://www.php.net/manual/en/class.norewinditerator.php

> This iterator ignores rewind operations. This allows processing an iterator in multiple partial foreach loops.

sanmai:
In my own pipeline work, I'm clearly biased: lazy operations exist only when eager array-based code becomes impossible, overly complex, or fragile. I considered adding iterators with caching and lazy-evaluation, but never found a concrete benchmark that justified the extra complexity (batching or plain arrays were always sufficient). I'd be interested in seeing benchmarks that demonstrate a clear win here.

src/Process/Runner/MutationTestingRunner.php
```diff
@@ -41,7 +41,7 @@
 use Infection\Event\MutantProcessWasFinished;
 use Infection\Event\MutationTestingWasFinished;
 use Infection\Event\MutationTestingWasStarted;
-use Infection\IterableCounter;
+use Infection\Framework\Iterable\IterableCounter;
```
sanmai, line 44: This change sounds like a good separate refactor PR.

src/Framework/Iterable/GeneratorFactory.php
theofidry: Instead of a generator you mean? 

Because whilst it won't fail with a "generator closed error", once exhausted, it simply yields an empty result: https://3v4l.org/looqc#vnull

src/Process/Runner/MutationTestingRunner.php
theofidry: I could do it after indeed, it's only moving that iterable utility next to the new one

theofidry:
>What is the cost of reading the whole dataset all at once into a plain old array?
>How much do you need to amortize this cost to justify the extra complexity?

Good questions :) And I think it depends.

One of the cases I encountered was for the `SourceCollector`. Whilst maybe there's a way to make it work there, I do not thing collecting the sources is a significant hog in an infection process. The source files were [before introducing `SourceCollector`] loaded eagerly, and whilst I liked the idea of making it lazy, I do not think the technical complexity involved justified it. Well, in fairness, I did not benchmark the source collection in an entire infection process, so maybe I'm wrong and it's a missed opportunity, but at the very least I was not ready to tackle this challenge.

The other two cases are:

1. the coverage report
2. the traces
3. the mutations

1) and 2) are closely related, but those are two distinct problems. What about there? It depends. Currently, despite using iterables, we are loading the entire documents in-memory, this can be several dozens of MBs of XML. IMO it's not pretty, and nothing unscallable, but I admit I don't like it. That said my immediate concern there is to address the coverage report API more than the performances. For the performances there is various axis of improvements, the parsing algorithm is only one of the few. TL:DR; I could live without this in 1).

In 2)... Well we have benchmarks. The current status is using `iterable`, I could adapt the implementation of the tracer in https://github.com/infection/infection/pull/2707 to check how much worse it would be by loading all the traces upfront in an array.

3): I don't think this is a scalable solution. Since we do not mutate the generators and we have only one subscriber, the problems highlighted _happened_ to not be a problem (I think at least – maybe we have subtle bug(s) there that we didn't notice). But I'm not touching that part yet so I'm not too concerned about that yet. However, **I don't think we can skip using non-array iterables of some sorts here.**

>And if you really want to make a dent, [fibers](https://www.php.net/manual/en/language.fibers.php) could be a better fit here

Maybe I'm missing the bigger picture here, but I admit I struggle to see how fibers would help. There provide a more controllable generator, but this does not really solve the problem of multiple subscribers & caching. A better higher level solution would rather be observables IMO, but I'd deffer the decision of using this until much further in my refactoring process.

In a more near-future, I was more imagining a different return type than `iterable` that would better convey the type of encapsulated value instead of "whatever" (which is what `iterable` is, what the underlying data is and how it behaves is just extremely diverse). But I'm not set on that either.

More immediate term, I just want better primitives to ensure our code using `iterable` works as intended, which is what this PR is about.

<hr/>

>[!IMPORTANT]
>**I'll benchmark the Tracer using an array instead, if the results are not terrible, we can go with arrays and defer this PR until the next usage comes up.**

sanmai:
So, how does it look like on benchmarks?

How does one benchmark this?

theofidry:
>So, how does it look like on benchmarks?

I was checking this out. I've rebased #2707.

>How does one benchmark this?

That is a good question... I kinda forgot on the moment, but our current benchmark is useless for this. Indeed in the current benchmark we just go through all the sources. But the strength of the algorithm, is to be able to stop at the beginning or half way if that is sufficient, if you traverse the whole source iterable, there is no difference with a regular foreach on an array...

So I'd need to sketch out a different scenario, but the order of the source files and the declaration of the tests in the coverage report matters... So this is tricky :/

theofidry:
time for me to go to bed, I'll sleep over it!

theofidry:
@sanmai the results are here. But I feel like I need to expend a bit more on the rationale.

I explained the inversion of logic in #2707, so I won't repeat this part. I'll focus more on the technical aspect of the flow.

First thing to clarify, #2707 is a minimal implementation. The `TraceProvider` is meant to go, i.e. further down the line, with #2543, I mean to have an easier to consume API for the coverage and the tracer will directly consume that coverage API to compute the traces. This is important to keep in mind in the following explanation.

We are doing a source file -> test lookup, and once we located the test we compute a trace. This means we have two axis for performance:

- the test lookup. Currently we parse the entire coverage, so the first lookup is expensive, but the remainder as pretty much free.
- the trace computation: we aim to make it free by lazily evaluating their values.

This means in the current configuration, with the trace provider, the benefits are _limited_.

Indeed, the practical case is: you are in a diff mode, so you want to get the traces of a handful of source files instead of the entirety of them. Ideally: you want to do just the necessary work, i.e. grab the corresponding tests and compute the traces for them.

If you compute the entire coverage, then there is little room for performance improvements in this scenario. **Currently we do not compute 100% of the coverage upfront, but more than necessary still.** So on that front we are halfway.

However, if you make the parsing of the coverage smarter, i.e. you parse only what's necessary, then you _may_ greatly improve things in this scenario. It is still a big question mark, not because I doubt it will, but because there is a great variance. If the test you declare is at the very end of the report, then you're unlucky and you needed to parse the entire coverage to get it even if that was the only test needed.

So in a sense for sure it is a lot of very conditional wins.

<hr/>

The benchmark I took is a very simplistic scenario that showcase the possible wins in the happy path. I parameterized the benchmark with a percentage (`0. < $percentage < 1.`) which picks the first % of the source files we have, and we do the tracing for them.

This is awfully skewed toward the happy path, again this doesn't mean this can't happen, it's just that it's unpredictable hence you cannot _expect_ this degree of benefit ahead of time. You get lucky or you're not.

With this in mind, the result for the array is:

```
+--------------+---------+-----+------+-----+----------+--------+--------+
| benchmark    | subject | set | revs | its | mem_peak | mode   | rstdev |
+--------------+---------+-----+------+-----+----------+--------+--------+
| TracingBench | bench   | 0   | 1    | 5   | 68.902mb | 5.371s | ±2.87% |
| TracingBench | bench   | 1   | 1    | 5   | 68.902mb | 5.527s | ±3.98% |
| TracingBench | bench   | 2   | 1    | 5   | 68.903mb | 5.537s | ±5.97% |
| TracingBench | bench   | 3   | 1    | 5   | 68.909mb | 5.769s | ±1.85% |
| TracingBench | bench   | 4   | 1    | 5   | 68.902mb | 5.582s | ±0.76% |
+--------------+---------+-----+------+-----+----------+--------+--------+
```

each bench is respectively with 10%, 25%, 50%, 75% and 100% of the sources. Unsurprisingly, there is no change for each of them because we compute them all regardless.

With the iterator approach:

```
+--------------+---------+-----+------+-----+----------+-----------+--------+
| benchmark    | subject | set | revs | its | mem_peak | mode      | rstdev |
+--------------+---------+-----+------+-----+----------+-----------+--------+
| TracingBench | bench   | 0   | 1    | 5   | 68.456mb | 632.814ms | ±6.45% |
| TracingBench | bench   | 1   | 1    | 5   | 68.456mb | 1.328s    | ±4.97% |
| TracingBench | bench   | 2   | 1    | 5   | 68.456mb | 2.662s    | ±5.12% |
| TracingBench | bench   | 3   | 1    | 5   | 68.456mb | 3.761s    | ±2.45% |
| TracingBench | bench   | 4   | 1    | 5   | 68.830mb | 5.338s    | ±2.04% |
+--------------+---------+-----+------+-----+----------+-----------+--------+
```

<hr/>

But I think we got a bit lost. The question is: do we want to go with `iterable` for the trace generation, which is what the current implementation of #2707 respects, or if we do not want in which case we would change `TraceProvider` to return an array instead (otherwise nothing changes for #2707 – you get an iterable as an input, the complexity lies from there).

I _really_ dislike the current status regarding our usage of `iterable`. Even for this stupid benchmark I got bitten once more because if you call the trace provider twice the 2nd one will not fail and just yield no result. So when you mention "iterable burnout" I am totally with you, it's tricky and annoying as f***.

However, I also do not think it is reasonable to do all the computation upfront. It is sacrificing too much. Hence this PR. Yes it's more code, but IMO it gives better tools to deal with `iterable` and bring back some sanity.

src/Framework/Iterable/GeneratorFactory.php
sanmai: It would be great to add a comment about this, because it is not obvious :thinking:

src/Framework/Iterable/GeneratorFactory.php
```diff
@@ -0,0 +1,76 @@
...
+        } elseif ($iterable instanceof IteratorAggregate) {
```
sanmai:
```suggestion
              return;
        } 
        
        if ($iterable instanceof IteratorAggregate) {
```

src/Framework/Iterable/GeneratorFactory.php
```diff
@@ -0,0 +1,76 @@
...
+            yield from self::fromIterable($iterable->getIterator());
+        } else {
```
sanmai:
```suggestion
             return;
        }
```

sanmai:
How does it play out in the whole project? How do you plan to use it?

Just to be clear, apart from a few nitpicks, the code makes sense, so I'm not here to hold you back; instead, I'm following up on your review request as someone who has had enough of iterators in PHP. I still use them, but very deliberately, only when it makes sense or simplifies the code.

src/Process/Runner/MutationTestingRunner.php
theofidry: Done via #2711

theofidry:
>How does it play out in the whole project? How do you plan to use it?

Short term my goal is to bring back some sanity if you need to consume an `iterable`.

The `GeneratorFactory` is there to transform it back into a nicer to use API. I defaulted to `Generator`, but I imagine we could use a different kind of `Iterable` or even a `SpecialIterator|array` or something in the future. I picked `Generator` fitted the bill for now, but it's not set in stone either.

And test-wise the new utilities already make testing a lot easier (see the tracer for #2707).

>Just to be clear, apart from a few nitpicks, the code makes sense, so I'm not here to hold you back; instead, I'm following up on your review request as someone who has had enough of iterators in PHP. I still use them, but very deliberately, only when it makes sense or simplifies the code.

I'm with you, streams and pipes are not easy, and iterable is akin to forcing streams without an API. I dare hope we'll have something better than `iterable` in the future without inventing an entire framework.

In any case, it's a painful journey so happy to have someone else to ramble with 😄

tests/phpunit/TestingUtility/Iterable/NonRewindableIterator.php
```diff
@@ -0,0 +1,87 @@
...
+     * @param Iterator<TKey, TValue> $iterator
+     */
+    public function __construct(
+        private Iterator $iterator,
+    ) {
+    }
+
+    public function rewind(): void
+    {
+        throw new DomainException('Cannot be rewind.');
```
sanmai, line 65: There's a hidden issue here. Some iterators such as `IteratorIterator` need `rewind()` to initialize. So if we skip that call entirely,  `IteratorIterator` may not work at all.

I learned about this just today while working [on this `cursor()` feature](https://github.com/sanmai/pipeline?tab=readme-ov-file#pipeline-cursor).

theofidry: Yes. I don't think it's an issue though, at least in the context of that testing utility. I hesitated to adapt the code to make it more usable, but there's no use case yet

PR #2709, opened 2025-12-14
refactor: Move Trace under a Tracing namespace

theofidry:
Currently we got `Trace` under `TestFramework\Coverage`. Since the first form of trace directly from coverage data, it made sense at that time.

In theory though: a trace is about tests in general, the coverage is what we use internally to create the traces, but it s a detail, the trace is otherwise not coupled to the coverage.

We've also had more trace-related value objects and trace implementations added over time. #2707 will also introduce a `Tracer` interface.

This PR proposes to re-organise things a bit and move the traces and their value-object related objects to `TestFramework\Tracing\Trace`. The current implementation still have a few issues because we have dependencies from `ProxyTrace` to `TestFramework\Coverage`, but this need further changes to be addressed, hence that should be done later.

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -137,7 +137,7 @@ return RectorConfig::configure()
         LocallyCalledStaticMethodToNonStaticRector::class,
         ReadOnlyPropertyRector::class => [
             // property can't be readonly as it's returned by reference and may be updated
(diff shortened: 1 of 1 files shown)


maks-rafalko (approved):

theofidry:
'ty!

PR #2711, opened 2025-12-16
refactor: Move IterableCounter to a dedicated namespace

theofidry:
In #2708 and likely upcoming PRs, we will have more `iterable` related utility, so I moving this to a dedicated namespace for better organisation.


theofidry:
This has been discussed and validated in https://github.com/infection/infection/pull/2708, but extracted out of it to reduce the noise of that PR.

PR #2624, opened 2025-11-30
test: Rework the PHPUnit end-to-end tests

theofidry:
## Backgorund

For #2543 I need to touch a lot of the coverage part and one thing I am missing is getting hands on a good enough sample of coverage for various test frameworks and different versions of those.

We've had a mix of fixtures:

- I introduced some for the tracing test in `/Users/tfidry/Project/Humbug/infection/tests/phpunit/TestFramework/Tracing/Fixtures` in ##2444.
- We have some comprehensive ones with a bit more information about what wee get from the coverage data with `/Users/tfidry/Project/Humbug/infection/tests/phpunit/Fixtures/TestFramework/PhpUnit/Coverage/*` (see #957)

I'm however not too happy with the current state of things: it's hard to find the right balance to maintain them.

## Solution

The idea of this PR is to leverage the end-to-end tests. We have e2e tests for the various test frameworks with their different setup/versions. What we can do is to actually enrich them a bit to be able to generate coverage from them, when needed.

We don't really need to go out of our way to automate it: getting the coverage for PHPUnit 9.3 for instance doesn't need to be done more. The day we need more out of it, e.g. we want to see how a trait is described in the XML coverage or something so we need to add a new class/tests. Then we can pick up the generated coverage and place it where needed.

## Implementation

This PR updates the code to offer a better range of covered and non-covered code as well as a `Makefile` command to make the test generation easy.

I plan to touch the PhpSpec & Codeception in a similar fashion too.

PR #2730, opened 2025-12-19
test(e2e): Move the artefacts of the test for some e2e tests

theofidry:
I want to introduce some improvements in #2624, hence I focused on the tests I'm working on and not all the e2e tests.

What I propose is to move the temporary artifact to a `var` directory to allow to keep the test a bit cleaner even when executed locally.

This is loosely related to https://github.com/infection/infection/issues/2498.

### DIFF
--- a/tests/phpunit/Console/E2ETest.php
+++ b/tests/phpunit/Console/E2ETest.php
@@ -208,7 +208,18 @@ final class E2ETest extends TestCase
         $expected = file_get_contents('expected-output.txt');
         $expected = Str::toSystemLineEndings($expected);
 
-        $this->assertStringEqualsFile('infection.log', $expected, sprintf('%s/expected-output.txt is not same as infection.log (if that is OK, run GOLDEN=1 vendor/bin/phpunit)', getcwd()));
+        $expectedFile = file_exists('var/infection.log')
+            ? 'var/infection.log'
+            : 'infection.log';
+
+        $this->assertStringEqualsFile(
+            $expectedFile,
+            $expected,
+            sprintf(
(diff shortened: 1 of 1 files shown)


tests/e2e/standard_script.bash
```diff
@@ -11,4 +11,8 @@ else
     php $INFECTION
 fi
 
-diff -w expected-output.txt infection.log
```
theofidry: the change of option has been extracted to #2729.

tests/e2e/standard_script.bash
```diff
@@ -11,4 +11,8 @@ else
...
+if [ -f "var/infection.log" ]; then
+    diff --ignore-all-space expected-output.txt var/infection.log
```
theofidry, line 15: This change deserves to be mentioned in the docs once #2728 is merged

maks-rafalko:
> What I propose is to move the temporary artifact to a `var` directory to allow to keep the test a bit cleaner even when executed locally.

I don't follow why do we need this? What was the reason for this change?

theofidry:
@maks-rafalko a couple of cases:

- PHPUnit releases a cache directory setting: by default it adds it to the root so if it's not part of `.gitignore`, then it's committed.
- I need to add PHPStan and I want its cache to be local to that project: same problem.
- I need to generate coverage for myself, but I don't really want to commit it neither pollute the gitignore for something I used locally for my own reasons.

So in general there's plenty of situations where you may end up with artifacts of all kinds. I think it's cleaner to just keep them in a `var` directory. It pollutes less the root, you don't need to constantly update the `.gitignore` and it's also easier to clean up.

maks-rafalko:
> I need to add PHPStan and I want its cache to be local to that project: same problem.

is it something you need in the future? I didn't notice it in this PR

theofidry:
Yes! For #2624

PR #2743, opened 2025-12-23
fix: Make the computation of the JUnitTestExecutionInfoAdder lazy

theofidry:
I didn't check but I think I broke it in #2738... In any case it is fixed & tested.


sanmai (commented):
Thanks!

Issue #2758, opened 2025-12-26
[0.32.0] excludes configuration ignored - excluded paths still being mutated


koraga:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.32.0
| Test Framework version | PHPUnit 12.5.4
| PHP version | 8.5.1
| Platform    | MacOS (Docker php-cli-alpine)
| Github Repo | -

## Description

After upgrading from Infection 0.31.9 to 0.32.0, the `excludes` configuration in `infection.json5` is being ignored. Files from excluded directories are still being mutated.

## Expected Behavior

Files in the following excluded directories should NOT be mutated:
- `bootstrap/cache`
- `app/Infrastructure/OpenApi/Scramble`
- `app/Infrastructure/Router`
- `app/Infrastructure/Eloquent`
- `app/Infrastructure/CLI`

## Actual Behavior

Infection 0.32.0 generates mutations for files in excluded directories. Examples from output:
```
1) /var/www/html/app/Infrastructure/CLI/CommandFinder.php:23
2) /var/www/html/app/Infrastructure/CLI/CommandFinder.php:44
...
11) /var/www/html/app/Infrastructure/Eloquent/EloquentServiceProvider.php:14
12) /var/www/html/app/Infrastructure/Eloquent/Models/Model.php:33
...
18) /var/www/html/app/Infrastructure/OpenApi/Scramble/DocumentTransformers/AddBearerSecurityScheme.php:18
```

All these paths are explicitly listed in `excludes` configuration but are still being processed.

## Additional Context

- This worked correctly in version **0.31.9**
- Downgrading to 0.31.9 resolves the issue
- The same configuration file works on 0.31.9 but fails on 0.32.0

<details>
 <summary>infection.json5</summary>

 ```json5
{
    "$schema": "vendor/infection/infection/resources/schema.json",
    "source": {
        "directories": [
            "app",
            "config",
            "bootstrap"
        ],
        "excludes": [
            "bootstrap/cache",
            "app/Infrastructure/OpenApi/Scramble",
            "app/Infrastructure/Router",
            "app/Infrastructure/Eloquent",
            "app/Infrastructure/CLI"
        ]
    },
    "threads": "max",
    "tmpDir": "var/infection",
    "timeout": 30,
    "minCoveredMsi": 90,
    "minMsi": 90,
    "mutators": {
        "@default": true
    }
}
 ```
</details>

<details>
 <summary>phpunit.xml</summary>

 ```xml
[...]

PR #2759, opened 2025-12-26
Make sure exclusion by folder works as expected by e2e test

maks-rafalko:
Covers https://github.com/infection/infection/issues/2758

### DIFF
--- a/tests/e2e/Exclude_By_Folder/expected-output.txt
+++ b/tests/e2e/Exclude_By_Folder/expected-output.txt
@@ -1,0 +1,11 @@
+Total: 1
+
+Killed by Test Framework: 1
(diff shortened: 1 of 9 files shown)


tests/e2e/Exclude_By_Folder/expected-output.txt
```diff
@@ -0,0 +1,11 @@
+Total: 1
```
theofidry, line 1: more of an FYI than anything: you could now use the `config:list-sources` command instead to check that the desired files are gonna be used

maks-rafalko: yeah, saw it, thanks. I think it can be useful for users

PR #2762, opened 2025-12-31
feat(command): Introduce a debug command to execute the initial tests

theofidry:
Sometimes getting the initial config of the test framework maybe a bit tricky and you do not want to run the whole infection to figure out why it's failing, and would like to have more information on what infection does that may be different than locally.

This PR introduces the command `initial-test:run` which supports all the options that can otherwise be used with the regular infection command. It has two distinct differences:

- It logs the test framework option used, regardless of the verbosity/debug.
- It automatically enables the debug command, as such executing it will create temporary files that are not deleted!

I am not happy with the output to be honest, but I'm not happy about it neither in the main execution process. I would also like to provide more information about the test framework used at some point:

- version detected
- executable used
- config file used for the initial tests

But there is no easy way to get those at the moment and may require further changes, hence I decided to keep it out of the scope of this PR.

## Reviewer notes

Note that there is no test for the command... Besides an end-to-end test, I do not see a way to implement one, do you think I should add one?

### DIFF
--- a/tests/phpunit/Command/InitialTest/InitialTestRunCommandTest.php
+++ b/tests/phpunit/Command/InitialTest/InitialTestRunCommandTest.php
@@ -40,6 +40,7 @@ use Infection\Command\InitialTest\InitialTestRunCommand;
 use Infection\Console\Application;
 use Infection\Container\Container;
 use Infection\Framework\Str;
+use Infection\Git\Git;
 use Infection\Process\Runner\InitialTestsFailed;
 use Infection\Process\Runner\InitialTestsRunner;
 use PHPUnit\Framework\Attributes\CoversClass;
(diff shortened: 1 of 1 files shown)


theofidry:
@infection/core ping for thoughts/feedback

maks-rafalko (approved):
would be useful for me as well. One thing that I personally needed many times is generated XML files during initial tests run.

Since Infection removes them during a common execution, I always updated `infection.json5`'s `tmpDir` to set it to `.` and executed Infection with `--debug` - this option doesn't remove temp files.

So if this new command will somehow help to avoid all of the above - it would be awesome.

theofidry:
Good point, I really dislike running it, taking a while to fail, see that I missed the verbosity, copy the command, see that I forgot the debug... 

I'm not sure this can be done out of the box yet, but I'll have a look

src/Command/Option/DebugOption.php
```diff
@@ -0,0 +1,73 @@
...
+     * @template T of Command
+     */
+    public static function addOption(
+        Command $command,
+        bool $default = Container::DEFAULT_DEBUG,
+    ): Command {
+        return $command->addOption(
+            self::NAME,
+            null,
+            InputOption::VALUE_NEGATABLE,
```
theofidry, line 63: to extract this change as two things are done here:

- Extracting the `--debug` option as a re-usable option.
- Changing `InputOption::VALUE_NONE` -> `InputOption::VALUE_NEGATABLE`.

This changes `--debug` to `--debug|--no-debug`. This was we can re-use it in the `InitialTestRunCommand ` and simply change the default value, but we would allow the possibility of the user to execute it without debug still.

src/Console/Application.php
```diff
@@ -105,8 +106,6 @@ public function getHelp(): string
 
     protected function getDefaultCommands(): array
     {
-        Container::create()->getGit();
```
theofidry: to extract; unrelated

src/Container/Container.php
```diff
@@ -901,6 +901,11 @@ public function getInitialStaticAnalysisProcessFactory(): InitialStaticAnalysisP
         return $this->get(InitialStaticAnalysisProcessFactory::class);
     }
 
+    public function getInitialTestsRunProcessFactory(): InitialTestsRunProcessFactory
```
theofidry, line 898: doesn't look to be necessary

theofidry:
Ready for a first round of review @infection/core

src/Command/Option/DebugOption.php
theofidry: https://github.com/infection/infection/pull/3004

theofidry: https://github.com/infection/infection/pull/3005

theofidry:
@maks-rafalko I could add very basic tests, but there is just too many paths for that command (with all the options).

PR #2763, opened 2026-01-01
refactor: Fix the Factory filtered source PHPStan warnings

theofidry:
`Factory::getFilteredSourceFilesToMutate()` was documenting returning a `list<SplFileInfo>` whilst in practice we were returning a `SplFileInfo[]`.

In general, I'm personally in favour of stricter types, so I am bias in favour of `list<SplFileInfo>`. However, this is yet another operation to do on top of all the files collected, which is why I made `SourceCollector::collect()` return a `SplFileInfo[]` instead of a `list<SplFileInfo>`.

I investigated if we really needed a list, and it turns out we do not. This is a good news because it means we do not have a hidden bug and that we can safely change `list<SplFileInfo>` to `SplFileInfo[]`. This is what this PR does.

### DIFF
--- a/src/TestFramework/Factory.php
+++ b/src/TestFramework/Factory.php
@@ -45,7 +45,6 @@ use Infection\TestFramework\Config\TestFrameworkConfigLocatorInterface;
 use Infection\TestFramework\PhpUnit\Adapter\PhpUnitAdapterFactory;
 use InvalidArgumentException;
 use function is_a;
-use function iterator_to_array;
 use SplFileInfo;
 use function sprintf;
 use Webmozart\Assert\Assert;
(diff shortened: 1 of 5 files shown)


src/TestFramework/Factory.php
```diff
@@ -132,25 +131,12 @@ public function create(string $adapterName, bool $skipCoverage): TestFrameworkAd
...
-            return [];
-        }
-
-        /**
-         * TODO: fix this warning, it is legit...
-         * @var list<SplFileInfo> $files
-         * @psalm-suppress InvalidArgument
-         * @phpstan-ignore varTag.type
-         */
-        $files = iterator_to_array(
```
theofidry, line 150: technically there is a minor unrelated refactor here...:

- `iterator_to_array()` is no longer necessary as `SourceCollector::collect()` returns an `array`
- the if condition is inversed
-  we have a one time usage variable which does not bring any clarity to the code 

Hence with the whole docblock gone with the errors, I figured we can rewrite it in this clearer form.

PR #2764, opened 2026-01-01
fix: Do not depend on the SourceCollector to know if the sources are filtered

theofidry:
The complete justification is a long one, the short version is: we do not need to know _exactly_ if the sources are filtered. I.e. there is no need to distinguish the case where the user applied a filter, but it did not result in anything to be filtered out, and something got filtered out. Only the fact that the user applied a some sort of filter is sufficient.

As such, we can get rid of whole bunch of code by removing `SourceCollector::isFiltered()` and use `Configuration#sourceFilter === null` instead.

<hr/>

Now the real thing that lead me there. `MutationGenerator` picks up the collected sources, iterates over it to hand it to `FileMutationGenerator` which then, if there is no tests for the trace associated to that source, skips it (unless we configured `--with-uncovered`).

Thanks to the `SourceCollector` API, actually, nothing would prevent us to do that filtering ahead of time. We can trivially implement another `SourceCollector` decorator that uses the `Tracer` and applies a filter... Can you?

Turns out we can't. The reason is simple, but not obvious. To get to `$tracer->trace($source)->hasTests()`, you need to execute the `Tracer`, which will need `IndexXmlCoverageParser`, which needs... `SourceCollector::isFiltered()`! So you end up in the situation where you would need the source collector to create the source collector, and laziness cannot help here (unless we're willing to make `IndexXmlCoverageParser#isSourceFiltered` lazy, but I don't think it really makes sense).

I don't like that situation. We could do without this of course, but it makes me think there is a problem at the model level, which bothers me a lot more than the actual technical problem. So I investigated why we need to know if the source is filtered, and there is two cases:

- `IndexXmlCoverageParser`: to throw a more understandable exception message upon failure. Here what matters is the _intent_: did the user try to filter the source at all? That the filter had an effect or not is irrelevant.
- `Factory::getFilteredSourceFilesToMutate()`: remove extra sources from the initial test run. Here it is more subtle. We _could_ benefit from knowing if the filter had an effect at all. I.e. if the user used a filter but had no effect, we could treat it as if no filter was applied. But looking at the code removed, independently from the original problem, I don't think the cost justifies it considering the case where there can actually be a difference is such a tiny edge case.


src/TestFramework/Factory.php
```diff
@@ -130,27 +129,16 @@ public function create(string $adapterName, bool $skipCoverage): TestFrameworkAd
...
+     * Get only those source files that will be mutated. If the source is filtered by the user,
+     * we do not need to execute the initial test run against all the sources, only the necessary
+     * subset.
      *
-     * @return list<SplFileInfo>
+     * @return SplFileInfo[]
      */
     private function getFilteredSourceFilesToMutate(): array
     {
-        // TODO: this looks incorrect to me... But it is what it is right now.
```
maks-rafalko: regarding this TODO - the code works correctly as for me. Just to double check the understanding: 

- if user uses some filter (`--filter=file1.php,file2.php`) - we return the list of such files and [update initial tests run](https://github.com/infection/infection/blob/533f119e963b86c2bf342f38f0e83ede8b2ec266/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php#L267) `phpunit.xml` to look only into these files for performance 
- if user doesn't use a filter - we return empty array, meaning no adjustments are needed to initial run `phpunit.xml` (checked [here](https://github.com/infection/infection/blob/533f119e963b86c2bf342f38f0e83ede8b2ec266/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php#L245))

theofidry: Indeed, it wasn't quite clear to me when I saw the code the first time. Checking the original code it made more sense, and I hope it is more obvious now as well, both from the form of the code and the documentation.

PR #2765, opened 2026-01-01
refactor: Exclude the container from Infection

theofidry:
I am a bit tired of false-positives like reported in https://github.com/infection/infection/pull/2764:

```
Warning: Escaped Mutant for Mutator "NotIdentical":

@@ @@
     {
         $container = new self([
             IndexXmlCoverageParser::class => static fn (self $container): IndexXmlCoverageParser => new IndexXmlCoverageParser(
-                isSourceFiltered: $container->getConfiguration()->sourceFilter !== null,
+                isSourceFiltered: $container->getConfiguration()->sourceFilter === null,
             ),
             Tracer::class => static fn (self $container) => new TraceProviderAdapterTracer(
                 $container->getTraceProvider(),
```

Technically it is correct, but covering this would require expensive integration tests, and in general we do not test the wiring of the services directly.

### DIFF
--- a/infection.json5
+++ b/infection.json5
@@ -7,6 +7,7 @@
             "src"
         ],
         excludes: [
+            "/^Container.php$/",
             "FileSystem/DummyFileSystem.php",
             "FileSystem/DummySymfony5FileSystem.php",
             "FileSystem/DummySymfony6FileSystem.php",


infection.json5
```diff
@@ -7,6 +7,7 @@
             "src"
         ],
         excludes: [
+            "/^Container.php$/",
```
theofidry, line 10: I don't like it tbh, but checking the result with `bin/infection config:list | grep Container`, we originally had:

```
src/Container.php
src/Process/Factory/MutantProcessContainerFactory.php
src/Process/MutantProcessContainer.php
src/Process/Runner/IndexedMutantProcessContainer.php
src/Testing/SingletonContainer.php
```

And just having `Container.php` resulted in:

```
src/Process/Factory/MutantProcessContainerFactory.php
```

So we were excluding more than desired.

theofidry: This is a known issue though, see https://github.com/infection/infection/issues/2758

maks-rafalko: didn't get how this is related to #2758, could you please clarify?

> And just having `Container.php` resulted in:
> 
> ```
> src/Process/Factory/MutantProcessContainerFactory.php
> ```

this is weird :|

theofidry: It's loosely related in the sense this has to do with how you configure infection as the user, with certain expectations, and how differently it behaves underneath because of how the Finder is configured/works.

Yes... I must say I just never had a good experience with the Symfony Finder. It's does the job but for every project, despite my best efforts (see the amount of tests for it in Box for example), I've always run into this sort of issues.

I hope I get to find a nicer solution or fix it at some point, but for now it is what it is.

theofidry:
Technically yes.

The current situation is that we do not test the container wiring at all. Even the integration test that we have that could test it, are not marked a covering the container.

So either we challenge the status quo and say this is something that must be addressed, or we keep it as is.

Personally I don't find testing the wiring of the container very interesting and I think it will only to expensive tests (annoying to write, hard to read, slow) for little value. So I propose to exclude it from the mutated files to not have failing builds for files we don't plan to cover.

Maybe someone feels strongly about the opposite, in which case I would encourage them to step up and fix it ASAP rather than letting the problem pester us any longer.

<hr/>

Actually thinking about it, it seems weird there is anything covering it at all in the first place. I will have a look into that.

infection.json5
sanmai: Symfony Finder is amazing when it works, but such a headache when it doesn't.

sanmai (commented):
I'd rather we nip it in the bud as in #2782; non-blocking.

theofidry:
@sanmai I opened https://github.com/infection/infection/pull/2784 as a replacement.

PR #2766, opened 2026-01-01
ci: Pin the GitHub Actions dependencies

theofidry:
GitHub Actions tags & branches are not safe to use. See PHPCSStandards/PHPCSDevTools#197.

There is also a repository setting to enforce pinned dependencies, but we cannot use it yet due to some deps not using pin dependencies themselves (notably https://github.com/ramsey/composer-install/issues/275).

Note that dependabot supports updating those just fine. So aside from the ugliness of the thing, it does not cause any issues.

### DIFF
--- a/.github/workflows/autoreview.yaml
+++ b/.github/workflows/autoreview.yaml
@@ -28,10 +28,10 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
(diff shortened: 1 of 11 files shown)


theofidry:
@sanmai dependabot, it does it just fine (updates hash + comment)

maks-rafalko:
> @sanmai dependabot, it does it just fine (updates hash + comment)

but it's not configured (just in case)

theofidry:
>but it's not configured (just in case)

I... shall fix that

PR #2767, opened 2026-01-01
ci: Add dependabot to update the GitHub Actions actions

theofidry:
This is super basic.

We could further configure it to update some other deps, but currently we have Conductor for PHP.


maks-rafalko (approved):
👍

Issue #2770, opened 2026-01-02
0 mutations were generated


knallcharge:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.32.0
| Test Framework version | PHPUnit 12.5.4
| PHP version | 8.5.0
| Platform    | Windows

After upgrading vom 0.31.9 to 0.32.0 mutations are no longer generated.
I tried generating a fresh infection.json5 (with only "source" and "mutators: default") but the result is the same.

<details>
 <summary>infection.json5</summary>

 ```json5
{
  "$schema": "vendor/infection/infection/resources/schema.json",
  "tmpDir": "build/infection",
  "source": {
    "directories": [
      "src/"
    ],
    "excludes": [
      "Enum",
      "EventSubscriber"
    ]
  },
  "timeout": 120,
  "logs": {
    "html": "build/infection/infection.html"
  },
  "mutators": {
    "@default": true,
    "MBString": false,
    "global-ignore": [
      "App\\Kernel",
      "App\\Admin\\*",
      "App\\Controller\\IndexController",
      "App\\Repository\\*::save",
      "App\\Security\\*",
      "App\\Service\\MailService",
      "App\\Twig\\AppExtension",
      "App\\ValueObject\\*"
    ],
    "global-ignoreSourceCodeByRegex": [
      "#\\[IsGranted\\(.*",
      ".*EnvService::isNotTesting\\(\\).*"
    ],
    "ArrayItemRemoval": {
      "ignoreSourceCodeByRegex": [
        ".*methods: \\[.*"
      ]
    },
    "CastFloat": {
      "ignore": [
        "App\\Entity\\Measurement::getStatusComparison"
      ]
    },
    "CastInt": {
      "ignore": [
        "App\\Entity\\Measurement::getSsc"
      ]
    },
    "CastString": {
      "ignore": [
        "App\\Entity\\User::setUserDataFromAdUser"
      ]
    },
    "NullSafeMethodCall": {
      "ignore": [
        "App\\Entity\\Device::isMethod"
      ]
    },
    "Continue_": {
      "ignore": [
        "App\\Entity\\Worklist::getFree"
      ]
    },
    "ArrayOneItem": {
      "ignore": [
        "App\\Entity\\Worklist::getFree"
      ]
    },
    "PublicVisibility": {
      "ignore": [
        "App\\Entity\\Worklist::getMeasurements"
      ]
    },
    "DecrementInteger": {
      "ignore": [
[...]

PR #2773, opened 2026-01-02
docs: Update the GitHub issues and PR templates

theofidry:
From https://github.com/infection/infection/issues/2770 I noticed that we did not suggest to copy the infection command used, which I think is a mistake.

From there I decided to have a quick look at the templates. I proposed some suggestions for in the content of each templates. Besides this, here is a recap of the proposed changes:

- Use an alpha-numeric prefix for the files to guarantee the order (see https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#changing-the-order-of-templates).
- Split the feature and RFC templates. This allows to makes the templates more custom, with the title prefix and label. In practice it is true a feature can be an RFC and vice-versa, but it's not always the case. And historically most of the RFCs written were not features but architectural changes.
- Delete an old (unused) issue template file.
- Note that the renaming of the files was suggested by the GitHub template editor. It originally included the emojis in the title which I decided to replace by the alphanumeric prefix for the order.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -959,16 +959,16 @@
         },
         {
             "name": "sanmai/di-container",
-            "version": "0.1.5",
+            "version": "0.1.6",
             "source": {
                 "type": "git",
                 "url": "https://github.com/sanmai/di-container.git",
-                "reference": "355534ad7970fc7dab4211ecaf2da5c546855ee8"
+                "reference": "e57d6de5309be56ca3a810987bf8279d69ce4661"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/04-custom.md
```diff
@@ -0,0 +1,20 @@
+---
+name: "\U0001F917 Support Question"
+about: "If you have a question \U0001F4AC, please check out our Discord, Gitter or
```
maks-rafalko: we don't use Gitter anymore, so we can remove it while we are here. 

Instead, we also have Discussion on GitHub.

maks-rafalko (approved):

sanmai (approved):
🚀

PR #2774, opened 2026-01-02
[PoC]: Evaluate what is missing for new AST archi

theofidry:
This PR is to continue the AST enrichment part first drafted in #2272.

The goal is to have a functional version with feature-parity and determine a plan for an incremental port.


src/TestFramework/Tracing/Trace/LineRangeCalculator.php
```diff
@@ -36,17 +36,18 @@
...
  * @internal
  */
 final class LineRangeCalculator
 {
-    public function calculateRange(Node $originalNode): NodeLineRangeData
-    {
-        if ($originalNode->getAttribute(ReflectionVisitor::IS_ON_FUNCTION_SIGNATURE, false) === true) {
+    public function calculateRange(
+        Node $originalNode,
+        bool $isOnFunctionSignature,
```
theofidry, line 48: to extract this: we try to memoize this call. I'm not sure if it's worth memoizing but this is clearly duplicated here

src/PhpParser/NodeTraverserFactory.php
```diff
@@ -87,11 +105,68 @@ public function create(NodeVisitor $mutationVisitor, array $nodeIgnorers): NodeT
         return $traverser;
     }
 
-    public function createPreTraverser(): NodeTraverserInterface
-    {
-        $traverser = new NodeTraverser();
-        $traverser->addVisitor(new NextConnectingVisitor());
+    /**
+     * TODO: this replaces the "createPreTraverser": this "pre" traverse, which
```
theofidry, line 70: I think this should be the starting point: port things done in the `NodeMutationGenerator` to the pre-traverse

PR #2778, opened 2026-01-03
refactor: Inject NameResolver into NodeTraverserFactory

sanmai:
## Description

Instead of `NodeTraverserFactory` creating `NameResolver` via static factory, inject it through the constructor following proper DI principles.  

## Changes

- Converts NameResolverFactory to implement Builder interface
- Registers NameResolver in Container with its factory
- Updates tests to use SingletonContainer
- Adds a test for the builder

Arguably this was a low-hanging fruit, but it is something we can do elsewhere to reduce the scope of the Container class that we struggle to test (see #2765). Unlike the container, builders are trivially testable.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -959,16 +959,16 @@
         },
         {
             "name": "sanmai/di-container",
-            "version": "0.1.7",
+            "version": "0.1.8",
             "source": {
                 "type": "git",
                 "url": "https://github.com/sanmai/di-container.git",
-                "reference": "f6c0de5f1c31d303148e46107284a16948b4736a"
+                "reference": "f7ea6a00692608f785fc0eabb1c54f5349d973e9"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/sanmai/di-container/zipball/f6c0de5f1c31d303148e46107284a16948b4736a",
-                "reference": "f6c0de5f1c31d303148e46107284a16948b4736a",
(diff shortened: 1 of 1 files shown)


src/PhpParser/Visitor/NameResolverBuilder.php
```diff
@@ -35,26 +35,27 @@
...
     {
         return new NameResolver(
             errorHandler: null, // Ensure we throw.
-            options: [
-                'preserveOriginalNames' => true,
-                // must be `false` for pretty-printing to work properly
-                // @see https://github.com/nikic/PHP-Parser/blob/master/doc/component/Pretty_printing.markdown#formatting-preserving-pretty-printing
-                'replaceNodes' => false,
-            ],
+            options: self::OPTIONS,
```
sanmai, line 58: Not something we can mutate yet.

src/PhpParser/Visitor/NameResolverBuilder.php
```diff
@@ -42,18 +42,20 @@
  * @internal
  * @implements Builder<NameResolver>
  */
-final class NameResolverFactory implements Builder
+final class NameResolverBuilder implements Builder
```
theofidry, line 45: doesn't it look super weird that this class now needs to be aware of the Container API?

sanmai: Agree, I'd rather extract only what is really need a test like #2782

PR #2780, opened 2026-01-03
test: Introduce a new test framework for PhpParser visitors

theofidry:
## Description

This PR adds a new testing framework for the visitors, and more generally the AST enrichment in Infection and migrate the existing tests to that new framework.

## Problem

Testing the AST visitors is a complicated and cumbersome business. Indeed, there is a lot of challenging constraints:

- You can't really create a comprehensive AST out of thin air easily: it is a ton of boilerplate code, and it is hard to come up with something sensible. Unless we need to apply code on 1-2 nodes, we never really on such a method because of it and instead parse a code string to get out the statements.
- ASTs tends to be massive. Because of it, we tend to do assumptions about what nodes we are getting, directly access specific nodes and check the state of those hoping that is all that is needed. The promise is quite fragile, and often wrong: there may be more side-effects than expected resulting in the rest of the tree not being in the expected state. `InterfaceIgnorerTest`, `NextConnectingVisitorTest` and `ReflectionVisitorTest` all do those partial checks for example.
- Without a visual representation, it is extremely hard to see what the AST looks like and its state, much less what the visitors traversed.

What pushed this initiative was #2272 where I wanted to change the way we enriched the AST and where I tried to do some sort of optimisations and avoid traversing parts of the AST we were not interested in. I also was confused about how some of the current code was behaving, and the existing tests were unsatisfactory. Without a visual support, this kind of things are incredibly hard to test properly in a readable and maintainable way.

I also believe that even without those architecture changes, we are already facing those challenges. More moderately, true, but they are already existent. In fact, while porting the existing tests to this new testing framework, I encountered more than one case where the tests was either misleading, or at least was not doing what it stated (I'll pin them in the diff).

## The solution

The fundamental idea of this is extremely simple: we leverage the PhpParser `NodeDumper` to get a string representation of the AST, and do our assertions against this instead of asserting state on individual nodes.

The idea is simple, but the practice required some tooling to make it smooth. The general implementation is as follows:

- A custom `NodeDumper`. Unfortunately (it was to be expected), the PhpParser `NodeDumper` doesn't quite fit the bill out of the box. That being said, the changes are not crazy, so I copy/pasted it, annotated the changes, and kept the changes to this class to the minimum. I captured all the modified behaviour. This class is bound to be changed in PhpParser in the future and so we must be able to back-port those changes easily. To this end, keeping the changes minimal is in our best interests. I'll see if we can work out a better solution, but I don't want us to be blocked by this either.
- A new `VisitorTestCase` base PHPUnit test case. It is tailored toward that usage of testing the dumped AST. It has too little similitude with `BaseVisitorTestCase` and the latter is still used for 2 specific tests (which I would like to change in the future but should stay for the time being).
- Little visitor utilities to make the AST output simpler:
  - `AddIdToTraversedNodesVisitor`: add IDs (a 0 starting sequence) to nodes. This makes it easier to identify at times and also allows the `NodeDumper` to replace a node found as an attribute by its IDs, which makes it less noisy, and allows to avoid circular dependencies (this was mostly relevant for `ReflectionVisitor` for which a node can have itself pointed by the `functionScope` attribute).
  - `KeepOnlyDesiredAttributesVisitor` which can help to remove attributes necessary for the execution of the test, but irrelevant to the final output (i.e. this is to remove noise).
  - `MarkTraversedNodesAsVisitedVisitor` a must have for visitors which may cause to skip a traversal (fully or partially). The node dumper can be configured to show all nodes, but otherwise, by default, it will render a node not having been marked as visited as `<skipped>`.

For a breakdown, you can check the documentation at `tests/phpunit/PhpParser/Visitor/README.md`.

## Examples

Due to how big the diff is, I feel like it will be easier to showcase a few examples.

<details>
<summary>IgnoreAllMutationsAnnotationReaderVisitor</summary>

Before:

```php
    public function test_it_toggles_ignorer_for_nodes_commented_with_expected_annotation(): void
    {
        $changingIgnorer = $this->createMock(ChangingIgnorer::class);
        $changingIgnorer
            ->expects($this->once())
            ->method('startIgnoring')
        ;

        $ignoredNodes = new SplObjectStorage();

        $visitor = new IgnoreAllMutationsAnnotationReaderVisitor($changingIgnorer, $ignoredNodes);

        $comment = $this->createMock(Comment::class);
        $comment
            ->expects($this->once())
            ->method('getText')
            ->willReturn('@infection-ignore-all')
        ;

        $nodeWithComments = $this->createMock(Node::class);

        $nodeWithComments
            ->expects($this->once())
            ->method('getComments')
            ->willReturn([$comment])
        ;

        $visitor->enterNode($nodeWithComments);

        $this->assertCount(1, $ignoredNodes);
    }
```

It works, but it hard to understand what is going on, and if you like visual representations you'll have to do some mental gymnastic to get there. Meanwhile, here's the new version:

```php
yield 'comment on a method' => [
    <<<'PHP'
        <?php

        namespace Infection\Tests\Virtual;

        class ClassWithExcludedMethod {
            function nonExcludedMethod() {}

            // @infection-ignore-all
            function excludedMethod() {}
        }

        PHP,
    <<<'AST'
        array(
            0: Stmt_Namespace(
                name: Name
                stmts: array(
                    0: Stmt_Class(
                        name: Identifier
                        stmts: array(
                            0: Stmt_ClassMethod(
                                name: Identifier
                            )
                            1: <skipped>
                        )
                    )
                )
                kind: 1
            )
        )
        AST,
];
```

</details>

<details>
<summary>InterfaceIgnorerTest</summary>

Before:

```php
        yield 'interfaces are ignored' => [
            <<<'PHP'
                <?php

                interface Bar
                {
                    public function nope(Bar $ignored): void;
                }
                PHP,
            0,
        ];

        yield 'classes arent ignored' => [
            <<<'PHP'
                <?php

                class Bar
                {
                    public function nope(Bar $counted)
                    {
                        $counted = true;
                    }
                }
                PHP,
            2,
        ];
```

This one is not very verbose, but also the test is fairly inaccurate: we only count the visited nodes and do an assertion on it. There is zero guarantee or indication about what node has been visited. Meanwhile the new version:

```php
yield 'it ignores only interfaces' => [
    <<<'PHP'
        <?php

        interface Bar
        {
            public function nope(Bar $ignored): void;
        }

        class Bar
        {
            public function nope(Bar $counted)
            {
                $counted = true;
            }
        }

        PHP,
    <<<'AST'
        array(
            0: <skipped>
            1: Stmt_Class(
                name: Identifier
                stmts: array(
                    0: Stmt_ClassMethod(
                        name: Identifier
                        params: array(
                            0: Param(
                                type: Name
                                var: Expr_Variable
                            )
                        )
                        stmts: array(
                            0: Stmt_Expression(
                                expr: Expr_Assign(
                                    var: Expr_Variable
                                    expr: Expr_ConstFetch(
                                        name: Name
                                    )
                                )
                            )
                        )
                    )
                )
            )
        )
        AST,
];
```

</details>

## A note on the code

Given the size of the PR, I do feel this deserves a mention. At the time of writing the diff is `+5,514 −877`.

It can seem like a lot, but I think this number is highly inflated. Here's a breakdown:

- +224L are for pure documentation (markdown files).
- +1,453L (26%) is for the `NodeDumper` and its tests
- +631L (11%) are the test utilities themselves and their tests (`VisitorTestCase` and `tests/phpunit/TestingUtilities/PhpParser/Visitor/*`)
- So on this +3,206L (60%), it is actually purely on the tests. Some are new (617, so 20% – actually more as I added more tests in some existing tests, not just new test cases), did not exist before, and others are simply the same tests, which tend to be more verbose.

IMO a certain verbosity is unavoidable with visuals like such, but personally at least, I'd pick them any day over a super procedural and partial existing tests.


tests/phpunit/TestingUtility/PhpParser/Visitor/AddIdToTraversedNodesVisitor/AddIdToTraversedNodesVisitor.php
```diff
@@ -35,74 +35,66 @@
...
-use PhpParser\NodeVisitorAbstract;
-use PHPUnit\Framework\Attributes\CoversClass;
-use PHPUnit\Framework\Attributes\Group;
+use PhpParser\NodeTraverser;
+use PhpParser\Parser;
+use PHPUnit\Framework\TestCase;
 
-#[Group('integration')]
-#[CoversClass(NonMutableNodesIgnorerVisitor::class)]
-final class NonMutableNodesIgnorerVisitorTest extends BaseVisitorTestCase
```
theofidry, line 48: git doing weird things again, this class is new, the moved `NonMutableNodesIgnorerVisitorTest` is at https://github.com/infection/infection/pull/2780/changes#diff-c0c5180781dc3b36fe5d15b294c0057f99f5eeb504cc6752632e832c03645356

tests/phpunit/PhpParser/Visitor/VisitorCollectorIntegration/VisitorCollectorIntegrationTest.php
```diff
@@ -0,0 +1,289 @@
...
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Group;
+use function Safe\file_get_contents;
+
+/**
+ * The goal of this test is to showcase the state of the tree as the MutationVisitor sees it.
+ */
+#[Group('integration')]
+#[CoversNothing]
+final class VisitorCollectorIntegrationTest extends VisitorTestCase
```
theofidry, line 51: This test is entirely new and quite important. A bit like #2444, it's a lot to take it, but I think is very valuable.

It should be used with care as it's _very_ verbose, but it should be able to help us to nail the most tricky scenarios.

tests/phpunit/PhpParser/Visitor/NextConnectingVisitorTest.php
```diff
@@ -36,309 +36,650 @@
...
+
+        yield 'it handles functions and closures as boundaries' => [
+            <<<'PHP'
+                <?php
+
+                $a = 1; // no next
+
+                function test() {
+                    $b = 2; // next = $c
+                    $c = 3; // no next ;TODO: current has next=$d!
```
theofidry, line 143: @sanmai I was quite surprised when seeing this, at least this is not what I expected when reading the code/usage of this class that you introduced in https://github.com/infection/infection/pull/2296.

Could you clarify if those are legit bugs or if it is my misunderstanding? I marked all the suspicious cases with `TODO` in this class.

tests/phpunit/PhpParser/Visitor/ReflectionVisitorTest.php
```diff
@@ -36,344 +36,1287 @@
...
-        $this->assertTrue($this->spyVisitor->isInsideFunction);
-    }
-
-    public function test_it_does_not_traverse_a_regular_global_or_namespaced_function(): void
-    {
-        $code = $this->getFileContent('Reflection/rv-inside-function.php');
-
-        $this->parseAndTraverse($code);
-
-        $this->assertFalse($this->spyVisitor->isInsideFunction);
```
theofidry, line 116: interestingly @maks-rafalko, this test which got updated with https://github.com/infection/infection/pull/508 does not do what it claims.

Actually the `NodeVisitor::DONT_TRAVERSE_CURRENT_AND_CHILDREN` appears to be dead code (which I'll fix separately, but I could only notice this issue with confidence thanks to the new tests).

tests/phpunit/PhpParser/Visitor/README.md
```diff
@@ -0,0 +1,217 @@
...
+    0: Stmt_Namespace(
+        name: Name(
+            nodeId: 1
+            parent: nodeId(0)
+        )
+        stmts: array(
+            0: Stmt_Class(
+                name: Identifier(
+                    nodeId: 3
+                    parent: nodeId(2)
```
maks-rafalko, line 179: where is node with `nodeId: 2` in this example?

tests/phpunit/PhpParser/Visitor/README.md
```diff
@@ -0,0 +1,217 @@
...
+    // 2. Add IDs for easier debugging
+    $this->addIdsToNodes($nodes);
+
+    // 3. Apply visitors
+    (new NodeTraverser(
+        new ParentConnectingVisitor(),
+        new NameResolver(),
+        new ReflectionVisitor(),
+        new MarkTraversedNodesAsVisitedVisitor(),
+    ))->traverse($nodes);
```
maks-rafalko, line 147: _this is the first file that I'm reading, so leaving comments as I go (probably it will be clearer later for me)_

1. just to be consistent, does it make sense to use `$this->tranverser->tranverse()` here?
2. and doing p.1 will let us always add `new MarkTraversedNodesAsVisitedVisitor(),` (as I understood from this docs - we always need to add it)

tests/phpunit/PhpParser/Visitor/README.md
```diff
@@ -0,0 +1,217 @@
...
+    string $code,
+    ?array $desiredAttributes,
+    string $expected,
+): void {
+    // 1. Parse the code
+    $nodes = $this->parser->parse($code);
+    $this->assertNotNull($nodes);
+
+    // 2. Add IDs for easier debugging
+    $this->addIdsToNodes($nodes);
```
maks-rafalko, line 139: just a question and to discuss: do we always need to add ids? If so, I would think about adding some helper `parse()` method (or whatever it will be named) that will do it for us

tests/phpunit/PhpParser/Visitor/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+```php
+public function test_it_annotates_nodes(
+    string $code,
+    ?array $desiredAttributes,
+    string $expected,
+): void {
+    // 1. Parse the code
+    $nodes = $this->parser->parse($code);
+    $this->assertNotNull($nodes);
```
maks-rafalko: just a question: do we need this call? looks like we can throw if it's null and this can be encapsulated to `parse()`? does it make sense?

tests/phpunit/PhpParser/Visitor/ReflectionVisitorTest.php
maks-rafalko: yeah let's extract to a separate PR. Are you sure it wasn't working at the time of creation as well? (probably it was broken later?... idk)

why I'm surprised, because I contributed this exact `NodeVisitor::DONT_TRAVERSE_CURRENT_AND_CHILDREN` to php-parser, so it did some job for us at the time of creation

tests/phpunit/TestingUtility/PhpParser/NodeDumper/NodeDumper.php
```diff
@@ -0,0 +1,560 @@
...
+{
+    private const IGNORE_ATTRIBUTES = [
+        'comments' => true,
+        'startLine' => true,
+        'endLine' => true,
+        'startFilePos' => true,
+        'endFilePos' => true,
+        'startTokenPos' => true,
+        'endTokenPos' => true,
+        // The following attribute(s) is/are Infection specific
```
maks-rafalko:
```suggestion
        // Infection specific attribute(s)
```

I suggest to follow `// Infection specific XXX` pattern so that it's easy to find overriding/new code created by us just by searching by `// Infection specific`

tests/phpunit/TestingUtility/PhpParser/NodeDumper/PotentialCircularDependencyDetected.php
```diff
@@ -0,0 +1,58 @@
...
+
+final class PotentialCircularDependencyDetected extends UnexpectedValueException
+{
+    public static function forAttribute(
+        string $key,
+        Node $value,
+    ): self {
+        return new self(
+            sprintf(
+                'The attribute "%s" found a node instance "%s" (#%s). The NodeDumper cannot support those as they may trigger circular dependencies. Either remove the attribute before dumping, do not dump extra attributes or add an ID the node.',
```
maks-rafalko:
```suggestion
                'The attribute "%s" found a node instance "%s" (#%s). The NodeDumper cannot support those as they may trigger circular dependencies. Either remove the attribute before dumping, do not dump extra attributes or add an ID to the node.',
```

tests/phpunit/TestingUtility/PhpParser/Visitor/MarkTraversedNodesAsVisitedVisitor/MarkTraversedNodesAsVisitedVisitor.php
```diff
@@ -0,0 +1,72 @@
...
+    public static function markAsVisited(Node $node): Node
+    {
+        $node->setAttribute(self::VISITED_ATTRIBUTE, true);
+
+        return $node;
+    }
+
+    public function enterNode(Node $node): void
+    {
+        self::markAsVisited($node);
```
maks-rafalko, line 70: btw, since then we started using static methods instead of instance methods? Is the any information about benefits? (just can't remember anything, did I miss it?)

tests/phpunit/TestingUtility/PhpParser/Visitor/MarkTraversedNodesAsVisitedVisitor/MarkTraversedNodesAsVisitedVisitor.php
```diff
@@ -0,0 +1,72 @@
...
+        return $node->hasAttribute(self::VISITED_ATTRIBUTE);
+    }
+
+    /**
+     * @template T of Node
+     *
+     * @param T $node
+     * @return T
+     */
+    public static function markAsVisited(Node $node): Node
```
maks-rafalko, line 61: it's a bit strange that we expose it and use this method in `enterNode()` instead of inlining the logic directly in `enterNode()`.

Is it exposed for testing purposes only?

feel free to resolve

tests/phpunit/PhpParser/Visitor/IgnoreNode/InterfaceIgnorerTest.php
```diff
@@ -78,13 +94,85 @@ public function nope(Bar $counted)
...
+                                                name: Name
+                                            )
+                                        )
+                                    )
+                                )
+                            )
+                        )
+                    )
+                )
+                AST,
```
maks-rafalko, line 175: not sure I got this test. Where is the interface in the dump output?

tests/phpunit/PhpParser/Visitor/NextConnectingVisitorTest.php
```diff
@@ -35,310 +35,689 @@
...
+                            expr: Scalar_Int(
+                                rawValue: 1
+                                kind: KIND_DEC (10)
+                                nodeId: 3
+                            )
+                            nodeId: 1
+                        )
+                        nodeId: 0
+                        next: nodeId(4)
+                    )
```
maks-rafalko, line 101: just thinking: it wasn't easy here to spot which `Stmt_Expression` is for `$a` which is for `$b` etc. And this is a very simple code.

I guess it will be even harder with the more advance and complex code. Would it be possible to use names (attribute) in such dump? (when needed).

Do you think it's a problem (or not)?

maks-rafalko (commented):
Generally, I like the visual representation, as for me - totally makes sense.

Left some questions / minor suggestions after the first review.

tests/phpunit/PhpParser/Visitor/ReflectionVisitorTest.php
theofidry: I wouldn't be surprised that it changed at some point in time, might be hard to revert though.

I don't think it needs immediate addressing either. Well I don't know. Currently my plan is get this through, then I'll be able to incrementally transfer things we currently do during the 2nd traverse (the original one, to mutate code) or in `NodeMutationGenerator`, to the pre-traverse.

I could do most of it in https://github.com/infection/infection/pull/2774, but I had concerns about breaking things or misunderstanding the original code. One thing I noticed there is that maybe there is parts of the AST that cannot be skipped as easily as I thought. I'm not entirely sure, but anyway, this can be done incrementally.

What I'm trying to say is: even if we restore the original skip traverse, I'm not confident this wouldn't break something else. But in any case, it is things we can tackle separately and hopefully this framework helps us more to navigate through those changes.

tests/phpunit/PhpParser/Visitor/README.md
theofidry: AI slop... my bad... I'm replacing it by a real example.

tests/phpunit/PhpParser/Visitor/README.md
theofidry: Actually we no longer need to.

We needed it because technically `PhpParser\Parser::parse()` is flexible. It allows to return `null`, but can also throw, it depends on the error handler given. So from PhpStan perspective, it looks like it can return `null` despite us configuring it so that it would throw an exception rather than returning `null`.

I got irritated by it and just adding `VisitorTestCase::parse()` to make PHPStan happy, but I forgot to update it here.

Actually, I'll update the `VisitorTestCase::parse()`  to add a comment about this.

tests/phpunit/PhpParser/Visitor/README.md
theofidry: Hmm.

Originally I would have said no, because I only really needed it in a handful of places. But since then, I figured actually I could use them for `NodeDumper` to deal with circular dependencies.

~~So yeah, make sense~~

There is actually quite a few cases (more than not) that don't need it, so it is better to leave alone. I updated the doc to clarify that it is optional and what it is for to not give the impression it should always be used.

tests/phpunit/PhpParser/Visitor/README.md
theofidry: I'm not certain is the short answer.

>new MarkTraversedNodesAsVisitedVisitor(), (as I understood from this docs - we always need to add it)

It is only necessary for visitors that will cause some nodes to be skipped, like the node ignorers ones. Otherwise (e.g. the `NextConnector`) it isn't.

From my experience so far, both here and with #2774, I've had to regularly use multiple traversers (with different visitors) to do multiple traverses. A good example is to add IDs: you do not want the sequence to be interrupted because some nodes where not traversed, so you add the IDs in a first traverse.

There is a bit less variety of this here with our current visitors, but I had the impression there was more with the ones I was experimenting in #2774.

Another thing is the visitor is what is tested here. Since it is what you need to give to the `NodeTraverser` and you may need other visitors and the order matters, you'll always have a different `NodeTraverser` with state that is bound in your test method. For example:

```php
public function test_it_marks_nodes_with_the_ignore_all_comment_as_ignored(
    string $code,
    string $expected,
): void {
    $nodes = $this->parse($code);

    $changingIgnorer = new ChangingIgnorer();

    $traverser = new NodeTraverser(
        new IgnoreAllMutationsAnnotationReaderVisitor(
            $changingIgnorer,
            new SplObjectStorage(),
        ),
        new NonMutableNodesIgnorerVisitor([$changingIgnorer]),
        new MarkTraversedNodesAsVisitedVisitor(),
    );
    $traverser->traverse($nodes);

    $actual = $this->dumper->dump($nodes);

}
```

I'm not convinced moving the instantiation of `NodeTraverser` in a separate method would be of much help here.

Maybe one thing that could make it slightly nicer is:

```php
// VisitorTestCase
/**
 * @param Node[] $nodes
 * @param NodeVisitor ...$visitors
 * 
 * @return Node[]
 */
final protected function traverse(
    array $nodes,
    NodeVisitor ...$visitors,
): array
{
    return (new NodeTraverser(...$visitors))->traverse($nodes);
}
```

Which would transform the above into:

```diff
public function test_it_marks_nodes_with_the_ignore_all_comment_as_ignored(
    string $code,
    string $expected,
): void {
    $nodes = $this->parse($code);

    $changingIgnorer = new ChangingIgnorer();

--    $traverser = (new NodeTraverser(
--        new IgnoreAllMutationsAnnotationReaderVisitor(
--            $changingIgnorer,
--            new SplObjectStorage(),
--        ),
--        new NonMutableNodesIgnorerVisitor([$changingIgnorer]),
--        new MarkTraversedNodesAsVisitedVisitor(),
--    ))->traverse($nodes);
++    $nodes = $this->traverse(
++        $nodes,
++        new IgnoreAllMutationsAnnotationReaderVisitor(
++            $changingIgnorer,
++            new SplObjectStorage(),
++        ),
++        new NonMutableNodesIgnorerVisitor([$changingIgnorer]),
++        new MarkTraversedNodesAsVisitedVisitor(),
++    );

    $actual = $this->dumper->dump($nodes);

    $this->assertSame($expected, $actual);
}
```

But we don't really save in lines either.

tests/phpunit/TestingUtility/PhpParser/NodeDumper/NodeDumper.php
theofidry: Applied

tests/phpunit/TestingUtility/PhpParser/Visitor/MarkTraversedNodesAsVisitedVisitor/MarkTraversedNodesAsVisitedVisitor.php
theofidry: >btw, since then we started using static methods instead of instance methods?

I don't think we've been super consistent with this. Personally I tend to reflex to use static as the default, and sometimes I like to use it even if there is some state and pass the state as an argument to make easier what state is injected where (very opinionated, and depends a lot of the cases).

Don't quote me on this but I think there is a micro optimization about having static functions, but more importantly IMO, it's a good signal that it is stateless (well, unless you're using static state ofc...)

tests/phpunit/TestingUtility/PhpParser/Visitor/MarkTraversedNodesAsVisitedVisitor/MarkTraversedNodesAsVisitedVisitor.php
theofidry: >it's a bit strange that we expose it and use this method in enterNode() instead of inlining the logic directly in enterNode().

Unless super duper obvious I do like the expressiveness of having a method, i.e. you read:

```
enterNode(Node $node) {
  self::markAsVisited($node);
}
```

Doesn't really matter how verbose/complex `markAsVisited()` is, you immediately understand.

Now I agree here it is overkill, but in this specific case this is done because `markAsVisited()` is used by `NodeDumper` too.

tests/phpunit/PhpParser/Visitor/IgnoreNode/InterfaceIgnorerTest.php
theofidry: `0: <skipped>` L149: it was to show it only skips the interface but otherwise continues the traversal (i.e. it doesn't entirely stop the traversal)

tests/phpunit/PhpParser/Visitor/NextConnectingVisitorTest.php
theofidry: >just thinking: it wasn't easy here to spot which Stmt_Expression is for $a which is for $b etc. And this is a very simple code.

I agree. I like having a visual AST, but it remains quite verbose and as a result it is sometimes hard to spot the info you are looking for. I had to rely on the IDE syntax highlighting to spot the `next:` nodes to double check.

>Would it be possible to use names (attribute) in such dump? (when needed).

What do you mean by that?

Originally I admit having such tests wasn't my goal per say, I just needed _something_ that could help me figure out if my skip nodes optimisations were working and I couldn't make sense of it without dumping it in such a way. So currently what we get is as close as it gets to the native `PhpParser\NodeDumper` output.

We could however decide to commit more to this approach and further customize `NodeDumper`. It means deviating more from the original implementation so backporting code will be harder, but it also means we could maybe work out more readable outputs sometimes.

For instance I'm not too found of `array()`, sometimes there is some properties of the node you may not care about, but I wasn't too sure what was the best approach there and if it was not just going to end up creating more confusion.

tests/phpunit/PhpParser/Visitor/IgnoreNode/InterfaceIgnorerTest.php
maks-rafalko: ok, it's not so obvious to spot it

tests/phpunit/PhpParser/Visitor/NextConnectingVisitorTest.php
maks-rafalko: > What do you mean by that?

Would it be possible to configure our dumper in such a way, that it prints node name in addition to the current attributes? So that it will be easier to understand where is the variable `$a`, where is `$b`, etc. (if needed)

Again: the reason of asking is because it was hard to understand the dump for very simple code, because we loos variable names

maks-rafalko (approved):

tests/phpunit/PhpParser/Visitor/IgnoreNode/InterfaceIgnorerTest.php
theofidry: no, it's a case where I wonder if there was would be a better way to render it because we don't really care about what the other nodes are, just that they are not skipped...

tests/phpunit/PhpParser/Visitor/NextConnectingVisitorTest.php
theofidry: I'll check this separately, I do think we'll iterate more on this solution for  better readability

PR #2782, opened 2026-01-03
refactor: Extract a builder from Container, refactor container namespace

sanmai:
## Description

Move the `IndexXmlCoverageParser` instantiation logic to a dedicated `Builder` class. This makes the `sourceFilter` null-check testable and reduces escaped mutants in `Container` (see #2765).

Also, refactor the container into own namespace and place the builder locally.

## Checklist

- [x] Tests added/updated

### DIFF
--- a/src/PhpParser/FileParser.php
+++ b/src/PhpParser/FileParser.php
@@ -43,12 +43,11 @@ use Throwable;
 
 /**
  * @internal
- * @final
  */
-class FileParser
+final readonly class FileParser
 {
     public function __construct(
-        private readonly Parser $parser,
+        private Parser $parser,
     ) {
     }
 
--- a/tests/phpunit/Mutation/FileMutationGenerator/FileMutationGeneratorTest.php
+++ b/tests/phpunit/Mutation/FileMutationGenerator/FileMutationGeneratorTest.php
@@ -49,6 +49,7 @@ use Infection\Tests\Fixtures\PhpParser\FakeNode;
 use Infection\Tests\PhpParser\FakeToken;
 use function iterator_to_array;
 use PhpParser\NodeTraverserInterface;
+use PhpParser\Parser;
 use PHPUnit\Framework\Attributes\CoversClass;
 use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\MockObject\MockObject;
@@ -58,7 +59,7 @@ use Symfony\Component\Finder\SplFileInfo;
 #[CoversClass(FileMutationGenerator::class)]
 final class FileMutationGeneratorTest extends TestCase
 {
-    private MockObject&FileParser $fileParserMock;
+    private MockObject&Parser $phpParserMock;
 
     private MockObject&NodeTraverserFactory $traverserFactoryMock;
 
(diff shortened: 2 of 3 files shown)


src/Container/Container.php
```diff
@@ -240,9 +241,7 @@ final class Container extends DIContainer
     public static function create(): self
     {
         $container = new self([
-            IndexXmlCoverageParser::class => static fn (self $container): IndexXmlCoverageParser => new IndexXmlCoverageParser(
-                isSourceFiltered: $container->getConfiguration()->sourceFilter !== null,
-            ),
+            IndexXmlCoverageParser::class => IndexXmlCoverageParserBuilder::class,
```
theofidry, line 244: this seems off to me tbh.

In the end what we're doing here is moving a container specific factory out of the container. Why not, but I struggle to see the goal here, because I'm not convinced moving all of those closures declared here to container builders will really improve the readability. The container itself will be simpler in the end for sure, but following what is going on won't (IMO).

I'm also not too crazy about the idea of having container services sprinkled a bit everywhere in the codebase. If we want to go down that route I would suggest:

- moving`src/Container.php` to `src/Container/Container.php`
- moving all the Container builders to `src/Container/Builder/*`

sanmai: Thanks, it makes total sense not to sprinkle. The goal is test the parts of the container without having to cover the container itself.

Let me move these things.

theofidry (commented):
Codewise LGTM. I'll leave the approval to @maks-rafalko because I'm really sure what to think of adding tests for (parts of) the service registration to the container.

maks-rafalko (approved):
> Codewise LGTM. I'll leave the approval to @maks-rafalko because I'm really sure what to think of adding tests for (parts of) the service registration to the container.

let's see if it will be irritating or not. (I don't have a strong opinion as I didn't write much code recently). But  the least we can do - give it a chance.

theofidry:
Hmm, I like having the checklist when creating the PR, but less so in the commit message. Still worth having in the template?

maks-rafalko:
> Hmm, I like having the checklist when creating the PR, but less so in the commit message. Still worth having in the template?

could you please clarify what checklist and commit message you are referring to? Which template?

theofidry:
The PR template:

Screenshot 2026-01-05 at 18 53 10

maks-rafalko:
https://github.com/infection/infection/blob/master/.github/PULL_REQUEST_TEMPLATE.md - we have a template with checklist (not the same that is posted in the image above though)

So what worth adding to the template? I didn't get the proposal

theofidry:
I don't know. I'm wondering if that checklist should be there at all because I don't think it makes sense for it to land in the commit message (we default to copy the title + description to the commit – which ofc you can change). Maybe it's not worth thinking over though I don't know.

maks-rafalko:
> we default to copy the title + description to the commit 

do you mean GitHub automatically copies all the content of the PR's description to a commit message during merge/squash?

maks-rafalko:
image

Has it been changed? I remember that previously my commit message had higher priority and wasn't overridden by title+description of the PR. Not sure how all of this works though now, need to test.

I feel that we had another option chosen before - title + commit details

theofidry:
>do you mean GitHub automatically copies all the content of the PR's description to a commit message during merge/squash?

As a default yes, you always have the option to adjust it as you click the "squash and merge" as before.

> I remember that previously my commit message had higher priority and wasn't overridden by title+description of the PR. 

I did change that because I think it's a better experience overall:

- if you do one or multiple messy commits, you can work on having the PR title & description clean and that is what will be taken as the default.
- if you do a nice commit title + description and you want this to be picked, as you open the PR, this is what will be picked

So from my PoV, this looks like preserving the previous behaviour we had + having a nicer experience with multi-commits PRs.

sanmai:
You can always switch to just "pull request title" and the long text will be gone.

theofidry:
I guess it's fine, I do like having the description because it's usually something that is refined to explain the PR.

PR #2784, opened 2026-01-03
test: Do not mark the Container as covered

theofidry:
## Description

We have some service definitions being covered by tests due to `ContainerTest` creating code coverage.

## Changes

- Replacement for #2765.

## Reviewer Notes

Personally I'm not really willing to go test all the services registered of the container: it sounds like a lot of work, for not much value, and likely very brittle test (how are you gonna test efficiently the laziness or caching?).

The overall soundness of the configuration _is_ tested, at least partially, via integration tests for some services and end to end tests. But it is not something I really want to add mutation testing on top of.

If someone believes strongly against it though, it's not something I really want to argue over and I can leave it alone.


sanmai:
Going to merge this to unblock further refactors.

PR #2788, opened 2026-01-04
perf: Do not fetch the file contents multiple times

theofidry:
## Description

For some reasons I thought the `Symfony\Finder\SplFileInfo::getContents()` was caching the result. It appears it doesn't, which means we fetches the content of the same file multiple times (unnecessarily as we do not modify neither do we expect those files to change).

## Solution

This PR introduces a `FileStore` service which decorates the `FileSystem` to leverage `::readFile()` and cache its result.

```php
// Before:
$finderFileInfo->getContents();

// After:
$fileStore->getContents($pathOrFileInfo);
```

There is two nice side-effects from this:

- Despite being an I/O operation, we leverage the existing `FileSystem` so we preserve the "same experience": the same type of exception will be thrown and if we are using a readonly or fake filesystem would those carry.
- We no longer need to depend on `Symfony\Finder\SplFileInfo` for this.

Note that we only store the last file content: indeed, currently we do not need more so it is more memory friendly to do so.

## Results

| benchmark               | its | mem_peak  | mode   | rstdev |
|-----------------------|-----|-----------|--------|--------|
| MutationGeneratorBench (no cache)   | 290.336mb | 6.767s | ±0.46% |
| MutationGeneratorBench (cache) | 293.012mb | 6.419s | ±1.56% |

### DIFF
--- a/tests/phpunit/Mutation/FileMutationGenerator/FileMutationGeneratorTest.php
+++ b/tests/phpunit/Mutation/FileMutationGenerator/FileMutationGeneratorTest.php
@@ -80,7 +80,7 @@ final class FileMutationGeneratorTest extends TestCase
         $fileSystemStub
             ->method('readFile')
             ->willReturnCallback(
-                static fn (string $path) => sprintf(
+                static fn (string $path): string => sprintf(
                     'contents(%s)',
                     $path,
                 ),


src/FileSystem/FileStore.php
```diff
@@ -0,0 +1,71 @@
...
+     */
+    public function getContents(SplFileInfo|string $file): string
+    {
+        $path = is_string($file)
+            ? $file
+            : $file->getRealPath();
+
+        Assert::notFalse($path);
+
+        return $this->contents[$path] ??= $this->fileSystem->readFile($path);
```
sanmai: I'd argue that it could be better to fetch the file again from the file system _cache_ rather than cache it ourselves indefinitely.

If we know we load the same file over and over, and only cache the most recently read file, we get the best of both worlds: avoiding duplicating the OS’s work while preserving some sanity. 

I.e. we need `$contents` and `$lastReadPath`.

theofidry: Even if the operating system caches it, PHP still does a call to the operating system, so not ideal is it?

theofidry: Locally I still get consistently 5-10% faster with this (although 3MB higher – +1%) with `make benchmark_mutation_generator`

sanmai: And what numbers do you get with "just one cached" approach?

My concern is that it caching everything can disproportionately affect memory usage on larger projects.

theofidry: | benchmark               | its | mem_peak  | mode   | rstdev |
|-----------------------|-----|-----------|--------|--------|
| MutationGeneratorBench (no cache)   | 290.336mb | 6.767s | ±0.46% |
| MutationGeneratorBench (cache) | 293.012mb | 6.419s | ±1.56% |

theofidry: It certainly doesn't seem too impactful at least on that sample, you can double check, but I think that sample should be sufficient to mimic "larger projects".

I'm not sure how PHP deals with those strings internally, but the file content lives for a long time as it's part of the trace too.

If we do not want that caching though, I'll adapt the PR to only switch to use to the filesystem to make 1) mocking easier 2) reduce the dependencies on the Finder SplFileInfo.

sanmai: I reviewed the benchmark. I can see why it does not show any downsides: it iterates all mutations to completion before measuring memory. This doesn't reflect concurrent mode where mutations are processed lazily (i.e. with `--no-progress`): for the released version of Infection on a very large project at any given moment it will only use as much memory as it requires (older files get GC'd). 

With this change as you intend it (i.e. a forever cache for all ever read files) Infection will keep using the memory even for files it will _never_ have to touch again.

A single-entry cache (`$lastPath`, `$lastContents`) gives the same performance benefit since the access pattern is strictly sequential. `FileParser` reads file A, then `FileMutationGenerator` reads file A, then we move to file B - we do not jump around.

theofidry: It'd work, maybe a bit surprising I guess (as it's not a long lasting cache), but it's more eleguant that explicitly evicting it. It's gonna be finicky/fragile anyway.

For the sake of it I would like to benchmark it with the benchmark-sources, but it's gonna take a while to run and I don't feel like waiting 1h on it right now so I'll come back to it later

sanmai: What makes it impossible for you to stick to a single-entry cache? It sounds like a good thing to me.

theofidry: it's possible (and don't doubt we'll get better results knowing the current code), I would just like to measure the impact of either changes

PR #2789, opened 2026-01-05
Canonicalize file pathnames in tracing logic

knallcharge:
## Description

Added `Path::canonicalize` to the trace logic to make the file collector work on Windows systems.

## Changes

Changed `src/TestFramework/Tracing/TraceProviderAdapterTracer.php`.

## Reviewer Notes

Needs to be checked on unix.

## Related issues

Fixes https://github.com/infection/infection/issues/2770.

### DIFF
--- a/src/TestFramework/Tracing/TraceProviderAdapterTracer.php
+++ b/src/TestFramework/Tracing/TraceProviderAdapterTracer.php
@@ -108,7 +108,7 @@ final class TraceProviderAdapterTracer implements Tracer
 
     private function tryToTrace(SplFileInfo $fileInfo): ?Trace
     {
-        $sourcePathname = $fileInfo->getPathname();
+        $sourcePathname = Path::canonicalize($fileInfo->getPathname());
 
         return array_key_exists($sourcePathname, $this->indexedTraces)
             ? $this->indexedTraces[$sourcePathname]
@@ -136,7 +136,7 @@ final class TraceProviderAdapterTracer implements Tracer
 
             $this->isEmpty = false;
 
-            $traceSourcePathname = $trace->getSourceFileInfo()->getPathname();
+            $traceSourcePathname = Path::canonicalize($trace->getSourceFileInfo()->getPathname());
             $this->indexedTraces[$traceSourcePathname] = $trace;
 
             if ($traceSourcePathname === $sourcePathname) {


theofidry:
@knallcharge just to make sure I understand everything, so despite using plain Windows, the paths got from the XML coverage report are normalized?

Could you try if getting the `$fileInfo->getRealPath()` rather than `->getPathname()` (without the normalize) works?

maks-rafalko:
> Could you please add tests to make sure this change works as expected?

let's cover this by test please, otherwise it will be (potentially) broken again

theofidry:
A minimal test case could look like this:

```php
// tests/phpunit/TestFramework/Tracing/TraceProviderAdapterTracerTest.php

    public function test_it_finds_traces_on_windows(): void
    {
        if (!OperatingSystem::isWindows()) {
            $this->markTestSkipped('Requires Windows.');
        }

        $fileInfo = self::createDummySplFileInfo('src/Service1.php');
        $anotherFileInfo = self::createDummySplFileInfo('src/Service2.php');

        $tracesIterator = new ArrayIterator([
            $trace = $this->createTraceMock($fileInfo),
            $this->createTraceMock($anotherFileInfo),
        ]);

        $this->traceProviderMock
            ->expects($this->once())
            ->method('provideTraces')
            ->willReturn($tracesIterator);

        $this->assertSame($trace, $this->tracer->trace($fileInfo));
    }
```

In this test you'll notice there is only one `SplFileInfo`: it is the same instance used for `$this->tracer->trace($fileInfo)` and `$tracesIterator`. There provenance differs however: in practice, the former is constructed by `SourceFileInfoProvider` which comes from the XML code coverage whilst the latter comes from the `SourceCollector`.

Your issues indicates that in one case we get backslashes, but not in the other, and I'm a bit confused as to why because both are `SplFileInfo` instances in the end.

One thing I do notice is that in `XmlCoverageParser` we do `new SplFileInfo($realPath)` whereas for the files coming from the `SourceCollector` they are created by the Symfony Finder.

Hence my previous question, because the fix choose would depend on the case, and how you write the test would differ too.

To be clear, I have no doubt that hammering `Path::canonicalize()` will work, but it is also 2 extra calls on a hotpath, so if we can avoid it it is better.

knallcharge:
> just to make sure I understand everything, so despite using plain Windows, the paths got from the XML coverage report are normalized?

I'm not familiar with the internal workings of infection at all, so I don't think I understand what you mean or how I can verify where which path-info comes from.

> Could you try if getting the $fileInfo->getRealPath() rather than ->getPathname() (without the normalize) works?

Removing both canonicalize-calls and adding `getRealPath` in `tryToTrace()` works, but I can't verify this with the test you added.
Using `getRealPath` will make the test fail as realPath in the mock-object is false and will trigger "_($sourcePathname) must be of type string, false given_"-errors.
Not changing anything in the `TraceProviderAdapterTracer` will make your test pass, but so does adding any of the two canonicalize-calls (or both) so we don't get any assurance from the test.

theofidry:
>I'm not familiar with the internal workings of infection at all, so I don't think I understand what you mean or how I can verify where which path-info comes from.

Apologies I wasn't clear enough with this. This is actually about PHPUnit, if you say execute `phpunit --coverage-xml=var/coverage-xml` and check the paths in any XML paths, are they with backslashes or forward slashes?

>Removing both canonicalize-calls and adding getRealPath in tryToTrace() works, but I can't verify this with the test you added.

Let's me try to adjust the test for it as soon as you can confirm the above about the Coverage XML.

knallcharge:
> check the paths in any XML paths, are they with backslashes or forward slashes?

I have backslashes in those files, this is from the index.xml in `build\artifacts\coverage-xml`:
`<project source="C:\Users\...\...\...\src">`

theofidry:
Ok I'm super confused then 🙈 Which source file ends up with forward slashes then?

knallcharge:
If I dump `$fileInfo` in function `trace()` in TraceProviderAdapterTracer, I get this, so the forward slashes are already in there (wherever that SplFileInfo comes from):

```
Symfony\Component\Finder\SplFileInfo^ {#1845
  -relativePath: "Admin"
  -relativePathname: "Admin\Admin.php"
  path: "C:/Users/.../.../.../src\Admin"
  filename: "Admin.php"
  basename: "Admin.php"
  pathname: "C:/Users/.../.../.../src\Admin\Admin.php"
  extension: "php"
  realPath: "C:\Users\...\...\...\src\Admin\Admin.php"
  aTime: 2026-01-05 12:33:54
  mTime: 2025-12-16 14:53:54
  cTime: 2025-09-08 13:42:18
  inode: 44473046320294275
  size: 9631
  perms: 0100666
  owner: 0
  group: 0
  type: "file"
  writable: true
  readable: true
  executable: false
  file: true
  dir: false
  link: false
  linkTarget: "C:\Users\...\...\...\src\Admin\Admin.php"
}
```
So using `getRealPath()` seems to be a valid solution as realPath already has backslashes in SplFileInfo, while pathName will break it for Windows.

theofidry:
Merged via #2790, thank you @knallcharge

PR #2791, opened 2026-01-06
fix: Use correct FileSystem class in MakeCustomMutatorCommandTest

sanmai:
## Description

The test was mocking `Symfony\Component\Filesystem\Filesystem` but the `Container` expects `Infection\FileSystem\FileSystem` so a real `FileSystem` instance was created instead of using the mock, causing actual files to be written to disk.

### DIFF
--- a/src/TestFramework/Tracing/TraceProviderAdapterTracer.php
+++ b/src/TestFramework/Tracing/TraceProviderAdapterTracer.php
@@ -108,7 +108,8 @@ final class TraceProviderAdapterTracer implements Tracer
 
     private function tryToTrace(SplFileInfo $fileInfo): ?Trace
     {
-        $sourcePathname = $fileInfo->getPathname();
+        $sourcePathname = $fileInfo->getRealPath();
+        Assert::notFalse($sourcePathname);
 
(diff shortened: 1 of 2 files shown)


maks-rafalko (approved):
```php
$fileSystemMock
            ->expects($this->exactly(2))
```

do you have an idea how this worked if it wasn't the right mock?

sanmai:
The container created the actual thing, and it just did what you wanted - created files and such.

tests/phpunit/Command/MakeCustomMutatorCommandTest.php
```diff
@@ -180,7 +180,7 @@ private function createFileSystemMock(): MockObject
     private function createContainer(): Container
     {
         return MockedContainer::createWithServices([
-            Filesystem::class => $this->createFileSystemMock(...),
+            FileSystem::class => $this->createFileSystemMock(...),
```
theofidry, line 183: Since MockedContainer is for testing purposes, I wonder if there would be an easy way to add an additional guard to prevent it from creating PHPUnit mock objects (since this is what happened here no?)

PR #2792, opened 2026-01-06
build: Add --no-progress to PHPUnit in CI environments

sanmai:
## Description

When CI environment variable is set, PHPUnit progress dots pollute the LLM context without adding value. This PR adds `--no-progress` conditionally.

How it works:
- `PHPUNIT_BIN` holds the binary path (for Make dependencies)
- `PHPUNIT` expands to include `--no-progress` when `CI` is non-empty
- `CI ?=` provides empty default to satisfy --warn-undefined-variables

Triggered by `CI=1` in environment, e.g. via `.claude/settings.json`

```json
$ grep -C4 CI ~/.claude/settings.json
{
  "env": {
    "GITHUB_ACTIONS": "1",
    "CI": "1"
  },
```
(With this setup Infection also runs without extra output when Claude starts it.)

### DIFF
--- a/src/TestFramework/Tracing/TraceProviderAdapterTracer.php
+++ b/src/TestFramework/Tracing/TraceProviderAdapterTracer.php
@@ -108,7 +108,8 @@ final class TraceProviderAdapterTracer implements Tracer
 
     private function tryToTrace(SplFileInfo $fileInfo): ?Trace
     {
-        $sourcePathname = $fileInfo->getPathname();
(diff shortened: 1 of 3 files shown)


theofidry:
do we really want no progress in the CI?

sanmai:
For GitHub Actions it does not matter if there's output or not (they will keep it chugging along for hours without any output), so there's no harm.

PR #2147, opened 2025-06-08
Remove dependencies that can be resolved automatically

sanmai:
This PR:

- [ ] Adds new feature ...
- [ ] Covered by tests
- [ ] Doc PR: https://github.com/infection/site/pull/XXX

Fixes https://github.com/infection/infection/issues/XXX

PR #2793, opened 2026-01-06
refactor: Remove dependencies that can be resolved automatically

sanmai:
## Description

Follow-up to #2147. Removes explicit factories for services that can be safely auto-wired by the DI container.

<details>
  <summary>Key Insight from Claude (c0871db3f00c77)</summary>

### Key Insight: Auto-wireable vs Correctly Auto-wireable

Auto-wiring resolves dependencies by **type**, not by **intent**. A service may technically auto-wire but produce incorrect behavior:

- Interface dependencies get wrong implementations
- Primitive arguments (strings, bools) can't be resolved
- Shared instances used where fresh ones were needed

### The Allowlist Model

Rather than auto-wire everything and exclude problems, we use an **allowlist** of services proven safe:

| Service | Why Safe |
|---------|----------|
| `StaticAnalysisToolFactory` | All concrete deps, no primitives |
| `MutantCodeFactory` | All concrete deps, no primitives |
| `MutantCodePrinter` | All concrete deps, no primitives |
| `FileMutationGenerator` | All concrete deps, no primitives |
| `SourceCollectorFactory` | All concrete deps, no primitives |

### Services That Cannot Be Auto-wired

| Service | Reason |
|---------|--------|
| `StaticAnalysisConfigLocator` | String path argument |
| `InitialStaticAnalysisRunConsoleLoggerSubscriberFactory` | Bool arguments |
| `SubscriberRegisterer` | Interface dependency (`EventDispatcher`) |
| `InitialTestsRunner` | Interface dependency (`EventDispatcher`) |
| `ConfigurationFactory` | Interface dependency (`CiDetectorInterface`) |

### Test Enforcement

- `test_service_can_be_auto_wired` - Allowlisted services must work without factories
- `test_factory_is_essential` - Factories must be necessary (prevents boilerplate creep)

See `src/Container/README.md` for full documentation.

</details>

## Checklist

- [x] Appropriate labels applied

## Related issues

- #2147 - Original auto-wiring PR


bin/find-redundant-deps.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+# Find redundant Container dependencies by testing each removal
+
+cd "$(dirname "$0")/.."
```
theofidry: instead of being a bash script this could be a PHP script?

I was thinking such script would be useful to have in the CI or to invoke regularly, but the current approach (try to remove + run tests, IIUC), does seem a bit heavy handed for that

sanmai: We had some kind of thing going on in ContainerTest, but I can't remember what was the problem with it. Let me see if I can fix it.

PR #2799, opened 2026-01-07
test: Standardize e2e test scripts

sanmai:
## Description

- Add explicit `cd "$(dirname "$0")"` to script directory for running from any location
- Add GOLDEN mode support for updating expected outputs
- Use unified diff (`-u`) for better error output
- Normalize quoting in dirname calls
- Print full path in e2e_tests runner for copy-paste convenience

### Explicit `cd` at script start

Previously, e2e scripts had to be run from their own directory:

```bash
cd tests/e2e/YieldValue && ./run_tests.bash bin/infection
```

Now scripts can be run from anywhere:

```bash
./tests/e2e/YieldValue/run_tests.bash bin/infection
```

This makes it easier to iterate on tests without constantly switching directories.

### Full path in test output

The e2e_tests runner now prints the full path to failing tests, making it easy to copy-paste and re-run individual tests:

```
Testing /home/user/path/to/Infection/tests/e2e/YieldValue
```

### GOLDEN mode

To update expected outputs after intentional changes:

```bash
GOLDEN=1 ./tests/e2e/YieldValue/run_tests.bash bin/infection
```

### Unified diff

Changed from `diff --ignore-all-space` to `diff -u --ignore-all-space` for context-rich output when tests fail.

### DIFF
--- a/tests/e2e/Without_Git/run_tests.bash
+++ b/tests/e2e/Without_Git/run_tests.bash
@@ -29,6 +29,16 @@ sed -i "s/\"infection\/infection\": \"dev-master\"/\"infection\/infection\": \"d
 set -e pipefail
 
 rm -f composer.lock
+
+# For local testing on forks, detect the fork's remote URL and add it as a repository
+# This allows Composer to find branches that exist on the fork but not on the main repo
+if [ -z "$GITHUB_EVENT_NAME" ]; then
+  fork_url=$(git -C ../../.. remote get-url origin 2>/dev/null || true)
+  if [ -n "$fork_url" ]; then
+    composer config repositories.fork vcs "$fork_url" --no-interaction
+  fi
+fi
+
 composer install
 
 docker run -t -v "$PWD":/opt -w /opt php:8.4-alpine vendor/bin/infection --coverage=infection-coverage


tests/e2e/Adapter_Installer/run_tests.bash
```diff
@@ -6,7 +6,7 @@ if test ! -f "$(which expect)"; then
     exit 1;
 fi
 
-cd $(dirname $0)
+cd "$(dirname "$0")"
```
theofidry, line 9: could you explain what that change is for?

tests/e2e/Composer_Xdebug_Handler_Version_1/run_tests.bash
```diff
@@ -17,4 +19,8 @@ else
     php $INFECTION
 fi
 
-diff --ignore-all-space expected-output.txt infection.log
+if [ -n "$GOLDEN" ]; then
+    cp -v infection.log expected-output.txt
+fi
+
+diff -u --ignore-all-space expected-output.txt infection.log
```
theofidry, line 26: if we can help it, in scripts, let's use the long option form (here `--unified` I think?)

tests/e2e/Composer_Xdebug_Handler_Version_1/run_tests.bash
```diff
@@ -17,4 +19,8 @@ else
     php $INFECTION
 fi
 
-diff --ignore-all-space expected-output.txt infection.log
+if [ -n "$GOLDEN" ]; then
+    cp -v infection.log expected-output.txt
```
theofidry, line 23: the expected file already exists so isn't `-f` necessary?

tests/e2e_tests
theofidry, line 60: does `4` need to be specified? Isn't there a way to tell it to use as many parallel processes as possible?

sanmai: I don't think we necessarily want just that; Infection spawns its own children so we don't want to suffocate them.

tests/e2e/Adapter_Installer/run_tests.bash
sanmai: If you have spaces in the path to where you have Infection, it would not work.

tests/e2e/Composer_Xdebug_Handler_Version_1/run_tests.bash
sanmai: No `-f` here 🤔

tests/e2e/Composer_Xdebug_Handler_Version_1/run_tests.bash
sanmai: Well, I don't recall the full form for `-u` either so it sounds like it will be harder to read at a glance with the full option 🤔

PR #2801, opened 2026-01-07
ci: Re-add paratest for parallel E2E test execution

sanmai:
## Description

Re-introduces brianium/paratest ^7.8.4 which was removed in 4c196bb3 when upgrading to PHPUnit 10.5 (wasn't compatible at the time).

Paratest 7.8.4 now supports PHPUnit ^11.5.11 and enables parallel test execution for the E2E test suite.

Local benchmark (27 E2E tests):
- Sequential: ~5-6 minutes
- Parallel (32 processes): ~35 seconds

## Changes

Required upgrading Symfony components to satisfy paratest's dependencies.

## Checklist

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -92,7 +92,7 @@ jobs:
         run: composer update --no-interaction --prefer-dist --no-progress --prefer-lowest
 
       - name: Try to install paratest for parallel test execution
-        run: composer require --dev brianium/paratest:^7.8.4 -W --no-interaction --no-progress || true
+        run: composer require --dev brianium/paratest:^7.8.4 --no-interaction --no-progress || true
 
       - name: Cache E2E tests dependencies
         if: runner.os == 'Windows'


theofidry:
Wouldn't it make sense to introduce it definitely, i.e. for us locally too so that we could benefit from it for our tests? I don't know if it would help with the unit tests, but even just our integration tests are quite slow.

maks-rafalko (approved):

sanmai:
@theofidry it sounds like there's no version of paratest that supports both PHPUnit 11 AND Symfony 8

| Version          | PHPUnit  | Symfony      |
|------------------|----------|--------------|
| v7.8.4           | ^11.5.24 | ^6.4 \|\| ^7.x |
| v7.9.0 - v7.14.x | ^12.x    | ^6.4 \|\| ^7.x |
| v7.15.0+         | ^12.x    | ^7.3 \|\| ^8.0 |

theofidry:
Ha yeah now I remember, unfortunately paratest is quite strict/aggressive about the versions they support :/

.github/workflows/e2e-tests.yaml
```diff
@@ -91,6 +91,9 @@ jobs:
         if: ${{ matrix.dependencies == 'lowest' }}
         run: composer update --no-interaction --prefer-dist --no-progress --prefer-lowest
 
+      - name: Try to install paratest for parallel test execution
+        run: composer require --dev brianium/paratest:^7.8.4 -W --no-interaction --no-progress || true
```
sanmai:
```suggestion
        run: composer require --dev brianium/paratest:^7.8.4 --no-interaction --no-progress || true
```

.github/workflows/e2e-tests.yaml
```diff
@@ -92,7 +92,7 @@ jobs:
         run: composer update --no-interaction --prefer-dist --no-progress --prefer-lowest
 
       - name: Try to install paratest for parallel test execution
-        run: composer require --dev brianium/paratest:^7.4 -W --no-interaction --no-progress || true
+        run: composer require --dev brianium/paratest:^7.8.4 -W --no-interaction --no-progress || true
```
theofidry, line 95: I would suggest to use:

- `--with-all-dependencies` instead of `-W`
- `--minimal-changes`: do not update things that shouldn't be updated

Issue #2803, opened 2026-01-07
RFC: Collect and Visualize Long-term Performance Metrics


sanmai:
<sup>(Mostly authored by Claude under my guidance and ideation, so forgive it for some AI sloppiness.)</sup>

## Motivation

As Infection evolves, we lack visibility into performance trends over time. Questions like "Did the last refactoring improve mutation generation speed?" or "Is memory usage creeping up?" currently require manual benchmarking to answer.

This proposal outlines a system to automatically collect, store, and visualize performance metrics from Infection runs.

### External Metrics (GNU time)

Using `/usr/bin/time -f '<format>'` we can capture system-level metrics in JSON format:

| Metric | Format Specifier | Description |
|--------|------------------|-------------|
| `user_time_sec` | `%U` | User CPU time (seconds) |
| `system_time_sec` | `%S` | Kernel CPU time (seconds) |
| `wall_clock_sec` | `%e` | Elapsed real time (seconds) |
| `max_rss_kb` | `%M` | Peak memory usage (KB) |
| `cpu_percent` | `%P` | CPU utilization percentage |
| `voluntary_ctx_switches` | `%w` | Voluntary context switches |
| `involuntary_ctx_switches` | `%c` | Involuntary context switches |
| `major_page_faults` | `%F` | Major page faults (I/O required) |
| `minor_page_faults` | `%R` | Minor page faults (no I/O) |
| `fs_inputs` | `%I` | File system inputs |
| `fs_outputs` | `%O` | File system outputs |
| `exit_status` | `%x` | Exit code |

**Collection command:**
```bash
/usr/bin/time -f '{"user_time_sec":%U,"system_time_sec":%S,"wall_clock_sec":%e,"max_rss_kb":%M,"cpu_percent":"%P","voluntary_ctx_switches":%w,"involuntary_ctx_switches":%c,"exit_status":%x}' \
  php bin/infection ... 2> timing.json
```

### Internal Metrics (Infection)

Using `--logger-summary-json=summary.json` (lightweight, no per-mutation details, implementation pending):

| Metric | Source | Description |
|--------|--------|-------------|
| `totalMutantsCount` | MetricsCalculator | Total mutations generated |
| `killedCount` | MetricsCalculator | Killed by test framework |
| `notCoveredCount` | MetricsCalculator | Mutations in uncovered code |
| `escapedCount` | MetricsCalculator | Mutations that survived tests |
| `errorCount` | MetricsCalculator | Process errors encountered |
[...]

PR #2804, opened 2026-01-07
feat: Add performance metrics collection workflow

sanmai:
## Description

PoC for https://github.com/infection/infection/issues/2803

Adds a workflow that collects performance metrics from Infection runs and commits them to [infection-metrics](https://github.com/infection/infection-metrics).

:link: https://infection.github.io/infection-metrics/

## Example Metrics Entry

```json
{
  "user_time_sec": 450.71,
  "system_time_sec": 409.79,
  "wall_clock_sec": 277,
  "max_rss_kb": 5176336,
  "cpu_percent": "310%",
  "exit_status": 0,
  "timestamp": "2026-01-11T11:30:39Z",
  "commit_date": "2026-01-11T20:20:50+09:00",
  "git_sha": "5d3f85371c97...",
  "git_ref": "refs/heads/feat/metrics-gh-pages-branch-poc",
  "trigger": "workflow_dispatch",
  "php_version": "8.2.30",
  "config": { "name": "baseline", "args": "--no-progress --threads=max" },
  "runner": {
    "os": "ubuntu-latest",
    "cpu_model": "AMD EPYC 7763 64-Core Processor",
    "cpu_cores": 4,
    "total_memory_kb": 16379472
  },
  "mutations": {
    "total": 5688,
    "killed": 5050,
    "msi": 91.6
  },
  "derived": {
    "wall_clock_per_core": 69.25,
    "cpu_time_total": 860.5,
    "wall_clock_per_mutation": 0.0487,
    "user_time_per_mutation": 0.0792,
    "memory_per_mutation_kb": 910.05
  }
}
```

(Mutation numbers collection depends on #2825 so we'll only see averages from the next release.)

## Key Metrics

| Metric | Value | Description |
|--------|-------|-------------|
| `wall_clock_sec` | 277s (~4.6 min) | Total elapsed time |
| `cpu_time_total` | 860s | User + system CPU time |
| `mutations.total` | 5688 | Total mutations generated |
| `wall_clock_per_mutation` | 0.049s | Time per mutation (normalized) |

## Optimizations

- **OPcache file cache** - Runs `setup-php` twice: first with pcov for coverage, then with `coverage: none` + OPcache for mutation testing (~39% faster)
- **Merged jobs** - Single job does coverage, mutation testing, and commit (no artifact overhead)
- **External script** - `prepare-metrics.php` lives in metrics repo

## Backward Compatibility

Uses `git grep` to detect `--logger-summary-json` support:
- New releases: full metrics including mutation counts
- Old tags: timing/memory metrics only (no `mutations` field)

## Testing the Visualization

https://infection.github.io/infection-metrics/

Or, if you want to view them locally, use:
```bash
cd infection-metrics
php -S localhost:8000
open http://localhost:8000
```

### DIFF
--- a/.github/workflows/mt.yaml
+++ b/.github/workflows/mt.yaml
@@ -74,8 +74,8 @@ jobs:
       - name: Collect coverage report
         run: |
           php vendor/phpunit/phpunit/phpunit --stop-on-failure \
-            --coverage-xml=var/logs/coverage-xml \
-            --log-junit=var/logs/junit.xml
+            --coverage-xml=var/coverage/coverage-xml \
+            --log-junit=var/coverage/junit.xml
 
       - name: Setup PHP for mutation testing
         uses: shivammathur/setup-php@44454db4f0199b8b9685a5d763dc37cbf79108e1 # 2.36.0
@@ -97,7 +97,7 @@ jobs:
             --skip-initial-tests \
             --min-msi=$MIN_MSI \
             --min-covered-msi=$MIN_COVERED_MSI \
-            --coverage=var/logs \
+            --coverage=var/coverage \
             --log-verbosity=none \
             --no-interaction \
             --no-progress
--- a/tests/phpunit/TestFramework/Tracing/PHPUnitCoverageTracerTest.php
+++ b/tests/phpunit/TestFramework/Tracing/PHPUnitCoverageTracerTest.php
@@ -58,6 +58,7 @@ use Infection\Tests\TestFramework\Tracing\Trace\SyntheticTrace;
 use Infection\Tests\TestFramework\Tracing\Trace\TraceAssertion;
 use PHPUnit\Framework\Attributes\CoversNothing;
 use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\TestCase;
 use function Safe\realpath;
 use SplFileInfo;
@@ -65,6 +66,7 @@ use function sprintf;
 use Symfony\Component\Filesystem\Path;
 use Symfony\Component\Process\Process;
 
+#[Group('integration')]
 #[CoversNothing]
 final class PHPUnitCoverageTracerTest extends TestCase
 {


.github/workflows/perf-metrics.yaml
```diff
@@ -0,0 +1,202 @@
...
+# yamllint disable rule:braces
+
+name: Performance Metrics
+
+on:
+  push:
+    branches:
+      - main
+      - master
+      - feat/metrics-gh-pages-branch-poc  # PoC branch for testing
```
sanmai:
```suggestion
```

sanmai:
Screenshot From 2026-01-07 16-42-26

.github/workflows/perf-metrics.yaml
```diff
@@ -189,11 +189,17 @@ jobs:
       (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
 
     steps:
-      - name: Checkout metrics branch
-        uses: actions/checkout@v4
+      - name: Setup SSH key for metrics repo
+        uses: webfactory/ssh-agent@v0.9.0
```
sanmai: the checkout action can use the ssh key directly

.github/workflows/perf-metrics.yaml
```diff
@@ -0,0 +1,212 @@
...
+
+jobs:
+  check:
+    name: Check if should run
+    runs-on: ubuntu-latest
+    outputs:
+      skip: ${{ steps.check.outputs.skip }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
theofidry: actions should be pinned (you can use `pinact run` with [panact](https://github.com/suzuki-shunsuke/pinact)). Note that sometimes the output is not perfect so you may need a review of it; e.g. it tends to mark the pin of `shivammathur/setup-php@v2` as `master` instead of the tagged version for some reasons

.github/workflows/perf-metrics.yaml
```diff
@@ -0,0 +1,212 @@
...
+        with:
+          fetch-depth: 2
+
+      - name: Check commit message
```
theofidry: I don't quite get how this works

.github/workflows/perf-metrics.yaml
```diff
@@ -0,0 +1,212 @@
...
+            composer-${{ runner.os }}-
+            composer-
+
+      - name: Install dependencies
+        run: composer install --no-interaction --prefer-dist --no-progress
+
+      - name: Collect coverage report
+        run: |
+          php vendor/phpunit/phpunit/phpunit --stop-on-failure \
+            --coverage-xml=var/logs/coverage-xml \
```
theofidry:
```suggestion
            --coverage-xml=var/coverage/coverage-xml \
```

? (would need adjusting in multiple places)

.github/workflows/perf-metrics.yaml
```diff
@@ -0,0 +1,212 @@
...
+            --log-junit=var/logs/junit.xml
+
+      - name: Setup PHP for mutation testing
```
theofidry: it is described in the PR description but I would add a comment here, something like:


```suggestion
      # Re-configure PHP without pcov and with opcache
      - name: Setup PHP for mutation testing
```

.github/workflows/perf-metrics.yaml
```diff
@@ -0,0 +1,212 @@
...
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ env.PHP_VERSION }}
+          coverage: none
+          ini-values: opcache.enable_cli=1, opcache.file_cache=/tmp/opcache, opcache.file_cache_only=1
```
theofidry, line 86: I feel like I've asked already. I don't remember if it was somewhere else or if it was when I started to review this PR and dropped it as it was still a draft, but I don't remember the answer at least.

Why do we need to configure the opcache file and explicitly create it? For caching it?

.github/workflows/perf-metrics.yaml
```diff
@@ -0,0 +1,212 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Create opcache directory
+        run: mkdir -p /tmp/opcache
+
+      - name: Collect runner specs
+        id: runner-specs
+        run: |
```
theofidry, line 95: maybe could be put in a `.ci/collect-runner-specs.sh` script?

I'm thinking if we need to adjust it, at least it would be easier to execute.

And whilst we're not too hot into testing bash scripts, it is something we could eventually do (I had a reasonable experience so far with https://github.com/bats-core/bats-core)

.github/workflows/perf-metrics.yaml
```diff
@@ -0,0 +1,212 @@
...
+          echo "cpu_model=$(grep -m1 'model name' /proc/cpuinfo | cut -d: -f2 | xargs)" >> $GITHUB_OUTPUT
+          echo "cpu_cores=$(nproc)" >> $GITHUB_OUTPUT
+          echo "cpu_arch=$(uname -m)" >> $GITHUB_OUTPUT
+          echo "total_memory_kb=$(grep MemTotal /proc/meminfo | awk '{print $2}')" >> $GITHUB_OUTPUT
+
+      - name: Check for summary logger support
+        id: feature-check
+        run: |
+          if git grep -q 'logger-summary-json' -- src/; then
```
theofidry, line 104: maybe I misunderstand what is done here, but rather than `git grep` wouldn't it be better to do `infection run --help | grep <whatever>` instead?

.github/workflows/perf-metrics.yaml
```diff
@@ -0,0 +1,212 @@
...
+
+      - name: Checkout metrics repository
+        if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
+        uses: actions/checkout@v4
+        with:
+          repository: infection/infection-metrics
+          ssh-key: ${{ secrets.INFECTION_METRICS_SSHKEY }}
+          path: metrics-repo
+
+      - name: Prepare and commit metrics
```
theofidry, line 137: I suggest to split this job.

Whilst it would mean more overhead, the job just did a pretty expensive computation to get valuable artefacts. Now if the commit/push just fails, we'll end up wasting all of this computing.

Instead, if we store the artefacts with the job (which IMO has less chance to go wrong), we could have the last "publishing" job as a job we can manually trigger and manually trigger it if the push failed for some reasons (at the end of the day, it's still a connection that can be interrupted, or the SSH key is wrong and had to be rotated or smth).

theofidry:
I'd add I'm still a bit unsure how certain metrics are meant to be read/interpreted, but I guess this can be clarified separately either in the RFC or the metrics dashboard

.github/workflows/perf-metrics.yaml
sanmai: We use `var/logs` in other workflows, so I suppose we can swap these in one go in all places.

.github/workflows/perf-metrics.yaml
sanmai: Done.

.github/workflows/perf-metrics.yaml
sanmai:
```suggestion
      - name: Re-configure PHP for mutation testing without pcov and with opcache
```

.github/workflows/perf-metrics.yaml
sanmai: This was in https://github.com/infection/infection/pull/2810.

Infection launches a bunch of processes, and they rely on the pre-compiled bytecode in the file cache. So we don't have to recompile the same PHP code in each sub process. We could've used shared memory, but it is not a thing for CLI, unfortunately.

.github/workflows/perf-metrics.yaml
sanmai: I'm kinda torn; one one hand I'd put these into scripts and subject to `shellcheck` for a good measure; on another hand, it is so much easier to read when it is all in one place.

.github/workflows/perf-metrics.yaml
sanmai: I wasn't sure it will work when [back-filling the stats for older versions](https://infection.github.io/infection-metrics/) (have we had it changed?), and I went with the safest approach possible. Also, starting a bunch of backfill workflows and finding they failed is not very pleasant, so here's another argument to make it as simple as possible. Make sense?

.github/workflows/perf-metrics.yaml
sanmai: You have a point. I'm hesitant to follow the approach since we are in very early stages, and optimizing this part before it is obvious it is clearly broken could be detrimental to anything else we'd want to do.

Besides, it is [only 4:30 to run the whole thing](https://infection.github.io/infection-metrics/) - not that expensive, from what I can tell.

.github/workflows/perf-metrics.yaml
sanmai: I think there's a way to negate the need of the file OPcache: if we could pre-read all the code Infecition runs (a simple class exists check could work) then all forks will inherit this compilation and would not need to rely on the cache.

.github/workflows/perf-metrics.yaml
sanmai: OK, that's done here.

.github/workflows/perf-metrics.yaml
sanmai: Removed; it was a hangover from a previous attempt.

.github/workflows/perf-metrics.yaml
sanmai: Moved this right into the title

sanmai:
<details>
  <summary>backfill-metrics.sh</summary>

```bash
#!/bin/bash
# backfill-metrics.sh - run locally, not committed
# Triggers perf-metrics workflow for all PHP 8.2+ releases

# Configuration
REPO="${REPO:-sanmai/infection}"  # Use sanmai/infection for PR branch
WORKFLOW_REF="${WORKFLOW_REF:-feat/metrics-gh-pages-branch-poc}"  # Branch with workflow file

TAGS=(
  0.29.10 0.29.11 0.29.12 0.29.13 0.29.14
  0.30.0 0.30.1 0.30.2 0.30.3
  0.31.0 0.31.1 0.31.2 0.31.3 0.31.4 0.31.5 0.31.6 0.31.7 0.31.8 0.31.9
  0.32.0 0.32.1
)

mapfile -t TAGS < <(printf '%s\n' "${TAGS[@]}" | shuf)

echo "Repository: $REPO"
echo "Workflow branch: $WORKFLOW_REF"
echo "Tags to process: ${#TAGS[@]}"
echo

for tag in "${TAGS[@]}"; do
  echo "=== Running metrics for $tag ==="

  # Trigger workflow:
  # --ref = branch containing the workflow file
  # -F ref = input parameter for checkout target
  gh workflow run perf-metrics.yaml \
    --repo "$REPO" \
    --ref "$WORKFLOW_REF" \
    -F ref="$tag" || { echo "Failed to trigger $tag, skipping..."; continue; }

  # Wait for run to be registered
  sleep 3

  # Find the run ID and wait for completion
  RUN_ID=$(gh run list --repo "$REPO" --workflow=perf-metrics.yaml --limit=1 --json databaseId -q '.[0].databaseId')

  echo -n "Run $RUN_ID: "
  if gh run watch "$RUN_ID" --repo "$REPO" --exit-status > /dev/null 2>&1; then
    echo "success"
  else
    echo "failed"
  fi
  echo "=== Completed $tag ==="
  echo
done

echo "All tags processed!"
echo

# Also run on the current branch
echo "=== Running metrics for current branch ($WORKFLOW_REF) ==="
gh workflow run perf-metrics.yaml \
  --repo "$REPO" \
  --ref "$WORKFLOW_REF" || { echo "Failed to trigger current branch"; }

sleep 3
RUN_ID=$(gh run list --repo "$REPO" --workflow=perf-metrics.yaml --limit=1 --json databaseId -q '.[0].databaseId')
echo -n "Run $RUN_ID: "
if gh run watch "$RUN_ID" --repo "$REPO" --exit-status > /dev/null 2>&1; then
  echo "success"
else
  echo "failed"
fi
echo "=== Completed current branch ==="
echo
echo "Next steps:"
echo "1. cd infection-metrics && git pull"
echo "2. jq -s 'sort_by(.commit_date)' metrics.jsonl | jq -c '.[]' > sorted.jsonl && mv sorted.jsonl metrics.jsonl"
echo "3. git add metrics.jsonl && git commit -m 'Sort metrics by commit date' && git push"
```

</details>

PR #2806, opened 2026-01-07
ci: Enable PHP OPcache and JIT for benchmark workflow 

sanmai:
## Summary

- Disable Xdebug via `coverage: none` to remove any possible profiling overhead
- Enable OPcache for CLI with `opcache.enable_cli=1`
- Enable JIT tracing mode with 100M buffer for CPU-intensive workloads

This reduces the time necessary to run benchmarks from 6 minutes to 2 minutes just so that y'all don't have to wait for it to finish last.

| Job        | Before (avg) | After (JIT) | Improvement |
|------------|--------------|-------------|-------------|
| Tests      | 15 sec       | 9 sec       | 40% faster   |
| Benchmarks | 6m 25s       | 2m 9s       | 67% faster :exploding_head:  |

## Why

The benchmark workflow was running without PHP performance optimizations. PHPBench performs CPU-intensive AST parsing and mutation generation - exactly the workload where JIT provides significant improvements (30-50% for CPU-bound tasks according to PHP documentation).

## Changes

Both `tests` and `benchmarks` jobs now use:
```yaml
coverage: none
ini-values: opcache.enable_cli=1, opcache.jit=tracing, opcache.jit_buffer_size=100M
```

## References

- PHP 8.4 JIT changes: https://php.watch/versions/8.4/opcache-jit-ini-default-changes
- JIT setup guide: https://stitcher.io/blog/php-8-jit-setup
- setup-php action: https://github.com/shivammathur/setup-php

## Checklist

### DIFF
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -28,6 +28,8 @@ jobs:
         with:
           php-version: '8.4'
           tools: composer
+          coverage: none
+          ini-values: opcache.enable_cli=1, opcache.jit=tracing, opcache.jit_buffer_size=100M
 
       - name: Install dependencies
         uses: ramsey/composer-install@3cf229dc2919194e9e36783941438d17239e8520 # 3.1.1
@@ -49,6 +51,8 @@ jobs:
             with:
                 php-version: '8.4'
                 tools: composer
+                coverage: none
+                ini-values: opcache.enable_cli=1, opcache.jit=tracing, opcache.jit_buffer_size=100M
 
           - name: Install dependencies
             uses: ramsey/composer-install@3cf229dc2919194e9e36783941438d17239e8520 # 3.1.1


theofidry:
but if you enable JIT there doesn't it completely skews the benchmark?

(I'm not super well informed on that subject, so maybe it is a dumb concern).

I also found out you can configure workflows to be manually triggered so we could make the benchmark manually triggerable but disable it for regular runs (still leave the test for checking that it executes).

This way we save on computing, we can still get the results for both a PR and main branch but would require to trigger them by hand.

sanmai:
JIT, as any other heavy optimization, may indeed mask some critical inefficiencies by reducing their _measurement scale_ (so a minute-long problem will look only a few seconds worse with JIT). Still, it won't hide brazen problems if we O(n<sup>2</sup>)-botch the mutation logic entirely. So we lose some sensitivity to micro-regressions, but we gain minutes of yours and mine time on every single build.

I don't have any hard feelings about getting these to run on-demand - but since we reduced the time they take to the levels, from what I can tell, where human-operator is not affected (you and I did not need to wait extra for the benchmark to finish), it should not be an issue to keep them working. If you want to take off JIT and run the benchmark on-demand and on-merge only, go ahead.

theofidry:
To clarify: my main concern here is to be able to assess the impact of a change.

Let' say we introduce a patch, there is no notable difference with JIT (because super optimized for it), but if you do not have JIT then it actually causes a 30% increase in time spent compare to before the patch.

I do not know if that can really happen, maybe it's non-sensical. But my concern is to not be able to notice such cases; especially considering that JIT not often enabled (at least I come across multiple cases of it not being enabled regularly)

PR #2810, opened 2026-01-08
ci: Enable OPcache for mutation testing workflow

sanmai:
## Summary

- Split mutation testing into two stages: coverage collection and mutation testing
- Enable OPcache CLI in the mutation testing stage
- Use GitHub artifacts to pass coverage between jobs

## Why

The mutation testing stage doesn't need pcov/xdebug - it only consumes pre-collected coverage data. By splitting the workflow:

1. **Coverage job**: Uses pcov to collect coverage (OPcache incompatible)
2. **Mutation testing job**: Uses `coverage: none` + OPcache + file cache

Based on #2807 benchmarks, JIT provides 40-67% performance improvement for CPU-intensive workloads like mutation testing. We'll see how [file-backed OPcache affects this workload](https://github.com/infection/infection/pull/2810#issuecomment-3723471260) in just a moment.

#### mt.yaml

| Configuration               | Infection Time | Total Time | vs Baseline |
|-----------------------------|----------------|------------|-------------|
| Baseline (pcov, no opcache) | 7m 27s         | ~8m 14s    | -           |
| Split + JIT                 | 11m 25s        | ~12m 31s   | +52% slower |
| Split + file_cache          | 4m 58s         | ~5m 58s    | -33% faster |
| Split + file_cache (run 2)  | 5m 4s          | ~6m 4s     | -31% faster |

#### mt-annotations.yaml

| Configuration         | Total Time | vs Baseline                       |
|-----------------------|------------|-----------------------------------|
| Baseline (single job) | ~40-50s    | -                                 |
| Split + file_cache    | ~1m 23s    | slower (expected - full coverage) |

(going to revert this one)

## How we did it

Coverage is passed between jobs via `actions/upload-artifact` and `actions/download-artifact`.

## Affected workflows

- `mt.yaml` - Full mutation testing on push to main/master

## Checklist

### DIFF
--- a/.github/workflows/mt.yaml
+++ b/.github/workflows/mt.yaml
@@ -13,7 +13,6 @@
 name: Mutation Testing
 
 on:
-  pull_request:
   push:
     branches:
       - main


.github/workflows/mt.yaml
```diff
@@ -13,6 +13,7 @@
 name: Mutation Testing
 
 on:
+  pull_request:
```
sanmai:
```suggestion
```
Planning to remove it after testing, before merging.

sanmai:
## The Technical Barriers

1. JIT and file_cache are mutually exclusive (https://github.com/php/php-src/issues/17770)
    - When JIT is enabled, opcache won't write to file cache
    - There's explicit code blocking this: if (JIT_G(on)) return FAILURE;
    - Even if fixed, file cache only stores bytecode, not JIT machine code
2. Shared memory doesn't work across CLI processes on Unix (https://www.npopov.com/2021/10/13/How-opcache-works.html)
    - Only Windows supports shared opcache between unrelated CLI processes
    - On Unix, shared memory is only inherited through fork(), not across separate invocations
3. JIT machine code is process-local (https://php.watch/articles/jit-in-depth)
    - JIT compiles to machine code with process-specific memory addresses (ASLR)
    - Cannot be serialized and shared between processes
    - Each process must JIT compile its own code
4. PHPUnit/testing workloads don't benefit from JIT (https://stitcher.io/blog/jit-in-real-life-web-applications)
    > CLI applications [...] network and file-intensive applications such as Composer and PHPUnit will not likely see a performance gain"

sanmai:
I was expecting more, but that's something.

## Performance Comparison

### Baseline: Single Job (pcov, no opcache optimization)

| Run | Commit | Total Time |
|-----|--------|------------|
| [20796664256](https://github.com/infection/infection/actions/runs/20796664256) | [Conductor] Update webmozart/assert to 2.1.1 (#2808) | **8m 14s** |
| [20783099472](https://github.com/infection/infection/actions/runs/20783099472) | ci: Enable PHP OPcache and JIT for several workflows (#2807) | **8m 20s** |
| [20783017208](https://github.com/infection/infection/actions/runs/20783017208) | ci: Enable PHP OPcache and JIT for benchmark workflow (#2806) | **7m 02s** |
| | **Average** | **7m 52s** |

### New: Split Workflow with `opcache.file_cache`

| Run | Coverage | Infection | Total |
|-----|----------|-----------|-------|
| [20811795539](https://github.com/infection/infection/actions/runs/20811795539) | 44s | 5m 04s | **5m 48s** |
| [20811446089](https://github.com/infection/infection/actions/runs/20811446089) | 49s | 5m 00s | **5m 49s** |
| [20810610389](https://github.com/infection/infection/actions/runs/20810610389) | 44s | 4m 58s | **5m 42s** |
| [20810359741](https://github.com/infection/infection/actions/runs/20810359741) | 50s | 5m 04s | **5m 54s** |
| | **Average** | | **5m 48s** |

### Failed Experiment: JIT (for reference)

| Run | Infection | Notes |
|-----|-----------|-------|
| [20808571399](https://github.com/infection/infection/actions/runs/20808571399) | 11m 25s | JIT compilation overhead hurts subprocess-heavy workloads |

### Summary

| Configuration | Avg Time | vs Baseline |
|--------------|----------|-------------|
| Baseline (single job, pcov) | 7m 52s | - |
| Split + JIT | ~11m 30s | +46% slower |
| **Split + file_cache** | **5m 48s** | **-26% faster** |

The `opcache.file_cache` optimization allows PHP subprocesses to read pre-compiled bytecode from disk, saving ~2 minutes per mutation testing run.

**P.S.:**

I suspect that cutting those 10 timeout runaway mutations (each taking up to 25 seconds, ~4 minutes) we can reduce the run-times further.

.github/workflows/mt.yaml
```diff
@@ -73,6 +75,60 @@ jobs:
...
+        with:
+          php-version: ${{ env.PHP_VERSION }}
+          coverage: none
+          tools: composer
+          ini-values: opcache.enable_cli=1, opcache.file_cache=/tmp/opcache, opcache.file_cache_only=1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Create opcache directory
+        run: mkdir -p /tmp/opcache
```
theofidry, line 106: Why do we need to specify a specific file for the opcache cache? It would make sense if we were to cache it but that doesn't appear to be the case here?

.github/workflows/mt.yaml
```diff
@@ -73,6 +75,60 @@ jobs:
...
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: composer-${{ runner.os }}-${{ env.PHP_VERSION }}-${{ hashFiles('composer.*') }}
+          restore-keys: |
+            composer-${{ runner.os }}-${{ env.PHP_VERSION }}-
+            composer-${{ runner.os }}-
+            composer-
+
+      - name: Install dependencies
+        run: composer install --no-interaction --prefer-dist --no-progress
```
theofidry, line 123: it's just me rambling but I still think it would make sense ot use https://github.com/ramsey/composer-install at some point...

.github/workflows/mt.yaml
```diff
@@ -73,6 +75,60 @@ jobs:
             --coverage-xml=var/logs/coverage-xml \
             --log-junit=var/logs/junit.xml
 
+      - name: Upload coverage
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: coverage
+          path: var/logs/
```
theofidry, line 81:
```suggestion
          path: var/coverage/
```

? (the other job needs to be adjusted too)

theofidry: Ah that's what it was before, nvm then

.github/workflows/mt.yaml
sanmai: Infection launches a bunch of processes, and they rely on the pre-compiled bytecode in the file cache.

Shared memory is not a thing for CLI, unfortunately.

maks-rafalko:
👍 

@staabm  just wanted to mention you as I think it can be interesting for you as a nice example of how MT can be improved, as you use it for PHPStan

.github/workflows/mt.yaml
```diff
@@ -73,6 +74,60 @@ jobs:
...
+          retention-days: 1
+
+  mutation-testing:
```
staabm, line 84: If you don't separate in 2 jobs but just invoke setup-php again in the previous job - it should save you the overhead of bootstrapping job 2

sanmai: Wow, that's awesome! Let me try this.

Issue #2797, opened 2026-01-06
RFC: Rich AST for Scalable Mutation Testing


theofidry:
## Summary

This RFC proposes separating AST processing into two distinct passes:

1. An enrichment pass that annotates nodes with eligibility, coverage, and context information.
2. A mutation pass that only generates mutations for eligible nodes.

This separation simplifies mutator logic, improves testability, and enables future features like symbol targeting
and weighted mutations.

## Motivation

Infection relies _heavily_ on the AST. As such, it is not a surprise that we have a lot
of code related to it. As time went by, new mutators were introduced, new features
were added, and evaluating what is happening and testing it became much harder.

In the future, I would like to introduce various features which will have to touch the AST. For example:

- Being able to target a _symbol_. We can already target a few selected files via `--filter`. The natural
  evolution is to have a more granular control and to be able to directly target a class or method. Maybe
  we could go even further and be able to target a specific node/line/expression.
- We have discussed this in various issues: not all code is equally interesting. As such, it would
  be interesting to introduce a concept of weight, or what at Google they like to call "dry areas".
- Have different mutation selection strategies: pick the first 5 most impactful for a node/statement/symbol/file.
- Improve performances: avoid processing unnecessary code.
- Improve the mutators smartness: this may require being able to enrich further the AST, e.g. to
  add types to variables/parameters to be able to prevent unproductive mutations.

Those are a few ideas, and each deserves its own specifications and process to be discussed.

## Goals

I would like the AST processing to be a process we can more easily understand, analyze, test, and enrich.

This RFC scope is about consolidating our foundations to enable the features described in the section above.

## Current solution

Not too long ago (before https://github.com/infection/infection/pull/2296), we had only one traverse.

Still, even with this feature introduced, most of the work is done in the 2nd traverse. The responsibilities of the
current visitors are blurry. For example `ReflectionVisitor` adds all kinds of information, but is also
responsible (supposedly; that code doesn't seem to work anymore) for skipping the traversal of irrelevant nodes.

We also have `NodeMutationGenerator`, where half the code is about computing information to know if it
can mutate the node at all, and a good chunk of code to compute the information necessary to create a mutation.

The current tests are also severely lacking. Complex visitors such as `ReflectionVisitor` are extremely hard to
test, even more so in a readable way.

The integration tests that we have also test more the final output (the mutations) without really offering a way
to analyse how the different visitors interacted and altered the AST.

## Proposed Solution

Enrichment Pass Responsibilities Flowchart:

[...]

PR #2795, opened 2026-01-06
feat(ast): Mark nodes as eligible

theofidry:
## Description

This PR is the step 3) of the implementation plan of https://github.com/infection/infection/issues/2797. It:

- Moves all the enrichment logic to the pre-traverse, that was renamed "enrichment" as per the nomenclature.
- To this initial traverse a `LabelNodesAsEligibleVisitor` that annotates each node with an attribute to signal the node is eligible for being mutated.
- `NodeMutationGenerator` now checks if the node is eligible first.

This is the most minimal and least disruptive implementation. The plan is to then move more of the logic currently implemented in `NodeMutationGenerator` to visitors that will be part of the enrichment traverse.

## Related issues

- https://github.com/infection/infection/issues/2797

PR #2811, opened 2026-01-08
test: Re-introduce an internal mutator test case

theofidry:
## Description

This PR copies `BaseMutatorTestCase` and `SimpleMutationsCollectorVisitor` from `src/Testing` to `tests`, and update the tests to use those instead.

## Motivation

The issue I am encountering in #2795 is that I need to update two classes. More generally, I am changing the way the behaviour of the mutation execution: there is different visitors, in different orders, the conditions upon which the mutators create mutation changes.

I am not introduce crazy changes, in #2795 I am focusing on the minimum before tackling more features mentioned in #2797. Nonetheless, however small, there is still changes.

More importantly, in the future, even simply adding a new visitor can alter the behaviour and will result in a BC break. This is not sustainable: we **cannot** not update our tests because of this.

## Solution

The immediate short-term solution, which I opted for here, is to copy those classes. This way we can update them without impacting our users.

More long term, I think the API of those tests need to change in a way that we can:

- avoid complete code duplication
- communicate changes to the user and allow them to migrate from an old behaviour to a new one

I have no idea of the specifics at the moment, but if we want to those classes to be part of the public API, we need to be able to give a heads up to the consumers one way or another. Ideally, we would be able to allow them to migrate smoothly in one version at least.

## Changes

I renamed `BaseMutatorTestCase` to `MutatorTestCase` to offer a distinction, and I feel "base" is now redundant with the fact that we name it "TestCase")... But maybe it is silly to do so. Happy to revert. I otherwise only copied the code, there is no other changes.

### DIFF
--- a/devTools/phpstan-baseline.neon
+++ b/devTools/phpstan-baseline.neon
@@ -181,6 +181,42 @@ parameters:
 			path: ../src/Testing/BaseMutatorTestCase.php
 
 		-
+			message: '#^Method Infection\\Tests\\Mutator\\MutatorTestCase\:\:createMutator\(\) has parameter \$settings with no value type specified in iterable type array\.$#'
+			identifier: missingType.iterableValue
+			count: 1
+			path: ../tests/phpunit/Mutator/MutatorTestCase.php
+
+		-
+			message: '#^Method Infection\\Tests\\Mutator\\MutatorTestCase\:\:createMutator\(\) return type with generic interface Infection\\Mutator\\Mutator does not specify its types\: TNode$#'
+			identifier: missingType.generics
+			count: 1
+			path: ../tests/phpunit/Mutator/MutatorTestCase.php
+
+		-
+			message: '#^Method Infection\\Tests\\Mutator\\MutatorTestCase\:\:getMutationsFromCode\(\) has parameter \$settings with no value type specified in iterable type array\.$#'
+			identifier: missingType.iterableValue
+			count: 1
+			path: ../tests/phpunit/Mutator/MutatorTestCase.php
+
+		-
+			message: '#^Method Infection\\Tests\\Mutator\\MutatorTestCase\:\:mutate\(\) has parameter \$settings with no value type specified in iterable type array\.$#'
+			identifier: missingType.iterableValue
+			count: 1
+			path: ../tests/phpunit/Mutator/MutatorTestCase.php
+
+		-
+			message: '#^Parameter \#1 \$resolvedMutators of method Infection\\Mutator\\MutatorFactory\:\:create\(\) expects array\<class\-string\<Infection\\Mutator\\ConfigurableMutator\<PhpParser\\Node\>\>, array\<mixed\>\>, non\-empty\-array\<string, array\{settings\: array\}\> given\.$#'
+			identifier: argument.type
+			count: 1
+			path: ../tests/phpunit/Mutator/MutatorTestCase.php
+
+		-
+			message: '#^Property Infection\\Tests\\Mutator\\MutatorTestCase\:\:\$mutator with generic interface Infection\\Mutator\\Mutator does not specify its types\: TNode$#'
+			identifier: missingType.generics
+			count: 1
+			path: ../tests/phpunit/Mutator/MutatorTestCase.php
+
+		-
 			message: '#^Class Infection\\Testing\\SimpleMutation extends @final class Infection\\Mutation\\Mutation\.$#'
 			identifier: class.extendsFinalByPhpDoc
 			count: 1
--- a/tests/phpunit/Mutator/MutatorFixturesProvider.php
+++ b/tests/phpunit/Mutator/MutatorFixturesProvider.php
@@ -57,15 +57,15 @@ final class MutatorFixturesProvider
     private static array $testCaseFixtureDirMapping = [];
 
     /**
-     * @param class-string<MutatorTestCase> $testCaseClassName
+     * @param class-string $class
      */
-    public static function getFixtureFileContent(string $testCaseClassName, string $file): string
+    public static function getFixtureFileContent(string $class, string $file): string
     {
-        Assert::isAOf($testCaseClassName, MutatorTestCase::class);
+        Assert::isAOf($class, MutatorTestCase::class);
 
         return file_get_contents(sprintf(
             '%s/%s',
-            self::getTestCaseFixtureDir($testCaseClassName),
+            self::getTestCaseFixtureDir($class),
             $file,
         ));
     }


theofidry:
>As it's hard to think about it without a real example.

Sure, I can give several of them.

### non-parity between NodeMutationGenerator & SimpleMutationsCollectorVisitor (bug)

Right this instant, `NodeMutationGenerator` has:

```php
if (!$this->isOnFunctionSignature()
    && !$this->isInsideFunction()
) {
    return;
}
```

This is not reflected in `SimpleMutationsCollectorVisitor`. This means currently when testing mutators (ours, or consumers), they may generate mutations that would normally be skipped.

Fixing this is trivial (source code wise), adjusting the tests is a bit of faff but nothing difficult, but it is a behaviour change impacting the consumers: whey they update their tests can break.


### Adding a visitor

Tomorrow we figure out a nice performance optimization with a new visitor that skips the relevant pieces of code e.g. we don't traverse functions which we currently can't mutate (specifically, we found that out with https://github.com/infection/infection/pull/2780#discussion_r2659244531), likewise it can break someone's tests.


### More AST changes

With https://github.com/infection/infection/pull/2795 and follow up PR, I would like to change the condition upon which the nodes are skipped, the idea being to extract behaviour out of `NodeMutationGenerator` (more details in https://github.com/infection/infection/issues/2797).

This again, would mean a difference between `NodeMutationGenerator` and `SimpleMutationsCollectorVisitor`, but changing the latter is a BC break.

**TL:DR;** even if we do not break the code signatures, any change on our side of `NodeTraverserFactory`, (`NodeMutationGenerator` as we need to reflect it in `SimpleMutationsCollectorVisitor`), `BaseTestCase` can cause consumers tests to break on update.

maks-rafalko:
I think I'm not following.

> Fixing this is trivial (source code wise), adjusting the tests is a bit of faff but nothing difficult, but it is a behaviour change impacting the consumers: whey they update their tests can break.

this is a bug fix, not a BC break. When we fix something in Infection and some mutations disappear, or vice-versa appear - we don't call it a BC break, especially if it's a bugfix.

> tomorrow we figure out a nice performance optimization with a new visitor that skips the relevant pieces of code

also not sold here. When Infection (as a core) skips generating mutations after we add some optimization, it must (IMO) be reflected in the tests. Be it our test or our consumer's test - doesn't matter, because this particular new Infection version works differently.

If we introduce something in mutating logic that require a major version upgrade (imagine we are on stable 1.x), then anyway user will need to upgrade to 2.x and updated tests logic is allowed. So what I'm trying to say is that changes in the core directly affect tests.

> This again, would mean a difference between NodeMutationGenerator and SimpleMutationsCollectorVisitor, but changing the latter is a BC break.

any differences between our real core mutation logic and our tests mutation logic should be avoided. This is exactly what I partially did during format-preserving pretty printing implementation - I started to use the same services both in the code and in the tests so that we have _the same behavior_

maks-rafalko:
Counter-example:

image phpstan tests their custom mutator and it changes state A to a state B. But we merge this PR, and then change our core logic so that state A is changed to state C (instead of B). Then we release.

phpstan's tests will still work with A->B, while Infection core will actually do A->C, which doesn't make any sense as for me.

Our exposed "test framework" for mutators should be affected each time we change a core logic, otherwise I don't see a sense in it.

theofidry:
@maks-rafalko I actually agree with all of your points, and I would say we can justify this in multiple ways:

- This test is intended to reflect how infection generates mutations, a difference of behaviour is a bug.
- How many mutations / what mutators or the state of the AST at the time at which we generate mutations is not part of the backward compatibility promise.
- Infection is not stable.

We _can_ justify breakage. But this _will_ cause friction (tests will fail – requires immediate action).

Now let's imagine: this is a stable part, rarely ever changes, it's may cause minor breakage for 2-3 users. OK, a bit of communication ahead of time can do the trick.

My issue here is that:

- I already identified two non-trivial changes (the nodes not being skipped when they should and applying mutations when it shouldn't – both mentioned in my previous answer, I'm not talking of new ones)
- I'm changing things in https://github.com/infection/infection/pull/2795.
- I plan to change more things as part of https://github.com/infection/infection/issues/2797.
- I plan to change more things as part of the features mentioned in https://github.com/infection/infection/issues/2797.
- While implementing #2795, I was a bit unhappy for the unhappy path, i.e. when you get mutations not being generated or failing tests. It is a bit hard to debug and understand what is going on. I don't have a clear idea of what the DX should be there, neither if that will require any change at all to `BaseVisitorTestCase`, but I don't really want to ship smth that will break tests of a user in a non-expected way.

And the issue here is that I don't have a timeline for it. If we sorted everything in a week, pack everything in one release, it's a one time breaking, may cause some frictions, but it's fine.

Now let's say we spread all of those changes over the year, resulting in not one time where tests need to be adjusted but in every single of them.

Again, we can justify this, but I'm uncomfortable with the idea of deliberately causing that much friction.

<hr/>

My idea was: dissociate the 2 right this instant, figure things out, and then we can reunite the two with a friendly upgrade path or smth.

maks-rafalko:
I think I still miss the point, let me give an example.

Let's imagine our consumer has a custom mutator that do the following mutation:

```diff
- $a->doA();
+ $a->doB();
```

and our consumer uses base mutator class that we expose with the following test case:

```php
yield 'It mutates doA() to doB() in a condition' => [
	<<<'PHP'
		<?php

		if (false && $a->doA()) {}
		PHP
,
	<<<'PHP'
		<?php

		if (false && $a->doB()) {}
		PHP
,
];
```

so this test passes - mutant is correctly generated.

Then we decide to update our architecture (core of Infection) to mark such conditions after `false &&` as a dead code and skip mutations. In the following code

```php
if (false && $a->doA()) {}
```

we mark `$a->doA()` as an arid node / dead code and Infection skips mutation. We do a release.

Now, if we merge this PR, this is what will happen:

- our customer upgrades to the new version with detection of the dead code
- our customer's tests still pass(!), because they use the old copy of base mutator class. They expect nothing is broken and mutator still mutates the code
- they run Infection and it turns out that this is not the case - Infection actually doesn't mutate anything (in this particular example)

I really don't get why would we want it. In this imaginary case the test suite of our customer's custom mutator must fail, because the core engine is changed and it does affect how mutator behaves.

theofidry:
>I really don't get why would we want it.

We're on the same page, we (ultimately) do not want this. We disagree on the _when_.

>In this imaginary case the test suite of our customer's custom mutator must fail, because the core engine is changed and it does affect how mutator behaves.

Correct. And my idea would be to address this, but in a timely manner. Otherwise:

- Any update of infection can cause this.
- Supporting multiple versions of infection would be an absolute nightmare.

And regardless, I think we need to clarify our backward compatibility policy on this, because it would not make sense to have it under a regular Semver BC policy: if we add a new visitor/mutation, the infection execution will change, the same should go for the tests that reflect this execution.

Ok I don't know many users whom are using this extension point but we know at least PHPStan. Maybe @staabm could you shim in?

I guess the question is, is it a big problem for PHPStan if:

- Any release (that include patch versions, even if it was stable) of infection can break the mutators tests?
- That it may not be possible to test the mutators with different versions.

staabm:
> Ok I don't know many users whom are using this extension point but we know at least PHPStan.

find a list of users here: https://github.com/search?q=%22Infection%5CTesting%5CBaseMutatorTestCase%22+language%3APHP++NOT+is%3Afork&type=code

----

Regarding PHPStan: I don't mind to change a few classes on a future release as long as there is a equivalent replacement.
the custom mutators we have implemented are not for PHPStan Endusers .. just for PHPStan itself.. so thats fine.

----

regarding the change itself. maybe you could leave `Infection\Testing\BaseMutatorTestCase` as is (maybe deprecate it) and introduce a new class so Infection can use the new base-class for tests without affecting externals.

I don't see/understand the value in changing `Infection\Testing\BaseMutatorTestCase` atm for custom mutators implementors... changing it looks like a unnecessary change without benefit atm

maks-rafalko:
> find a list of users here: https://github.com/search?q=%22Infection%5CTesting%5CBaseMutatorTestCase%22+language%3APHP++NOT+is%3Afork&type=code

This one is a bit more filtered (Infection is excluded): https://github.com/search?q=%22Infection%5CTesting%5CBaseMutatorTestCase%22+language%3APHP++NOT+is%3Afork+NOT+org%3Ainfection&type=code

It boils down to PHPStan and [DwiHandoyo/mutant-killer-alpha-2](https://github.com/DwiHandoyo/mutant-killer-alpha-2), so 2 consumers (public).

> We disagree on the _when_.

that's also important, cause as you say it can take year(s). 

Anyway, my take is: we need to treat custom mutators' tests the same way as our internal tests. If core is changed in such a way that it affects AST mutation process and it affects **our tests**, then this must affect consumers' custom mutator tests. There is no difference between these tests - they work with the same engine. Custom mutator can be an internal one, as well as the opposite - some mutator can become a custom.

@theofidry do you have any examples how PHPStan's tests suite can be affected? Like literally what would be a diff?

theofidry:
>do you have any examples how PHPStan's tests suite can be affected? Like literally what would be a diff?

I do not at the moment.

However, I expect that when this happens, we will have at least one of our test affected, so that will be a diff that is available.

OK then I'll just proceed and maybe add a extra warning/mention for how it can affect users.

theofidry:
>I don't see/understand the value in changing Infection\Testing\BaseMutatorTestCase atm for custom mutators implementors... changing it looks like a unnecessary change without benefit atm

Do you mean as you to extend its behaviour as a user?

maks-rafalko:
> OK then I'll just proceed and maybe add a extra warning/mention for how it can affect users.

then, we can get back to this PR at the moment when you will have a real diff, to discuss based on it and see if and how PHPStan's testsuite can be affected

theofidry:
I think the current alternative is okish. Let's close this for now.

PR #2812, opened 2026-01-08
refactor: Remove dependency on FinderSplFileInfo

theofidry:
## Description

Thanks to #2788, we do not need to depend explicitly on `Symfony\Component\Finder\SplFileInfo`. We still use it of course as we are using the Symfony Finder, but we do not need to let that implementation detail of `BasicSourceFileInfo` leak to the rest of the codebase.

## Changes

- Use `SplFileInfo` instead of `Symfony\Component\Finder\SplFileInfo` where possible.
- Leverage `MockSplFileInfo` in a few places where we were mocking `SplFileInfo` instead.
- Remove the `Finder\MockSplFileInfo` in favour of `MockSplFileInfo`.
- Simplify `MockSplFileInfo`: we have been using only the pathname and realpath in practice, so we can simply pass those two properties instead:

```php
// before
new MockSplFileInfo(['name' => '...', 'realPath' => '...']);

// after:
new MockSplFileInfo('file', 'realPath/file');
```


theofidry:
@sanmai if you have a moment

maks-rafalko (approved):
techincally, the code is ok.

but I struggle to understand the reason of doing it.

What was wrong with finder's SplFileInfo? Why do we need all these changes?

theofidry:
>What was wrong with finder's SplFileInfo? Why do we need all these changes?

Ha, well this one is a long story 😄 

So disclaimer, I have a bit of a bias against `symfony/finder`: I always find that it behaves a bit unexpectedly, the DX seems good at first, but it's hard to really nail the behaviour you want and properly testing it requires a lot of efforts.

Given the number of issues we've had on this, and that I've had on PHP-Scoper & Box too, I don't think I'm alone there... We can refer to #2770 as the most recent one, but there is more.

The finder `SplFileInfo` is also a bit more complicated can than traditional `SplFileInfo`, so it is always more code when we want to use it (there is 2 more parameters), it is more code when we want to mock it (3 more methods to cover) and it can be quite confusing at times what those values are for and if used at all. So I think if we can avoid using it, it is better as it is less friction.

Another angle is architecture: we have a hard dependency on `symfony/finder` for the collection of source code. If tomorrow we decide to ditch it because we found something better, a few month ago this would have been straight up impossible: it is a dependency that leaked deep, everywhere. We putted it in the trace and everything that consumed a trace that needed something about the source file ended up using it, even if it did not really need the `Finder\SplFileInfo` and could have used the simpler `SplFileInfo`. In general, this kind of things tend to ring a bell in my mind and tell me there is potentially a design issue somewhere (and IMO there was – as some RFC highlighted, e.g. https://github.com/infection/infection/issues/2380).

In the past months there was a conscious effort on my end to limit its usage and use the native `SplFileInfo` when we can help it. I didn't go with the goal to remove it – this was not the drive for some re-structural work, but it certainly was always in the back of my mind.

This PR is the conclusion of this background work. It is not a big change, but it fixes a leak, simplifies and removes some bits of code.

sidz:
**I'm not against this PR** especially as Theo has already put some efforts into it but this story sounds like a personal choice rather than something else.

> Another angle is architecture: we have a hard dependency on symfony/finder for the collection of source code. If tomorrow we decide to ditch it because we found something better, a few month ago this would have been straight up impossible: 

It is possible to tell the same about any Infection dependency, for example what if tomorrow we will decide to get rid of symfony/process, symfony/console, etc. We obviously will have to touch the half of the project.

sanmai:
I approved because it looks like a no-brainer to me because we don't use Symfony's SplFileInfo for the features it has, just incidentally. If we were using some features of specifically Symfony's SplFileInfo, it would have been a different story.

theofidry:
>It is possible to tell the same about any Infection dependency, for example what if tomorrow we will decide to get rid of symfony/process, symfony/console, etc. We obviously will have to touch the half of the project.

Absolutely! It is a question of (conscious) coupling and to what parts of your application it spreads to. There is always a fine line, and this is a mistake often done by inexperienced people being a bit too enthusiastic about hexagonal/onion architecture. You do not want to abstract way every single piece of third-party, otherwise you end up with an unbearable mess of all kind of abstraction.

>but this story sounds like a personal choice rather than something else.

The part of `symfony/finder` very much is. I'm biased against it and personally use it because I didn't find better. But that bias aside, the real changes happened for other reasons (which I hope are detailed and justified enough in the RFCs). This PR is more of a side-effect bonus from having introduced (what I think is) better abstractions/contracts.

theofidry:
Ok I'm merging this now because I don't want to have to rebase it for a 3rd time 😅.

PR #2815, opened 2026-01-09
test: Improve the code coverage tests

theofidry:
## Description

This PR reworks the tests related to the coverage parsing. It leverages the new reports we got from our end-to-end tests and add a lot more tests to cover the current behaviour. This includes:

- PHPUnit 9, 10, 11, 12.0 and 12.5. I have yet to analyse if there is any difference between 10, 11 and 12.0. 9 & 12.5 for sure are different.
- Codeception
- PhpSpec

The complete coverage reports have been moved to `tests/phpunit/TestFramework/Coverage/Fixtures`. The exact process to retrieve them and process them before committing is described in the `README.md` file found in that directory. It essentially amounts to go to the e2e test, generate the coverage report, copy it and execute the Makefile to anonymise the paths.

## Motivations

It is not trivial to code meaningful code coverage output from different test frameworks, and it is something we do not have direct control over. While experimenting with #2543, I was quite frustrated within finding out that some assumptions we made about what the coverage report looked like was were incorrect.

I drafted some changes for https://github.com/infection/infection/issues/2816, but I do not think the current tests are sufficient to ensure we are not breaking things on the way.

## Changes

- For each test framework, besides some extra tests testing something very specific (e.g. "test x with invalid XML"), the tests have been replaced to use those.
- Obsolete fixtures have been removed

## Note to the reviewers

This is HUGE. This is partly due to:

- We are adding a lot of coverage reports (18,997 lines against 2.2 kL before).
- The tests _are_ verbose: there is a lot of data.
- Instead of having a few select tests, we have exhaustive tests for all adapters and different versions of PHPUnit.

Now the question is how do we deal with this. There is several strategies:

- I could have a deeper review to remove reports that we identified as redundant. For example if there is no difference for the JUnit report between PHPUnit 10 & 11, we could remove one of them for the JUnit test. The pro is less code, the cons are that if we do change the implementation and we may do this in a way that affects only specific versions, we could break things without knowing.
- Take the risk and remove a lot more tests.

Before you make up your mind on this, there is a few points to consider:

- The coverage data is a part that will change _a lot_ (the motivations were described in https://github.com/infection/infection/issues/2816). Tests that may seem redundant right now may prevent regressions.
- `TracerIntegrationTest` is arguably the most verbose, but it is also the most important: this is what can guarantee the integrity of the whole tracing service.
- You may notice missing data from the PhpSpec and PHPUnit 12.5. This is caused by the lack of JUnit report. For PhpSpec it is expected, for PHPUnit 12.5 it is because PHPUnit introduced some features that should remove the need for it (and I plan to integrate them). I am still a bit suspicious that the TestLocation file path is missing and I would like to check that, but it is a separate problem and thus out of scope of this PR.

## References

- https://github.com/infection/infection/pull/2816
- https://github.com/infection/infection/pull/2543

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -30,6 +30,7 @@
 
             <exclude>tests/phpunit/AutoReview</exclude>
             <exclude>tests/phpunit/Fixtures</exclude>
+            <exclude>tests/phpunit/TestFramework/Coverage/Fixtures</exclude>
         </testsuite>
     </testsuites>
 


tests/phpunit/TestingUtility/PHPUnit/DataProviderFactoryTest.php
```diff
@@ -58,4 +58,21 @@ public function test_it_creates_a_phpunit_data_provider_from_an_iterable(): void
 
         $this->assertSame($expected, $actual);
     }
+
+    public function test_it_can_add_a_prefix_to_the_title_of_the_scenarios_of_a_provider(): void
```
theofidry, line 62: could be used elsewhere

maks-rafalko (approved):
I have to admit GitHub sucks at displaying such big PRs, comparing to GitLab where you just click to a file and see it alone. Even clicking to a file doesn't work here for me :|

`tests/phpunit/TestFramework/Coverage/Fixtures` should be excluded in phpunit.xml I think same way as we excluded `tests/phpunit/Fixtures` before

theofidry:
>I have to admit GitHub sucks at displaying such big PR

Agreed, I did my best to reduce the noise as much as possible, but GitHub really sucks there.

>tests/phpunit/TestFramework/Coverage/Fixtures should be excluded in phpunit.xml I think same way as we excluded tests/phpunit/Fixtures before

There is no PHP files in there, but you think it should be excluded still?

maks-rafalko:
> There is no PHP files in there, but you think it should be excluded still?

tbh, I'm not sure right now. I think it's better to exclude, and here is why I think so:

back in 2019, I was debugging a major performance degradation of PHPUnit start time, see https://github.com/infection/infection/pull/780

> PHPUnit scans ALL files inside tests folder in order to find tests and it takes up to 1 minute on my machine because of huge amount of vendor folders inside tests/Fixtures/e2e/*. I have reported a bug https://github.com/sebastianbergmann/php-file-iterator/issues/61

I suspect that PHPUnit will go over the files inside fixtures folders and will try to find tests there. Probably it is quick, as files are filtered out by `.php` extentions. Probably not. I didn't do any performance checks now, but since we already have some fixtures exclusions, it's not a big deal to add similar folders. If we ever add PHP files to `.../Fixtures/...`, this will not impact PHPUnit run time.

PR #2817, opened 2026-01-09
Add --with-timeouts and --max-timeouts options

### LINKED ISSUE #1133: Add config option to count timeouts as mutant escapes

### LINKED ISSUE #2221: timed-out mutants limit

sanmai:
## Summary

- Add `--with-timeouts` CLI option / `timeoutsAsEscaped` config to treat timed out mutants as escaped (affects MSI calculation)
- Add `--max-timeouts=N` CLI option / `maxTimeouts` config to fail build if timeout count exceeds threshold
- Show timed out mutations with `-s` flag when `--with-timeouts` is enabled

### Tasks

- [x] Doc PR https://github.com/infection/site/pull/293

## Motivation

Timeouts on CI (due to weaker CPU) can inflate MSI scores since they're counted as "killed" when they might actually be escapees that just ran too slow. These options give users control over how timeouts affect their metrics.

## Usage Examples

```bash
# Treat timeouts as escaped (stricter MSI)
infection --with-timeouts --min-msi=80

# Fail if any timeouts occur (useful for PRs)
infection --git-diff-lines --max-timeouts=0

# Both combined
infection --with-timeouts --max-timeouts=5 --min-msi=75
```

Config file:
```json5
{
    "timeoutsAsEscaped": true,
    "maxTimeouts": 10,
    "minMsi": 77
}
```

## Test plan

- [x] Verify `--with-timeouts` includes timeouts from MSI calculation
- [x] Verify `--max-timeouts=N` fails when timeout count exceeds N
- [x] Verify timed out mutations display with `-s` when `--with-timeouts` enabled
- [x] Run existing test suite

Closes #1133
Closes #2221
Partially addresses #1145

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -959,16 +959,16 @@
         },
         {
             "name": "sanmai/di-container",
-            "version": "0.1.8",
+            "version": "0.1.10",
             "source": {
                 "type": "git",
                 "url": "https://github.com/sanmai/di-container.git",
-                "reference": "f7ea6a00692608f785fc0eabb1c54f5349d973e9"
+                "reference": "7a7f82ee7084de134b9f0d5e2e8926693cc977a2"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/sanmai/di-container/zipball/f7ea6a00692608f785fc0eabb1c54f5349d973e9",
-                "reference": "f7ea6a00692608f785fc0eabb1c54f5349d973e9",
+                "url": "https://api.github.com/repos/sanmai/di-container/zipball/7a7f82ee7084de134b9f0d5e2e8926693cc977a2",
+                "reference": "7a7f82ee7084de134b9f0d5e2e8926693cc977a2",
                 "shasum": ""
             },
             "require": {
@@ -1022,7 +1022,7 @@
             ],
             "support": {
                 "issues": "https://github.com/sanmai/di-container/issues",
-                "source": "https://github.com/sanmai/di-container/tree/0.1.8"
+                "source": "https://github.com/sanmai/di-container/tree/0.1.10"
             },
             "funding": [
                 {
@@ -1030,7 +1030,7 @@
                     "type": "github"
                 }
             ],
-            "time": "2026-01-03T14:10:40+00:00"
+            "time": "2026-01-12T23:28:14+00:00"
         },
         {
             "name": "sanmai/duoclock",
--- a/devTools/phpstan-baseline.neon
+++ b/devTools/phpstan-baseline.neon
@@ -13,12 +13,6 @@ parameters:
 			path: ../src/Command/MakeCustomMutatorCommand.php
 
 		-
-			message: '#^Parameter \#2 \$value of method DIContainer\\Container\:\:set\(\) expects \(callable\(DIContainer\\Container\)\: object\)\|class\-string\<DIContainer\\Builder\<object\>\>, callable\(Infection\\Container\\Container\)\: object given\.$#'
-			identifier: argument.type
-			count: 1
-			path: ../src/Container/Container.php
-
-		-
 			message: '#^Variable \$end might not be defined\.$#'
 			identifier: variable.undefined
 			count: 1


theofidry:
Whilst I understand the motivation, I don't know what to think of this option.

Fundamentally a user has little control over a timeout. Sometimes they do, because it can be about improving the speed of the tests executed, but otherwise you have legitimate cases of timeouts where the mutation causes an infinite loop. So what is the expected course of action for them there? Just allow an arbitrary number of timeouts? Ignore the mutation?

I do see the problem, and I think we don't highlight enough that timeouts (like fatals) can hide something despite covering the mutation, but I'm not sure what the solution is for the MSI...

I'll have to give it some thoughts.

sanmai:
Well, it is an option - not a requirement in any way. 

Personally, I despise timeouts with my guts. Say, I fixed all timeouts on a project, and Infection runs in no time. I added a new feature, and - boom - now it again takes _minutes_ because there are new 20-something timeouts. It is not that hard to fix them: rather, there's no way to know they exist without looking into the logs.

resources/schema.json
```diff
@@ -168,6 +168,16 @@
...
         },
+        "timeoutsAsEscaped": {
+            "type": "boolean",
+            "default": false,
+            "description": "Treat timed out mutants as escaped (affects MSI calculation)."
+        },
+        "maxTimeouts": {
+            "type": "integer",
+            "minimum": 0,
+            "description": "Maximum allowed timeouts. Build fails if exceeded."
```
theofidry: I think the current name/phrasing is a bit confusing in regards of the `timeout` option

src/Configuration/ConfigurationFactory.php
```diff
@@ -313,6 +317,16 @@ private static function retrieveMinCoveredMsi(?float $minCoveredMsi, SchemaConfi
         return $minCoveredMsi ?? $schema->minCoveredMsi;
     }
 
+    private static function retrieveTimeoutsAsEscaped(bool $timeoutsAsEscaped, SchemaConfiguration $schema): bool
+    {
+        return $timeoutsAsEscaped || ($schema->timeoutsAsEscaped ?? false);
+    }
+
+    private static function retrieveMaxTimeouts(?int $maxTimeouts, SchemaConfiguration $schema): ?int
```
theofidry, line 326: probably a `positive-int|0`?

src/Metrics/MaxTimeoutsChecker.php
```diff
@@ -0,0 +1,62 @@
...
+
+declare(strict_types=1);
+
+namespace Infection\Metrics;
+
+/**
+ * @internal
+ * @final
+ */
+class MaxTimeoutsChecker
```
theofidry: can't it be `final readonly`?

src/Metrics/MaxTimeoutsCheckFailed.php
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Infection\Metrics;
+
+use function sprintf;
+use UnexpectedValueException;
+
+/**
+ * @internal
+ */
+final class MaxTimeoutsCheckFailed extends UnexpectedValueException
```
theofidry:
```suggestion
final class MaxAllowedTimeoutCountReached extends UnexpectedValueException
```

? (yes it is a bit more verbose...)

theofidry:
@sanmai what happened there?

src/Configuration/ConfigurationFactory.php
sanmai: It is going to balloon through the entire chain :grimacing:

src/Metrics/MaxTimeoutsChecker.php
sanmai: Made `readonly`, we are mocking it - so soft-final only.

src/Metrics/MaxTimeoutsCheckFailed.php
sanmai: How about `MaxTimeoutCountReached`?

src/Metrics/MetricsCalculator.php
```diff
@@ -218,7 +212,7 @@ public function getStaticAnalysisMaximumRuntime(): float
 
     private function getCalculator(): Calculator
     {
-        return $this->calculator ??= Calculator::fromMetrics($this);
+        return Calculator::fromMetrics($this, $this->timeoutsAsEscaped);
```
sanmai: Not super fond of this

src/Metrics/TargetDetectionStatusesProvider.php
```diff
@@ -80,6 +81,10 @@ private function findRequired(): Generator
     {
         if ($this->numberOfShownMutations !== 0) {
             yield DetectionStatus::ESCAPED->name => DetectionStatus::ESCAPED;
+
+            if ($this->withTimeouts) {
+                yield DetectionStatus::TIMED_OUT->name => DetectionStatus::TIMED_OUT;
+            }
```
sanmai, line 87: Feels odd

resources/schema.json
sanmai:
```suggestion
            "description": "Maximum allowed number of timed-out mutants. Build fails if exceeded."
```

resources/schema.json
```diff
@@ -168,6 +168,16 @@
             "maximum": 100,
             "description": "Minimum Covered Code Mutation Score Indicator (MSI) percentage value. Should be used in CI server."
         },
+        "timeoutsAsEscaped": {
+            "type": "boolean",
+            "default": false,
+            "description": "Treat timed out mutants as escaped (affects MSI calculation)."
```
sanmai:
```suggestion
            "description": "Treat timed-out mutants as escaped (affects MSI calculation)."
```

src/Command/RunCommandHelper.php
```diff
@@ -186,4 +186,17 @@ public function getIgnoreMsiWithNoMutations(): ?bool
             default => true,
         };
     }
+
+    public function getTimeoutsAsEscaped(): bool
+    {
+        return (bool) $this->input->getOption(RunCommand::OPTION_WITH_TIMEOUTS);
+    }
+
+    public function getMaxTimeouts(): ?int
```
theofidry, line 195: I would make the type a bit more strict (and propagate it downstream too). I feel it would be a nice too have but not a blocker either if you feel strongly against it.

```suggestion
    /**
     * @return natural|null
     */
    public function getMaxTimeouts(): ?int
```

src/Configuration/ConfigurationFactory.php
theofidry: ah I already wrote it down here. Yeah... If you don't feel like it it's fine.

src/Event/Subscriber/MutationTestingConsoleLoggerSubscriber.php
```diff
@@ -82,6 +82,7 @@ public function __construct(
         private readonly FederatedLogger $mutationTestingResultsLogger,
         private readonly ?int $numberOfShownMutations,
         private readonly bool $withUncovered,
+        private readonly bool $withTimeouts = false,
```
theofidry: is the default value necessary here? It is always configured explicitly in the container no? (same for the other places).

To be clear: whilst I do see the benefit as it avoids updating some tests, it also increases the risk of forgetting to pass it somewhere (I mean it for here and a few other places where it is introduced as a default value).

tests/phpunit/Event/Subscriber/MutationTestingConsoleLoggerSubscriberTest.php
```diff
@@ -712,6 +712,92 @@ public function test_it_hides_mutation_score_indicator_when_flag_is_false(): voi
         $this->assertStringContainsString('Covered Code MSI: <high>90%</high>', $displayOutput);
     }
 
+    public function test_it_does_not_show_timed_out_mutants_by_default(): void
+    {
+        $output = new StreamOutput(fopen('php://memory', 'w'));
+
+        $this->resultsCollector->expects($this->once())
```
theofidry: _nitpick – feel free to ignore_

I personally like a bit better the form:


```suggestion
        $this->resultsCollector
            ->expects($this->once())
            ->method('getEscapedExecutionResults')
            ->willReturn([]);
```

As you get all the method calls aligned

tests/phpunit/Event/Subscriber/MutationTestingConsoleLoggerSubscriberTest.php
```diff
@@ -712,6 +712,92 @@ public function test_it_hides_mutation_score_indicator_when_flag_is_false(): voi
...
+        // withTimeouts parameter is NOT passed, so it uses the default value of false
+        $dispatcher->addSubscriber(new MutationTestingConsoleLoggerSubscriber(
+            $output,
+            $this->outputFormatter,
+            $this->metricsCalculator,
+            $this->resultsCollector,
+            $this->diffColorizer,
+            new FederatedLogger(),
+            20,
+            false, // withUncovered
```
theofidry: this is a case where named parameters would be better (as it's more reliable than a comment)

tests/phpunit/Event/Subscriber/MutationTestingConsoleLoggerSubscriberTest.php
```diff
@@ -712,6 +712,92 @@ public function test_it_hides_mutation_score_indicator_when_flag_is_false(): voi
...
 
+    public function test_it_does_not_show_timed_out_mutants_by_default(): void
+    {
+        $output = new StreamOutput(fopen('php://memory', 'w'));
+
+        $this->resultsCollector->expects($this->once())
+            ->method('getEscapedExecutionResults')
+            ->willReturn([]);
+
+        // getTimedOutExecutionResults should NOT be called when withTimeouts defaults to false
```
theofidry: this comment and maybe some others of that test feel a bit too verbose, I feel the intent of the test is clear enough from the test title. I don't feel strongly about it though so feel free to ignore

src/Event/Subscriber/MutationTestingConsoleLoggerSubscriber.php
```diff
@@ -110,6 +111,10 @@ public function onMutationTestingWasFinished(MutationTestingWasFinished $event):
             if ($this->withUncovered) {
                 $this->showMutations($this->resultsCollector->getNotCoveredExecutionResults(), 'Not covered');
             }
+
+            if ($this->withTimeouts) {
+                $this->showMutations($this->resultsCollector->getTimedOutExecutionResults(), 'Timed out');
```
theofidry, line 116: Do I understand correctly that this now appears in:

```
140 mutations were generated:
     118 mutants were killed by Test Framework
      18 mutants were configured to be ignored
       4 covered mutants were not detected
```

if you pass `--with-timeouts`?

src/Event/Subscriber/MutationTestingConsoleLoggerSubscriberFactory.php
```diff
@@ -55,6 +55,7 @@ public function __construct(
         private ?int $numberOfShownMutations,
         private OutputFormatter $formatter,
         private bool $withUncovered,
+        private bool $withTimeouts = false,
```
theofidry: same as https://github.com/infection/infection/pull/2817/changes#r2679692783

tests/phpunit/Event/Subscriber/MutationTestingConsoleLoggerSubscriberFactoryTest.php
```diff
@@ -108,4 +120,115 @@ public static function showMutationsProvider(): iterable
...
+        );
+    }
+
+    private function getDisplay(StreamOutput $output): string
+    {
+        rewind($output->getStream());
+
+        $display = stream_get_contents($output->getStream());
+
+        return str_replace(PHP_EOL, "\n", $display);
```
theofidry:
```suggestion
        return Str::toUnixLineEndings($display);
```

src/Command/RunCommand.php
```diff
@@ -488,6 +509,8 @@ private function createContainer(
             ignoreMsiWithNoMutations: $commandHelper->getIgnoreMsiWithNoMutations(),
             minMsi: MsiParser::parse($minMsi, $msiPrecision, self::OPTION_MIN_MSI),
             minCoveredMsi: MsiParser::parse($minCoveredMsi, $msiPrecision, self::OPTION_MIN_COVERED_MSI),
+            timeoutsAsEscaped: $commandHelper->getTimeoutsAsEscaped(),
+            maxTimeouts: $commandHelper->getMaxTimeouts(),
```
theofidry, line 513: This means that currently if you pass `--max-timeouts=10`, currently `timeoutsAsEscaped=false`. Should we implicitly set it? Alternatively we could fail, but I think it is a worst UX.

src/Metrics/TargetDetectionStatusesProvider.php
theofidry: what does?

tests/phpunit/Metrics/MaxTimeoutsCheckerTest.php
```diff
@@ -0,0 +1,104 @@
...
+#[CoversClass(MaxTimeoutsChecker::class)]
+final class MaxTimeoutsCheckerTest extends TestCase
+{
+    public function test_it_does_nothing_when_max_timeouts_is_null(): void
+    {
+        $checker = new MaxTimeoutsChecker(null);
+
+        $checker->checkTimeouts(100);
+
+        $this->addToAssertionCount(1);
```
theofidry: I suggest using `expectNotToPerformAssertions()` which is a bit more explicit and not marked as `@internal`

tests/phpunit/Metrics/MaxTimeoutsCheckerTest.php
```diff
@@ -0,0 +1,104 @@
...
+
+use Infection\Metrics\MaxTimeoutCountReached;
+use Infection\Metrics\MaxTimeoutsChecker;
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\TestCase;
+
+#[CoversClass(MaxTimeoutsChecker::class)]
+final class MaxTimeoutsCheckerTest extends TestCase
+{
+    public function test_it_does_nothing_when_max_timeouts_is_null(): void
```
theofidry: I tend to prefer one test with a data provider, but it feels like Claude really likes this style instead...

tests/phpunit/Metrics/MaxTimeoutsCheckerTest.php
```diff
@@ -0,0 +1,104 @@
...
+    public function test_it_throws_when_timed_out_count_exceeds_limit(): void
+    {
+        $checker = new MaxTimeoutsChecker(5);
+
+        try {
+            $checker->checkTimeouts(10);
+
+            $this->fail('Expected MaxTimeoutCountReached to be thrown');
+        } catch (MaxTimeoutCountReached $exception) {
+            $this->assertSame(
```
theofidry: it would be a bit less verbose to use `::expectExceptionObject()` (same for the next test)

tests/phpunit/Metrics/MetricsCalculatorTest.php
```diff
@@ -125,4 +126,28 @@ public function test_its_metrics_are_properly_updated_when_adding_a_new_process(
         $this->assertSame(100.0, $calculator->getCoverageRate());
         $this->assertSame(100.0, $calculator->getCoveredCodeMutationScoreIndicator());
     }
+
+    public function test_calculator_is_memoized(): void
```
theofidry, line 130: I did use "memoized" too by reflex in some places, but wouldn't be "cached" be more correct?

tests/phpunit/Event/Subscriber/MutationTestingConsoleLoggerSubscriberTest.php
sanmai: That's what Claude added; I don't feel strongly about it and also apparently these comments help Claude grasp the purpose of the tests better, so I'm usually inclined to let them be.

src/Event/Subscriber/MutationTestingConsoleLoggerSubscriber.php
sanmai: The summary does not change, but it prints the escaped mutants including the timeouts:
```diff
Timed out mutants:
==================

1) src/Helper/WindowIterator.php:100    [M] GreaterThanNegotiation [ID] c37dd83f4

@@ @@
             return;
         }
 
-        while ($this->count() > $this->maxSize) {
+        while ($this->count() <= $this->maxSize) {
             unset($this->buffer[$this->headKey]);
             ++$this->headKey;
             --$this->position;

```

tests/phpunit/Metrics/MaxTimeoutsCheckerTest.php
sanmai: Yep. Asked to fix it.

src/Command/RunCommand.php
sanmai: That would not be great; say you have 9 timeouts and 100% MSI. If you tag `timeoutsAsEscaped=true` automatically, then MSI will drop below 100% as the escaped mutants are now accounted in the MSI. Build fails. So it should be the user's call to enable the other option. In other words:

  - `--max-timeouts=10` = "fail-fast if something is badly wrong (infrastructure issue, infinite loops)"
  - `--with-timeouts` = "timeouts are test quality issues, count them in MSI"
  - Or it could be both things if a user desires

What do you think?

src/Metrics/MaxTimeoutsCheckFailed.php
theofidry: 👍

src/Command/RunCommand.php
theofidry: No I agree. `--max-timeouts` is just a hard limit check regardless of how you treat timeout mutations

PR #2823, opened 2026-01-11
test: Relocate specific coverage fixtures

theofidry:
## Context

We have Coverage XML/Junit files stored under `tests/phpunit/Fixtures`. The issue is that we have a few files meant to be re-used (and are), and others that are very specific to on peculiar test.

When the time comes to clean up, it is hard to tell. Or rather it is easy, but slow: delete a file, run all the tests. It's a big waste of time.

## Problem

Does this need to change? Until now no, so it was fine that it grew a bit and that there was useless files. I'm however adjusting a lot of those tests and a lot more complete and accurate coverage fixture files will be added. So it is time to clean it up.

## Description

In this PR I re-locate the fixtures files that are specific to peculiar services/tests. `tests/phpunit/Fixtures` now contains only fixture files that are re-used across multiple tests (notably the individual tests touching the different part of the code coverage generation, `JUnitTestFileDataProvider`, `SourceFileInfoProvider`, `IndexXmlCoverageParser` etc.)

Note that although those files will be replaced in #2815, the principle will remain the same: I will add more reports from various e2e tests for different versions of PHPUnit, PhpSpec and other, and the usage will be the same too. Hence the current split helps a lot to know what can be removed and what cannot*

*: I will re-iterate: we do want to remove unnecessary files, because there will be _a lot_ added.


sanmai (approved):
Moving fixtures close to the tests 👍

sanmai:
```
./devTools/check_trailing_whitespaces.sh
fatal: :!tests/phpunit/Fixtures/Files/phpunit/format-whitespace/original-phpunit.xml: no such path in the working tree.
```
Feels related :thinking:

theofidry:
Arg, I may have forgotten to run that one, lemme fix it immediately

theofidry:
Ha actually I did, but it doesn't make the process fail I see... Well, one thing at a time

theofidry:
@sanmai https://github.com/infection/infection/pull/2828

PR #2825, opened 2026-01-11
feat: Add --logger-summary-json CLI option

sanmai:
## Description

Add CLI support for specifying the summary JSON log file path directly via `--logger-summary-json`, matching existing patterns for `--logger-gitlab`, `--logger-html`, and `--logger-text` options.

Extracted from #2804

## Changes

- Adds `--logger-summary-json` CLI option to `RunCommand`
- Adds `getSummaryJsonLogFilePath()` to `RunCommandHelper` with reusable `trimmedStringOrNull()` utility
- Threads `summaryJsonLogFilePath` through `Container` and `ConfigurationFactory`
- Adds `setSummaryJsonLogFilePath()` to `Logs` entry class for CLI overrides

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (https://github.com/infection/site/pull/292)

## Testing steps

```
./dist/infection.phar --logger-summary-json=/dev/stdout
```
Outputs:
```json
{"stats":{"totalMutantsCount":100,"killedCount":99, ... }
```


src/Command/RunCommandHelper.php
```diff
@@ -171,4 +172,19 @@ public function getIgnoreMsiWithNoMutations(): ?bool
...
+
+    public function getSummaryJsonLogFilePath(): ?string
+    {
+        return self::trimmedStringOrNull($this->input->getOption(RunCommand::OPTION_LOGGER_SUMMARY_JSON));
+    }
+
+    /**
+     * Converts an input option to a trimmed string, or null if empty.
+     */
+    public static function trimmedStringOrNull(string|bool|int|float|null $value): ?string
```
theofidry: in this peculiar case only having `?string` as an input, and then this utility could be in `Str` (you could still mark the method as `@interna` if we do not want to expose it to the public API)

src/Configuration/Entry/Logs.php
```diff
@@ -51,7 +51,7 @@ public function __construct(
         private readonly ?string $perMutatorFilePath,
         private bool $useGitHubAnnotationsLogger,
         private readonly ?StrykerConfig $strykerConfig,
-        private readonly ?string $summaryJsonLogFilePath,
+        private ?string $summaryJsonLogFilePath,
```
theofidry, line 54: so where is this used?

src/Command/RunCommandHelper.php
sanmai: I actually do not like this claudeizm at all. Adding #2826 as a replacement.

maks-rafalko (approved):
Out of curiosity: is it needed for metrics RFC?

sanmai:
@maks-rafalko Exactly!

src/Configuration/Entry/Logs.php
sanmai: In `setSummaryJsonLogFilePath` so we can override it from the CLI option.

PR #2827, opened 2026-01-11
test: Restructure the coverage tests to leverage more data providers

theofidry:
In this PR, I restructure the coverage related tests to leverage more the data providers and better tests the actual outcomes. For example in `JUnitTestFileDataProviderTest` we check the exact results rather than just the file path (i.e. we also check the timings) for the other JUnit paths.

The goal of this PR is to make it easier to add more scenarios in a structured way, which is what will be done in #2815. In the latter, we will no longer use `XmlCoverageFixtures`. This means the structure of the data providers changes a bit compared to right now (details in the comments of the diff). As a result, a bit of adapter code needs to be done. For example, when we test the test locations parsed from `XmlCoverageParser`, we now check the `TestLocations` instances rather than the normalized tests. This means we need to convert back the normalized tests to a `TestLocations` instance. This would not be ideal if that code is meant to live a long time, but this is really just to provide that re-structuring whilst preserving all the existing tests.

Besides 1-2 test being added, no test (expect one commented) was deleted.


tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestFileDataProviderTest.php
```diff
@@ -59,140 +68,120 @@ final class JUnitTestFileDataProviderTest extends TestCase
...
+    private string $generatedJunitPath;
 
     protected function setUp(): void
     {
-        $this->jUnitLocatorMock = $this->createMock(ReportLocator::class);
-
-        $this->provider = new JUnitTestFileDataProvider($this->jUnitLocatorMock);
-
-        $this->tempfile = tempnam('', '');
+        $this->generatedJunitPath = FS::tmpFile('JUnitTestFileDataProviderTest');
```
theofidry, line 73: this makes the file slightly more identifiable but if you're unhappy about it I can switch back to `tempnam()`. Technically `FS` is more robust (as it relies on the Symfony filesystem, you can check `Filesystem::tempnam()`) but I don't really care enough to argue over this either.

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestFileDataProviderTest.php
```diff
@@ -59,140 +68,120 @@ final class JUnitTestFileDataProviderTest extends TestCase
 
     private const JUNIT_CODECEPTION_CEST_FORMAT = __DIR__ . '/../../../Fixtures/Files/phpunit/junit_codeception_cest.xml';
 
-    private MockObject&ReportLocator $jUnitLocatorMock;
-
-    private JUnitTestFileDataProvider $provider;
-
-    private string $tempfile;
```
theofidry, line 66: renamed `tempFile` to `generatedJUnitPath` to reflect what this "temp file" is about, which was not immediately obvious.

tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestFileDataProviderTest.php
```diff
@@ -59,140 +68,120 @@ final class JUnitTestFileDataProviderTest extends TestCase
...
-        $testFileInfo = $this->provider->getTestFileInfo('app\controllers\ExampleCest:FeatureA');
-
-        $this->assertSame('/app/controllers/ExampleCest.php', $testFileInfo->path);
+        return new JUnitTestFileDataProvider(
+            new FixedLocator($file),
+        );
     }
 
-    #[DataProvider('xmlProvider')]
-    public function test_it_does_not_trigger_count_assertion(string $xml): void
```
theofidry, line 158: this test was NOT ported. I simply don't understand what it's supposed to do...

tests/phpunit/TestFramework/Coverage/XmlReport/IndexXmlCoverageParser/IndexXmlCoverageParserTest.php
```diff
@@ -66,106 +75,164 @@ final class IndexXmlCoverageParserTest extends TestCase
...
     {
         $this->fileSystem = new FileSystem();
         $this->parser = new IndexXmlCoverageParser(
-            false,
-            $this->fileSystem,
+            isSourceFiltered: false,
+            fileSystem: new FakeFileSystem(),
         );
+
+        $this->generatedIndexXmlPath = FS::tmpFile('IndexXmlCoverageParserTest');
```
theofidry, line 73: This test had a `$filename = __DIR__ . '/generated_index.xml';` file which was used as a temporary file. I feel like the approach we had in `JUnitTestFileDataProviderTest`, which was to ask the FS a temporary file a bit better (and if we are to parallelise tests ever, would also work better in that context), so i copied it here.

This mean this test no longer needs this `generated_index.xml`. Well currently it still does because the "old coverage test", but that will go away in the next PR, it's not important enough to try to tweak this test any further.

tests/phpunit/TestFramework/Coverage/XmlReport/SourceFileInfoProviderTest.php
```diff
@@ -52,94 +54,165 @@ final class SourceFileInfoProviderTest extends TestCase
...
-                    $incorrectCoverageSrcDir,
-                    Path::canonicalize(XmlCoverageFixtures::FIXTURES_COVERAGE_DIR),
-                ),
-                $exception->getMessage(),
-            );
-            $this->assertSame(0, $exception->getCode());
-            $this->assertNull($exception->getPrevious());
-        }
+        // TODO: this is not ideal...
+        $this->expectException(ValueError::class);
```
theofidry, line 193: this is a new test; I would like to fix it but it's quite and edge case so neither critical neither need to be fixed in the next PR. But I think it' still worth documenting the status quo.

tests/phpunit/TestFramework/Coverage/XmlReport/XmlCoverageParserTest.php
```diff
@@ -59,139 +62,126 @@ protected function setUp(): void
...
-    /**
-     * @param array<string, mixed> $expectedTests
-     */
-    #[DataProvider('sourceFileInfoProviderProvider')]
-    public function test_it_reads_every_type_of_fixture(
-        SourceFileInfoProvider $provider,
-        array $expectedTests,
+    #[DataProvider('lineCoverageProvider')]
+    public function test_it_can_get_the_line_coverage(
+        SafeDOMXPath $xPath,
```
theofidry, line 67: in the current form of the API, providing the xPath is a lot more flexible because we can create it either from a real file (`SafeDOMXPath::fromFile()`) or from a string. This allowed to easily combined the scenarios from the previous data provider (which was using real files) and the ones from the other tests which were defining the XML string.

sidz (approved):

PR #2832, opened 2026-01-11
fix(test): Synchronize the skip node behaviour of the tests with the real code

theofidry:
## Description

When testing our mutators, we use `SimpleMutationsCollectorVisitor` instead of `NodeMutationGenerator`. At some point `NodeMutationGenerator` got updated, but not `SimpleMutationsCollectorVisitor`. As a result, the tests we have no longer reflect how our mutators behave when applied to real code.

The goal of this PR is to address this.

## Changes

`NodeMutationGenerator` skips code that is in the global scope or functions, hence `SimpleMutationsCollectorVisitor` applies the same. This results in a lot of mutator no longer testing anything because they were touching code that is no longer touched. 

The solution opted is to introduce a utility `BaseMutatorTestCase::wrapCodeInMethod(string $code, ?string $namespace = null): string`. This transforms:

```php
$class->getObject()?->getName();
```

into:

```php
<?php

namespace Acme;

class WrappingClass {
    public function wrappedTestedCode() {
        $class->getObject()?->getName();
    }
}

```

There is a few outliners, but that's the gist of the changes.

## Note to the reviewers

In https://github.com/infection/infection/pull/2811 we extensively discussed about our intention to have out mutator tests behaving the same way as production code, and this applies to our consumers too.

This overall change is pushed by #2797.

This is a massive PR but the changes are simple. The initial changes we pushed using AI to drive the refactoring and polished by hand. I try to review each file already, I hope I did not miss anything.


maks-rafalko:
> So from that experience, I wonder if we need any mutator "unit test".

what is the alternative? 😮 

if we want our mutators unit tests to be as close to the real code as possible, then we have to wrap them with class+methods, to be mutated

> While working on #2795 I was constantly battling weird issues that should not be, and couldn't figure it out for a while.

what are the issues and how are they related to this change?

theofidry:
>>  So from that experience, I wonder if we need any mutator "unit test".

>what is the alternative? 😮

If we had a "unit test version" we wouldn't need to adapt those tests (I think?). But then again I'm not sure if it's the right solution either.

Let's go with the original plan which is to synchronize the behaviour even if it means we have a lot of tests to fix.

>what are the issues and how are they related to this change?

Doing what I thought would be minor changes and causing a whole lot of things to break, but not because my code is bad, but because the assumption that the tests we had were testing the real behaviour was incorrect in the first place.

It's been a real long time since I touched the mutator part and it's been a bit confusing to me how certain pieces evolved and wasn't clear if I was doing something wrong or if there were bugs.

Evidently from this PR there is at least one bug, and I'll need to fix those. We'll get a complete diff of the difference before/after from it.

theofidry:
In theory it is green, it is locally at least... But this was done using a lot Claude, I need to do a review still.

tests/phpunit/Mutator/Arithmetic/ShiftRightTest.php
```diff
@@ -55,38 +55,36 @@ public function test_it_can_mutate(string $input, string|array|null $expected =
...
+            self::wrapCodeInMethod(
+                <<<'PHP'
+                    $a = 1;
+                    $a << 2;
+                    PHP,
+            ),
         ];
     }
 
     public function test_replaces_post_decrement(): void
```
theofidry: Moving this to the data provider; but to extract this change

tests/phpunit/Mutator/Cast/CastStringTest.php
```diff
@@ -55,86 +55,86 @@ public function test_it_can_mutate(string $input, string|array|null $expected =
...
-                <?php
-
-                function noReturnType()
-                {
-                    return random_int();
-                }
-                PHP,
+            self::wrapCodeInMethod(
+                <<<'PHP'
+                    function noReturnType()
```
theofidry, line 103: previously I updated that code to make it a method instead of a function but actually you can declare a function within a method just fine so I left it alone afterwards.

I only changed it if we used `declare()` as this needs to be declared outside a class/function.

tests/phpunit/Mutator/FunctionSignature/PublicVisibilityTest.php
```diff
@@ -344,7 +344,7 @@ protected function foo(): bool
         ];
 
         yield 'It does not remove attributes' => [
-            <<<'PHP_WRAP'
+            <<<'PHP'
```
theofidry, line 347: to extract this change

theofidry:
@maks-rafalko ready for review... if this can be reviewed... I did go through all files before handing it over to you, but it is a lot of them.

tests/phpunit/Mutator/Arithmetic/AssignmentEqualTest.php
```diff
@@ -55,36 +55,36 @@ public function test_it_can_mutate(string $input, string|array|null $expected =
...
+                    if ($a == $b) {
+                    }
+                    PHP,
+            ),
+            self::wrapCodeInMethod(
+                <<<'PHP'
+                    if ($a = $b) {
+                    }
+                    PHP,
+            ),
```
maks-rafalko, line 69: what if we do `wrapCodeInMethod()` later in the base class, not here?

- it will allow leave the code as it was
- consumers will not need to updated their tests
- we will not need to use this in the future tests

----

UPD: after some thinking, I realized that 99% that you considered these, but not all tests would work with this approach: for example those wouldn't that already have class+method "wrapping"

So, the comment above can be ignored, but leaving just in case

theofidry: >but not all tests would work with this approach: for example those wouldn't that already have class+method "wrapping"

It is nonetheless the majority so it is legitimate to ask if we should flip it. But personally I do not like that idea: it is certainly not a transformation that I would expect (as a user or uninformed developer) to be done.

PR #2833, opened 2026-01-11
refactor: Abstract the trace from the filesystem

theofidry:
## Description

This removes the last piece of dependency the trace has on the native filesystem (`file_get_contents()`, `file_put_contents()`, etc., which cannot be mocked).

This will allow to adjust `PHPUnitCoverageTracerTest` in the future to no longer require the source files parsed by the coverage data to actually exist on the filesystem.

## Motivation

The `IndexXmlCoverageParser` creates `SourceFileInfoProvider`, and as reworked in #2820, we can properly mock the filesystem there, which makes testing `IndexXmlCoverageParser` easier. The same was done for `SourceFileInfoProvider` in #2820.

This worked brilliantly for all pieces when I worked on #2815. All? ~~Not quite! Because a village inhabited by unwavering Gauls is still resisting the invaders.~~ Sorry French joke. Anyway one piece is still missing, which is for `PHPUnitCoverageTracerTest` (which IMO is the most important test).

Indeed, `SourceFileInfoProvider` creates a real `SplFileInfo` instance, which means when we do `SourceFileInfoProvider->getSplFileInfo->getRealPath()`... this cannot be mocked... And although we have `Trace::getRealPath()`, `ProxyTrace` just aliases `$this->splFileInfo->getRealPath()`...

## Changes

The simplest solution is to make `XmlCoverageParser` resolve the real path when creating the `ProxyTrace` with the injected filesystem instance.

### DIFF
--- a/tests/phpunit/TestFramework/Tracing/TracerIntegrationTest.php
+++ b/tests/phpunit/TestFramework/Tracing/TracerIntegrationTest.php
@@ -93,14 +93,14 @@ final class TracerIntegrationTest extends TestCase
                         fileSystem: new FileSystem(),
                     ),
                     parser: new XmlCoverageParser(
-                    new FileSystem(),
+                        new FileSystem(),
+                    ),
                 ),
-            ),
-            new JUnitTestExecutionInfoAdder(
-                $testFrameworkAdapterStub,
-                new MemoizedTestFileDataProvider(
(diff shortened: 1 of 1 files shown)


theofidry:
This is fairly blocking and I would really like to get the test coverage saga to end tonight. I promise this PR is out of necessity, the next PR will make use of it.

PR #2837, opened 2026-01-12
refactor(test): Use the Container configured tracer for `TracerIntegrationTest`

theofidry:
This is currently not working, trying to figure out what we're missing to get there.

Depends on:

- #2835
- https://github.com/infection/infection/pull/2838


src/Container/Container.php
```diff
@@ -1174,4 +1174,17 @@ private function offsetSet(string $id, callable $value): void
     {
         $this->set($id, $value);
     }
+
+    /**
+     * @param class-string<object> $id
+     * @param callable(static): object $value
+     */
+    public function withService(string $id, callable $value): self
```
sanmai, line 1182:
```suggestion
    public function withService(string $id, object $value): self
```
looks like we can reduce some boilerplate in the tests

sanmai:
Apologies for the PR hijack. PHPStan should be happy now.

src/Container/Container.php
```diff
@@ -1174,4 +1174,17 @@ private function offsetSet(string $id, callable $value): void
...
+    {
+        $clone = clone $this;
+
+        $clone->set($id, $value);
+
+        return $clone;
+    }
```
theofidry, line 1189: Actually should be:

```suggestion
    /**
     * @template T
     * 
     * @param id|class-string<T> $id
     * @param T|(callable(static): T) $value
     */
    public function withService(string $id, object|callable $value): self {
    {
        $clone = clone $this;

        $clone->set(
          $id,
          is_callable($value)
            ? $value
            : static fn () => $value,
        );

        return $clone;
    }
```

WDYT @sanmai? Should it be here or should it be moved to the container package?

sanmai: Do you mean specifically `withService` or a way to set prebuilt dependencies?

sanmai: I'm not sure immutable `withService` for everyone, but directly injecting is good:

https://github.com/sanmai/di-container/pull/47

sanmai: OK, it is out.

theofidry: I actually meant with either.

But it is fine, you have `::inject()` for a synthetic service and `::offsetSet()` for a regular service.

theofidry: Thanks!

tests/phpunit/TestFramework/Tracing/TracerIntegrationTest.php
```diff
@@ -85,24 +89,48 @@ protected function setUp(): void
             ->method('hasJUnitReport')
             ->willReturn(true);
 
-        $this->provider = new CoveredTraceProvider(
```
theofidry, line 88:
```php
return SingletonContainer::getContainer()
            ->withValues(
                logger: new NullLogger(),
                output: new NullOutput(),
                skipInitialTests: true,
                existingCoveragePath: dirname($indexXmlPath),
            )
            ->withService(
                'app.coverage.index_locator',
                new FixedLocator($indexXmlPath),
            )
            ->withService(
                FileSystem::class,
                $fileSystemStub,
            )
            ->withService(
                TestFrameworkAdapter::class,
                $testFrameworkAdapterStub,
            )
            ->withService(
                TestFileDataProvider::class,
                $junitFileDataProvider,
            )
            ->getTracer();
```


But unfortunately this appears insufficient as this requires a configuration file too...

PR #2838, opened 2026-01-12
refactor: Leverage the CachedLocator

theofidry:
## Description

This code was the one intended to be proposed in #2588 but was blocked by https://github.com/sanmai/di-container/issues/25 at the time.

I got back to it as I was experimenting with #2837 where I am blocked with the same issue. [^1]

## Changes

This changes the index & JUnit XML locators services for a string ID based service, which allows us to configure them as a `ReportLocator` instead of a specific concrete implementation, which in turns allows us to move out the caching out of the base locator, simplifying that class and some tests.

`MemoizedLocator` was merged in #2588 although unused as the blocker was meant to be dealt with soonish (it now has). I did rename it to `CachedLocator` as it also works (memoizing is a form of caching) and has the merit of being a real word ("memoized does not exist in the dictionary) and is probably understood more easily by people;

[^1]: The issue being we currently cannot swap the index or JUnit XML locator service for another `ReportLocator` implementation (e.g. `FixedLocator`), as the container does a check `service ID = a class -> instantiated service must be of that class` – regardless of whether there is a common interface like `ReportLocator` here).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -59,7 +59,7 @@
         "nikic/php-parser": "^5.6.2",
         "ondram/ci-detector": "^4.1.0",
         "psr/log": "^2.0 || ^3.0",
-        "sanmai/di-container": "^0.1.4",
+        "sanmai/di-container": "^0.1.9",
         "sanmai/duoclock": "^0.1.0",
         "sanmai/later": "^0.1.7",
         "sanmai/pipeline": "^7.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "a48210ad9db94dfc72a572d0cda57233",
+    "content-hash": "76630056a78ce5d963800c84d3747646",
     "packages": [
         {
             "name": "colinodell/json5",


src/Container/Container.php
```diff
@@ -243,6 +245,7 @@ final class Container extends DIContainer
 
     public static function create(): self
     {
+        // @phpstan-ignore argument.type
```
theofidry: @sanmai I'm not too happy with those phpstan suppression statements, but that's what it took to use that feature right now. I don't know if you want to check if that can be fixed within the container types

tests/phpunit/Container/ContainerTest.php
```diff
@@ -137,12 +138,19 @@ public function test_it_provides_a_friendly_error_when_attempting_to_configure_i
 
     public static function provideServicesWithReflection(): iterable
     {
+        $nonClassServiceIds = [
```
theofidry, line 141: not too thrilled about that either tbh, but I can live with it

src/Container/Container.php
sanmai: https://github.com/sanmai/di-container/releases/tag/0.1.9 should fix this

sanmai:
I've been looking around and I wonder if there's a simpler alternative that avoids the string-based service IDs, hand-wiring, and so on.

For #2837's test swapping needs, we could use inheritance instead:

```php
// In tests/...
trait FixedLocatorTrait {
    public function __construct(private readonly string $fixedPath) {}

    public function locate(): string {
        return $this->fixedPath;
    }
}

final class FixedIndexXmlCoverageLocator extends IndexXmlCoverageLocator {
    use FixedLocatorTrait;
}

final class FixedJUnitReportLocator extends JUnitReportLocator {
    use FixedLocatorTrait;
}
```

Then in tests:
```php
$container->set(
    IndexXmlCoverageLocator::class,
    static fn () => new FixedIndexXmlCoverageLocator($coveragePath . '/xml/index.xml'),
)
```

This way `FixedIndexXmlCoverageLocator` IS-A `IndexXmlCoverageLocator`, so the container accepts it under `IndexXmlCoverageLocator::class`, and likewise for another locator.


What do you think?

sanmai:
Actually, you can even do this:

```php
$locatorMock = $this->createMock(IndexXmlCoverageLocator::class);
$locatorMock->method('locate')->willReturn($coveragePath . '/xml/index.xml');

$container->set(
    IndexXmlCoverageLocator::class,
    static fn () => $locatorMock,
);
```
So none of the refactoring will be required.

theofidry:
@sanmai the workaround would do if I unmarked them as final indeed, and I certainly would prefer that over a trait/inheritance "hack".

>avoids the string-based service IDs

Just to clarify: I do feel that this is quite a basic feature for a container. The problem stems more from us having enforced some class-ID as service ID conventions – which is something we added.

So I guess my question is more: are you against it as the idea itself, or fixing it? Because there is 3 options:

1. we don't change the status quo (here for example it means making `IndexXmlCoverageLocator` non-final and keeping the cache within the base locator.
2. the current state of the container actually works, it's just phpstan not being happy about some pieces but we can just ignore them like done in this PR. It is not a big deal, just an annoyance.
3. we want to fix it (and if you don't want to spend time on it I can look into it).

sanmai:
No, I'm not against the idea of non-class-name service IDs.

I'm good with making `IndexXmlCoverageLocator` soft-final (`@final`) and keeping the cache within the base locator, _unless_ we have a practical need to separate them. I did not get if we are.

theofidry:
>I'm good with [...] keeping the cache within the base locator, unless we have a practical need to separate them. I did not get if we are.

I like to have implementations to focus on one thing, so I tend to prefer the composition of decorators. So when I thought of the `ReportLocator` interface, and I liked the idea of having the base implementation not to have to worry about caching and just have an explicit caching decorator with it.

As a result of it is that the unit tests are simpler as you have less behaviour to test, and less side-effects. For instance it is easier to write up the test to ensure the report is cached in the dedicated `CachedLocatorTest`, rather than in `BaseReportLocatorTest`, where there is more things going on.

But ultimately it's fairly minor, even more so in this case. I find the mock workaroud "ugly", but it's nitpicks in the end.

PR #2841, opened 2026-01-12
feat: Add file-based OPcache support for mutation testing

sanmai:
## Description

Add automatic file-based OPcache configuration to improve mutation testing performance by ~8%. When XdebugHandler restarts PHP without Xdebug, OpcacheConfigurer appends opcache settings to the temporary ini file, enabling bytecode caching for all child processes.

The implementation leverages the same mechanism used by MemoryLimiter to inject settings into child processes via PHPRC environment variable.

**Caveat:** This PoC optimization _only benefits users who have Xdebug enabled_. XdebugHandler only creates the temporary ini file when it restarts PHP to disable Xdebug. Users running without Xdebug won't see any improvement from this change. Naturally, we would want everyone to benefit from it.

## Changes

- Add `useOpcache` config option (default: true) to enable/disable
- Create OpcacheConfigurer service that configures file-based cache
- Opcache directory uses Infection's tmpDir for automatic cleanup
- Only activates when XdebugHandler creates temp ini file

## Benchmark

To verify the improvement, run Infection with and without the feature:

```bash
# With opcache (xdebug enabled, XdebugHandler restarts)
sudo phpenmod xdebug
/usr/bin/time -f "Time: %E" ./bin/infection --coverage=var/coverage --skip-initial-tests --no-interaction --no-progress 2>&1 | tail -5

# Without opcache (xdebug disabled, no XdebugHandler restart)
sudo phpdismod xdebug
/usr/bin/time -f "Time: %E" ./bin/infection --coverage=var/coverage --skip-initial-tests --no-interaction --no-progress  2>&1 | tail -5
sudo phpenmod xdebug
```

Results on this codebase:
- With opcache: ~43s
- Without opcache: ~47s
- Improvement: ~8.5%

### Full potential analysis

To measure the theoretical maximum, we tested with PHPRC set before Infection starts (so parent process also benefits):

| Scenario | Time | .bin files | Improvement |
|----------|------|------------|-------------|
| PHPRC opcache (parent + children) | 44.5s | 2717 | **11%** |
| Current implementation (children only) | 43s | 2429 | **8.5%** |
| No opcache baseline | 50s | 0 | - |

Key findings:
- **Full potential is ~11%**, not 15% as initially estimated
- **`-d` CLI options don't propagate** to child processes - PHPRC is required for children to inherit settings
- **Warm vs cold cache difference is negligible** (~0.2s) - most benefit comes from caching during the run itself
- **Current implementation captures ~77% of the theoretical maximum** (8.5% vs 11%)

The ~2.5% gap is from the parent Infection process not benefiting. Capturing this would require a wrapper script or external configuration - not worth the complexity.

## Checklist

- [ ] Tests added/updated

## Reviewer Notes

The implementation activates only when XdebugHandler restarts PHP, which means users without Xdebug won't benefit. This is a pragmatic trade-off - modifying the system php.ini is not safe, and the XdebugHandler mechanism provides a clean way to inject settings via PHPRC.


staabm:
Hiding this feature behind xdebug doesn‘t make sense to me. 

IMO people learnt to disable xdebug whenever possible because it has such a big perf impact on the runtime. 

I think a better idea would be adding a new cli switch to infection which can e.g. auto restart processes and append opcache settings when enabled.

opcache sometimes has bugs and whether opcache can/should be enabled depends on the used php version and/or development  environment details

sanmai:
@staabm It sounds like the idea itself is worthy, but the implementation is lacking, am I reading you wrong?

(I don't really have any attachment to this specific way of doing it, as it was an easy path for a PoC.)

staabm:
> It sounds like the idea itself is worthy, but the implementation is lacking?

yes, I think so too. the idea is great.

staabm:
detail wise: when running in CI we need to make sure to put the opcache file into a folder which gets/can be persisted between CI runs

sanmai:
@staabm How did you get opcache persistence to work? From what I recall from testing, every git clone resets file timestamps so the cache goes bust.

theofidry:
@sanmai you don't think the idea is worth pursuing still? If it is (to me it looks interesting), I would suggest to create an issue to document it including the failed attempt

sanmai:
To many iffs to make it work, TBH. You have to have a very specific setup to make it work, or we have to add XdebugHandler-like contraption for PCOV. I'll look into documenting it.

staabm:
fwiw, in PHPStan we currently play with implementing a native extension for hot paths in PHPStan.

https://github.com/phpstan/phpstan-src/tree/2.2.x/turbo-ext

it looks like this can speedup the overall experience by 20-30%. maybe thats something you want to look into as well.
its still early days though and ondrej is finalizing it for the upcoming release.

IMO opcache support is not stable enough yet to promote for our CLI tools and requiers manual fiddling by the end user. as it stands right now, it seem we can ship the extension and make its use transparent to the end user.

PR #2847, opened 2026-01-13
test: Make adding more scenarios for TracerIntegrationTest easier

theofidry:
This PR is the follow up of #2833 and the last preparatory step for #2815.

Now that we abstracted the filesystem entirely[^1] entirely in the tracing process, we can have a much more flexible test.

I adapted the existing test scenario to accommodate for that new setup. You will notice, from a readability PoV at least there is no benefit, if anything it is slightly worst. But there is concrete benefits:

- The JUnit location and index XML can be specified on a case per case basis. As per the footnote however, the file XML need to be in the right location relative to the index XML.
- **The source file does not need to exist anymore**. In the current test it does, and there is no interest in removing it as we have a micro-project we can update to generate new coverage data. But if you remove the `realpath()` function call in the test data provider and remove the `DemoCounterService.php`, the test will remain green.

And with this, we can easily add a coverage report and add a scenario for it to test what the trace looks like. No need to care about the absolute paths anymore nor to have the source files described from the coverage report.

[^1]: There is one caveat, which is the file lookup for the file XML in `SourceFileInfoProvider`: the path resolution is done in `SourceFileInfoProvider` and then we get it from the filesystem. Actually, the necessary abstraction is already there, but it requires a more complex setup for the FileSystem stub used in `TracerIntegrationTest` – the one I introduced is fairly simple, we only remove the file existence check from `::realPath()`.  So it is possible in theory, but in practice at the moment I do not see the need for it, hence I did not pursue that.


tests/phpunit/TestFramework/Tracing/TracerIntegrationTest.php
```diff
@@ -73,59 +72,28 @@ final class TracerIntegrationTest extends TestCase
...
-        );
-
-        $this->copyReportFromTemplateIfMissing($coveragePath);
-    }
-
     #[DataProvider('traceProvider')]
     public function test_it_can_create_a_trace(
+        string $indexXmlPath,
+        string $junitXmlPath,
         SplFileInfo $fileInfo,
```
theofidry: could simplify this, remove the need to pass `SplFileInfo` as it is the one that should end up in the expected, so we could remove it from here and do:

```diff
- $actual = $tracer->trace($fileInfo);
+ $actual = $tracer->trace($expected->getSplFileInfo));
```

tests/phpunit/TestFramework/Tracing/TracerIntegrationTest.php
```diff
@@ -426,20 +398,63 @@ public static function traceProvider(): iterable
         ];
     }
 
-    private function copyReportFromTemplateIfMissing(string $coveragePath): void
```
theofidry, line 429: Ha this copyReport is still necessary!

theofidry:
It's been two weeks I would like to move forward a bit...

Issue #2440, opened 2025-10-10
"Class XX is not a valid target for code coverage"


staabm:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.31.2
| Test Framework version | PHPUnit 12.4
| PHP version | 8.3.26 (cli) (built: Sep 23 2025 17:57:26) (NTS)
| Platform    | MacOS
| Github Repo | https://github.com/staabm/infection-example/tree/283908c1d48554f5f28bd5c8417aa8843ae82ce2

after running `composer install` I want to run infection with a simple filter like `vendor/bin/infection --filter=User.php` but as a result I get

```
         1 test triggered 1 PHPUnit warning:                
                                                            
         1) FibonacciTest::testFibonacci
 Class Fibonacci is not a valid target for code coverage
 /Users/m.staab/dvl/infection-example/tests/FibonacciTest.php:8                                                 
         
OK, but there were issues!
Tests: 1, Assertions: 1, PHPUnit Warnings: 1.   
```

I don't see why this error happens and whats wrong in either the phpunit.xml or the sources involved

### Reproducer

```
git clone git@github.com:staabm/infection-example.git
cd infection-example
git checkout 283908c1d48554f
composer install
composer require --dev -W phpunit/phpunit:^12
```

Fails:
```
vendor/bin/infection --filter=src/Fibonacci.php
```
With:
```
1 test triggered 1 PHPUnit warning:                           
                                                              
1) UserAgeFilterTest::testFilter                              
Class UserAgeFilter is not a valid target for code coverage   
```

While the following does not:
```
vendor/bin/infection --filter=src/Fibonacci.php --test-framework-options="--filter='Fibonacci'"
```

Downgrade to PHPUnit 11:

```
composer require --dev -W phpunit/phpunit:^11
vendor/bin/infection --filter=src/Fibonacci.php
```
No longer fails.

sanmai:
Infection generates the following `phpunitConfiguration.initial.infection.xml`

```
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #2848, opened 2026-01-14
fix: Auto-add --covers option for PHPUnit 10+ when requireCoverageMetadata is true

sanmai:
Closes #2440.

**Note: this exploration was ultimately a** _dead end._ **Do not attempt to make it work, it does not.**

## Problem

When executing infection in a filtered mode (`--filter` or `--git-diff-lines`), infection will execute PHPUnit for the initial test run with the following configuration

<details>
<summary>phpunitConfiguration.initial.infection.xml</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/phpunit.xsd"
         bootstrap="/path/to/project/tests/e2e/PHPUnit_12-0/vendor/autoload.php"
         colors="false"
         executionOrder="defects,random"
         resolveDependencies="true"
         failOnRisky="true"
         failOnWarning="true"
         stopOnDefect="true"
         cacheResult="false"
         stderr="false">
    <testsuites>
        <testsuite name="Test Suite">
            <directory>/path/to/project/tests/e2e/PHPUnit_12-0/tests
            </directory>
        </testsuite>
    </testsuites>
    <source>
        <include>
            <file>
                /path/to/project/tests/e2e/PHPUnit_12-0/src/Uncovered/Calculator.php
            </file>
            <file>
                /path/to/project/tests/e2e/PHPUnit_12-0/src/Covered/Calculator.php
            </file>
        </include>
    </source>
</phpunit>

```

</details>

The `source` element is updated to include only the relevant source files.

When `requireCoverageMetadata=true` is enabled, PHPUnit will mark a test as risky if it does not indicate the code it intends to cover as an attribute. Another case not mentioned in the doc, is if the test does have a cover as an attribute, but the targetted code is not part of the PHPUnit sources.

However, in our case in the filtered mode, whilst we do restrict the sources for performance reasons (see https://github.com/infection/infection/pull/1543), we do not restrict the tests executed. This means some tests will be marked as risky:

```
1) Infection\E2ETests\PHPUnit_12_0\Tests\Covered\FunctionsTest::test_format_name_with_first_name_only
Function Infection\E2ETests\PHPUnit_12_0\Covered\formatName is not a valid target for code coverage
```

## Solution

This PR adds `--covers=SourceFQCN` for each filtered files when:
[...]

PR #2851, opened 2026-01-14
Add a tip in PhpUnitAdapter on how to handle "not a valid target for code coverage"

sanmai:
Ref #2440

## Description

Let's add a tip on how users can resolve the issue.

## Checklist

- [x] Tests added/updated
- [ ] Deal with escaping mutations


src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
```diff
@@ -210,4 +202,30 @@ public function getInitialRunOnlyOptions(): array
...
+    }
+
+    /**
+     * @return iterable<string>
+     * @see https://github.com/infection/infection/issues/2440
+     */
+    private function getPHPUnit12Recommendations(): iterable
+    {
+        yield "PHPUnit 12+ validates coverage targets strictly. If you see \"not a valid target for code coverage\" errors,\n"
+            . "narrow tests with --test-framework-options=\"--filter='ClassName'\"";
```
sanmai, line 229: I'm not hyper positive about this line; although it cost me literally nothing to add it (I extracted it from #2848) it could be seen as  debt: once we fix #2440 (and if we fix) this tip won't be required anymore.

Strip or keep, what do you think @theofidry?

theofidry: Sorry missed that comment. I need to give it some more thoughts, I would like to know if that's the appropriate tip. It is definitely a UX degradation and I would like to know what's the path to address it

theofidry:
Closed by https://github.com/infection/infection/pull/3043.

Issue #2358, opened 2025-08-05
RFC: Switch to Options Pattern


sanmai:
**Is your feature request related to a problem? Please describe.**

It pains immensely to add new command line toggles and configuration file options. Although the final diff is not long, the amount of code you had to sift through to get your new option through is immense. Few recent example PRs:

- https://github.com/infection/infection/pull/2216
- https://github.com/infection/infection/pull/2336

**Describe the solution you'd like**

I'd prefer if we use the Options Pattern:

> It's a design pattern where you create a dedicated class, often a simple data object, whose properties directly map to the structure and keys of your configuration file. The application then "binds" the configuration data to an instance of this class. Then these configuration option can be further overridden by  by other higher‑priority inputs such as CLI arguments.

Then we could dependency-inject this options object anywhere we need it, and eliminate the need to needle through just a specific option. This options object can be mocked as much as we need. And strongly-typed, too.

**Additional context**

I suppose the most straightforward would be to use `jms/serializer`  and although there are some gotchas with serializing `bool|Type` things and [using existing object](https://github.com/schmittjoh/serializer/blob/27f89514148f7adbdabb6dec559a5265db1f7a2b/doc/cookbook/object_constructor.rst), there are few [ways to handle them](https://gist.github.com/sanmai/db0db14ef49faa5a941ba3b0a0ddb866) too.

This change could be introduced incrementally. We can build the new `InfectionConfig` object and start using it for new features, while slowly refactoring existing parts to use it over time. This would avoid the need for a single, massive refactoring PR.

Using `symfony/config` and `symfony/options-resolver` did not make sense to me since we already have the JSON schema validation, and adopting these packages will double the effort required to add an option.

staabm:
Agree that adding a new option is a pain. 👍 

having a POC of the mentioned pattern would help to get everyone on the same page, whether it will improve our current state.

maks-rafalko:
> Agree that adding a new option is a pain. 👍

It is much less painful than it was before after this https://github.com/infection/infection/pull/2304

I agree PoC would be nice to see as I don't fully get the idea yet, but if it helps, why not?

One thing to underline here is we not only have CLI options, but we have a config file as well, so the same configuration setting can be

- declared via CLI option
- or declared via config
- or declared both in a config and via CLI option, in this case CLI option has a higher priority

And now Infection's configuration object has the final normalized value, so that later in the code we just do `$config->getSomeOption()`

PR #2857, opened 2026-01-15
refactor: Extract the options out of the RunCommand

theofidry:
## Description

When I introduced the debug git commands in #2669, I gave a go at extracting the options into a dedicated class for re-usability. Whilst it's not perfect, for our usage I am quite pleased with it and plan to continue to follow that pattern to extract re-usable options.

This PR extracts the options necessary for #2762 to reduce the noise.

## Changes

Unlike #2669, I took it one step further and added tests for them. Indeed we had basic tests in `RunCommandHelper` and it did not feel right to outright delete them, neither does it make sense to extract this code to re-use it if we can't even test it at all.

Testing a command option is not that difficult, but a bit boilerplaty. You need to create an `InputDefinition` object to which you can add `InputArgument` & `InputOption` objects. Then bind the used input to that definition and validate it.

There is an issue with this though... We actually do not create an `InputOption` object for our command, we directly use `Command::addOption()`, which does the create `InputOption` object and add it to its `InputDefinition`. Whilst we could adapt the code, I found it was simpler to just keep something that looked more similar to our source code hence I've leveraged a dummy command.


src/Command/RunCommand.php
```diff
@@ -151,9 +146,6 @@ final class RunCommand extends BaseCommand
     /** @var string */
     private const OPTION_LOG_VERBOSITY = 'log-verbosity';
 
-    /** @var string */
-    private const OPTION_INITIAL_TESTS_PHP_OPTIONS = 'initial-tests-php-options';
```
sanmai, line 155: My biggest concern is that options are all over the place now; say, an agent will have to do a whole lot of work to find where all options are defined. What if leave these constants in place, exported as public, and let the various option use these constants defined in a canonical location?

theofidry: I developed those commands with Claude, it did just fine at extracting, finding and re-use those.

>and let the various option use these constants defined in a canonical location?

I'm not convinced having a circular dependency (RunCommand depends on the option which depends on RunCommand and the option is what other commands depend on) would make it better for anyone, human or LLM.

It's not a system I would generalise and apply blindly: it is inflexible if you wish to change the name of the option for example. And it would be cleaner to make the return an `InputOption` (but then adding it to a command is quite awkward). But it's quite effective for those and an effective solution is needed if we want to share options names and validate them the same way across multiple commands.

`RunCommandHelper` was kinda fine to move some of the parsing onto, but I admit I wasn't found of how it was becoming a god class, and that was without declaring the option definition which is required here to re-use options elsewhere.

I would also imagine that if we go with something like you suggested in https://github.com/infection/infection/issues/2358 it would be a lot easier to generate such objects and re-use them.

PR #2868, opened 2026-01-20
feat(teamcity): Add a debug command

theofidry:
_This description has been heavily updated. This PR was originally a PoC for adding TeamCity logs support. Most of the work involved has been extracted to dedicated PRs since then._

Blocked by https://github.com/j-plugins/infection-plugin/issues/28.

<hr/>

Sometimes I want to test how the infection plugin behaves for a given TeamCity log. This command allows to easily do that. A few examples:

```shell
infection debug:mock-teamcity teamcity.log
# or
cat teamcity.log | infection debug:mock-teamcity
# or
infection debug:mock-teamcity < teamcity.log
```

PR #2879, opened 2026-01-23
refactor(logger): Move UNKNOWN_COUNT constant to IterableCounter

theofidry:
## Description

This PR documents the type of `$mutationCount`, which is passed around and consumed in multiple places, and relocates the `UNKNOWN_COUNT` constant to a more appropriate location.

### Problem

The `UNKNOWN_COUNT` constant is currently declared in `AbstractMutationAnalysisLogger`. This placement creates several dependency issues:

- `MutationAnalysisLogger` (the interface) must reference a constant from `AbstractMutationAnalysisLogger` (its implementation), violating the abstraction.
- `IterableCounter`, a generic utility, depends on `AbstractMutationAnalysisLogger`, an unrelated contract implementation.
- Any new `MutationAnalysisLogger` implementation that does not extend `AbstractMutationAnalysisLogger` would still need to depend on it to access the constant. This is not theoretical, it will occur with #2868.

## Changes

- Document `$mutationCount` as `positive-int|UNKNOWN_COUNT`. Technically the type is `positive-int|0`, but using the constant name is clearer and ensures the documentation remains accurate if `UNKNOWN_COUNT` is changed in the future.
- Move `UNKNOWN_COUNT` from `AbstractMutationAnalysisLogger` to `IterableCounter`, where it logically belongs as a generic counting utility constant.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3260,11 +3260,11 @@
         },
         {
             "name": "phpstan/phpstan",
-            "version": "2.1.36",
+            "version": "2.1.37",
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/phpstan/phpstan/zipball/2132e5e2361d11d40af4c17faa16f043269a4cf3",
-                "reference": "2132e5e2361d11d40af4c17faa16f043269a4cf3",
+                "url": "https://api.github.com/repos/phpstan/phpstan/zipball/28cd424c5ea984128c95cfa7ea658808e8954e49",
+                "reference": "28cd424c5ea984128c95cfa7ea658808e8954e49",
                 "shasum": ""
             },
             "require": {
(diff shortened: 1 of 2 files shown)


src/Framework/Iterable/IterableCounter.php
```diff
@@ -48,14 +47,21 @@ final class IterableCounter
 {
     use CannotBeInstantiated;
 
+    // To facilitate the usage with the console which is the primary consumer,
+    // we use the same value as for the Symfony ProgressBar where 0 stands for
+    // an unknown number of steps.
+    public const UNKNOWN_COUNT = 0;
```
sanmai: This wording is hard to parse. Possible alternative:

> Follows the Symfony ProgressBar convention: 0 indicates an unknown number of steps.

PR #2880, opened 2026-01-23
refactor!(logger): Remove redundant `$mutationCount` from `MutationAnalysisLogger::advance()`

theofidry:
## Description

While working on #2868, I was wondering what `$mutationCount` was in `MutationAnalysisLogger::advance()`. Turns out it is exactly the same value as `::start()`, hence is just redundant.

## Changes

Remove `$mutationCount` from `MutationAnalysisLogger::advance()`.

## Additional Notes

Technically a BC break, like #2877, but as explained there, not relevant.


maks-rafalko:
why is it a BC break? is it part of our public api?

theofidry:
>why is it a BC break? is it part of our public api?

`MutationAnalysisLogger` is, although not used/usable in practice, at least not without great gymnastics (see the "Risks" from https://github.com/infection/infection/pull/2877)

PR #2881, opened 2026-01-23
feat!(event,logger): Introduce new events

theofidry:
## Description

In order to implement https://github.com/infection/infection/issues/2446, we are missing two pieces of information:

- **When are we starting to evaluate a mutation** (i.e. the mutation was generated, and we will execute heuristics to try to suppress it and if not sufficient, materialize the mutant and evaluate it). This is necessary to dispatch the **TestStarted** TeamCity event.
- When did we evaluate _all_ the mutations  of a source file. This is necessary to dispatch the **TestSuiteFinished** TeamCity event.

## Changes

- Introduce the event `MutationEvaluationWasStarted`.
- Add the IDs of all the mutations to `MutableFileWasProcessed`.
- Make `MutationTestingConsoleLoggerSubscriber` subscribe to both `MutationEvaluationWasStarted` and `MutableFileWasProcessed`. The information got is forwarded to the logger.
- Update `MutationAnalysisLogger`:
  - Add two methods `::finishMutationGenerationForFile()` and `::startEvaluation()`. The former is to know when to dispatch **TestSuiteFinished**  and the latter for **TestStarted**.
  - Renamed the methods to remove some confusion. See the table bellow.

| Before | After | 
|--|--|
| start | startAnalysis |
|  | startEvaluation |
| advance | finishEvaluation |
| finish | finishAnalysis |

## Additional Notes

Technically a BC break, like https://github.com/infection/infection/pull/2877, but as explained there, not relevant.

### DIFF
--- a/src/Command/InitialTest/Option/InitialTestsPhpOptionsOption.php
+++ b/src/Command/InitialTest/Option/InitialTestsPhpOptionsOption.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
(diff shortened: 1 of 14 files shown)


src/Event/Events/MutationAnalysis/MutationGeneration/MutableFileWasProcessed.php
```diff
@@ -0,0 +1,50 @@
...
+final readonly class MutationGenerationForSourceFileWasFinished
+{
+    /**
+     * @param list<string> $sourceFileMutationIds
+     */
+    public function __construct(
+        public array $sourceFileMutationIds,
+    ) {
+    }
+}
```
maks-rafalko, line 51: in the plugin, for what purpose is it needed?

theofidry: Currently we have (in TeamCity terms) we have:

- TestSuite = a source file
- Test: a mutation of the source file

Currently the main issue is that we do not know when we processed all the mutations of a source file, hence when we can close the TestSuite.

It wasn't a problem in an earlier draft of the teamcity log because I was logging everything at the end of the execution, i.e. without real time reporting, which was sufficient to test things out but not good enough for a final product.

So with this event we could track what mutations the file has and once all the tests of that file have been finished, we can mark the test suite as finished.

maks-rafalko: you mean we will compare `sourceFileMutationIds` with finished processes for each mutant?

src/Event/Events/MutationAnalysis/MutationEvaluation/MutationEvaluationWasFinished.php
```diff
@@ -41,7 +41,7 @@
  * @internal
  * @final
  */
-class MutantProcessWasFinished
+class MutationEvaluationWasFinished
```
maks-rafalko: while we already merged an updated nomenclature, I feel like it's hard for me to (quickly) understand what the Evaluation is.

- MutantProcessWasFinished - clear, we know that a separate Process that we created for a Mutant has been finished
- MutantEvaluationWasFinished - even a day after reviewing nomenclature, I already don't know what is it. Especially when we have materialization + evaluation (2 steps related to a Process).

I would try to somehow keep the Process in these terms if it's possible.

What I would like to avoid is

- going to a nomenclature to understand the event name
- searching the place where it's dispatched to understand the event name

it should be clear just by it's name (as it was before)

src/Event/Events/MutationAnalysis/MutationGeneration/MutableFileWasProcessed.php
theofidry: Yes, mutant évaluation finished -> keep track of that and check if all the mutations of the file were processed

src/Event/Events/MutationAnalysis/MutationEvaluation/MutationEvaluationWasFinished.php
theofidry: Valid concerns.

>MutantProcessWasFinished - clear, we know that a separate Process that we created for a Mutant has been finished

Actually that was the point of confusion for me: it is also dispatched when the mutation is suppressed by heuristics, i.e. before we spawn it as a process. At that point it's already a mutant because I think it was simpler to do so, but technically that [suppressing a mutation with heuristics] could be done before we make it into a mutant.

But I hear you, and I want to make sure this is not creating more confusion. I'm cleaning up the TeamCity PoC PR first and validating I have all the pieces needed for this to work and I'll come back to this PR and review the naming too.

theofidry: I reverted it for now, there is too many events I would like to rename, and some weird stuff going on with the subscribers. I would also add more events for the telemetry.

But in the end, none of those things are strictly necessary to move forward so I stripped them down for now in order to be able to land the TeamCity feature ASAP.

Issue #2869, opened 2026-01-20
Nomenclature Proposal


theofidry:
_I still need to process the results and do a proposal... Writing this issue to not loose track of it._

I would like to write down the terminology used for the following phases:

1. Collecting the source files.
2. Gathering the required artefacts. Usually involves a coverage report for the
  source files.
3. Creating an AST for a source file.
4. Traversing the AST to enrich it: add context, mark dry nodes, mark eligible nodes, etc.
5. Traverse the AST to yield mutations for each eligible node.
6. Execute heuristics to attempt to suppress the mutation
7. When the mutation could not be suppressed by heuristics:
  A. Dump the mutant to the filesystem
  B. Create a mutant process with the mutated code
8. Execute tests against the mutant
9. Generate a report about the whole execution

And the following groups:

- Group1: 3) and 4)
- Group2: 5), 6), 7), 8)
- Group3: 7), 8)

Once we commit to those, it will be easier when introducing new services or to rename some events to follow those (we have a few missing events so it is a problem that will come sooner rather than later).

<details>

<summary>AI generated results</summary>

# Mutation Testing Phases Across Tools and Sources

| Phase | Stryker | Mutant (Ruby) | PIT (Pitest) | Google (ChatGPT) | Claude/Literature | Scientific Literature |
|-------|---------|---------------|--------------|------------------|-------------------|----------------------|
| Part 1 | Source File Resolution / Input Resolution / File Discovery | Scope Matching / Matchable Scopes Initialization / Namespace Discovery | Target Class Resolution / Class Path Scanning / Code Under Test Discovery | Source Discovery / Program Under Test Identification / Compilation Unit Collection | Source Discovery / Source Collection / File Enumeration | Program Corpus Collection / Source File Discovery / Compilation Unit Collection |
| Part 2 | Initial Test Run / Dry Run / Coverage Capture | Infection / Bootstrap / Integration Setup | Coverage Generation / Line Coverage Analysis / Test Coverage Gathering | Artefact Resolution / Coverage Data Acquisition / Test Context Preparation | Coverage Analysis / Artifact Collection / Prerequisites Gathering | Test Artefact Acquisition / Coverage Data Collection / Test-suite & Instrumentation Gathering |
[...]

PR #2887, opened 2026-01-24
PoC: refactor(event): Introduce dedicated event subscribers

theofidry:
Touching the event subscribers again recently, I would like to revisit two things.

This PR implements the all proposed changes to easily give a diff and save a bit of time, but not all of it needs to make it.

## Events grouping

I applied the changed for all of our current events to showcase it. Currently we do not have too many events so it's still ok, but I think we'll have 3 more events for TeamCity and then depending of how much we go with the Telemetry we'll get anywhere between 2 to 10 events (to capture the different phases, to track the time taken for it).

I do not fancy the idea of having a lot more events added without any structure, I find the current amount bearable but finding your way with more than that starts being annoying.

To address this, I propose to group them in directories, following the terminology choosen in https://github.com/infection/infection/issues/2869.

## Subscriber duck typing

I've not been very fond of the duck typed subscription methods:

```php
class MySubscriber implements EventSubscriber
{
    public function onInitialTestSuiteWasStarted(InitialTestSuiteWasStarted $event): void { ... }
}
```

I find it annoying that changing an event name means I need to do a search & replace in multiple places instead of being able to use the refactor function of an IDE.

While working with PHPUnit codebase, I encountered the simple idea of having a typed subscriber interface:

```php
interface InitialTestSuiteWasStartedSubscriber extends EventSubscriber
{
    public function onInitialTestSuiteWasStarted(InitialTestSuiteWasStarted $event): void;
}
```

Then your subscriber implements implement an interface method. You still need to update the method name if you change the event name (with that pattern), but at least the refactor function works.

Note that PHPUnit name all of the subscribers methods `notify()`, but IMO that turns out more annoying than anything if you want to implement multiple subscriber at once, which we do.

## Event names

Disclaimer: personally I've used the form `*WasStarted`, `*WasFinished` on private projects, so I don't have an issue with it.

But working recently on PHPUnit, I noticed it does omit the `Was` & so does Laravel (Symfony uses another form that fits neither, likewise for Doctrine).

Hence for the sake of it I'm putting it here as well. 

<hr/>

I wrote the points in order of importance (to me). I don't feel super strongly about any of them, but I think the first two would be nice. The last one I really don't care about, I proposed it for completeness.


theofidry:
Merged via:

- https://github.com/infection/infection/pull/2888
- https://github.com/infection/infection/pull/2889

PR #2889, opened 2026-01-25
feat(event): Introduce subscribers for each events

theofidry:
Extracted from #2887 for visibility.

Currently, declaring a subscribers requires blindly typing the method name:

```php
class MySubscriber implements EventSubscriber
{
    public function onInitialTestSuiteWasStarted(InitialTestSuiteWasStarted $event): void { ... }
}
```

It follows a convention, and the name must match. This was a system introduced in 4fc177817e90be4a291f43470f03e9fe8b5fb8b5 and it has done its job. It is however a bit annoying to not have a proper contract and updating the name of the event also involves more manual search & replace operations.

Recently working on the PHPUnit codebase, I encountered a simple pattern that help with that issue which I quite like. The idea is to declare a subscriber interface:

I find it annoying that changing an event name means I need to do a search & replace in multiple places instead of being able to use the refactor function of an IDE.

While working with PHPUnit codebase, I encountered the simple idea of having a typed subscriber interface:

```php
interface InitialTestSuiteWasStartedSubscriber extends EventSubscriber
{
    public function onInitialTestSuiteWasStarted(InitialTestSuiteWasStarted $event): void;
}
```

This PR introduces a subscriber for each event we have and leverages it.

### DIFF
--- a/src/Event/Subscriber/InitialStaticAnalysisRunConsoleLoggerSubscriber.php
+++ b/src/Event/Subscriber/InitialStaticAnalysisRunConsoleLoggerSubscriber.php
@@ -87,11 +87,6 @@ final readonly class InitialStaticAnalysisRunConsoleLoggerSubscriber implements 
         $this->progressBar->start();
     }
 
-    public function onInitialStaticAnalysisSubStepWasCompleted(InitialStaticAnalysisSubStepWasCompleted $event): void
-    {
-        $this->progressBar->advance();
-    }
-
     public function onInitialStaticAnalysisRunWasFinished(InitialStaticAnalysisRunWasFinished $event): void
     {
         $this->progressBar->finish();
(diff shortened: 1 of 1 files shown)


src/Event/Subscriber/CiInitialStaticAnalysisRunWasConsoleLoggerSubscriber.php
```diff
@@ -44,7 +45,7 @@
 /**
  * @internal
  */
-final readonly class CiInitialStaticAnalysisRunConsoleLoggerSubscriber implements EventSubscriber
+final readonly class CiInitialStaticAnalysisRunWasConsoleLoggerSubscriber implements InitialStaticAnalysisRunWasStartedSubscriber
```
theofidry: arg, invalid renames

theofidry:
Merging this as this was already discussed & agreed in #2887.

PR #2898, opened 2026-01-28
feat(logger): Add a TeamCity logger

theofidry:
## Description

Implement a basic TeamCity logger for the Infection PhpStorm plugin. A basic explanation about this piece is available in a `README.md` file.

There is still a few question marks, points to solve and features to add. However I wanted to reduce the scope to have a basic yet usable base to work on as a starting point.

Closes https://github.com/infection/infection/issues/2446.

## Changes

- Moved `Infection\Logger\ConsoleLogger` to `Infection\Logger\Console\ConsoleLogger` as a different console logger was needed. I hope we can change this in the future as I do not think we really need two of them, but I did not want this to be a blocker for this feature.
- Add the `TeamCityLogger` implementation for `MutationAnalysisLogger`.
- Besides being able to use it with `--formatter=teamcity`, since I suspect we may change the latter option at some point, I added `--teamcity`.

### DIFF
--- a/tests/phpunit/Mutant/MutantExecutionResultBuilder.php
+++ b/tests/phpunit/Mutant/MutantExecutionResultBuilder.php
@@ -231,7 +231,7 @@ final class MutantExecutionResultBuilder
     }
 
     /**
-     * @param Deferred<string> $mutantDiff
+     * @param Deferred<string>|string $mutantDiff
      */
     public function withMutantDiff(Deferred|string $mutantDiff): self
     {


src/Logger/MutationAnalysis/TeamCity/Test.php
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    private static function getExecutionDurationInMs(MutantExecutionResult $executionResult): string
+    {
+        // TODO: this is actually incorrect! Or is it?
+        //  this could be either the (singular) process, but what about:
+        //  - the other processes executed prior?
+        //  - the time taken by the heuristics?
+        //  - the waiting time in-between being generated and processes?
+        return (string) round($executionResult->getProcessRuntime() * self::MILLISECONDS_PER_SECOND);
```
theofidry, line 124: Replace the todo with a link to https://github.com/infection/infection/issues/2900

src/Logger/MutationAnalysis/TeamCity/FlowIdFactory.php
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Infection\Logger\MutationAnalysis\TeamCity;
+
+use function hash;
+use Infection\CannotBeInstantiated;
+
+/**
+ * @internal
+ */
+final class FlowIdFactory
```
theofidry: Open question: do we want to keep the "flow ID" terminology? It is the one explained in the docs but as we found out in practice for the plugin we use "node ID"

src/Logger/MutationAnalysis/TeamCity/NodeIdFactory.php
```diff
@@ -0,0 +1,53 @@
...
+ * @internal
+ */
+final class FlowIdFactory
+{
+    use CannotBeInstantiated;
+
+    public static function create(string $value): string
+    {
+        // Any hash which avoids collision, is fast and deterministic will do.
+        return hash('xxh3', $value);
```
theofidry, line 61: Dunno if xxh3 is the best here, I've had that suggested, I did not double check if there is better options.

src/Logger/MutationAnalysis/TeamCity/TeamCity.php
```diff
@@ -0,0 +1,213 @@
...
+{
+    // `|` must be escaped FIRST to avoid double-escaping.
+    private const CHARACTERS_TO_ESCAPE = ['|', "'", "\n", "\r", '[', ']'];
+
+    private const ESCAPED_CHARACTERS = ['||', "|'", '|n', '|r', '|[', '|]'];
+
+    private const UNICODE_CHARACTER_REGEX = '/\\\\u(?<hexadecimalDigits>[0-9A-Fa-f]{4})/';
+
+    public function __construct(
+        private bool $timeoutsAsEscaped,
```
theofidry, line 68: proposal: we could rename it as `treatTimeoutsAsEscaped`, WDYT @sanmai ?

src/Logger/MutationAnalysis/TeamCity/TeamCity.php
```diff
@@ -0,0 +1,213 @@
...
+            MessageName::TEST_COUNT,
+            ['count' => $count],
+        );
+    }
+
+    public function testSuiteStarted(TestSuite $suite): string
+    {
+        return $this->write(
+            MessageName::TEST_SUITE_STARTED,
+            $suite->toAttributes() + ['parentNodeId' => '0'],
```
theofidry, line 87: not entirely sure about this, I opened https://github.com/infection/infection/issues/2901 to ask for clarification

src/Logger/MutationAnalysis/TeamCity/TeamCity.php
```diff
@@ -0,0 +1,213 @@
...
+        );
+    }
+
+    public function testSuiteFinished(TestSuite $suite): string
+    {
+        return $this->write(
+            MessageName::TEST_SUITE_FINISHED,
+            $suite->toAttributes(),
```
theofidry, line 95: not entirely sure about this, I opened https://github.com/infection/infection/issues/2901 to ask for clarification

src/Logger/MutationAnalysis/TeamCity/Test.php
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * @return MessageAttributes
+     */
+    public function toAttributes(): array
+    {
+        return [
+            'name' => $this->name,
+            'nodeId' => $this->flowId,
+            'parentNodeId' => $this->parentFlowId,
```
theofidry: nodeId & parentNodeId are also passed to the finished message, not entirely sure if it is necessary I opened https://github.com/infection/infection/issues/2901 to ask for clarification

src/Logger/MutationAnalysis/TeamCity/Test.php
```diff
@@ -0,0 +1,130 @@
...
+        ];
+    }
+
+    /**
+     * @return MessageAttributes
+     */
+    public function toFinishedAttributes(MutantExecutionResult $executionResult): array
+    {
+        return $this->toAttributes() + [
+            // TODO: looks like this information is not used when the test is marked as successful or ignored :/
```
theofidry, line 90: opened https://github.com/infection/infection/issues/2902 to this effect; to add the todo here unless it gets sorted out since then

src/Logger/MutationAnalysis/TeamCity/TeamCity.php
```diff
@@ -0,0 +1,213 @@
...
+use Infection\Mutation\Mutation;
+use function is_int;
+use function is_string;
+use function Safe\preg_replace;
+use function sprintf;
+use function str_contains;
+use function str_replace;
+
+/**
+ * This service provides the primitives to write a TeamCity log record. It is
```
theofidry: not too happy about this comments, would like to do a review

src/Logger/MutationAnalysis/TeamCity/FlowIdFactory.php
maks-rafalko: > do we want to keep the "flow ID" terminology?

I wouldn't as it's confusing. The docs explaining `flowId` is irrelevant to us, I hope JB will document the approach we use at some point

src/Container/Container.php
```diff
@@ -1066,6 +1075,26 @@ public function getFileStore(): FileStore
...
 
+    /**
+     * TODO: should allow string as the ID
+     *
+     * @template T of object
+     *
+     * @param class-string<T> $id
+     * @param T $value
+     */
+    public function withService(string $id, object $value): self
```
maks-rafalko: where is it used?

tests/phpunit/Mutant/MutantExecutionResultBuilder.php
```diff
@@ -232,10 +233,12 @@ public function withDetectionStatus(DetectionStatus $detectionStatus): self
     /**
      * @param Deferred<string> $mutantDiff
      */
-    public function withMutantDiff(Deferred $mutantDiff): self
+    public function withMutantDiff(Deferred|string $mutantDiff): self
```
maks-rafalko, line 236: is this change required? it also conflicts with the phpdoc

maks-rafalko (approved):
did a brief code-review and test locally (don't have much time to dig deeper)

could you please point to a place how/where we update our loggers so that they don't produce any messages in the middle of the TeamCity logs? (that issues that we discussed today)

theofidry:
@maks-rafalko this is not done yet. I'll open an issue and address this separately.

Otherwise I would block this PR to address that point first, but I would like to be able to address other points in parallel too besides that problem.

src/Container/Container.php
theofidry: Nowhere, it's a leftover that should have been removed good catch :)

src/Logger/MutationAnalysis/TeamCity/FlowIdFactory.php
theofidry: Removed.

src/Logger/MutationAnalysis/TeamCity/TeamCity.php
theofidry: created https://github.com/infection/infection/issues/2906 to this effect, will do that later

tests/phpunit/Mutant/MutantExecutionResultBuilder.php
theofidry: Yes and no.

We could do without, but the goal of those test builders is to make things easier, it is more convenient to pass a string in some tests than a `Deferred` object.


>it also conflicts with the phpdoc

It is curious that PHPStan did not catch that.

PR #2910, opened 2026-01-29
build(ci): Use cached apt action for expect package

sanmai:
## Description

Switch from apt-get install to `awalsh128/cache-apt-pkgs-action` for installing the expect package in E2E tests. This caches the installed package across workflow runs.

I also added `chronic` so that we would not have to fish for one failed dependency install next time.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -2018,16 +2018,16 @@
         },
         {
             "name": "symfony/process",
-            "version": "v7.4.3",
+            "version": "v7.4.5",
             "source": {
                 "type": "git",
                 "url": "https://github.com/symfony/process.git",
-                "reference": "2f8e1a6cdf590ca63715da4d3a7a3327404a523f"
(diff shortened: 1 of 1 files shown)


sanmai:
I also added `chronic` so that we would not have to fish for one failed dependency install next time:

Screenshot From 2026-01-29 19-19-10

The successful install looks like this:

Screenshot From 2026-01-29 19-38-45

maks-rafalko:
> I also added `chronic` so that we would not have to fish for one failed dependency install next time.

I don't see it in the diff

sanmai:
My bad, should be there now.

theofidry:
@sanmai I don't get it, what was the cause of the failure?

sanmai:
@theofidry https://github.com/infection/infection/pull/2912 fixed it, and https://github.com/infection/infection/pull/2913 is a follow up

PR #2912, opened 2026-01-29
Update PHPUnit_12-0 E2E test to permit PHPUnit versions with CVE

sanmai:
## Description

Should fix the build failures.


sanmai:
I wonder if 12.3.x will ever get released with a patch.

PR #2913, opened 2026-01-29
Remove PHPUnit_12-0 E2E as PHPUnit 12.4 no longer installable

sanmai:
## Description

PHPUnit >=12.0.0,<12.5.8 has [CVE-2026-24765](http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2026-24765) so users can't install it with Composer. 

As an alternative to #2912, let's remove the tests case altogether.

The PHPUnit_12-0 test was intended to verify compatibility with early PHPUnit 12.x versions (12.0-12.3). However:    
                                                                                                                     
1. CVE-2026-24765 affects PHPUnit versions before 12.5.8, making these versions unsuitable for testing               
2. The test is functionally identical to PHPUnit_12-5 (only namespace differences)                                   
3. No breaking changes between PHPUnit 12.0 and 12.5 require separate compatibility testing                          
                                                                                                                     
The PHPUnit_12-5 test provides sufficient PHPUnit 12.x coverage.                                                     

**That said, I wonder if 12.3.x will ever get released with a patch. If it happens, we still need this test.**

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -55,6 +55,22 @@ jobs:
           ini-values: memory_limit=512M, xdebug.mode=off
           tools: composer
 
+      - name: Get composer cache directory
+        id: composer-cache
+        shell: bash
+        run: echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
+
+      - name: Cache dependencies
+        uses: actions/cache@8b402f58fbc84540c8b491a91e594a4576fec3d7 # v5.0.2
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: composer-${{ runner.os }}-${{ matrix.php-version }}-${{ matrix.dependencies }} }}
+          restore-keys: |
+            composer-${{ runner.os }}-${{ matrix.php-version }}-${{ matrix.dependencies }} }}-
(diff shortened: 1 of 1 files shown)


sanmai:
@maks-rafalko What do you think about this argument?

sanmai:
@theofidry You added this test, so it appears to me it should be your call as to if nuke it. 

I merged #2912 meanwhile. Think worth keeping?

theofidry:
>https://github.com/advisories/GHSA-vvj3-c3rp-c85p affects PHPUnit versions before 12.5.8, making these versions unsuitable for testing

Noteworthy: this vulnerability does not affect us in this scenario.

>The test is functionally identical to PHPUnit_12-5 (only namespace differences)
>No breaking changes between PHPUnit 12.0 and 12.5 require separate compatibility testing

Whilst there is no breakage from our side, there is actually a difference in the output of the coverage reports, hence I'd be in favour of keeping it. Our coverage implementation does. not care, but it doesn't leverage the new changes introduced in there yet either.

So I think the solution of https://github.com/infection/infection/pull/2912 is more adequate.

There is an argument to be made about potential redundant coverage reports, but we'll touch on this in #2815, as this is beyond just PHPUnit 12.0.

sanmai:
Let's keep it :+1:

Issue #2911, opened 2026-01-29
feat: Improve the TeamCity diff output


theofidry:
As per https://github.com/infection/infection/issues/2903, currently we log the diff in the test message, which makes the colouring of it awkward.

Turns out there is a nicer way which is to use the `expected` & `actual` attributes:

```
##teamcity[testFailed nodeId='2.A' message='Failed asserting that two strings are equal.' details='/Users/xepozz/IdeaProjects/j-plugins/infection-plugin/playground/tests/BirdTest.php:28|n' duration='14' type='comparisonFailure' actual='|'hello\\n|nI|'m from\\n|ntest|'' expected='|'hello\\n|nfrom\\n|ntest|'']
```

theofidry:
I gave it a look  and I have doubts.

Currently what we are passing to the message is `Mutant#diff`, which is the computed diff. As a quick experiment, I updated the non-successful test closing messages (`testFailed` & `testIgnored`) with:

```
type='comparisonFailure'
expected='<sourceFileCode>'
actual='<mutatedCode>'
```

Having a look at the raw teamcity log the problem is obvious: this is passing the _entire_ file, so it is an awful amount of text passed for each log record.

It would not be hard to adjust the size to show only say 5 lines before the first modified line and 5 lines after the last modified line, but I actually found something I like about having the whole file:

<img width="1579" height="759" alt="Image" src="https://github.com/user-attachments/assets/99ae9eef-1fa0-4ba2-8efb-7b0cfbe4fd76" />

You can very easily identify which piece of code is affected, you can get the function name & co. and the right line number.

Maybe it's a silly idea, but I'm wondering if it would be possible to let the plugin get the source file code (as it would need to do it once – scoped to the run ofc, as another run might change the code of the file and we don't want it to live in-memory forever either) which would avoid passing it to the log, and that for the 

And then for the mutated code, we could pass the diff with the line affected, and computing the modified code could be done by the plugin too.

But maybe we do not have that much control on the plugin side or maybe it's too error-prone or a stupid idea...

xepozz:
This is nice idea, I'll try to look if it's possible now.

I would suggest to use "short diff" as Infection drop it in the console output in the expect/actual properties for now.
If we did this "real file comparison" we still will use this patch to apply changes to the final version of source code.

"short diff" example:
<img width="3018" height="704" alt="Image" src="https://github.com/user-attachments/assets/76a0580a-e43e-4d0c-8d1d-63b97d99f92e" />

[...]

PR #2918, opened 2026-01-30
feat(teamcity): Let PhpStorm propose the diff for mutations

theofidry:
Related to #2911.

### DIFF
--- a/src/Logger/MutationAnalysis/TeamCity/TeamCity.php
+++ b/src/Logger/MutationAnalysis/TeamCity/TeamCity.php
@@ -138,11 +138,8 @@ final readonly class TeamCity
             $attributes['expected'] = $to;
         }
 
-        return $this->write(
-            /** @psalm-suppress InvalidArgument */
-            $messageName,
-            $attributes,
-        );
+        /** @psalm-suppress InvalidArgument */
+        return $this->write($messageName, $attributes);
     }
 
     /**


src/Logger/MutationAnalysis/TeamCity/TeamCity.php
```diff
@@ -122,9 +124,24 @@ public function testFinished(
...
     ): string {
         /** @psalm-suppress InvalidArgument */
+        $messageName = $this->mapExecutionResultToTestStatus($executionResult);
+        $attributes = $test->toFinishedAttributes($executionResult);
+
+        if ($messageName !== MessageName::TEST_FINISHED) {
+            unset($attributes['details']);
+            [$from, $to] = $this->differ->diffToArray(
+                from: $executionResult->getMutatedCode(),
+                to: $executionResult->getOriginalCode(),
```
maks-rafalko: I didn't test it, but just from the names of these variables: shouldn't it be the other way around, where `from` is the original code and `to` is the mutated code?

tests/phpunit/Differ/DifferTest.php
```diff
@@ -98,6 +121,24 @@ public function echo(): void
...
+                    PHP,
+                <<<'PHP'
+
+                    public function echo(): void
+                    {
+                        echo 15;
+                    }
+
+                    PHP,
+            ],
```
maks-rafalko, line 141: not sure it's a good idea to mix these 2 tests with the same provider, because

- `string $_expectedStringDiff,` is not used in one test
- `array $expected,` is not used in another test. And which is worse - it needs to be added. So imagine in the future you will update only `test_it_shows_the_diff_between_two_sources_but_limiting_the_displ` test method, adding/updating it's data provider. It will be super confusing to have to add the fourth parameter `$expected` in the data provider, when it's not used.

WDYT?

src/Logger/MutationAnalysis/TeamCity/TeamCity.php
theofidry: This is how it looks right now:

Screenshot 2026-02-12 at 19 37 47

My reasoning was the expectation is that no code is mutated hence, hence expected = original code.

I am fine with inverting it if you think it's better though, I don't really have an opinion there

tests/phpunit/Differ/DifferTest.php
theofidry: >So imagine in the future you will update only test_it_shows_the_diff_between_two_sources_but_limiting_the_displ test method

Actually this is precisely what I wanted to avoid here.

At least when I implemented it, I wanted to cover all the scenarios we were using for `diff()`: if there is a scenario worthy of having in diff, I didn't see why it shouldn't be added to the `diffToArray()`.

That being said, whilst it was true at the time I implemented, I don't feel super strongly of it being the case in the future, so I don't mind splitting it if you prefer.

src/Logger/MutationAnalysis/TeamCity/TeamCity.php
maks-rafalko: > I am fine with inverting it if you think it's better though, I don't really have an opinion there

when you compare file A with B, and B with A - you will get completely different diff. So yeah I think it's more logical if we compare original code with a mutant, so

- `from` - this is an original code
- `to` - is a mutant

tests/phpunit/Differ/DifferTest.php
maks-rafalko: ok let's leave as is, thanks

PR #2919, opened 2026-01-30
refactor(teamcity): Improve the type of TeamCity::mapExecutionResultToTestStatus()

theofidry:
Minor refactor to make the return type `TeamCity::mapExecutionResultToTestStatus()` more accurate.

### DIFF
--- a/src/Logger/MutationAnalysis/TeamCity/TeamCity.php
+++ b/src/Logger/MutationAnalysis/TeamCity/TeamCity.php
@@ -122,6 +122,7 @@ final readonly class TeamCity
         MutantExecutionResult $executionResult,
     ): string {
         return $this->write(
+            /** @psalm-suppress InvalidArgument */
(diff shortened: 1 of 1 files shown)


src/Logger/MutationAnalysis/TeamCity/TeamCity.php
```diff
@@ -147,6 +148,11 @@ public function write(
         );
     }
 
+    /**
+     * @psalm-suppress InvalidReturnType,InvalidReturnStatement
```
theofidry, line 153: unfortunately even with `@psalm-type` and `@psalm-import-type` Psalm does not seem to understand this bit

maks-rafalko: then do we need to keep psalm type / import type?

theofidry: there isn't, only the phpstan ones remain

PR #2921, opened 2026-01-30
test: Use the same differ as configured for the application for DifferTest

theofidry:


src/Container/Container.php
```diff
@@ -1140,11 +1145,6 @@ private function getProcessRunner(): ProcessRunner
         ;
     }
 
-    private function getDiffer(): Differ
-    {
-        return $this->get(Differ::class);
-    }
-
```
maks-rafalko, line 1147: is this change needed (moving the method)?

theofidry: it is done by the CS fixer

maks-rafalko: how come? why didn't it do it before?

theofidry: because it was private, we have some visibility based ordering (not sure the exact order)

PR #2923, opened 2026-01-31
refactor: Allow to override a Container service on the fly

theofidry:
Follow through with the idea from https://github.com/infection/infection/pull/2868#discussion_r2732940983 which was also attempted in https://github.com/infection/infection/pull/2837.

I didn't manage to make it work for the Tracer yet as this requires to have a configuration file ready too, which I didn't really want to do (maybe we could have it optional at some point?), so I left it out for now.

The idea is to be able to selectively override a service, allowing us to still use the Container to get a specific service instead of instantiating it ourselves. On one hand this makes those tests more brittle as Container changes have a wider impact, but it also helps with having those integration tests behaviour closer to how our application is configured.

### DIFF
--- a/src/Container/Container.php
+++ b/src/Container/Container.php
@@ -145,7 +145,6 @@ use Infection\TestFramework\Coverage\JUnit\JUnitTestExecutionInfoAdder;
 use Infection\TestFramework\Coverage\JUnit\JUnitTestFileDataProvider;
 use Infection\TestFramework\Coverage\JUnit\MemoizedTestFileDataProvider;
 use Infection\TestFramework\Coverage\JUnit\TestFileDataProvider;
-use Infection\TestFramework\Coverage\Locator\ReportLocator;
 use Infection\TestFramework\Coverage\XmlReport\IndexXmlCoverageLocator;
 use Infection\TestFramework\Coverage\XmlReport\IndexXmlCoverageParser;
 use Infection\TestFramework\Coverage\XmlReport\PhpUnitXmlCoverageTraceProvider;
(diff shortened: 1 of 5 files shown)


src/Container/Container.php
sanmai, line 1175: What made you go away from class name service ids?

src/Container/Container.php
```diff
@@ -1080,6 +1082,25 @@ public function getDiffer(): Differ
...
+     * @param T|(Closure(static): T) $value
+     */
+    public function withService(string $id, object $value): self
+    {
+        $clone = clone $this;
+
+        if ($value instanceof Closure) {
+            $clone->offsetSet($id, $value);
+        } else {
+            $clone->inject($id, $value);
```
sanmai, line 1097: That's not great, I'll fix this

theofidry: WDYM?

I think it's ok for the Container to not make assumption on the usage. Currently `::withValues()` does not want an immutable container for example.

That said maybe it better belongs to `SingletonContainer` which is for test purposes only

src/Container/Container.php
theofidry: It was for the tracer as per https://github.com/infection/infection/pull/2837. For example `IndexXmlReportLocator` right now is a concrete class, but the contract use is `ReportLocator` and for there I would swap it for a `FixedLocator` instance, so a service string ID is necessary

src/Container/Container.php
theofidry: Ah nvm `SingletonContainer` won't work, ok I'll rename it to `cloneWithService` then

sanmai (approved):

PR #2926, opened 2026-02-01
feat(ast): Introduce a NodeAnnotator

theofidry:
## Description

This PR proposes an alternative approach for declaring the node enrichment API within the respective visitors. It introduces a `NodeAnnotator` and `Annotation` enum to:

- Provide an enum for type-safe manipulation of node attributes, replacing arbitrary strings that may be defined across multiple locations.
- Centralise node manipulation into a single location.

## Motivation

We use various visitors to enrich the traversed nodes, including both our own implementations and third-party visitors such as `ParentConnectingVisitor`.

A challenge when working within a mutator is the lack of discoverability regarding which information is available on a node. Whilst it is possible to inspect the attributes by dumping them, this approach is neither straightforward nor efficient. Additionally, some information is inferred from the absence of an attribute, for example `::hasNextNode()` or `::isInsideFunction()`.

With `Annotation`, the exhaustive list of attributes becomes immediately visible. Via `NodeAnnotator`, a single location provides a rich API for manipulating nodes, rather than requiring lookups across multiple visitors. (Note that not all functionality resides within visitors; for example, `ParentConnector` is separate from `ParentConnectingVisitor`, as PHP-Parser provides the latter.)

## Changes

**Before:**

```php
// Setting attributes
// NextConnectingVisitor.php
$this->previous?->setAttribute(NextConnectingVisitor::NEXT_ATTRIBUTE, $node);
// ReflectorVisitor.php
$node->setAttribute(IS_STRICT_TYPES::STRICT_TYPES_KEY, $this->isDeclareStrictTypes);

// Getting attributes
// NextConnectingVisitor.php
NextConnectingVisitor::hasNextStmtNode($node);
// ReflectorVisitor.php
ReflectionVisitor::isStrictTypesEnabled($functionScope)
```

**After:**

```php
// Setting attributes
NodeAnnotator::annotate($this->previous, Annotation::NEXT_NODE, $node);
NodeAnnotator::annotate($node, Annotation::IS_STRICT_TYPES, $this->isDeclareStrictTypes);

// Getting attributes
NodeAnnotator::hasNextStmtNode($node);
NodeAnnotator::isStrictTypesEnabled($functionScope)
```

## Related issues

### DIFF
--- a/src/PhpParser/Metadata/NodeAnnotator.php
+++ b/src/PhpParser/Metadata/NodeAnnotator.php
@@ -60,7 +60,7 @@ final class NodeAnnotator
         Annotation $annotation,
         mixed $value = null,
     ): void {
-        $node?->setAttribute($annotation->value, $value ?? $annotation->value);
+        $node?->setAttribute($annotation->value, $value ?? true);
     }
 
     public static function getParent(Node $node): Node


tests/phpunit/PhpParser/Visitor/NameResolverTest.php
```diff
@@ -0,0 +1,213 @@
...
+                        nodeId: 8
+                    )
+                )
+                AST,
+        ];
+    }
+
+    public function test_it_can_provide_the_node_fqcn(): void
+    {
+        $fqcn = new FullyQualified('Acme\Foo');
```
theofidry, line 202: to rework those tests using the node IDs

tests/phpunit/Mutator/MutatorFactoryTest.php
```diff
@@ -179,8 +179,8 @@ private function createBoolNode(string $boolean, string $functionName, ClassRefl
         return new Node\Expr\ConstFetch(
             new Node\Name($boolean),
             [
-                ReflectionVisitor::REFLECTION_CLASS_KEY => $reflectionMock,
-                ReflectionVisitor::FUNCTION_NAME => $functionName,
+                Annotation::REFLECTION_CLASS->value => $reflectionMock,
```
theofidry, line 182: see if that could be made better, maybe `::annotate()` can be adapted

tests/phpunit/Mutator/Util/NameResolverTest.php
theofidry, line 46: to check what this `NameResolver` is about

tests/phpunit/Mutator/Util/ReturnTypeHelperTest.php
```diff
@@ -49,7 +49,7 @@ public function test_it_detects_default_attributes(): void
     {
         $returnNode = new Node\Stmt\Return_();
         $returnNode->setAttribute(
-            ReflectionVisitor::FUNCTION_SCOPE_KEY,
```
theofidry, line 52: should use the annotator here; check the other places too, `Annotation::*->value` should not really be used outside the `NodeAnnotator`

src/PhpParser/Visitor/ReflectionVisitor.php
```diff
@@ -52,16 +54,6 @@
...
-
-    public const REFLECTION_CLASS_KEY = 'reflectionClass';
-
-    public const IS_INSIDE_FUNCTION_KEY = 'isInsideFunction';
-
-    public const IS_ON_FUNCTION_SIGNATURE = 'isOnFunctionSignature';
-
-    public const FUNCTION_SCOPE_KEY = 'functionScope';
-
     public const FUNCTION_NAME = 'functionName';
```
theofidry: check why `FUNCTION_NAME` is still being used here

tests/phpunit/PhpParser/Visitor/NameResolverTest.php
theofidry: https://github.com/infection/infection/pull/2973

theofidry:
@sanmai @maks-rafalko let me know what you think.

It's a pattern that came out as I worked more on the AST in https://github.com/infection/infection/pull/2795 and others, but I don't feel like this is a definite big win either. So if you do not like it/ don't see much the point do say so.

https://github.com/infection/infection/pull/2795 doesn't really require this PR, but it uses this pattern, so I would like to avoid unnecessary rebases here.

theofidry:
>probably except that NodeAnnotator becomes a god object

Yes, it certainly would. Limited to whatever state we put in the node attributes though.

Issue #2866, opened 2026-01-18
RFC: From Loggers to Reporters


theofidry:
## Summary

This RFC proposes refactoring the Logger/Reporter subsystem to address code duplication, Single Responsibility violations,
and confusing naming in the current implementation.

## Motivation

While investigating the reports subsystem to add new reporters, I found the current architecture difficult to navigate.
The mixing of concerns and inconsistent naming make it harder than necessary to understand or extend.

> [!NOTE]
> This is not critical, nor difficult to fix. This RFC documents the identified issues and proposes a straightforward
solution.

## Goals

The goal of this RFC is to improve the naming and organisation of the `src/Logger` code. It should make it easy:

- to understand the code without navigating between multiple files
- to update shared information without duplicating changes across reporters
- to add new output destinations (e.g., stdout instead of a file)
- to create new reporters
- to test individual components, independently of the destination

### Non-Goals

- Modifying the `ConsoleLogger` (PSR logger for runtime diagnostics).

## Current solution

Here is a diagram of the current solution:

<img width="8191" height="2736" alt="Image" src="https://github.com/user-attachments/assets/4b2cd2bc-2422-4fcb-9ff4-1a391db02013" />

<details>

<summary>Current solution mermaid code</summary>

```mermaid
classDiagram
direction BT

%% Interfaces
class MutationTestingResultsLogger {
    <<interface>>
}
class LineMutationTestingResultsLogger {
    <<interface>>
}

%% Abstract classes
class BaseTextFileLogger {
    <<abstract>>
}

%% External (PSR)
class AbstractLogger["Psr\Log\AbstractLogger"] {
    <<external>>
}

%% Root level classes
class DebugFileLogger
class FederatedLogger
class FileLogger
class FileLoggerFactory
class PerMutatorLogger
class TextFileLogger


namespace Stryker {
    class StrykerLogger
    class StrykerLoggerFactory
    class StrykerCurlClient
    class StrykerDashboardClient
    class StrykerHtmlReportBuilder
    class Response
}

%% Logger (PSR-like)
namespace Logger {
    class AbstractLogger
    class ConsoleLogger
}

namespace GitHub {
    class GitHubAnnotationsLogger
    class GitHubActionsLogTextFileLogger
}

namespace Summary {
    class SummaryFileLogger
    class SummaryJsonLogger
}

namespace Html {
    class HtmlFileLogger
}

namespace Json {
    class JsonLogger
}

namespace GitLab {
[...]

PR #2932, opened 2026-02-04
refactor(reporter): Rename the "loggers" to reporters

theofidry:
## Description

This PR renames the currently named loggers into reporters. For a breakdown of the motivations and rationale, please checkout https://github.com/infection/infection/issues/2866.

I originally intended to leave them be for longer, but due to the Teamcity backlog I'm diving deeper into the subscribers/loggers and having those "fake loggers" in the middle just keep making things confusing. Nonetheless, it remains not my priority, so I kept the changes to a minimum.

## Changes

- `MutationTestingResultsLogger` -> `Reporter`
- `LineMutationTestingResultsLogger::getLogLines()` -> `LineMutationTestingResultsReporter::getLines()`
- Change of the namespace

## Related issues

https://github.com/infection/infection/issues/2866

### DIFF
--- a/devTools/phpstan-baseline.neon
+++ b/devTools/phpstan-baseline.neon
@@ -151,36 +151,6 @@ parameters:
 			path: ../src/Logger/Console/ConsoleLogger.php
 
 		-
-			rawMessage: 'Method Infection\Logger\Html\StrykerHtmlReportBuilder::build() return type has no value type specified in iterable type array.'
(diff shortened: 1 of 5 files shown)


sanmai (approved):
So we are looking at what they produce instead of what they consume. Makes sense.

theofidry:
Thanks a nice way to put it :)

PR #2935, opened 2026-02-05
refactor(event): Widen reporter type to Reporter in console logger subscriber

theofidry:
For some strange reasons, we forced `MutationTestingConsoleLoggerSubscriber#reporter` to be an instance of `FederatedReporter`. This action, besides being unnecessary, was causing some type contortions.

I plan to refactor this piece of code to move it out of  `MutationTestingConsoleLoggerSubscriber`, so in the future the reporter instance injected will not be a `FederatedReporter` either.

This PR changes `#reporter` to be a `Reporter` instead which also allows to fix a few issues that were reported by PHPStan (and sometimes silence via hard overriding the declared type with phpdoc).

I had a cursory look and this was the only case I could find.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3812,16 +3812,16 @@
         },
         {
             "name": "phpunit/phpunit",
-            "version": "11.5.50",
+            "version": "11.5.51",
             "source": {
                 "type": "git",
                 "url": "https://github.com/sebastianbergmann/phpunit.git",
-                "reference": "fdfc727f0fcacfeb8fcb30c7e5da173125b58be3"
+                "reference": "ad14159f92910b0f0e3928c13e9b2077529de091"
             },
             "dist": {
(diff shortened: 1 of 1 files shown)


sanmai (approved):

src/Event/Subscriber/MutationTestingConsoleLoggerSubscriber.php
```diff
@@ -91,7 +91,7 @@ public function __construct(
         private readonly MetricsCalculator $metricsCalculator,
         private readonly ResultsCollector $resultsCollector,
         private readonly DiffColorizer $diffColorizer,
-        private readonly FederatedReporter $reporter,
+        private readonly Reporter $reporter,
```
theofidry, line 94: I've just noticed, some tests can be simplified as a result

PR #2927, opened 2026-02-01
poc: try a different logger/subscriber configuration

theofidry:
Trying to fix https://github.com/infection/infection/issues/2907.

Goal: outputting information during the execution? Loggers. Maybe the logger can be interactive, be a progress bar, be silent, or be dedicated to TeamCity or other. One time information about the execution result? Reporter.

## Changes

- Adds/Fixes/Updates ...

## Checklist

- [ ] Tests added/updated
- [ ] Documentation updated (provide link to PR: https://github.com/infection/site/pull/XXX)
- [ ] CHANGELOG.md updated (if deprecations/breaking changes)
- [ ] Appropriate labels applied (e.g. `performance`, `feature`).

## Related issues

Fixes https://github.com/infection/infection/issues/XXX.

This PR:

- [ ] Adds new feature ...
- [ ] Covered by tests
- [ ] Doc PR: https://github.com/infection/site/pull/XXX

Fixes https://github.com/infection/infection/issues/XXX

PR #2937, opened 2026-02-05
refactor(reporter): Move the show generated report files to a dedicated reporter

theofidry:
Moves out the logic of `MutationTestingConsoleLoggerSubscriber::showGeneratedLogFiles()` to `FileLocationReporter`. The implementation has been kept as close as possible to the original to ease the review, only the tests were a bit more heavily edited to ensure we cover all cases of `FileLocationReporter`.

### Related

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3260,11 +3260,11 @@
         },
         {
             "name": "phpstan/phpstan",
-            "version": "2.1.38",
+            "version": "2.1.39",
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/phpstan/phpstan/zipball/dfaf1f530e1663aa167bc3e52197adb221582629",
-                "reference": "dfaf1f530e1663aa167bc3e52197adb221582629",
+                "url": "https://api.github.com/repos/phpstan/phpstan/zipball/c6f73a2af4cbcd99c931d0fb8f08548cc0fa8224",
+                "reference": "c6f73a2af4cbcd99c931d0fb8f08548cc0fa8224",
                 "shasum": ""
             },
             "require": {
@@ -3309,20 +3309,20 @@
                     "type": "github"
                 }
             ],
-            "time": "2026-01-30T17:12:46+00:00"
+            "time": "2026-02-11T14:48:56+00:00"
         },
         {
             "name": "phpstan/phpstan-phpunit",
-            "version": "2.0.14",
+            "version": "2.0.15",
             "source": {
                 "type": "git",
                 "url": "https://github.com/phpstan/phpstan-phpunit.git",
-                "reference": "f7553d6c613878d04f7e7ef129d4607118cd7cd4"
+                "reference": "0f6ba8e22a3e919353a8f4615ea3cd2e91626c7f"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/phpstan/phpstan-phpunit/zipball/f7553d6c613878d04f7e7ef129d4607118cd7cd4",
-                "reference": "f7553d6c613878d04f7e7ef129d4607118cd7cd4",
+                "url": "https://api.github.com/repos/phpstan/phpstan-phpunit/zipball/0f6ba8e22a3e919353a8f4615ea3cd2e91626c7f",
+                "reference": "0f6ba8e22a3e919353a8f4615ea3cd2e91626c7f",
                 "shasum": ""
             },
             "require": {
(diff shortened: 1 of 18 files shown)


src/Event/Subscriber/MutationTestingConsoleLoggerSubscriber.php
```diff
@@ -145,7 +141,7 @@ public function onMutationTestingWasFinished(MutationTestingWasFinished $event):
         }
 
         $this->showMetrics();
-        $this->showGeneratedLogFiles();
+        $this->reporter->report();
```
maks-rafalko, line 144: isn't it much less readable than before?

my thoughts

- what is a "reporter" here?
- why does it report here after a good named `showMetrics()`?
- ...and before other `writeln()`

src/Reporter/FileLocationReporter.php
```diff
@@ -0,0 +1,103 @@
...
+
+use Generator;
+use function sprintf;
+use function str_repeat;
+use function str_starts_with;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Decorator that generates a list of generated report file locations to the console.
+ * It does NOT execute the decorated report.
```
maks-rafalko, line 46: sounds like.. potential issue with the chosen pattern, no?

- is it really a decorator then?

The main question: what is the purpose of Reporters? If we have this interface for our FILE reporters, then I think this particular `FileLocationReporter` does not suit this architecture.

If `Reporter` interface is not only for file reporters, then could you please describe the idea of a reporter?

UPD: found https://github.com/infection/infection/issues/2866, still questions are relevant

theofidry: >>It does NOT execute the decorated report.

>sounds like.. potential issue with the chosen pattern, no?
> is it really a decorator then?

Yes agreed. And IMO it _is_ currently an issue. Tbh I didn't even realise this until I worked on the description of the service. I totally though it was it doing it until then.

theofidry:
@maks-rafalko I'll reply to it all here because I think it's actually all related. And apologies, I realise I did do a PoC for myself get the direction but failed to communicate it here.

>The main question: what is the purpose of Reporters?  [...] If Reporter interface is not only for file reporters, then could you please describe the idea of a reporter?

I do hope https://github.com/infection/infection/issues/2866 gave some clarifications, but to repeat some pieces of information, my idea of a reporter is a service that can process the execution information to deliver different pieces of information to different supports. Just to be clear, it is nothing new, it is what our previously named "Loggers" were doing. We provides a summary of the execution, a detailed breakdown or a more or less detailed description of the escaped mutations to different supports (a text file, a JSON file, the stdout, an remove service like Stryker).

For clarity between us, I'll use "reporter" now, even if I'm talking of a them at the time they were called "loggers".

To that effect, what currently `MutationTestingConsoleLoggerSubscriber` does is:

- call the `MutationAnalysisLogger`
- do some reporting:
  - show some mutations to the stdout
  - show execution metrics to the stdout
  - show generated reports (which btw are not generated yet at the point it is called – that part is currently done in `MutationTestingResultsLoggerSubscriber` which is executed after)

From what I see the reporters were introduced later, or at least at the time `MutationTestingConsoleLoggerSubscriber` was written, the responsibility of the reporters was not well defined. But now I think this entire reporting part belongs, well, to reporters.

### Motivations

This was not directly asked but that is the natural next question: why bother moving that code around. The answer is two fold.

First, there are several issues regarding how reporters are currently written/designed. This is more of a general point and I hope sufficiently articulated in #2866, so I will not dwell on it here.

Second, when experimenting/investigating the TeamCity output I've been struggling a lot to control the different outputs and make sense of what is doing what, from where, and how to control it. The issue is that some goes in the loggers, which can be configured, some is directly on the stdout, some in the reporters (which can output to the stdout too).

What I would like to achieve here is clearer boundaries of what does what, which should make it easier to control what can do what and when.

For example: outputting information during the execution? Loggers. Maybe the logger can be interactive, be a progress bar, be silent, or be dedicated to TeamCity or other. There is still question marks for me on how we want to control them, i.e. one setting to configure all loggers or multiple ones, to be discussed when the time comes. 

Outputting information (stdout, files or others) once the whole execution is done? Reporters.

That's it: nothing else should randomly print something.

maks-rafalko:
generally, makes sense, thanks. 

However

> For example: outputting information during the execution? Loggers.
> Outputting information (stdout, files or others) once the whole execution is done? Reporters.

ok, but now what the extracted [reporter `FileLocationReporter`](https://github.com/infection/infection/pull/2937/changes#diff-865d4b19a6d444222ce3653242ef535694914d4164c748ef22e89bc083093e82R67) does is really logs something, it doesn't "output information once the whole execution is done".. or am I missing something? So why is it reporter but not the logger? :)

theofidry:
>but now what the extracted [reporter FileLocationReporter](https://github.com/infection/infection/pull/2937/changes#diff-865d4b19a6d444222ce3653242ef535694914d4164c748ef22e89bc083093e82R67) does is really logs something, it doesn't "output information once the whole execution is done"

It's outputting the file reporters path once the execution is done. So whilst it could be the log of entry of a logger that logs the end of the execution, it also fits the definition of reporter which is outputting something at the end of the execution.

To be clear though, I plan to:

- move this reporter as part of the reporter chain, so there is only one `$reporter->report()` call done ever
- make this reporter a proper decorator, i.e. it would not only log the reported files but also call the decorated reporter.

theofidry:
if you want a peek you can check https://github.com/infection/infection/pull/2927.

Breaking this done arguably means the general direction may be a bit harder to follow, but equally it's hard to ensure that nothing breaks on the way without more incremental changes

PR #2945, opened 2026-02-08
test: Migrate to PHPUnit stubs

theofidry:
Addresses the issues reported by the Rector upgrade from https://github.com/infection/infection/pull/2938.

I could not find the exact change in PHPUnit that triggers this change of behaviour, it is described however in
https://getrector.com/blog/upgrade-to-phpunit-125-in-7-diffs.

It required some additional changes because Rectors goes one setup further [here](https://getrector.com/blog/upgrade-to-phpunit-125-in-7-diffs#content-2-inline-once-used-mocks-property-to-a-variable) and in our case this result in invalid code (since we compare by reference).

### DIFF
--- a/devTools/phpstan-baseline.neon
+++ b/devTools/phpstan-baseline.neon
@@ -1081,7 +1081,7 @@ parameters:
 			path: ../tests/phpunit/Mutator/ProfileListTest.php
 
 		-
-			rawMessage: 'Property Infection\Tests\Mutator\Util\AbstractValueToNullReturnValueTest::$testSubject with generic class Infection\Mutator\Util\AbstractValueToNullReturnValue does not specify its types: TNode'
+			rawMessage: 'Property Infection\Tests\Mutator\Util\AbstractValueToNullReturnValueTest::$testSubjectStub with generic class Infection\Mutator\Util\AbstractValueToNullReturnValue does not specify its types: TNode'
 			identifier: missingType.generics
 			count: 1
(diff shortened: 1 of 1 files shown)


tests/phpunit/Event/Subscriber/MutationTestingConsoleLoggerSubscriberTest.php
```diff
@@ -96,18 +90,18 @@ public function test_it_reacts_on_mutation_testing_started(): void
 
         $dispatcher = new SyncEventDispatcher();
         $dispatcher->addSubscriber(new MutationTestingConsoleLoggerSubscriber(
-            $this->output,
+            $this->createStub(OutputInterface::class),
             $this->logger,
             $this->metricsCalculator,
             $this->resultsCollector,
-            $this->diffColorizer,
+            $this->createStub(DiffColorizer::class),
```
sanmai, line 97: This just calls dirty diffs and merge conflicts if you are working on something you want to ship in turns.

sanmai (approved):
Hopefully it is not going to be a problem later.

tests/phpunit/Event/Subscriber/MutationTestingConsoleLoggerSubscriberTest.php
theofidry: Yes I'm not too fond of it but at the same time not creating unnecessary properties is not too bad and I don't really want to go fix each individual cases atm either

PR #2948, opened 2026-02-09
test(ast): Add a node ID map for visitor tests

theofidry:
## Description

Sometimes when testing our AST we want to check the state of a specific node. For example right now, for resolving the FQCN we have:

- `NameResolverFactory` configured a certain way.
- `FullyQualifiedClassNameManipulator::getFqcn()` which gives the resolved name.

The former has not test and the tests of the latter look like this:

```php
public function test_it_can_provide_the_node_fqcn(): void
{
    $fqcn = new FullyQualified('Acme\Foo');
    $node = new Nop(['resolvedName' => $fqcn]);

    $this->assertSame($fqcn, FullyQualifiedClassNameManipulator::getFqcn($node));
}

public function test_it_can_provide_the_node_fqcn_for_an_anonymous_class(): void
{
    $node = new Nop(['resolvedName' => null]);

    $this->assertNull(FullyQualifiedClassNameManipulator::getFqcn($node));
}
```

Needless to say, this is not testing anything. Instead, it would be more interesting to have something along those lines:

```php
$nodes = $this->parser->parse('...');

$nodesById = $this->addIdsToNodes($nodes);
(new NodeTraverser(
    NameResolverFactory::create(),
))->traverse($nodes);

// This assertion is not for the test itself but to provide a visible representation of
// the nodes traversed to the reader.
$this->assertSame(
    '...',
    $this->dumper->dump($nodes, onlyVisitedNodes: false),
);

$specificNodeIWantToTest = $nodesById[10]; // I know what ID from the dump just above

// Do my tests against $specificNodeIWantToTest
$this->assertSame(
    '...',
    FullyQualifiedClassNameManipulator::get($specificNodeIWantToTest),
);
```

This allows to easily retrieve a specific node without having to mess with the whole node hierarchy, which if you doubt is annoying/verbose I invite you to look at this PR `AddIdToTraversedNodesVisitorTest` new tests.

## Changes

Add a map to `AddIdToTraversedNodesVisitor` to track the ids assigned to nodes during the traverse, allowing to provide a map of nodes by their IDs at the end of the traverse.

This allows to adapt `VisitorTestCase::addIdsToNodes()` to return this map for visitor tests.


maks-rafalko (approved):
having an ID map itself is fine, while maintaining this code

```php
$specificNodeIWantToTest = $nodesById[10]; // I know what ID from the dump just above
```

can be a PITA.

I just imagine the case where 10 is changes to 9 or 11 because of some changes in the nodes, and this will be extremely difficult to understand & fix, no?

theofidry:
>I just imagine the case where 10 is changes to 9 or 11 because of some changes in the nodes, and this will be extremely difficult to understand & fix, no?

But what would be the alternative? Right now there is:

- not doing anything, you end up with the massive boilerplate shown in the test because you need to go through the entire node tree yourself
- have a different ID labelling strategy. But what?

So yeah I agree it can easily be annoying, but at the same time I don't really see a different way.

Or... Actually maybe there is one... I'll experiment with it. But in any case I think this PR offers a better alternative than the status quo

PR #2955, opened 2026-02-11
test(ast): Allow to label specific nodes

theofidry:
List of places spotted that can leverage this:

- `MarkTraversedEligibleNodesAsVisitedVisitor`
- `FullyQualifiedClassNameManipulatorTest`
- `NextConnectingVisitorTest`

<hr/>

## Description

This is a follow up of https://github.com/infection/infection/pull/2948.

The current `VisitorTestCase::addIdsToNodes()` returns nodes indexed by sequential numeric IDs (0, 1, 2, ...). When the parsed code changes, all IDs shift, causing tests to break even if the specific node under test didn't change.

This PR introduces a feature to allow labelling specific nodes via comments in the test code. Labels serve as stable identifiers that don't change when unrelated code is modified.

```php
<?php

function foo() {  // @label:Stmt_Function:foo-function
    $x = 1;  // @label:Expr_Variable:x-var
    return $x;  // @label:Stmt_Return:return-stmt
}
```

There is still value in the sequential IDs as they are automated and can be used when dumping the AST, but those labels can offer a better API for when one wants to select a specific node.

## Reviewer Notes

I had a look at the code, I briefly let Claude do its thing after babysitting its specs. Not happy with it, I'll update it when I get. a chance, but on principle it works.

### DIFF
--- a/tests/phpunit/PhpParser/Visitor/VisitorTestCase.php
+++ b/tests/phpunit/PhpParser/Visitor/VisitorTestCase.php
@@ -36,8 +36,10 @@ declare(strict_types=1);
 namespace Infection\Tests\PhpParser\Visitor;
 
 use Infection\Testing\SingletonContainer;
+use Infection\Tests\TestingUtility\PhpParser\LabelParser\LabelParser;
 use Infection\Tests\TestingUtility\PhpParser\NodeDumper\NodeDumper;
 use Infection\Tests\TestingUtility\PhpParser\Visitor\AddIdToTraversedNodesVisitor\AddIdToTraversedNodesVisitor;
+use Infection\Tests\TestingUtility\PhpParser\Visitor\AssignLabelsToNodesVisitor\AssignLabelsToNodesVisitor;
 use Infection\Tests\TestingUtility\PhpParser\Visitor\KeepOnlyDesiredAttributesVisitor\KeepOnlyDesiredAttributesVisitor;
 use Infection\Tests\TestingUtility\PhpParser\Visitor\MarkTraversedNodesAsVisitedVisitor\MarkTraversedNodesAsVisitedVisitor;
 use PhpParser\Node;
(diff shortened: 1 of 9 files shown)


theofidry:
One example of usage, in https://github.com/infection/infection/pull/2973/changes#diff-bd135fb23d1bbbcbcdfc63058e0e440a2cd07bcb7810320f042abac7033dea8aR160 we refer to the node ID to specify what the name should be. If the test code changes a bit, the IDs get moved around, requiring to update the test, making it a bit fragile. With this, we could label the node we want to target and use that label instead:

Before:

```
yield 'function declaration and call' => [
    <<<'PHP'
        <?php

        function calculate() {}
        calculate();

        PHP,
    <<<'AST'
        array(
            0: Stmt_Function(
                name: Identifier(
                    nodeId: 1
                )
                nodeId: 0
            )
            1: Stmt_Expression(
                expr: Expr_FuncCall(
                    name: Name(
                        nodeId: 4
                        resolvedName: nodeId(4)
                    )
                    nodeId: 3
                )
                nodeId: 2
            )
        )
        AST,
    [
        4 => new FullyQualified('calculate'),
    ],
];
```

After, with a bit more work we can make the output a bit leaner (e.g. we don't need to list all the IDs):

```
yield 'function declaration and call' => [
    <<<'PHP'
        <?php

        function calculate() {}
        calculate();    // @label:Name:function-name

        PHP,
    <<<'AST'
        array(
            0: Stmt_Function(
                name: Identifier(...)
                nodeId: 0
            )
            1: Stmt_Expression(
                expr: Expr_FuncCall(
                    name: Name(
                        resolvedName: ...
                    )
                )
            )
        )
        AST,
    [
        'function-name' => new FullyQualified('calculate'),
    ],
];
```

Another example is in the following case:

```php
public function test_it_can_provide_the_next_node(): void
{
    $nodes = $this->parse(
        <<<'PHP'
            <?php

            $a = 1;
            $b = 2;
            PHP,
    );

    $this->addIdsToNodes($nodes);
    (new NodeTraverser(
        new NextConnectingVisitor(),
    ))->traverse($nodes);

    $expressionA = $nodes[0];
    $this->assertInstanceOf(Expression::class, $expressionA);

    $expressionB = $nodes[1];
    $this->assertInstanceOf(Expression::class, $expressionB);

    $this->assertTrue(NextConnectingVisitor::hasNextNode($expressionA));
    $this->assertFalse(NextConnectingVisitor::hasNextNode($expressionB));
}
```

In this instance the test is perfectly fine as it is, but partly because we constrained ourselves to keep the PHP code extremely simple precisely to avoid having to go through the whole tree code code. With the above feature we can do the following:

```php
public function test_it_can_provide_the_next_node(): void
{
    $nodes = $this->parse(
        <<<'PHP'
            <?php

            $a = 1; // @label:Expression:expressionA
            $b = 2; // @label:Expression:expressionB
            PHP,
    );

    $nodesByIds = $this->addIdsToNodes($nodes);
    (new NodeTraverser(
        new NextConnectingVisitor(),
    ))->traverse($nodes);

    $expressionA = $nodesByIds['expressionA'];
    $expressionB = $nodesByIds['expressionB'];

    $this->assertTrue(NextConnectingVisitor::hasNextNode($expressionA));
    $this->assertFalse(NextConnectingVisitor::hasNextNode($expressionB));
}
```

I do not have much concrete examples right now, mostly because well... it was so annoying, we simply avoided those situations. We only recently introduced the ability to add IDs to nodes and use them, this feature is more about removing one more annoying constraints.

I don't have a strong need for it at the moment though so I'm fine putting this on hold, it is more to know if you're open to the idea so that I can invest time to introduce once I hit a point that it becomes valuable.

PR #2957, opened 2026-02-12
test: Allow to set the mutation hash in the builder

theofidry:
Sometimes you want a mutation with a specific hash without caring about what it takes to have this specific hash.

This will be needed for multiple upcoming PRs. I've also already felt the need for it when implementing the `TeamCityLogger`, but tried to do without at that time. I refrained myself to update the existing test as I'm not sure it will be that big of an improvement and there is still pending PRs like https://github.com/infection/infection/pull/2918 which would then need to have conflicts resolved.

### DIFF
--- a/devTools/phpstan-baseline.neon
+++ b/devTools/phpstan-baseline.neon
@@ -1216,12 +1216,6 @@ parameters:
 			rawMessage: 'Call to internal method PHPUnit\Framework\TestCase::addToAssertionCount() from outside its root namespace PHPUnit.'
 			identifier: method.internal
 			count: 1
-			path: ../tests/phpunit/TestFramework/Coverage/JUnit/JUnitTestFileDataProviderTest.php
-
-		-
-			rawMessage: 'Call to internal method PHPUnit\Framework\TestCase::addToAssertionCount() from outside its root namespace PHPUnit.'
-			identifier: method.internal
-			count: 1
 			path: ../tests/phpunit/TestFramework/Coverage/JUnit/TestLocationBucketSorterTest.php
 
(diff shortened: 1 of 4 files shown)


maks-rafalko:
> Sometimes you want a mutation with a specific hash without caring about what it takes to have this specific hash.

could you please describe in which cases? As now it looks suspicious :) 

> This will be needed for multiple upcoming PRs

are we going to pass the cache NOT built with the algorithm but some random other hash? For what purpose?

theofidry:
>could you please describe in which cases? As now it looks suspicious :)

I'm not sure what's suspicious about it, we do not care of the coherence/integrity of the data at all times. For example if you test a service that displays some information of the mutations, your test probably only cares about those elements, not whether this mutation is realistic. To this effect, having a tool to easily build a mutation in a given state makes it easier, and that is the purpose of that (test) builder.

When I was working on the `TeamCityLogger` for example, we display the hash, being able to set one arbitrary human readable value directly is a lot nicer than guessing what values used behind it are and set all of them (even if you are not interested in some/any of them) because otherwise you would end up with duplicate hashes.

>are we going to pass the cache NOT built with the algorithm but some random other hash? For what purpose?

To set an arbitrary value that is both more convenient for the test and readable.

<hr/>

Turns out the case I needed was not necessary. I'm manipulating `MutantExecutionResult` for which we already have `MutantExecutionResultBuilder::withMutantHash()`. The PR in question is https://github.com/infection/infection/pull/2959 where previously a mock was used and where the builder was nicer to use (IMO).

So until necessary we can hold off on the PR.

PR #2964, opened 2026-02-13
perf(ast): Remove the node memoization from NodeMutationGenerator

theofidry:
I have delayed quite a bit a few changes I wanted to do on the AST due to the potential issues caused by this change, hence I extracted it out from https://github.com/infection/infection/pull/2926.

## Description

We introduced memoization of some properties in `NodeMutationGeneration` in #1423.

The PR claims some performance gains, but I could not find similar results when using either our benchmarks or blackfire profiles.

## Motivations

As per https://github.com/infection/infection/issues/2797, more state will be added to the nodes, and more state should be removed from `NodeMutationGenerator`. 

I do not think this sort of memoization can be replicated everywhere we would use state from a node attribute.


sanmai (approved):
I suppose we moved the memoization layer to somewhere else since 2020, so this optimization became redundant.

PR #2966, opened 2026-02-16
test(event): Add more tests for the MutationTestingConsoleSubscriber show metrics

theofidry:
## Description

This PR adds tests to cover more extensively `MutationTestingConsoleSubscriber::showMetrics()`. Like done in https://github.com/infection/infection/pull/2959, it will be extracted as a reporter as the next step.

## Related issues

- https://github.com/infection/infection/pull/2927

### DIFF
--- a/tests/phpunit/Event/Subscriber/MetricsScenario.php
+++ b/tests/phpunit/Event/Subscriber/MetricsScenario.php
@@ -1,0 +1,139 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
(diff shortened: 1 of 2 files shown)


theofidry:
The failures are unrelated to this change, they are caused by https://github.com/webmozarts/assert/issues/347. It affects our pipelines for which we do a composer update which picked up this new version.

I'd opt to wait a bit see if that gets solved, if it doesn't, we can add a conflict for this version. I would prefer to avoid having to do that, as it also impacts our users, and the change affects edge code that the regular user should not encounter so blocking them from updating in this case is a bit extreme.

This PR is a boring bit that follows the same logic as https://github.com/infection/infection/pull/2959 so I won't bother waiting for a review.

Issue #2958, opened 2026-02-12
bug: `global-ignoreSourceCodeByRegex` is not ignoring multiline code since 0.32.0


LeoVie:
| Question    | Answer
| ------------| ---------------
| Infection version | >= 0.32.0 (`infection.phar --version`)
| Test Framework version | PHPUnit
| CI | no
| PHP version | 8.4.17
| Platform    | MacOS
| Github Repo | https://github.com/LeoVie/infection-issue

Since infection 0.32.0, I am getting escaped mutators for code, that should be ignored via `global-ignoreSourceCodeByRegex`.
Until 0.31.9, no mutators were escaped.

<!-- Please past your phpunit.xml[.dist] if no Github link to the repo provided -->

<details>
<summary>phpunit.xml</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>

<!-- https://docs.phpunit.de/en/latest/configuration.html -->
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
         failOnDeprecation="true"
         failOnNotice="true"
         failOnWarning="true"
         colors="true"
         bootstrap="tests/bootstrap.php"
>
    <php>
        <ini name="display_errors" value="1"/>
    </php>

    <testsuites>
        <testsuite name="All">
            <directory>./tests/</directory>
        </testsuite>
    </testsuites>


    <source ignoreSuppressionOfDeprecations="true"
    >
        <include>
            <directory>src</directory>
        </include>
    </source>

    <coverage>
        <report>
            <html outputDirectory="build/coverage/html" />
            <clover outputFile="build/coverage/clover.xml" />
        </report>
    </coverage>
</phpunit>
```
</details>

<details>
<summary>infection.json5</summary>

```json5
{
  "$schema": "vendor/infection/infection/resources/schema.json",
  "timeout": 10,
  "source": {
    "directories": [
      "src"
    ],
  },
  "mutators": {
    "@default": true,
    "global-ignoreSourceCodeByRegex": [
      "\\$this->logger->info.*"
    ]
  }
}
```
</details>

<!-- Remove this section if not needed -->
<details>
 <summary>Output with issue (0.32.0)</summary>

 ```
    ____      ____          __  _
   /  _/___  / __/__  _____/ /_(_)___  ____
   / // __ \/ /_/ _ \/ ___/ __/ / __ \/ __ \
[...]

PR #2971, opened 2026-02-16
Make sure one-line logger call is ignored with different mutators on this line

maks-rafalko:
Related to 

- https://github.com/infection/infection/discussions/2970
- https://github.com/infection/infection/issues/2958

This test proves one-line call is correctly ignored.

### DIFF
--- a/tests/e2e/Without_Git/composer.json
+++ b/tests/e2e/Without_Git/composer.json
@@ -14,6 +14,7 @@
         }
     },
     "minimum-stability": "dev",
+    "prefer-stable": true,
     "config": {
         "allow-plugins": {
             "infection/extension-installer": true


maks-rafalko:
damn, have no idea why `Without_Git` test fails on this PR but doesn't fail on others :|

sanmai:
It fails for me locally all the time FWIW

maks-rafalko:
> It fails for me locally all the time FWIW

you mean with exactly this error? or just this test with any error?

cause I can't reproduce it locally with PHP 8.4

maks-rafalko:
ok, thanks to Claude it's fixed

>  Root cause: The e2e test's composer.json has "minimum-stability": "dev" (needed to install infection/infection from a branch) but no "prefer-stable": true. This causes Composer to resolve symfony/console to 8.1.x-dev instead of stable 7.4.x.

this test is quite expensive, not the first time it took hours to debug. But at the same time it's very useful, so I'm not sure how to simplify it..

PR #2976, opened 2026-02-19
Fix logical or mutator failing on sigil comparison

### LINKED ISSUE #2975: bug: Weirness in LogicalOr
| Question    | Answer
| Infection version | 0.32.0
| Test Framework version | PHPUnit 10.5.10
| CI | playground
| PHP version | -
| Platform    | -
| Github Repo | -

$leftSigil and $rightSigil are both "<", so there are no matches.

Managed to reproduce in the playground:
- https://infection-php.dev/r/x2mo

BackEndTea:
## Description

Bug #2975 ran into an issue where the following statement was mutated `$a < 0 || 23 < $a`. (Which is an equivalent of `$a < 0 || $a > 23`). This is a valid check, as it is a negative range check, meaning it checks if $a is outside of the range 0-23.

Mutating this to an `&&` does not make sense, as it would mean checking if $a is both smaller than 0 and larger than 23.

## Changes

Adds a check in logical or, to not mutate when the direction of the sigils are equivalent. e.g. when comparing `>=` and `>`.

Also updated the thrown exception to include the sigils, in case another bug ever occurs. This way there is more information about exactly what went wrong.

## Checklist

- [X] Tests added/updated
- [x] Appropriate labels applied (bugfix)

## Reviewer Notes

I'm not sure if the dataprovider name is the correct way to do things, or if it should be grouped differently

## Related issues

Fixes #2975 

This PR:
- [X] Covered by tests


BackEndTea:
The escaped mutant is about the Logic exception in the match arm, and is intended to never be triggered.

tests/phpunit/Mutator/Boolean/LogicalOrTest.php
```diff
@@ -731,4 +733,132 @@ private static function instanceOfMutationsProvider(): iterable
...
+                <<<'PHP'
+                    0 > $a || $a >= 23;
+                    PHP,
+            ),
+        ];
+
+        yield 'It does mutate different variables are tested against "Smaller" and "Smaller' => [
+            self::wrapCodeInMethod(
+                <<<'PHP'
+                    $a < 0 || 23 < $b;
```
maks-rafalko, line 853: it looks like this still makes sense to mutate, no?

```php
$a < 0 && 23 < $b
```

is a valid and good mutation IMO

tests/phpunit/Mutator/Boolean/LogicalOrTest.php
```diff
@@ -731,4 +733,132 @@ private static function instanceOfMutationsProvider(): iterable
...
+                            throw new \Exception('Wrong');
+                        }
+                        PHP,
+                ),
+                self::wrapCodeInMethod(
+                    <<<'PHP'
+                         if (
+                            $a < 0 ||
+                            23 < $a ||
+                            $b < 0 && 59 < $b
```
maks-rafalko, line 769: could you please explain why we still mutate to this `$b < 0 && 59 < $b`? It's not real

tests/phpunit/Mutator/Boolean/LogicalOrTest.php
BackEndTea: That is mutated, on line 857 is the mutated code

maks-rafalko: oh, misread. Thanks!

tests/phpunit/Mutator/Boolean/LogicalOrTest.php
BackEndTea: Due to operator precedence these two statements are the same. So mutating the final || to && makes sense here

```
$a || $b || $c && $d

($a || $b || $c) && $d

```

maks-rafalko: nice, thanks for the explanation. 

Optionally, I would add it to the comment as it's not so obvious about the precedence.

maks-rafalko:
Thank you @BackEndTea

PR #2986, opened 2026-02-24
refactor(event): Introduce InitialStaticAnalysisExecutionLogger

theofidry:
## Description

This PR is the counter part of the `InitialTestsExecutionLogger`, see https://github.com/infection/infection/pull/2978.

Note that there is **a lot** of similitude between the two. But further simplifications are avoided here to reduce the noise and will be done in a separate PR for better visibility.

## Changes

- Introduce `InitialStaticAnalysisExecutionLogger`
- `InitialStaticAnalysisRunConsoleLoggerSubscriber` -> `InitialStaticAnalysisExecutionLogger`
- `CiInitialStaticAnalysisRunConsoleLoggerSubscriber` -> `ConsoleNoProgressLogger` 
- `InitialStaticAnalysisRunConsoleLoggerSubscriber` -> `ConsoleProgressBarLogger`

Note that I opted for "execution" instead of "run" as both are valid and "execution" was already picked for the tests, then it made sense to keep the same terminology.


theofidry:
It is exactly the same as https://github.com/infection/infection/pull/2978, hence won't wait for a review. The subsequent one will require a review though.

Issue #2907, opened 2026-01-29
bug: Polluted TeamCity logs


theofidry:
| Question    | Answer
| ------------| ---------------
| Infection version | master (unreleased)
| Test Framework version | any
| CI | yes/no (local only)/either
| PHP version | any
| Platform    | any
| Github Repo | -

When testing the infection plugin, I noticed some tests that were marked as "stopped". I think this is caused by some logs being printed before all the teamcity logs are out.

I need to investigate further and figure out a solution.

xepozz:
Are there examples?

theofidry:
Not yet, I didn't really get a good reproducer yet.

But I'm pretty certain it is an issue on our side. We'll be able to tell with more accuracy once I managed to get a reliable reproducer.

theofidry:
I could confirm the issue. This is happening sporadically but consistently with escaped mutations. The issue is that despite picking the TeamCity logger, we still log stuff to the stdout, which we should not.

theofidry:
@xepozz I am a bit lost there, and actually I think the issue is on the plugin side, I am a bit confused though so I'll have to work a bit more to provide a reliable reproducer to allow you to investigate.

What I did in https://github.com/infection/infection/pull/2931 was

- decorating the output we use to not only output in the stdout but also append any output in a file `var/teamcity/debug.log`
- this file is emptied whenever invoking `bin/infection`, otherwise anything is in append only mode so that if there is multiple processes logging things to that file there is no issues.
- adjust the detection status to provide more escaped results, allowing to reduce the scope (with `--filter=src/Command` for example) to not have to run the whole test suite
- generate the coverage data: XDEBUG_MODE=coverage vendor/bin/phpunit --coverage-xml=var/coverage/coverage-xml --log-junit=var/coverage/junit.xml
- configure the command to use it and skip the initial tests for a faster feedback loop: `-n --no-progress --teamcity --coverage=var/coverage --skip-initial-tests --filter=src/Command --threads=max`

Executing this result in this:

<img width="631" height="768" alt="Image" src="https://github.com/user-attachments/assets/bc5d5fa4-5a6b-421c-bdbb-2e0e47bb2c78" />

You can notice for example the test for the mutation `89d9113405e0263c6bc001482b50030a` was marked as "stopped", which looks incorrect.

From this execution, I recovered the following files:

- [the output printed by the plugin](https://github.com/user-attachments/files/25075529/row-output.log)
[...]

Issue #2863, opened 2026-01-18
RFC: TestFramework Interface Unification


theofidry:
# Summary

The test framework adapter is one of the oldest components in the codebase. It existed before
the fork of Infection from Humbug and has evolved significantly since then.

Initially, it was a simple interface:

```php
interface TestFrameworkAdapter {
    public function getExecutableCommandLine() : string;    // e.g. 'vendor/phpunit/phpunit/phpunit', 'bin/phpspec'.
}
```

This API quickly evolved to accommodate additional use cases. Eventually, it was extracted to a separate package
(#922) to allow adapters to live outside the codebase. However, since #519
in late 2018 – nearly 8 years at the time of writing!, the API has remained unchanged.

Although the API remained stable, several long-standing issues and feature requests were never addressed. Other changes
were implemented by working around the API rather than evolving it.

The goal of this RFC is to lay out the current issues and propose solutions to address them.

# Motivations

This RFC is motivated by two goals:

- Support for multiple test frameworks. Users should be able to run Infection with more than just PHPUnit, and, for
  example, use it with PHPUnit + Behat. Likewise, I would like to be able to have support for more than one static
  analysis tool at the same time.
- Easier and richer test framework adapters. There is a variety of concerns when integrating a new
  test framework: how it works, if it needs to be launched in separate processes or if it can be done
  in-memory, what kind of coverage it handles, what kind of cache or other artefacts it needs, what
  are its testing capabilities: is it a yes/no, or we have more granular results and can infer more
  information from it (e.g. warnings/notices). Adding new test framework adapters is non-trivial, and
  the current API does not accommodate all execution models.

_The exact details as to what is a problem and for what purpose will be detailed in the next section._

# Issues

## Multi-framework support

Initially requested in https://github.com/infection/infection/issues/157 in 2018, this has been a long-standing feature request (and a feature I've been wanting
for longer than that to be honest).

The idea is very simple, we want to be able to use more than one test framework at once. While the concept is
straightforward, the implementation presents several challenges. It was actually demonstrated by the recent integration of the [PHPStan adapter][phpstan-introduction].
The PHPStan integration highlighted several limitations:

- It could not be a regular test framework adapter, because then it would not be usable with PHPUnit or another test
  framework adapter. Precisely because infection currently supports only one test framework.
- The current API of `TestFrameworkAdapter` is very similar, yet has some differences with the introduced `StaticAnalysisToolAdapter`.
[...]

PR #2987, opened 2026-02-24
refactor(logger): Reduce duplicated code between the initial tests & static analysis execution loggers

theofidry:
## Description

This PR is a follow up of #2978, #2986 and https://github.com/infection/infection/pull/2991. Those recent refactoring made more obvious the similitude between the two and this PR proposes to adapt the code to avoid such duplication.

Whilst there is more duplication remaining, this is due to the similitude between the `StaticAnalysisAdapter` and `TestFrameworkAdapter` which is out of scope and covered by https://github.com/infection/infection/issues/2863.

## Changes

- Merge `InitialStaticAnalysisExecution\ConsoleNoProgressLogger` and `InitialTestsExecution\ConsoleNoProgressLogger` into `ConsoleNoProgressLogger`.
- Merge `InitialStaticAnalysisExecution\ConsoleProgressBarLogger` and `InitialTestsExecution\ConsoleProgressBarLogger` into `ConsoleProgressBarLogger`.

## Related issues

- https://github.com/infection/infection/issues/2907

### DIFF
--- a/tests/phpunit/Logger/ArtefactCollection/ConsoleNoProgressLoggerTest.php
+++ b/tests/phpunit/Logger/ArtefactCollection/ConsoleNoProgressLoggerTest.php
@@ -78,7 +78,7 @@ final class ConsoleNoProgressLoggerTest extends TestCase
 
         $expected = <<<'EOF'
 
-            Initial execution of PHPUnit version: 6.5.4
+            Running initial tests with PHPUnit version 6.5.4
 
             EOF;
 
--- a/src/Logger/ArtefactCollection/ConsoleNoProgressLogger.php
+++ b/src/Logger/ArtefactCollection/ConsoleNoProgressLogger.php
@@ -65,7 +65,7 @@ final readonly class ConsoleNoProgressLogger implements InitialStaticAnalysisExe
         $this->output->writeln([
             '',
             sprintf(
-                'Initial execution of %s version: %s',
+                'Running initial tests with %s version %s',
                 $this->testFramework->getName(),
                 $version,
             ),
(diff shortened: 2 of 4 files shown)


tests/phpunit/Logger/ArtefactCollection/ConsoleNoProgressLoggerTest.php
```diff
@@ -78,9 +78,7 @@ public function test_it_logs_on_start(): void
 
         $expected = <<<'EOF'
 
-            Running initial test suite...
```
maks-rafalko: not sure I like this change

- "initial tests" phase is a well-known wording since day 0
- "initial execution" is a bit ambiguous / not clear

but I don't have a better suggestion atm.. If you don't have as well, let's go

theofidry: I don't feel strongly about it, but I did not like changing this in a "refactor" since IMO it should not change behaviour...

I don't mind to have "Running" instead of "execution". What do you think about `Running initial tests with PHPUnit/PHPStan version X`? Maybe @sanmai has an idea too

maks-rafalko: > What do you think about `Running initial tests with PHPUnit/PHPStan version X`

for me it's a better option (that I'm personally get used to). But yeah, I also don't feel strongly about it

theofidry: Extracted it to https://github.com/infection/infection/pull/2991 for better visibility. I'll make this PR into a pure (and simple) refactoring

PR #2989, opened 2026-02-24
refactor(event): Rename the reporter subscriber

theofidry:
`Infection\Logger\MutationTestingResultsLogger` has been renamed into `Infection\Reporter\Reporter` in
https://github.com/infection/infection/pull/2932, hence the "Logger" part does not make much sense.

Unlike `MutationAnalysisLoggerSubscriber` for example, which operates for the whole `MutationAnalysis` phase, this subscriber reacts to only one event. As a result, we could follow the existing pattern, like `CleanUpAfterMutationTestingFinishedSubscriber`, to name it after the event subscribed to.

This PR proposes to rename it to `ReportAfterMutationTestingFinishedSubscriber`.

### DIFF
--- a/src/Container/Container.php
+++ b/src/Container/Container.php
@@ -63,7 +63,7 @@ use Infection\Event\Subscriber\DispatchPcntlSignalSubscriber;
 use Infection\Event\Subscriber\InitialStaticAnalysisExecutionLoggerSubscriber;
 use Infection\Event\Subscriber\InitialTestsExecutionLoggerSubscriber;
 use Infection\Event\Subscriber\MutationAnalysisLoggerSubscriber;
-use Infection\Event\Subscriber\MutationGeneratingConsoleLoggerSubscriberFactory;
+use Infection\Event\Subscriber\MutationGenerationLoggerSubscriber;
 use Infection\Event\Subscriber\MutationTestingResultsCollectorSubscriber;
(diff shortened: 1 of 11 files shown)


maks-rafalko (commented):
I feel like it's a downgrade in the clarity, isn't it?

When I see `MutationTestingResultsLoggerSubscriber` - it's clear that at which point this subscriber is triggered.

When I see `ReporterSubscriber`, I have no idea where/when it's going to report.

But I'm not sure, probably I miss something or doesn't see the whole picture

theofidry:
>it's clear that at which point this subscriber is triggered.

I admit I did not consider this, actually it appears it's the pattern for single event subscribers, e.g. `CleanUpAfterMutationTestingFinishedSubscriber`. I didn't really notice it because the others were more about a phase than a single event.

`MutationTestingResultsLoggerSubscriber` is not correct IMO because, well, it doesn't log anything... And I didn't pick up on the when because next to it we have `MutationTestingResultsCollectorSubscriber` which has the same prefix, but happens at a different time.

So taking your suggestion in consideration, what do you think of `ReportAfterMutationTestingFinishedSubscriber`?

maks-rafalko:
> So taking your suggestion in consideration, what do you think of `ReportAfterMutationTestingFinishedSubscriber`?

nice one, thanks

maks-rafalko (approved):

Issue #2990, opened 2026-02-24
bug: BaseMutatorTestCase fails when custom mutator  `getName()`  is not the class name


to-mo-ki:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.32.5
| Test Framework version | PHPUnit 11.5.51
| CI | local only
| PHP version | 8.2.30 (`PHP 8.2.30 (cli)`)
| Platform    | macOS (Docker)
| Github Repo | - |

When creating a custom mutator, `BaseMutatorTestCase` appears to fail if the mutator `getName()` is not the class name.

I confirmed this in Docker using the current `master` branch.  
If `getName()` returns a custom string (e.g. `'my-custom-name'`), `BaseMutatorTestCase::createMutator()` fails with an undefined array key.  
If `getName()` returns `self::class`, the same setup works.

## Steps to reproduce

1. Create a custom mutator that implements `Mutator` and returns a non-class name from `getName()`:
   - `public function getName(): string { return 'my-custom-name'; }`
2. Create a test extending `BaseMutatorTestCase`:
   - Override `getTestedMutatorClassName()` to return the custom mutator class.
3. Trigger mutator creation (e.g. through `setUp()` or calling `createMutator()` from a public wrapper).

## Actual result

Failure:

```text
Warning: Undefined array key "DemoCustomMutator" in /opt/infection/src/Testing/BaseMutatorTestCase.php on line 162
TypeError
Infection\Testing\BaseMutatorTestCase::createMutator(): Return value must be of type Infection\Mutator\Mutator, null returned
```
## Expected result
BaseMutatorTestCase should work for custom mutators regardless of getName() value, or the constraint should be explicitly documented/enforced with a clearer error message.

## Why this happens (analysis)
`MutatorFactory::create()` returns an array where keys are the names of the mutators (the return value of `$mutator->getName()`). 
However, `BaseMutatorTestCase::createMutator()` tries to access the mutator using the fully qualified class name as a key:
```php
// src/Testing/BaseMutatorTestCase.php
return SingletonContainer::getContainer()
    ->getMutatorFactory()
    ->create([
        $mutatorClassName => ['settings' => $settings],
    ], false)[$mutatorName] // <--- $mutatorName is $mutatorClassName here for custom mutators
;
```

## Additional Context
I have a local fix for this and I'm happy to open a Pull Request if the maintainers agree with this issue.

theofidry:
[...]

PR #2992, opened 2026-02-25
fix(testing): Ensure BaseMutatorTestCase handles custom mutator names

to-mo-ki:
## Description
This PR fixes a bug in `BaseMutatorTestCase::createMutator()` where it was incorrectly using the mutator's FQCN as a lookup key for the MutatorFactory output, while MutatorFactory results are keyed by the mutator name (returned by `getName()`).

## Changes
- Fixes `BaseMutatorTestCase::createMutator()` to retrieve the first (and only) element from the `MutatorFactory::create()` result.
- Since `createMutator()` always requests exactly one mutator class, the resulting array will contain exactly one element, making the lookup by name unnecessary and error-prone for custom mutators.
- Adds tests/phpunit/Fixtures/Mutator/CustomNameMutator.php as a test fixture.
- Adds tests/phpunit/Testing/MutatorTestCaseIntegrationTest.php to ensure BaseMutatorTestCase correctly handles mutators with custom names.

## Checklist
- [x] Tests added/updated
- [ ] Documentation updated - N/A
- [ ] CHANGELOG.md updated - N/A
- [ ] Appropriate labels applied

## Related issues

#2990

### DIFF
--- a/src/Testing/BaseMutatorTestCase.php
+++ b/src/Testing/BaseMutatorTestCase.php
@@ -35,8 +35,6 @@ declare(strict_types=1);
 
 namespace Infection\Testing;
 
-use function array_flip;
-use function array_key_exists;
 use function array_map;
 use function array_shift;
 use function count;
@@ -47,7 +45,6 @@ use Infection\Framework\Str;
 use Infection\Mutation\Mutation;
 use Infection\Mutator\Mutator;
 use Infection\Mutator\NodeMutationGenerator;
-use Infection\Mutator\ProfileList;
 use Infection\PhpParser\NodeTraverserFactory;
 use Infection\PhpParser\Visitor\MutationCollectorVisitor;
 use Infection\PhpParser\Visitor\MutatorVisitor;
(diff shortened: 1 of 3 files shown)


maks-rafalko (approved):
Thanks!

Can you confirm that this branch works for you custom mutators in your project?

Happy to merge and release

----

By the way, what are your custom mutators? Curious to know why would you need them to understand the use case :)

to-mo-ki:
@maks-rafalko Thanks for the quick review and reaching out!

I've confirmed that this branch works perfectly with my project's custom mutators. By using `getName()`, I can now provide clear identifiers for them, which is essential for our workflow.

**The Use Case**:
I am developing several **framework-specific mutators** to verify if certain framework method calls are **actually meaningful** in our logic. 

A prime example is checking the significance of arguments, such as a grouping key. If replacing a specific key with an empty string doesn't break any tests, it reveals that the grouping might be redundant or lacks proper test coverage.

We don't apply these to everything; instead, we target specific patterns or common mistakes we've encountered frequently. Since these mutators are highly framework-dependent, they wouldn't be suitable for the Infection core. This is exactly why having robust support for custom mutators is so valuable to us.

**Why custom names?**:
We prefer assigning descriptive custom names (like `discard-collection-merge` or `empty-grouping-key`) because they make the **Infection reports** much more actionable.

Instead of displaying long class names, these human-readable names allow our team to immediately understand the **intent** of the mutation. It helps us quickly distinguish whether a survive mutant is a sign of redundant code or a missing test.

I'm happy with this progress. Thanks again for maintaining this great tool!

maks-rafalko:
Wow, thank you for such a detailed answer, it's really cool use-case, totally makes sense to me.

Let's get it merged, I will release ASAP

maks-rafalko:
Released https://github.com/infection/infection/releases/tag/0.32.6

to-mo-ki:
Thank you for the super fast release! 
Happy to see it out in v0.32.6. I really appreciate your support! 🎉

PR #3004, opened 2026-03-06
refactor(command): Extract the debug option into a re-usable option

theofidry:
This PR extracts the run command debug option into a re-usable option object. This is a preparatory step for https://github.com/infection/infection/pull/2762 where we will use this option in another command.

There is no change of behaviour whatsoever.


theofidry:
Skipping the review as there is nothing interesting here and the PR requiring it was already approved.

PR #3007, opened 2026-03-07
test(ast): Introduce LabelMutationCandidatesVisitor for more realistic AST testing

theofidry:
## Description

As part of the ongoing effort to enrich the AST with additional behaviour and state (see https://github.com/infection/infection/issues/2797), introducing even simple changes remains difficult without risking regressions.

Significant work has already been done to improve matters, most notably https://github.com/infection/infection/pull/2780. However, this has proven insufficient. A key part of the problem is that `NodeMutationGenerator` still encapsulates a great deal of behaviour, making it impractical to reuse directly in tests. Consequently, replicating the real traversal in tests is considerably more involved than it should be.

For example, `EnrichmentTraverseIntegrationTest` currently correctly captures which nodes are visited by `NodeMutationGenerator`, and via the `eligible` attribute, we can also see which ones may be mutated. However, `NodeMutationGenerator` still has additional behaviour, **the eligibility alone is insufficient**.

To bridge this gap, this PR introduces a new visitor, `LabelMutationCandidatesVisitor`. The concept is similar to the `SimpleCollectionVisitor` removed in https://github.com/infection/infection/pull/2861, but designed to be more amenable to AST testing. This visitor marks all visited nodes as such and duplicates the logic of `NodeMutationGenerator`[^1] to mark all nodes that would be passed to the mutators as mutation candidates

With this in place, `EnrichmentTraverseIntegrationTest` now clearly shows:

- which nodes `NodeMutationGenerator` would visit
- which nodes would be checked by the mutators (excluding additional filtering such as skipping unchanged lines or non-covered nodes)

## Changes

- Introduce `LabelMutationCandidatesVisitor`.
- Add more tests to `EnrichmentTraverseIntegrationTest`.
- Adapt `NodeDumper` and the `debug:ast-dumper` command to highlight mutation candidates.

## Related issues

[^1]: Except the changed lines of code via the source matcher and the tests check.

### DIFF
--- a/src/Container/Container.php
+++ b/src/Container/Container.php
@@ -108,6 +108,7 @@ use Infection\Mutation\MutationGenerator;
 use Infection\Mutator\MutatorFactory;
 use Infection\Mutator\MutatorResolver;
 use Infection\PhpParser\FileParser;
+use Infection\PhpParser\InfectionPrettyPrinter;
 use Infection\PhpParser\NodeDumper\NodeDumper;
 use Infection\PhpParser\NodeTraverserFactory;
 use Infection\Process\Factory\InitialStaticAnalysisProcessFactory;
@@ -170,7 +171,6 @@ use OndraM\CiDetector\CiDetector;
 use function php_ini_loaded_file;
 use PhpParser\Parser;
 use PhpParser\ParserFactory;
-use PhpParser\PrettyPrinter\Standard;
 use PhpParser\PrettyPrinterAbstract;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
@@ -332,7 +332,7 @@ final class Container extends DIContainer
                 new JUnitTestFileDataProvider($container->getJUnitReportLocator()),
             ),
             Parser::class => static fn (): Parser => (new ParserFactory())->createForHostVersion(),
-            PrettyPrinterAbstract::class => static fn (): Standard => new Standard(),
+            PrettyPrinterAbstract::class => static fn (): InfectionPrettyPrinter => new InfectionPrettyPrinter(),
             MetricsCalculator::class => static fn (self $container): MetricsCalculator => new MetricsCalculator(
                 $container->getConfiguration()->msiPrecision,
                 $container->getConfiguration()->timeoutsAsEscaped,
--- a/src/PhpParser/InfectionPrettyPrinter.php
+++ b/src/PhpParser/InfectionPrettyPrinter.php
@@ -33,48 +33,27 @@
 
 declare(strict_types=1);
 
-namespace Infection\Testing;
+namespace Infection\PhpParser;
 
-use Infection\Container\Container;
-use Infection\Mutant\MutantCodePrinter;
-use Infection\PhpParser\InfectionPrettyPrinter;
-use Infection\Tests\AutoReview\PhpDoc\PHPDocParser;
-use PhpParser\NodeDumper;
+use PhpParser\PrettyPrinter\Standard;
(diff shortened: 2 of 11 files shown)


tests/phpunit/PhpParser/Visitor/EnrichmentTraverse/Fixtures/AbstractMethod.php
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Infection\Tests\PhpParser\Visitor\EnrichmentTraverse\Fixtures;
+
+trait TraitExample
+{
+}
+
+abstract class ClassWithAbstractMethod
+{
+    // use TraitExample;
```
maks-rafalko: do we need this?

same is below

tests/phpunit/PhpParser/Visitor/EnrichmentTraverse/Fixtures/Function_.php
```diff
@@ -0,0 +1,46 @@
...
+namespace Infection\Tests\PhpParser\Visitor\EnrichmentTraverse\Fixtures;
+
+function equals(mixed $left, mixed $right): bool
+{
+    return $left === $right;
+}
+
+function identical(mixed $left, mixed $right): bool
+{
+    return $left === $right;
```
maks-rafalko, line 40: not sure where this is used yet, but is it an issue that function names do not reflect it's logic? see `equals()`

tests/phpunit/PhpParser/Visitor/EnrichmentTraverse/Fixtures/AbstractMethod.php
theofidry: not sure, I can remove it here is no reason to keep it right now. I'll also remove the commented tests and unused fixtures while I'm at it

theofidry:
>could you please explain in more details what problem does it solve?

The problem is that right now you look at the AST, you have an idea of the state yes, and you see which nodes are visited by the mutation visitor (which means you can see if a node/branch is skipped), but you have no idea which nodes the mutators will see.

When I did my PoC, I migrated all the behaviour of the rich AST, so in there the eligibility of a node was sufficient to know what could be mutated or not.

Currently however, there is more logic in `NodeMutationGenerator`, so the eligibility is not sufficient yet. And part of my trouble is, as I introduce the migration to move out the logic from `NodeMutationGenerator`, I'm afraid to break things; and I noticed on more than one occasion I broke something despite not being having the `EnrichmentTraverseIntegrationTest` – meaning in its current form it is insufficient to capture the current behavour of our traverses.

So `LabelMutationCandidatesVisitor` comes in. Like previously, the visitor we pass to `createMutationTraverser()` marks the node `NodeMutationGenerator` would pass to the mutator as mutation candidates.

...In theory, and this is where I realise I fucked up `LabelMutationCandidatesVisitor` and took it from my more advanced branch and forgot to add some logic that `NodeMutationGenerator` still has 🤦

tests/phpunit/PhpParser/Visitor/EnrichmentTraverse/Fixtures/Function_.php
theofidry: I thought it was just me being distracted, but no it's the CS fixer

theofidry: well the equality does not really matter tbh, so I'll just remove the equal function for now

theofidry:
@maks-rafalko updated, and updated the description too.

PR #3008, opened 2026-03-08
feat(command): Add a debug command for the AST

theofidry:
## Description

This PR adds a debug command `debug:dump-ast` that allows to dump the AST of a given file to the stdout. The AST dumped is the same one as used for testing in for example `EnrichmentTraverseIntegrationTest`.

## Motivations

In the context of #2797, I have been struggling with ensuring nothing breaks upon any refactoring or addition of optimisations. This is a part that is quite fragile, that we most likely broke without noticing in the past (I do not remember in which PR, but I noticed at some point that has some dead path using `NodeVisitor::DONT_TRAVERSE_CURRENT_AND_CHILDREN` which should have worked at least when they were introduced).

As part of a solution to this problem, we introduced a test now called `EnrichmentTraverseIntegrationTest` in https://github.com/infection/infection/pull/2780.

However, updating this test is a bit cumbersome: you come up with a file, write an empty expectation, execute the test to get the diff, copy the diff and double check that it somewhat makes sense.

With this command, you instead can directly get the dumped AST like so:

```shell
bin/infection debug:dump-ast <path-to-file> | pbcopy
```

Another benefit IMO, is that if a user encounters suspicious behaviour, we can always ask them to dump the AST of the problematic file.

Another scenario would be the playground, for debugging purposes.

## Changes

- Introduce the `DumpAstCommand` command.
- Move the `NodeDumper` and some related utilities to the source code and mark them as `@internal` for them to be available in the distributed package/PHAR (otherwise the command would not be usable for a user).

## Related issues

- https://github.com/infection/infection/pull/2797
- https://github.com/infection/infection/pull/3007

### DIFF
--- a/src/Command/Debug/DumpAstCommand.php
+++ b/src/Command/Debug/DumpAstCommand.php
@@ -103,7 +103,7 @@ final class DumpAstCommand extends BaseCommand
 
         $nodes = $this->createAst($container, $file);
 
-        $io->writeln(
+        $io->write(
             $container->getNodeDumper()->dump($nodes),
         );
 
--- a/tests/phpunit/Command/Debug/DumpAstCommand/DumpAstCommandTest.php
+++ b/tests/phpunit/Command/Debug/DumpAstCommand/DumpAstCommandTest.php
@@ -121,7 +121,6 @@ final class DumpAstCommandTest extends FileSystemTestCase
                     origNode: nodeId(4)
                 )
             )
-
             AST;
 
         $tester->execute([


src/Command/Debug/DumpAstCommand.php
```diff
@@ -0,0 +1,172 @@
...
+            ->getContainer()
+            ->withValues(
+                logger: $logger,
+                output: $io->getOutput(),
+                configFile: $configFile,
+            );
+
+        $nodes = $this->createAst($container, $file);
+
+        $io->writeln(
```
theofidry:
```suggestion
        $io->write(
```

theofidry:
I would like to move a bit forward with this, since this is not critical and can be easily reverted, we can revert it later if you are not happy with it.

PR #3009, opened 2026-03-10
feat!: Improve the project directory resolution to the ConfigurationFactory

theofidry:
## Description

Currently the project directory resolution is partially done (and copy/pasted) in the `GitHubAnnotationsReporter` and `GitLabCodeQualityReporter` with an ugly in-line and some unsafe `shell_exec()`.

The project directory already comes from the `Configuration` as it can be forwarded by the `--logger-project-root-directory` option.

This PR proposes to move the resolution of the project directory to the configuration.

## Changes

- Add validation to what the project directory value should be in the `RunCommand` (**BC break**: can fail when it previously did not).
- Extract the CI environment variable check and add validation (**BC break**: can fail when it previously did not).
- Move the project directory resolution via git to the `Git` service.
- Move the resolution of the project directory from the respective reporters to `ConfigurationFactory` (BC break: now failure means infection cannot be executed whilst before it would happen when doing the reporting).
- Rename "logger project root directory" to "project directory": it is what it actually is, it is not logger specific. The CI variable is also named `CI_PROJECT_DIR` which matches. The option was left alone however to not cause a hard BC to the users.

## Breaking Changes

- Infection will now fail if the project directory value used is invalid, i.e. the path found is not an absolute path pointing at a readable directory.
- Infection will now fail when computing the configuration (i.e. early, before the initial tests) instead of at the reporting (which is not transactional, so some reports could have been executed).
- If the user was using infection without git and no environment variables, it will not longer work.

## Reviewer Notes

I tried to keep things a bit tidy in the commits to help the review if so desired. It is not perfect though.

I renamed this from "refactor" to "feat" as there is too many – even if subtle and edge-case, BC breaks.

I am not happy about the last BC break. Previously this was not an issue because the git command executed when no environment variable was executed only if the GitHub annotations logger or GitLab one were registered.

An alternative is to allow `Configuration#projectDirectory = null` and fail only when instantiating the reporters if the reporters needed it are registered. Otherwise, we should accept this limitation. Note that whilst this is a BC break, the user can still work with it without installing git by passing the `--logger-project-root-directory` option.


src/Configuration/CiProjectDirectoryProvider.php
```diff
@@ -0,0 +1,83 @@
...
+use function getenv;
+use Infection\FileSystem\FileSystem;
+use function sprintf;
+use Webmozart\Assert\Assert;
+
+/**
+ * @final
+ * @internal
+ */
+readonly class CiProjectDirectoryProvider
```
theofidry: missing a test for it

tests/phpunit/Reporter/GitLabCodeQualityReporterTest.php
```diff
@@ -155,21 +158,8 @@ public static function metricsProvider(): iterable
...
-
-        $reporter = new GitLabCodeQualityReporter($resultsCollector, null);
-
-        $this->assertStringContainsString('"path":"foo\/bar"', $reporter->getLines()[0]);
-    }
-
     public function test_it_reports_correctly_with_custom_project_dir(): void
     {
-        \Safe\putenv('CI_PROJECT_DIR=/my/project/dir');
         self::setOriginalFilePrefix('/custom/project/dir/');
```
theofidry, line 153: to double check, this appears to have no effect, but it is out of scope of the PR

theofidry:
The end-to-end test failure was legit and I had to change a few things.

maks-rafalko (commented):
seen, make sense

PR #1632, opened 2021-12-30
Add new `--git-diff-lines` option to generate Mutants only for the changed *lines*

maks-rafalko:
I've posted a question recently in twitter (https://twitter.com/infection_php/status/1475841561695072259) asking what is the  most wanted feature needed in Infection for developers.

One of the interesting feature request was to mutate only changed/added **lines** (note: currently in `master` as a minimum, we can mutate only added/changed **files** by `--git-diff-filter=AM`)

So, here is an implementation of the new option - `--git-diff-lines` that parsers `git diff` output, finds changed/added lines and filters out all mutants that are not for these lines.

This option automatically mutates only **A**dded and **M**odified files (`--git-diff-filter=AM` under the hood).

Useful to check how your changes impacts MSI in a feature branch. Useful for those who do not want / can't write tests for the whole legacy file, but wants to cover their own changes.

@Ocramius - this is your feature request, if you have some time, please help us testing on one of your projects. I've prepared a PHAR for this PR: 

[infection.phar.zip](https://github.com/infection/infection/files/7793116/infection.phar.zip)

### How to use (as an example):

```bash
infection.phar --git-diff-lines -j4  -s 
```

by default it does a diff with `master` branch, but this can be changed:

```bash
infection.phar --git-diff-lines --git-diff-base=main -j4  -s 
```

----

- [x] Doc PR: [TBD](https://github.com/infection/site/pull/228)

PR #3028, opened 2026-03-24
test(ast): Add tests with changed lines ranges

theofidry:
## Description

When the ability to mutate only changed code was introduced (see https://github.com/infection/infection/pull/1632), there was no practical way to visualise or verify its behaviour. Since then, various refactorings (notably #2689, #2622 and #2380) have improved confidence in how the individual components behave.

However, there is currently no easy way to observe the impact that changed line ranges have on the traverse. https://github.com/infection/infection/pull/3007, for example, deliberately omitted that aspect, as the tooling to support it did not yet exist.

As a reminder, https://github.com/infection/infection/issues/2797 will introduce significant changes to the traverse. The less test coverage we have, the more likely regressions become, and they may not be immediately obvious (e.g. nodes that should be mutated no longer are). Such issues could remain undetected for a considerable time.

This PR aims to capture the impact of changed line ranges on the traverse.

## Changes

This section provides additional context on the approach taken, which may be useful for other contributors.

The primary way to visualise the traverse is currently the `debug:dump-ast` command, and its output can be captured in tests via `EnrichmentTraverseIntegrationTest`.

### Visualising the impact of diff filters

The first step is to visualise the effect of the various filtering and configuration options on the traverse of a file. This is achieved by adding those options to the `debug:dump-ast` command:

```shell
infection debug:dump-ast src/Git/CommandLineGit.php
# => AST where we see some eligible and mutation candidates

infection debug:dump-ast src/Git/CommandLineGit.php --git-diff-lines
# => as the file is not part of the diff, there is no longer any mutation candidates

infection debug:dump-ast src/Git/CommandLineGit.php --filter=CommandLineGit
# => AST where we see some eligible and mutation candidates
```

### Finer control over changed lines

The above is useful, but there are cases where one wants to observe the effect of specific changed lines (e.g. "what happens if the git diff reports lines X, Y and Z as changed?") without modifying the file system. To support this, the PR introduces the `--changed-lines-ranges` option.

When `--changed-lines-ranges` is provided, a `SimpleSourceLineMatcher` instance is registered in place of the git-based `SourceLineMatcher`, allowing the desired line ranges to be passed directly:

```shell
infection debug:dump-ast src/Git/CommandLineGit.php --changed-lines-ranges=102:124
```

This approach can be replicated in `EnrichmentTraverseIntegrationTest` by using `SimpleSourceLineMatcher` there as well, and adapting `LabelMutationCandidatesVisitor` to include the source line matching logic from `NodeMutationGenerator`.

### Other changes

To support the above, line numbers need to be visible in the AST dump. As the output is already fairly verbose, this is not enabled by default; a new option has been added to opt in.

## Reviewer Notes

Whilst this capability is valuable, tests involving changed lines tend to be quite verbose. It is recommended to keep them to a minimum and add them only where they provide meaningful coverage.

The `debug:dump-ast` command has been used extensively throughout this work to produce more readable diffs. Its UX is not yet optimal, if you encounter any rough edges whilst using it, please report them. The command's primary goals are (1) to provide a visual representation of the AST and (2) to allow copy-pasting the output into tests. Beyond those constraints, improvements are welcome.

## Related issues


theofidry:
@sanmai any thoughts on this?

theofidry:
The end-to-end error seems persistent, I will look into it, but it is entirely unrelated to this feature.

maks-rafalko (approved):
Makes sense  👍

sanmai (approved):
Tested, works 👍

PR #3033, opened 2026-03-30
refactor(tracing): Adjust the API of JUnitTestCaseSorter

theofidry:
## Description

Whilst resuming work on https://github.com/infection/infection/pull/2543 (recently unblocked by https://github.com/infection/infection/pull/2815), I investigated the remaining usages of the JUnit services that were left untouched at the time.

`JUnitTestCaseSorter` accumulates several naming, location, and documentation issues. I would like to address them separately from #2543 to keep that PR focused. The problems are not bugs – the behaviour is correct, but the naming has drifted far enough from what the service actually does that it warrants a dedicated clean-up.

`JUnitTestCaseTimeAdder` is an internal utility of `JUnitTestCaseSorter` and suffers from the same issues.

## Changes

This PR:

- Renames `JUnitTestCaseSorter` to `TestRunOrderResolver`. The rationale:
  - The service has no dependency on JUnit. The fact that its input may originate from a JUnit report is an implementation detail irrelevant to its API and behaviour.
  - It operates on `TestLocation` objects, not test cases.
  - "Sort/Sorter" implies an in-place re-ordering of the input. This service _extracts_ an ordered, deduplicated list — the output differs in both shape and content from the input.
- Since the service has no relationship with JUnit, it was moved out of that namespace. Co-locating it with `TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder` (the sole consumer) was considered, but that class may itself be restructured (see https://github.com/infection/infection/issues/2863). Placing it alongside `TestLocation` is not possible either, as that class lives in the separate `infection/abstract-testframework-adapter` package. Given that the service is directly related to trace information, it was placed in the `Trace` namespace instead.

### DIFF
--- a/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
+++ b/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
@@ -62,7 +62,7 @@ class MutationConfigBuilder extends ConfigBuilder
         private readonly string $originalXmlConfigContent,
         private readonly XmlConfigurationManipulator $configManipulator,
         private readonly string $projectDir,
-        private readonly TestRunOrderResolver $jUnitTestCaseSorter,
+        private readonly TestRunOrderResolver $testRunOrderResolver,
     ) {
     }
 
@@ -235,7 +235,7 @@ class MutationConfigBuilder extends ConfigBuilder
         $testSuite = $xPath->document->createElement('testsuite');
         $testSuite->setAttribute('name', 'Infection testsuite with filtered tests');
 
-        $orderedTestFilePaths = $this->jUnitTestCaseSorter->getOrderedTestFilePaths($tests);
+        $orderedTestFilePaths = $this->testRunOrderResolver->resolve($tests);
 
         foreach ($orderedTestFilePaths as $testFilePath) {
             $fileElement = $xPath->document->createElement('file', $testFilePath);
--- a/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
+++ b/src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
@@ -41,9 +41,9 @@ use DOMNodeList;
 use Infection\AbstractTestFramework\Coverage\TestLocation;
 use Infection\StreamWrapper\IncludeInterceptor;
 use Infection\TestFramework\Config\MutationConfigBuilder as ConfigBuilder;
-use Infection\TestFramework\Coverage\JUnit\JUnitTestCaseSorter;
 use Infection\TestFramework\PhpUnit\Config\XmlConfigurationManipulator;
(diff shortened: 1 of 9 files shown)


src/TestFramework/PhpUnit/Config/Builder/MutationConfigBuilder.php
```diff
@@ -235,12 +235,12 @@ private function addTestSuiteWithFilteredTestFiles(
         $testSuite = $xPath->document->createElement('testsuite');
         $testSuite->setAttribute('name', 'Infection testsuite with filtered tests');
 
-        $uniqueTestFilePaths = $this->jUnitTestCaseSorter->getUniqueSortedFileNames($tests);
+        $orderedTestFilePaths = $this->jUnitTestCaseSorter->getOrderedTestFilePaths($tests);
```
theofidry: forgot to commit the rename here


```suggestion
        $orderedTestFilePaths = $this->testRunOrderResolver->getOrderedTestFilePaths($tests);
```

tests/phpunit/TestFramework/Tracing/TestLocationSorterTest.php
```diff
@@ -33,85 +33,86 @@
...
+use Infection\TestFramework\Tracing\TestRunOrderResolver;
 use function log;
 use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
 
-#[CoversClass(JUnitTestCaseSorter::class)]
-final class JUnitTestCaseSorterTest extends TestCase
+#[CoversClass(TestRunOrderResolver::class)]
+final class TestLocationSorterTest extends TestCase
```
theofidry: forgot to rename here too

src/TestFramework/Tracing/TestRunOrderResolver.php
```diff
@@ -59,11 +63,13 @@ final class JUnitTestCaseSorter
...
     /**
+     * Returns unique test file paths ordered by execution time (fastest first).
+     *
      * @param TestLocation[] $tests
      *
-     * @return iterable<string>
+     * @return string[]
      */
-    public function getUniqueSortedFileNames(array $tests): iterable
+    public function getOrderedTestFilePaths(array $tests): array
```
theofidry: will sleep on this but `TestRunOrderResolver::getOrderedTestFilePaths()` does not sound amazing either. If I have a "resolver" I like to be able to just call "resolve", the same way that if you have a factory then you have a "create" and it's self explanatory. Here it feels a bit more mixed

theofidry:
@sanmai actually I notice `JUnitTestCaseTimeAdder` suffers from the same issue (as in: it is not related to JUnit and manipulates test locations), so I will likely propose the same changes to it too.

sanmai (approved):
👍🏻

PR #3026, opened 2026-03-23
refactor(ast): Move more logic out of the NodeMutationGenerator

theofidry:
## Description

As part of #2797, this is a continuation to move state out of the `NodeMutationGenerator` to the first (enrichment) traverse.

_This PR does not change which node is mutated_. There is no change of behaviour from that perspective. However, as show by `EnrichmentTraverseIntegrationTest`, some nodes that were previously marked as eligible (but not mutation candidates) are now marked as ineligible.

## Changes

- Instead of skipping nodes that are not part of a function signature or body in `NodeMutationGenerator`, this PR introduces `ExcludeNonMutableCodeVisitor` which marks such nodes as ineligible.
- Tests are now added as an attribute in `AddTestsVisitor` instead of being resolved in `NodeMutationGenerator`.
- Nodes that are not part of the changed lines are now marked as ineligible via `ExcludeUnchangedLinesVisitor` instead of directly skipped in `NodeMutationGenerator`.
- Remove `LabelMutationCandidatesVisitor` that was introduced in https://github.com/infection/infection/pull/3007. With this PR, the eligibility is sufficient to determine which node can be mutated.

## Reviewer Notes

I found out during the implementation that it was more practical to mark some nodes as eligible first, and then have a visitor mark it as ineligible. Indeed, if the node is not eligible in the first place, doing extra computing to know if it may be ineligible is a waste.

For the AST changes, the expected changes are as follows:

- A new attribute `tests` is available. Note that this does not indicate there is tests, the value is a `Closure` to allow the tests resolution to be lazily evaluated and the result is memoized to not re-compute the closure.
- The attribute `mutationCandidate` is removed. Any `eligible=true` node becomes `eligible=false` _unless_ it had `mutationCandidate=true`.

## Related issues

- https://github.com/infection/infection/issues/2797

PR #3039, opened 2026-04-03
fix(ast): Prevent invalid usage of @infection-ignore-all to corrupt the name resolution

theofidry:
## Description

Closes #3032.

Note that this comes at the cost of performance degradation. To what degree, I do not know, it's probably not gonna be noticeable unless someone had a heavy usage of `@infection-ignore-all` though, so it should not affect many users.

## Changes

- `IgnoreAllMutationsAnnotationReaderVisitor` has been renamed to `ExcludeIgnoredNodesVisitor`. A few notes on the name:
  - Currently it is the only way to mark nodes as ignored in the AST so there is ambiguity on what "ignore" refer to.
  - We no longer "ignore" the node, but mark is as ineligible. It is distinct from `Skip` which implies we skip the traversal of some nodes.
- Allowing to mark nodes as ineligible requires to either to mark nodes as eligible first and then mark it as ineligible (what this PR does) or mark as ineligible first and then when marking as eligible skip the nodes that were already marked as ineligible. From my experiments in #3026, the former was an simpler approach.


tests/phpunit/PhpParser/Visitor/ExcludeIgnoredNodesVisitor/MarkAllButIneligibleNodesAsVisitedVisitor.php
```diff
@@ -33,30 +33,24 @@
...
+use Infection\PhpParser\Visitor\MarkTraversedNodesAsVisitedVisitor;
 use PhpParser\Node;
+use PhpParser\NodeVisitorAbstract;
 
 /**
  * @internal
- * @final
  */
-class ChangingIgnorer implements NodeIgnorer
+final class MarkAllButIneligibleNodesAsVisitedVisitor extends NodeVisitorAbstract
```
theofidry, line 46: once again I do not know what GitHub is smoking here, this diff is incorrect, the previous class was deleted and `MarkAllButIneligibleNodesAsVisitedVisitor` is completely new.

src/PhpParser/Visitor/ExcludeIgnoredNodesVisitor.php
```diff
@@ -44,26 +44,24 @@
...
 {
     private const IGNORE_ALL_MUTATIONS_ANNOTATION = '@infection-ignore-all';
 
     /**
      * @param SplObjectStorage<object, mixed> $ignoredNodes
      */
     public function __construct(
-        private readonly ChangingIgnorer $changingIgnorer,
-        private readonly SplObjectStorage $ignoredNodes,
+        private readonly SplObjectStorage $ignoredNodes = new SplObjectStorage(),
```
theofidry: in the canonical usage it should not be passed, but for tests it's useful to be able to inject it to ensure it is properly cleaned up

theofidry:
Hmm found more issues, need more stuff before that

src/PhpParser/Visitor/ExcludeIgnoredNodesVisitor.php
```diff
@@ -44,38 +44,58 @@
...
     public function enterNode(Node $node): ?Node
     {
-        foreach ($node->getComments() as $comment) {
-            if (str_contains($comment->getText(), self::IGNORE_ALL_MUTATIONS_ANNOTATION)) {
-                $this->changingIgnorer->startIgnoring();
-                $this->ignoredNodes->offsetSet($node);
-            }
+        if (null !== $this->excludedStartNode) {
+            LabelNodesAsEligibleVisitor::markAsIneligible($node);
+        } elseif (self::hasIgnoreAnnotation($node)) {
```
sanmai: I kinda wish it would return early instead of elseif.

sanmai (approved):
Reviwed the new ExcludeIgnoredNodesVisitor, makes sense.

src/PhpParser/Visitor/ExcludeIgnoredNodesVisitor.php
theofidry: I wish we could just do return on `void` à la TypeScript tbh

theofidry:
Depends on https://github.com/infection/infection/pull/3038 and https://github.com/infection/infection/pull/3040

PR #3043, opened 2026-04-12
fix(PHPUnit): Do not modify the inital coverage configuration

### LINKED ISSUE #2440: "Class XX is not a valid target for code coverage"

### LINKED ISSUE #6472: Option to suppress code coverage target validation
References: https://github.com/infection/infection/issues/2440 and https://github.com/infection/infection/pull/2848

**Quick summary:**

Infection is a mutation testing tool. It makes small changes to a project's source code (such as inverting the condition inside an `if`), runs the test suite and checks whether the tests still pass.
Before doing any of this, Infection needs to do an initial run of the test suite to gather coverage information. It does this by generating a new, temporary PHPUnit config (based on the project's config) and invoking PHPUnit.

Infection supports a `--filter` option that allows users to specifcy a list of files or directories to consider. It will only generate mutations for those files.
If this option is used, the generated PHPUnit config is modified to only list the relevant files or directories in `<source><include>`.

**Problem:**

If a project uses `#[CoversClass()]`/`#[UsesClass()]`, PHPUnit (since version 12 I believe) will run some validation on those attributes in https://github.com/sebastianbergmann/phpunit/blob/197fc3445df847e7771bd8574e7f55d46fa1e44f/src/Runner/CodeCoverage.php#L228-L256
Because the full test suite is being run but the config `<source>` includes only a small subset of source files, an exception is thrown in https://github.com/sebastianbergmann/php-code-coverage/blob/3c97001bca11db4a3e6eee52f6e6924617364c3a/src/Target/Mapper.php#L82 and a "X is not a valid target for code coverage" warning is emitted.
This, combined with `failOnPhpunitWarning="true"` (the default and probably generally advisable setting), leads to a failed initial test run, which in turn leads Infection to abort its run.

**Possible solution:**

A number of options were explored in the issues I linked above. The TLDR is that this is difficult to properly handle on Infection's end.
The most straightforward solution would be to add a "don't validate code coverage targets" config option to PHPUnit.
Is that something PHPUnit would consider adding?

theofidry:
## Description

As of PHPUnit 12.0, the [source code configured to PHPUnit](https://docs.phpunit.de/en/12.5/code-coverage.html#including-files) is validated and fails when a condition is not met. This results in #2440 where, when infection execute the initial tests in "diff mode" (i.e. a filter is applied to execute a subset of the source code), if any test executed end up covering code that is not registered as part of the source code registered to PHPUnit.

A minimal reproducer is already doable via Infection's end-to-end tests:

<details>
<summary>Reproducer</summary>

Execute the following:

```shell
cd tests/e2e/PHPUnit_12-0
composer update
XDEBUG_MODE=coverage ../../../bin/infection # works
XDEBUG_MODE=coverage ../../../bin/infection --filter=Calculator # fails
```

It will fail with:

```
Runtime:       PHP 8.3.19 with Xdebug 3.4.2
Configuration:
/path/to/tests/e2e/PHPUnit_12-0/var/infection-tmp/infection/phpunitConfiguration.initial.infection.xml
Random Seed:   1776035509

............W

Time: 00:00.032, Memory: 10.00 MB

1 test triggered 1 PHPUnit warning:

1) Infection\E2ETests\PHPUnit_12_0\Tests\Covered\FunctionsTest::test_format_name_with_both_names
Function Infection\E2ETests\PHPUnit_12_0\Covered\formatName is not a valid target for code coverage

/path/to/tests/e2e/PHPUnit_12-0/tests/Covered/FunctionsTest.php:13

OK, but there were issues!
Tests: 13, Assertions: 24, PHPUnit Warnings: 1.

Generating code coverage report in PHPUnit XML format ... done [00:00.006]
```

The dumped XML looks like this:

```xml
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/phpunit.xsd" bootstrap="/Users/tfidry/Project/Humbug/infection/tests/e2e/PHPUnit_12-0/vendor/autoload.php" colors="false" testdoxSummary="true" cacheDirectory="var/phpunit-cache" executionOrder="defects,random" resolveDependencies="true" failOnRisky="true" failOnWarning="true" stopOnDefect="true" cacheResult="false" stderr="false">
  <testsuites>
    <testsuite name="Test Suite">
      <directory>/path/to/tests/e2e/PHPUnit_12-0/tests</directory>
    </testsuite>
  </testsuites>
  <source>
    <include>
      <file>/path/to/tests/e2e/PHPUnit_12-0/src/Uncovered/Calculator.php</file>
      <file>/path/to/tests/e2e/PHPUnit_12-0/src/Covered/Calculator.php</file>
    </include>
  </source>
</phpunit>
```

As a result, when `FunctionsTest` is checked, it covers `Infection\E2ETests\PHPUnit_12_0\Covered\formatName ` which is not part of the configured source code, causing the failure.

</details>

This change in the initial configuration file used was introduced in #1543 for performance reasons. Unfortunately, as of PHPUnit 12.0, this means we are introducing a performance degradation. I evaluated an egregious case via our benchmark source.

<details>
<summary>Performance impact assessment</summary>

```shell
git clone https://github.com/infection/benchmark-source
cd benchmark-source
composer install

# note we execute against `tests/AutoGenerated/Calculator` to save some time, as it is already a lot of tests
# and takes ~5min to run this benchmark on my machine
hyperfine 'XDEBUG_MODE=coverage vendor/bin/phpunit tests/AutoGenerated/Calculator --coverage-text'

# edit the phpunit.xml file to update the source code from `src` to `src/AutoGenerated/Calculator`

hyperfine 'XDEBUG_MODE=coverage vendor/bin/phpunit tests/AutoGenerated/Calculator --coverage-text'
```

The results are:

```
Benchmark 1: XDEBUG_MODE=coverage vendor/bin/phpunit tests/AutoGenerated/Calculator --coverage-text
  Time (mean ± σ):     11.039 s ±  0.420 s    [User: 9.665 s, System: 1.021 s]
  Range (min … max):   10.592 s … 11.981 s    10 runs

Benchmark 2: XDEBUG_MODE=coverage vendor/bin/phpunit tests/AutoGenerated/Calculator --coverage-text
  Time (mean ± σ):     11.033 s ±  1.062 s    [User: 9.484 s, System: 1.015 s]
  Range (min … max):   10.219 s … 13.864 s    10 runs
```

I was curious about the memory usage too so repeating the process but only executing PHPUnit once:

```
# with source=src
Time: 00:09.961, Memory: 129.00 MB
# with source=src/AutoGenerated/Calculator
Time: 00:09.983, Memory: 78.00 MB
```

</details>

The conclusion is that narrowing the source code defined to PHPUnit does not improve the execution time but does improve significantly the memory usage.

For context, PHPUnit improved a lot the coverage data collection and coverage report generation and I did not check the impact on the previous PHPUnit versions. So by precaution, I did not apply this change if the PHPUnit version detected is less than 12.0.

## Changes

No longer restrict the PHPUnit `source` element to the files/directories collected when executing infection with a filter, if PHPUnit 12.0 is detected.

## Related issues

- Closes https://github.com/infection/infection/issues/2440.
- Closes https://github.com/sebastianbergmann/phpunit/issues/6472
- Related to https://github.com/infection/infection/pull/2848 (a previous attempt).


theofidry:
ping @maks-rafalko @sanmai

sanmai (approved):
Makes sense

src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
```diff
@@ -106,6 +106,17 @@ private function buildPath(): string
 
     private function addCoverageNodes(string $version, SafeDOMXPath $xPath): void
     {
+        if (version_compare($version, '12.0', '>=')) {
+            // For PHPUnit 12.0+, preserve the original coverage configuration as-is.
```
maks-rafalko, line 110: can we add a reason here as it will be not clear why in some time without reading the PR description

theofidry: Done in https://github.com/infection/infection/pull/3071

PR #3046, opened 2026-04-15
feat(mago): Add support for mago static analysis 

### LINKED ISSUE #2393: Add support for `mago` as an option for `--static-analysis-tool`

Currently `infection` offers `phpstan` as an option, I would like to instead use `mago`

Support for `--static-analysis-tool=mago`

N/A

**Additional context**

We migrated our codebase from `psalm` to `mago` ( https://github.com/azjezz/psl/pull/527 ), previously we used `roave-infection-static-analysis-plugin` to integrate `psalm`, however, for mago, there is no such plugin, and we would rather support be added here instead of making an external third-party plugin.

blackwolf12333:
## Description

This introduces mago as a second static analysis tool we can use to kill
mutants that aren't killed by tests. It works pretty much the same as
the phpstan analysis except it uses a different tool.

For now we require mago >=1.20.0 because that's the only version I've
tested this on. But we can probably lower that version number a bit if
someone askes for it.

## Changes

- Adds support for mago static analysis tool

## Checklist

- [x] Tests added/updated
- [x] CHANGELOG.md updated (if deprecations/breaking changes)

## Reviewer Notes

I only checked that things work in the e2e test but I'm not 100% convinced that everything works as you'd expect.

- ~Is the mago output shown anywhere when a mutant is(n't) killed by mago~ Yes this is useful output which is saved in the logs to see *why* a mutant was killed.
- ~Mago does have a method to "replace" an existing file with a new one but that relies on stdin. I now have it running on the mutated file without any config changes and without providing the original file, I'm not sure that this will work in larger code bases where more things depend on context from the full code base analysis. Could use some insights on this.~ It does as of 1.23.0

## Related issues

Fixes https://github.com/infection/infection/issues/2393

### DIFF
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -152,6 +152,21 @@
                 }
             }
         },
+        "mago": {
+            "type": "object",
+            "description": "Mago additional settings.",
+            "additionalProperties": false,
+            "properties": {
+                "configDir": {
+                    "type": "string",
+                    "description": "Path to directory containing mago configuration file."
+                },
+                "customPath": {
+                    "type": "string",
+                    "description": "Custom path to mago executable."
+                }
+            }
+        },
         "ignoreMsiWithNoMutations": {
             "type": "boolean",
             "description": "Ignore MSI violations with zero mutations."
--- a/src/StaticAnalysis/Config/StaticAnalysisConfigLocator.php
+++ b/src/StaticAnalysis/Config/StaticAnalysisConfigLocator.php
@@ -51,6 +51,9 @@ readonly class StaticAnalysisConfigLocator implements TestFrameworkConfigLocator
         'neon',
         'neon.dist',
         'dist.neon',
+        'toml',
+        'json',
+        'yaml',
     ];
 
     public function __construct(
--- a/src/StaticAnalysis/Mago/Adapter/MagoAdapter.php
+++ b/src/StaticAnalysis/Mago/Adapter/MagoAdapter.php
@@ -1,0 +1,148 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+declare(strict_types=1);
+
+namespace Infection\StaticAnalysis\Mago\Adapter;
+
+use function array_merge;
+use Infection\Process\Factory\LazyMutantProcessFactory;
+use Infection\StaticAnalysis\Mago\Mutant\MagoMutantExecutionResultFactory;
+use Infection\StaticAnalysis\Mago\Process\MagoMutantProcessFactory;
+use Infection\StaticAnalysis\StaticAnalysisToolAdapter;
+use Infection\TestFramework\CommandLineBuilder;
+use Infection\TestFramework\VersionParser;
+use RuntimeException;
+use Safe\Exceptions\PcreException;
+use function sprintf;
+use Symfony\Component\Process\Exception\ProcessFailedException;
+use Symfony\Component\Process\Exception\ProcessSignaledException;
+use Symfony\Component\Process\Exception\ProcessTimedOutException;
+use Symfony\Component\Process\Process;
+use function version_compare;
+
+/**
+ * @internal
+ */
+final class MagoAdapter implements StaticAnalysisToolAdapter
+{
+    /**
+     * @param list<string> $staticAnalysisToolOptions
+     */
+    public function __construct(
+        private readonly MagoMutantExecutionResultFactory $mutantExecutionResultFactory,
+        private readonly string $staticAnalysisConfigPath,
+        private readonly string $staticAnalysisToolExecutable,
+        private readonly CommandLineBuilder $commandLineBuilder,
+        private readonly VersionParser $versionParser,
+        private readonly float $timeout,
+        private readonly array $staticAnalysisToolOptions,
+        private ?string $version = null,
+    ) {
+    }
+
+    public function getName(): string
+    {
+        return 'Mago';
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getInitialRunCommandLine(): array
+    {
+        $options = array_merge([
+            "--config=$this->staticAnalysisConfigPath",
+            'analyze',
+        ], $this->staticAnalysisToolOptions);
+
+        return $this->commandLineBuilder->build(
+            $this->staticAnalysisToolExecutable,
+            [],
+            $options,
+        );
+    }
+
+    public function createMutantProcessFactory(): LazyMutantProcessFactory
+    {
+        return new MagoMutantProcessFactory(
+            $this->mutantExecutionResultFactory,
+            $this->staticAnalysisToolExecutable,
+            $this->commandLineBuilder,
+            $this->timeout,
+            $this->staticAnalysisToolOptions,
+        );
+    }
+
+    /**
+     * @throws PcreException|ProcessTimedOutException|RuntimeException|ProcessSignaledException|ProcessFailedException
+     */
+    public function getVersion(): string
+    {
+        return $this->version ??= $this->retrieveVersion();
+    }
+
+    /**
+     * @throws RuntimeException|PcreException
+     */
+    public function assertMinimumVersionSatisfied(): void
+    {
+        $version = $this->getVersion();
+
+        if (version_compare($version, '1.23.0', '>=')) {
+            return;
+        }
+
+        throw new RuntimeException(sprintf(
+            'Infection requires Mago version >=1.23.0, but "%s" is installed.',
+            $version,
+        ));
+    }
+
+    /**
+     * @throws PcreException|ProcessTimedOutException|RuntimeException|ProcessSignaledException|ProcessFailedException
+     */
+    private function retrieveVersion(): string
+    {
+        $testFrameworkVersionExecutable = $this->commandLineBuilder->build(
+            $this->staticAnalysisToolExecutable,
+            [],
+            ['--version'],
+        );
+
+        $process = new Process($testFrameworkVersionExecutable);
+        $process->mustRun();
+
+        return $this->versionParser->parse($process->getOutput());
+    }
+}
--- a/src/StaticAnalysis/Mago/Process/MagoMutantProcessFactory.php
+++ b/src/StaticAnalysis/Mago/Process/MagoMutantProcessFactory.php
@@ -1,0 +1,104 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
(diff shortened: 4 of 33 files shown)


resources/schema.json
```diff
@@ -152,6 +152,17 @@
                 }
             }
         },
+        "mago": {
+            "type": "object",
+            "description": "Mago additional settings.",
+            "additionalProperties": false,
+            "properties": {
+                "customPath": {
```
maks-rafalko, line 164: for consistency, I would suggest to support the same config properties as in PHPStan. So `configDir` would also be useful here (if it makes sense in `mago` world)

theofidry (approved):
LGTM.

I couldn't bring myself to have a super deep review of it tbh. It is at the end of day a copy of PHPStan, and as such it suffers from the same issues, but those issues are to be discussed/addressed in #2863.

I played around a little with the tool and seems to be working. How well, to what extend, I'm not quite sure yet. I've also created #3049 to include it in our pipeline to increase familiarity with the tool.

Thanks a lot of the contribution, it is very appreciated!

tests/e2e/Mago_Integration/tests/SourceClassTest.php
```diff
@@ -0,0 +1,22 @@
...
+{
+    public function test_hello()
+    {
+        $sourceClass = new SourceClass();
+
+        $list = $sourceClass->makeAList(['a' => 'b', 'c' => 'd']);
+
+        $this->assertCount(2, $list);
+
+        // try to kill to not run phpstan
```
maks-rafalko:
```suggestion
        // try to kill to not run mago
```

tests/phpunit/Configuration/Entry/PhpStanTest.php
```diff
@@ -71,25 +71,25 @@ public static function basePathProvider(): iterable
 
         yield 'both paths are relative' => [
             new PhpStan(
-                'devTools/phpstan',
-                'devTools/phpstan/bin/phpstan',
+                'devTools/mago',
```
maks-rafalko: not sure this is the right file for these changes, it's for phpstan

src/StaticAnalysis/Mago/Process/MagoMutantProcessFactory.php
```diff
@@ -0,0 +1,103 @@
...
+    ): array {
+        $options = array_merge(
+            [
+                '--colors=never',
+                'analyze',
+                '--reporting-format=short',
+            ],
+            $this->staticAnalysisToolOptions,
+            [
+                $mutatedFilePath,
```
maks-rafalko: could you please explain how does it work? Does `mago` check only 1 file with this command line?

maks-rafalko (commented):
Thanks. I didn't try it yet, only did a brief code review. The main question from my side is how we run `mago` under the hood

- just for 1 file?
- or for the all files with one mutated being overridden

----

cc @azjezz - you were a feature requester and as the author of `mago` probably the best person to review this 🙏

maks-rafalko (changes requested):

resources/schema.json
azjezz: We don't support config directory, we only support `--config` arg in the binary for passing a config file. however, you can make it consistent by accepting only a directory, and appending `/mago.toml` to it. the problem could be if someone is naming the file differently, which mago allows if you are using `--config` arg.

src/StaticAnalysis/Config/StaticAnalysisConfigLocator.php
```diff
@@ -51,6 +51,7 @@
         'neon',
         'neon.dist',
         'dist.neon',
+        'toml',
```
azjezz, line 54: Note: mago also allows `mago.yaml` and `mago.json`.

src/StaticAnalysis/Mago/Adapter/MagoAdapter.php
```diff
@@ -0,0 +1,133 @@
...
+        $version = $this->getVersion();
+
+        if (version_compare($version, '1.20.0', '>=')) {
+            return;
+        }
+
+        throw new RuntimeException(sprintf(
+            'Infection requires Mago version >=1.20.0, but "%s" is installed.',
+            $version,
+        ));
```
azjezz: why 1.20 minimum? I don't mind any version constraint, as i always use latest, but i don't think we added anything in 1.20 that is significant.

src/StaticAnalysis/Mago/Process/MagoMutantProcessFactory.php
azjezz: Yes, Mago would consider the rest of the project to be part of vendor/ basically, and only analyze this file.

src/StaticAnalysis/Mago/Process/MagoMutantProcessFactory.php
```diff
@@ -0,0 +1,103 @@
...
+     * @return string[]
+     */
+    private function getMutantCommandLine(
+        string $mutatedFilePath,
+    ): array {
+        $options = array_merge(
+            [
+                '--colors=never',
+                'analyze',
+                '--reporting-format=short',
```
azjezz, line 91: Note: if the goal is to discard output here, its better to use `--reporting-format=count`, its the fastest option.

src/StaticAnalysis/Mago/Process/MagoMutantProcessFactory.php
maks-rafalko: but if e.g. we do a mutation in let's say base class

```diff
- public function foo()
+ protected function foo()
```

would `mago` in this case notice that non-public method is being called **in other files**? With PHPStan integration, it would re-analyze all the affected files (not only the one that is being mutated)

azjezz: Ah, yea, in that case, i think analyzing only the mutated file is not a good idea, i think analyzing the full project is fine, mago is fast so it shouldn't slow down infection by analyzing. e.g we use Infection in Psl, and `mago analyze` on the whole project is `0.16s`, trying to save few extra milliseconds yeilds false positives as you described. since files using the mutated file won't report an error.

src/StaticAnalysis/Mago/Adapter/MagoAdapter.php
blackwolf12333: Mostly that that is the version I tested this with and haven't had time yet to test with older versions.

src/StaticAnalysis/Mago/Process/MagoMutantProcessFactory.php
blackwolf12333: @maks-rafalko I wasn't really sure about this, do we care about the output of the killed mutants at all or is that discarded?

maks-rafalko: we do save the output of static analysis tool and IMO it's convenient to see in the logs **why** particular mutant has been killed, so yes we do use it, I would not cut it off

src/StaticAnalysis/Mago/Process/MagoMutantProcessFactory.php
maks-rafalko: ok, then what is our approach to run `mago` for the whole project replacing just 1 file (mutated one)? How to achieve that in `mago`?

blackwolf12333: @azjezz the `--stdin-input` option allows overriding a file inside the current project right? We could probably use that but then I need to hook up the mutant file through stdin.

azjezz: Just re-run the full analysis. mago does not do caching because it is fast. caching just adds trouble to us, we will always read from the filesystem. `--stdin` is not a good idea because it would be analzing only the mutated file, and not the rest of the project.

e.g. full analysis on magento using mago is 10x faster than fully cached phpstan: https://carthage-software.github.io/php-toolchain-benchmarks/?project=magento&kind=Analyzers 

Trying to scope analysis to a single changed file and its dependant would just introduce more headache.

maks-rafalko: > Just re-run the full analysis.

but we need to replace 1 file with a mutated one. How could we achieve it in `mago`?

azjezz: You mean re-write a single file without modifying it in the file system?

I think for this, the best way would be first class support for this in mago, something like `mago analyze --mutate /path/to/original/file.php:/path/to/mutated/file.php` - wdyt? If that's a good approach, i can implement it now and release a new version, and that would be the minimum supported version for that analyzer here.

azjezz: Ah! i see PHPStan has this already: https://github.com/infection/infection/blob/4ac75f98df36510040347a827237f7499ad0ae1e/src/StaticAnalysis/PHPStan/Process/PHPStanMutantProcessFactory.php#L97-L98 

Mago does not. 

Give me 10 minutes, i'm adding it :D

maks-rafalko: yeah, you found the correct place in PHPStan implementation :)

> You mean re-write a single file without modifying it in the file system?

Infection writes mutated file to the filesystem to configured temp directory, and PHPStan replaces 1 file by those `--tmp-file` / `--insteaf-of` options.

When we do the same with PHPUnit, we use stream wrapper to replace the file https://github.com/infection/include-interceptor

In both cases, original file is not modified.

azjezz: Added, releasing in a bit, `mago analyze --substitute /path/to/original.php=/path/to/mutated.php`

azjezz: ref https://github.com/carthage-software/mago/commit/a898fb400f7eb2c86b70e528eb288f41ce2a0631

maks-rafalko: Awesome, we will wait for a release to test it here and properly set min version. Looks promising!

azjezz: released 1.23.0, that should be used now, and be the minimum version ( note, build still in progress, and 1.23.1 is getting released right after - i'm bumping rust version ), 

ref https://mago.carthage.software/tools/analyzer/command-reference#substituting-files

src/StaticAnalysis/Mago/Adapter/MagoAdapter.php
blackwolf12333: Bumped to 1.23.0 for the new --substitute flag

src/StaticAnalysis/Mago/Process/MagoMutantProcessFactory.php
blackwolf12333: Thanks for the quick turnaround @azjezz!

src/StaticAnalysis/Mago/Adapter/MagoAdapter.php
```diff
@@ -78,7 +77,6 @@ public function getName(): string
     public function getInitialRunCommandLine(): array
     {
         $options = array_merge([
-            "--config=$this->staticAnalysisConfigPath",
```
azjezz: @blackwolf12333 i disagree with this change. `mago.toml` won't always be in root directory, for example, in https://github.com/php-standard-library/php-standard-library, we need to pass `--config config/mago.toml` as we put all tool configs in `config/` directory :/

blackwolf12333: Yeah, I was just about to revert it!

maks-rafalko (approved):
I've tested it locally, now it works as expected. Thank you!

```
cd tests/e2e/Mago_Integration
../../../bin/infection --logger-text=php://stdout --log-verbosity=all --debug
```

outputs

```
Killed by Static Analysis mutants:
==================================

1) /opt/infection/tests/e2e/Mago_Integration/src/SourceClass.php:19    [M] CastInt [ID] fe1d6aa33d552ab6dbf75301a506b728

@@ @@
         $strings = ['1'];

         $ints = array_map(function ($value): int {
-            return (int) $value;
+            return $value;
         }, $strings);

         $nonEmptyArray = ['1'];


$ '/opt/infection/tests/e2e/Mago_Integration/vendor/bin/mago' '--colors=never' 'analyze' '--reporting-format=short' '--substitute' '/opt/infection/tests/e2e/Mago_Integration/src/SourceClass.php=/opt/infection/tests/e2e/Mago_Integration/./infection/mutant.fe1d6aa33d552ab6dbf75301a506b728.infection.php'
  infection/mutant.fe1d6aa33d552ab6dbf75301a506b728.infection.php:19:20: error[invalid-return-statement]: Invalid return type for function `10229611493037670587:305`: expected `int`, but found `string('1')`.
  error: found 1 issues: 1 error(s)


2) /opt/infection/tests/e2e/Mago_Integration/src/SourceClass.php:28    [M] UnwrapArrayValues [ID] 400ddfa84f4f5076c65246652621690e

@@ @@

         $inlineNonEmpty = ['1'];

-        return array_values($values);
+        return $values;
     }

     /**

$ '/opt/infection/tests/e2e/Mago_Integration/vendor/bin/mago' '--colors=never' 'analyze' '--reporting-format=short' '--substitute' '/opt/infection/tests/e2e/Mago_Integration/src/SourceClass.php=/opt/infection/tests/e2e/Mago_Integration/./infection/mutant.400ddfa84f4f5076c65246652621690e.infection.php'
  infection/mutant.400ddfa84f4f5076c65246652621690e.infection.php:28:16: error[invalid-return-statement]: Invalid return type for function `Mago_Integration\SourceClass::makeAList`: expected `list<('T.mago_integration\sourceclass::makealist() extends mixed)>`, but found `array<array-key, ('T.mago_integration\sourceclass::makealist() extends mixed)>`.
  error: found 1 issues: 1 error(s)
```

The build failure is suspicious though:

```
Help: Ensure the `assertsame` method is defined in the `Infection\Tests\StaticAnalysis\Mago\Process\MagoMutantProcessFactoryTest` class-like.
Error: Method `assertfalse` does not exist on type `Infection\Tests\StaticAnalysis\Mago\Process\MagoMutantProcessFactoryTest`.
```

why does `mago` think `assertSame()` doesn't exist? (tests pass)

blackwolf12333:
@maks-rafalko I was just typing a comment about that. I think there might be a bug in mago here cause those methods obviously do exist otherwise the test would fail (@azjezz).

I am writing a commit that updates the baseline to include these issues for now (as well as a bunch of other things that I think got moved around by my additions and now no longer match the baseline).

azjezz:
@blackwolf12333 i will investigate afterwards, definitly a false positive, and high number of issues in the baseline is suspecious, i will take care of that.

theofidry:
@azjezz I hesitated to open a long list of issue because knowing SA, you often have the case that ignoring a bunch of issues causes a lot of confusion and false-positives or non-sense.

But if you have any question going through the list feel free to ping me on Discord too, or we could have a meta issue somewhere either here on infection or mago, as I don't think it should hold back this PR either

azjezz:
Yea, not a blocker for this PR. I'm working on fixing some other unrelated issues in Mago, then will look into it.

maks-rafalko:
Let's get it merged. Thanks @blackwolf12333 and @azjezz

PR #3048, opened 2026-04-16
fix(html-report): handle out-of-bounds line index in StrykerHtmlReportBuilder

medeirosinacio:
## Description

When a mutant has an `originalStartingLine` that exceeds the number of lines in the source file (e.g. due to a fatal error mutant), the array access in `StrykerHtmlReportBuilder::retrieveMutants()` returns `null`. This is then passed to `strlen()`, which throws a `TypeError` on PHP 8.1+ since it no longer accepts `null`.

This crashes the entire HTML report generation with exit code 255, even though the mutation testing itself completed successfully and all other report formats (JSON, log, per-mutator) were generated without issues.

## Steps to reproduce

1. Run mutation testing on a codebase where some mutants produce fatal errors (`E` status)
2. Enable the HTML report logger
3. Observe the crash:

```
PHP Warning: Undefined array key 80 in src/Reporter/Html/StrykerHtmlReportBuilder.php on line 237
PHP Fatal error: Uncaught TypeError: strlen(): Argument #1 ($string) must be of type string, null given in src/Reporter/Html/StrykerHtmlReportBuilder.php:238
```

## Changes

- Adds a null coalescing fallback (`''`) when accessing the source file lines array by index, so that an out-of-bounds line index results in a safe default (empty string) instead of a fatal error.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated — N/A (bug fix, no public API change)
- [x] CHANGELOG.md updated — N/A (no deprecations or breaking changes)
- [ ] Appropriate labels applied — `bug` (no permission to add labels, maintainers please apply)

## Reviewer Notes

The fix is intentionally minimal — a single `''` fallback. The root cause (why a mutant reports a line index beyond the file length) may warrant further investigation, but the report builder should be resilient to this regardless.

The regression test uses `ForHtmlReport.php` (49 lines) with `originalStartingLine = 80` to reproduce the out-of-bounds scenario and asserts the report is generated without errors with safe defaults for the column positions.

This PR:

- [x] Fixes a bug
- [x] Covered by tests

### DIFF
--- a/src/Reporter/Html/StrykerHtmlReportBuilder.php
+++ b/src/Reporter/Html/StrykerHtmlReportBuilder.php
@@ -234,7 +234,8 @@ final readonly class StrykerHtmlReportBuilder
                 $fileAsArrayOfLines = preg_split('/\n|\r\n?/', $originalCode);
                 $replacement = $this->retrieveReplacementFromDiff($result->getMutantDiff());
 
-                $originalCodeLine = $fileAsArrayOfLines[$result->getOriginalStartingLine() - 1];
+                $lineIndex = $result->getOriginalStartingLine() - 1;
+                $originalCodeLine = $fileAsArrayOfLines[$lineIndex] ?? '';
                 $originalCodeLineLength = strlen($originalCodeLine) + 1;
 
                 $startingColumn = $originalCodeLineLength - strlen(ltrim($originalCodeLine));
--- a/tests/phpunit/Reporter/Html/StrykerHtmlReportBuilderTest.php
+++ b/tests/phpunit/Reporter/Html/StrykerHtmlReportBuilderTest.php
@@ -450,6 +450,47 @@ final class StrykerHtmlReportBuilderTest extends TestCase
         );
     }
 
+    public function test_it_does_not_crash_when_mutant_line_exceeds_file_length(): void
+    {
+        $metricsCalculator = new MetricsCalculator(2);
+        $resultsCollector = new ResultsCollector();
+
+        // ForHtmlReport.php has 49 lines; use line 80 to trigger out-of-bounds
+        $result = self::createMutantExecutionResult(
+            DetectionStatus::ERROR,
+            <<<'DIFF'
+                @@ @@
+                -    $a = 1;
+                +    $a = 2;
+                DIFF,
+            'abc123outofbounds',
+            IgnoreMutator::class,
+            realpath(__DIR__ . '/../../Fixtures/ForHtmlReport.php'),
+            80,
+            80,
+            0,
+            0,
+            [
+                new TestLocation('TestClass::test_method1', '/infection/path/to/TestClass.php', 0.123),
+            ],
+            'Fatal error output',
+        );
+
+        $metricsCalculator->collect($result);
+        $resultsCollector->collect($result);
+
+        $report = (new StrykerHtmlReportBuilder($metricsCalculator, $resultsCollector))->build();
+
+        $mutants = current($report['files'])['mutants'];
+
+        $this->assertCount(1, $mutants);
+        $this->assertSame('abc123outofbounds', $mutants[0]['id']);
+        $this->assertSame(80, $mutants[0]['location']['start']['line']);
+        // column defaults to 0→1 since the line doesn't exist
+        $this->assertSame(0, $mutants[0]['location']['start']['column']);
+        $this->assertSame(1, $mutants[0]['location']['end']['column']);
+    }
+
     /**
      * @param array<int, TestLocation> $testLocations
      */
--- a/tests/phpunit/Reporter/Html/StrykerHtmlReportBuilderTest.php
+++ b/tests/phpunit/Reporter/Html/StrykerHtmlReportBuilderTest.php
@@ -464,7 +464,7 @@ final class StrykerHtmlReportBuilderTest extends TestCase
                 +    $a = 2;
                 DIFF,
             'abc123outofbounds',
-            IgnoreMutator::class,
+            PublicVisibility::class,
             realpath(__DIR__ . '/../../Fixtures/ForHtmlReport.php'),
             0,
             0,
--- a/tests/phpunit/Reporter/Html/StrykerHtmlReportBuilderTest.php
+++ b/tests/phpunit/Reporter/Html/StrykerHtmlReportBuilderTest.php
@@ -455,7 +455,7 @@ final class StrykerHtmlReportBuilderTest extends TestCase
         $metricsCalculator = new MetricsCalculator(2);
         $resultsCollector = new ResultsCollector();
 
-        // ForHtmlReport.php has 49 lines; use line 80 to trigger out-of-bounds
+        // Errored mutants can report originalStartingLine=0, making the index -1
         $result = self::createMutantExecutionResult(
             DetectionStatus::ERROR,
             <<<'DIFF'
@@ -466,8 +466,8 @@ final class StrykerHtmlReportBuilderTest extends TestCase
             'abc123outofbounds',
             IgnoreMutator::class,
             realpath(__DIR__ . '/../../Fixtures/ForHtmlReport.php'),
-            80,
-            80,
+            0,
+            0,
             0,
             0,
             [
@@ -485,8 +485,7 @@ final class StrykerHtmlReportBuilderTest extends TestCase
 
         $this->assertCount(1, $mutants);
         $this->assertSame('abc123outofbounds', $mutants[0]['id']);
-        $this->assertSame(80, $mutants[0]['location']['start']['line']);
-        // column defaults to 0→1 since the line doesn't exist
+        $this->assertSame(0, $mutants[0]['location']['start']['line']);
         $this->assertSame(0, $mutants[0]['location']['start']['column']);
         $this->assertSame(1, $mutants[0]['location']['end']['column']);
     }


theofidry:
Thanks for the bug report & the fix!

Could you clarify when that happens? Because I'm struggling to see what a fatal error mutant would trigger it and not another one. If you have a concrete example of the source code + the mutation applied that would be super helpful!

medeirosinacio:
> Thanks for the bug report & the fix!
> 
> Could you clarify when that happens? Because I'm struggling to see what a fatal error mutant would trigger it and not another one. If you have a concrete example of the source code + the mutation applied that would be super helpful!

output

```shel
    ____      ____          __  _

#StandWithUkraine

Infection - PHP Mutation Testing Framework version 0.27.11

[notice] You are running Infection with PCOV enabled.

Running initial test suite...

PHPUnit version: 10.5.63

    0 [>---------------------------] < 1 ms
    1 [->--------------------------]    3 s
    2 [--->------------------------]    3 s
   16 [----->----------------------]    3 s
   52 [------->--------------------]    3 s
   72 [--------->------------------]    3 s
   90 [----------->----------------]    3 s
  113 [------------->--------------]    3 s
  133 [-------------->-------------]    3 s
  183 [---------------->-----------]    3 s
  241 [------------------>---------]    3 s
  292 [-------------------->-------]    3 s
  327 [---------------------->-----]    3 s
  360 [------------------------>---]    4 s
  420 [-------------------------->-]    4 s
  474 [>---------------------------]    4 s
  538 [->--------------------------]    4 s
  570 [--->------------------------]    4 s
  609 [----->----------------------]    4 s
  659 [------->--------------------]    4 s
  691 [--------->------------------]    4 s
  718 [----------->----------------]    4 s
 1007 [------------->--------------]    4 sProcessing source code files:  0/78

Generate mutants...


Processing source code files:  8/78
Processing source code files: 16/78
Processing source code files: 32/78
Processing source code files: 55/78
Processing source code files: 63/78
Processing source code files: 71/78
Processing source code files: 78/78
.: killed, M: escaped, U: uncovered, E: fatal error, X: syntax error, T: timed out, S: skipped, I: ignored

UUUUUUU...........................................   (  50 / 1041)
.........M..................................M..M..   ( 100 / 1041)
......M...........................MM.......MMM....   ( 150 / 1041)
........................MMM...............M.MM....   ( 200 / 1041)
.....M...............................M.MM.....MM.M   ( 250 / 1041)
E......................M....M.....................   ( 300 / 1041)
.....................M....E......E.MM....M.......M   ( 350 / 1041)
..................................................   ( 400 / 1041)
........M.................M......M.....MM.........   ( 450 / 1041)
.....................................M.M..........   ( 500 / 1041)
........................M..........M.M............   ( 550 / 1041)
.................................................M   ( 600 / 1041)
...........M.............M........................   ( 650 / 1041)
..................MM...M..M..M..M............M....   ( 700 / 1041)
........M.M.......................................   ( 750 / 1041)
M..M..M.MM.MM.....................................   ( 800 / 1041)
..........................M......................M   ( 850 / 1041)
.....M....MM..M.M..............E........M........M   ( 900 / 1041)
E.......MMMMM...M....M..M...M...MM....MMM..MM.....   ( 950 / 1041)
..MM......M.MM.......M............................   (1000 / 1041)
..............MM.........M..MMM...M...MM.            (1041 / 1041)

1041 mutations were generated:
     931 mutants were killed
       7 mutants were not covered by tests
      98 covered mutants were not detected
       5 errors were encountered
       0 time outs were encountered

Metrics:
         Mutation Score Indicator (MSI): 89%
         Mutation Code Coverage: 99%
         Covered Code MSI: 90%

Generated Reports:
         - /app/tests/reports/infection/infection.log
         - /app/tests/reports/infection/infection.html
         - /app/tests/reports/infection/summary.log
         - /app/tests/reports/infection/infection-log.json
         - /app/tests/reports/infection/per-mutator.md

PHP Warning:  Undefined array key 80 in /app/vendor/infection/infection/src/Logger/Html/StrykerHtmlReportBuilder.php on line 226
PHP Fatal error:  Uncaught TypeError: strlen(): Argument #1 ($string) must be of type string, null given in /app/vendor/infection/infection/src/Logger/Html/StrykerHtmlReportBuilder.php:227
Stack trace:
#0 [internal function]: Infection\Logger\Html\StrykerHtmlReportBuilder->Infection\Logger\Html\{closure}()
#1 /app/vendor/infection/infection/src/Logger/Html/StrykerHtmlReportBuilder.php(221): array_map()
#2 /app/vendor/infection/infection/src/Logger/Html/StrykerHtmlReportBuilder.php(195): Infection\Logger\Html\StrykerHtmlReportBuilder->retrieveMutants()
#3 /app/vendor/infection/infection/src/Logger/Html/StrykerHtmlReportBuilder.php(169): Infection\Logger\Html\StrykerHtmlReportBuilder->retrieveFiles()
#4 /app/vendor/infection/infection/src/Logger/Html/StrykerHtmlReportBuilder.php(106): Infection\Logger\Html\StrykerHtmlReportBuilder->getFiles()
#5 /app/vendor/infection/infection/src/Logger/Html/HtmlFileLogger.php(79): Infection\Logger\Html\StrykerHtmlReportBuilder->build()
#6 /app/vendor/infection/infection/src/Logger/Html/HtmlFileLogger.php(68): Infection\Logger\Html\HtmlFileLogger->getMutationTestingReport()
#7 /app/vendor/infection/infection/src/Logger/FileLogger.php(61): Infection\Logger\Html\HtmlFileLogger->getLogLines()
#8 /app/vendor/infection/infection/src/Logger/FederatedLogger.php(56): Infection\Logger\FileLogger->log()
#9 /app/vendor/infection/infection/src/Logger/FederatedLogger.php(56): Infection\Logger\FederatedLogger->log()
#10 /app/vendor/infection/infection/src/Event/Subscriber/MutationTestingResultsLoggerSubscriber.php(52): Infection\Logger\FederatedLogger->log()
#11 /app/vendor/infection/infection/src/Event/EventDispatcher/SyncEventDispatcher.php(59): Infection\Event\Subscriber\MutationTestingResultsLoggerSubscriber->onMutationTestingWasFinished()
#12 /app/vendor/infection/infection/src/Process/Runner/MutationTestingRunner.php(123): Infection\Event\EventDispatcher\SyncEventDispatcher->dispatch()
#13 /app/vendor/infection/infection/src/Engine.php(134): Infection\Process\Runner\MutationTestingRunner->run()
#14 /app/vendor/infection/infection/src/Engine.php(74): Infection\Engine->runMutationAnalysis()
#15 /app/vendor/infection/infection/src/Command/RunCommand.php(397): Infection\Engine->execute()
#16 /app/vendor/infection/infection/src/Command/BaseCommand.php(75): Infection\Command\RunCommand->executeCommand()
#17 /app/vendor/symfony/console/Command/Command.php(341): Infection\Command\BaseCommand->execute()
#18 /app/vendor/symfony/console/Application.php(1094): Symfony\Component\Console\Command\Command->run()
#19 /app/vendor/symfony/console/Application.php(356): Symfony\Component\Console\Application->doRunCommand()
#20 /app/vendor/symfony/console/Application.php(195): Symfony\Component\Console\Application->doRun()
#21 /app/vendor/infection/infection/bin/infection(85): Symfony\Component\Console\Application->run()
#22 /app/vendor/bin/infection(119): include('...')
#23 {main}
  thrown in /app/vendor/infection/infection/src/Logger/Html/StrykerHtmlReportBuilder.php on line 227
Script infection --threads=max --test-framework-options='--exclude-group=swoole' handling the test:mutation event returned with error code 255
Script composer test:mutation handling the ci event returned with error code 255

make: *** [ci] Error 255

Process finished with exit code 2
```

medeirosinacio:
I dug into the `infection-log.json` from that run. All 5 errored mutants have `originalStartingLine: 0` and `originalEndingLine: 0`. The mutators involved were `Increment`, `Foreach_`, `NotIdentical`, `PublicVisibility`, `Coalesce`, `UnwrapArrayValues`, `LogicalOrNegation`, `LogicalAnd` — nothing exotic, just regular mutators that happened to produce fatal errors at runtime.

So the crash path is: `getOriginalStartingLine()` returns `0` → line index becomes `0 - 1 = -1` → `$fileAsArrayOfLines[-1]` is an undefined key (PHP arrays from `preg_split` are 0-indexed, no negative keys) → `null` gets passed to `strlen()` → `TypeError` on PHP 8.1+.

The "Undefined array key 80" in the stack trace confused me at first too — I think it might be a different mutant in the same `array_map()` batch, or possibly the key numbering shifts after `array_filter`/`array_slice` in `retrieveReplacementFromDiff()` (those preserve original keys). But the root cause is the same: out-of-bounds access on the lines array.

I do not have a minimal reproducer for _why_ those mutants end up with `originalStartingLine: 0` — the source files are regular PHP classes with nothing unusual. Could be related to how the process crash is captured, or maybe the line info is lost when the mutant process exits abnormally. That might be worth a separate investigation.

In any case, the report builder shouldn't crash regardless of what line value it gets from the mutant result. I'll update the test to use `originalStartingLine: 0` since that's what actually happens in practice.

theofidry:
>I do not have a minimal reproducer for why those mutants end up with originalStartingLine: 0

I feel that's unfortunately the root of the issue.

Your patch looks fine on the surface, and I'm fine with it, but I can't really make sense of how this could happen and I suspect there is a deeper issue here. So I would be keen to address the real issue rather than a symptom.

`originalStartingLine=0` is super weird: this would mean you're updated the very first line which usually is `<?php`, so either you have some exotic code (which would be helpful to have a sample of) or there is something wrong in the way those properties are calculated/applied, in which case there may be more bugs as a result, but not necessarily causing a straight up error.

medeirosinacio:
@theofidry I went deeper into this and I owe you a correction on my earlier analysis.

The JSON log actually shows valid `originalStartLine` values for all errored mutants (184, 56, 75, 291, 89). They're not zero in the serialized output. But `Mutation::getOriginalStartingLine()` does `(int) $this->attributes['startLine']` — if that key is missing from the PhpParser node attributes, the cast silently returns `0`, and then `0 - 1 = -1` is an undefined key.

I wrote a standalone reproducer to confirm the mechanics:

```php
$code = '<?php
final class Foo {
    public function bar(): int {
        for ($i = 0; $i < 10; ++$i) { echo $i; }
        return 42;
    }
}';

$lines = preg_split('/\n|\r\n?/', $code);
// 8 lines → keys 0..7

$lines[0 - 1];   // Warning: Undefined array key -1 → null
strlen(null);      // TypeError on PHP 8.1+

$lines[80 - 1];   // Warning: Undefined array key 79 → null
strlen(null);      // same crash
```

Both paths crash the same way — the `?? ''` fix handles both.

Now, I couldn't reproduce `originalStartingLine = 0` inside Infection itself with a minimal example. My project runs on Hyperf 3.x / Swoole 5.x and the errored mutants all crash during framework bootstrap — the process output shows `unserialize()` warnings from Hyperf's DI annotation scanner before any test even runs. I suspect something about that early crash path causes the node attributes to be incomplete, but I haven't pinned down exactly where.

The attributes come from `$node->getAttributes()` in `NodeMutationGenerator`, which should always include `startLine` from PhpParser. Why it would be missing for these specific nodes, I don't know yet. I'm going to try to set up a minimal Hyperf project and reproduce this end-to-end — if I can nail down the root cause I'll open a separate issue for it.

In the meantime, the `?? ''` at least prevents the report builder from taking down the whole run, which feels like the right thing regardless of what's happening upstream.

medeirosinacio:
@theofidry I dug into the root cause properly this time. Went through the code paths, ran some tests, and here's where I landed.

`Mutation::getOriginalStartingLine()` casts `$this->attributes['startLine']` to int. If that key is missing, the cast silently returns `0`. The attributes come from `$node->getAttributes()` in `NodeMutationGenerator` (line 148), passed straight into the `Mutation` constructor without validation.

I parsed 5,203 PhpParser nodes from our project, including Hyperf proxy classes that go through a full parse-transform-print-reparse cycle. All of them had `startLine`. PhpParser's lexer always populates it. I also traced all four `MutantExecutionResult` construction paths (`createFromProcess`, `createFromNonCoveredMutant`, `createFromTimeSkippedMutant`, `createFromIgnoredMutant`) — they all go through `$mutation->getOriginalStartingLine()`, no alternative path that could zero it out. Our 10 errored mutants in the JSON log all have valid `originalStartLine` values (56, 75, 98, etc.) — the errors are from Hyperf's annotation scanner failing on mutated source (`unserialize()` failures), not from missing line info.

So in the standard flow with parser-produced nodes, `originalStartingLine = 0` shouldn't happen. The scenario that would produce it: a `Mutation` built with node attributes missing `startLine` — nodes created programmatically by custom mutators, test fixtures, or PhpParser plugins that don't set position attributes. The `Mutation` constructor doesn't enforce that `startLine` exists, though fixing that feels like a separate issue.

The crash itself I was able to reproduce directly. Under `strict_types=1`:

- `originalStartingLine = 0` → index `-1` → `null` from `preg_split` result → `strlen(null)` → `TypeError`
- Same thing for any line number past the end of the file

The `?? ''` guards against both without changing behavior for valid inputs.

Let me know if you'd rather take a different approach here.

theofidry:
OK I can see several problems here.

First thing first, the version you are using is [0.27.11](https://github.com/infection/infection/releases/tag/0.27.11) which is from last year. The latest is [0.32.6](https://github.com/infection/infection/releases/tag/0.32.6).

Please update before anything else: even if we were to accept your patch as is, it would only be released in 0.32.7, not 0.27.1. We are not actively maintaining multiple branches.

<hr/>

If it is established that the problem persists, before anything, identify which file causes the issue and try to use the following command:

```shell
infection debug:dump-ast <file> --show-attributes
```

This should print the complete AST and will show if `startLine` is properly set or not.

>The scenario that would produce it: a Mutation built with node attributes missing startLine — nodes created programmatically by custom mutators, test fixtures, or PhpParser plugins that don't set position attributes

Which is not your scenario, so irrelevant. Mutations are created by `NodeMutationGenerator` which operates on an enriched AST.

To me this indicates (unless you the dumped command shows missing attributes) that this is more related to your setup with hyperf/swoole. I'm not suggesting this is an invalid scenario, but it's unclear to me where the bug lies. If the problem is there, the suggested solution is no longer fixing a symptom, but hiding a bigger problem. As such, I think it is critical to be able to establish a minimal reproducer to be able to correctly identify the problem, and fix it in the right place. Maybe it's a bug in infection, maybe it's a mis-use in your setup, maybe it's infection not orchestrating such tests correctly with parallelism, maybe it's a serialization bug in hyperf/swoole, who knows. To be able to properly investigate, a reproducer is needed.

PR #3049, opened 2026-04-16
build(mago): Add Mago

theofidry:
In light of #3046 I propose to adapt Mago on top of the existing PHPStan and Psalm integrations.

I do not have an opinion on the quality of usefulness of the tool, but having it alongside PHPStan and Psalm will allow us to have a better feel for how it works out for us.

As always, there is the risk of it getting more in the way than anything. I quickly tried it and it supports baseline, so I do not think it will get more in the way than PHPStan or Psalm can, silencing undesired issues can be done by updated the baseline.

Note that I included as a dev dependency since the package is here just to provide a rust binary, there is no other package dependency, hence having it as a dev dependency allows to keep things simple.


mago.toml
```diff
@@ -0,0 +1,51 @@
...
+
+[source]
+workspace = "."
+paths = [
+    "src",
+    "tests/autoloaded",
+    "tests/benchmark",
+    "tests/phpunit"
+]
+includes = ["vendor"]
```
maks-rafalko, line 14: I haven't used `mago` before, but didn't find `vendor` in the docs. What does it mean here? Why is it included?

https://mago.carthage.software/tools/analyzer/configuration-reference

mago.toml
```diff
@@ -0,0 +1,51 @@
...
+excludes = [
+    "tests/benchmark/MutationGenerator/sources",
+    "tests/benchmark/Tracing/benchmark-source",
+    "tests/benchmark/Tracing/coverage"
+]
+
+[source.glob]
+literal-separator = true
+
+[formatter]
```
maks-rafalko: does it mean it works as alternative to PHP-CS-Fixer as well? Do we need it? (as from the description, I thought we use it as a static analyzer).

I don't have an opinion yet, just asking was it added on purpose or not.

mago.toml
```diff
@@ -0,0 +1,51 @@
...
+analyze-dead-code = true
+memoize-properties = true
+allow-possibly-undefined-array-keys = true
+check-throws = true
+check-missing-override = false
+find-unused-parameters = true
+strict-list-index-checks = false
+no-boolean-literal-comparison = true
+check-missing-type-hints = true
+register-super-globals = true
```
maks-rafalko, line 46: Where these rules come from? How were they chosen? Or is it some default list?

I found this strict config, not sure we need it, again - asking out of curiosity as I didn't use it before

https://mago.carthage.software/tools/analyzer/configuration-reference#maximum-strictness-configuration

sidz:
My 5 cents: I don't think we need to bring every static analysis tool into our pipeline for me it sounds like redundant and mostly doesn't have any value.

p.s. I don't have anything against mago or rust.. Just don't see any benefit.

If we will give it a go and merge to master branch: should we also include [github.com/phan/phan](https://github.com/phan/phan)?

mago.toml
azjezz: [docs](https://mago.carthage.software/guide/configuration#source-section)


TL;DR: this is where mago will look for dependencies, i.e code the analyzer needs for type inference, but won't analyze.

mago.toml
azjezz: ref https://mago.carthage.software/tools/analyzer/configuration-reference#feature-flags

theofidry: Those were the defaults prompted with `mago init`, I just saw some names where I thought this seems like something we want so I changed it from `false` to `true`.

mago.toml
theofidry: It does offer that alternative, I did not attempt to use it neither was I looking into replacing PHP-CS-Fixer with it at that point.

This was added as part of the init command, I'll try to remove it, no idea if it's wise to do so

azjezz: You can remove anything that you dont need, mago is fine with a completely empty `mago.toml` too :D

theofidry:
>I don't think we need to bring every static analysis tool into our pipeline for me it sounds like redundant and mostly doesn't have any value.

Agreed.

>If we will give it a go and merge to master branch: should we also include [github.com/phan/phan](https://github.com/phan/phan)?

If someone in the teams really wants it, why not?

Personally I've never had a good experience with Phan, so no, and we have no bridge for phan in the work either.

The reason I propose Mago is because to date, the only good alternative to PHPStan was Psalm. Unfortunately the latter has grown increasingly disappointing, to the point where we cannot even upgrade to the latest version and the issue, despite affecting many, was not addressed yet (see https://github.com/infection/infection/pull/2565). Mago on the other hand seems promising, but I've never used it myself, so hard to judge.

Given that it is fast enough to not cause more slowdowns and that the baseline feature akin to the PHPStan one that is very easy to keep up to date, I think it's in a good position to give it a go and see it side by side.

sanmai:
I'd seriously consider dropping Psalm at this point.

sidz:
Fine with me. We could give mago a go, monitor it for a while and if we think all good with it just drop psalm after.

theofidry:
Also to be clear: if Mago just gets in the way and always prove to be subpar, we can drop it; I'm proposing this as a test because it requires very little efforts to add **and** remove it.

maks-rafalko (approved):

maks-rafalko:
> Fine with me. We could give mago a go, monitor it for a while and if we think all good with it just drop psalm after.

I would say removing Psalm shouldn't be connected to adding (or removing) mago  :). We can do it at any point of time, which I also vote for

PR #3042, opened 2026-04-12
fix(PHPUnit): Remove PHPUnit XSD validation when executing the initial tests

theofidry:
- [ ] Extract unrelated changes
- [ ] Remove XSD E2E test

Closes #2303.

My takeaway from the discussion is that PHPUnit already validates its configuration, there is no need for us to validate it on top of it.

Note that only affects the PHPUnit XML file used to execute the initial tests, we did not, AFAICS, validate the XML files for each configuration file dumped for the respective mutants.

PR #3051, opened 2026-04-16
feat(xml): Throw a dedicated InvalidXml exception when the XML loaded by XPath is invalid

theofidry:
This PR builds on top of https://github.com/infection/infection/pull/3050.

## Description

Make `SafeDOMXPath` throw a dedicated `InvalidXml` exception that is a `RuntimeException` rather than, previously, a `LogicException` or `Error`, when the XML to load is invalid.

This is a preparatory step for https://github.com/infection/infection/pull/3042 (or an extract PR) where we will catch this exception in order to provide a more helpful feedback to the user when using an invalid PHPUnit configuration.

## Changes

- Move `TestFramework\SafeDOMXPath` to `TestFramework\XML\SafeDOMXPath`.
- Introduce `TestFramework\XML\InvalidXml`.
- Replace thrown `InvalidArgumentException` and `ValueError` by `InvalidXml`.


theofidry:
~~Marking as a draft to not accidentally merge it before https://github.com/infection/infection/pull/3050 but it is otherwise ready for review.~~

If you only want to see the exception change I recommend to check the last commit only, it is significantly less noisy.

sanmai (approved):

PR #3052, opened 2026-04-17
refactor: Leverage the Filesystem to dump files

theofidry:
## Description

This PR replaces the usages of `file_put_contents` in `InitialConfigBuilder`, `MutationConfigBuilder` and `ConfigureCommand` with the `Filesystem`.

Like previous FS refactors done in the past, the goal is to provide a better abstraction towards the FileSystem, making it easier to give some guarantees on the application side (e.g. not to write anything when using `--dry-run`) and making it possible to mock the filesystem for tests (see #2833) and also helps to speed up the tests (less I/O).

Note however that due to our overly simplistic implementation of IoDetector, we cannot remove the usage of the `integration` group despite the updated tests not requiring any I/O anymore.

With this PR, the only usage of `file_put_contents()` left is `FileReporter` to write to a stream, which `Filesystem::dumpFile()` does not handle.

There is a few more non-abstracted calls to the I/O, e.g. `realpath()` or `file_exists()`, but I prefer to address them separately.

## Changes

- Leverage `Filesystem` in `InitialConfigBuilder`, `MutationConfigBuilder` and `ConfigureCommand`.
- Introduce a `InMemoryFileSystem` to make it easier to mock `Filesystem`.

## Reviewer Notes

This PR uncovered a few type related issues in InitialConfigBuilder` and `MutationConfigBuilder` about the `saveXml()` potentially returning `false`.

I left this alone to reduce the amount of noise and plan to address it in a dedicated PR.


devTools/phpstan-baseline.neon
```diff
@@ -102,6 +102,65 @@ parameters:
...
+			rawMessage: 'Method Infection\FileSystem\InMemoryFileSystem::remove() has parameter $files with no value type specified in iterable type iterable.'
+			identifier: missingType.iterableValue
+			count: 1
+			path: ../src/FileSystem/InMemoryFileSystem.php
+
+		-
+			rawMessage: 'Method Infection\FileSystem\InMemoryFileSystem::touch() has parameter $files with no value type specified in iterable type iterable.'
+			identifier: missingType.iterableValue
+			count: 1
+			path: ../src/FileSystem/InMemoryFileSystem.php
```
sanmai: looks like these should be as easy to fix as `iterable<string>`

sanmai (approved):

devTools/phpstan-baseline.neon
theofidry: It is but it's an inherited method so the proper place to fix it is in the Symfony `Filesystem` rather. But I realise we had a cleaner fix already for `FakeFileSystem`:

```neon
        # We are not too interested in making the FakeFileSystem correct typewise.
        -
            message: "#Do not use magic number as default parameter#"
            path: ../src/FileSystem/FakeFileSystem.php
        -
            identifier: missingType.iterableValue
            path: ../src/FileSystem/FakeFileSystem.php
```

PR #3061, opened 2026-04-18
build: Permit insecure PHPUnit 9.x-11.x in E2E tests

sanmai:
## Description

Looking at this build error:

>    - Root composer.json requires phpunit/phpunit ^9.6.20, found phpunit/phpunit[9.6.20, ..., 9.6.34] but these were not loaded, because they are affected by security advisories ("PKSA-5jz8-6tcw-pbk4", "PKSA-z3gr-8qht-p93v"). Go to https://packagist.org/security-advisories/ to find advisory details. To ignore the advisories, add them to the audit "ignore" config. To turn the feature off entirely, you can set "block-insecure" to false in your "audit" config.
441

The advisory is [PKSA-5jz8-6tcw-pbk4](https://github.com/advisories/GHSA-qrr6-mg7r-m243), affected versions >=13.0.0,<=13.1.5|<=12.5.21. That covers every 9.x, 10.x, 11.x, and 12.x up to 12.5.21 - so no 9.x fix exists. Only 12.5.22+ / 13.1.6+ are safe. 

> pull request from an untrusted contributor that modifies phpunit.xml to include a newline-containing INI value, executed by a CI system that runs PHPUnit against the PR without isolation ... inject arbitrary additional directives into the child's configuration, including auto_prepend_file

There are proposed workarounds, but we still need to permit installing these versions. So let's do just that.

I also had to modify the main composer.json - review my comments below.

## Changes

- Adds `audit.block-insecure = false` for the affected E2E tests.

```
for d in $(git grep phpunit/phpunit tests/e2e | grep -F composer.json  | cut -f1 -d: | xargs -I{} dirname {}); do (cd $d; composer config audit.block-insecure false); done
```

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -104,6 +104,12 @@
         "allow-plugins": {
             "phpstan/extension-installer": true,
             "infection/extension-installer": true
+        },
+        "audit": {
+            "ignore": {
+                "PKSA-5jz8-6tcw-pbk4": "PHPUnit argument injection - no fix in 11.x line",
+                "PKSA-z3gr-8qht-p93v": "PHPUnit security advisory - no fix in 11.x line"
+            }
         }
     },
     "autoload": {
--- a/tests/e2e/Adapter_Installer/composer.json
+++ b/tests/e2e/Adapter_Installer/composer.json
@@ -11,5 +11,10 @@
         "psr-4": {
             "Adapter_Installer\\Test\\": "tests/"
         }
+    },
+    "config": {
+        "audit": {
+            "block-insecure": false
+        }
     }
 }
--- a/tests/e2e/Composer_Xdebug_Handler_Version_1/composer.json
+++ b/tests/e2e/Composer_Xdebug_Handler_Version_1/composer.json
@@ -14,5 +14,10 @@
     },
     "require": {
         "composer/xdebug-handler": "^1"
+    },
+    "config": {
+        "audit": {
+            "block-insecure": false
+        }
     }
 }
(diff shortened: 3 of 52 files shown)


composer.json
```diff
@@ -104,6 +104,12 @@
         "allow-plugins": {
             "phpstan/extension-installer": true,
             "infection/extension-installer": true
+        },
+        "audit": {
+            "ignore": {
+                "PKSA-5jz8-6tcw-pbk4": "PHPUnit argument injection - no fix in 11.x line",
+                "PKSA-z3gr-8qht-p93v": "PHPUnit security advisory - no fix in 11.x line"
+            }
```
sanmai, line 112: It sounds like we either need to upgrade to 12.5.22 or wait for 11.x to get fixed.

maks-rafalko (approved):

theofidry:
Instead of adding `audit.block-insecure = false` can't we add the relevant CVEs to the list of ignored? I am wondering if it's wise to ignore _any_ CVEs are some can still be relevant for local development or CIs.

I don't feel strongly about it though, so feel free to ignore this.

theofidry (approved):

composer.json
```diff
@@ -104,6 +104,12 @@
         "allow-plugins": {
             "phpstan/extension-installer": true,
             "infection/extension-installer": true
+        },
+        "audit": {
```
theofidry, line 108: this is no longer about the e2e tests though

sanmai: Right, but we don't just allow any vulnerable version in the main composer.json

theofidry: sure, I'd rather put it in a dedicated PR for better visibility though.

But it's a nitpick, feel free to ignore that

staabm:
Cve has been updated. You may revert this PR


See https://phpc.social/@sebastian/116424686237276260

PR #3065, opened 2026-04-18
build(mago): Align excludes with PHPStan

sanmai:
## Description

Mago was analyzing paths that PHPStan and `devTools/check_trailing_whitespaces.sh` have always excluded: vendored Psalm sources under `tests/benchmark/Tracing/sources`, test fixtures, template stubs, and the `DummyFileSystem` placeholder. These files are not meant to pass static analysis and were only silent because of baseline entries. Also, polluted the baseline.

## Changes

Mirror the existing PHPStan exclude list in mago.toml so the three tools agree on what is "analysis surface" versus fixture/vendored material.

## Related issues

#3049

### DIFF
--- a/mago.toml
+++ b/mago.toml
@@ -24,7 +24,6 @@ excludes = [
     "tests/phpunit/Command/Debug/DumpAstCommand/EchoGreeter.php",
     "tests/phpunit/Mutation/FileMutationGenerator/Fixtures",
     "tests/phpunit/PhpParser/Visitor/EnrichmentTraverse/Fixtures",
-    "tests/phpunit/WithConsecutive.php",
 ]
 
 [source.glob]
--- a/mago.toml
+++ b/mago.toml
@@ -13,9 +13,18 @@ paths = [
 ]
 includes = ["vendor"]
 excludes = [
+    "src/FileSystem/DummyFileSystem.php",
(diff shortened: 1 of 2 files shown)


mago.toml
```diff
@@ -13,9 +13,18 @@ paths = [
...
-    "tests/benchmark/Tracing/coverage"
+    "tests/benchmark/Tracing/coverage",
+    "tests/benchmark/Tracing/sources",
+    "tests/phpunit/Fixtures",
+    "tests/phpunit/Command/Debug/DumpAstCommand/Greeter.php",
+    "tests/phpunit/Command/Debug/DumpAstCommand/EchoGreeter.php",
+    "tests/phpunit/Mutation/FileMutationGenerator/Fixtures",
+    "tests/phpunit/PhpParser/Visitor/EnrichmentTraverse/Fixtures",
+    "tests/phpunit/WithConsecutive.php",
 ]
```
sanmai, line 27: https://github.com/infection/infection/blob/642c8fe4d271b645f3415e3bd5df1b4a3ce1894f/devTools/phpstan.neon#L202-L219

theofidry (approved):

theofidry:
Good catch, thank you @sanmai

PR #3068, opened 2026-04-18
chore: Drop support for PHP 8.2

### LINKED ISSUE #3064: feat: Drop support for PHP 8.2
## Problem Statement

PHPUnit 12 requires PHP 8.3.
PHPUnit 13 requires PHP 8.4.

PHP 8.2 is in security support only and its support ends the 31 Dec 2026.
PHP 8.3 is in security support only and its support ends the 31 Dec 2027.

PHPUnit 11 end of bugfix support ended the February 6, 2026.

We cannot upgrade to PHPUnit 12 at the moment because Infection still supports PHP 8.2.

## Proposed Solution

Drop PHP 8.2.

## Technical Considerations

From the [trends from Brent](https://stitcher.io/blog/php-version-stats-june-2025) from last year, PHP 8.2 was already at a very sharp decline at that time. We do not have the data at the moment for 2026, but if there is any regularities with the previous years, then we can estimate the current PHP 8.2 adoption is fairly minimal.

theofidry:
Closes #3064.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ services:
     build:
       context: devTools
       args:
-        PHP_VERSION: '8.2'
+        PHP_VERSION: '8.3'
         XDEBUG_VERSION: '3.4.1'
     volumes:
       - .:/opt/infection
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,5 @@
 services:
-  php82:
+  php83:
     build:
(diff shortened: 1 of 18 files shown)


sanmai (approved):

docker-compose.yml
maks-rafalko: php82
```suggestion
  php83:
```

maks-rafalko (approved):

PR #3073, opened 2026-04-19
refactor: Add types to constants

theofidry:
Now that #3068 has been merged, we can add types to constants. This PR achieves this using the `AddTypeToConstRector` rector rule, but it will later be enabled automatically by using the PHP 8.3 preset instead of PHP 8.2.

### DIFF
--- a/devTools/mago-baseline.toml
+++ b/devTools/mago-baseline.toml
@@ -32,18 +32,6 @@ count = 1
 
 [[issues]]
 file = "src/Command/ConfigureCommand.php"
-code = "missing-constant-type"
-message = "Class constant `NONINTERACTIVE_MODE_ERROR` is missing a type hint."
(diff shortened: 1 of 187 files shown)


maks-rafalko (approved):
the diff itself makes sense.

Not sure I understand mago failures tbh, for example this https://github.com/infection/infection/pull/3073/changes#diff-0473474fdc0fb9c13ab167218609bd57def1c710dbdc28db514f6c9e188c6be4R569

theofidry:
>Not sure I understand mago failures tbh

No me neither. But atm the baseline is too big as well, there is several issues that hint there is something not analyse correctly. I'm slowing reporting them see where that gets us. Some I'm wondering why it's not been reported before, but it gets fixed fast to can't complain

PR #3084, opened 2026-04-23
Move files, used in src to src from tests. Introduce phpat

### LINKED ISSUE #3081: bug: MockSplFileInfo is autoloadable for consumers
| Question    | Answer
| Infection version | introduced in 0.32.5 
| Test Framework version | PHPUnit 13
| CI | no
| PHP version | 8.4
| Platform    |  Ubuntu
| Github Repo | -

### Summary

`Infection\Testing\BaseMutatorTestCase` (shipped in `src/Testing/`) imports `Infection\Tests\TestingUtility\FileSystem\MockSplFileInfo`, a class that lives under the `autoload-dev` namespace (`tests/phpunit/`). Since Composer does **not** distribute `autoload-dev` classes with installed packages, the class is absent for any downstream consumer of `infection/infection`.

This causes a **fatal `Class not found` error** whenever a consumer test that extends BaseMutatorTestCase calls mutate()or assertMutatesInput(), because getMutationsFromCode() instantiates `new MockSplFileInfo(...)`.

### Root cause

The import was introduced as part of a refactor to BaseMutatorTestCase::getMutationsFromCode()(PR #2108) where EmptyTrace was changed to accept an `SplFileInfo`. The test utility MockSplFileInfo (which overrides getRealPath() to return a fake path instead of hitting the filesystem) was used — but it belongs to the `Infection\Tests\` namespace which maps to `tests/phpunit/` via `autoload-dev`:

```json
"autoload-dev": {
    "psr-4": {
        "Infection\\Tests\\": "tests/phpunit"
    }
}
```

The production `autoload` section only maps `Infection\\` → `src/`, so `Infection\Tests\TestingUtility\FileSystem\MockSplFileInfo` is never autoloadable for consumers.

### Relevant code

```php
// src/Testing/BaseMutatorTestCase.php line 54
use Infection\Tests\TestingUtility\FileSystem\MockSplFileInfo;
```

```php
// src/Testing/BaseMutatorTestCase.php lines 243-245
trace: new EmptyTrace(
    new MockSplFileInfo('/path/to/test-file.php'),
),
```

EmptyTrace::__construct()accepts a plain `\SplFileInfo`:

```php
// src/TestFramework/Tracing/Trace/EmptyTrace.php line 45
public function __construct(private SplFileInfo $sourceFileInfo)
```

### Steps to reproduce

1. `composer require --dev infection/infection:^0.32.5`
2. Create a custom mutator and a test extending BaseMutatorTestCase
3. Run the test → **Fatal error**: `Class "Infection\Tests\TestingUtility\FileSystem\MockSplFileInfo" not found`

maks-rafalko:
Fixes https://github.com/infection/infection/issues/3081

Introduces `phpat` (phpstan extension) to prevent such bugs in the future


devTools/PHPat/SrcShouldNotDependOnTestsTest.php
```diff
@@ -0,0 +1,63 @@
...
+
+declare(strict_types=1);
+
+namespace Infection\DevTools\PHPat;
+
+use PHPat\Selector\Selector;
+use PHPat\Test\Builder\Rule;
+use PHPat\Test\PHPat;
+
+final class SrcShouldNotDependOnTestsTest
```
theofidry, line 42: wouldn't it make more sense to have it as part of `tests`?

We have `tests/Architecture`

devTools/PHPat/SrcShouldNotDependOnTestsTest.php
```diff
@@ -0,0 +1,63 @@
...
+                Selector::inNamespace('Infection\\Tests'),
+                Selector::inNamespace('Infection\\Benchmark'),
+                Selector::inNamespace('Infection\\DevTools'),
+            )
+            ->shouldNot()
+            ->dependOn()
+            ->classes(
+                Selector::AnyOf(
+                    Selector::inNamespace('Infection\\Tests'),
+                    Selector::inNamespace('Infection\\Benchmark'),
```
theofidry, line 58: To be consistent:

```suggestion
                    Selector::inNamespace('Infection\\Benchmark'),
                    Selector::inNamespace('Infection\\DevTools'),
```

devTools/PHPat/SrcShouldNotDependOnTestsTest.php
```diff
@@ -0,0 +1,63 @@
...
+                ),
+            )
+            ->because('Production code under src/ must not depend on tests/ or benchmarks.');
```
theofidry, line 61:
```suggestion
            ->because('Production code should not depend on development only code.');
```

theofidry (commented):
Left some suggestions

devTools/PHPat/SrcShouldNotDependOnTestsTest.php
maks-rafalko: I'm not sure, because it's really not a test, but phpstan rule. It's only written in this, hm interesting, "test-like" syntax.

We have another PHPStan rule on the same level: 

https://github.com/infection/infection/blob/b89eaf35ded4a5b914923855d80e221969526707/devTools/PHPStan/Rules/InfectionContainerRule.php#L51

theofidry: At the end of the day it is still code that can, like the rest, benefit from CS fixing & autoreview.

It is IMO akin to the autoreview tests that we have: some are the actual checks, some are pieces to enable those checks. Having it here means more less discoverable code that is entirely exempt from our usual checks.

I appreciate it's only two classes for now, and whilst I personally do not plan to work on additional PHPStan rules myself, I have a few PHPAt ones I'd like to add.

I'm happy to take care of that (unless you're against it)

maks-rafalko: sounds good

PR #3086, opened 2026-04-28
perf(ast): Avoid traversing branches which do not contain eligible nodes

theofidry:
_This PR is under heavy development, it is not ready for review in any way._

Notes for myself:

- Extract the benchmark, ensure it is comprehensive enough to capture the AST processing phase without prior overhead and without the overhead of generating mutations. For completeness the parsed files should be partially covered.
- Verify that the PR works as expected (it does at first glance but requires a more in-depth review).
- Measure the performance gains.

## Description

As of #3026, we now have a complete separation between the first traverse which is to enrich the AST and the second traverse which is for generating mutations.

As a result of the first mutation, we know which nodes are eligible, hence we can also deduce if a node contains any eligible child. If it does not, we could skip this branch entirely during the 2nd traverse.

## Changes

- Adds/Fixes/Updates ...

## Checklist

- [ ] Tests added/updated

## Related issues

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,6 +142,7 @@ validate:
 profile: 	 	## Runs Blackfire
 profile:
 	$(MAKE) profile_mutation_generator
+	$(MAKE) profile_ast_processing
 	$(MAKE) profile_parse_git_diff
 	$(MAKE) profile_tracing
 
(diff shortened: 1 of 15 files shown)


theofidry:
I'll be closing this one as it has prove that whilst possible, the performance gain did not materialise. Maybe the idea is fine and it is my implementation that is left to be desired, but in its current state I saw a ~+10% slowness, memory & CPU usage instead of a performance improvement.

Despite this, I think https://github.com/infection/infection/pull/3088 is still relevant, as it is this kind of benchmark that will reliably demonstrate if such changes are beneficial or not.

theofidry:
@sanmai maybe if you wish to have another look at it, I'm short on ideas implementation wise so I decided to give up on it (see my previous comment).

PR #3088, opened 2026-04-28
test(ast): Add a benchmark for AST processing

theofidry:
## Description

This PR adds a benchmark dedicated to the AST processing, i.e. generating the rich AST and traversing the AST the 2nd time for mutation generation (but without generating any mutation). This second part is important to include to be able to measure any potential performance optimisation done during the second traverse, without including the noise of generating the mutations themselves.

As for the other benchmarks, this PR introduces two commands:

- `make benchmark_ast_processing`
- `make profile_ast_processing`

As well as a `tests/benchmark/AstProcessing/README.md` for more detailed instructions.

## Reviewer Notes

There is a noticeable overlap with the existing `MutationGeneration` benchmark. However, while evaluating https://github.com/infection/infection/pull/3086, I encountered two issues with the latter:

- Generating the mutations themselves is an overhead in which we are not interested for this kind of change (optimising the 2nd traverse).
- It generates all possible mutations, regardless of the coverage data. This is problematic as a significant supposed gain of #3086 is to be able to skip ineligible branches, and one source of ineligible branches is the lack of tests.

For this reason I believe a different specific benchmark is more appropriate.

Also note that it re-uses the `benchmark-source` from the `Tracing` benchmark. It may be better to move the latter out of `Tracing` as it is being re-used across multiple benchmarks now. This will be done in a separate PR as likely noisy.

## Related issues

- #3086


tests/benchmark/AstProcessing/create-main.php
```diff
@@ -0,0 +1,260 @@
...
+                ->traverse($initialStatements);
+
+            $remainingNodeCount = $maxCount - $count;
+            Assert::positiveInteger($remainingNodeCount);
+
+            $nodeCountingVisitor = new NodeCountingVisitor($remainingNodeCount);
+
+            $traverserFactory
+                ->createMutationTraverser($nodeCountingVisitor)
+                ->traverse($initialStatements);
```
maks-rafalko, line 249: not sure I understand what we are benchmarking here.

```php
    public function createMutationTraverser(NodeVisitor $mutationVisitor): NodeTraverserInterface
    {
        return new NodeTraverser(
            new NodeVisitor\CloningVisitor(),
            $mutationVisitor,
        );
    }
```

so we basically benchmark `CloningVisitor` and `NodeCountingVisitor`. What do I miss?

maks-rafalko (commented):
> This second part is important to include to be able to measure any potential performance optimisation done during the second traverse

do you already have an example of such optimization? what could it be?

theofidry:
@maks-rafalko 

>do you already have an example of such optimization? what could it be?

Yes, https://github.com/infection/infection/pull/3086. At least that is one idea that could be explored, I'll details the findings there. Apologies I linked the telemetry PoC instead of that PR.

tests/benchmark/AstProcessing/create-main.php
theofidry: >so we basically benchmark CloningVisitor and NodeCountingVisitor. What do I miss?

Correct. The reason being:

>This second part is important to include to be able to measure any potential performance optimisation done during the second traverse, without including the noise of generating the mutations themselves.

`NodeCountingVisitor` could very well be a `NullVisitor`, but I thought it would be good to have the count of mutations entered as a naive metric to know if it is working as expected or see a quick difference.

You are right that it does not do much, but if you exclude it and like in #3086 implement something to no longer fully traverse the AST, then the benchmark would not be able to capture that at all.

maks-rafalko (commented):
so, if I got it right

1. this new benchmark would measure the changes like https://github.com/infection/infection/pull/3086
2. currently it doesn't measure anything (which is why I was confused initially)

if p. 1 is true, then makes sense

theofidry:
>this new benchmark would measure the changes like https://github.com/infection/infection/pull/3086

Yes.

>currently it doesn't measure anything (which is why I was confused initially)

I don't understand what you mean by that. The benchmark is functional, regardless of #3086.

I don't have any change planned queued up yet, but for any change touching the AST, rather than the mutation generation itself, it is a lot more accurate.

maks-rafalko:
> I don't understand what you mean by that. The benchmark is functional, regardless of #3086.

my understanding: currently, this benchmark measures nothing actually, except `CloningVisitor` (php-parser's visitor) and `NodeCountingVisitor` (created in benchmark). 

With #3086 it measures one more visitor, which does make sense for me. Without #3086 - it make sense to have this benchmark only if/when we add something to the 2nd traverser.

If I miss something, could you please explain in more details?

The code under benchmark is:

https://github.com/infection/infection/blob/57e9d503cc1727ba9ce11001dbe6fe9d5404171a/src/PhpParser/NodeTraverserFactory.php#L110-L117

theofidry:
>my understanding: currently, this benchmark measures nothing actually, except CloningVisitor (php-parser's visitor) and NodeCountingVisitor (created in benchmark).

The first traverse is included in the benchmark.

As a concrete example, https://github.com/infection/infection/pull/3086 added two visitors. The first visitor was in the first traverse (the rich AST pass) to add a property to signal a node did not contain any eligible node, and the second traverse (the mutation generation pass) where it skipped the traverse of the node if it did not contain any eligible node.

It is the two traverses that need to be benchmarked, otherwise you would get an incomplete picture. Here with #3086 it showed that, at least with the source files used, adding those two visitors with this small additional computation and stopping the traverse of some nodes was more expensive than traversing all nodes.

PR #3089, opened 2026-04-28
Disable code review status comments

maks-rafalko:
See https://github.com/infection/infection/pull/3086#issuecomment-4334153437

See https://docs.coderabbit.ai/reference/configuration#param-review-status

from claude:

> CodeRabbit may still post a minimal walkthrough comment on the PR. The only fully-clean way to eliminate it is via the dashboard (Repositories → Reviews → "Disable Walkthrough"), which isn't exposed as a YAML key.

^ we will probably need to test this later as well. Posting 2 comments after PR is craeted (see below) is ... too much.

Disabled poem in Walkthrough 🤦  (`poem: false`)

image

Also, disabled this "Summary by CodeRabbit" 👇 (`high_level_summary: false`)

## Summary by CodeRabbit

* **Chores**
  * Updated configuration settings to adjust review behavior.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,2 @@
+reviews:
+  review_status: false
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,2 +1,3 @@
 reviews:
   review_status: false
+  high_level_summary: false
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,3 +1,4 @@
 reviews:
   review_status: false
   high_level_summary: false
+  poem: false


theofidry:
>Disabled poem in Walkthrough 🤦 (poem: false)

Strange I thought I did that in the UI already? I didn't see it for the newer settings, but I'm fine with having it programmatically on the repository instead of on their SaaS.

>^ we will probably need to test this later as well. Posting 2 comments after PR is craeted (see below) is ... too much.

I agree it's annoying, although I do like the result. It's more noisy than I like but it has had useful tips. I think it would be nicer if it was not doing a review automatically and rather be on demand.

.coderabbit.yaml
```diff
@@ -0,0 +1,4 @@
+reviews:
+  review_status: false
+  high_level_summary: false
```
theofidry, line 3: the high level summary is fine I think, if you disable it it will only disable a _section_ of the comment you referred, and another setting is controlling if that section is in the PR description (which I really dislike because I don't want AI to edit my PR without my demand) or in that first comment.

theofidry:
@maks-rafalko I disabled it, we can still run it by requesting it manually. See #3094

.coderabbit.yaml
maks-rafalko: > if you disable it it will only disable a section of the comment you referred

I know, still I don't see many of the sections useful, like `poem`. If you think high_level_summary is ok for now, I'm fine, let's see

PR #3095, opened 2026-04-30
feat(reporter): Introduce the Reporter framework

theofidry:
## Summary

Introduces the first reporting framework primitives described in the [From Loggers to Reporters](https://github.com/infection/infection/issues/2866).

- `DataProducer` for producing report content independently of its destination
- `ReportWriter` for writing report content to a destination
- `ComposableReporter` to connect a producer to a writer for the common case
- `AggregateReporter` and `AggregateReporterFactory` to compose several configured reporters
- `FileWriter`, `StreamWriter`, and `NullReporter` as reusable baseline implementations

The aim is to simplify adding new reports without duplicating content-generation logic across output destinations, whilst allowing specialised reports to implement `Reporter` directly where the producer/writer split is not appropriate.

**This PR voluntarily does not change any of the existing reporters.**

The goal is, at the moment, to be able to add new reporters with this new API. The existing ones can be adapted and ported incrementally over time.

But as a result, it may be a bit hard to visualise those changes, so what does it change?

- No need for a ever growing complex `FileReporterFactory`, instead each entry could use their own factory. For example:

```php
final readonly class GitHubAnnotationsReporterFactory implements ReporterFactory
{
    public function __construct(
        private ResultsCollector $resultsCollector,
        private ?string $loggerProjectRootDirectory,
        private bool $decoratedOutput,
    ) {
    }

    public function create(Logs $logConfig): Reporter
    {
        if (!$logConfig->getUseGitHubAnnotationsLogger()) {
            return new NullReporter();
        }

        return new ComposableReporter(
            new GitHubAnnotationsProducer(...),
            writer: StreamWriter::createForStream(...),
        );
    }
}
```

Then `GitHubAnnotationsReporterFactory` is just part of the chain of Reporter factories like another service. Note that `FileReporterFactory` does allow quite a few simplifications, so not necessarily interesting for this peculiar example, but it gives a sense of how some reporters can be composed with re-usable pieces.

## Details

- Added `Infection\Report\Framework` contracts for report data production, writing, and reporter creation.
- Added file and stream writer implementations supporting string content and iterable line output.
- Added aggregate and null reporter implementations for composition and empty reporter cases.
- Added unit tests covering the new reporters, factory, and writer behaviour.

## Related issues


maks-rafalko (approved):
makes sense, as we discussed. Thank you

Issue #2437, opened 2025-10-10
PHPStorm / Intellij IDEA plugin Roadmap


xepozz:
Blocked by https://github.com/infection/infection/issues/2446.

<hr/>

PHPStorm / Intellij IDEA plugin is available: 
Marketplace https://plugins.jetbrains.com/plugin/28650-infection
Sources: https://github.com/j-plugins/infection-plugin

There's simple integration which is moving forward, copying PHPUnit behavior.

I'd like to listen any ideas to be implement.

What's currently in my mind:

## Test runner

Add a marker to run tests with infection, like PHPUnit or Codeception do

Example:
<img width="323" height="218" alt="Image" src="https://github.com/user-attachments/assets/5fc1643a-bca2-464a-bfed-0b52724ef74b" />

## Code Coverage

Infection could mark code coverage panel with its results.
I didn't dive enough there, but looks helpful to discover mutated tests in a "files view".

I mean this panel:

<img width="503" height="376" alt="Image" src="https://github.com/user-attachments/assets/d4839c5f-eb98-4b26-8e76-2016dbbc9e44" />

## Errors highlighting

I've created a code highlighter which uses results of the run.
<img width="590" height="190" alt="Image" src="https://github.com/user-attachments/assets/4f3a85ff-0f1c-4d04-add3-8cfd3775c2ac" />

But seems it will be very annoying because many project don't have full CC and there always live mutants.
There should an easy way to show/hide such errors, compared to regular "Disable inspection", because users will disable it in 99% and won't enable it back again.

What else?
- `@infection-ignore` PHPdoc tags?
- `infection.json` schema provider?
- Any test method markers? `function testMe(): void` `(5 mutants)`
- Any code completions? Any Type providers?
- Anything else?

xepozz:
@maks-rafalko is there svg flat icon? 
I've found only png which isn't acceptable for IDEA

maks-rafalko:
Thanks to @staabm we do have SVG icon.

![Image](https://github.com/user-attachments/assets/d895f6a0-bdb6-41bf-a6fc-eaf5c4e27854)

I will go through all the questions later, thank you.

xepozz:
![Image](https://github.com/user-attachments/assets/94d093d4-d660-47ef-83f0-53630d2f0e0e)
![Image](https://github.com/user-attachments/assets/bff72c3c-0e1a-4d80-8f72-52aae8af212c)

[...]

PR #3105, opened 2026-05-04
feat(teamcity): Add the source file path to the test finished message

theofidry:
_PR ready but waiting on feedback, see https://github.com/infection/infection/issues/2911#issuecomment-4369585109_

## Description

Add the source file path to the TeamCity message so that the plugin can work out a nicer diff when checking a mutation.

## Related issues

- #2437


theofidry:
⚠️ Putting in Draft to not merge it, waiting on feedback.

PR #3107, opened 2026-05-05
feat: Register a new testing framework Testo

### LINKED ISSUE #3053: feat: Testo support
## Proposed Solution

[Testo](https://github.com/php-testo/testo) is a new testing framework in PHP ecosystem, it would be nice to collaborate and support testo as well as phpunit.

## Technical Considerations

I'm not sure I'm ready to talk about implementations, it's not my surface.

roxblnfk:
## Description

Adds support for [Testo](https://github.com/php-testo/testo) as a recognized test framework in Infection. The bridge lives in a separate package, [`testo/bridge-infection`](https://github.com/php-testo/bridge-infection), following the shape of `infection/codeception-adapter`; this PR adds the minimal core changes so `testFramework: "testo"` validates against the schema and `AdapterInstaller` can resolve the bridge package.

Testo's config is a PHP file (`testo.php`), so the config locator is also extended to recognize `.php` as a supported extension.

With these changes, Infection [runs end-to-end against the Testo repository](https://github.com/php-testo/testo/actions/workflows/infection.yml), where the integration is already in use.

## Changes
                                                                                                                                                                                                                                                                                                            
- Registered Testo as a new test framework type (`TestFrameworkTypes::TESTO`) and added it to the default types list.
- Mapped Testo to its official adapter package `testo/bridge-infection` in `AdapterInstaller::OFFICIAL_ADAPTERS_MAP`.
- Added `"testo"` as an allowed value for `testFramework` in `resources/schema.json`.
- Added `"php"` to the supported config file extensions in `TestFrameworkConfigLocator`, enabling PHP-based framework configs.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (provide link to PR: https://github.com/infection/site/pull/297)
- [x] CHANGELOG.md updated (if deprecations/breaking changes)

## Reviewer Notes

Adding `php` to the locator's supported extensions applies to all frameworks, not just Testo. `xml`/`yml` variants are still tried first, so existing setups are unaffected; the change only matters when no `xml`/`yml` config is present and a `<framework>.php` file exists in the config dir. Discussed in #3053.

## Related issues

Fixes #3053

### DIFF
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -641,7 +641,8 @@
             "anyOf": [
                 "phpunit",
                 "phpspec",
-                "codeception"
+                "codeception",
+                "testo"
             ]
         },
         "staticAnalysisTool": {
--- a/src/TestFramework/AdapterInstaller.php
+++ b/src/TestFramework/AdapterInstaller.php
@@ -48,6 +48,7 @@ final readonly class AdapterInstaller
     public const array OFFICIAL_ADAPTERS_MAP = [
         TestFrameworkTypes::CODECEPTION => 'infection/codeception-adapter',
         TestFrameworkTypes::PHPSPEC => 'infection/phpspec-adapter',
+        TestFrameworkTypes::TESTO => 'testo/bridge-infection',
     ];
 
     // 2 minutes
--- a/src/TestFramework/Config/TestFrameworkConfigLocator.php
+++ b/src/TestFramework/Config/TestFrameworkConfigLocator.php
@@ -52,6 +52,7 @@ final readonly class TestFrameworkConfigLocator implements TestFrameworkConfigLo
         'yml.dist',
         'dist.xml',
         'dist.yml',
+        'php',
     ];
 
     public function __construct(
--- a/src/TestFramework/TestFrameworkTypes.php
+++ b/src/TestFramework/TestFrameworkTypes.php
@@ -51,6 +51,8 @@ final class TestFrameworkTypes
 
     public const string CODECEPTION = 'codeception';
 
+    public const string TESTO = 'testo';
+
     /**
      * @var string[]
      */
@@ -58,6 +60,7 @@ final class TestFrameworkTypes
         self::PHPUNIT,
         self::PHPSPEC,
         self::CODECEPTION,
+        self::TESTO,
     ];
 
     /**
--- a/Makefile
+++ b/Makefile
@@ -306,13 +306,13 @@ $(MAGO): vendor
 	touch -c $@
 
 $(INFECTION): vendor $(shell find bin/ src/ -type f) $(BOX) box.json.dist .git/HEAD
-	composer require infection/codeception-adapter infection/phpspec-adapter
+	composer require infection/codeception-adapter infection/phpspec-adapter testo/bridge-infection
 	# Workaround for https://github.com/box-project/box/issues/580
 	composer install --no-dev
 	$(BOX) --version
 	$(BOX) validate
 	$(BOX) compile
-	composer remove infection/codeception-adapter infection/phpspec-adapter
+	composer remove infection/codeception-adapter infection/phpspec-adapter testo/bridge-infection
 	composer install
 	touch -c $@
 
--- a/src/Command/RunCommand.php
+++ b/src/Command/RunCommand.php
@@ -63,6 +63,7 @@ use Infection\Metrics\MinMsiCheckFailed;
 use Infection\Process\Runner\InitialTestsFailed;
 use Infection\Source\Exception\NoSourceFound;
 use Infection\StaticAnalysis\StaticAnalysisToolTypes;
+use Infection\TestFramework\AdapterInstaller;
 use Infection\TestFramework\TestFrameworkTypes;
 use InvalidArgumentException;
 use const PHP_SAPI;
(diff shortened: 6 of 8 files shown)


maks-rafalko (approved):
Not sure if it's ready for review yet or you just forgot to move it from draft, but looks good with the current state.

Regarding outstanding TODO items (checkboxes), could you please create a Docs PR?

This is a good place to start.

https://github.com/infection/site/blob/a45fef68e47ecc40247d36c3a1264d234b3fdc3e/src/guide/supported-test-frameworks.md

I would recommend searching "codeception" and add Testo in all the relevant places

Image

resources/schema.json
```diff
@@ -641,7 +641,8 @@
             "anyOf": [
                 "phpunit",
                 "phpspec",
-                "codeception"
+                "codeception",
+                "testo"
```
theofidry, line 645: not to self: we probably need to allow an arbitrary string here to allow one to register their own test framework without requesting changes within infection

src/TestFramework/Config/TestFrameworkConfigLocator.php
```diff
@@ -52,6 +52,7 @@
         'yml.dist',
         'dist.xml',
         'dist.yml',
+        'php',
```
theofidry, line 55: note to self: check if the necessity of that change is well captured in #2863

src/TestFramework/AdapterInstaller.php
```diff
@@ -48,6 +48,7 @@
     public const array OFFICIAL_ADAPTERS_MAP = [
         TestFrameworkTypes::CODECEPTION => 'infection/codeception-adapter',
         TestFrameworkTypes::PHPSPEC => 'infection/phpspec-adapter',
+        TestFrameworkTypes::TESTO => 'testo/bridge-infection',
```
theofidry, line 51: just to be clear: this does not hurt to have, but it is otherwise unnecessary right? As in: the extension installer would correctly detect that adapter based on the composer package type

src/TestFramework/TestFrameworkTypes.php
```diff
@@ -51,13 +51,16 @@ final class TestFrameworkTypes
...
+    public const string TESTO = 'testo';
+
     /**
      * @var string[]
      */
     private static array $defaultTypes = [
         self::PHPUNIT,
         self::PHPSPEC,
         self::CODECEPTION,
+        self::TESTO,
```
theofidry, line 63: same question for here

src/TestFramework/AdapterInstaller.php
maks-rafalko: we have a feature when you run `bin/infection --test-framework=testo` and `testo` adapter is not yet installed, it would automatically install it in place and continue. I'm pretty sure this is for this purpose.

At least this is how I developed it for Codeception/PhpSpec back in that time

https://github.com/infection/infection/blob/cc5f5bcd7a85ffd109d992c7b813dd167a669bd9/src/TestFramework/AdapterInstallationDecider.php#L73-L74

and then

https://github.com/infection/infection/blob/b4fa5e25830a6e14816e828a402fde35e29bf0f3/src/Command/RunCommand.php#L511

This is done to improve Developer Experience (DX) so you don't have to install it manually, which is cool.

We can probably split this behavior and allow installing any adapter, as well as supporting auto-installation for "known" adapters, but this is out of the scope of this PR - what is done here is correct for now IMO

src/TestFramework/TestFrameworkTypes.php
maks-rafalko: same answer: https://github.com/infection/infection/pull/3107#discussion_r3189064060

src/TestFramework/AdapterInstaller.php
roxblnfk: I'm not sure these changes are sufficient for installing the package. But they also aren't needed to run infection when the bridge is installed.

image

maks-rafalko: Then something is broken. I can take a look only tonight. The code that should allow it is here:

https://github.com/infection/infection/blob/17cf4fa449b79413e9608ce201cecf266d9a0d69/src/TestFramework/Factory.php#L97-L122

Extension installer populates `Infection\ExtensionInstaller\GeneratedExtensionsConfig::EXTENSIONS` const right in the vendor folder, so this is where I would start debugging.

theofidry: In fact I vaguely remember it not working correctly with phpspec/codeception too. I didn't mind any mind at that time, but it's likely not a testo only issue

roxblnfk: Found the reason. Also found some hardcoded thing
image

maks-rafalko: Nice! Sorry you had to debug it, but at the same time it's cool to know now `testo` can be automatically installed! 👍

maks-rafalko (changes requested):
One thing that I noticed after reviewing docs: it says PHAR automatically bundles all the adapters, while it's not - `testo` is now missed.

Could you please update this place in a `Makefile` to include `testo` into PHAR?

https://github.com/infection/infection/blob/6f8e1847affa513411ed48af8b18246f5627316f/Makefile#L309

Notes: when we release Infection, we have a job that automatically compiles PHAR and uploads it to a release page, using this exact make target. 

https://github.com/infection/infection/blob/e3594a996ae7f073e1f1e1f1944475ea567803c9/.github/workflows/release.yaml#L43

maks-rafalko (approved):
I didn't have a chance to test it myself, but will certainly do before releasing. 

Code-wise, looks good. Thank you!

maks-rafalko:
Thank you @roxblnfk

Issue #3090, opened 2026-04-29
RFC: Telemetry Traces and Metrics


theofidry:
## Summary

This RFC proposes adding an optional [OpenTelemetry]-based telemetry system to Infection. It would collect structured execution traces and resource metrics during a mutation testing run, expose the results as a reporter artefact or via an [OpenTelemetry] exporter, and provide sufficient post-run querying and visualisation to understand where time and resources are spent.

The initial target audience is the Infection team for dogfooding, and users who wish to track Infection-specific KPIs or investigate expensive mutations.

## Motivation

Infection currently exposes the final mutation testing result via reports. Those reports give insights into the outcome, but very little into the execution itself: what operations were performed, in what context, and how expensive they were.

### Use case for large code-bases

Users running Infection on large codebases often cannot afford full runs. They need to understand which source files, mutators, mutations, heuristics, or test executions consume the most time so that they can tune their configuration, adjust their code, prioritise work, or track KPIs over time.

In this context, reports provide information only about a partial run. A more comprehensive overview would require collecting and aggregating those partial reports to extract and track metrics. No existing report provides this without significant investment in custom code and infrastructure. Supporting [OpenTelemetry] would allow users to leverage these features with any compatible system, such as [Datadog], [Sentry], or [NewRelic].

### An Observability Standard

[OpenTelemetry], also known as OTel, is a vendor-neutral open-source observability framework for instrumenting, generating, collecting, and exporting telemetry data such as traces, metrics, and logs.

The range of metrics and data of interest varies by use case, but implementing this standard gives considerable flexibility.

### Better Visualisation and Event Model

Infection relies on events to implement hooks for logging or recording state for reporting purposes. There are known gaps in the event system (see [#2327][issue-2327] and [#1862][issue-1862]). Such gaps can be addressed by adding events, but doing so is difficult without a clear visualisation of the current execution, an understanding of the distinct units of execution, and agreed-upon terminology. Some progress was made in [#2869][issue-2869], but gaps remain. A trace visualisation tool would address this directly.

For duration-reporting issues such as [#2900][issue-2900] specifically, span-level tracing makes it possible to identify which component of a reported time is incorrect: whether it corresponds to active execution, queue wait, or reporting overhead.

### More Debugging Capabilities

[...]

PR #3111, opened 2026-05-07
feat(telemetry): Add support for INFECTION_TELEMETRY

theofidry:
## Description

Controlling the `OTEL_*` environment variables is quite powerful and flexible. However, users may not want telemetry to be enabled for any application just because an `OTEL_` environment variable is set. Additionally, telemetry was never part of infection before and we do not have a SaaS service either, so **telemetry should be off by default and enabled only if requested**. An `OTEL_` is not explicit enough.

This PR introduces `INFECTION_TELEMETRY=true`. Without it, not telemetry will is activated. Note that this sort of environment variable is not defined by the OpenTelemetry specification.

A few example of applications working this way:

- Claude Code (`CLAUDE_CODE_ENABLE_TELEMETRY=1`)
- next.js (`NEXT_TELEMETRY_DISABLED=1`)
- GitHub CLI (`GH_TELEMETRY=false`)
- Homebrew (`HOMEBREW_NO_ANALYTICS=1`)

## Related issues

- https://github.com/infection/infection/issues/3090


theofidry:
@coderabbitai full review

PR #3113, opened 2026-05-08
fix(git): Fix the project directory resolution when git is not available

theofidry:
## Description

https://github.com/infection/infection/pull/3009 introduced a BC break change:

>If the user was using infection without git and no environment variables, it will not longer work.

I however did not realise we expect infection to work in those conditions (see `tests/e2e/Without_Git`) – at least to a degree. So this PR addresses this BC (which at the time of writing is not released yet).

This PR swallows the exception thrown by the Git project provider (still logs the exception) and falls back to the current working directory.

## Changes

- Allow `Git::getProjectDirectory()` to throw a `NoGitProjectFound` exception.
- Make the `Git` implementation throw `NoGitProjectFound` upon any process exception thrown.
- Add a new `ProjectDirectoryProvider` to fallback to the current working directory.

## Reviewer Notes

I am not clear why this e2e test did not fail in #3009. I also noticed it failing in some other builds (e.g. https://github.com/infection/infection/actions/runs/25544899235/job/74978721841?pr=3112 or https://github.com/infection/infection/actions/runs/25455767137/job/74684687591?pr=3091) but ran fine in #3009 and still does (at the time of writing) on the master branch too: https://github.com/infection/infection/actions/runs/25308273563/job/74189031708

## Related issues

- #3009

### DIFF
--- a/src/Git/CommandLineGit.php
+++ b/src/Git/CommandLineGit.php
@@ -52,6 +52,7 @@ use function Safe\preg_split;
 use function sprintf;
 use function str_starts_with;
 use Symfony\Component\Process\Exception\ExceptionInterface as ProcessException;
+use Symfony\Component\Process\Exception\ProcessFailedException;
 use Symfony\Component\Process\Exception\ProcessTimedOutException;
 use Webmozart\Assert\Assert;
 
@@ -150,22 +151,20 @@ final readonly class CommandLineGit implements Git
         return $base;
     }
 
-    /**
-     * {@inheritdoc}
-     *
-     * @throws ProcessTimedOutException
-     * @throws ProcessException
-     */
     public function getProjectDirectory(): string
     {
         // An error here should really not happen, so we are fine to let it
         // bubble up instead of throwing a dedicated exception or providing a
         // fallback.
-        $directory = $this->shellCommandLineExecutor->execute([
-            'git',
-            'rev-parse',
-            '--show-toplevel',
-        ]);
+        try {
+            $directory = $this->shellCommandLineExecutor->execute([
+                'git',
+                'rev-parse',
+                '--show-toplevel',
+            ]);
+        } catch (ProcessTimedOutException|ProcessFailedException|ProcessException $exception) {
+            throw NoGitProjectFound::create($exception);
+        }
 
         Assert::stringNotEmpty($directory);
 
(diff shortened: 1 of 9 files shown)


src/Git/CommandLineGit.php
```diff
@@ -150,22 +151,20 @@ public function getBaseReference(string $base): string
...
-            'rev-parse',
-            '--show-toplevel',
-        ]);
+        try {
+            $directory = $this->shellCommandLineExecutor->execute([
+                'git',
+                'rev-parse',
+                '--show-toplevel',
+            ]);
+        } catch (ProcessTimedOutException|ProcessFailedException|ProcessException $exception) {
```
sanmai: not sure `ProcessTimedOutException` really means git isn't there 🤔

src/Configuration/ProjectDirectoryProvider/GitProjectDirectoryProvider.php
```diff
@@ -44,11 +46,21 @@
...
+        try {
+            return $this->git->getProjectDirectory();
+        } catch (NoGitProjectFound $exception) {
+            $this->logger->info(
+                'Could not determine the project directory from Git.',
+                ['exception' => $exception],
+            );
+
+            return null;
+        }
```
sanmai, line 64: So the gist is that we swallow this exception to let Infection proceed along

sanmai (approved):

src/Git/CommandLineGit.php
theofidry: Indeed.

I tried to have the message generic enough, i.e. that we could not get the result we needed with the command, regardless of the reason. The exact reason should be conveyed by the previous exception (being forwarded).

I'm also not clear of why `ProcessTimedOutException` and `ProcessFailedException` are not `ProcessException` but maybe I looked at the wrong thing at the time of implementation. I would double check

src/Configuration/ProjectDirectoryProvider/GitProjectDirectoryProvider.php
theofidry: Indeed, apologies that should have been in the description, I updated it.

src/Git/CommandLineGit.php
theofidry: Ok it does so I can simplify it

maks-rafalko:
how did you notice it if tests didn't fail? 🤔

theofidry:
@maks-rafalko the e2e test was failing locally for me, so I could easily reproduce the issue and verify the fix. What I couldn't figure out is why it is sometimes passing in the CI

PR #3129, opened 2026-05-12
feat(logger): Make dots-per-row configurable in the dot progress formatter

### LINKED ISSUE #3122: feat: configurable dots-per-row in ConsoleDotLogger
## Problem Statement

Running Infection on a mid-size codebase (around 2k mutants) on a wide terminal, I noticed the progress output stacks up far more rows than it needs to. `ConsoleDotLogger` ships with `DOTS_PER_ROW = 50` as a private const, so even on a 200-column terminal I'm scrolling through 40 rows when I could fit it in 10.

It's a small papercut, but PHPUnit already solved this with `columns="max"` in its config; it would be nice to have parity.

## Proposed Solution

A new optional key in `infection.json{,5}`, defaulting to 50 so nothing changes for anyone who doesn't touch it:

```json
{
    "dotsPerRow": 50
}
```

I'd also let it take `"max"` to auto-fit the terminal width, capped at the total mutation count so a short run doesn't leave trailing whitespace. `symfony/console` is already a direct dep, so `Terminal::getWidth()` is on hand.

A `--dots-per-row` CLI flag matching the config key would round it out, same shape as `--threads` (which already accepts `max` from both the config and the flag).

## Technical Considerations

The constant lives at `ConsoleDotLogger.php:54` and is read in two places: the modulo at line 103 and the trailing-pad at line 108. Both already go through `self::DOTS_PER_ROW`, so swapping it for a constructor-injected `int` is a contained change.

Schema gets a `dotsPerRow` entry that's `oneOf` an integer >= 1 or the literal `"max"`. `ConsoleDotLoggerTest` will need its fixture strings parameterised on width (the current ones bake in 50 dots per row), but I don't think anything else shifts.

Happy to put up a PR if there's interest.

## Related Issues

#222 and #2866 both touch the logger area but neither covers this specifically.

paulbalandan:
## Description

Adds a new `dotsPerRow` option to the configuration schema and a matching `--dots-per-row` / `-r` CLI flag for the dot progress formatter. The default stays at 50, so existing users see no change. Set an explicit positive integer for a fixed width, or `"max"` to fit the terminal width.

`infection.json5`:

```json5
{
    "source": { "directories": ["src"] },
    "dotsPerRow": 80
}
```

CLI:

```bash
vendor/bin/infection --dots-per-row=max
vendor/bin/infection -r 20
```

## Reviewer Notes

The `int|"max"` setting is carried unresolved through `Configuration::$dotsPerRow` and resolved inside `ConsoleDotLogger::startAnalysis()` rather than at config build time. This is the only place where both the real terminal width and the mutation count are known, which the `"max"` semantics need; the resolved width subtracts the `   (n / total)` suffix so the counter still fits, and is capped at the total mutation count to avoid trailing whitespace on short runs.

The new `int|"max"` argument lives on `MutationAnalysisLoggerFactory::create()` rather than the constructor; the DI container's reflection-based autowiring rejects union types on factory constructors.

`Symfony\Component\Console\Terminal` is injected as an optional constructor argument on `ConsoleDotLogger` so tests can mock the terminal width.

## Related issues

Closes #3122.

### DIFF
--- a/resources/schema.json
+++ b/resources/schema.json
@@ -26,6 +26,19 @@
                 }
             ]
         },
+        "dotsPerRow": {
+            "description": "Number of dots per row in the dot progress formatter. Use \"max\" to fit the terminal width.",
+            "oneOf": [
+                {
+                    "type": "integer",
+                    "minimum": 1
+                },
+                {
+                    "type": "string",
+                    "enum": ["max"]
+                }
+            ]
+        },
         "source": {
             "type": "object",
             "required": [
--- a/src/Command/RunCommand.php
+++ b/src/Command/RunCommand.php
@@ -82,6 +82,8 @@ final class RunCommand extends BaseCommand
 {
     public const string OPTION_THREADS = 'threads';
 
+    public const string OPTION_DOTS_PER_ROW = 'dots-per-row';
+
     public const string OPTION_LOGGER_GITHUB = 'logger-github';
 
     public const string OPTION_SHOW_MUTATIONS = 'show-mutations';
@@ -175,6 +177,13 @@ final class RunCommand extends BaseCommand
                 InputOption::VALUE_REQUIRED,
                 'Number of threads to use by the runner when executing the mutations. Use "max" to auto calculate it.',
                 Container::DEFAULT_THREAD_COUNT,
+            )
+            ->addOption(
+                self::OPTION_DOTS_PER_ROW,
+                'r',
+                InputOption::VALUE_REQUIRED,
+                'Number of dots per row in the dot progress formatter. Use "max" to fit the terminal width.',
+                Container::DEFAULT_DOTS_PER_ROW,
             )
             ->addOption(
                 self::OPTION_WITH_UNCOVERED,
@@ -474,6 +483,7 @@ final class RunCommand extends BaseCommand
             staticAnalysisToolOptions: $commandHelper->getStringOption(self::OPTION_STATIC_ANALYSIS_TOOL_OPTIONS, Container::DEFAULT_STATIC_ANALYSIS_TOOL_OPTIONS),
             sourceFilter: SourceFilterOptions::get($io),
             threadCount: $commandHelper->getThreadCount(),
+            dotsPerRow: $commandHelper->getDotsPerRow(),
             // To keep in sync with Container::DEFAULT_DRY_RUN
             dryRun: (bool) $input->getOption(self::OPTION_DRY_RUN),
             useGitHubLogger: $commandHelper->getUseGitHubLogger(),
--- a/src/Command/RunCommand.php
+++ b/src/Command/RunCommand.php
@@ -180,7 +180,7 @@ final class RunCommand extends BaseCommand
             )
             ->addOption(
                 self::OPTION_DOTS_PER_ROW,
-                'r',
+                null,
                 InputOption::VALUE_REQUIRED,
                 'Number of dots per row in the dot progress formatter. Use "max" to fit the terminal width.',
                 Container::DEFAULT_DOTS_PER_ROW,
--- a/src/Command/RunCommandHelper.php
+++ b/src/Command/RunCommandHelper.php
@@ -35,11 +35,13 @@ declare(strict_types=1);
 
 namespace Infection\Command;
 
+use function ctype_digit;
 use function getenv;
 use Infection\Container\Container;
 use Infection\Resource\Processor\CpuCoresCountProvider;
 use InvalidArgumentException;
 use function is_numeric;
+use function is_string;
 use function max;
 use function sprintf;
 use Symfony\Component\Console\Input\InputInterface;
@@ -133,22 +135,21 @@ final readonly class RunCommandHelper
     {
         $dotsPerRow = $this->input->getOption(RunCommand::OPTION_DOTS_PER_ROW);
 
-        // user didn't pass `--dots-per-row` option
         if ($dotsPerRow === null) {
             return null;
         }
 
-        // user passed `--dots-per-row=<int>` option
-        if (is_numeric($dotsPerRow)) {
+        $error = sprintf('The value of option `--dots-per-row` must be a positive integer or string "max". "%s" provided.', $dotsPerRow);
+
+        if (is_string($dotsPerRow) && ctype_digit($dotsPerRow)) {
             $value = (int) $dotsPerRow;
 
-            Assert::positiveInteger($value, sprintf('The value of option `--dots-per-row` must be a positive integer or string "max". %d provided.', $value));
+            Assert::positiveInteger($value, $error);
 
             return $value;
         }
 
-        // user passed `--dots-per-row=max` option
-        Assert::same($dotsPerRow, 'max', sprintf('The value of option `--dots-per-row` must be a positive integer or string "max". String "%s" provided.', $dotsPerRow));
+        Assert::same($dotsPerRow, 'max', $error);
 
         return 'max';
     }
--- a/src/Command/RunCommandHelper.php
+++ b/src/Command/RunCommandHelper.php
@@ -126,6 +126,33 @@ final readonly class RunCommandHelper
         return max(1, CpuCoresCountProvider::provide() - 1);
     }
 
+    /**
+     * @return positive-int|'max'|null
+     */
+    public function getDotsPerRow(): string|int|null
+    {
+        $dotsPerRow = $this->input->getOption(RunCommand::OPTION_DOTS_PER_ROW);
+
+        // user didn't pass `--dots-per-row` option
+        if ($dotsPerRow === null) {
+            return null;
+        }
+
+        // user passed `--dots-per-row=<int>` option
+        if (is_numeric($dotsPerRow)) {
+            $value = (int) $dotsPerRow;
+
+            Assert::positiveInteger($value, sprintf('The value of option `--dots-per-row` must be a positive integer or string "max". %d provided.', $value));
+
+            return $value;
+        }
+
+        // user passed `--dots-per-row=max` option
+        Assert::same($dotsPerRow, 'max', sprintf('The value of option `--dots-per-row` must be a positive integer or string "max". String "%s" provided.', $dotsPerRow));
+
+        return 'max';
+    }
+
     public function getNumberOfShownMutations(): ?int
     {
         $shownMutations = $this->input->getOption(RunCommand::OPTION_SHOW_MUTATIONS);
--- a/src/Configuration/ConfigurationFactory.php
+++ b/src/Configuration/ConfigurationFactory.php
@@ -55,7 +55,6 @@ use Infection\Configuration\SourceFilter\SourceFilter;
 use Infection\FileSystem\Locator\FileOrDirectoryNotFound;
 use Infection\FileSystem\TmpDirProvider;
 use Infection\Git\Git;
-use Infection\Logger\MutationAnalysis\ConsoleDotLogger;
 use Infection\Mutator\ConfigurableMutator;
 use Infection\Mutator\Mutator;
 use Infection\Mutator\MutatorFactory;
@@ -65,7 +64,6 @@ use Infection\Reporter\FileReporter;
 use Infection\Resource\Processor\CpuCoresCountProvider;
 use Infection\Source\Exception\NoSourceFound;
 use Infection\TestFramework\TestFrameworkTypes;
-use function is_int;
 use function is_numeric;
 use function max;
 use OndraM\CiDetector\CiDetector;
@@ -88,6 +86,8 @@ class ConfigurationFactory
      */
     private const int DEFAULT_TIMEOUT = 10;
 
+    private const int DEFAULT_DOTS_PER_ROW = 50;
+
     public function __construct(
         private readonly TmpDirProvider $tmpDirProvider,
         private readonly MutatorResolver $mutatorResolver,
@@ -101,6 +101,7 @@ class ConfigurationFactory
 
     /**
      * @param non-empty-string|null $projectDirectory Absolute path.
+     * @param positive-int|'max'|null $dotsPerRow
      *
      * @throws FileOrDirectoryNotFound
      * @throws NoSourceFound
@@ -474,21 +475,13 @@ class ConfigurationFactory
     }
 
     /**
+     * @param positive-int|'max'|null $dotsPerRow
+     *
      * @return positive-int|'max'
      */
     private static function retrieveDotsPerRow(string|int|null $dotsPerRow, SchemaConfiguration $schema): int|string
     {
-        $value = $dotsPerRow ?? $schema->dotsPerRow ?? ConsoleDotLogger::DEFAULT_DOTS_PER_ROW;
-
-        if (is_int($value)) {
-            Assert::positiveInteger($value, sprintf('The value of `dotsPerRow` must be a positive integer or string "max". %d provided.', $value));
-
-            return $value;
-        }
-
-        Assert::same($value, 'max', sprintf('The value of `dotsPerRow` must be a positive integer or string "max". String "%s" provided.', $value));
-
-        return 'max';
+        return $dotsPerRow ?? $schema->dotsPerRow ?? self::DEFAULT_DOTS_PER_ROW;
     }
 
     /**
--- a/src/Configuration/ConfigurationFactory.php
+++ b/src/Configuration/ConfigurationFactory.php
@@ -55,6 +55,7 @@ use Infection\Configuration\SourceFilter\SourceFilter;
 use Infection\FileSystem\Locator\FileOrDirectoryNotFound;
 use Infection\FileSystem\TmpDirProvider;
 use Infection\Git\Git;
+use Infection\Logger\MutationAnalysis\ConsoleDotLogger;
 use Infection\Mutator\ConfigurableMutator;
 use Infection\Mutator\Mutator;
 use Infection\Mutator\MutatorFactory;
@@ -64,6 +65,7 @@ use Infection\Reporter\FileReporter;
 use Infection\Resource\Processor\CpuCoresCountProvider;
 use Infection\Source\Exception\NoSourceFound;
 use Infection\TestFramework\TestFrameworkTypes;
+use function is_int;
 use function is_numeric;
 use function max;
 use OndraM\CiDetector\CiDetector;
@@ -125,6 +127,7 @@ class ConfigurationFactory
         ?string $staticAnalysisToolOptions,
         PlainFilter|IncompleteGitDiffFilter|null $sourceFilter,
         ?int $threadCount,
+        string|int|null $dotsPerRow,
         bool $dryRun,
         ?bool $useGitHubLogger,
         ?string $gitlabLogFilePath,
@@ -192,6 +195,7 @@ class ConfigurationFactory
             maxTimeouts: self::retrieveMaxTimeouts($maxTimeouts, $schema),
             msiPrecision: $msiPrecision,
             threadCount: $this->retrieveThreadCount($threadCount, $schema),
+            dotsPerRow: self::retrieveDotsPerRow($dotsPerRow, $schema),
             isDryRun: $dryRun,
             ignoreSourceCodeMutatorsMap: $ignoreSourceCodeMutatorsMap,
             executeOnlyCoveringTestCases: $executeOnlyCoveringTestCases,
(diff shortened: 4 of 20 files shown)


src/Command/RunCommand.php
```diff
@@ -176,6 +178,13 @@ protected function configure(): void
                 'Number of threads to use by the runner when executing the mutations. Use "max" to auto calculate it.',
                 Container::DEFAULT_THREAD_COUNT,
             )
+            ->addOption(
+                self::OPTION_DOTS_PER_ROW,
+                'r',
```
theofidry: let's avoid a short option, `r` is not really intuitive

src/Command/RunCommandHelper.php
```diff
@@ -126,6 +126,33 @@ public function getThreadCount(): ?int
...
     }
 
+    /**
+     * @return positive-int|'max'|null
+     */
+    public function getDotsPerRow(): string|int|null
+    {
+        $dotsPerRow = $this->input->getOption(RunCommand::OPTION_DOTS_PER_ROW);
+
+        // user didn't pass `--dots-per-row` option
```
theofidry: no need for the superfluous comments here (same L141 and L150)

src/Logger/MutationAnalysis/ConsoleDotLogger.php
```diff
@@ -39,25 +39,50 @@
...
+     * `<total>` widths are added on top of this.
+     */
+    private const int KNOWN_COUNT_SUFFIX_FIXED_LENGTH = 8;
 
     private ?int $mutationCount = null;
 
+    private int $dotsPerRow = self::DEFAULT_DOTS_PER_ROW;
+
+    /**
+     * @param int|'max' $dotsPerRowSetting
```
theofidry:
```suggestion
     * @param positive-int|'max' $dotsPerRowSetting
```

src/Logger/MutationAnalysis/ConsoleDotLogger.php
```diff
@@ -39,25 +39,50 @@
...
+     */
     public function __construct(
         private readonly OutputInterface $output,
+        private readonly int|string $dotsPerRowSetting = self::DEFAULT_DOTS_PER_ROW,
+        private readonly ?Terminal $terminal = null,
     ) {
+        if (is_string($dotsPerRowSetting)) {
```
theofidry: no need for the extra validation here: this is ensured by SA, validating the user input is already done by that point

src/Logger/MutationAnalysis/ConsoleDotLogger.php
```diff
@@ -66,6 +91,7 @@ public function startAnalysis(int $mutationCount): void
         parent::startAnalysis($mutationCount);
 
         $this->mutationCount = $mutationCount;
+        $this->dotsPerRow = $this->resolveDotsPerRow($mutationCount);
```
theofidry: A different property should be used rather than overriding the current one. The `#dotsPerRow` could be kept, but you could have a private `::getDotsPerRow()` which uses `#dotsPerRow` if already set and otherwise resolves and set it.

src/Logger/MutationAnalysis/ConsoleDotLogger.php
```diff
@@ -124,6 +150,33 @@ public function finishEvaluation(MutantExecutionResult $executionResult): void
         }
     }
 
+    private function resolveDotsPerRow(int $mutationCount): int
```
theofidry, line 157:
```suggestion
    /**
     * @param positive-int|IterableCounter::UNKNOWN_COUNT $mutationCount
     */
    private function resolveDotsPerRow(int $mutationCount): int
```

src/Logger/MutationAnalysis/ConsoleDotLogger.php
```diff
@@ -124,6 +150,33 @@ public function finishEvaluation(MutantExecutionResult $executionResult): void
...
+    {
+        if (!is_string($this->dotsPerRowSetting)) {
+            return $this->dotsPerRowSetting;
+        }
+
+        $terminalWidth = ($this->terminal ?? new Terminal())->getWidth();
```
theofidry: let's make the terminal non-optional.

Otherwise what is happening here in the current run (not the tests) is that `#terminal = null`, so you _always_ use this new `Terminal` instance. In practice it is not a problem since `::startAnalysis()` should be called only once per run, but it creates unnecessary confusion.

src/Logger/MutationAnalysis/ConsoleDotLogger.php
```diff
@@ -124,6 +150,33 @@ public function finishEvaluation(MutantExecutionResult $executionResult): void
         }
     }
 
+    private function resolveDotsPerRow(int $mutationCount): int
+    {
+        if (!is_string($this->dotsPerRowSetting)) {
```
theofidry:
```suggestion
        if (is_int($this->dotsPerRowSetting)) {
```

src/Logger/MutationAnalysis/ConsoleDotLogger.php
```diff
@@ -39,25 +39,50 @@
...
+use Symfony\Component\Console\Terminal;
 use Webmozart\Assert\Assert;
 
 /**
  * @internal
  */
 final class ConsoleDotLogger extends AbstractMutationAnalysisLogger
 {
-    private const int DOTS_PER_ROW = 50;
+    public const int DEFAULT_DOTS_PER_ROW = 50;
```
theofidry: We are now moving this setting from an internal service configuration to a general project configuration, so this constant should be moved to. As such, this constant should be moved to `ConfigurationFactory`, like we do for the `DEFAULT_TIMEOUT`.

src/Logger/MutationAnalysis/ConsoleDotLogger.php
```diff
@@ -124,6 +150,33 @@ public function finishEvaluation(MutantExecutionResult $executionResult): void
...
+        $resolved = max(1, $terminalWidth - self::suffixLength($mutationCount));
+
+        if ($mutationCount === IterableCounter::UNKNOWN_COUNT) {
+            return $resolved;
+        }
+
+        return min($resolved, $mutationCount);
+    }
+
+    private static function suffixLength(int $mutationCount): int
```
theofidry, line 175:
```suggestion
    /**
     * @param positive-int|IterableCounter::UNKNOWN_COUNT $mutationCount
     * 
     * @return positive-int
     */
    private static function suffixLength(int $mutationCount): int
```

theofidry: (ignore the return type if that causes more SA annoyances)

src/Logger/MutationAnalysis/MutationAnalysisLoggerFactory.php
```diff
@@ -55,13 +55,18 @@ public function __construct(
...
     }
 
-    public function create(MutationAnalysisLoggerName $name): MutationAnalysisLogger
-    {
+    /**
+     * @param positive-int|'max' $dotsPerRow
+     */
+    public function create(
+        MutationAnalysisLoggerName $name,
+        int|string $dotsPerRow = ConsoleDotLogger::DEFAULT_DOTS_PER_ROW,
```
theofidry: No need for a default value

```suggestion
        int|string $dotsPerRow,
```

tests/phpunit/Command/RunCommandHelperTest.php
```diff
@@ -96,6 +96,33 @@ public static function providesThreadCount(): iterable
...
+    /**
+     * @return iterable<string, array{0: int|'max'|null, 1: string|null}>
+     */
+    public static function providesDotsPerRow(): iterable
+    {
+        yield 'not provided' => [null, null];
+
+        yield 'provided as numeric string' => [20, '20'];
+
+        yield 'provided as max' => ['max', 'max'];
```
theofidry, line 127: it would be good to have a non-valid case (e.g. non-natural – negative value and float, or a random and empty string) added here.

This can be added to this provider rather than another test (see `MapSourceClassToTestOptionTest` which can allow an exception as the expected result)

tests/phpunit/Command/RunCommandHelperTest.php
```diff
@@ -96,6 +96,33 @@ public static function providesThreadCount(): iterable
...
+            ->with(RunCommand::OPTION_DOTS_PER_ROW)
+            ->willReturn($optionValue);
+
+        $commandHelper = new RunCommandHelper($this->inputMock);
+        $this->assertSame($expected, $commandHelper->getDotsPerRow());
+    }
+
+    /**
+     * @return iterable<string, array{0: int|'max'|null, 1: string|null}>
+     */
```
theofidry: We do not add PHPDoc to the data providers (although currently Mago complains about it)

```suggestion

```

tests/phpunit/Command/RunCommandHelperTest.php
```diff
@@ -96,6 +96,33 @@ public static function providesThreadCount(): iterable
         yield [5, '5'];
     }
 
+    /**
+     * @param int|'max'|null $expected
+     */
```
theofidry: This phpdoc only narrows the type of `string` but this is the test about validating the user input, the user input can be a string different than `'max'`. Mago will complain about those cases there but it is a case where we need to work out the mago configuration (out of scope of this PR, so ignore it) rather than trying to please it

```suggestion

```

src/Configuration/Schema/SchemaConfiguration.php
```diff
@@ -76,6 +76,7 @@ public function __construct(
         public ?string $testFrameworkExtraOptions,
         public ?string $staticAnalysisToolOptions,
         public string|int|null $threads,
+        public string|int|null $dotsPerRow,
```
theofidry, line 80: missing the PHPDoc here

src/Configuration/ConfigurationFactory.php
```diff
@@ -469,6 +473,24 @@ private function retrieveThreadCount(?int $threadCount, SchemaConfiguration $sch
...
     }
 
+    /**
+     * @return positive-int|'max'
+     */
+    private static function retrieveDotsPerRow(string|int|null $dotsPerRow, SchemaConfiguration $schema): int|string
+    {
+        $value = $dotsPerRow ?? $schema->dotsPerRow ?? ConsoleDotLogger::DEFAULT_DOTS_PER_ROW;
+
+        if (is_int($value)) {
```
theofidry: with the updated values (default constant, the schema configuration phpdoc, etc.), those assertions should now be unnecessary

tests/phpunit/Logger/MutationAnalysis/ConsoleDotLoggerTest.php
```diff
@@ -232,10 +233,155 @@ public function test_it_prints_current_number_of_processed_mutations_when_the_to
         $this->assertSame($expected, $actual);
     }
 
+    public function test_it_honors_a_custom_dots_per_row(): void
```
theofidry: I would rather enrich `test_it_prints_total_number_of_mutations`:


```suggestion
    /**
     * @param 'max'|positive-int $dotsPerRowSetting
     */
    #[DataProvider('dotsPerRowProvider')]
    public function test_it_prints_total_number_of_mutations(
        string|int $dotsPerRowSetting,
        string $expected,
    ): void
```

tests/phpunit/Logger/MutationAnalysis/ConsoleDotLoggerTest.php
```diff
@@ -232,10 +233,155 @@ public function test_it_prints_current_number_of_processed_mutations_when_the_to
...
+                ...          (23 / 23)
+                TXT,
+        );
+
+        $actual = strip_tags($output->fetch());
+
+        $this->assertSame($expected, $actual);
+    }
+
+    public function test_it_resolves_max_to_terminal_width_minus_suffix(): void
```
theofidry: this could leverage `test_it_prints_total_number_of_mutations` instead too (i.e. enrich the test & data provider) by allowing to pass the terminal width

tests/phpunit/Logger/MutationAnalysis/ConsoleDotLoggerTest.php
```diff
@@ -232,10 +233,155 @@ public function test_it_prints_current_number_of_processed_mutations_when_the_to
...
+                .......              (25 / 25)
+                TXT,
+        );
+
+        $actual = strip_tags($output->fetch());
+
+        $this->assertSame($expected, $actual);
+    }
+
+    public function test_it_falls_back_to_one_dot_per_row_when_the_terminal_is_narrower_than_the_suffix(): void
```
theofidry: likewise

tests/phpunit/Logger/MutationAnalysis/ConsoleDotLoggerTest.php
```diff
@@ -232,10 +233,155 @@ public function test_it_prints_current_number_of_processed_mutations_when_the_to
...
+                .   (3 / 3)
+                TXT,
+        );
+
+        $actual = strip_tags($output->fetch());
+
+        $this->assertSame($expected, $actual);
+    }
+
+    public function test_it_caps_max_at_the_total_mutation_count(): void
```
theofidry: likewise

tests/phpunit/Logger/MutationAnalysis/ConsoleDotLoggerTest.php
```diff
@@ -232,10 +233,155 @@ public function test_it_prints_current_number_of_processed_mutations_when_the_to
...
+    }
+
     private function createMutantExecutionResultOfType(DetectionStatus $detectionStatus): MutantExecutionResult
     {
         return MutantExecutionResultBuilder::withMinimalTestData()
             ->withDetectionStatus($detectionStatus)
             ->build();
     }
+
+    private function terminalWithWidth(int $width): Terminal
```
theofidry:
```suggestion
    private function createTerminal(int $width): Terminal
```

tests/phpunit/Logger/MutationAnalysis/ConsoleDotLoggerTest.php
```diff
@@ -232,10 +233,155 @@ public function test_it_prints_current_number_of_processed_mutations_when_the_to
...
+                .   (2 / 3)
+                .   (3 / 3)
+                TXT,
+        );
+
+        $actual = strip_tags($output->fetch());
+
+        $this->assertSame($expected, $actual);
+    }
+
```
theofidry, line 278: We are missing the analogous cases where the mutation counts are unknown (see `test_it_prints_current_number_of_processed_mutations_when_the_total_number_is_not_known`). It probably need to be adjusted the same way `test_it_prints_total_number_of_mutations` is

theofidry (commented):
There is quite a few little nitpicks, but overall the direction of the PR is good 👍

resources/schema.json
```diff
@@ -26,6 +26,19 @@
...
+        "dotsPerRow": {
+            "description": "Number of dots per row in the dot progress formatter. Use \"max\" to fit the terminal width.",
+            "oneOf": [
+                {
+                    "type": "integer",
+                    "minimum": 1
+                },
+                {
+                    "type": "string",
+                    "enum": ["max"]
```
sanmai, line 38: what about `auto` instead of `max`

theofidry: With `auto` you'll probably wonder what width will be picked up, `max` has the advantage to leave no ambiguity: it is the maximum width available.

src/Command/RunCommand.php
paulbalandan: Removed.

src/Command/RunCommandHelper.php
paulbalandan: Removed.

src/Logger/MutationAnalysis/ConsoleDotLogger.php
paulbalandan: Fixed.

src/Logger/MutationAnalysis/ConsoleDotLogger.php
paulbalandan: Removed.

src/Logger/MutationAnalysis/ConsoleDotLogger.php
paulbalandan: Added `$resolvedDotsPerRow`

src/Logger/MutationAnalysis/ConsoleDotLogger.php
paulbalandan: Added

src/Logger/MutationAnalysis/ConsoleDotLogger.php
paulbalandan: Changed

src/Logger/MutationAnalysis/ConsoleDotLogger.php
paulbalandan: Done.

src/Logger/MutationAnalysis/ConsoleDotLogger.php
paulbalandan: Fixed

src/Logger/MutationAnalysis/ConsoleDotLogger.php
paulbalandan: Done.

src/Logger/MutationAnalysis/MutationAnalysisLoggerFactory.php
paulbalandan: Fixed

tests/phpunit/Command/RunCommandHelperTest.php
paulbalandan: Removed.

tests/phpunit/Command/RunCommandHelperTest.php
paulbalandan: Removed.

src/Configuration/ConfigurationFactory.php
paulbalandan: Fixed.

tests/phpunit/Logger/MutationAnalysis/ConsoleDotLoggerTest.php
paulbalandan: Done.

tests/phpunit/Logger/MutationAnalysis/ConsoleDotLoggerTest.php
paulbalandan: Done.

tests/phpunit/Logger/MutationAnalysis/ConsoleDotLoggerTest.php
paulbalandan: Done.

tests/phpunit/Logger/MutationAnalysis/ConsoleDotLoggerTest.php
paulbalandan: Done.

tests/phpunit/Logger/MutationAnalysis/ConsoleDotLoggerTest.php
paulbalandan: Renamed.

src/Configuration/Schema/SchemaConfiguration.php
paulbalandan: Added

tests/phpunit/Logger/MutationAnalysis/ConsoleDotLoggerTest.php
paulbalandan: Added to the tests.

theofidry (approved):

theofidry:
@coderabbitai full review

tests/phpunit/Command/RunCommandHelperTest.php
paulbalandan: Added all invalids

theofidry:
It is a bit late so may have missed something, LGTM otherwise. Thanks @paulbalandan?

src/Command/RunCommand.php
```diff
@@ -82,6 +82,8 @@ final class RunCommand extends BaseCommand
 {
     public const string OPTION_THREADS = 'threads';
 
+    public const string OPTION_DOTS_PER_ROW = 'dots-per-row';
```
maks-rafalko, line 85: Could you please create docs for both new option and configuration.

Option can be documented here: https://github.com/infection/site/blob/master/src/guide/command-line-options.md

Config changes here: https://github.com/infection/site/blob/master/src/guide/usage.md

tests/phpunit/Command/RunCommandHelperTest.php
```diff
@@ -96,6 +97,56 @@ public static function providesThreadCount(): iterable
...
+
+        if ($expected instanceof InvalidArgumentException) {
+            $this->expectExceptionObject($expected);
+        }
+
+        $actual = $commandHelper->getDotsPerRow();
+
+        if (!$expected instanceof InvalidArgumentException) {
+            $this->assertSame($expected, $actual);
+        }
```
maks-rafalko, line 118: sorry if it has been already discussed (if so, could you point me to the discussion), but looks like we need 2 separate tests here - one for exceptions and one for other cases.

UPD: I found the [discussion](https://github.com/infection/infection/pull/3129#discussion_r3224868071), still I don't like it. But since this is out of the scope of this PR, let's keep as it is

maks-rafalko (commented):
Awesome work. Tested, works as expected

Image

We need to update the docs and happy to merge then

src/Command/RunCommand.php
paulbalandan: Thank you! Raised a PR in the infection site repo.

maks-rafalko (approved):
Nice, thank you for working on this and thanks all for the review

PR #3131, opened 2026-05-12
test(ast): Fix the naming of a few traverser related variables

theofidry:
Those renames were missed in #2795.


sanmai (approved):
code makes sense 👍

theofidry:
I'm just really confused by the failure 😓 but I'll check that tonight

PR #3132, opened 2026-05-12
feat(testFramework): Add test framework extra-args

### LINKED ISSUE #2924: bug: Allow infection to execute a specific test
| Question    | Answer
| Infection version | 0.32.3
| CI | any
| PHP version | any
| Platform    | any
| Github Repo | infection

Currently executing:

```shell
infection --test-framework-options="/path/to/test.php"
```

Will fail because the executed command will be:

```shell
/path/to/phpunit --configuration=<dumped-config> --/path/to/test.php ...
```

Notice the leading `--`

I don't think we should alter the passed options this way, although it would no longer be options but arguments... And this would be a BC and we need to check that it works for all test frameworks.

The ability to execute infection for a given test would be very useful for https://github.com/j-plugins/infection-plugin/issues/33.

theofidry:
## Description

Adds `testFrameworkExtraArgs` / `--test-framework-extra-args` as the supported way to pass raw test framework arguments through Infection.

The existing `testFrameworkOptions` / `--test-framework-options` name is kept as a deprecated fallback, but the new name makes the contract clearer: these values are user-provided test framework arguments, not Infection-generated options. The implementation also rejects configurations and CLI invocations that provide both names, so precedence is explicit rather than silently ambiguous.

## Name choice

In the context of the CLI, "argument" refers to any token after the command name. For example `$argv` (argument vector, from the C `argv`), contains the script name (`$argv[0]`) and then all raw tokens. What "argument" refers to when distinguishing arguments from options, e.g. when adding arguments and options to a Symfony Console, is "positional arguments".

So the name chosen is correct but indeed ambiguous. It is unfortunatly a case that doesn't seem to have a nice solution without being overly verbose. 

## Changes

- Adds the `--test-framework-extra-args` CLI option and wires it through `run` and `initial-tests`.
- Adds the `testFrameworkExtraArgs` JSON configuration key and marks `testFrameworkOptions` as deprecated in the schema.
- Rejects configurations or CLI invocations that provide both the legacy and new names.
- Keeps legacy `testFrameworkOptions` behaviour as a fallback when the new option is not provided.

## Breaking Changes

Since the `initial-test:run` is mostly for debugging, this PR removed support for `--test-framework-options` for it entirely, whilst preserving it (and deprecating it) for the default command.

Otherwise, no immediate breaking change is intended. `testFrameworkOptions` / `--test-framework-options` remains supported as a deprecated fallback.

## Reviewer Notes

In the context of the CLI, "argument" refers to any token after the command name. For example $argv (argument vector, from the C argv), contains the script name ($argv[0]) and then all raw tokens.

What "argument" refers to when distinguishing arguments from options is "positional arguments".

So not an easy one.

## Related issues

- Fixes #2924.
- doc PR: https://github.com/infection/site/pull/300


src/TestFramework/TestFrameworkExtraArgs.php
```diff
@@ -0,0 +1,175 @@
...
+use function substr;
+use Symfony\Component\Console\Exception\InvalidArgumentException;
+use Symfony\Component\Console\Input\StringInput;
+use function trim;
+use Webmozart\Assert\Assert;
+
+/**
+ * @internal
+ */
+final readonly class TestFrameworkExtraArgs
```
theofidry: this whole thing needs to be reviewed in extra details this seems off

src/TestFramework/PhpUnit/Adapter/PhpUnitAdapter.php
theofidry, line 93: this seems incorrect

src/Command/Option/TestFrameworkExtraArgsOption.php
```diff
@@ -33,21 +33,43 @@
...
+    public static function addOption(Command $command): Command
     {
-        return new self(
-            'Could not find a git project.',
-            previous: $previous,
+        return $command->addOption(
+            self::NAME,
+            null,
+            InputOption::VALUE_REQUIRED,
+            'Raw PHPUnit arguments to pass before Infection-generated arguments',
```
maks-rafalko, line 65: I thought we need to support both arguments and options here, not only arguments (see https://github.com/infection/infection/issues/2924)

then the name of this options is also wrong

theofidry: It depends...

In the context of the CLI, "argument" refers to any token after the command name. For example `$argv` (argument vector, from the C `argv`), contains the script name (`$argv[0]`) and then all raw tokens.

What "argument" refers to when distinguishing arguments from options is "positional arguments".

So not an easy one.

resources/schema.json
```diff
@@ -679,7 +679,12 @@
         },
         "testFrameworkOptions": {
             "type": "string",
-            "description": "Specify an additional options to pass to the test framework (e.g. enabling Verbose Mode). --test-framework-options will override this option."
+            "deprecated": true,
+            "description": "Deprecated. Use testFrameworkExtraArgs instead."
+        },
+        "testFrameworkExtraArgs": {
+            "type": "string",
+            "description": "Specify raw additional arguments to pass to PHPUnit."
```
maks-rafalko:
```suggestion
            "description": "Specify raw additional arguments to pass to Test Framework."
```

src/Command/InitialTest/InitialTestRunCommand.php
```diff
@@ -71,7 +71,7 @@ protected function configure(): void
         FilterOption::addOption($this);
         InitialTestsPhpOptionsOption::addOption($this);
         TestFrameworkOption::addOption($this);
-        TestFrameworkOptionsOption::addOption($this);
+        TestFrameworkExtraArgsOption::addOption($this);
```
maks-rafalko, line 74: shouldn't it be added in addition to `TestFrameworkOptionsOption::addOption($this);`? RunCommand has both still

src/Configuration/Schema/SchemaConfigurationFactory.php
```diff
@@ -82,6 +85,7 @@ public function create(string $pathname, stdClass $rawConfig): SchemaConfigurati
             bootstrap: self::normalizeString($rawConfig->bootstrap ?? null),
             initialTestsPhpOptions: self::normalizeString($rawConfig->initialTestsPhpOptions ?? null),
             testFrameworkExtraOptions: self::normalizeString($rawConfig->testFrameworkOptions ?? null),
+            testFrameworkExtraArgs: self::normalizeString($rawConfig->testFrameworkExtraArgs ?? null),
```
maks-rafalko, line 88: do we need it here when we have this guard in `RunCommand.php` that is triggered earlier?

src/TestFramework/TestFrameworkExtraArgs.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * @param list<string> $argvTokens
+     */
+    private function __construct(
+        public array $argvTokens,
+    ) {
+    }
+
+    public static function raw(?string $value): self
```
maks-rafalko: this seems unused except tests

maks-rafalko: removed. This was a duplicate (at least by file name) for `src/TestFramework/PhpUnit/CommandLine/TestFrameworkExtraArgs.php`

src/Command/InitialTest/InitialTestRunCommand.php
maks-rafalko: added in addition, not instead of.  Same as in RunCommand

resources/schema.json
maks-rafalko: updated

maks-rafalko (approved):
compiled PHAR from this branch and tried on my target project - it works!

This is how I used it on a project that takes ~10minutes to execute full test suite:

```shell
./infection.phar \
   --test-framework-extra-args="tests/path/to/DocumentLinesTest.php" \
   --filter=path/to/DocumentLinesTest.php
```

with this command

- we only run 1 test by PHPUnit
- PHPUnit generates coverage only from this test
- we mutate only 1 desired file

and it works fast.

src/Configuration/Schema/SchemaConfigurationFactory.php
theofidry: maybe the comment got removed: yes, the `RunCommand` checks the options passed whereas `SchemaConfiguration` checks the options registered in the configuration file

resources/schema.json
```diff
@@ -679,7 +679,12 @@
         },
         "testFrameworkOptions": {
             "type": "string",
-            "description": "Specify an additional options to pass to the test framework (e.g. enabling Verbose Mode). --test-framework-options will override this option."
+            "deprecated": true,
```
theofidry, line 682: note that deprecated has no effect there, it's purely informative

src/Command/InitialTest/InitialTestRunCommand.php
theofidry: Then the change L92 is incorrect as the old option is not forwarded.

My thoughts were since this is a debug command we can be a bit more aggressive and not bother with a deprecation path. This also allowed to avoid to duplicate the guard check from `RunCommand`

theofidry:
@maks-rafalko sorry for being a bit slow on this, replied

src/Command/InitialTest/InitialTestRunCommand.php
maks-rafalko: got it, ok reverter

maks-rafalko:
cool, let's get it merged. If anything else you think needs to be addressed, let me know please.

Thank for working on this! I will continue with https://github.com/infection/infection/issues/2191 to make this even more DX friendly

PR #3133, opened 2026-05-13
test(phpunit): Expand PHPUnit adapter command-line scenario coverage

### LINKED ISSUE #2924: bug: Allow infection to execute a specific test

theofidry:
_Status: having some trouble with Windows; checking how to properly mock the PHP executable on it_

## Description

While implementing #3132 to address #2924, several confusing PHPUnit adapter command-line cases came up.

The broader test framework adapter API issues are discussed in #2863. This PR does not try to solve those design problems directly. Instead, it improves test coverage for the current PHPUnit adapter behaviour, making the existing transformations easier to review, reason about, and change safely.

Several cases are already broken or ambiguous. The immediate objectives are therefore:

- Make it possible to fix #2924, even if the resulting implementation is still constrained by the current adapter API.
- Document the current behaviour well enough to support a proper follow-up fix through #2863.

To support this, the PR introduces a focused PHPUnit adapter test covering the command-line argument and option transformations currently performed by the adapter. The test makes the existing behaviour explicit and documents the bugs identified in these scenarios.

I opened https://github.com/infection/infection/issues/3138 for the discovered bugs.

## Changes

- Adds scenario coverage for PHPUnit adapter command-line argument and option transformations.
- Documents currently broken or ambiguous transformations as explicit test cases.
- Provides a safer baseline for fixing #2924 and for the broader adapter work in #2863.

## Related issues

- https://github.com/infection/infection/issues/2924
- https://github.com/infection/infection/pull/3132


sanmai (approved):
Scanned through the code, makes sense

theofidry:
@coderabbitai full review

PR #3135, opened 2026-05-14
test(mago): Fix the mago e2e test

theofidry:
## Description

Skip the `Mago_Integration` e2e test when the Infection executable passed to the e2e runner is not a PHAR.

The fixture loads its own Mago dependency. When running Infection through `bin/infection`, the root project Mago dependency is already loaded before the fixture autoloader is included, which can trigger duplicate Mago function declarations. The PHAR path avoids that dependency collision and remains the supported mode for this fixture.

## Reviewer Notes

Verified the source-binary skip path with:

```shell
./tests/e2e_tests bin/infection Mago_Integration
# no infection.log is generated

make compile
./tests/e2e_tests dist/infection.phar Mago_Integration
# infection.log is generated and checked
```

The PHAR execution path was not verified locally because `dist/infection.phar` was not present.

### DIFF
--- a/tests/e2e/Mago_Integration/run_tests.bash
+++ b/tests/e2e/Mago_Integration/run_tests.bash
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+cd "$(dirname "$0")"
+
+if [[ "${1}" != *.phar ]]
+then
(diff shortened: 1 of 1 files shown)


maks-rafalko:
why didn't we have CI failures then? 🤔

theofidry:
@maks-rafalko because of the issue described in https://github.com/infection/infection/pull/3136

PR #3136, opened 2026-05-14
test(ci): Run PHAR e2e jobs with the compiled artifact path

theofidry:
## Description

Fix the PHAR e2e CI matrix so the PHAR jobs actually match the workflow step that runs the full e2e suite.

The matrix was using `bin/infection.phar`, while the PHAR execution step is gated on `dist/infection.phar`. As a result, PHP PHAR jobs with either coverage driver could skip every e2e execution step.

The issue was introduced in #3068, my mistake.

## Changes

- Updates the PHP 8.3 PHAR matrix entries from `bin/infection.phar` to `dist/infection.phar`.


maks-rafalko:
not sure I got it, why doesn't it fail with incorrect path?

theofidry:
@maks-rafalko I am not sure yet, I still need to check that

PR #3147, opened 2026-05-16
fix(phpunit): Parse PHPUnit extra options as console input tokens

### LINKED ISSUE #3141: bug(PHPUnit): Fix extra options parsing
See the parent task.

theofidry:
## Description

Fixes PHPUnit `testFrameworkExtraArgs` / `--test-framework-extra-args` parsing so quoted values, paths containing spaces, and non-space whitespace separators are converted into the argv tokens PHPUnit expects instead of being split with a plain space delimiter.

This solves a whole lot of issues identified in https://github.com/infection/infection/issues/3141.

## Changes

- Adds `TestFrameworkExtraArgs` to parse raw PHPUnit extra args with Symfony Console `StringInput`. The name will probably change in the future as the solution is expended to other test frameworks.
- Updates `ArgumentsAndOptionsBuilder` to append parsed argv tokens instead of `explode(' ', ...)` output as the latter was insufficient to properly delimitate tokens.
- Updates PHPUnit adapter expectations for quoted filters, test suites, paths with spaces, option-like text inside quoted values, and newline/tab-separated args.

## Related issues

Fixes https://github.com/infection/infection/issues/3141.

Preparatory to #3132.


theofidry:
@coderabbitai full review

tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapter/PhpUnitAdapterTest.php
```diff
@@ -341,11 +337,7 @@ public static function initialTestRunProvider(): iterable
                     '/path/to/phpunit',
                     '--configuration',
                     '/tmp/phpunitConfiguration.initial.infection.xml',
-                    // Incorrect current behaviour: the shell quotes are preserved and split across argv tokens.
-                    '--filter="a',
-                    'test',
-                    'with',
-                    'spaces"',
+                    '--filter=a test with spaces',
```
maks-rafalko, line 340: is it a correct value though? why?

```shell
vendor/bin/phpunit --filter=a test
```

```shell
PHPUnit 13.1.8 by Sebastian Bergmann and contributors.

Test file "test" not found
```

tests/phpunit/TestFramework/PhpUnit/CommandLine/TestFrameworkExtraArgsTest.php
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * @return iterable<string, array{string, list<string>}>
+     */
+    public static function rawTokensProvider(): iterable
+    {
+        yield 'quoted value' => [
+            ' tests/FooTest.php --filter="a test" --colors=always ',
+            [
+                'tests/FooTest.php',
+                '--filter=a test',
```
maks-rafalko, line 80: same question, looks incorrect

src/TestFramework/PhpUnit/CommandLine/TestFrameworkExtraArgs.php
```diff
@@ -0,0 +1,155 @@
...
+        // @phpstan-ignore function.alreadyNarrowedType
+        return method_exists(StringInput::class, 'getRawTokens')
+            ? (new StringInput($value))->getRawTokens()
+            : self::tokenize($value);
+    }
+
+    /**
+     * @return list<string>
+     */
+    private static function tokenize(string $input): array
```
maks-rafalko, line 103: could you please explain

- why do we we have a condition for Symfony 6.4
- why do we have our own copy of `tokenize` here?

tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapter/PhpUnitAdapterTest.php
theofidry: > `vendor/bin/phpunit --filter=a test`

Indeed that is incorrect. But this is not what we are doing here, what is done is:

```php
new Symfony\Component\Process\Process([
  'vendor/bin/phpunit',
  '--filter=a test with spaces',
]);
```

Which will become (you can see it if you call `::getCommandLine()`):

```shell
'vendor/bin/phpunit' '--filter=a test with spaces'
```

Which does work.

src/TestFramework/PhpUnit/CommandLine/TestFrameworkExtraArgs.php
theofidry: `StringInput::getRawTokens()` is available only as of Symfony 7.1, hence we need a fallback for 6.4 or 7.0 (and hence why there is a bit more tests than otherwise necessary as we have some custom code).

Note that I've not been too demanding regarding the fallback code: this was generated by Codex, it worked in the few tested cases, but it's not otherwise code I'm happy with. But it is for a temporary polyfill so I did not spend much time on it either.

maks-rafalko: > `StringInput::getRawTokens()` is available only as of Symfony 7.1,

hm.. I don't see this method in any of the versions, from 6.4 to 8.0:

- https://github.com/symfony/console/blob/7.1/Input/StringInput.php
- https://github.com/symfony/console/blob/7.1/Input/ArrayInput.php

theofidry: it's in `ArgvInput`: https://github.com/symfony/console/blob/7.1/Input/ArgvInput.php#L355

tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapter/PhpUnitAdapterTest.php
maks-rafalko: oh nice, but it looks weird in tests unfortunately, which is not for sure your fault, just saying that it's confusing.

src/TestFramework/PhpUnit/CommandLine/TestFrameworkExtraArgs.php
maks-rafalko: damn, I checked incorrect `ArrayInput` instead of `ArgvInput`, misclicked

thanks

maks-rafalko (approved):

tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapter/PhpUnitAdapterTest.php
theofidry: I agree :/ But in the end that is what we are passing to `Process`, and that is how `ArgvInput/StringInput` format it, so I don't think it's wise to fight it either

PR #3158, opened 2026-05-17
ci: Introduce zizmor GitHub Actions security analysis

### LINKED ISSUE #3134: ci: introduce zizmor GitHub Actions security analysis
## Problem Statement

GitHub Actions workflows are security-sensitive and currently lack automated security linting. Regressions such as broad token permissions, unsafe triggers, persisted checkout credentials, or unpinned actions can slip into review.

## Proposed Solution

Add a dedicated workflow running [zizmor](https://github.com/zizmorcore/zizmor) on GitHub Actions changes, with PR annotations enabled.

## Technical Considerations

- Fix initial findings or add justified local ignores.
- Keep workflow permissions and checkout credential persistence explicit.

## Related Issues

- Composer reference implementation: https://github.com/composer/composer/pull/12821
- PIE reference implementation: https://github.com/composer/composer/pull/12821
- zizmor project: https://github.com/zizmorcore/zizmor
- zizmor GitHub Action: https://github.com/zizmorcore/zizmor-action

theofidry:
## Description

Adds a local Docker Compose entry and a GitHub Actions workflow for running [zizmor](https://docs.zizmor.sh), then addresses the **high-severity** findings it reported in the existing workflows. Warnings and lower are left alone for now.

The changes focus on removing the actionable high-risk findings while keeping existing workflow behavior intact where a direct replacement would change semantics.

To execute it, for now:

```shell
make zizmor
```

In the future this may be added to the `autoreview` command, but there is still too many warnings to address, which will be done in a separate PR.

### Introducing zizmor

- Adds a `zizmor` Docker Compose service to run the audit with `make zizmor`.
  This makes the security audit reproducible locally without requiring every contributor to install `zizmor` directly.
- Pins the Docker Compose `zizmor` image to `ghcr.io/zizmorcore/zizmor:1.25.2` instead of `latest`.
  Floating container tags can change without a repository diff, so pinning the local audit image keeps local results reproducible.
- Adds a Dependabot `docker-compose` entry for the repository root.
  Dependabot can now propose updates when a newer pinned `zizmor` image tag is available, avoiding a stale manual pin; see [dependabot support for docker-compose](https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories).
- Adds a `zizmor` GitHub Actions workflow for changes under `.github/**` and `docker-compose.yml`.
  This keeps high-severity GitHub Actions security findings visible in CI when workflow or audit runner files change. The action is pinned by commit, annotations are enabled, and the workflow is scoped to `min-severity: high` so existing medium warnings do not fail unrelated CI.

### Workflow fixes

- Pins `awalsh128/cache-apt-pkgs-action` to the commit currently behind `v1`.
  GitHub Actions tags can be moved, so pinning third-party actions to immutable commits avoids silently executing different code in CI.
  See https://docs.zizmor.sh/audits/#unpinned-uses.
- Adds Dependabot cooldowns for GitHub Actions and Docker Compose updates. This avoids immediate follow-up update PRs when upstreams publish a quick succession of releases; see https://docs.zizmor.sh/audits/#dependabot-cooldown.
- Moves GitHub expression values used by shell scripts in `perf-metrics.yaml` into step environment variables.
  Expanding `${{ ... }}` directly inside `run` blocks can turn attacker-controlled values into shell syntax; reading them through environment variables keeps them as data.
  See https://docs.zizmor.sh/audits/#template-injection.
- Keeps the `pull_request_target`-based require-tests workflow, but documents the scoped `zizmor` ignore and restricts the workflow token to `pull-requests: write`.
  The workflow needs to create a PR review, so it still needs a privileged trigger for fork PRs, but the workflow does not checkout or execute PR code. The explicit permission avoids relying on GitHub's broader default token permissions.
  See https://docs.zizmor.sh/audits/#dangerous-triggers and https://docs.zizmor.sh/audits/#excessive-permissions.
- Splits the Docker sandbox `container-structure-test` cache into explicit restore and save steps, and only saves the cache on trusted `push` runs.
  Pull request and release runs can restore an existing cache, but they cannot create or update that cache entry. This keeps the download cached while preventing less trusted runs from poisoning runtime artifacts used later.
  See https://docs.zizmor.sh/audits/#cache-poisoning.

## Reviewer Notes

The `pull_request_target` finding is intentionally ignored inline because the workflow does not checkout or execute pull request code; it only asks the pinned `infection/tests-checker-action` to inspect PR file names and create a review.

The Docker sandbox workflow keeps caching `container-structure-test`, but cache writes are restricted to trusted push runs. This is why the workflow uses `actions/cache/restore` and `actions/cache/save` instead of the combined `actions/cache` action.

## Related issues

Fixes #3134.


docker-compose.yml
```diff
@@ -1,4 +1,14 @@
 services:
+  # May require you to login to ghcr.io first:
+  # $ docker login ghcr.io --username=<username>
+  # If you need a token, a short-lived classic token with read:packages is enough.
+  zizmor:
```
maks-rafalko, line 5: I would add something similar to what we have in the description of this PR

> Adds a zizmor Docker Compose service to run the audit with docker compose run zizmor.
> This makes the security audit reproducible locally without requiring every contributor to install zizmor directly.

somewhere so we now how to run it and why. Here in the comment?

maks-rafalko (approved):
Nice!

docker-compose.yml
theofidry: Actually should just be available as a make command...

maks-rafalko (approved):

PR #3165, opened 2026-05-18
ci: Use gh for the release

theofidry:
## Description

Removes the remaining unsuppressed `zizmor` finding from the release workflow.

The release job no longer depends on `svenstaro/upload-release-action` for functionality that is already available on GitHub-hosted runners. It now uses the preinstalled GitHub CLI to create the tag release when needed and upload the PHAR artifacts.

## Changes

- Replaces `svenstaro/upload-release-action` with `gh release view/create` and `gh release upload`.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -59,9 +59,10 @@ jobs:
               dist/infection.phar
 
       - name: Upload PHAR to release
-        uses: svenstaro/upload-release-action@29e53e917877a24fad85510ded594ab3c9ca12de # latest
-        with:
-          repo_token: ${{ secrets.GITHUB_TOKEN }}
-          file: dist/infection.phar*
-          file_glob: true
-          tag: ${{ github.ref }}
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release view "${GITHUB_REF_NAME}" > /dev/null 2>&1 \
+            || gh release create "${GITHUB_REF_NAME}" --verify-tag --title "${GITHUB_REF_NAME}"
+
+          gh release upload "${GITHUB_REF_NAME}" dist/infection.phar*


.github/workflows/release.yaml
```diff
@@ -59,9 +59,10 @@ jobs:
...
-          file: dist/infection.phar*
-          file_glob: true
-          tag: ${{ github.ref }}
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release view "${GITHUB_REF_NAME}" > /dev/null 2>&1 \
+            || gh release create "${GITHUB_REF_NAME}" --verify-tag --title "${GITHUB_REF_NAME}"
+
+          gh release upload "${GITHUB_REF_NAME}" dist/infection.phar*
```
sanmai, line 68: how's that going to work with immutable releases?

theofidry: No idea, but we are not using them yet either, so not an immediate problem at least.

I think PIE solved that problem but I didn't get a chance to take a look yet

sanmai: No problem, just wanted to highlight that the current approach can bite us if we ever need immutable releases.

theofidry: Looks like the solution is to use a draft: https://github.com/php/pie/blob/1.5.x/.github/workflows/release.yml

.github/workflows/release.yaml
```diff
@@ -59,9 +59,10 @@ jobs:
...
-        with:
-          repo_token: ${{ secrets.GITHUB_TOKEN }}
-          file: dist/infection.phar*
-          file_glob: true
-          tag: ${{ github.ref }}
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release view "${GITHUB_REF_NAME}" > /dev/null 2>&1 \
+            || gh release create "${GITHUB_REF_NAME}" --verify-tag --title "${GITHUB_REF_NAME}"
```
maks-rafalko, line 66: could you explain it? why do we need to create a release here if this job is created when we create a release (together with the tag)? We didn't have this before so it's unclear why we need it here

theofidry: From what I understood, `svenstaro/upload-release-action` was doing that.

Additionally this job is triggered on tag push, not release, although in practice I suspect we are all just publish the release in the GitHub UI.

But at least that was the suggested change, which was on part with how `svenstaro/upload-release-action` worked, so I left it alone.

maks-rafalko: ok, thanks, I will test it

> Additionally this job is triggered on tag push, not release, although in practice I suspect we are all just publish the release in the GitHub UI.

yes, 100% of times I do it through UI

maks-rafalko: it works https://github.com/infection/infection/releases/tag/0.33.2 👍

PR #3167, opened 2026-05-18
ci: Make Zizmor fail in the CI on warning severity

theofidry:
- Mirror the configuration of the `make zizmor` command.
- Rename the workflow to make it a bit less lengthy.

Note that the workflow is not added to the branch protection rules, this is because its run is conditional.

### DIFF
--- a/.github/workflows/zizmor.yaml
+++ b/.github/workflows/zizmor.yaml
@@ -38,4 +38,3 @@ jobs:
           advanced-security: false
           annotations: true
           inputs: .github
-          min-severity: high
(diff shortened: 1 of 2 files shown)


.github/workflows/zizmor.yaml
```diff
@@ -38,4 +38,3 @@ jobs:
           advanced-security: false
           annotations: true
           inputs: .github
-          min-severity: high
```
maks-rafalko, line 41: does it make sense to keep it explicitly with warning level? otherwise it's not obvious

theofidry: Done in https://github.com/infection/infection/pull/3185.

PR #3210, opened 2026-05-24
feat(telemetry): Add support for OpenTelemetry metrics

theofidry:
## Description

Adds OpenTelemetry metrics support to Infection telemetry, alongside the existing tracing support. Metrics remain opt-in through `INFECTION_TELEMETRY=true` and an explicit `OTEL_METRICS_EXPORTER`, so enabling telemetry does not start exporting metrics unless the user configures a supported metrics exporter.

Note that unlike for the traces, Sentry does not support an OTEL endpoint for metrics. This means our testing, for now, will be limited to the e2e test and locally.

## Changes

- Adds `OpenTelemetryMetrics` to derive low-cardinality counters and histograms from completed telemetry spans.
- Wires metrics creation, no-op handling, shutdown, and exporter validation into `OpenTelemetryTracerFactory`.
- Adds a failing meter provider factory so invalid metrics exporter configuration fails clearly instead of silently
  degrading to no-op metrics.

## Related issues

- #3090


theofidry:
@coderabbitai full review

Issue #3120, opened 2026-05-11
feat: add phpstan infection to CI


staabm:
## Problem Statement

to catch breaking changes against custom mutators early I would suggest adding https://github.com/phpstan/build-infection to the infection CI pipeline.

that way we could detect earlier when things break.

## motivation

I tried updating the phpstan integration to 0.32.6 which doesn't work because of https://github.com/infection/infection/issues/3119

after that I now dropped PHP 8.2 and raised the min version to PHP 8.3.
after that I was able to update to 0.32.7 and get a new error now :/

```
1) PHPStan\Infection\LooseBooleanMutatorTest::testMutator with data set "It mutates isTrue() into loose comparison" ('<?php\n\n$type->isTrue()->yes();', '<?php\n\n$type->toBoolean()->...yes();')
Failed asserting that the number of code samples (1) equals the number of mutants (0) created by the mutator. Make sure mutator is enabled and mutates the source code. Mutants are: 
Failed asserting that actual size 1 matches expected size 0.

/home/runner/work/build-infection/build-infection/vendor/infection/infection/src/Testing/BaseMutatorTestCase.php:115
/home/runner/work/build-infection/build-infection/tests/Infection/LooseBooleanMutatorTest.php:19

2) PHPStan\Infection\LooseBooleanMutatorTest::testMutator with data set "It mutates isFalse() into loose comparison" ('<?php\n\n$type->isFalse()->yes();', '<?php\n\n$type->toBoolean()->...yes();')
Failed asserting that the number of code samples (1) equals the number of mutants (0) created by the mutator. Make sure mutator is enabled and mutates the source code. Mutants are: 
Failed asserting that actual size 1 matches expected size 0.

/home/runner/work/build-infection/build-infection/vendor/infection/infection/src/Testing/BaseMutatorTestCase.php:115
/home/runner/work/build-infection/build-infection/tests/Infection/LooseBooleanMutatorTest.php:19

...
```

see https://github.com/phpstan/build-infection/pull/79

I guess something in `BaseMutatorTestCase` broke in 0.32.7

theofidry:
That's a good idea

theofidry:
@staabm as for the actual issue, isn't it related to https://github.com/infection/infection/pull/2832 ?

staabm:
Thanks to Théo the tests in PHPStan custom mutators were fixed with https://github.com/phpstan/build-infection/pull/81

[...]

PR #3219, opened 2026-05-26
ci: Add PHPStan build to the Infection pipeline

theofidry:
See the discussion in #3120.


theofidry:
cc @staabm

staabm:
in case you have a minute left, it would be great if you would fix the build on phpstan-build-infection (we can raise php min version to 8.3 to ease the upgrade)

.github/workflows/e2e-tests.yaml
```diff
@@ -149,13 +149,82 @@ jobs:
           TERM: xterm-256color
         run: make test-e2e E2E_PHPUNIT_GROUP=e2e INFECTION=dist/infection.phar BENCHMARK_SOURCES=
 
+  phpstan-build-infection:
```
maks-rafalko, line 152: could you please explain why do we have this job and a standalone `phpstan-build-infection.yaml` file?

theofidry: ah, I meant to remove `phpstan-build-infection.yaml`

.github/workflows/e2e-tests.yaml
```diff
@@ -149,13 +149,82 @@ jobs:
...
+          key: composer-${{ runner.os }}-phpstan-build-infection-${{ hashFiles('build-infection/composer.*', 'composer.*') }}
+          restore-keys: |
+            composer-${{ runner.os }}-phpstan-build-infection-
+            composer-${{ runner.os }}-
+            composer-
+
+      - name: Configure local Infection dependency
+        working-directory: build-infection
+        run: |
+          composer config repositories.local-infection '{"type":"path","url":"..","options":{"symlink":true,"versions":{"infection/infection":"0.32.99"}}}'
```
maks-rafalko: I guess this is why we have build failure? Should it be

```suggestion
          composer config repositories.local-infection '{"type":"path","url":"..","options":{"symlink":true,"versions":{"infection/infection":"0.33.99"}}}'
```

?

theofidry: probably... I don't really like that it's buried there... I would prefer if it picked the latest regardless

staabm: Couldn't you require latest dev-master instead of a fixed version?

theofidry: I could indeed

staabm (approved):

.github/workflows/e2e-tests.yaml
```diff
@@ -149,13 +149,83 @@ jobs:
...
+          key: composer-${{ runner.os }}-phpstan-build-infection-${{ hashFiles('build-infection/composer.*', 'composer.*') }}
+          restore-keys: |
+            composer-${{ runner.os }}-phpstan-build-infection-
+            composer-${{ runner.os }}-
+            composer-
+
+      - name: Configure local Infection dependency
+        working-directory: build-infection
+        run: |
+          composer config repositories.local-infection '{"type":"path","url":"..","options":{"symlink":true,"versions":{"infection/infection":"dev-master"}}}'
```
staabm, line 201: having dev-master here, might mean it only makes sense to run this job after merge on the master branch?

I guess we will not be able to define a version here, which fits/reflects a in-progress PR

maks-rafalko: or, do something like here?

https://github.com/infection/infection/blob/d9331f9a3b04ad03048e253ece3acf8a01710937/tests/e2e/Without_Git/run_tests.bash#L27

but this test is one of the most annoying one 😄

maks-rafalko (approved):

.github/workflows/e2e-tests.yaml
theofidry: Hmm I don't think? It means we replace whatever version PHPStan wanted to `dev-master` and then we map it to the local version

PR #3227, opened 2026-05-26
refactor(autoreview): Move tests are final checks to PHPAt

theofidry:
With [PHPAt](https://www.phpat.dev/) being introduced via https://github.com/infection/infection/pull/3084, we can now take it a step further and leverage it for existing architecture checks we have been doing with plain PHPUnit. The advantage of this, is that checks can be done more reliable (less home-made code operating on finicky regexes) and we can also leverage the PHPStan baseline.

This PR moves the check verifying that test classes are final unless they are traits or abstract (i.e. not applicable).


sanmai (approved):
Nice!

PR #3228, opened 2026-05-26
refactor(autoreview): Make classes final when possible

theofidry:
## Summary

Leverages [PHPat](https://www.phpat.dev/) being introduced via https://github.com/infection/infection/pull/3084 to check if classes that implement an interface are marked as final.

## Changes

- Add the `InterfaceImplementationsShouldBeFinalTest` test.
- `ParallelProcessRunner` can no longer have its `@final` (otherwise PHPStan does understand it as `final`) when it is not. So a dedicated test has been added for it.
- Make the classes `final` when easy to do so. Note that rector updates them to be `readonly` too afterwards in some cases.
- Left cases that were not trivial alone, an entry was added to the PHPStan baseline instead.

### DIFF
--- a/tests/phpunit/StaticAnalysis/PHPStan/Adapter/PHPStanAdapterTest.php
+++ b/tests/phpunit/StaticAnalysis/PHPStan/Adapter/PHPStanAdapterTest.php
@@ -35,8 +35,8 @@ declare(strict_types=1);
 
 namespace Infection\Tests\StaticAnalysis\PHPStan\Adapter;
 
+use Infection\Mutant\MutantExecutionResultFactory;
 use Infection\StaticAnalysis\PHPStan\Adapter\PHPStanAdapter;
-use Infection\StaticAnalysis\PHPStan\Mutant\PHPStanMutantExecutionResultFactory;
 use Infection\StaticAnalysis\PHPStan\Process\PHPStanMutantProcessFactory;
 use Infection\TestFramework\CommandLineBuilder;
 use Infection\TestFramework\VersionParser;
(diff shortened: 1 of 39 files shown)


tests/phpunit/TestFramework/Coverage/CoverageChecker/CoverageCheckerTest.php
```diff
@@ -296,9 +295,6 @@ public function test_it_does_not_pass_existence_check_if_junit_file_is_missing_w
 
     public function test_it_does_not_pass_existence_check_if_junit_file_is_missing_with_junit_report_with_phpunit_test_framework_adapter(): void
     {
-        $phpUnitAdapterMock = $this->createMock(PhpUnitAdapter::class);
```
theofidry: this change looks suspicious, to extract

tests/phpunit/StaticAnalysis/StaticAnalysisToolFactoryTest.php
```diff
@@ -51,7 +51,7 @@ public function test_it_throws_an_exception_if_it_cant_find_sa_tool(): void
         $factory = new StaticAnalysisToolFactory(
             ConfigurationBuilder::withMinimalTestData()->build(),
             $this->createStub(StaticAnalysisToolExecutableFinder::class),
-            $this->createStub(StaticAnalysisConfigLocator::class),
+            new StaticAnalysisConfigLocator(__DIR__),
```
theofidry: to extract this change, `StaticAnalysisToolFactory` should use `TestFrameworkConfigLocatorInterface` instead here

tests/phpunit/StaticAnalysis/PHPStan/Adapter/PHPStanAdapterTest.php
```diff
@@ -241,7 +241,7 @@ public function test_it_rejects_invalid_versions(string $version): void
     {
         $adapter = new PHPStanAdapter(
             $this->createStub(Filesystem::class),
-            $this->createStub(PHPStanMutantExecutionResultFactory::class),
```
theofidry: kinda unrelated, to extract

PR #3233, opened 2026-05-27
feat(autoreview): Introduce re-usable PHPat selectors

theofidry:
## Description

This PR introduces `InfectionSelector` to complement PHPat's native `Selector` class. The aim is to make the PHPat tests more readable and reliable while simplifying future refactoring.

### New Selector Framework

- **`InfectionSelector`**: Acts as a proxy to custom or preconfigured selectors. It provides a simple, readable API with an experience similar to PHPat's native `Selector`.
- **`InfectionCode`**: Selector matching all project code controlled by Infection, excluding vendor code and source code used as benchmark fixtures.
- **`InfectionSourceCode`**: Selector matching code found in `src/`.
- **`InfectionTestCode`**: Selector matching test code found in `tests/`.
- **`IsAnonymousClass`**: Selector identifying anonymous class reflections.
- **`ClassReflectionAccessor`**: Helper for safely accessing reflection metadata.

### Test Infrastructure

- **`SelectorTestCase`**: Abstract base class for selector tests. Initialises PHPStan container and provides utilities to create `ClassReflection` instances, including support for anonymous classes from fixture files.
- Comprehensive unit tests for each new selector class with multiple test cases covering expected matches and exclusions.

### Configuration Updates

- Exclude this code from Mago analysis. Mago is not aware of PHPStan's PHAR, and there does not appear to be a straightforward way to expose PHPStan's public API to it. It is therefore simpler to avoid scanning this code with Mago, so the baseline is not inflated unnecessarily.
- Adapt the PHPUnit configuration so `Architecture` is handled in the same way as `AutoReview`: executed through a separate configuration.

## Reviewer Notes

Note that this is an (internal) feature, not a refactor. Indeed, whilst the _intent_ of the currently used selectors is the same, there is slight changes. For example some code from `tests/` could depend on `src/` without being caught by `SrcShouldNotDependOnTestsTest`.

Basic tests were included for the selectors. They were relatively easy to generate and adapt. They are limited in scope and are not sufficient to guarantee that the PHPat rules work as expected. However, they should be enough to catch the most significant issues.

I am not certain what would be required to add more complete tests, but it would likely require significantly more infrastructure, which would be harder and more expensive to maintain. I would therefore advise against it. Ultimately, this supports quality-rule enforcement. If it remains somewhat limited, as is already the case for the current regular AutoReview tests, that is acceptable. The goal is to pursue low-cost improvements to the IDX, not to provide an expensive, exhaustive solution.

### DIFF
--- a/mago.toml
+++ b/mago.toml
@@ -16,11 +16,12 @@ includes = ["vendor"]
 excludes = [
     "src/FileSystem/DummyFileSystem.php",
     "src/CustomMutator/templates",
-    "tests/Architecture",
     "tests/benchmark/MutationGenerator/sources",
     "tests/benchmark/Tracing/benchmark-source",
     "tests/benchmark/Tracing/coverage",
     "tests/benchmark/Tracing/sources",
+    "tests/Architecture",
+    "tests/phpunit/Architecture",
     "tests/phpunit/Fixtures",
     "tests/phpunit/Command/Debug/DumpAstCommand/Greeter.php",
     "tests/phpunit/Command/Debug/DumpAstCommand/EchoGreeter.php",
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -28,6 +28,7 @@
         <testsuite name="Infection Test Suite">
             <directory>tests/phpunit</directory>
 
+            <exclude>tests/phpunit/Architecture</exclude>
             <exclude>tests/phpunit/AutoReview</exclude>
             <exclude>tests/phpunit/Fixtures</exclude>
             <exclude>tests/phpunit/TestFramework/Coverage/Fixtures</exclude>
--- a/phpunit_autoreview.xml
+++ b/phpunit_autoreview.xml
@@ -22,12 +22,15 @@
 
     <testsuites>
         <testsuite name="Auto-Review Test Suite">
+            <directory>tests/phpunit/Architecture</directory>
             <directory>tests/phpunit/AutoReview</directory>
         </testsuite>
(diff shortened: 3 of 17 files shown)


theofidry:
@coderabbitai full review

tests/Architecture/PHPat/SrcShouldNotDependOnTestsTest.php
```diff
@@ -44,19 +44,10 @@ final class SrcShouldNotDependOnTestsTest
...
             ->shouldNot()
             ->dependOn()
-            ->classes(
-                Selector::AnyOf(
-                    Selector::inNamespace('Infection\\Tests'),
-                    Selector::inNamespace('Infection\\Benchmark'),
-                ),
-            )
-            ->because('Production code under src/ must not depend on tests/ or benchmarks.');
+            ->classes(InfectionSelector::testCode())
```
maks-rafalko, line 50: on the left, we also had benchmark. Is it a mistake? or why the logic has changed?

maks-rafalko (approved):

tests/Architecture/PHPat/SrcShouldNotDependOnTestsTest.php
theofidry: I'm not sure I understand that comment correctly.

But `testCode()` contains `Infection\Benchmark\*` classes that are not fixtures, i.e. code like `Instrumentor` are part of the test code, but code like `MutationGeneratorBench` is not

PR #3235, opened 2026-05-29
refactor(autoreview): Ignore Trivial Implementations in Missing-Test Architecture Rules

theofidry:
## Description

This PR narrows the PHPat architecture rules that require concrete classes to have canonical tests. Classes with trivial implementations are now excluded, so fake, null-object, unsupported-method, and constant-returning classes do not need dedicated tests solely to satisfy the rule.

## Motivations

I find the rule that detects missing tests very useful. However, one recurring friction is that we have a number of classes that are simply not interesting to test. A non-exhaustive list:

- Plain value objects that only list properties and have no behaviour. Most events follow this pattern, but it is not limited to them.
- Null, fake and dummy implementations.

Before this PR, and before moving the autoreview rule from PHPUnit to PHPat, we had to keep extending `ProjectCodeProvider::NON_TESTED_CONCRETE_CLASSES`, which was quite irritating.

## Changes

- Adds a `HasTrivialImplementation` selector and wires it into the source and PHPUnit test-support missing-test architecture rules.
- Adds an analyser that reads a concrete class file, parses it, and traverses the AST to detect meaningful method implementation.
- Treats empty classes, declarations-only classes, simple scalar/static returns, and throw-only unsupported-method implementations as trivial.
- Adds focused analyser and visitor tests with physical fixture files for PHPStan-backed class reflection.

## Reviewer Notes

As you may notice, I am introducing `Analyser`, which adds AST-based file analysis. The current implementation is relatively simple, so false positives may arise, but I believe this should provide a better experience overall.

I am also planning to port IO usage detection to this analyser in the future. That should make usage detection simpler and more reliable, reducing the number of false positives.


composer.lock
```diff
@@ -5712,30 +5712,30 @@
         },
         {
             "name": "webmozarts/strict-phpunit",
-            "version": "7.15.5",
+            "version": "7.24.1",
```
theofidry, line 5715: this was necessary otherwise due to some PHPStan shenanigans, `InstalledVersions::getPrettyVersion('phpunit/phpunit')` was returning 9.x from somewhere... Which caused `webmozarts/strict-phpunit` to use incorrect code resulting in an exception. This code was there to have PHPUnit 10+ & 9 compatibility, which is no longer present in that newer version.

theofidry (commented):
Disclaimer: I leverage a lot more than usual Codex for this PR, notably for the visitor. I did intervene quite a bit to ensure the implementation is not complete non-sense with nested while loops, but I also have not been as careful as I usually am.

I do not think it is a particularly critical piece of code so as long as the tests are fine, I think the implementation of the visitor matters little.

maks-rafalko (approved):

Issue #3138, opened 2026-05-14
bug(testFramework): PHPUnit adapter mis-parses test framework options


theofidry:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.32.7
| Test Framework version | All
| CI | either
| PHP version | any
| Platform    | any
| Github Repo | https://github.com/infection/infection |

While expanding PHPUnit adapter command-line scenario coverage in https://github.com/infection/infection/pull/3133, several broken or ambiguous transformations were identified.

The affected input is `testFrameworkOptions` from `infection.json5`, or the equivalent `--test-framework-options` CLI option. The same parsing path is used when building both the initial PHPUnit command and mutant PHPUnit commands.

This issue may also affect the other test framework adapters.

## Current behaviour

Examples from PR #3133:

- Short PHPUnit options are rewritten as long options: `-v` becomes `--v`.
- Space-separated option values are kept as a single raw token: `--filter "a test with spaces"` becomes one token containing the quotes, rather than `--filter` plus `a test with spaces`.
- Quoted values keep their shell quotes: `--filter="a test with spaces"` still contains the quote characters in the generated argv token.
- Option values containing ` --` are split as if the value introduced a new option: `--filter="a test -- with option-like text" --group=default` is split inside the filter value.
- Positional PHPUnit arguments are rewritten as long options: `tests/FooTest.php --filter=Foo` becomes `--tests/FooTest.php --filter=Foo`.
- Quoted positional arguments have both problems: `"tests/Foo Test.php"` becomes `--"tests/Foo Test.php"`.
- User-provided filters are not detected consistently when Infection also generates class-to-test filters: `--filter Foo` and `--filter=Foo` do not prevent the generated `--filter FooTest|BazTest` from being appended.
- Repeated spaces are preserved in option tokens: `--group=default  --filter=Foo` leaves a trailing space on the `--group` token.
- Tabs and newlines are not treated as argument separators: `--group=default\t--filter=Foo` and `--group=default\n--filter=Foo` remain single tokens.
- Other options with separated values have the same issue, for example, `--testsuite "Unit Tests"` and `--configuration custom-phpunit.xml`.

## Expected behaviour

Infection should preserve the user-provided PHPUnit command-line semantics when building PHPUnit command arrays:

- short options remain short options;
- quoted values are parsed into argv values without shell quote characters;
- option-like text inside quoted values is preserved as value text;
- positional PHPUnit arguments remain positional arguments;
[...]

PR #3254, opened 2026-05-31
fix(process): Return Composer commands as argv tokens

### LINKED ISSUE #3253: bug: Invalid Composer executable usage
| Question    | Answer
| Infection version | 0.33.2
| CI | ?
| PHP version | any
| Platform    | any
| Github Repo | -

`ComposerExecutableFinder::find()` returns a `string`. In practice, it means we could end up with `php /path/to/composer.phar`. This code is used in `TestFrameworkFinder` like so:

```php
try {
    $process = new Process([
        ...$this->findComposer(),
        'config',
        'bin-dir',
    ]);

    $process->mustRun();
    $vendorPath = trim($process->getOutput());
} catch (RuntimeException) {
    $candidate = getcwd() . '/vendor/bin';

    if (file_exists($candidate)) {
        $vendorPath = $candidate;
    }
}
```

This means we are constructing the `Process` with the following arguments `['php /path/to/composer', 'config', 'bin-dir']`, which is incorrect. It should be:

```diff
[
-- 'php /path/to/composer',
++ 'php',
++ '/path/to/composer',
   'config',
   'bin-dir',
]`
```

Note that we have a similar issue in different places: https://github.com/infection/infection/issues/3138.

## Impact

In practice, it means Infection cannot ask Composer for `bin-dir`, then falls back to `$cwd/vendor/bin`. This only becomes user-visible if the tool is not discoverable through that fallback or existing `PATH`. This leads to the test framework or static analysis to fail, e.g.:

```
Infection\FileSystem\Finder\Exception\FinderException: Unable to locate a phpunit executable on local system. Ensure that phpunit is installed and available.
```

So the case of failures are not very common:

- An unconventional Composer bin directory.
- An execution outside of the usual working directory.

But they do exist.

theofidry:
## Description

This PR changes Composer executable resolution to return an argv prefix instead of a single string. This avoids treating commands such as `php /path/to/composer.phar` as one executable name when they are passed to Symfony Process in array mode.

The previous `string` contract was ambiguous: it could represent either an executable path or a complete command fragment. Returning `list<string>` lets callers append Composer arguments safely while preserving the correct executable and script tokens.

## Changes

- Updates `ComposerExecutableFinder::find()` to return `list<string>`.
- Updates the concrete and memoized Composer finders to preserve tokenised Composer commands.
- Spreads the Composer command prefix in test framework, static analysis, adapter installation, and E2E Composer process creation.

## Related issues

Fixes https://github.com/infection/infection/issues/3253.

### DIFF
--- a/devTools/mago-baseline.toml
+++ b/devTools/mago-baseline.toml
@@ -5071,18 +5071,6 @@ message = "Assigning `mixed` type to a variable may lead to unexpected behavior.
 count = 3
 
 [[issues]]
-file = "tests/benchmark/Tracing/profile.php"
-code = "redundant-docblock-type"
-message = "Redundant docblock type for variable `$maxTraceCount`."
-count = 1
-
-[[issues]]
-file = "tests/benchmark/Tracing/profile.php"
-code = "redundant-docblock-type"
-message = "Redundant docblock type for variable `$percentage`."
-count = 1
-
-[[issues]]
 file = "tests/phpunit/AutoReview/BuildConfigYmlTest.php"
 code = "imprecise-type"
 message = "Type `iterable` in return type of `providesYamlFilesForTesting` is imprecise, equivalent to `iterable<mixed, mixed>`."
@@ -7342,6 +7330,18 @@ count = 5
 file = "tests/phpunit/FileSystem/Finder/TestFrameworkFinderTest.php"
 code = "string-member-selector"
 message = "This member selector uses a non-literal string type (`string`); its specific value cannot be statically determined."
+count = 1
+
+[[issues]]
+file = "tests/phpunit/FileSystem/Finder/TestFrameworkFinderTest.php"
+code = "unhandled-thrown-type"
+message = 'Potentially unhandled exception `Symfony\Component\Filesystem\Exception\IOException` in `Infection\Tests\FileSystem\Finder\TestFrameworkFinderTest::createComposerExecutableFixture`.'
+count = 1
+
+[[issues]]
+file = "tests/phpunit/FileSystem/Finder/TestFrameworkFinderTest.php"
+code = "unhandled-thrown-type"
+message = 'Potentially unhandled exception `Symfony\Component\Filesystem\Exception\IOException` in `Infection\Tests\FileSystem\Finder\TestFrameworkFinderTest::createPhpUnitExecutableFixture`.'
 count = 1
 
 [[issues]]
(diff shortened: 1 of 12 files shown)


maks-rafalko:
from the related issue:

```diff
[
-- 'php /path/to/composer',
++ 'php',
++ '/path/to/composer',
   'config',
   'bin-dir',
]`
```

could you please explain, how does it lead to incorrect executable being used?

maks-rafalko:
> That breaks for PHAR and non-executable Composer files, because `Process` does not split array elements on spaces.

I think I don't follow this sentence, so appreciate the explanation.

theofidry:
@maks-rafalko I removed it, the phrasing is too confusing. The impact and broken cases are detailed in https://github.com/infection/infection/issues/3253.

maks-rafalko:
> The impact and broken cases are detailed in #3253.

I've read it a couple of times before asking 😄 and still don't get what is the issue, so let me ask step by step

1. with previous `'php /path/to/composer',` string, do we have an exception that is caught in `catch` and `$candidate = getcwd() . '/vendor/bin';` is executed as a fallback?
2. if yes, what is the original exception? why it fails? what is the message of the exception you are experiencing?
3. do you have something other than `vendor/bin` as a "bin" directory of composer?

theofidry:
>1) with previous 'php /path/to/composer', string, do we have an exception that is caught in catch and $candidate = getcwd() . '/vendor/bin'; is executed as a fallback?

Exactly.

> 2) if yes, what is the original exception? why it fails?

If you type `php /path/to/composer` in your shell, it will be parsed as two argv arguments: `php` and `/path/to/composer`. By passing `'php /path/to/composer'` though, what you are doing is passing _one_ argument, so it is as if you did `php\ /path/to/composer` (notice the space escape).

You check for yourself very easily actually:

```shell
which php
which composer
```

then:

```php
$validProcess = new Process(['<php>', '<composer>']);
$invalidProcess = new Process(['<php> <composer>']);

$validProcess->mustRun(); // works
$invalidProcess->mustRun(); // fails, ProcessFailedException "the command ... failed"
```

>do you have something other than vendor/bin as a "bin" directory of composer?

I do not no.

_But_, I vaguely remember seeing some weirdness due to not having the right working directory or using a specific path to a PHAR. I can't remember exactly the case. I could already pinpoint the issue to `TestFrameworkFinder` but didn't have the energy or motivation to dig further.

It's only stuck me now with https://github.com/infection/infection/issues/3138.

maks-rafalko (approved):
Thank you 👍

PR #3255, opened 2026-05-31
refactor(autoreview): Add PHPat rules for the events architecture

theofidry:
## Description

Adds PHPat architecture coverage for `src/Event/Events` so event classes and their single-event subscriber interfaces stay paired and follow the expected shape.

The rules make the event model explicit:

- Files under `src/Event/Events` should be either event classes or single-event subscriber interfaces.
- Each event should have a matching subscriber interface.
- Each subscriber interface should point back to an existing event.
- Subscriber methods should follow the `on<EventName>(<EventName> $event): void` convention.

With those rules, the PHPUnit AutoReview ones (for events) are now obsolete. I would not call this a migration as the AutoReview were a lot more limited:

- Checking that the method named was `on<EventName>(<EventName> $<notChecked>): <notChecked>`.
- Checking that the event subscriber implementers followed the naming convention for their method names. This has been enforced by using interface, which was introduced in https://github.com/infection/infection/pull/2889.

## Changes

- Registers `EventClassesShouldFollowConventionsTest` as a PHPat/PHPStan architecture test.
- Adds various selectors.
- Adds reusable event-architecture selector support.

## Reviewer Notes

Note that I chose the term "single event". This is because we have two kinds of subscribers:

- The subscribers for each events (the ones I call "single event"), which have very clear conventions.
- Other subscribers which are consumers of those "single event", e.g. `PerformanceLoggerSubscriber` which implements multiple (single event) subscribers.

Also note that whilst this is significantly more code that the old AutoReview rules, the extend of the checks, as well as their test coverage, is much higher.


sanmai (approved):
Looks like a lot of unit tests 👍

PR #3108, opened 2026-05-07
feat(telemetry): Add OpenTelemetry observability support

theofidry:
## Description

This PR introduces an [OpenTelemetry][otel]-based observability framework. The feature is opt-in and requires `INFECTION_TELEMETRY=true` to be enabled.

The motivation is detailed in RFC #3090.

### Introduction to OpenTelemetry

[OpenTelemetry][otel] is an open source observability framework. It defines a standard set of primitives that observability tooling can consume, allowing an application to support various backends such as [Sentry][sentry], [DataDog][datadog] or [NewRelic][newrelic] without integrating against each vendor's SDK individually.

OpenTelemetry organises data into [Signals][signals], the units exported to a backend. The three main ones are:

- [Traces][traces]: capture the path of a request (CLI or HTTP) through the application, providing a picture of what happened during execution. The level of detail depends on the implementation. A trace is composed of [spans][spans] to which attributes are attached at the implementer's discretion. [Our traces are available here][infection-traces].
- [Metrics][metrics]: measurements captured at runtime. Unlike traces, which aim to give a detailed picture, metrics aggregate indicators and typically have a longer retention period. Our Sentry instance does not currently accept OTLP metrics, but this PR still adds support so that users running a compatible backend can benefit.
- [Logs][logs]: traditional application logs.

### OpenTelemetry library

This PR uses the [official OpenTelemetry SDK][otel-sdk]. @norberttech has raised concerns about this choice and the selection is not final. The official SDK was chosen because it supports the full `OTEL_*` environment variable surface; relying on an alternative would have made it harder to distinguish genuine issues from gaps in that library's implementation.

### Usage

Configuration is driven entirely by environment variables. Most OpenTelemetry tooling honours the standard `OTEL_*` variables, and this implementation follows the same convention.

Because those variables are universal, exporting them in a shell session could inadvertently enable telemetry across multiple applications. To guard against accidental activation, a common pattern is to require an application-specific opt-in: in this case, `INFECTION_TELEMETRY=true`. This variable only controls activation; telemetry can still be disabled through other mechanisms (for example, `OTEL_SDK_DISABLED=true`).

An example:

```shell
INFECTION_TELEMETRY=true \
    OTEL_TRACES_EXPORTER=console \
    bin/infection --no-interaction --quiet --no-progress
```

This outputs traces as JSON. The trace span hierarchy looks as follows:

```
- `infection.run` (root span; covers the full execution)
    - `infection.artefact_collection`
        - `infection.initial_tests`
[...]

PR #3258, opened 2026-06-02
refactor(autoreview): Better I/O detection

theofidry:
## Description

Moves PHPUnit integration-group I/O detection from the old AutoReview string scanner into the PHPat selector analyser. The new analyser parses class ASTs, evaluates both test code and covered source code, and detects I/O through native filesystem and network functions, selected XML reader/writer factory methods, filesystem instantiation, and the test `FS` utility.

This makes the `#[Group('integration')]` architecture rules less dependent on broad text matches, such as merely importing `Filesystem`, and allows mocked filesystem dependencies to be treated as non-I/O.

## Changes

- Migrates `IoCodeDetector` under the PHPat selector analyser support namespace.
- Extends `Analyser` and `AnalysisResult` so a single parse pass can report both trivial implementation and I/O usage.
- Updates `PHPUnitTestIoRequirements` to analyse test classes, parent test classes, covered source classes, and parent source classes through the shared analyser.
- Updates PHPat result-cache metadata to track the selector support files instead of the removed detector file.
- Updates integration-group selector fixtures and tests for the new I/O semantics, including filesystem mocks, real filesystem instantiation, `FS::tmpFile()`/`FS::tmpDir()`, Safe functions, grouped imports, and `FileSystemTestCase` children.
- Updates PHPStan and Mago baselines after moving/removing the old detector and changing PHPat rule outcomes.

**This PR does not fix the baseline.** Tests that have a redundant `#[Group('integration')]` attribute are not updated in this PR to keep it focused. They will be fixed separately.

## Reviewer Notes

Although this PR covers `Filesystem`, similar logic could be applied to other abstractions such as `Process` and `HttpClient`. This is intentionally left for a dedicated PR if we decide to improve that area.


devTools/phpstan-baseline.neon
```diff
@@ -762,27 +762,33 @@ parameters:
 			count: 1
 			path: ../src/Testing/SingletonContainer.php
 
+		-
+			rawMessage: 'Property Infection\Tests\Architecture\PHPat\Selector\Support\PHPUnitTestIoRequirements::$classUsesIoCache (array<class-string, bool>) does not accept array<string, bool>.'
```
theofidry: this should be addressed

devTools/phpstan-baseline.neon
```diff
@@ -762,27 +762,33 @@ parameters:
...
+			rawMessage: Infection\Tests\Architecture\PHPat\Selector\Support\PHPUnitTestIoRequirementsTest should not exist
 			identifier: phpat.testPHPUnitTestsRequiringIoBelongToIntegrationGroup
 			count: 1
-			path: ../tests/phpunit/Architecture/PHPat/Selector/PHPUnitTestSupportConcreteClassWithoutCanonicalTestTest.php
+			path: ../tests/phpunit/Architecture/PHPat/Selector/Support/PHPUnitTestIoRequirementsTest.php
 
 		-
-			rawMessage: Infection\Tests\Architecture\PHPat\Selector\Support\Analyser\AnalyserTest should not exist
-			identifier: phpat.testPHPUnitTestsRequiringIoBelongToIntegrationGroup
+			rawMessage: 'Parameter #1 $analyser of class Infection\Tests\Architecture\PHPat\Selector\Support\PHPUnitTestIoRequirements constructor expects Infection\Tests\Architecture\PHPat\Selector\Support\Analyser\Analyser, Infection\FileSystem\FileSystem given.'
```
theofidry: should be addressed

theofidry (commented):
needs a more careful review and probably some minor rework

theofidry:
@coderabbitai full review

tests/Architecture/PHPat/Selector/Support/Analyser/IoCodeDetector.php
```diff
@@ -0,0 +1,528 @@
...
+    private function detectUseItem(UseItem $useItem): void
+    {
+        $type = $useItem->type === Use_::TYPE_UNKNOWN ? $this->currentUseDefaultType : $useItem->type;
+        $usedName = $this->getUsedName($useItem, $this->currentUsePrefix);
+
+        if ($type === Use_::TYPE_FUNCTION && $this->isIoFunctionName($usedName)) {
+            $this->hasIoOperations = true;
+
+            return;
+        }
```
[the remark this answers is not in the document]
theofidry: In practice, we do not allow unused functions (i.e. they are cleaned up by the CS). As a result, the only case where this may be true is if we were using the function-name à la class way, e.g. `ClassName::class`, i.e. a usage of the function that is other than invoking it. PHP does not allow it, and if eventually it does, the consequences are only a false positive that would affect maybe one file or two, not a big deal.

maks-rafalko (commented):
Probably I'm tired, but I have strange feelings with all these `PHPAt` PRs, as I really don't get why we need so many changes (many PRs, many changed files, too much new code to understand, which I didn't manage to do in this particular PR -> so I can't click an approve). 

Reminder, I look into all PRs, even the merged ones, so it's important for me to keep track and efficiently spend my time.

Questions I have in mind:

- what is the main driver and pain points we are trying to solve here?
- is it important now?
- do we need so many PRs and changes for the things that worked for years?

Don't get me wrong, I believe the changes are good and the direction is good, but I miss the point of **why** we need all of this.

Many of the things take me hours to read / review / trying to understand, so I have to ask it - is it (the whole sequence of changes) really needed? I may miss something, so apologies and would appreciate clarification.

tests/Architecture/PHPStan/ResultCache/PHPatArchitectureResultCacheMetaExtension.php
```diff
@@ -62,8 +60,7 @@ public function getHash(): string
             ->files()
             ->name('*.php')
             ->in(self::SELECTOR_DIRECTORY)
-            ->sortByName()
-            ->append([self::IO_CODE_DETECTOR_FILE]);
```
maks-rafalko, line 66: could you please explain what is this change for?

maks-rafalko (approved):

tests/Architecture/PHPStan/ResultCache/PHPatArchitectureResultCacheMetaExtension.php
theofidry: The class itself was introduced in https://github.com/infection/infection/pull/3261, which TLDR is we use custom selectors, and changing the code of those selectors did not necessarily trigger the PHPStan cache to go stale which meant the rules were not updated like they should.

Here we remove `tests/phpunit/AutoReview/IntegrationGroup/IoCodeDetector.php` completely, so this line is no longer needed.

theofidry (commented):
That's quite a feedback, so the answer is not short 😅 

>Probably I'm tired, but I have strange feelings with all these PHPAt PRs, as I really don't get why we need so many changes

Whilst I did not create a dedicated RFC for it, I did try to add an explanation as to why and its effect in the description of each PR. That said, maybe it was too many and it's easily to loose context when not following the whole train.

>what is the main driver and pain points we are trying to solve here?

We have AutoReview tests to codify conventions, to help maintaining boundaries, categorising tests (integration or not), defining if the code is part of the public API or internal, checking that implementation have tests.

They serve us well, there is value in those, but it's a lot of home-brewed code with its flaws, be it in the brittleness, accuracy or feedback. We recently introduced PHPat which is designed for this. So there is two goals here:

- Port rules, which should mean less home-brewed code, i.e. less maintenance for us.
- Add rules which otherwise we did not want to add because it meant more home-brewed code (e.g. recently you added `SrcShouldNotDependOnTestsTest` because it is trivial to do so with PHPat).

Other benefits IMO:

- Being able to improve some existing rules to reduce the amount of false-positives. False-positives can either contribute in the rule not operating efficiently (e.g. not spotting a file is missing a test) or in friction (e.g. asking for a test when it should not or categorising a test as integration when it should not) – An example of such an improvement is https://github.com/infection/infection/pull/3235.
- Violations are now plain PHPStan violations. Which means false positives should go to `phpstan.neon` or `// @phpstan-ignore`, and others that we don't want to address right at this moment can go to the baseline, with the other issues that should be fixed _eventually_. This is, IMO, a net improvement over `ProjectCodeProvider::MY_CONST_THAT_CONTAINS_FALSE_POSITIVES_AND_LEGITIMATE_STUFF_THAT_WE_WILL_NEVER_LOOK_AT`.

>is it important now?

Not really. I'm more interested in the test framework rework atm because I'm really keen on having those PHPUnit coverage improvements. However, the latter requires more careful attention.

Migrating PHPUnit tests to PHPat though, is for the most part, boring work trivialised by AI. It's easy to prompt and requires less time and thinking to review or test.

That comes with caveats, it's easy for slop to slip in, maybe not all the code is as readable as I would like. For the most part though, I think the result is pretty good, the code in the rules itself is quite expressive (much better than the PHPUnit ones at least) and it is a lot more tested.

>do we need so many PRs and changes for the things that worked for years?

That it worked for years doesn't mean it can be made better. For example right now, every time you add a event, a `Null*`, `Dummy*` or `Fake*` implementation, you will have the autoreview step telling you off that you're missing a test for them even when there is nothing to test. Regularly you need to correct a test to categorise it as `integration` when it is not because there is an IO function that slipped in somehow.

I'd also add that while it's a lot of PRs, I keep them relatively small and scoped for easy review and making it easy to trace back what the change was for. Like the question you just asked above, for which the answer was made easy.

Another point: while the PR diffs look like it's a lot more code added, most of it is tests/fixtures. A lot of the rules introduced could be "inlined", but personally find _a lot_ better to have an expressive and tested API, e.g. https://github.com/infection/infection/pull/3233, even if it's a bit boilerplaty.

maks-rafalko:
> That's quite a feedback, 

didn't want to hurt you in any way, just to be clear :) So apologies if it sounded wrong. 

> so the answer is not short 😅

Thank you for the explanations.

---

This

> > is it important now?
>
> Not really.

and this

> Migrating PHPUnit tests to PHPat though, is for the most part, boring work **trivialised by AI**.

are the major things led me to all the questions I asked in the previous message.

Context: I noticed that I spend literally half of the day looking into AI generated information (articles, code review on a primary job and here, code review comments from AI, the code generated by AI, Claude answers with _walls of texts_ where it could be written 2x-5x times shorter and so on) - so I'm trying to reduce all this load to the brain and free up some time to develop new features. What I end up though is reading many of the PRs with tens of files, which are not really important comparing to other features or bugs.

Again, _please_ don't get me wrong, everything you did does make sense, it's well scoped and split with many PRs (thank you), it improves the things. The problem is not in you, but probably in me not having enough time to absorb this, so I'm trying to prioritize (at least in my head) all of this, and I fail to do it right now.

The good example: I have a new feature in my branch already for 2 weeks, and each time I try to work on it, I see a list of many not-reviewed PRs that _I have to_ review to not loose the track. 

That's why I believe that such questions that I asked do make sense. Probably we need to prioritize the things across the whole bunch of issues/PRs to have it manageable and reduce the cognitive load / frustration.

theofidry:
>So apologies if it sounded wrong.

No offence taken nw.

>Context: [...]

I know the feeling. The AI-generated review fatigue is real, I feel it a lot as well.

Side note: my walls of texts are still very much hand-written, I tend to use AI to _reduce_ them, but human me is very good at writing walls 😅 

>Probably we need to prioritize the things across the whole bunch of issues/PRs to have it manageable and reduce the cognitive load / frustration.

My question there is: does this code needs your full attention?

For context, this is for AutoReview rules. I think what really matters here is if the general rule itself makes sense. Then there is two other sides to this:

- How annoying or costly this code is to maintain.
- How much friction this rule introduces, especially in the case of false-positives.

The actual implementation is otherwise irrelevant. In theory it could be a big pile of unreadable garbage slop that we can't maintain and decide to either reworking it if we want to keep it or ditch it entirely, this should not cause problems. To be clear though, although I feel a lot less attachment and care about those pieces of code, I do still pay a good deal of attention so that it's not any harder to maintain/update.

Because otherwise yes, it will be very distracting. Especially considering that even outside of the current feature you are working on, there is at least 2-3 RFCs that were not reviewed yet and quite a big feature (https://github.com/infection/infection/pull/3108) waiting for feedback; that's without the current test framework changes that I'm about to introduce

maks-rafalko:
> My question there is: does this code needs your full attention?

yeah, this is something new in AI era that I didn't think through. Probably, this particular code is, as you said, not quite important, agree, and shouldn't be reviewed line by line to completely understand it. Anyway, I've already did that and approved 😄.

> Side note: my walls of texts are still very much hand-written

yeah I see it 👍 

> Especially considering that even outside of the current feature you are working on, there is at least 2-3 RFCs that were not reviewed yet and quite a big feature (https://github.com/infection/infection/pull/3108) waiting for feedback; that's without the current test framework changes that I'm about to introduce

don't you want to have a vacation? 🤣 (joking..)

Ok, let's get it merged. Thank you.

Issue #3082, opened 2026-04-22
RFC: Test Framework and Version-specific Tracing


theofidry:
## Summary

This RFC replaces [#2816][issue-2816-coverage-rfc]. The latter was shaped and motivated by the work done in [#2543][issue-2543-xml-rework]. Whilst there is merit in improving the coverage-related API and some pieces of [#2816][issue-2816-coverage-rfc] may still be necessary, this RFC concludes that the API proposed in [#2816][issue-2816-coverage-rfc] is the wrong abstraction and will not solve the fundamental problem.

This RFC proposes a different direction to solve the main remaining issue, which is to **enable an easier adoption of new coverage formats**.

## Motivations

Coverage report parsing and the information derived from it has been a relatively stable part of Infection in recent years. Whilst stable, improvements are still possible but are difficult to accommodate within the current architecture.

Following discussions with Sebastian Bergmann, several features have landed in the code coverage tooling:

- [PHPUnit no longer ships source code in the XML report, making it significantly lighter.][pr-2604-xml-lighter]
- PHPUnit added the `hash` property to the `file` element to support integrity checks.
- PHPUnit added the `time` (execution time) attribute to the `test` element, making JUnit obsolete.

In the context of [issue #2600][issue-2600-coverage-feedback], Sebastian Bergmann expressed openness to further changes, up to and including an entirely different code coverage report format.

The recent request to support [Testo][testo] reinforces this: it may well require a coverage format with no overlap with the current PHPUnit-based model.

**This demonstrates the need for test-framework-version-specific tracer implementations.**

## Proposed Solution

There is a very simple solution, enabled by the `Tracer` introduced in [#2751][issue-2751-tracer].

`Tracer` and `TestLocation` form the domain-facing API for consuming test-related information. Everything behind that boundary – JUnit, XML coverage, `TestFileDataProvider`, and so on, is an implementation detail.

This means we could simply use a different `Tracer` implementation when desired. While the idea is simple, there are a few practical considerations.

### Introduce framework-specific tracers

Each adapter should own a `Tracer` implementation responsible for translating its own coverage artefacts into Infection's domain objects:

```diff
interface TestFrameworkAdapter {
    // ...
    
--  public function hasJUnitReport(): bool;
++  public function getTracer(): Tracer;
}
```

While this API change is simple, it has further consequences, for example the factories invoked in `Infection\TestFramework\Factory` may require different arguments.

#### Version detection and tracer selection

[...]

PR #3269, opened 2026-06-04
feat(autoreview): Codify adapter package boundaries with PHPat

theofidry:
## Description

This PR contributes to https://github.com/infection/infection/issues/2863 by establishing
the boundaries of the future packages and enforcing them with PHPat.

Updating `infection/abstract-testframework-adapter` is sensitive because of its downstream impact. This PR therefore keeps as much preparatory work as possible in the main repository first. The PHPUnit adapter, the static analysis contract, and the Mago and PHPStan adapters currently live in this repository, so we can make a meaningful part of the transition here before updating and extracting packages.

This PR proposes the following:

- Changes planned for `Infection\AbstractTestFramework\*` will be located in `Infection\TestFramework\Contracts`. It should not depend on code from the main Infection application, only vendor code*.
- The PHPUnit adapter will still live under `Infection\TestFramework\PhpUnit`.
- The PHPStan and Mago adapters will live under `Infection\TestFramework\PhpStan` and `Infection\TestFramework\Mago`, respectively. This follows the goal described in https://github.com/infection/infection/issues/2863 to model the static analysis contract and adapters as regular test frameworks.
- `Infection\TestFramework\Common` will contain low-dependency utilities that can be shared across adapters. The first two candidates are `Infection\TestFramework\CommandLineBuilder` and `Infection\TestFramework\VersionParser`, which are currently copied in each adapter.

*: note that this is true only because none of the adapters or adapter utilities from the future packages are installed as dev dependencies. Otherwise, this rule should be tightened, for example `vendor` minus `infection/*`, plus a small set of explicitly selected `infection/*` packages.

This PR focuses on establishing the rules, not fixing the violations they introduce. Even simple violations, such as `CommandLineBuilder` or `VersionParser` requiring only a namespace change, will be addressed in future PRs.

## Changes

- Adds `AdapterExtractionBoundariesTest` to enforce future package boundaries with PHPat.
- Adds selector helpers for future adapter, contract, common utility, and non-source dependency groups.
- Adds `ClassNamedAny` for selecting individual shared classes that are not grouped by namespace.
- Adds unit coverage for `ClassNamedAny`.
- Registers the new PHPat test in `devTools/phpstan.neon`.
- Updates the PHPStan baseline with the existing package-boundary violations.
- Documents the intended future package ownership for `TestFramework/Common`, `TestFramework/Contracts`, `TestFramework/Mago`, and `TestFramework/PhpStan`.

## Related issues


maks-rafalko:
I'm not sure I follow the idea with the current explanation.

- what is the final goal of this change?
- what are we going to do with ignored (in baseline) errors? are we going to fix them? if not - why do we need them to be in the baseline? I don't think we are going to extract PHPUnit, for example

the diagram without explanation is not clear for me, to be honest

theofidry:
>what is the final goal of this change?

The goal is the RFC.

>what are we going to do with ignored (in baseline) errors? are we going to fix them? if not - why do we need them to be in the baseline? I don't think we are going to extract PHPUnit, for example

The goal is to fix them yes. They show for instance that if tomorrow we want to extract the PHPStan or other static analysis adapter, to allow one to introduce their own, we would not be able to in the current form.

Likewise there is several things here that are PHPUnit specific, or bound to it, that should not. For example the tracer (https://github.com/infection/infection/issues/3082), or some utilities that are otherwise copy/pasted (e.g. `CommandLineBuilder` or `VersionParser`) which means new test framework require to copy them (most likely) and makes patching them much harder.

>I don't think we are going to extract PHPUnit, for example

I don't know if it's necessary in the current form, but on principle we should be able to at least.

maks-rafalko:
> The goal is the RFC.

no, I mean the goal of this PR in particular, I miss the point.

> The goal is to fix them yes.

How are we going to fix PHPUnit's ones if we don't plan to extract it? Let's have a concrete example with concrete errors / classes so it's clear where we are and where we want to be.

> to allow one to introduce their own, we would not be able to in the current form.

sorry, what do you mean exactly here? Could you please give a concrete example of what will be impossible and why?

> I don't know if it's necessary in the current form, but on principle we should be able to at least.

that's the question. If we are not going to extract it, I don't get why we need errors in the baseline for PHPUnit adapter. I probably miss something, so appreciate examples / explanation.

theofidry:
>no, I mean the goal of this PR in particular, I miss the point.

If tomorrow we are to extract a static analysis test framework contract/abstract, and bridges, like we did for the test frameworks, we would not be able to due to the current API relying on Infection's core code. Whilst my goal is a bit different (making static analysis a regular test framework instead), the problem is the same.

Those PHPat rules codifies the rules that must be respected for that goal. So now instead of having to look at the code and get pack to understand what is fine to depend on what – which is not necessarily clear and easy, especially when it's not a task that can be carried out in one day, we have PHPStan violations to stick to. Violations that can be fixed one after the other.

>How are we going to fix PHPUnit's ones if we don't plan to extract it? Let's have a concrete example with concrete errors / classes so it's clear where we are and where we want to be.

I'm not following here. But there's also a mea culpa to do.

The "concrete example" for me there is the interface proposed in the RFC, this is what I would like `TestFrameworkAdapter` to look like and that it would fit both current test frameworks and static analysis tools alike.

Regarding PHPUnit specifically, there is two sides to this.

The first side is, like the static analysis tools, living in the core means it's easy to have leaky abstractions, i.e. have the adapter that relies on infection's core when it should not. (this is more true for the static analysis tool atm than the PHPUnit adapter), or vice-versa (https://github.com/infection/infection/issues/3082). As long as PHPUnit is not about to be extracted, it is not an immediate problem.

The second side to this is, and for which the mea culpa is for, is that actually I _do_ want to extract the PHPUnit package (I completely forgot when writing my reply). I want to do it because of https://github.com/infection/infection/issues/3082, which means the PHPSpec and Codeception adapters will want to leverage the PHPUnit adapter, mostly for the tracer parts. (Reminder that this simply reflects the fact that both PHPSpec and Codeception themselves rely on PHPUnit for the coverage part).

>> to allow one to introduce their own, we would not be able to in the current form.

> sorry, what do you mean exactly here? Could you please give a concrete example of what will be impossible and why?

From the violations added in this PR (non-exhaustive list):

- `StaticAnalysisToolAdapter` depends on `Infection\Process\Factory\LazyMutantProcessFactory`
- `{PhpStan,Mago}MutantExecutionResultFactory` depends on:
  - `Infection\Mutant\DetectionStatus`
  - `Infection\Mutant\MutantExecutionResult`
  - `Infection\Mutant\MutantExecutionResultFactory`
  - `Infection\Process\MutantProcess`

theofidry:
@maks-rafalko I updated the PR. Hopefully the structure, intent and rules are clearer now.

theofidry:
@sanmai @sidz this PR is ready for review. Let me know if you prefer to opt-out of the review of this feature if you're not peculiarly interested in it.

maks-rafalko (approved):

PR #3274, opened 2026-06-07
chore: Upgrade to PHPUnit 12

theofidry:
## Description

Since we dropped support for PHP 8.2, we can now upgrade to PHPUnit 12 which supports PHP 8.3 as the minimum version.

This upgrade requires several changes. Most are preparatory step for PHPUnit 13 and are about tightening some usages.

## Changes

- `#[CoversNothing]` can no longer be used on a method (currently a deprecration, not a failure).
- PHPUnit `Mock`s expects assertions, otherwise a stub should be used. I switch to stub where we could, but there is various places where we create a mock in `::setUp()` and later use it either as a stub or a mock depending of the test method. In those cases, the attribute `#[AllowMockObjectsWithoutExpectations]` was added (currently a deprecration, not a failure).
- `::assertEqualsCanonicalizing()` changed and no longer ignore map vs list differences, requiring some re-indexing. There is new PHPUnit array assert methods that would work better, e.g. [`::assertArraysAreEqualIgnoringOrder()`](https://docs.phpunit.de/en/13.2/assertions.html#assertarraysareequalignoringorder), but those are only available as of PHPUnit 13.
- PHPUnit no longer allows the PHPUnit data providers to provide more arguments than necessary (currently a deprecration, not a failure). This was happening when we were re-using a data provider (e.g. one was providing 3 arguments, but another test method only needed the first 2).


tests/phpunit/Differ/DifferTest.php
```diff
@@ -83,6 +83,13 @@ public function test_it_can_diff_the_code_as_arrays(
         $this->assertSame($expected, $actual);
     }
 
+    public static function diffStringProvider(): iterable
+    {
+        foreach (self::diffProvider() as $label => [$sourceA, $sourceB, $expected]) {
+            yield $label => [$sourceA, $sourceB, $expected];
+        }
```
sanmai: Not sure if I understand the goal of this contraption 🤔

tests/phpunit/FileSystem/Finder/StaticAnalysisToolExecutableFinderTest.php
```diff
@@ -93,8 +93,8 @@ protected function setUp(): void
 
         $this->fileSystem = new Filesystem();
 
-        $this->composerFinder = $this->createMock(ComposerExecutableFinder::class);
-        $this->composerFinder->method('find')
+        $this->composerFinderStub = $this->createStub(ComposerExecutableFinder::class);
+        $this->composerFinderStub->method('find')
```
sanmai:
```suggestion
        $this->composerFinder->method('find')
```
In some places we did not add the suffix, and in some places we did; I'd rather keep it without suffix because it reduced the diff radius.

tests/phpunit/Mutant/MutantCodeFactoryTest.php
```diff
@@ -130,6 +130,13 @@ public function test_it_creates_the_mutant_code_without_altering_the_original_no
         $this->assertSame($originalNodesDump, $originalNodesDumpAfterMutation);
     }
 
+    public static function mutationOnlyProvider(): iterable
+    {
+        foreach (self::mutationProvider() as $label => [$mutation]) {
+            yield $label => [$mutation];
+        }
```
sanmai: Looks redundant.

tests/phpunit/Mutator/ProfileListProvider.php
```diff
@@ -77,6 +77,13 @@ public static function mutatorNameAndClassProvider(): iterable
         }
     }
 
+    public static function implementedMutatorFileAndClassProvider(): iterable
+    {
+        foreach (self::implementedMutatorProvider() as $label => [$filePath, $className]) {
+            yield $label => [$filePath, $className];
+        }
+    }
```
sanmai, line 87: could be `yield from` if we actually need this

tests/phpunit/Differ/DifferTest.php
theofidry: `self::diffProvider()` has a 4th argument (not shown here as not needed). This contraption is about exposing a new provider with one argument less; otherwise PHPUnit complains that the provider has more parameters than the method needs

tests/phpunit/Mutant/MutantCodeFactoryTest.php
theofidry: similar to https://github.com/infection/infection/pull/3274#discussion_r3380184334

That said I appreciate it's a bit annoying to read/understand. Maybe I can extract that in a PHPUnit data provider utility

tests/phpunit/Differ/DifferTest.php
sanmai: yeah that's very annoying; we can also add a extra argument to the test itself - it will be unused, yes, but one less data provider

tests/phpunit/AutoReview/EnvVariableManipulation/EnvTestCasesProvider.php
```diff
@@ -138,6 +141,10 @@ private static function checkTestCaseForEnvManipulations(string $testCaseClass):
         while ($parentTestCaseClassReflection->getName() !== TestCase::class) {
             $parentTestCaseFileName = $parentTestCaseClassReflection->getFileName();
 
+            if ($parentTestCaseFileName === false || !self::isProjectCodeFile($parentTestCaseFileName)) {
```
theofidry: this is a bug and should be extracted

tests/phpunit/Differ/DifferTest.php
```diff
@@ -83,6 +84,14 @@ public function test_it_can_diff_the_code_as_arrays(
         $this->assertSame($expected, $actual);
     }
 
+    public static function diffStringProvider(): iterable
```
theofidry, line 87: we could extract this class of fix

tests/phpunit/AutoReview/EnvVariableManipulation/EnvTestCasesProvider.php
theofidry: done in https://github.com/infection/infection/pull/3282

phpunit.xml.dist
```diff
@@ -8,11 +8,7 @@
          failOnDeprecation="true"
          failOnRisky="true"
          requireCoverageMetadata="true"
-         displayDetailsOnIncompleteTests="true"
-         displayDetailsOnPhpunitDeprecations="true"
-         displayDetailsOnSkippedTests="true"
-         displayDetailsOnTestsThatTriggerDeprecations="true"
-         displayDetailsOnTestsThatTriggerWarnings="true"
+         displayDetailsOnAllIssues="true"
```
theofidry, line 11: Extracted to https://github.com/infection/infection/pull/3283.

tests/phpunit/FileSystem/Finder/StaticAnalysisToolExecutableFinderTest.php
theofidry: Updated to not change the name when we did not have the prefix.

I think we can discuss of whether we want to have the stub/mock prefix, but that should be done separately (and personally I'm not too interested in it so I'll not pursue it either).

tests/phpunit/Differ/DifferTest.php
theofidry: I proposed a different form, but your suggestion is valid too. I don't have a strong opinion so happy to adjust based on what @maks-rafalko thinks.

To recap, the problem here is that the data provider `diffProvider`, which is re-used for multiple tests, declares 4 parameters whereas `test_it_shows_the_diff_between_two_sources_but_limiting_the_displayed_lines` only uses 3 of them. PHPUnit 12 complains about it. So to fix it there is two options:

<details>
<summary>Option 1: new data provider</summary>

```diff
-   #[DataProvider('diffProvider')]
+   #[DataProvider('diffStringProvider')]
    public function test_it_shows_the_diff_between_two_sources_but_limiting_the_displayed_lines(
        string $sourceA,
        string $sourceB,
        string $expected,
    ): void {

+   public static function diffStringProvider(): iterable
+   {
+       yield from DataProviderFactory::takeArguments(3, self::diffProvider());
+   }
```

</details>

<details>
<summary>Option 2: extra (unused) argument</summary>

```diff
    #[DataProvider('diffProvider')]
    public function test_it_shows_the_diff_between_two_sources_but_limiting_the_displayed_lines(
        string $sourceA,
        string $sourceB,
        string $expected,
+       mixed $_extraArg,
    ): void {
```

</details>

theofidry: I'd like to move forward there. I'll merge as it is (option 1) and will update the code to option 2 if you prefer it, afterwards.

PR #3275, opened 2026-06-08
ci: Prevent GitHub rate limiting when downloading mago in e2e fixtures

theofidry:
I noticed an E2E test fail due to the mago binary not being downloaded. Given not `GITHUB_TOKEN` was available in the failing task, this is likely due to a rate limiting.

This PR fixes this issue the same way we did for the autoreview step.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -123,6 +123,12 @@ jobs:
           ls tests/e2e/*/composer.json | xargs dirname |
              xargs -I{} -P 4 $CHRONIC composer --working-dir={} install --no-interaction --prefer-dist --no-progress || true
 
+      - name: Download mago used in some E2E tests
+        run: vendor/bin/mago --version
+        working-directory: tests/e2e/Mago_Integration
+        env:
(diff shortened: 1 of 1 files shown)


sanmai (approved):

.github/workflows/e2e-tests.yaml
```diff
@@ -123,6 +123,12 @@ jobs:
           ls tests/e2e/*/composer.json | xargs dirname |
              xargs -I{} -P 4 $CHRONIC composer --working-dir={} install --no-interaction --prefer-dist --no-progress || true
 
+      - name: Download mago used in some E2E tests
+        run: vendor/bin/mago --version
+        working-directory: tests/e2e/Mago_Integration
+        env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
sanmai, line 130: we could also cache this as part of the composer cache

theofidry: the actual file name used changes from a version to another and arch to another, so in practice, whilst possible, I couldn't find a neat way to do so :/

Maybe a better solution is to suggest to Mago to put the downloaded artefact in the Composer cache instead but maybe that's too sketchy.

Alternatively we could ask for a more well known path.

PR #3276, opened 2026-06-08
fix(autoreview): Fix PHPat I/O detection for PHPUnit covers attributes

theofidry:
When detecting I/O, we currently do not check the target symbol for `#[CoversTrait]`, resulting in some false-positives (e.g. `ImplodableEnumTest`) which will be marked as requiring I/O when they do not (note that when a test covers no code, e.g. it uses `#[CoversNothing]`, we require it to belong to the `integration` PHPUnit group, see https://github.com/infection/infection/pull/3277).

This PR extends the covers attribute support to `#[CoversTrait]`, resulting in the deletion of some false-positives from the PHPStan baseline.

As a bonus, I added support for `#[CoversMethod]` (new in PHPUnit 11) too as it was easy to do, although unused at the moment. `#[CoversFunction]` seems more complicated, so I left it alone.

### DIFF
--- a/devTools/phpstan-baseline.neon
+++ b/devTools/phpstan-baseline.neon
@@ -991,12 +991,6 @@ parameters:
 			path: ../tests/phpunit/FileSystem/Finder/TestFrameworkFinderTest.php
 
 		-
-			rawMessage: Infection\Tests\Framework\Enum\ImplodableEnum\ImplodableEnumTest should not exist
-			identifier: phpat.testPHPUnitTestsRequiringIoBelongToIntegrationGroup
-			count: 1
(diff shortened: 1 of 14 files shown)


theofidry:
Note to self: to verify the second change, I'm not sure it make sense as `#[CoversNothing]` does warrant an `integration` PHPUnit group.

Likewise we can skip the analysis if it has that covers annotation...

sanmai:
`CoversNothing` could be a sanity test that does not imply strictly `integration` group

theofidry:
@sanmai true.

I think there's two questions:

- historically what was the behaviour?
- is it more often than not the case.

Because a non-respect of this rule is a violation, which we can silence. There is flexibility in how we enforce it

theofidry:
@sanmai in case you missed it, I clarified this in #3277

PR #3277, opened 2026-06-09
refactor(autoreview): Enforce CoversNothing integration rules separately

theofidry:
# Description

In https://github.com/infection/infection/pull/3276, it was unclear whether tests using `#[CoversNothing]` should belong to the `integration` PHPUnit group. I no longer considered the changes proposed in #3276 sufficient:

- I was not sure whether the currently enforced policy made sense.
- The policy is enforced indirectly through the I/O requirement rules.

I verified our usage of `#[CoversNothing]`, which is very limited. All tests using it, at least at class level, belong to the `integration` PHPUnit group. **Based on current practice, the existing rule that `#[CoversNothing]` warrants the `integration` group makes sense.**

This leaves the second point, which is what this PR addresses: it introduces `PHPUnitTestsWithCoversNothingShouldBelongToIntegrationGroupTest` to enforce the policy that `#[CoversNothing]` requires `#[Group('integration')]`.

## Changes

- Registers a dedicated PHPat rule requiring concrete PHPUnit tests with `#[CoversNothing]` to belong to the `integration` group.
- Adds analyser-backed selectors for PHPUnit tests with `#[CoversNothing]` and for integration PHPUnit tests.
- Extends `AnalysisResult` so analysed state can be reused.

## Reviewer Notes

This is a _flexible_ rule. Exceptions and false positives can be handled with `@phpstan-ignore` or by registering them in `phpstan.neon`. The baseline should remain reserved for violations that should eventually be addressed.

### DIFF
--- a/tests/Architecture/PHPat/Selector/InfectionSelector.php
+++ b/tests/Architecture/PHPat/Selector/InfectionSelector.php
@@ -51,6 +51,8 @@ final class InfectionSelector
 
     private static ?EventArchitecture $eventArchitecture = null;
 
+    private static ?Analyser $analyser = null;
+
     private static ?PHPUnitTestIoRequirements $phpUnitTestIoRequirements = null;
 
     private static ?ReflectionProvider $phpUnitTestIoRequirementsReflectionProvider = null;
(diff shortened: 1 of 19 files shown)


tests/Architecture/PHPat/Selector/InfectionSelector.php
```diff
@@ -158,14 +171,7 @@ public static function eventDirectoryClassWithoutExpectedShape(): SelectorInterf
 
     public static function hasTrivialImplementation(): SelectorInterface
     {
-        $container = SingletonContainer::getContainer();
-
-        return new HasTrivialImplementation(
```
theofidry: should be extracted

PR #3279, opened 2026-06-09
refactor(autoreview): Extract PHPUnit test analysis into a dedicated helper

theofidry:
## Description

Future PHPat rules will need more PHPUnit-related information from PHPStan `ClassReflection` objects.

This PR takes the first step by extracting existing PHPUnit-specific checks into a reusable helper, `PHPUnitTestClassAnalysis`. The goal is not to reduce much code immediately, but to make the intent explicit and keep the next behaviour changes small and easier to review.

## Changes

Adds `PHPUnitTestClassAnalysis`, which:

- Determines whether a PHPStan `ClassReflection` represents a PHPUnit test case.
- Moves the `#[Group('integration')]` attribute parsing out of `PHPUnitTestIoRequirements`.
- Gives future PHPat rules a dedicated place for PHPUnit test-class analysis.

The PHPUnit test-case check is intentionally unchanged for now. It remains a small wrapper around the existing logic so that later fixes can be reviewed as behaviour changes, rather than being mixed into this extraction.

## Reviewer Notes

This PR is intended as a refactor with no behaviour change.

This is a small step to untangle #3277 which itself is a PR made to untangle things. It's partly about refining the modelisation of our rules with PHPat and partly adjusting AI generated code I previously did not care as much about.

### DIFF
--- a/tests/Architecture/PHPat/Selector/ConcretePHPUnitTestClass.php
+++ b/tests/Architecture/PHPat/Selector/ConcretePHPUnitTestClass.php
@@ -36,9 +36,9 @@ declare(strict_types=1);
 namespace Infection\Tests\Architecture\PHPat\Selector;
 
 use Infection\Tests\Architecture\PHPat\Selector\Support\ConcreteClassReflection;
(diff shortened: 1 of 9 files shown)


theofidry:
@coderabbitai full reveiw

Issue #2237, opened 2025-06-19
Reduce mutation scope to class method or line


maks-rafalko:
This is a continuation of our work on reducing the feedback loop working with Infection, see:

- https://github.com/infection/infection/issues/2126

Extracted from https://github.com/infection/infection/pull/2226 and originally suggested by @staabm here https://github.com/infection/infection/pull/2226#issuecomment-2982922609

The goal is to reduce mutation scope to class method or line when you are working on killing mutants, to get the very fast and noiseless feedback.

----

> what [...] if we use a filter which is similar to the ones used in PHPUnit namely phpunit --filter {TestMethodName} {FilePath}. I think this would already bring us a big step forward and it would also prevent the "if-source-changes" problem.
> 
> In infection we could use infection --filter {methodName} {FilePath}
>
> Pros
>
> -  the identifier can be guessed by the user, even without running infection beforehand
> -  the identifer will not change as long as the methods' name not change (and even then the user can predict how a new identifier needs to look like)
> -  the identifier format is known to people already using PHPUnit
> 
> Cons
> 
> -  its not a single mutant, but a handful or something -> this could also be a advantage as you see how your change affects the code arround the location you are testing (nearby mutants) -> so its not too focused on a single problem

Then this idea was extended a bit:

> Another feature that we can add to your idea is to run mutations on particular line ([like we have for ignore config](https://infection.github.io/guide/how-to.html#Disable-in-particular-class-or-method-or-line)), e.g.:
>
> - `SourceClass::someMethod` - mutate the whole method `someMethod` method in class `SourceClass`
> -  `SourceClass::someMethod:32` or just `SourceClass::32` - mutate line 32 on class `SourceClass` on line 32
> [...]
> Also, `SourceClass::someMethod::SomeMutator`

----

I personally think that we need to extend our `--filter` functionality to allow passing `SourceClass::someMethod` / `SourceClass::32` / `SourceClass::someMethod::SomeMutator`, but yet to be discussed & analyzed.

theofidry:
I think there is two issues to extract from here actually:

1. Related to the feedback loop, you want to more easily scope the executed/tested/mutated code.
2. Reduce the noise: you want to be able to more easily/smartly ignore certain mutations for pieces of code.

Regarding the 2. I also think there is a DX issue to consider (which should be a different epic), i.e. make it easier to show how to turn of a specific mutation when it is escaped.

[...]

Issue #2595, opened 2025-11-22
Rework the source & filtering API


theofidry:
I was doing this work in #2395 but I started to loose myself there. I think in the end the problem is just bigger than I thought and will require multiple changes. I think it makes more sense to capture the context, explanation and documentation in a dedicated issue and refer to it rather than that PoC merge request.

Partial relevant history:

- #1632: introduced `--git-diff-filter`.
- #2289: infer the git base instead of the hardcoded `origin/master`.
- #2542: improved the readability of the processing of the filter and git option inputs in the `RunCommand`.
- Related PoCs: #2395 and #2381.

# Background

Currently, the command allows the following options:

- `--filter`: Allow to filter on the source files, e.g. `$ infection --filter=src/Console,src/FileSystem`.
- `--git-diff-filter`: When present, is passed to the underlying `git diff --diff-filter=...` command. See [`--diff-filter`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---diff-filterACDMRTUXB).
- `--git-diff-lines`: An alias for `--git-diff-filter=AM`.
- `--git-diff-base`: E.g. `origin/main`. When not present we do a `git` process to figure it out.

Only one type of filter can be applied, which means `--filter`, `--git-diff-filter` and `--git-diff-lines` are mutually exclusive.

The values are processed (checked, sanitized & validated) in `RunCommand` and forwarded to `ConfigurationFactory`.

On top of that we have the settings coming from the configuration file, `source.{directories, excludes}`. Both saved as is in `Configuration#sourceDirectories` and `Configuration#sourceFilesExcludes`.

The `ConfigurationFactory` processes the configuration file `source` to collect the source files – without accounting for the filter, and save them as `Configuration#sourceFiles`. It also processes the filter (coming from `--filter` or git) and saves the result, which is a comma separated list of file names, as `Configuration#sourceFilesFilter`.

Where are those values used:

| value | usage |
|--|--|
| `isForGitDiffLines` | `IndexXmlCoverageParser`, `FileMutationGenerator`, `NodeMutationGenerator`|
| `gitDiffBase` | `FileMutationGenerator`, `NodeMutationGenerator`|
| `sourceFilesFilter` | `SourceFileFilter` |

How those values are used:

- `ConfigurationFactory` uses `$filter`, `$gitDiffFilter`, `$isForGitDiffLines`, `$gitDiffBase` and `$sourceDirectories` to compute `sourceFilesFilter`.
[...]

PR #3284, opened 2026-06-14
Support positional arguments to mimic `--filter` and `--test-framewor-extra-args` options

### LINKED ISSUE #2191: Add alias for `--test-framework-options="--filter='tests/MyExampleText.php'"` or rename options

maks-rafalko:
## Description

- [x] Doc PR https://github.com/infection/site/pull/303
- [ ] create a ticket for our JB IDE Plugin to support this

TL;DR

This feature allows to do the following:

```diff
- infection --filter=src/Mutator/Number/ --test-framework-extra-args="tests/phpunit/Mutator/Number/"
+ infection run src/Mutator/Number/ tests/phpunit/Mutator/Number/

- infection --filter=src/Mutator/Number/
+ infection run src/Mutator/Number/

- infection --test-framework-extra-args="tests/phpunit/Mutator/Number/"
+ infection run tests/phpunit/Mutator/Number/
```

Fixes https://github.com/infection/infection/issues/2191 and implements it the way I described in this comment: https://github.com/infection/infection/issues/2191#issuecomment-4527920620

This is a feature that allows users to path either or both **source** and **tests** files or folders as Infection arguments, without the need to type `--filter` and/or `--test-framework-extra-args`. For more details see the related issue.

I specifically designed it in such a way so we can pass source and tests arguments in any order, which greatly improves DX (I've been tested this for a week now for a real project).

### What is currently supported
- 1 argument - mutate some source folder/file/path
  - `infection run Plus_` - mutates file(s) matched by this name
  - `infection run Plus_.php` - mutates file(s) matched by this name
  - `infection run src/Service/Mailer.php` - mutates 1 particular file
  - `infection run src/Service/` - mutates the whole folder
  - mutate 2 files
    - `infection run src/Service/Mailer.php src/Entity/Foobar.php`
    - `infection run Mailer.php Foobar.php` - mutates file(s) matched by these names
- 1 argument - test file path/folder path. This is an equivalent to `infection --test-framework-extra-args="..."`. Passes file/folder to `phpunit ...` as argument, reducing the executed tests and significantly improving performance
  - `infection run tests/Unit/Service/MailerTest.php`
  - `infection run tests/Unit/Service/`
  - `infection run tests/Unit/`
  - `infection run tests/`
  - `infection run tests`
- N arguments
  - `infection run tests/Differ tests/Mutator` - many tests folders (or files)
  - `infection run src/Differ/Differ.php tests/Differ src/Mutator/Plus_.php tests/Mutator/Plus_.php` - mixed order of source and tests paths
- 2 arguments - source and test paths
  - `infection run src/Service/Mailer.php tests/Unit/Service/MailerTest.php`
  - `infection run src/Service/Mailer.php tests/Unit/Service/`
  - `infection run src/Service/ tests/Unit/Service/MailerTest.php`
  - `infection run src/Service/ tests/Unit/Service/`
  - `infection run Mailer.php tests/Unit/Service/`
  - `infection run Mailer tests/Unit/Service/`
  - mirror everything above with different order
    - `infection run tests/Unit/Service/MailerTest.php src/Service/Mailer.php`
    - `infection run tests/Unit/Service/ src/Service/Mailer.php`
    - `infection run tests/Unit/Service/MailerTest.php src/Service/`
    - `infection run tests/Unit/Service/ src/Service/`
    - `infection run tests/Unit/Service/ Mailer.php`
    - `infection run tests/Unit/Service/ Mailer`

## Related issues

- https://github.com/infection/infection/issues/2191
- https://github.com/infection/infection/issues/2237 - as the next step, I will add support to run 
  - `infection run App\Some\class` 
  - and even `infection run App\Some\class::method`
  - or `App\Some\class::method::33`

## For reviewers

I would ask you to try this for your project(s) if possible, reporting any issues here.

Compiled zipped PHAR if needed is attached:

[infection.phar.zip](https://github.com/user-attachments/files/29015385/infection.phar.zip)

To understand why particular syntax is supported, make sure to read `--filter` docs:

- https://infection.github.io/guide/command-line-options.html#filter

#### Extracted PRs

- https://github.com/infection/infection/pull/3285


theofidry:
First round of feedback: this focuses more on the general feature and not the code.

I am not sure I understand the rational behind enforcing two arguments, I feel like it drives towards a weird UX. For example, we expect the user to do:

```shell
infection run src/Differ,src/Source
```

instead of (what I think is more natural):


```shell
infection run src/Differ src/Source
```

Which can be achieved by making the input argument a `InputArgument::IS_ARRAY` type.

I understand that behind this is inherited behaviour from `--filter`, but I am questioning if we should actually follow that path.

Another limitation is not being able to pass more than one test directory:

```shell
# none of those are allowed
infection run tests/phpunit/Differ tests/phpunit/Source
infection run tests/phpunit/Differ,tests/phpunit/Source
```

I thought it was a limitation of the current `--test-framewor-extra-args` option, but actually using `--test-framework-extra-args="tests/phpunit/Differ tests/phpunit/Source"` works just fine. You can also verify it with:

```shell
phpunit tests/phpunit/Source tests/phpunit/Differ --list-tests
```

That being said, I wouldn't surprised this does not work with all test frameworks nor all PHPUnit versions.

maks-rafalko:
> am not sure I understand the rational behind enforcing two arguments

this is because original idea was to just use the same behavior without the need to type `--filter` and `--test-framework-extra-args`. Example [from the related issue](https://github.com/infection/infection/issues/2191#issuecomment-4527920620)

```diff
- infection 
-    --filter=path/to/DocumentLines.php 
-    --test-framework-extra-args="tests/path/to/DocumentLinesTest.php"  

+ infection run
+    path/to/DocumentLines.php 
+    tests/path/to/DocumentLinesTest.php
```

> Another limitation is not being able to pass more than one test directory:

I wanted to address this in a separate PR, but I agree, we can think about dropping backward-compatibility with `--filter` syntax and allow

```shell
infection source1 source2 test1 test2 ... testN sourceN
```

I will play with it. This shouldn't block the testing though, the look and feel will not change much comparing to the current state of the PR if you use 1 source and 1 test paths.

maks-rafalko:
applied the changes, updated PR description

maks-rafalko:
@coderabbitai review

src/Command/Option/PathsArgument.php
```diff
@@ -0,0 +1,76 @@
...
+    use CannotBeInstantiated;
+
+    public const string NAME = 'paths';
+
+    public static function addArgument(Command $command): Command
+    {
+        return $command->addArgument(
+            self::NAME,
+            InputArgument::IS_ARRAY | InputArgument::OPTIONAL,
+            'Source and/or test paths to focus mutation testing on. Source paths are forwarded to "--filter"; test paths are forwarded to "--test-framework-extra-args". Kind is auto-detected against the configured "source.directories". Pass as many paths as needed: `infection run src/A.php src/B.php tests/ATest.php`.',
```
theofidry: I wonder if we should mention `--filter` here. Personally I would like to deprecate it in favour of this feature.

theofidry: Phrasing suggestion, also note for options declared as a dedicated class you can use their constants:


```suggestion
            sprintf(
                'Limit mutation testing to the given source and/or test paths. Source paths are used as "%s" values; test paths are passed to the test framework (%s). Path kind is inferred from the configured source directories and test-like paths.',
                SourceFilterOptions::PLAIN_FILTER_NAME,
                TestFrameworkExtraArgsOption::NAME,
            ),
```

src/Command/Option/PathsArgument.php
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * @return list<non-empty-string>
+     */
+    public static function get(IO $io): array
+    {
+        return array_values(
+            array_filter(
+                $io->getInput()->getArgument(self::NAME),
+                static fn (string $path): bool => trim($path) !== '',
```
theofidry: you do check for the trimmed string here but the value is not trimmed, and `$io->getInput()->getArgument(self::NAME),` returns untrimmed string.

src/Configuration/ConfigurationFactory.php
```diff
@@ -170,6 +171,8 @@ public function create(
         $mutators = $this->mutatorFactory->create($resolvedMutatorsArray, $useNoopMutators);
         $ignoreSourceCodeMutatorsMap = $this->retrieveIgnoreSourceCodeMutatorsMap($resolvedMutatorsArray);
 
+        $sourceFilter = self::mergePositionalSourcePaths($sourceFilter, $classifiedPaths->sourcePaths);
```
theofidry: I find this line a bit strange, conceptually. If there is any "classified paths", then `--filter` should be `null` no?

Oh, actually no, whilst it probably doesn't make sense to allow `--filter`, `--git-diff-lines` for example may still be desired

src/Command/RunCommand.php
```diff
@@ -505,9 +508,35 @@ private function createContainer(
             projectDirectory: $this->getProjectDirectory($io),
             staticAnalysisTool: $commandHelper->getStringOption(self::OPTION_STATIC_ANALYSIS_TOOL, Container::DEFAULT_STATIC_ANALYSIS_TOOL),
             mutantId: $commandHelper->getStringOption(self::OPTION_MUTANT_ID, Container::DEFAULT_MUTANT_ID),
+            positionalPaths: PathsArgument::get($io),
```
theofidry: There was a lot of work done not too long ago to try to simplify the filtering workflow (https://github.com/infection/infection/issues/2595, https://github.com/infection/infection/pull/2395) and I feel this is a step back from that direction.

I would be really keen to try to re-use the `SourceFilter` interface instead. The rational being that:

- It would make it easier to re-use it elsewhere (e.g. `config:list-sources` or `initial-test:run`).
- It would make it easier to enforce exclusion rules (e.g. you cannot use `--filter` with this).
- It will make the removal of `--filter` less awkward (I presume)
- It would deffer the complexity to `ConfigurationFactory`, instead of pushing another value object to the container.

The only problem I would have with it is that then it's not strictly about source files, hence "SourceFilter" is a misnomer, but that can be changed afterwards.

src/Configuration/PositionalPathsClassifier.php
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        // at this point, both for `source` and for `test` slots we must have a real file path for provided value
+        $absolutePath = Path::isAbsolute($path)
+            ? $path
+            : Path::join($configDir, $path);
+
+        if (!$fileSystem->isReadableFile($absolutePath) && !$fileSystem->isReadableDirectory($absolutePath)) {
+            throw new InvalidArgumentException(sprintf(
+                'Positional path "%s" does not exist (resolved to "%s"). Check the path, or pass it via "--%s" / "--%s" explicitly.',
```
theofidry:
```suggestion
                'Invalid path argument "%s": multiple paths must be passed as separate arguments.',
```

I don't think "positional path" is very understandable to the user

src/Configuration/PositionalPathsClassifier.php
```diff
@@ -0,0 +1,238 @@
...
+            }
+        }
+
+        return new ClassifiedPaths($sourcePaths, $testPaths);
+    }
+
+    /**
+     * @return list<non-empty-string>
+     */
+    private static function resolveAbsoluteSourceDirectories(
```
theofidry, line 99: this is awfully familiar to `BasicSourceCollector::makePathsAbsolute()`. I think it's not too trivial of a piece of code, there is a very conscious decision here that the base is the configuration. However, we do want to keep relative paths in other places.

`Infection\Configuration\Entry\Source` should remain simple, it represents the entry from the configuration file. But maybe we could introduce `Infection\Configuration\Source`:

```
Source {
  + relativePaths
  + absolutePaths
  + excludedPaths
}
```

(AFACS we don't need that yet for excluded paths hence left alone). Another advantage is that we can take the change to clarify the naming, since `Source#directories` can in theory contain plain file paths... Likewise for the exclude.

We could also move the normalization of the paths (`Path::canonicalize()` and removing the trailing `/`) there too.

src/Configuration/PositionalPathsClassifier.php
```diff
@@ -0,0 +1,238 @@
...
+        if (self::looksLikeClassOrFileName($path)) {
+            return self::KIND_SOURCE;
+        }
+
+        // at this point, both for `source` and for `test` slots we must have a real file path for provided value
+        $absolutePath = Path::isAbsolute($path)
+            ? $path
+            : Path::join($configDir, $path);
+
+        if (!$fileSystem->isReadableFile($absolutePath) && !$fileSystem->isReadableDirectory($absolutePath)) {
```
theofidry: I'd recommend to move this block into a dedicated method for better readability:


```suggestion
        self::assertIsValidPath(...);
```

theofidry: Can't it just be:


```suggestion
        if (!$fileSystem->isReadable($absolutePath)) {
```

Then? The `*File` and `*Directory` where for when a stricter assertion was needed, but in this case the Symfony original redable seems sufficient

src/Configuration/PositionalPathsClassifier.php
```diff
@@ -0,0 +1,238 @@
...
+        return self::KIND_TEST;
+    }
+
+    /**
+     * \SomeNamespace\Class
+     * \SomeNamespace\Class::method
+     * \SomeNamespace\Class::method::34
+     * App\Foo (bare, unbackslashed)
+     */
+    private static function looksLikeFqcn(string $value): bool
```
theofidry: I think this is the sort of utility that would be suited for `ClassName`, e.g. `ClassName::isFqcn()`

src/Configuration/PositionalPathsClassifier.php
```diff
@@ -0,0 +1,238 @@
...
+    private static function looksLikeClassOrFileName(string $value): bool
+    {
+        return ctype_upper($value[0]) // class and file name starts with Pascal Case (A-Z only, not digits)
+            && !str_contains($value, '/')
+            && !str_contains($value, '\\');
+    }
+
+    private static function looksLikeTestPath(string $value): bool
+    {
+        $normalized = str_replace('\\', '/', $value);
```
theofidry: I do not understand this `$normalized` here. Are we trying to translate a FQCN into a path?

If so, i.e. if I understood that correctly, then I think it opens up a discussion that we will have at some point with that feature which is how we handle the test framework extra arguments.

Indeed, currently our policy is pretty much "forward X to the test framework". We do not really validate it, if the test framework is not happy with something, it is up to it to fail with an error message. Whilst this may result in a non-ideal user experience, it frees our hands of otherwise a lot of responsibility.

I think we should try to keep it that way, so in that sense, I do not think we should do a FQCN -> path transformation.

That said, I really like the idea of being able to do:

```shell
bin/infection 'Infection\Tests\Framework\ClassNameTest::test_it_can_get_a_class_short_name'
```

(or any test ID found e.g. in a coverage report or something)

But then we do need this extra layer to:

- know how to distinguish a test ID from a source selection
- map this to the test framework – and the test framework may not support it so we need to decide what to do in that situation

Either way, I do not think it's a straightforward answer/solution, so I'd recommend to keep this initial PR more scoped and opt for the first option.

theofidry (commented):
I did a review on the source code only, there is quite a few big questions.

I hope I was clear enough with them, and if you feel you're lacking too much time and something is not clear, I'm happy to provide a patch too.

src/Command/Option/PathsArgument.php
maks-rafalko: removed

maks-rafalko: updated a bit

src/Command/Option/PathsArgument.php
maks-rafalko: done

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: done

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: done

maks-rafalko: `isReadable()` is a private method

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: it's not only `isFqcn`, it also finds a `::metho` or `::method::line`, so left as is for now. Renamed this method

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: Short answer: we _do not_ try to translate FQCN into a path here. 

As I explained [here](https://github.com/infection/infection/issues/2191#issuecomment-4528131479), infering a source from a test and vice-versa is a another big topic that is out of scope of this particular PR.

> That said, I really like the idea of being able to do:
>
> `bin/infection 'Infection\Tests\Framework\ClassNameTest::test_it_can_get_a_class_short_name'`

we will have it, and I have a todo on line 159 for this. Method `looksLikeTestPath` isn't related to it.

src/Configuration/ConfigurationFactory.php
maks-rafalko: could you please elaborate on this? What is wrong?

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: I don't really get what is wrong here. If I got you right, you want to have both relative paths and absolute paths in a `Source` VO and reuse it here?

If so, sounds like out of scope of this PR and can be done separately

src/Command/RunCommand.php
maks-rafalko: PoC attempt: https://github.com/infection/infection/pull/3294

maks-rafalko:
@theofidry is there anything else you want me to look at?

src/Command/RunCommand.php
maks-rafalko: PoC has been approved and merged in here

src/Command/RunCommand.php
```diff
@@ -489,7 +492,7 @@ private function createContainer(
             testFrameworkExtraOptions: TestFrameworkOptionsOption::get($io),
             testFrameworkExtraArgs: TestFrameworkExtraArgsOption::get($io),
             staticAnalysisToolOptions: $commandHelper->getStringOption(self::OPTION_STATIC_ANALYSIS_TOOL_OPTIONS, Container::DEFAULT_STATIC_ANALYSIS_TOOL_OPTIONS),
-            sourceFilter: SourceFilterOptions::get($io),
+            sourceFilter: SourceFilterOptions::get($io, PathsArgument::get($io)),
```
theofidry, line 495: Maybe a silly suggestion, but can't this be part of the `SourceFilterOptions`? I.e. remove `PathsArgument` entirely, have its add arguments as part of `SourceFilterOptions`?

If yes then `SourceFilterOptions` would need to be renamed to something like `SourceFilterArgumentsAndOptions` though...

A benefit of that IMO is that it would immediately carry over to the other relevant commands. So `initial-test:run` would automatically work with this new feature for example.

Another side effect is that `SourceFilterOptions::get()` no longer needs the second optional parameter.

src/Configuration/ConfigurationFactory.php
theofidry: My comment was that it looked weird because I do not see how we could end up with a "merge" situation, as `--filter` and positional arguments are exclusive (cannot be used together).

But I corrected myself as in fact you could very well use a positional argument with `--git-diff-lines` (I think?).

src/Command/RunCommand.php
```diff
@@ -153,6 +154,8 @@ protected function configure(): void
             ->setName('run')
             ->setDescription('Runs the mutation testing.');
 
+        PathsArgument::addArgument($this);
```
theofidry, line 157: Adding a separate comment here for better visibility as it was once reported in https://github.com/infection/infection/pull/3284#discussion_r3433980369.

An issue here is that those arguments are not available/propagated to the other relevant commands. Indeed, I think it should also work with `config:list-sources` and `initial-test:run`. Maybe I am missing another command.

However, there is multiple ways this could be addressed, one discussed in the comment bellow.

src/Configuration/SourceFilter/PositionalPathsFilter.php
```diff
@@ -0,0 +1,53 @@
...
+namespace Infection\Configuration\SourceFilter;
+
+/**
+ * Wraps raw positional CLI path arguments before they are classified into source
+ * and test paths. ConfigurationFactory resolves this into a PlainFilter (source
+ * paths) and forwards test paths to testFrameworkExtraArgs.
+ *
+ * @internal
+ */
+final readonly class PositionalPathsFilter implements SourceFilter
```
theofidry, line 45: What do you think of `UnclassifiedPathFilter`? IMO it conveys more that it's various filtering paths that we do not know if they are tests or source yet, without specifying that this comes from positional CLI path arguments, which is more of a detail of the invoker, rather than a concern of this class.

src/Command/Option/SourceFilterOptions.php
```diff
@@ -247,5 +259,24 @@ private static function assertOnlyOneTypeOfFiltering(
...
+        if ($positionalPaths !== [] && $plainFilter !== null) {
+            throw new InvalidArgumentException(
+                sprintf(
+                    'Cannot pass source paths as positional arguments together with the "--%s" option. Use either form, not both.',
+                    self::PLAIN_FILTER_NAME,
+                ),
+            );
+        }
+
+        if ($positionalPaths !== [] && $gitFilter !== null) {
```
theofidry, line 272: I wonder if this assertion is really wrong.

I could imagine myself typing:

```
infection tests/.../MyClassTest.php --git-diff-filter
```

The rational being that `--git-diff-filter` is good at limiting the scope of the source files processed, but this does not limit what tests are executed (unless you register a testing strategy `--map-source-class-to-test`). Passing the positional arguments here help with that.

But I think this has further implications, so I think it is wiser to consider this as a dedicated feature to implement in a separate PR.

src/Configuration/ConfigurationFactory.php
```diff
@@ -400,6 +412,37 @@ private function retrieveIgnoreSourceCodeMutatorsMap(array $resolvedMutatorsMap)
...
 
+    /**
+     * @return array{0: PlainFilter|null, 1: string|null}
+     */
+    private function resolvePositionalPathsFilter(
+        PositionalPathsFilter $sourceFilter,
+        SchemaConfiguration $schema,
+        ?string $testFrameworkExtraArgs,
+    ): array {
+        $classified = PositionalPathsClassifier::fromPaths(
```
theofidry: In the end `PositionalPathsClassifier` does a bit more than maybe initially expected and requires the usage of some services (`Filesystem`). So I think it would be more conventional to make into a regular service.

We are in `ConfigurationFactory` which is a service too, so injecting another service is not a problem. And this way ``PositionalPathsClassifier::fromPaths()` only requires the actual input, the services it relies on can be injected as part of its constructor.

src/Configuration/ConfigurationFactory.php
```diff
@@ -170,6 +174,14 @@ public function create(
         $mutators = $this->mutatorFactory->create($resolvedMutatorsArray, $useNoopMutators);
         $ignoreSourceCodeMutatorsMap = $this->retrieveIgnoreSourceCodeMutatorsMap($resolvedMutatorsArray);
 
+        if ($sourceFilter instanceof PositionalPathsFilter) {
+            [$sourceFilter, $testFrameworkExtraArgs] = $this->resolvePositionalPathsFilter(
```
theofidry:
```suggestion
            [$sourceFilter, $testFrameworkExtraArgs] = $this->resolveUnclassifiedPaths(
```

src/Configuration/ConfigurationFactory.php
```diff
@@ -170,6 +174,14 @@ public function create(
         $mutators = $this->mutatorFactory->create($resolvedMutatorsArray, $useNoopMutators);
         $ignoreSourceCodeMutatorsMap = $this->retrieveIgnoreSourceCodeMutatorsMap($resolvedMutatorsArray);
 
+        if ($sourceFilter instanceof PositionalPathsFilter) {
```
theofidry: I think this could be moved to `refineFilterIfNecessary()`. It does mean the signature of the latter changes, but I think it conveys better the meaning.

Previously, filtering only affected the source code, no tests, with:

```
[$sourceFilter, $testFrameworkExtraArgs] = $this->refineFilterIfNecessary($sourceFilter, $schema);
```

We are explicit that the filtering also affect the test framework in some ways.

For clarity: I am not suggesting to remove `::resolvePositionalPathsFilter()`, just to "hide" it in `::refineFilterIfNecessary()`.

src/Configuration/PositionalPathsClassifier.php
```diff
@@ -0,0 +1,236 @@
...
+    use CannotBeInstantiated;
+
+    private const string KIND_SOURCE = 'source';
+
+    private const string KIND_TEST = 'test';
+
+    /**
+     * @param list<non-empty-string> $paths
+     */
+    public static function fromPaths(
```
theofidry:
```suggestion
    public static function classify(
```

?

src/Configuration/PositionalPathsClassifier.php
```diff
@@ -0,0 +1,236 @@
...
+use function strtolower;
+use Symfony\Component\Filesystem\Path;
+
+/**
+ * Classifies positional `path` arguments into source-filter and
+ * test-framework-extra-args buckets.
+ *
+ * @internal
+ */
+final class PositionalPathsClassifier
```
theofidry:
```suggestion
final readonly class PositionalPathsClassifier
```

src/Configuration/PositionalPathsClassifier.php
theofidry: >If so, sounds like out of scope of this PR and can be done separately

Yes. I would do it either before or after. I'm happy to do it

src/Configuration/PositionalPathsClassifier.php
theofidry: OK. I have further comments on this piece of code but will open it in a separate comment.

src/Configuration/PositionalPathsClassifier.php
theofidry: oh, nvm then

src/Configuration/PositionalPathsClassifier.php
```diff
@@ -0,0 +1,236 @@
...
+        string $configDir,
+    ): array {
+        $absolute = [];
+
+        foreach ($schema->source->directories as $directory) {
+            $resolved = Path::isAbsolute($directory)
+                ? $directory
+                : Path::join($configDir, $directory);
+
+            $canonical = rtrim(Path::canonicalize($resolved), '/');
```
theofidry: this mutation is legit, `Path::canonicalize()` will remove the trailing `/`

theofidry: `Path::join()` canonicalise the path already, so this piece of code could be simplified to:


```suggestion
            $canonical = Path::isAbsolute($directory)
                ? Path::canonicalize($directory)
                : Path::join($configDir, $directory);
```

src/Configuration/PositionalPathsClassifier.php
```diff
@@ -0,0 +1,236 @@
...
+     */
+    private static function isInsideSourceDirectories(
+        string $path,
+        array $absoluteSourceDirs,
+    ): bool {
+        if ($absoluteSourceDirs === []) {
+            return false;
+        }
+
+        $candidate = rtrim(Path::canonicalize($path), '/');
```
theofidry: likewise:


```suggestion
        $candidate = Path::canonicalize($path);
```

src/Configuration/PositionalPathsClassifier.php
```diff
@@ -0,0 +1,236 @@
...
+    private static function looksLikeClassOrFileName(string $value): bool
+    {
+        return ctype_upper($value[0]) // class and file name starts with Pascal Case (A-Z only, not digits)
+            && !str_contains($value, '/')
+            && !str_contains($value, '\\');
+    }
+
+    private static function looksLikeTestPath(string $value): bool
+    {
+        $lowered = strtolower($value);
```
theofidry: I'm having a fresh look and I'm having strong doubts about this. This feels to tailored to PHPUnit. Indeed, if you pick up Behat or PhpSpec, none of the assumptions made about what a "test file path" look hold.

Part of the issue is, by design, we do not know what the tests paths are. Indeed, we only require the user to register the test framework (or we guess it) and we orchestrate it. Anything require knowing the test paths is done much later when we collected them from the test coverage, and even then it is test framework specific logic, i.e. none of it would belong or be usable here.

So what about simplifying this by assuming that any file path that we know is not part of the source, is the path of a test file?

src/Configuration/PositionalPathsClassifier.php
```diff
@@ -0,0 +1,236 @@
...
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * @param list<non-empty-string> $absoluteSourceDirs
+     *
+     * @return 'source'|'test'
```
theofidry:
```suggestion
     * @return self::KIND_SOURCE|self::KIND_TEST
```

src/Configuration/PositionalPathsClassifier.php
```diff
@@ -0,0 +1,236 @@
...
+        // TODO: FQCN-style arguments (e.g. "\App\Foo" or "\App\Foo::method::45") will
+        // be supported via https://github.com/infection/infection/issues/2237
+        if (self::looksLikeFqcnWithOptionalMethodOrLine($path)) {
+            throw new InvalidArgumentException(sprintf(
+                'FQCN-style arguments like "%s" are not yet supported. See https://github.com/infection/infection/issues/2237.',
+                $path,
+            ));
+        }
+
+        if (self::looksLikeTestPath($path)) {
```
theofidry: as per my comment in `::looksLikeTestPath()` (below), I think the return statement L190 makes sense and we could simplify the above code to remove `::looksLikeTestPath()` and `::looksLikeClassOrFileName()`

src/Configuration/PositionalPathsClassifier.php
```diff
@@ -0,0 +1,236 @@
...
+        if (self::looksLikeClassOrFileName($path)) {
+            return self::KIND_SOURCE;
+        }
+
+        // at this point, both for `source` and for `test` slots we must have a real file path for provided value
+        $absolutePath = Path::isAbsolute($path)
+            ? $path
+            : Path::join($configDir, $path);
+
+        if (self::isValidPath($fileSystem, $absolutePath)) {
```
theofidry: I would rather move this into a `assertIsValidPath()` here which could:

- Check if there is a comma in the path and throw the current exception in this case – technically you can have a file name with a comma but I think it's rare enough that the degraded UX (as in the explanation we give to the user is incorrect) is acceptable.
- Additionally, like currently done, check that the path is readable just to ensure it exists. Realistically any failure is more likely to be caused by a mistake in `::looksLikeFqcnWithOptionalMethodOrLine()`, .e.g. the user passes `\App\Foo::method::45` and we failed to detect it properly. But we need a guard somewhere so I think this file existence check at this point is a good way to do it.

A benefit of this is we can have more detailed or noisy checks in there without reducing the readability of this method here.

theofidry (commented):
I had a fresh review.

src/Configuration/ConfigurationFactory.php
maks-rafalko: this code doesn't exist anymore, resolving

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: resolving, out of scope

src/Command/RunCommand.php
maks-rafalko: this won't works and I don't really like it for a couple of reasons:

- https://github.com/infection/infection/blob/57e9d503cc1727ba9ce11001dbe6fe9d5404171a/src/Command/Debug/DumpAstCommand.php#L117 - this command has its own argument AND uses `SourceFilterOptions`. If we add `PathsArgument` inside `SourceFilterOptions`, this will be broken
- when you combine them, for me it breaks the clarity of this class and it does many things in this case - adds `--filter`, adds positional arguments. I don't see why we would need this. 

Left as is

src/Command/RunCommand.php
maks-rafalko: Left as is, see https://github.com/infection/infection/pull/3284#discussion_r3486308071

src/Configuration/SourceFilter/PositionalPathsFilter.php
maks-rafalko: I intentionally named classes `Positional***` because reading this I understand what we are working with. So since this is a matter of taste, I leave it as is

src/Command/Option/SourceFilterOptions.php
maks-rafalko: indeed this is not an easy topic. It's better to think/work on it separately. Resolving as out of scope

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: fixed

src/Configuration/ConfigurationFactory.php
maks-rafalko: as I wrote in other comments, I intentionally use positional wording. Left as is

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: fixed

maks-rafalko: done

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: > So what about simplifying this by assuming that any file path that we know is not part of the source, is the path of a test file?

while I rewrote it, it's not that easy. Please note

- we do need to support `infection Plus_` notation. `Plus_` is a part of a file name, it's not a valid path and it's not a test. So the logic "if it's not source path then it's a test" - is incorrect. 
- I left `looksLikeClassOrFileName`, but removed `looksLikeTestPath` 
- also, we can have a test inside a source directory, like `src/UserBoundedContext/tests/SomeTest.php` - this is now broken (but worked before). I will work on it separately

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: done

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: see https://github.com/infection/infection/pull/3284#discussion_r3486329946

- `looksLikeTestPath()` is removed
- `looksLikeClassOrFileName` can't be removed

src/Configuration/ConfigurationFactory.php
maks-rafalko: done

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: done

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: I find `::fromPaths()` more expressive and self-explaining, so left as is

src/Configuration/ConfigurationFactory.php
maks-rafalko: made as a service

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: this method has been rewrote. In its current state, I don't think extracting is needed, for me it's clear enough

maks-rafalko:
@theofidry please take a look, fixed some of the comments, answered to others

src/Configuration/SourceFilter/PositionalPathsFilter.php
theofidry: >because reading this I understand what we are working with.

It's an argument.

I'm gonna give some extra counter arguments just in case it may change your mind, because I feel there is both a precedent about similar usage and problems it may cause, see the case of [`JUnitTestCaseSorter` to `TestRunOrderResolver`](https://github.com/infection/infection/pull/3033).

I find "positional path filter" confusing because:

- Having the context and fresh in my mind, I understand "positional" here refers to the "positional CLI command arguments", but as it is qualifies "paths", and personally I don't know what a "positional path" means.
- It describes more where it is used (the CLI positional arguments) than what it is (a collection of source or tests paths).
- "positional argument" is not a term we come across super often neither, nor is it exposed to the user. So here for example `$testFrameworkExtraArgs`, is easy to mentally map back to the CLI option because the CLI option is called `--test-framework-extra-args`. Here "positional" argument"... Not so much. Well maybe. I think personally I'm good because I had to dig into this how to qualify a generic console argument vs a positional argument with you.

That being said, if you really prefer `PositionalPathsFilter`, I can live with it.

src/Configuration/PositionalPathsClassifier.php
theofidry: In the first iteration `::fromPaths()` made sense to me and I immediately understood it, when reading its content, but looking at its call:

```php
$classified = $this->positionalPathsClassifier->fromPaths(...)
```

it reads a bit weirdly. In such cases I feel the two most self-explanatory forms are:

```php
$classified = $this->positionalPathsClassifier->classify(...)
```

And:

```php
$classified = $this->classifier->classifyPositionalPaths(...)
```

I tend to prefer the former because reading `PositionalPathsClassifier::classifyPositionalPaths()` feels very redundant.

But I appreciate it is a matter of taste.

src/Configuration/PositionalPathsClassifier.php
theofidry: >we do need to support infection Plus_ notation. Plus_ is a part of a file name, it's not a valid path and it's not a test. So the logic "if it's not source path then it's a test" - is incorrect.

Ahh I see.

Hm just wondering, I understand it was working for `--filter`, but is it really something we want to support? This looks like a super confusing scenario to me because it can resolve, at the same time, to:

- A class name (from the global namespace sure, but that is valid).
- Source file(s).
- Test file(s).

So it really screams confusion at first glance.

maks-rafalko: > Hm just wondering, I understand it was working for --filter, but is it really something we want to support? 

yeah, I would really like to support it. How I use it: (I compiled PHAR from this branch weeks ago, so had some time using it on a real project)

- you write a new source file
- you write a test for it
- then you want to run Infection only for this new source file and against this new test

How I do it:

- I press my custom shortcut `ctrl + shift + c` to copy a test path from project root, let's say `tests/SomeNewFileTest.php`
- then, in a terminal, I type `infection <cmd + v>` which gives me `infection tests/SomeNewFileTest.php`
- this already gives me what I want if the project uses `#[CoversClass(...)]`
- but if it doesn't, I need to provide also a source file path OR the part of its file name. In this case I do this: `infection <cmd + v> SomeNewFile`

As a result, the command line is `infection tests/SomeNewFileTest.php SomeNewFile` and it works exactly as I need it.

So it was quite convenient so far.

> Test file(s).

this notation only supports source files, not tests file

tests/phpunit/Configuration/PositionalPathsClassifierTest.php
```diff
@@ -0,0 +1,682 @@
...
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\TestCase;
+use function str_starts_with;
+
+#[CoversClass(PositionalPathsClassifier::class)]
+#[CoversClass(ClassifiedPaths::class)]
+#[Group('integration')]
+final class PositionalPathsClassifierTest extends TestCase
+{
+    public function test_it_returns_empty_buckets_when_paths_is_empty(): void
```
theofidry, line 59: I do not like this test very much, I find it quite hard to read, follow the different cases and I think it's overly complicated.

That being said I'm keen to see this PR move forward so I'd suggest to go with that and I can propose a refactor of the test in a separate PR – it does not need to block the feature.

theofidry (approved):

src/Configuration/PositionalPathsClassifier.php
maks-rafalko: > $classified = $this->positionalPathsClassifier->classify(...)

0k, this can be a compromise. Updated it to `classify()`, but kept the class and property name

src/Configuration/SourceFilter/PositionalPathsFilter.php
maks-rafalko: I see the sense in what you describe, thanks. But I don't like the suggested alternative :) 

Combining it with your other comment, I renamed the method `fromPaths()` to `classify()` but kept the class name for now, so we have, as a compromise. Let's see how it goes.

tests/phpunit/Configuration/PositionalPathsClassifierTest.php
maks-rafalko: oh, interesting. I find it quite simple and I like it haha. Let's see how you would refactor it in a separate PR, thanks.

maks-rafalko:
thanks @theofidry for the thorough review, I think this is a good start for this great feature, I like it a lot these days.

Let's go 🚀

src/Configuration/PositionalPathsClassifier.php
theofidry: I'm not entirely convinced, but I also feel I'm getting things mixed up with the test support, so let's see how that goes, we can still change things in the future after all.

PR #3285, opened 2026-06-14
Allow skipping `run` command with positional args

maks-rafalko:
In https://github.com/infection/infection/pull/3284 we add support for passing positional arguments to `infection` binary, like:

```shell
infection run src/Mutator/Number/ tests/phpunit/Mutator/Number/
```

Note, that when you pass positional arguments, Symfony console requires you to pass our default `run` command, because our application has many commands.

This PR checks, if the first positional argument is a known command, and if not  - it means we pass `run`'s command positional arguments so we can default to `run` command.

Having this allows us to do:

```diff
- infection run src/Mutator/Number/ tests/phpunit/Mutator/Number/
+ infection src/Mutator/Number/ tests/phpunit/Mutator/Number/
```

which is even more easier to work with

Related to:

- https://github.com/infection/infection/pull/3284

### DIFF
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -91,20 +91,30 @@ final class Application extends BaseApplication
     #[Override]
     public function doRun(InputInterface $input, OutputInterface $output): int
     {
-        $firstArgument = $input->getFirstArgument();
-
-        // this allows to run `infection tests/` instead of `infection run tests/`
-        if ($firstArgument !== null && !$this->has($firstArgument)) {
-            $newInput = new StringInput('run ' . $input);
-
-            if (!$input->isInteractive()) {
-                $newInput->setInteractive(false);
-            }
-
-            $input = $newInput;
+        // This allows to run `infection tests/` instead of `infection run tests/`.
+        if ($this->isUnknownCommand($input)) {
+            $input = $this->defaultToRunCommand($input);
         }
 
         return parent::doRun($input, $output);
+    }
+
+    private function isUnknownCommand(InputInterface $input): bool
+    {
+        $firstArgument = $input->getFirstArgument();
+
+        return $firstArgument !== null && !$this->has($firstArgument);
+    }
+
+    private function defaultToRunCommand(InputInterface $input): InputInterface
+    {
+        $newInput = new StringInput('run ' . $input);
+
+        if (!$input->isInteractive()) {
+            $newInput->setInteractive(false);
+        }
+
+        return $newInput;
     }
 
     public function getContainer(): Container
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -54,6 +54,7 @@ use OutOfBoundsException;
 use Override;
 use function sprintf;
 use Symfony\Component\Console\Application as BaseApplication;
+use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\StringInput;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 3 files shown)


src/Console/Application.php
```diff
@@ -87,6 +88,25 @@ public function __construct(
...
     }
 
+    #[Override]
+    public function doRun(InputInterface $input, OutputInterface $output): int
+    {
+        $firstArgument = $input->getFirstArgument();
+
+        // this allows to run `infection tests/` instead of `infection run tests/`
+        if ($firstArgument !== null && !$this->has($firstArgument)) {
+            $newInput = new StringInput('run ' . $input);
```
theofidry: this code is not awfully complicated, but it's not immediately clear. WDYT of using some private methods to make it more explicit?


```suggestion
    #[Override]
    public function doRun(InputInterface $input, OutputInterface $output): int
    {
        // This allows to run `infection tests/` instead of `infection run tests/`.
        if (!self::isKnownCommand($input)) {
            $input = self::fallbackToDefaultCommand($input);
        }
        
        return parent::doRun($input, $output);
    }
```

src/Console/Application.php
```diff
@@ -87,6 +88,25 @@ public function __construct(
...
+
+            if (!$input->isInteractive()) {
```
theofidry: In practice, the `Application` either receives an already configured input, or it create an `ArgvInput`. From `bin/infection.php` we appear to do the latter, so the former is in practice only happening for our tests.

I wonder if `::doRun()` is the appropriate hook because it happens after `::configureIO()`. I don't really know if it's a problem. I just wanted to point out that there is also `::configureIO()` and `::run()` as possible hooks too.

theofidry (commented):
👍 for the feature; but maybe this deserves an e2e test?

src/Console/Application.php
maks-rafalko: hooked into `::run()`

src/Console/Application.php
maks-rafalko: rewrote a bit

theofidry:
@maks-rafalko no e2e test?

maks-rafalko:
> no e2e test?

is it needed though? I covered this new logic by unit test, but probably I miss something

theofidry:
>is it needed though? I covered this new logic by unit test, but probably I miss something

I would say so. `Application` and `Command` are both 1kLoC behemoths with a lot of side effects. So I feel unit tests are good for a fast feedback loop but to not enough of a guarantee in themselves.

Also partly why I was asking was that you reacted 👍 to my first suggestion to add them 😅.

I'm happy to add it though (and you can review if you think it's good or too much). I'd like to add an AI skill to write e2e tests.

maks-rafalko:
> Also partly why I was asking was that you reacted 👍 to my first suggestion to add them 😅.

yeah, because I pushed the first commit with no tests at all. So the 👍 was to tests in general, not specifically to e2e. But then I decided to cover the logic not by e2e.

> I'm happy to add it though (and you can review if you think it's good or too much). I'd like to add an AI skill to write e2e tests.

I can do it myself. But if you want to try your skill - feel free.

Btw, starting adding skills was my next "big thing to work on" as I lack them here and think we definitely need skills in our repo

- for us (maintainers)
- for users to be exported from our package to their projects (this shouldn't be exactly how it's described, by I liked the idea of exported skills from this blog post https://php-testo.github.io/blog/skills#llm-skills-%E2%80%94-a-composer-plugin-for-skills)

theofidry:
That looks interesting indeed.

I have a few skills, but I did not publish them because they felt a bit too personal 😅 Also I've been juggling between Claude & Codex and it is a bit annoying to do so, I regularly had to translate one to the other.

But, I think it's worth trying to create a common baseline for maintainers at least. If we can have a good skill for the user, I think that would be awesome. I do not know if skill or MCP is better at this point either...

sanmai:
A well-structured AGENTS.md would be a great start. I can take it on if y'all would not mind.

First thing, what do we want it to cover?

I infer that Fable will be available soon, and I still have about a month of 5x plan. It did some great work making very detailed AGENTS.md on some other projects.

theofidry:
>First thing, what do we want it to cover?

I think that's the hard part to figure out 😅

A few things on top of my head:

- A way to allow the AI to use infection easily, i.e. don't do something like running the entire test suite or knowing how to understand the output. I suspect there is work to be done in infection itself to achieve this, but we gotta start somewhere.
- A skill to write e2e tests.
- I believe the project contains some conventions that the AI often needs to be told to respect, i.e. using the `*Builder`, the tests `Scenario` pattern, using data providers when possible to reduce boilerplate, this sort of stuff.

Maybe it would be good to add new issues for those with an AI label and see if there is more or refine/specify those?

maks-rafalko:
> Maybe it would be good to add new issues for those with an AI label and see if there is more or refine/specify those?

agree. 

Also, one thing that I wanted to add for AI/claude, is to use our new feature `infection sourceN testN`. So when the new code is generated, it can be checked by Infection for an escaped mutants, but with performance in mind -> so running `infection src/NewFile.php tests/NewFileTest.php`, so that it limits initial tests run to just 1 file

Issue #3295, opened 2026-06-20
feat: phpunit/diff >= 9.0 removed UnifiedDiffOutputBuilder


sidz:
## Problem Statement

Sebastian [removed](https://github.com/sebastianbergmann/diff/releases/tag/9.0.0) `UnifiedDiffOutputBuilder` class in the diff >= 9.0 in a favor of `StrictUnifiedDiffOutputBuilder`.

## Proposed Solution

Currently I see only 2 possible solutions:
1. Switch to `StrictUnifiedDiffOutputBuilder` but it will break compatibility with current behavior as it is not possible to suppress the header.
2. Support `UnifiedDiffOutputBuilder` on our end

WDYT?

maks-rafalko:
it's quite hard to decide without concrete examples to understand the difference.

I quickly tried to compare those 2 classes, seems like "header" is this: 

```
string $header = "--- Original\n+++ New\n"
```

if this is what we are talking about, I don't think we need to support our own and we can stick to Strict* one, but if you have more details, feel free to share :)

sidz:
`string $header = "--- Original\n+++ New\n"` - default header in the `UnifiedDiffOutputBuilder` and currently we just override it with an empty string.
`UnifiedDiffOutputBuilder` output:

<img width="2374" height="472" alt="Image" src="https://github.com/user-attachments/assets/44b04bc5-6e3b-4f2a-aaef-8271a0f0d855" />

`StrictUnifiedDiffOutputBuilder` output:

<img width="2397" height="559" alt="Image" src="https://github.com/user-attachments/assets/85882212-73bd-4e76-a692-bef6301e1f6f" />



`StrictUnifiedDiffOutputBuilder` requires to provide `fromFile` and `toFile` array keys. We could pass `Original` (or `Source`) string as a `fromFile` and `New` (or  `Mutated`) string as a `toFile`.

<img width="1736" height="428" alt="Image" src="https://github.com/user-attachments/assets/c3053a5e-18ac-49bc-9c49-aac759d8e468" />





Raised PR: https://github.com/infection/infection/pull/3298

maks-rafalko:
oh, now it's clear, thank you. For some reason I incorrectly understood it and thought we won't have header with `Strict*` one, but it's the opposite.

My thoughts:

- having `---Original` / `--New` for each mutant, where we can have thousands of them, doesn't make sense and makes the diff and the whole log file/output more bigger and noisy, especially given these lines are highlighted
[...]

PR #3298, opened 2026-06-21
Support sebastian/diff: 9

sidz:
## Description
This PR brings [sebastian/diff: 9](https://github.com/sebastianbergmann/diff/releases/tag/9.0.0) support.

## Changes

Replaces `UnifiedDiffOutputBuilder` with `StrictUnifiedDiffOutputBuilder` as `UnifiedDiffOutputBuilder` has been removed in the diff >= 9.

## Checklist

- [x] Tests added/updated

## Breaking Changes

Output changed from:
Screenshot from 2026-06-21 16-01-11

To:
Screenshot from 2026-06-21 15-57-50

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -63,7 +63,7 @@
         "sanmai/duoclock": "^0.1.0",
         "sanmai/later": "^0.1.7",
         "sanmai/pipeline": "^7.2",
-        "sebastian/diff": "^4.0 || ^5.0 || ^6.0 || ^7.0 || ^8.0",
+        "sebastian/diff": "^4.0 || ^5.0 || ^6.0 || ^7.0 || ^8.0 || ^9.0",
         "symfony/console": "^6.4 || ^7.4 || ^8.0",
         "symfony/filesystem": "^6.4 || ^7.4 || ^8.0",
         "symfony/finder": "^6.4 || ^7.4 || ^8.0",
--- a/tests/phpunit/Differ/DifferTest.php
+++ b/tests/phpunit/Differ/DifferTest.php
@@ -112,7 +112,9 @@ final class DifferTest extends TestCase
 
                 PHP,
             <<<'PHP'
-                @@ @@
+                --- Original
+                +++ New
+                @@ -1,5 +1,5 @@
 
                  public function echo(): void
                  {
@@ -209,7 +211,9 @@ final class DifferTest extends TestCase
                 15
                 PHP,
             <<<'PHP'
-                @@ @@
+                --- Original
+                +++ New
+                @@ -4,7 +4,7 @@
                  3
                  4
                  5
@@ -282,7 +286,9 @@ final class DifferTest extends TestCase
                 15
                 PHP,
             <<<'PHP'
-                @@ @@
+                --- Original
+                +++ New
+                @@ -4,10 +4,10 @@
                  3
                  4
                  5
@@ -340,12 +346,14 @@ final class DifferTest extends TestCase
                 1{$windowsLineReturn}2
                 PHP,
             <<<'PHP'
-                @@ @@
-                 #Warning: Strings contain different line endings!
+                --- Original
+                +++ New
+                @@ -1,3 +1,3 @@
                  0
                 -1
                 +1
                  2
+                \ No newline at end of file
 
                 PHP,
             [
@@ -372,12 +380,14 @@ final class DifferTest extends TestCase
                 (1){$windowsLineReturn}2
                 PHP,
             <<<'PHP'
-                @@ @@
-                 #Warning: Strings contain different line endings!
+                --- Original
+                +++ New
+                @@ -1,3 +1,3 @@
                  0
                 -1
                 +(1)
                  2
+                \ No newline at end of file
 
                 PHP,
             [
--- a/tests/phpunit/Differ/DifferTest.php
+++ b/tests/phpunit/Differ/DifferTest.php
@@ -113,7 +113,7 @@ final class DifferTest extends TestCase
                 PHP,
             <<<'PHP'
                 --- Original
-                +++ New
+                +++ Mutated
                 @@ -1,5 +1,5 @@
 
                  public function echo(): void
@@ -212,7 +212,7 @@ final class DifferTest extends TestCase
                 PHP,
             <<<'PHP'
                 --- Original
-                +++ New
+                +++ Mutated
                 @@ -4,7 +4,7 @@
                  3
                  4
@@ -287,7 +287,7 @@ final class DifferTest extends TestCase
                 PHP,
             <<<'PHP'
                 --- Original
-                +++ New
+                +++ Mutated
                 @@ -4,10 +4,10 @@
                  3
                  4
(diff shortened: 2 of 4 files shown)


src/Differ/Differ.php
```diff
@@ -74,7 +73,7 @@ public function diff(string $from, string $to): string
      *
      * @return array{string, string}
      */
-    public function diffToArray(string $from, string $to, ?LongestCommonSubsequenceCalculator $lcs = null): array
```
sidz: `LongestCommonSubsequenceCalculator` unused

maks-rafalko (approved):
Added detailed thoughts here: https://github.com/infection/infection/issues/3295#issuecomment-4762171237

> - having ---Original / --New for each mutant, where we can have thousands of them, doesn't make sense and makes the diff and the whole log file/output more bigger and noisy, especially given these lines are highlighted
> - if you are blocked from upgrading because of this, we can merge your PR as is, make a release and unblock you, and then create our one differ at some point that removes this header.
> 
> And, looking into the first image from your message, we could also get rid of @@ @@ which doesn't make sense for us as well. Even @@ -117,7 +117,7 @@ doesn't make sense for our diffs, we are interested in the minimal source code changes diff only IMO

so, eventually I think it's better to not have these headers and `@@` lines

composer.json
```diff
@@ -63,7 +63,7 @@
         "sanmai/duoclock": "^0.1.0",
         "sanmai/later": "^0.1.7",
         "sanmai/pipeline": "^7.2",
-        "sebastian/diff": "^4.0 || ^5.0 || ^6.0 || ^7.0 || ^8.0",
+        "sebastian/diff": "^4.0 || ^5.0 || ^6.0 || ^7.0 || ^8.0 || ^9.0",
```
staabm, line 66: we may drop old versions 4,5,6 since 7.x is the first one supporting php 8.3+, which is what infection supports

staabm: (same can be said for symfony 6 & 7)

staabm: hmm.. I might be wrong.. its only the min php version which got raised in those versions

sidz:
> having ---Original / --New for each mutant, where we can have thousands of them, doesn't make sense and makes the diff and the whole log file/output more bigger and noisy, especially given these lines are highlighted

And this is an issue we can't solve without writing own OutputBuilder.

Same for:
> And, looking into the first image from your message, we could also get rid of @@ @@ which doesn't make sense for us as well. Even @@ -117,7 +117,7 @@ doesn't make sense for our diffs, we are interested in the minimal source code changes diff only IMO

> if you are blocked from upgrading because of this, we can merge your PR as is, make a release and unblock you, and then create our one differ at some point that removes this header.

Currently it doesn't blocking me. 


As an idea: we could technically use https://github.com/localheinz/diff as it's a fork of Sebastian `diff` and still use `UnifiedDiffOutputBuilder`.

FYI: `UnifiedDiffOutputBuilder` still writes `@@` into the output.

sanmai (approved):
Goods stuff; agents will appreciate a clean patch format with no guesswork around where the change is.

It wasn't a big deal for the likes of Claude/Codes, but for lesser LLMs like Qwen/Gemma every bit of detail helps.

sidz:
I'd agree with @sanmai that we will have standard diff output everybody knows how to work with. I believe that's the reason why Sebastian decided deprecate and remove `UnifiedDiffOutputBuilder`.

tests/phpunit/Differ/DifferTest.php
```diff
@@ -112,7 +112,9 @@ public function echo(): void
 
                 PHP,
             <<<'PHP'
-                @@ @@
+                --- Original
+                +++ New
```
theofidry: Rather than `New`, what do you think of having `Mutated`?

src/Differ/Differ.php
theofidry: This requires to update the Mago baseline too. But actually it's unrelated to the PR and begs the question of why it was there in the first place.

I propose to extract it to https://github.com/infection/infection/pull/3299 (so you would only need to update your PR), but if you prefer you can close it and update the baseline in your PR instead.

theofidry (approved):

tests/phpunit/Differ/DifferTest.php
sidz: Fine with me. I suggested `Mutated` as an option in the opened issue: https://github.com/infection/infection/issues/3295#issuecomment-4762109455

I'm ok to change if everybody agree.
cc @maks-rafalko, @sanmai

maks-rafalko:
I don't have strict objections _right now,_ but if we merge this one, I will definitely look into copying (with license) this diff output builder from differ package and make those lines conditional.

> Goods stuff; agents will appreciate a clean patch format with no guesswork around where the change is.
> 
> It wasn't a big deal for the likes of Claude/Codes, but for lesser LLMs like Qwen/Gemma every bit of detail helps.

I'm still not convinced TBH we need this addition:

```
--- Original
+++ New
```

that's for sure only my opinion, by I'm so tired of the extra content we need to read because of AI. Moreover,

- we didn't have these lines before, and from what I experienced - we didn't have any issues so far with AI reading our diffs with the lack of these 2 lines
- this is a noise that brings nothing to a human when you quickly read the output, and I don't think it brings anything to the AI. If it does - do we have examples where a model fails to parse this diff?

Even if we need it for AI, this is better to display conditionally, detecting AI agent. For a human, I don't think it's needed.

image

my eyes immediately stuck at highlighted `--- Original` / `+++ New` for this output, which is not what we are looking for in this diff. Our attention should be focused on the real difference in the code.

sanmai:
TBH for my own consumption I'd prefer an even more compact view, with zero lines of context. So just two lines. And I think it could be done with a simple regex filter like `/^[+-]/` so you would not even need to for anything.

```diff
1) src/SourceClass.php:11 [M] Increment [ID] 123123hex
- $a++;
+ $a--;

2) src/SourceClass.php ....
```
And I'd say let the whitespace go as well. The first line already shows the position (`:11`) in file, and that's all you need to locate the change in the file.

theofidry:
Currently this output is re-used for the console too, not just file and I think the primary consumer is humans.

Whilst optimising output for the AI is a concern, I think we should explore this separately, maybe it's own RFC. There is indeed various other improvements that could be considered, e.g. the usage of [ECH sequence](https://github.com/symfony/symfony/pull/63862) or adapting the log output in various places.

sanmai:
FWIW I'm not super concerned with AI. What we have here - it works for humans and AI alike. But if we can make it work better for our biological brains - even better.

maks-rafalko:
> TBH for my own consumption I'd prefer an even more compact view, with zero lines of context. So just two lines. And I think it could be done with a simple regex filter like `/^[+-]/` so you would not even need to for anything.

btw, there is a `DiffOnlyOutputBuilder` builder that does exactly that

image

I personally wouldn't use it (as I like a bit of context lines before/after), but just saying that it can be, for example, configurable in a config file if one prefers this format. As, reflecting on our discussion, how the diff should look like is a matter of taste

maks-rafalko:
thanks @sidz, as discussed here (https://github.com/infection/infection/pull/3300#discussion_r3473344548) let's merge it

PR #3300, opened 2026-06-24
feat: Add our own local unified diff output builder

theofidry:
## Description

This PR introduces our own `UnifiedDiffOutputBuilder`. It is taken from [sebastian/diff](https://github.com/sebastianbergmann/diff), with a few minor changes to satisfy PHPStan or greater code coverage.

This reverts the switch of differ made in #3298, reverting any change for the user in terms of output.

## Changes

- Adds `Infection\Differ\UnifiedDiffOutputBuilder`.
- Wires `Infection\Container\Container` to instantiate `SebastianBergmann\Diff\Differ` with Infection's local `UnifiedDiffOutputBuilder`.
- Excludes `src/Differ/UnifiedDiffOutputBuilder.php` from PHP-CS-Fixer so the copied upstream license header is not replaced by Infection's project header.
- Adds `UnifiedDiffOutputBuilderTest`

## Differences from the original `sebastian/diff`

The new `Infection\Differ\UnifiedDiffOutputBuilder` is intentionally based on the legacy `SebastianBergmann\Diff\Output\UnifiedDiffOutputBuilder`, but it is not a byte-for-byte copy. The behavioural goal is to preserve the legacy output Infection relied on; the local changes below are Infection-specific adaptations.

The changes:

- The original implementation extends `AbstractChunkOutputBuilder`. However, `AbstractChunkOutputBuilder` was also removed in v9. Instead, this abstract class has been inlined.
- Remove support for the header, which we did not use.
- Usage of the `Safe\*` functions instead of the non-safe ones.
- Added a few extra tests for better test coverage.
- Moved some constant values as constants.
- The original `::getDiff(array $diff)` assumes the diff array shape produced by `SebastianBergmann\Diff\Differ`. To satisfy PHPStan, this `array` shape is validated in our side.
- Some other minor style issues.

There is no behaviour drift intended.

## License concerns

The original file is under BSD-3-Clause, like our project, so I inlined `sebastian/diff`'s license in that file and excluded that file from PHP-CS-Fixer to not accidentally remove the license.

## Reviewer notes

This PR does not aim to erase the considerations raised in #3298. I think there is a valid point about maintainability or user familiarity about using a more well-known or established diff output or tweaking the output for AI.

However, it is a problem orthogonal to the `sebastian/diff` v9 compatibility. This PR is only a fix to allow the upgrade without the user facing any change, nothing more.

## Related issues

- https://github.com/infection/infection/pull/3298
- https://github.com/infection/infection/issues/3295

### DIFF
--- a/src/Differ/UnifiedDiffOutputBuilder.php
+++ b/src/Differ/UnifiedDiffOutputBuilder.php
@@ -1,12 +1,38 @@
-<?php declare(strict_types=1);
-/*
- * This file is part of sebastian/diff.
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
  *
- * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ * Copyright (c) 2002-2024, Sebastian Bergmann
+ * All rights reserved.
  *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
+
+declare(strict_types=1);
+
 namespace Infection\Differ;
 
 use function array_splice;
--- a/src/Differ/UnifiedDiffOutputBuilder.php
+++ b/src/Differ/UnifiedDiffOutputBuilder.php
@@ -47,7 +47,6 @@ use function Safe\fclose;
 use function Safe\fopen;
 use function Safe\fwrite;
 use function Safe\stream_get_contents;
-use function str_ends_with;
 use function substr;
 use InvalidArgumentException;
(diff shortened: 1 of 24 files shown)


tests/phpunit/Differ/UnifiedDiffOutputBuilderTest.php
```diff
@@ -0,0 +1,384 @@
...
+
+        $this->assertSame('', $builder->getDiff([]));
+    }
+
+    public function test_it_keeps_the_configured_header_for_empty_diffs(): void
+    {
+        $builder = new UnifiedDiffOutputBuilder("--- Original\n+++ New\n");
+
+        $this->assertSame(
+            "--- Original\n+++ New\n",
```
maks-rafalko: when it's now an implementation that we control, why would we need the support for this header? As well as `@@ @@` that is completely useless IMO

src/Differ/UnifiedDiffOutputBuilder.php
```diff
@@ -0,0 +1,339 @@
...
+use InvalidArgumentException;
+use SebastianBergmann\Diff\Differ;
+use SebastianBergmann\Diff\Output\DiffOutputBuilderInterface;
+
+/**
+ * Builds a diff string representation in unified diff format in chunks.
+ *
+ * @internal
+ */
+final readonly class UnifiedDiffOutputBuilder implements DiffOutputBuilderInterface
```
maks-rafalko, line 60: we can update `composer.json/lock` and remove

```
"sebastian/diff": "^4.0 || ^5.0 || ^6.0 || ^7.0 || ^8.0"
```

altogether

maks-rafalko: oh, wait, we use `Differ` as well from that package.

Then the comment in https://github.com/infection/infection/blob/a642bc951addf00d30c4ed45faf1b5c6d41e1cf6/src/Differ/Differ.php#L50 is no longer valid

>  * Tiny adapter for Sebastian's differ. **It is technically no longer necessary as of now,** but there
> * has been several changes with Sebastian's differ hence having an adapter against which we test
> * the behaviour is always a nice to have.

the behavior of `diffToArray()` is not the same as in original differ, so we do have a different implementation.

theofidry: >we can update composer.json/lock and remove sebastian/diff

It would require a few more changes, but in the current state, we still use a few symbols and it was not the goal to remove it entirely.

>the behavior of diffToArray() is not the same as in original differ, so we do have a different implementation.

WDYM? The behavior should be the same.

tests/phpunit/Differ/UnifiedDiffOutputBuilderTest.php
theofidry: I focused on copying `UnifiedDiffOutputBuilder` without trying to alter much from it. Some changes were necessary to satisfy PHPStan and code coverage, but otherwise no change.

But I do not mind simplifying it immediately.

src/Differ/UnifiedDiffOutputBuilder.php
maks-rafalko: > WDYM? The behavior should be the same.

- we have a wrapper `Differ` around package's `Differ` class
- in that file we have a comment, `... It is technically no longer necessary as of now, ...`
- we override the behavior of `diffToArray()`
- I'm saying that if I remove our custom logic from `diffToArray()`, tests fail -> so we do technically require this class, thus the comment is incorrect now

> It would require a few more changes, but in the current state, we still use a few symbols and it was not the goal to remove it entirely.

should we allow ` || ^9.0` then as it's done in other PR? that was the main purpose of that PR

theofidry: >I'm saying that if I remove our custom logic from diffToArray(), tests fail -> so we do technically require this class, thus the comment is incorrect now

Ah I see, but that's actually completely related to the change of this PR. Will update it.

>should we allow  || ^9.0 then as it's done in other PR? that was the main purpose of that PR

I was planning to do it in a separate PR but yes that's the goal.

sidz: > I was planning to do it in a separate PR but yes that's the goal.

I'd suggest to merge https://github.com/infection/infection/pull/3298 first and after that rebase this PR against latest main.

tests/phpunit/Differ/DifferTest.php
```diff
@@ -211,9 +209,7 @@ public function echo(): void
                 15
                 PHP,
             <<<'PHP'
-                --- Original
-                +++ Mutated
-                @@ -4,7 +4,7 @@
+                @@ @@
```
maks-rafalko, line 212: I would also remove this useless characters. Fine to be delayed though

maks-rafalko (approved):

PR #3308, opened 2026-06-27
Deprecate `--filter` options in favor of new positional arguments for filtering source

maks-rafalko:
- [x] Doc PR https://github.com/infection/site/pull/304

There was an interesting idea during code review of https://github.com/infection/infection/pull/3284 - to deprecate `--filter` in favor of `infection source1 test1 ... sourceN testN` syntax, which makes sense to me.

This PR deprecates `--filter` by showing a yellow warning:

image

### DIFF
--- a/src/Command/Option/SourceFilterOptions.php
+++ b/src/Command/Option/SourceFilterOptions.php
@@ -35,6 +35,7 @@ declare(strict_types=1);
 
 namespace Infection\Command\Option;
 
+use const E_USER_DEPRECATED;
 use Infection\CannotBeInstantiated;
 use Infection\Configuration\SourceFilter\IncompleteGitDiffFilter;
 use Infection\Configuration\SourceFilter\PlainFilter;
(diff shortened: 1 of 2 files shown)


src/Command/Option/SourceFilterOptions.php
```diff
@@ -126,9 +126,15 @@ public static function get(IO $io, array $positionalPaths = []): PlainFilter|Inc
...
 
-    private static function getPlainFilter(InputInterface $input): ?PlainFilter
+    private static function getPlainFilter(IO $io): ?PlainFilter
     {
-        $value = trim((string) $input->getOption(self::PLAIN_FILTER_NAME));
+        $value = trim((string) $io->getInput()->getOption(self::PLAIN_FILTER_NAME));
+
+        if ($value !== '') {
+            $io->warning(
+                'The "--filter" option is deprecated since 0.34.0 and will be removed in 0.35.0. Use positional arguments instead: infection <filter>',
```
theofidry: Maybe we can be more vague about when we will remove it? Allowing it on more than one version makes transitioning easier.

theofidry (approved):

src/Command/Option/SourceFilterOptions.php
maks-rafalko: I thought that this is quite an easy upgrade so it shouldn't be too hard for a human, not saying for an AI.

```diff
- infection --filter=src/A.php,src/B.php
+ infection src/A.php src/B.php
```

but ok, I will remove 0.35 from the warning message

PR #3309, opened 2026-06-28
test: Refactor positional path classifier tests

theofidry:
## Description

Follow-up to my comment in https://github.com/infection/infection/pull/3284#discussion_r3486721470.

The previous test covered a simple `PositionalPathsClassifier` API, but its structure made the intent harder to follow than necessary:

- The same API surface was tested partly through dedicated test methods and partly through data providers, without a clear reason for choosing one form over the other.
- The `::classify()` test helper obscured the act step. Hiding repetitive setup can improve readability, but hiding the SUT call makes it harder to see which behaviour the test exercises.
- The assertions checked individual `ClassifiedPaths` properties with several `::assertSame()` calls. The canonical assertion style for these value objects is to compare the expected and actual objects with `::assertEquals()` (see https://github.com/infection/infection/pull/2486).
- The filesystem mock used case-specific logic that did not reflect the behaviour being tested. Using `InMemoryFileSystem` makes the fixture closer to a real implementation and easier to reason about. This required adding the missing directory-related operations, which are small and included in this PR.
- The test built an arbitrary `SchemaConfiguration` instance directly. Using `SchemaConfigurationBuilder` is simpler.
- The new `Scenario` object groups each case's inputs and expected result. I think this improves readability by keeping each scenario self-contained, but I am happy to remove it if the additional type feels heavier than the test needs.
- I kept separate data providers for successful and failing paths because that matches @maks-rafalko's preferred style. They could otherwise be merged later; personally, I prefer a single data provider.

## Changes

- Adds directory support to `InMemoryFileSystem`, including `mkdir()` and `isReadableDirectory()`.
- Moves `PositionalPathsClassifierTest` under a dedicated namespace directory so that the `Scenario` object can live alongside it.
- Reworks the positional path classifier cases around a `Scenario` value object and `InMemoryFileSystem` instead of filesystem mocks.
- Uses `SchemaConfigurationBuilder` through the scenario setup rather than hand-building schema configuration in the test.


tests/phpunit/Configuration/PositionalPathsClassifierTest.php
```diff
@@ -1,682 +0,0 @@
-<?php
```
theofidry, line 1: ofc GitHub would not render it as a moved file...

maks-rafalko (approved):
Oh, it's a good proof that this is just a matter of taste. I see literally 0 reasons of why these tests are better than they were before 😄

TL;DR: Since this is minor and you already spent your time, I have no objections to merge it

----

My bet that the good part of description was generated by AI, many of the points just don't make sense to me. Apologies if not 😄 

For example:

> The same API surface was tested partly through dedicated test methods and partly through data providers, without a clear reason for choosing one form over the other.

from what times this is considered bad?

> The filesystem mock used case-specific logic that did not reflect the behaviour being tested. Using InMemoryFileSystem makes the fixture closer to a real implementation and easier to reason about.

in which way this is "closer to a real implementation" if it's not a real implementation?

> The test built an arbitrary SchemaConfiguration instance directly. Using SchemaConfigurationBuilder is simpler.

for me it looks much more verbose, not sure where it is simpler in this particular test.

No need to answer, I'm just thinking outloud.

theofidry:
>My bet that the good part of description was generated by AI, many of the points just don't make sense to me. Apologies if not 😄

None actually :D

>> The same API surface was tested partly through dedicated test methods and partly through data providers, without a clear reason for choosing one form over the other.

> from what times this is considered bad?

I find it bad on two levels:

- Using multiple methods for testing something where the only changes is the input/output, is more boilerplate than with data providers. So IMO, also less nice to read.
- When you are looking for a test or to add a test, you do not know where to look. You can not care and randomly add a test, but personally I find it confusing.

> in which way this is "closer to a real implementation" if it's not a real implementation?

In the same way using a virtual FS is closer to using a FS than mocking the FS.

The mock used previously for example, was written in a way that validates the implementation, not describing a behaviour. In this case, it was "if the path starts to match an existing path, then it is a readable file and directory".

Whilst I would not call `InMemoryFileSystem` a  virtual FS because of how little of the subset it implements, and because it is kept simple since the only need is for test purposes (e.g. it does not try to mimic any FS boxing that the Symfony FS does which can throw `IOException` under certain circumstances), its behaviour is a lot closer to how FS acts – and its behaviour can be, and is, tested.

>> The test built an arbitrary SchemaConfiguration instance directly. Using SchemaConfigurationBuilder is simpler.

> for me it looks much more verbose, not sure where it is simpler in this particular test.

Ok, we must not be talking of the same thing I think. I was specifically referring to using:

```php
    private static function createSchema(array $sourceDirectories): SchemaConfiguration
    {
        return SchemaConfigurationBuilder::withMinimalTestData()
            ->withPathname('/project/infection.json5')
            ->withSource(new Source($sourceDirectories, []))
            ->build();
    }
```

Instead of:

```php
    /**
     * @param list<non-empty-string> $sourceDirectories
     */
    private function createSchema(array $sourceDirectories): SchemaConfiguration
    {
        return new SchemaConfiguration(
            pathname: '/project/infection.json5',
            timeout: null,
            source: new Source($sourceDirectories, []),
            logs: Logs::createEmpty(),
            tmpDir: null,
            phpUnit: new PhpUnit(null, null),
            phpStan: new PhpStan(null, null),
            mago: new Mago(null, null),
            ignoreMsiWithNoMutations: null,
            minMsi: null,
            minCoveredMsi: null,
            timeoutsAsEscaped: null,
            maxTimeouts: null,
            mutators: [],
            testFramework: null,
            bootstrap: null,
            initialTestsPhpOptions: null,
            testFrameworkExtraOptions: null,
            testFrameworkExtraArgs: null,
            staticAnalysisToolOptions: null,
            threads: null,
            dotsPerRow: null,
            staticAnalysisTool: null,
        );
    }
```

For `Scenario` in this case though, I do agree it's a lot more debatable. Personally I do prefer:

```php
        yield 'source folder and test file' => [
            $baseScenario
                ->withPaths(['src/Service/', 'tests/Unit/Service/MailerTest.php'])
                ->withExistingFiles(['/project/tests/Unit/Service/MailerTest.php'])
                ->withExpected(new ClassifiedPaths(['src/Service/'], ['tests/Unit/Service/MailerTest.php'])),
        ];
```

Over:

```php
        yield 'source folder + test file' => [
            ['src/Service/', 'tests/Unit/Service/MailerTest.php'],
            ['src/Service/'],
            ['tests/Unit/Service/MailerTest.php'],
        ];
```

Because I immediately can tell what is what, but for <= 4 parameters I can live without.

maks-rafalko:
> Because I immediately can tell what is what, but for <= 4 parameters I can live without.

you can use named arguments if this is a problem

```php
yield 'source folder + test file' => [
            'someName1' => ['src/Service/', 'tests/Unit/Service/MailerTest.php'],
            'someName2' => ['src/Service/'],
            'someName3' => ['tests/Unit/Service/MailerTest.php'],
        ];
```

theofidry:
I do occasionally, I find it really annoying to use though as any change/rename is harder. It also doesn't help in some other cases like for `ConfigurationFactory`

PR #3312, opened 2026-07-01
feat(test-framework): Introduce the new TestFramework contract

theofidry:
## Description

Introduce a new internal `TestFramework` contract. It implements the proposed new API for `TestFrameworkAdapter` described in https://github.com/infection/infection/issues/2863. Whether the name is kept or `TestFrameworkAdapter` is updated remains open, but does not need to be addressed in this PR.

The goal of this PR is to introduce this new interface _with minimal changes_. It does so by introducing `LegacyTestFrameworkBridge`, allowing adapters to be migrated progressively from the legacy API to the new one. This lets Infection use the new API without requiring any other adapter changes.

Three important notes:

- This PR introduces new violations of the guards from https://github.com/infection/infection/pull/3269. These are legitimate and should be fixed, but are left out of this PR to keep it focused. They will be addressed in follow-up PRs.
- Only the `TestFrameworkAdapter` path is ported in this PR. I will port `StaticAnalysisToolAdapter` in a separate PR.
- The implementation of `LegacyTestFrameworkBridge` has been kept as close to the original code as possible, although further changes and improvements are planned.

## Changes

- Adds `TestFramework`, `InitialRunResults`, and contract-level throwable types.
- Adds `LegacyTestFrameworkBridge` for the existing adapter-based implementation.
- Updates `Engine` to depend on `TestFramework` instead of coordinating the adapter, coverage checker, initial test runner, and mutant extra-options filtering directly.
- Updates `MutationTestingRunner` to ask the configured `TestFramework` to build mutant process containers.
- Updates `MemoryLimiter` to consume `InitialRunResults` instead of querying adapter memory usage itself.
- Registers the legacy bridge as the `TestFramework` service in the container.

## Breaking Changes

None identified from the diff. The new contract is marked internal, and the legacy adapter path remains wired through `LegacyTestFrameworkBridge`.

## Reviewer Notes

Please focus on the new responsibility boundaries:

- `Engine` now delegates framework requirements and initial runs to `TestFramework`.
- `MutationTestingRunner` materializes mutants before delegating process creation to `TestFramework::test()`.
- `LegacyTestFrameworkBridge` keeps the current adapter semantics, including skip-initial-tests coverage checks, initial-run-only option filtering, and unknown memory usage normalization.
- `MemoryLimiter` now treats `null` memory usage as unknown and skips limiting in that case.

Also note that the proposed API is not definitive. The one proposed in the RFC gives the general direction, but it may be changed as we work out the details.

## Related issues


tests/phpunit/Resource/Memory/MemoryLimiterTest.php
```diff
@@ -146,18 +124,6 @@ public static function memoryLimitProvider(): iterable
             20.,
             40.,
         ];
-
-        // This is a silly case. We do not care about it, it just to capture the current behaviour
-        yield 'negative' => [
```
theofidry, line 153: those cases are no longer possible due to the validation happening in `InitialRunResults` (`positive-float|null`)

src/TestFramework/Contracts/InitialRunResults.php
```diff
@@ -0,0 +1,54 @@
...
+declare(strict_types=1);
+
+namespace Infection\TestFramework\Contracts;
+
+use Webmozart\Assert\Assert;
+
+/**
+ * @internal
+ */
+final readonly class InitialRunResults
```
theofidry, line 45: Note to the reviewers: this is an alternative to:

```php
namespace Infection\AbstractTestFramework;

interface MemoryUsageAware
{
    /**
     * Reports memory used by a test suite, or -1 if it cant detect the memory usage.
     */
    public function getMemoryUsed(string $output): float;
}
```

Indeed the latter caused two issues:

- First the responsibility falls on the adapter. This means as we have a bridge between the two APIs, there is no easy way to do so (unless we duplicate the adapters, which is not ideal). Indeed, right now we have `LegacyTestFrameworkBridge` which takes `TestFramework`. Whether it should implement `MemoryUsageAware` depends of the adapter it decorates.
- The interface relied on the output. Whilst it's true for PHPUnit, a test framework could declare this result a different way, especially if it's an in-memory process of some kind.

Instead having this value object nicely solves both issues.

Additionally, if we need extra properties here in the future to add more test framework capabilities (e.g. the timeout is different from a test framework to another), we can easily do so by adding an optional property (it needs to be optional otherwise existing code would need to be updated, i.e. this would be a BC break).

src/Resource/Memory/MemoryLimiter.php
```diff
@@ -55,9 +54,11 @@ public function __construct(
     ) {
     }
 
-    public function limitMemory(string $processOutput, TestFrameworkAdapter $adapter): void
+    public function limitMemory(?InitialRunResults $initialRunResults): void
     {
-        if (!$adapter instanceof MemoryUsageAware
+        $memoryUsage = $initialRunResults?->memoryUsage;
+
+        if ($memoryUsage === null
```
theofidry, line 61: this `null` may be extractable

tests/phpunit/EngineTest.php
```diff
@@ -731,104 +363,62 @@ public function test_application_execution_was_finished_is_dispatched_when_min_m
...
+            ->with($this->isInstanceOf(ApplicationExecutionWasFinished::class))
+        ;
 
         $this->expectException(MinMsiCheckFailed::class);
 
-        $engine->execute();
+        $this->createEngine($config)->execute();
+    }
+
+    private function createEngine(
```
theofidry, line 389: should extract this change, it makes the diff of this test very hard

theofidry: done so in https://github.com/infection/infection/pull/3313

sanmai (commented):
I like where this is going. I wonder if there's a way to start splitting it away, at least in some part.

theofidry:
> I like where this is going. I wonder if there's a way to start splitting it away, at least in some part.

That is the split version 😄 

There is _a lot_ more, but I tried to make the minimal possible amount of change. This PR API change is significant, but I hoped to keep it understandable bu not altering any behaviour at this point.

src/TestFramework/Contracts/InitialRunResults.php
```diff
@@ -0,0 +1,58 @@
...
+declare(strict_types=1);
+
+namespace Infection\TestFramework\Contracts;
+
+use Webmozart\Assert\Assert;
+
+/**
+ * Carries the observable outcome of a test framework initial run for later orchestration steps.
+ *
+ * @internal
```
sanmai, line 43: So it won't be internal after all?

theofidry: Indeed. Nothing in `src/TestFramework/Contracts` should be as it will be extracted to the `infection/abstract-testframework-adapter` package.

I added `@internal` because it was less noisy (you don't need to update `ProjectCodeProvider` or add more baseline violations), but if that is too confusing I can adopt any of those two alternatives

sanmai: Makes sense, just wanted to confirm it.

src/TestFramework/Contracts/TestFramework.php
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * A test framework is the tool that will be used to evaluate mutations to check if they are
+ * covered.
+ *
+ * It can be a standard test framework such as PHPUnit, PhpSpec, a static analyser like PHPStan
+ * or Psalm or something else entirely!
+ *
+ * @internal
+ */
+interface TestFramework
```
maks-rafalko, line 53: the current `TestFrameworkAdapter` has `getVersion()` method that is used, for example, in this output line:

```
Running initial tests with PHPUnit version 9.6.34
```

How is it supposed to work with the new interface?

From the big picture view, it could be done only by 2 ways

- either by supporting the same `getVersion()` method in a new `TestFramework` 
- or by encapsulating the output logic to `TestFramework` implementation, which requires passing the some IO object. I don't think this option is a good one

so IMO `getVersion()` must be here, otherwise I don't understand how it will work.

I checked out this branch and 

```
Running initial tests with PHPUnit version 9.6.34
```

still uses `TestFrameworkAdapter` instance, not `TestFramework` - so this PR doesn't actually shows all the problems we have to fix/consider

src/TestFramework/Contracts/TestFramework.php
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public function executeInitialRun(): InitialRunResults;
+
+    /**
+     * Evaluates the Mutant. Some test frameworks may be able to do this in-memory, e.g.
+     * Psalm, or it requires to launch a process in which case the process execution is
+     * delegated to an orchestrator. A process encapsulates how the result of the process
+     * is interpreted.
+     */
+    public function test(Mutant $mutant): MutantProcessContainer;
```
maks-rafalko, line 90: this API can be a real problem, to be honest.

```
Infection\TestFramework\Contracts\TestFramework should not depend on Infection\Mutant\Mutant
```

this is, unfortunately, a very dangerous error reported by PHPStan/PHPAt.

Imagine we move PHPUnit (or any other) adapter to its own package. `Mutant` is not there - it's here, in `infection/infection`.

Moreover, `Mutant` currently depends on `Mutation` - yet another class that should be available in the adapter package.

(!) `Mutation` in its turn depends on

```
use PhpParser\Node;
use PhpParser\Token;
```

So, basically now we say that TestFramework depends on PHPParser, while previously it wasn't the case.

I think this is the most critical question right now (=blocker).

We should probably decouple `Mutant` from `TestFramework` by some DTO.

This is what currently needed inside a `TestFramework::test(Mutant $mutant)` method under the hood

https://github.com/infection/infection/blob/1fa32e56834707ae42d302898778ff8efc4f9367/src/Process/Factory/MutantProcessContainerFactory.php#L77-L80

and 

https://github.com/infection/infection/blob/1fa32e56834707ae42d302898778ff8efc4f9367/src/Mutant/TestFrameworkMutantExecutionResultFactory.php#L59-L78

I didn't dig into it to further, but this definitely requires a lot of thinking / experimenting, cause otherwise our extracted adapter will need a lot of internals from `infection/infection`

----

After some thinking, probably instead of trying to do the very minimum changes in this PR ignoring PHPAt violations in the baseline, we need to do a PoC where none violations are left in a baseline - proving we are correctly decouple `infection/xxx-adapter` from `infection/infection`

src/TestFramework/LegacyTestFrameworkBridge.php
```diff
@@ -0,0 +1,144 @@
...
+final readonly class LegacyTestFrameworkBridge implements TestFramework
+{
+    public function __construct(
+        private TestFrameworkAdapter $adapter,
+        private ConsoleOutput $consoleOutput,
+        private CoverageChecker $coverageChecker,
+        private InitialTestsRunner $initialTestsRunner,
+        private Configuration $config,
+        private MutantProcessContainerFactory $processFactory,
+        private TestFrameworkExtraOptionsFilter $testFrameworkExtraOptionsFilter,
```
maks-rafalko, line 66: As I understood, we have boundary checks only for `src/TestFramework/Contracts`.

But this file is not there, it's in `src/TestFramework/`.

As a result, we don't have reported violations for this class - it now depends on `Configuration`, `MutantProcessContainerFactory`, `TestFrameworkExtraOptionsFilter` etc.

Imagine for a moment that it is extracted to a separate package - this won't work.

Possible 2 action points

- to extend PHPAt rules to check this class too
- to fix any violations are this is a PoC of what would be in the `infection/phpunit-adapter` - so it must not depend on `infection/infection` classes

src/Engine.php
```diff
@@ -106,17 +97,15 @@ public function execute(): void
...
 
         /*
          * Limit the memory used for the mutation processes based on the memory
          * used for the initial test run.
          * This is done AFTER static analysis to avoid restricting PHPStan's memory.
          */
-        if ($initialTestSuiteOutput !== null) {
-            $this->memoryLimiter->limitMemory($initialTestSuiteOutput, $this->adapter);
-        }
+        $this->memoryLimiter->limitMemory($initialRunResults);
```
maks-rafalko, line 108: it's not related to this particular PR, but I have a feeling that this should be test framework adapter's responsibility now to limit memory. Especially as you say it can vary from one TF to another (PHPUnit has its own limit, PHPStan has its own - much bigger)

src/TestFramework/Contracts/TestFramework.php
maks-rafalko, line 53: if our goal is Multi-framework support, as it's stated in the RFC, then IMO we should understand if this interface suits the needs of static analysis tool. 

With the current state of PR, we can't prove it. And I'm pretty sure there will be more questions / PHPAt violations when we touch SA interface.

maks-rafalko, line 53: Since our goal is Multi-framework support (I hope this is the main driver, not the refactoring and non-important limitations of the current design of TF existing interface), then:

did you consider implementing Multi-framework support first, *before* touching existing interfaces?

If so - were there any blockers? What are they? 

If not - why? 

By asking these questions, I would like to understand and find *exact* problems that currently block us and force us doing this refactoring. So it's like doing the opposite to this PR - not refactoring the existing code, but trying to implement a Feature and understand what really limits us in the current design.

You may not like it, for sure, but this could potentially save all of us hundreds of human-hours of work, as changing the API, designing the packages with clear boundaries, extracting reusable packages is an extremely hard activity.

To be clear: I'm not saying we don't need to do it, I'm just deliberately acting as a "skeptic"`*` here to find the right answers and prevent the work we potentially don't need yet: for example, to avoid preliminary over-architecting. 

`*` being a skeptic here is a role, a "hat", to produce more questions like "Do we really..?" / "Have we considered...?" / "What if...?" / "Can we avoid...?".

theofidry: > the current TestFrameworkAdapter has getVersion() method that is used, for example, in this output line:
> Running initial tests with PHPUnit version 9.6.34
> How is it supposed to work with the new interface?

That is a good question, and it touches a bit of a broader question: how should the test framework log statements and what level of control we want to give it?

And I do not have a clear answer for it.

What I would like to do is give a `Psr\Logger` instance to the test frameworks to allow them to log various things (I am mostly thinking of debug statements for... well debugging purposes). A common feedback I've had with Infection is that sometimes something looks funky with PHPUnit, but users have no way to easily see or debug things. Since then we improved a bit things, e.g. being able to execute the initial tests alone, but I can see further improvements there still.

Now with a logger passed, the question is the degree of control. The less control we give, the more specific `TestFramework` needs to be because it needs to be aware of more details (e.g. here name and version to display that message) or knowing the progress to display a progress bar instead of streaming the raw output.
The less control, the less we need to be aware of the details, which means more freedom but as a drawback, also means you could more easily have very different outputs from one test framework to another.

Personally I'm not quite decided yet. I'm leaning towards the latter but it's not definite, and I think if what we need is still expose `::getVersion()` then no big deal.

> still uses TestFrameworkAdapter instance, not TestFramework - so this PR doesn't actually shows all the problems we have to fix/consider
> so this PR doesn't actually shows all the problems we have to fix/consider

Correct! It is also still using the "legacy" `PhpUnitAdapter` and `InitialTestsRunner`. But the goal of this PR is to introduce a first version of the API and backport the current adapter to it _with minimal changes_. There is still quite a few changes to do after that.

Proving that this new API can work was preparatory work done before and that was used to create the RFC. A public, a bit more advanced version is available in #3290.

That being said, I do not have a branch or PR where I solved 100% of the problems and everything is ready and there is no more question to answer to. Simply because getting there is a ton of work and I'm afraid to go there and not be able to break reviewable pieces apart later or easily integrate feedback happening halfway. However, if you do feel strongly about viewing how a specific change can work, I'm happy to add it to the PoC for you to assess it.

src/TestFramework/Contracts/TestFramework.php
theofidry: > Imagine we move PHPUnit (or any other) adapter to its own package. Mutant is not there - it's here, in infection/infection.
> Moreover, Mutant currently depends on Mutation - yet another class that should be available in the adapter package.
> I think this is the most critical question right now (=blocker).

Indeed, and it is certainly not trivial, and this sort of problem is why I felt strongly about https://github.com/infection/infection/pull/3269 as they clearly surface this sort of issue in the baseline rather than being able to easily miss it during a review.

But back on `Mutant`. It can seem like a big problem, but I actually do not think it will be hard to fix. The reason why it seems like a big one is because `Mutant` sitting in `infection/infection`, was designed in a way that was very convenient to consume (it has a lot of information, more than maybe strictly needed) – and that was a valid choice, it allowed to keep a lot of things simple.

However, the information we need for `TestFramework` of `Mutant`, and what we expect the test framework to return is a lot less information:

- the tests
- the hash
- the original file path
- the materalised mutation file path

Tests is not a trivial one, but that's actually covered by https://github.com/infection/infection/issues/3082. We already have some of the tests API in the abstract package and I do not think it will require problematic things to be added there.

Likewise when we create `MutantExecutionResult`, the only things we really need from it are:

- the timings (`$mutantProcess->getFinishedAt() - $process->getStartTime()`)
- the process output
- the detection status

None of those things require a deep dependency hierarchy going back to PHP-Parser or other. 

> After some thinking, probably instead of trying to do the very minimum changes in this PR ignoring PHPAt violations in the baseline, we need to do a PoC where none violations are left in a baseline

See my stance on the PoC at the end of my previous reply, if my reply here does not convince you and it's a point you explicit want to see I'm happy to do so in that PoC PR.

src/TestFramework/LegacyTestFrameworkBridge.php
theofidry: > Imagine for a moment that it is extracted to a separate package - this won't work.

I think there is a misconception here, `LegacyTestFrameworkBridge` should not be extracted.

There is two path for the new API: either it's a new package, or it's a new version of `infection/abstract-testframework-adapter`. This decision is not interesting at this point so let's assume the latter, and that this version with the new API is `0.7.0`.

How I envisioned it here is that for PHPUnit specifically, its adapter implementation directly implements the new `TestFramework` API, no usage of `LegacyTestFrameworkBridge`. So when we create `infection/phpunit-adapter`, it can depend on `^0.7.0` and everything is fine.

For existing adapters however, it is a different story. We now need to do two things:

- Have a transition period where the users can use their existing adapter versions without things breaking completely apart.
- Release a new version where we migrate the adapters to the new API.

What API the adapter supports will be ensured by their dependency on infection/abstract-testframework-adapter`, and when we want to remove the bridge code, we will simply have to make Infection require `infection/abstract-testframework-adapter` being `^0.7.0` instead of previously `<previous-version> || ^0.7.0`.

In fact, we could even say we do not need a transition period because it's too much work and skip it entirely, in which case `LegacyTestFrameworkBridge` will be deleted sooner.

src/Engine.php
theofidry: Yup, definitely.

I tried to address this in https://github.com/infection/infection/pull/3290/changes/677c5f0322ee7ad5025a020128fa6d34bb52615b#diff-bdde882ce051503a559d314b02b72761e55a075bb96d3772abea731b054a714d but I did not assess yet if that was sufficient or working out as I expected.

I thought in the end how hard can it be, we pass `-d memory_limit=-1` for PHPStan and Mago, surely we can just do `-d memory_limit=X` if we need to.

So it would be an implementation detail of the test framework adapter and would not rely on Xdebug either.

src/TestFramework/Contracts/TestFramework.php
theofidry: My idea was to apply the same strategy as for the test framework adapter: https://github.com/infection/infection/pull/3290/changes/677c5f0322ee7ad5025a020128fa6d34bb52615b

So having `LegacyStaticAnalysisBridge` and port the existing PHPStan and Mago adapters to that API. Since we do not have any external package for those it's a lot more straightforward as it means as long as there is violations we can't extract those packages, but also that it does not hold back the release of the new API (but I would not recommend doing that).

In terms of violations the current issue is `LazyMutantProcessFactory`.

Solving this is actually simple, what is a lot less so is in the context of multi-test-framework support. Indeed, right now registering the SA mutant process is done via the container:

```php
MutantProcessContainerFactory::class => static function (self $container): MutantProcessContainerFactory {

    $mutantProcessKillerFactories = [];

    if ($config->isStaticAnalysisEnabled()) {
        $mutantProcessKillerFactories[] = $container->getStaticAnalysisToolAdapter()->createMutantProcessFactory();
    }

    $configuration = $container->getConfiguration();

    return new MutantProcessContainerFactory(
        $container->getTestFrameworkAdapter(),
        $configuration->processTimeout,
        $container->getMutantExecutionResultFactory(),
        $mutantProcessKillerFactories,
        $container->getConfiguration(),
    );
},
```

This is because unlike the test frameworks for which we have a `Factory`, currently the SA adapters are entirely part of the `Container`. If we want to extract the SA adapters (even in their current state, i.e. without this new `TestFramework` API), this design will not work and we instead will need something similar to the test frameworks.

So in the context of multi-test frameworks support, there is the challenge of creating a `MutantProcessContainer` that can carry each test framework `MutantProcess`.

Another part of the challenge is dealing with `symfony/process`. The simple solution would be to accept it as a dependency of `infection/infection/abstract-testframework-adapter`, the other would be to have our own API. I'm not exactly set on that either tbh. Note that this would already be a problem if we were to try to extract the SA API in its own package right now.

src/TestFramework/LegacyTestFrameworkBridge.php
maks-rafalko: > I think there is a misconception here, LegacyTestFrameworkBridge should not be extracted.

Yes, I understand that, but our current architecture is written in such way that our imaginary new PHPUnit adapter (right now it's the `LegacyTestFrameworkBridge`) does depend on Configuration, MutantProcessContainerFactory, TestFrameworkExtraOptionsFilter etc.

So at this point it's not clear who all of this will work. 

We said that we will keep PHPUnit implementation (the new one) in `infection/infection` for some time to simplify development, but by the moment we decide to extract it, it should be 100% ready for an extraction - so all the boundary violations should be resolved. It's not the case right now, and we don't check them for this class - that's my point.

I will look into PoC, probably there are some answers. But now I think I don't quite get why we don't check by PHPAt this `LegacyTestFrameworkBridge`.

TL;DR

- **(already fine)** now `src/TestFramework/Contracts` must not depend on `infection/infection` classes
- **(to do)** and I believe that any _concrete_ implementation (e.g. `LegacyTestFrameworkBridge`) must not depend on `infection/infection` classes as well, so we need to have those PHPAt violations - they will show us the real problems

src/TestFramework/Contracts/TestFramework.php
maks-rafalko: what about this line?

https://github.com/infection/infection/blob/1fa32e56834707ae42d302898778ff8efc4f9367/src/Process/Factory/MutantProcessContainerFactory.php#L93

`LegacyTestFrameworkBridge::test()` passes `Mutant` to 

```php
new MutantProcess(
    $process,
    $mutant, // <----------------------------- here
    $this->mutantExecutionResultFactory,
),
```

and this `$mutant` is then used in the 

https://github.com/infection/infection/blob/1fa32e56834707ae42d302898778ff8efc4f9367/src/Mutant/TestFrameworkMutantExecutionResultFactory.php#L60-L78

So also diff / mutated code / etc. are also required.

src/TestFramework/Contracts/TestFramework.php
theofidry: ~~I'm struggling a bit to come up with a comprehensive answer that will not end up being a 3 pages thing, but let's try.~~ Booo, failed.

>So it's like doing the opposite to this PR [...] You may not like it, for sure, but this could potentially save all of us hundreds of human-hours of work, as changing the API, designing the packages with clear boundaries, extracting reusable packages is an extremely hard activity.

Agreed on all front: that approach is the correct way to go about it and I absolutely dread the idea of doing the package design + extracting + updating those packages.

**In a vacuum**: adding the multi-test framework support _should_ be possible with the current API without any further changes. In its current state, it will suffer from the following limitations:

- It would require to address the `MemoryUsageAware` current usage. You noted earlier you think it should be a test framework specific thing instead of applying it globally. I agree, it is a necessity to address it first, otherwise we end up with the same problem as with PHPStan and Mago except that in this case it is a lot more awkward as the affected adapters are outside of `infection/infection`.
- It requires to address the `MutantProcessContainer` situation which I mentioned in https://github.com/infection/infection/pull/3312#discussion_r3836143224. That said, this is strictly code from `infection/infection`, so doable first.

And a big question mark for me is how all the code that is currently done outside of the test framework adapter will behave, e.g. checking the requirements, the error message when the user is missing a coverage report, how to passing the coverage report of both test frameworks, how to adjust the test framework options for each test framework, etc.

Now, I emphasised **in a vaccum**, for a couple of reasons:

- This is strictly about `TestFrameworkAdapter`. So if you want the same for the SA, you need to duplicate the work. Not a major deal, personally it would sadden me that we're willing to duplicate yet so much code rather than addressing the problem at the root, but I could live with it.
- https://github.com/infection/infection/issues/3082. If we talk about priority, I think the latter is **much more important** than multi-test framework support, for two reasons:
  - It is a blocker for PhpSpec9 which does not use PHPUnit coverage anymore. Now to give the same level of skepticism, we could do without, but like we do for PHPUnit that implies having a PhpSpec specific `Tracer` implementation within `infection/infection`, which is doable, but not without tradeoffs (especially from a testability PoV – it means bringing back stuff that is otherwise done in the PhpSpec adapter, like an e2e test). 
  - It opens up more performance features. Benjamin implemented a few improvements in the PHPUnit coverage report in PHPUnit 12.5 which we cannot leverage yet because of the current design. _I think_ it could be doable without, but would require a bunch of ugly patchwork because that means registering a different Tracer depending of which version of PHPUnit is registered.
- There _are_ a list of bugs in the each respective adapters (e.g. https://github.com/infection/infection/issues/3138 or https://github.com/infection/infection/pull/3474), including within `infection/infection`, which are still present. In theory they are all fixable, the "blocker" to me is that they all come from the same duplicated code because we decided back then that it was a better solution to just duplicate stuff. And I'm not really of going across all packages, adding the tests to capture those problems, fix them, for each individual package, rather than fixing definitely once and for all in a robust manner.

So all in all, there is various issues that do call for a change in `TestFrameworkAdapter`. Quite a few, individualy one may be unimportant and very edge case, but as you said, changing `TestFrameworkAdapter` is very annoying for everyone involved, hence the attempt to fix it in one go.

<hr/>

Now more as a conclusion. A lot of things here are from things I either had in mind forever (multi-test framework) or I things I actively worked on and tried to introduce for a while already (the PHPUnit coverage perf. improvements – https://github.com/infection/infection/issues/3082, which comes from https://github.com/infection/infection/issues/2816, better PHPUnit integration https://github.com/infection/infection/issues/2874). Until recently, there was no time pressure for such things: PHPUnit improvements were back then more of a long-term goal, and the outstanding design issues from the test framework adapters were not blockers.

And this specific PR is also soon 2 months old, and the initial draft of this design is 3-4 months old. This is a while ago already.

Last year I could talk with Sebastian and get some changes in which means some of the performance improvements for PHPUnit are available, from PHPUnit side, since 12.5.

PhpSpec9 is now actively being worked on and no longer relies on the PHPUnit coverage report, which means the current design of having test framework adapters expected to produce a PHPUnit coverage report no longer works (and this would have happened sooner if for example Testo required their own too).

I believe the changes specified in #2863 and #3082 are simply too ambitious to be done in a reasonable timeline, at the very least to be in time for PhpSpec9. And I also do not want to rush those changes: the current "way" is meant to allow changes in a way we can release at any time and adjust the API as we uncover more scenarios.

There was the possibility of doing a giant PoC to try to address all answers ahead of time, but I've had to do my share of big PoCs already, it's really tedious to backport things, and I'm also not keen on feature branches. We already have #3090 which is already in a state I cannot rebase anymore and whilst I'm still confident in backporting it (the changes are much more simple and scoped thankfully), the test framework API is an entire different beast.

So now I think we may need to do some patchwork first even if that means delaying some of this work.

I do want to be clear though: I do strongly feel that the number of patchworks (i.e. situational code that is here because of an API/abstraction issue) is already quite high. Doing all the desired feature with patchworks (i.e. no major refactoring before) will _definitely_ push it to very unreasonable levels and  IMO.

src/TestFramework/Contracts/TestFramework.php
theofidry: > So also diff / mutated code / etc. are also required.

It is necessary for creating the current `MutantExcecutionResultFactory` object, but the only data that actually comes from the test framework is:

> - the timings ($mutantProcess->getFinishedAt() - $process->getStartTime())
> - the process output
> - the detection status

So essentially I think the test framework will return a subset of that information and within `infection/infection` we re-combine it.

src/TestFramework/LegacyTestFrameworkBridge.php
theofidry: > but our current architecture is written in such way that our imaginary new PHPUnit adapter (right now it's the LegacyTestFrameworkBridge)

_If_ we currently didn't have any violations (we do), we could already extract the PHPUnit adapter, relying on `LegacyTestFrameworkBridge` (which stays in `infection/infection`), and despite the latter using `Configuration` and more, it would still work.

The limiting factor for extracting PHPUnit are the current violations, not the violations of `LegacyTestFrameworkBridge`.

It is not the dependencies of `LegacyTestFrameworkBridge` that shows us the problem, but the usage of it in the first place.

src/TestFramework/Contracts/TestFramework.php
maks-rafalko: > So essentially I think the test framework will return a subset of that information and within `infection/infection` we re-combine it.

hopefully :)

src/TestFramework/Contracts/TestFramework.php
maks-rafalko: thanks for the detailed response

> https://github.com/infection/infection/issues/3082. If we talk about priority, I think the latter is much more important than multi-test framework support, for two reasons

now I'm a bit lost :) 

does that mean we should prioritize it above this RFC/PR? Just checked, it's not even on the board for review, so I personally didn't read it (carefully) before. Now I did

----

One idea, just to have another thought for you to consider: you often mention Psalm "in-memory" implementation in this RFC and now PhpSpec9 which is "completely different" from coverage point of view related to the tracer RFC. 

While it's true, I would like to highlight a very simple, but still very important idea: we do not need to support everything from a PHP community. Even if we supported/support it right now (we do support PhpSpec, but not sure if we need to in the future tbh. We do not support Psalm, and IMO we (maintainers of Infection) shouldn't - more on it below).

With all the respect to the libs and their maintainers: It may sound weird, but let's be honest

- **Psalm** is ~nearly dead, especially comparing to PHPStan, which is now more popular at least in my bubble, but I'm pretty sure the same is on average in the whole community
- **Psalm** is dead, comparing to Mago (performance-wise). And I'm not even sure what will happen next with the raise of AI, where different tools written in PHP are being rewritten to Rust/Go alternatives. It will be interesting to re-read this comment in a couple of years :). Mago is a great example of potentially replacing those 2 SA tools in a long distance, especially after adding support to [custom extensions](https://mago.carthage.software/1.47.0/en/extensions/overview/) (written in any language). I expect many of the PHPStan rules will be ported to Mago rules, still written in PHP(!), which will bring you for example 10x-100x performance boost. So, spending now time to think how to support Psalm in Infection is probably not a good idea.
- **PhpSpec**... I saw literally 0 projects in my professional career during the last 17 years that used PhpSpec for their code bases. Again, this is for sure my bubble, but there were tens of legacy and new projects. None of them used PhpSpec. It's so rare, that thinking about how to support it is probably waste of our, Infection maintainers, time. I may be wrong, don't want to offend anyone, especially maintainers who spend their time, but at the same time I'm pragmatic here - my idea is to work on the things that are demanded by Infection _users_. If it's not demanded - I don't care about it. I use PHPUnit, PHPStan, Behat, Mago personally - this is my (personal) focus. If you use Psalm and PhpSpec - no problem, you can spend the time to support it. But I believe we shouldn't care about those tools *in advance* unless there is a demand in it.

theofidry: >One idea, just to have another thought for you to consider: you often mention Psalm "in-memory" implementation in this RFC and now PhpSpec9 which is "completely different" from coverage point of view related to the tracer RFC.

Whilst that's true, there are two things to take into consideration.

The first one is that that, at least from my observation, the current limitation for "in-memory" test frameworks is pretty free in the context of changing the `TestFramework` API. Alone, I would not pursue it. But looking at the Static Analysis API, there is little to do to allow it.

I also think that even for fast tools like Mago will not write off that approach. See the "Server-based approach".

Regarding PhpSpec9, true. Maybe I'm too hasty about this.

I guess overall I don't really mind either way; either we have:

1. patch works to bring those new features, which accumulate a lot of tech debt and we clean it up after
2. do a cleaner way which tackles a fair share of tech debt and is groundwork for those new features

The only difference is the order in which features are brought in, and my only two personal concerns are:

- limiting the overall tech debt (which I feel for the Test Framework is _really_ high, as shown by the struggle of #2816 and the list of written and unwritten outstanding issues at the Test Framework)
- limiting the number of updates to do on `infection/abstract-testframework-adapter` because it's very tedious work.

maks-rafalko: well, I'm not trying to discard this PR, just wanted to make sure we thought about all the possible alternatives. I got the answers, thank you. Let's first try to fix TF API then

src/TestFramework/LegacyTestFrameworkBridge.php
maks-rafalko: > The limiting factor for extracting PHPUnit are the current violations, not the violations of `LegacyTestFrameworkBridge`.

I'm afraid I don't understand that, sorry. 

PHPAt violations correctly highlight that `TestFramework` adapter currently depends on `Mutant`, which can't be extracted as is - this is yet to be fixed.

But if we had PHPAt violations set up for `LegacyTestFrameworkBridge` (or whatever _concrete_ implementation, be it final PHPUnit implementation or any other) - it would highlight us another violations, and the example we have right now is:

```
private Configuration $config,
private MutantProcessContainerFactory $processFactory,
private TestFrameworkExtraOptionsFilter $testFrameworkExtraOptionsFilter,
```

For example `Configuration`, `MutantProcessContainerFactory` are currently non-extractable. Why do you think this shouldn't be reported as violations?

theofidry: >For example Configuration, MutantProcessContainerFactory are currently non-extractable. Why do you think this shouldn't be reported as violations?

Let's imagine with go with the grace period for the test frameworks, then we will release _with_ `LegacyTestFrameworkBridge`, so the violations 

Because `LegacyTestFrameworkBridge` lives and will live in `infection/infection`. So those violations would be incorrect.

_If_ there was no other violation for PHPUnit, then in theory, despite the dependencies of `LegacyTestFrameworkBridge`, the extraction would still work (it would be silly as it would be extracting it against the old API – but still that would work).

Now, if instead of using `LegacyTestFrameworkBridge` we were to "in-line" it to make the PHPUnitAdapter implement the new API and no longer depend on the old one, _then_ it would show up those violations. And that would be one of the next steps in order to gradually migrate the PHPUnit adapter.

So you're gut feeling that those needs to be addressed is correct, and we will address them for PHPUnit which will show how to address them for PhpSpec and Codeception too.

If it's too confusing then I think I need to sketch something, or I'm missing an obvious explanation in my tiredness.

maks-rafalko: discussed in PM: 

- PHPUnitAdapter already covered by PHPAt (`testPHPUnitAdapterStaysWithinFuturePackageBoundary`)
- legace bridge will be removed - no need to cover it.

Resolving

src/TestFramework/Contracts/TestFramework.php
maks-rafalko: > Another part of the challenge is dealing with `symfony/process`.

Not sure I like this, given `abstract-testframework-adapter` is an abstract package. But let's think about it a bit later. I anticipate the API of `TestFramework::test()` method can (or will have to) be significantly changed during our experiments

src/TestFramework/Contracts/InitialRunResults.php
theofidry: Edit: as discussed in another comment, the memory limit will most definitely be test framework specific and become an internal implementation detail of each adapter. So maybe `InitialRunResults` will actually not be necessary in the end. But it can be removed once we know for sure, right now it is being used still.

src/TestFramework/Contracts/TestFramework.php
maks-rafalko: > But the goal of this PR is to introduce a first version of the API and backport the current adapter to it with minimal changes.

but since `::getVersion()` doesn't work properly (yet), we can't say that this PR contains "a first version of the API", because API is not ready - that's my point. 

I mean it's fine to have the first version of API and then change it, but what we have in this PR is a *part* of the first version (=non-complete).

Or what do you mean by "the first version"?

I very briefly checked #3290 and see `TestFramework::getVersion()` there, which makes sense to me (at this moment) and answers my question. 

Just want to highlight that the current `TestFramework` API from this PR isn't complete.

Concrete questions (so the plan is better understood):

- what would be the next PR(s) given this one is merged? 
- are we going to extend this `TestFramework` interface to cover for example `::getVersion()`
- are we going to replace all instances of `TestFrameworkAdapter` with `TestFramework` so we know it works in all the places?

Example:

https://github.com/infection/infection/blob/440966b27c38b37fa5f65ed4d3100ba0534deabc/src/Logger/ArtefactCollection/ConsoleProgressBarLogger.php#L56

theofidry: > I mean it's fine to have the first version of API and then change it, but what we have in this PR is a part of the first version (=non-complete). [...] Just want to highlight that the current TestFramework API from this PR isn't complete.

It is definitely not the final version. I don't expect _major_ changes, and so far, and as recently as via the PoC of #3290 in which I went a lot further, there has not been any unexpected change.

I do expect some adjustments, a new parameter being necessary, a new exception being documented, or for example #3082 will likely introduce a `::getTracer()`. But none of those are problematic.

In the end in #3082 I kept `::getVersion()` so I added it back here. I had hopes to get rid of it but it's neither critical nor important.

> what would be the next PR(s) given this one is merged?

[I recorded the steps (in order) that I did in the PoC.](https://github.com/infection/infection/pull/3290/changes#diff-5136ef13af2a186aad8fe7aba15de9028d3931d5e8f836df90d187697fc570ab). I want to review them later and determine what is the best way to go about it.

I think there is two (conflicting) directions:

1. Complete the "migration" ASAP: this is essentially trying to arrive to the state I arrived at in #3290 (obviously incrementally, and being more careful with some changes and have proper reviews at each step).
2. Complete the _blocking_ pieces ASAP: the goal there would be to identify which intermediate milestones enables new features which would allow us to work on those features, without having to complete the entirety of the migration.

I say they are conflicting because they compete with each other in terms of timeline. I am not certain which way is better. Currently, from my PoC and my mental state, I'm more leaning towards 2). 

In the end, I expect that the limiting factor will rather be the review more than for which direction I submit the changes, so in this case changing course will be a matter of changing the priority of the PRs, which is easy to do at any time.

> are we going to extend this TestFramework interface to cover for example ::getVersion()

As mentioned above, I added it back (for now at least).

> are we going to replace all instances of TestFrameworkAdapter with TestFramework so we know it works in all the places?

That is the intent indeed. Roughly my idea is:

- Introduce the interface for `TestFrameworkFactory`.
- Turn the static analysis in a `TestFramework`. Unlike `TestFrameworkAdapter`, this is `infection/infection` code so we can complete the migration without having a bridge hanging around.
- Introduce a `CombinedTestFramework` implementation – the first implementation of the multi-TestFramework support. This is most likely going to be done in 3 steps: first a skeleton implementation, then a phase where we fix multi-TestFramework related issues (on top of my head there is https://github.com/infection/infection/issues/2327 which is no longer a nice to have but a blocker and something about `MutantProcessContainer` which I don't recall right this instant). And then we can adjust the code to allow to register more test frameworks via the configuration.

Migrating the remaining `TestFrameworkAdapter` with `TestFramework` will be done in parallel.

maks-rafalko (approved):
thank you @theofidry for explaining in details

Issue #3315, opened 2026-07-02
bug: Unable to locate a phpunit executable on local system


verfriemelt-dot-org:
| Question    | Answer
| ------------| ---------------
| Infection version | 0.33.3 and newer
| Test Framework version | PHPUnit 13.1
| CI | yes
| PHP version | 8.5.7
| Platform    | Ubuntu
| Github Repo | (private)

starting with infection 0.33.3 or 0.34.0 i get a consistent failure in the pipeline (and only there unfortunately) of infection being unable to locate phpunit even though its installed (the pipeline run with phpunit tests in isolation work :) ) 0.33.2 and before work fine.

```
Infection - PHP Mutation Testing Framework version 0.34.0
[notice] You are running Infection with Xdebug enabled.
In FinderException.php line 62:
                                                                               
  Unable to locate a phpunit executable on local system. Ensure that phpunit   
  is installed and available.                                                  
```

<details>
 <summary>infection.json5</summary>

 ```json5
{
    "$schema": "vendor/infection/infection/resources/schema.json",
    "source": {
        "directories": [
            "src",
            "lib"
        ],
        excludes: [
            ...
        ]
    },

    "mutators": {
        "global-ignore": [
            ...
        ],
        "@default": true,
        PublicVisibility: false,
        ProtectedVisibility: false,
        ConcatOperandRemoval: false,
        ReturnRemoval: false,
        MatchArmRemoval: false,
        CastString: false,
        Concat: false,
        ArrayOneItem: {
            ignore: [
                "*::toArray"
            ]
        },
        Coalesce: {
            ignore: [
                ...
            ]
        },
    },
    "testFramework": "phpunit",
    "testFrameworkOptions": "--testsuite=Unit --exclude-group=no-infection",
    "staticAnalysisTool":"phpstan",
}

 ```
</details>

<details>
 <summary>phpunit.xml</summary>

 ```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
         beStrictAboutCoverageMetadata="true"
         beStrictAboutOutputDuringTests="true"
         bootstrap="tests/bootstrap.php"

         colors="true"
         defaultTestSuite="Unit"

         displayDetailsOnTestsThatTriggerWarnings="true"
         displayDetailsOnTestsThatTriggerErrors="true"
[...]

PR #3318, opened 2026-07-03
fix(test-framework): Correctly leverage the vendor bin path fallback when Composer could not provide it

theofidry:
## Description

As of now, if `composer config bin-dir` returns an empty output, the fallback that we have (`$PWD/vendor/bin`) did not kick in. I think it is a mistake.

From #3315, it was identified that if the composer executable was incorrect (`php /app/composer`), but `/app/composer` was a directory, not the Composer executable.

A surprising behaviour is that PHP does not bail out:

```
# src is an existing directory, not an executable!
php src config bin-dir
# exists 0, no output
```

The conclusion is that **executing `php <composer> config bin-dir` fails if `<composer>` points to a directory**. That in itself is a bug too that should be addressed, but will be done separately.

## Changes

TODO: Extract the introduction of utility methods to make the diff less noisy.

## Related issues

- https://github.com/infection/infection/issues/3315

### DIFF
--- a/src/FileSystem/Finder/TestFrameworkFinder.php
+++ b/src/FileSystem/Finder/TestFrameworkFinder.php
@@ -92,12 +92,12 @@ class TestFrameworkFinder
         return $this->cachedPath[$testFrameworkName];
     }
 
-    public function getVendorDirProcess(): Process
+    private function getBinDirProcess(): Process
     {
         return new Process([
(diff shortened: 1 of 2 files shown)


theofidry:
Closed in favour of https://github.com/infection/infection/pull/3511.

PR #3320, opened 2026-07-03
test: Exclude create-main.php from PHPStan

theofidry:
Currently PHPStan complains about `tests/benchmark/MutationGenerator/create-main.php` relying on `sources/autoload.php` but the latter does not exist.

Indeed, the mutation generator benchmark fixtures may not be materialized hence the issue is legitimate. However, silencing it would also cause an issue when those fixtures are materialized.

Consequently, this PR proposes to exclude this file from the inspection. It is preferable than materializing the fixtures all the time IMO.


sanmai:
👍🏻

PR #3328, opened 2026-07-05
docs: Introduce an ADR for the cache naming convention

theofidry:
This PR puts into an ADR the convention we follow about naming cache (memoization) related classes. This convention is driven by the existing code rather than an active decision taken recently.


adr/0006-memoized-over-cached-nomenclature.md
```diff
@@ -0,0 +1,53 @@
...
+classes, such as storage backends, cache keys, invalidation, eviction, warmup,
+cleanup, configuration, persistence across process runs or sharing across object
+instances. Infection already uses cache terminology for these broader concerns:
+PHPUnit result cache files, PHPStan cache, Rector cache and files under
+`var/cache`, for example.
+
+Using `Cached*` for object-local memoization makes these classes appear closer
+to operational cache infrastructure than they are. `Memoized*` is more precise
+and makes it clear that the behaviour is a local implementation detail, not a
+cache lifecycle boundary.
```
maks-rafalko: I don't get this paragraph. Any examples / details?

adr/0006-memoized-over-cached-nomenclature.md
```diff
@@ -0,0 +1,53 @@
...
+
+
+## Decision
+
+Use `Memoized*` rather than `Cached*` when naming classes and decorators whose
+purpose is to reuse a previously computed value during the same object lifetime.
```
maks-rafalko: hm.. 

> whose
purpose is to reuse a previously computed value during the same object lifetime

we can say it about any caching. In which way this is more `Memoize*` than `Cached*`?

maks-rafalko (commented):
I find it hard to understand to be honest, probably it would be nice to have more examples about what the cached and memoized things are.

adr/0006-memoized-over-cached-nomenclature.md
theofidry: It is actually explained from the previous paragraph when defining `memoization`:

> the stored value is local to the object lifetime, with no explicit invalidation, eviction, TTL, persistence or shared cache backend.

adr/0006-memoized-over-cached-nomenclature.md
theofidry: When the cache lifecycle is exposed to the callers.

So if you have any shape or form of cache validation, eviction, lifetime, etc., IMO `Cached*` would be more appropriate.

In practice, in our codebase, this would translate to use a [`PSR-6`](https://www.php-fig.org/psr/psr-6/) or [`PSR-16`](https://www.php-fig.org/psr/psr-16/) abstraction.

maks-rafalko: > In practice, in our codebase, this would translate to use a [`PSR-6`](https://www.php-fig.org/psr/psr-6/) or [`PSR-16`](https://www.php-fig.org/psr/psr-16/) abstraction.

isn't it a counter-example of this RFC? These PRSr use `chache`, not `memoize` wording. Or I completely miss the point

theofidry:
@maks-rafalko updated and added some examples

maks-rafalko (approved):
Thank you

adr/0006-memoized-over-cached-nomenclature.md
theofidry: I hope the added examples clarify that point

PR #3331, opened 2026-07-05
chore(autoreview): Introduce PHPStan deprecration rules

theofidry:
While working on PHPat I noticed some deprecated code being used, but it did not trigger an PHPStan warning.

It turns out it is not a native feature, it requires [phpstan/phpstan-deprecation-rules](https://github.com/phpstan/phpstan-deprecation-rules). So this PR installs it and updates the baseline; no deprecation is addressed in this PR.


maks-rafalko (approved):
the build should be fixed as it will fail master

PR #3333, opened 2026-07-05
test: Fix PHPUnit configuration for coverage

theofidry:
We have various test classes (utilities, i.e. code for the test cases themselves, fixtures or implementations used in tests only). Since those are not part of the configured PHPUnit code source target, using `Covers` attributes against them result in issues as of PHPUnit 12.

I did not notice the issue on #3274 because I did not look properly at the CI and they show only if coverage is enabled.

This PR addresses this by registering `tests/phpunit` as part of the source code, i.e. classes within it may require coverage. In practice it aligns with our usage as we tend to cover those test classes by code too.

For `DefinitionTest` and `MutationCategoryTest`, those are now third-party code (located in `vendor/`) hence `CoversClass()` is no longer valid too.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -34,6 +34,7 @@
     <source ignoreIndirectDeprecations="true">
         <include>
             <directory>src</directory>
+            <directory>tests/phpunit</directory>
         </include>
     </source>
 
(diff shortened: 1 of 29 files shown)


maks-rafalko:
> we tend to cover those test classes by code too.

could you please explain what does this mean "cover tests by code"?

theofidry:
@maks-rafalko I do not remember. I think you're looking at it via the email, I since then edited the description to be clearer.

maks-rafalko:
no, it's in the current description

> This PR addresses this by registering tests/phpunit as part of the source code, i.e. classes within it may require coverage. **In practice it aligns with our usage as we tend to cover those test classes by code too.**

theofidry:
That if you have utility classes in `tests/phpunit`, e.g. `NonRewindableIterator`, we have the corresponding `NonRewindableIteratorTest` too.

PR #3334, opened 2026-07-05
ai: Add a Docker Sandbox skill

theofidry:
Add a skill that can be used actively (i.e. invoking it directly) or passively (i.e. the AI figures it should leverage that skill when performing a task) for Docker Sandbox.

There is nothing too deep in there, it is mostly to stop the AI from being confused when asked about Docker Sandbox as it tends to confuse it with the regular Docker product, and to be pointed to our go-to documentation in `devTools/sbx/README.md`.

Currently I copy/pasted it for both Claude and Codex. It would be nicer to figure out a way to have both with one source of truth, but I do not have a solution for that at the moment.


maks-rafalko:
> but I do not have a solution for that at the moment.

adding a symlink to a git? before we find a general way of reusing skills across different agents. I anticipate many skills to be added in the near future, duplicating them is not reliable.

theofidry:
>adding a symlink to a git?

But both skills slightly differ from Claude to Codex... Codex has the agent for example.

maks-rafalko:
> But both skills slightly differ from Claude to Codex... Codex has the agent for example.

but the files are identical. Probably, we can add a file reference and point to a shared `@some/file` from the `SKILL.md`.

I'm not sure as well. Fine with postponing it a bit.

maks-rafalko (approved):

theofidry:
>but the files are identical.

Yeah, fair enough. I suppose it's at least that.

theofidry:
Urg Codex doesn't follow symlinks... So I'll add it to Codex and symlink it for Claude

Issue #3337, opened 2026-07-05
bug: Static analysis kill-stage processes crash under the test-derived memory cap and are silently counted as detected


sebastianbergmann:
| Question              | Answer                                                    |
|-----------------------|-----------------------------------------------------------|
| Infection version     | 0.34.0                                                    |
| Test framework        | PHPUnit 13.2.2                                            |
| Static analysis tool  | PHPStan 2.2.4 (`staticAnalysisTool: phpstan`)             |
| PHP version           | 8.5.8                                                     |
| Platform              | GitHub Actions, `ubuntu-latest` (reproduces anywhere the conditions below hold) |
| Coverage driver       | Xdebug (via `shivammathur/setup-php`, `coverage: xdebug`) |

## Summary

When the PHPStan integration is enabled, the per-mutant PHPStan processes inherit the memory limit that `MemoryLimiter` derives from the *initial PHPUnit run* (twice its reported memory usage, 95M in my case). Loading PHPStan's result cache alone can exceed that limit, so on our CI every single mutant that reached the static analysis stage crashed with a PHP fatal error instead of being analysed.

The consequence is worse than a crash: `PHPStanMutantExecutionResultFactory` classifies any exit code above 100 as `DetectionStatus::ERROR`, and errors count as detected. Mutants that genuinely escape both the tests and PHPStan are therefore reported as detected, the MSI reads 100%, and the build stays green. We ran for some time with 19 genuinely escaped mutants (verified by applying the diffs by hand and running the full suite) that CI reported as "42 errors were encountered" and a passing 100% MSI gate.

## What happens

Every mutant that survives its covering tests produces the same log entry (from `--logger-text` with `--log-verbosity=all --debug`):

```
$ '.../tools/.phpstan/vendor/bin/phpstan' '--tmp-file=/tmp/infection/mutant.<id>.infection.php'
  '--instead-of=.../src/Compiler/CodeGenerator.php'
  '--configuration=/tmp/infection/phpstan.<id>.infection.neon'
  '--error-format=json' '--no-progress' '-vv'

  PHP Fatal error:  Allowed memory size of 99614720 bytes exhausted (tried to allocate 32768 bytes)
  in /tmp/phpstan/resultCache.php on line 52329
  Stack trace:
  #0 phar://.../phpstan.phar/src/Command/AnalyseApplication.php(79): ResultCacheManager->restore()
  ...
  PHPStan process crashed because it reached configured PHP memory limit: 95M
[...]

PR #3341, opened 2026-07-06
test: Capture PHPStan failing due to the PHPUnit memory limit

theofidry:
This PR adapts the PHPStan integration test to demonstrate the bug reported from #3337.


theofidry:
I'm hesitating with moving it to a dedicated e2e test instead of polluting the existing one. Also note that there is more comments left in the `README.md` than I would like, but it would be cleaned up by the PR introducing the fix.

tests/e2e/PHPStan_Memory_Limit/tests/bootstrap.php
```diff
@@ -0,0 +1,3 @@
+<?php
+
+require_once __DIR__ . '/../vendor/autoload.php';
```
maks-rafalko: why is it needed?

maks-rafalko (changes requested):
> I'm hesitating with moving it to a dedicated e2e test instead of polluting the existing one.

The existing test shouldn't be updated, I don't see any reason to do it. We should create a new e2e test for this bug.

Also, not sure why we are asserting **bad** behavior and make it green. 

IMO, as we did it many times before, this test should fail and we can add a fix here so that it becomes green.

theofidry:
> Also, not sure why we are asserting bad behavior and make it green.

It works both ways.

The advantage of this one IMO is that you can merge it separately and it's easy to come back to and verify if it worked or not. And when you fix it, you see only the relevant diff, not the noise about the test structure itself.

With the other, besides the noise, you need to go back to a commit that may just be squashed and no longer exist in the future.

In this peculiar case not only the test is quite noisy, it was a bit hard to get right too. So verifying that it worked because my fixed work rather than a mistake in the test was not easy.

maks-rafalko (approved):

PR #3348, opened 2026-07-07
test: Update the PHPStan baseline

theofidry:
Something got merged since I rebased https://github.com/infection/infection/pull/3331 and did not notice the branch was not up to date when hitting auto-merge.

As a result, the baseline is not up to date and fails on master. This PR re-generates it.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,588 @@
+# AGENTS.md
+
+Operating guide for AI agents contributing to Infection. It is read once per session, so it is
+intentionally dense. Each line must help an agent avoid a known mistake. Claims are anchored
+to files - trust the tree over your training data. This codebase was heavily re-architected
+through 2025-2026, so old assumptions can be wrong. When your change makes a line here false
+- a path moves, a version gate shifts, a convention evolves - update that line in the same
+task. If you find a stale line while reading, fix it too.
+
+## What you are working on
+
+Infection is the mutation testing framework for PHP - 29 million Packagist installs, running
+inside the CI of thousands of projects. It parses source into ASTs (nikic/php-parser - API
+refresher: `vendor/nikic/php-parser/README.md` and its `doc/` folder), applies
+small mutations, and checks whether the project's tests - and optionally a static analyser -
+notice. A bug here silently corrupts other projects' quality gates. A slowdown here multiplies
+by thousands of mutant processes per user run. Reviewers weigh correctness, memory, and
+per-process overhead equally.
+
+The project tests itself with Infection: CI runs Infection against this repository with a
+minimum-MSI gate (`.github/workflows/mt.yaml`) and annotates escaped mutants inline on PRs.
+That fact explains most of the coding rules below: code is written so that every mutation of
+it dies.
+
+## Vocabulary
+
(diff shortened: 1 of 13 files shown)


theofidry:
@ondrejmirtes @staabm I have the impression there is an issue on PHPStan side here.

My understanding is:

- The CI attempts to restore the cache; it is a new PR so it fallbacks on an older (and at least partially outdated) cache.
- PHPStan executes and re-analyse only a subset of the changes (not obvious from the diff, but 0.537 seconds is quite short).

In https://github.com/infection/infection/pull/3331, I added `phpstan/phpstan-deprecation-rules`. 

But locally, stale cache or new cache, updating the baseline results in the one currently available in the PR. The CI however, finds differences.

Any idea?

staabm:
this sounds like the behaviour introduced in https://github.com/phpstan/phpstan-src/pull/5933

@sandermuller can you clarify what we see is expected behaviour

theofidry:
For reference, I had a similar issue previously with PHPat, see https://github.com/infection/infection/pull/3261. Maybe the solution is for `phpstan/phpstan-deprecation-rules` to declare its own cache invalidator.

I think the general idea of https://github.com/phpstan/phpstan-src/pull/5933 is sensible, but depending of the plugins/extensions enabled, the rationale no longer holds true, but it is the responsibility of those plugins/extensions to adapt the cache invalidation.

SanderMuller:
I poked at the phpstan-src side. One thing that might narrow it down: PHPStan hashes the registered extension/rule files into the result cache separately from the composer.lock logic, and that check still runs regardless of #5933. So a phpstan-deprecation-rules bump that changes its rule code should still invalidate, and vendor `@deprecated` changes are covered too, since the file calling the deprecated API records a dependency on that package.

The gap #5933 can actually open looks narrower: a package whose changed code is neither reflected by any analysed file nor registered as an extension service. That's basically the PHPat selector situation (your own classes, not registered services), and fwiw #3261 landed a few weeks before #5933 merged, so that one wasn't caused by it. So a dedicated ResultCacheMetaExtension is nice hardening, but I'm not sure deprecation-rules actually needs one. If you can get a fresh vs warm cache to disagree with only deprecation-rules enabled, that'd pin it down. Happy to dig further either way.

SanderMuller:
Circling back on the cache-staleness angle: I found a concrete gap. PHPStan wasn't tracking global/namespaced constants as cache dependencies (only class constants, functions, methods and properties), so a file that reads a global constant records no dependency on where it's declared. When that constant changes the reader isn't re-analysed, and the result cache serves a stale result. With package-granular invalidation (phpstan/phpstan-src#5933) the same happens when the declaring package is bumped.

Opened a fix: phpstan/phpstan-src#6023.

It's separate from the rebase issue in this PR, but if the baseline drift you've been seeing is cache staleness, that branch is worth a try. It also means a per-plugin ResultCacheMetaExtension isn't the general answer for this; the missing dependency edge was the root cause for constants.

theofidry:
The stale baseline has accidentally been merged to `master` so hard to this PR cannot reproduce it anymore atm. We'll update PHPStan ASAP as soon as the patch is merged and I'll keep an eye on the issue.

PR #3327, opened 2026-07-05
Add AGENTS.md

sanmai:
## Description

I bootstrapped Fable 5 with a very detailed and lived-in file I pulled from a proprietary project, to use it as a source of inspiration, and later I steered the file towards more simple language (so "load-bearing" has minimal presence here) using [ASD-STE100](https://asd-web-be-prod.azurewebsites.net/media/wunhmi5y/asd-ste100-issue-9.pdf) [Simplified Technical English](https://en.wikipedia.org/wiki/Simplified_Technical_English) (airspace engineering English basically) as a guiding star, not a goal - the goal is always to enable the agents. 

The final file looks long (~12000 tokens) but not only it is not that expensive (that's $0.06 of Opus 4.8 and even less for other models) but also it is on a par with the system prompt used by Claude Code itself (usually tens of thousands of tokens when several MCP servers are configured). Alas, this is a one-per-session cost with cache reads being 90% cheaper.

I verified the result with multiple proprietary and open-source models:

- Claude Opus 4.6
- ChatGPT 5.5
- [Qwen3.6-27B](https://huggingface.co/Qwen/Qwen3.6-27B)
- [Gemma 4 31B Dense](https://huggingface.co/google/gemma-4-31B-it)

They all basically said:

> AGENTS.md is accurate and correctly describes the current state of the architecture, conventions, and invariants of the Infection codebase.

Now, the real test of this file will be later, when working with agents on improving the project.

## Costs to make it

```
Total cost:            $104.87
Total duration (API):  2h 54m 32s
Total duration (wall): 1d 21h 29m
Total code changes:    2422 lines added, 355 lines removed
Usage by model:              
    claude-haiku-4-5:  1.2k input, 32 output, 0 cache read, 0 cache write ($0.0014)
     claude-opus-4-8:  105.5k input, 479.7k output, 9.9m cache read, 1.4m cache write ($25.90)
      claude-fable-5:  145.6k input, 235.1k output, 25.9m cache read, 2.1m cache write ($78.96)
```

FWIW I typed this description myself with the exception of the quotes.

## Checklist

- [x] Appropriate labels applied (e.g. `performance`, `feature`).

PR #3349, opened 2026-07-07
docs: Add an implementation map for the execution phases

theofidry:
Add an implementation map to give a quick view of where to locate the different pieces corresponding to the domain phases to our implementation in Infection.

Prompted by https://github.com/infection/infection/pull/3327#discussion_r3535607265.

### DIFF
--- a/doc/nomenclature.md
+++ b/doc/nomenclature.md
@@ -3,6 +3,7 @@
 ## Table of Contents
 
 - [Execution Phases](#execution-phases)
+    - [Implementation Map](#implementation-map)
 - Index
     - [A](#a)
         - [Arid Node][arid-node]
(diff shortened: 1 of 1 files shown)


sanmai (approved):
Matches my memories.

I guess the problem I wanted to avoid by keeping all agent-relevant docs in  AGENTS.md is to avoid the situation where things go out of date. For AGENTS.md this situation is straightforward (it is all agent authored, and it can keep being such - no humans required) but for any documentation outside of AGENTS.md I would really want to review thoroughly or get it authored/proofed by a human.

maks-rafalko (approved):

theofidry:
>but for any documentation outside of AGENTS.md I would really want to review thoroughly or get it authored/proofed by a human.

Totally agree.

At the same time if there is pieces of docs that can be valuable for maintainers/contributors, I think it's good too.

PR #3358, opened 2026-07-09
fix(test-framework): Do not limit the PHPStan process by the memory usage detected for PHPUnit

### LINKED ISSUE #3337: bug: Static analysis kill-stage processes crash under the test-derived memory cap and are silently counted as detected

theofidry:
## Description

This PR prevents Infection's generated PHP `memory_limit` cap from constraining PHPStan executions.

Infection may append a memory limit to the temporary `php.ini` after the initial test run. PHPStan then inherits that cap for initial and mutant static analysis runs, which can cause PHPStan to fail because of Infection's process-level memory guard rather than because of static analysis results.

This PR makes PHPStan command lines now pass `-d memory_limit=-1` so PHPStan keeps its own memory-limit behavior independent from Infection's generated cap.

**This is not a definitive solution.** A better solution to consider is to allow test frameworks/static analysis adapters to decide what memory cap to apply. This however require more groundwork and will be part of #2863 and #3018.

## Changes

- Adds shared PHP extra args for PHPStan adapter commands and PHPStan mutant process commands.
- Passes `-d memory_limit=-1` when building PHPStan analyze and version commands.
- Updates PHPStan unit expectations for the new PHP extra args.

## Reviewer Notes

The decision about why this solution rather than another can be found in https://github.com/infection/infection/issues/3337.

An alternative solution considered was to use the PHPStan `--memory-limit` option. The problem with the latter is that the memory is set once the sub-process is booted. As a result, if the sub process did not have enough memory to boot up in the first place, it will crash.

## Related issues

Fixes https://github.com/infection/infection/issues/3337.


tests/e2e/PHPStan_Memory_Limit/README.md
```diff
@@ -10,21 +10,12 @@ The test forces those preconditions in `run_tests.bash`:
 
 The PHPUnit test prints `Memory: 16.00 MB` during the initial test run. PHPUnit's own final memory line may vary between environments, but Infection reads the first matching memory line from the process output. This makes `MemoryLimiter` append `memory_limit = 32M` to the temporary php.ini after the initial static analysis run and before mutant execution.
 
-`phpstan-bootstrap.php` is loaded by PHPStan. It is a no-op for the initial PHPStan run. For mutant PHPStan runs, identified by `--tmp-file=...`, it asserts the regression setup:
+`phpstan-bootstrap.php` is loaded by PHPStan. It is a no-op for the initial PHPStan run. For mutant PHPStan runs, identified by `--tmp-file=...`, it asserts both sides of the fixed behavior:
 
 - the temporary `php.ini` contains the expected `memory_limit = 32M` cap, proving `MemoryLimiter` ran;
-- the effective PHP memory limit is still `-1`.
+- the effective PHP memory limit is still `-1`, proving the PHPStan process opted out of that cap.
```
sanmai, line 16: What if a user has their own cap? I wonder if there's a way to skip adding our memory limit instead of adding a non-bounded limit.

sanmai (approved):
Comment is not blocking.

tests/e2e/PHPStan_Memory_Limit/README.md
theofidry: I know PHPStan allows to apply their own cap, I do not know how it works behind the scenes. If it's a restart, I actually think it is bad for the mutant processes as it means for each PHPStan process we launch, PHPStan does a restart...

But that's a separate problem.

I am not clear on how caps should carry nor when they should apply tbh. I think we should create a dedicated RFC to establish this. Even if what we currently are doing is correct, IMO it deserves to be documented.

src/StaticAnalysis/PHPStan/Adapter/PHPStanAdapter.php
```diff
@@ -55,6 +55,8 @@
  */
 final class PHPStanAdapter implements StaticAnalysisToolAdapter
 {
+    private const array PHP_EXTRA_ARGS = ['-d memory_limit=-1'];
```
maks-rafalko, line 68: PHPStan has `--memory-limit` CLI option: https://phpstan.org/user-guide/command-line-usage#--memory-limit

Do you know is it better (or worse) than `-d memory_limit=-1` on PHP args level? If not, we can ask Ondrej/staab, but I will tag them only if you don't know the answer.

My thoughts: given this options exists, there should be a reason, and it's better for us to understand this reason.

maks-rafalko (changes requested):

src/StaticAnalysis/PHPStan/Adapter/PHPStanAdapter.php
theofidry: When PHPStan parses `--memory_limit`, it uses `ini_set('memory_limit', '-1')`, so in that sense there is no overhead of using that instead of `-d memory_limit=-1`. Indeed, it PHPStan will not do a process restart to apply it.

But there is a subtle difference, which is about whether the process memory limit we set is sufficient for PHPStan to boot itself up. For example if we detected 32M from PHPUnit, because it has a lower footprint, and PHPStan takes 50MB to boot itself up, then PHPStan would crash still maybe before `--memory_limit` is applied.

I am not knowledgeable enough about the PHPStan internals to assess this. So maybe @staabm may want to shim in.

staabm: I am not sure what to reply (or what reply you expect) :-).

only thing which comes to mind is, that I would use the`--memory-limit` flag, because it exists and it seems there is a reason for it (which I don't know). in case the php.ini's memory-limit would be good enough, the CLI parameter would not exist in the first place, I guess.

My bet is the `--memory-limit` might have a different effect on subprocesses.
it is passed thru to workers, while this does not happen with any `-d` parameter you pass to the main process:
https://github.com/phpstan/phpstan-src/blob/3f716d44cde31076dc33a2f835872d7e2d26f0e7/src/Process/ProcessHelper.php#L41-L44

theofidry: The code you show actually shows the opposite, `-d memory_limit` (or rather, the limit of the main process) is forwarded to the sub-processes when `--memory-limit` is passed.

So I tried to implement the suggestion, it did not work. The problem appears to be that (in the e2e test) PHPStan tries to set the memory limit once the sub-process is started. In the scenario where the memory limit is too little, it will crash before that.

So the original fix (the current state of the PR) is a more robust solution.

maks-rafalko: let's add this explanation to the code explaining why we are using `-d memory_limit` instead of `--memory-limit` so we understand it in the future

maks-rafalko (approved):

src/StaticAnalysis/PHPStan/Adapter/PHPStanAdapter.php
theofidry: done.

PR #1983, opened 2024-06-11
Add e2e tests to cover bug from #1980

maks-rafalko:
Follow up for https://github.com/infection/infection/pull/1982

Confirms, that Infection works as expected in a docker container without `git`

PR #3359, opened 2026-07-09
fix(e2e): Fix Without_Git by emulating an empty PATH

sanmai:
## Description

- Infection starts out `git` as a bare argv command without a full path so the executable is resolved through PATH.
- If we point PATH at an empty directory then we will create an environment without git in PATH.
- But we can't just unset or blank PATH because it will then reset to glibc's default of `/bin:/usr/bin` and `git` is there exactly.

Bonus point: `run_tests.bash` accepted and respects `$1` now (so it can be tested with .phar)

Previously:

- #1983
- #2002
- #2798
- #3356 (where we started this fix)

## Changes

- Fixes `Without_Git` to just work, all the time

## Checklist

- [x] Tests added/updated


tests/e2e/Without_Git/README.md
```diff
@@ -1,4 +1,7 @@
-Run Infection in a docker container **without** git. This ensures it works and doesn't fail because of git absense.
+Run Infection **without** git. This ensures it works and doesn't fail because of git absence.
 
-Covers https://github.com/infection/infection/pull/1981 and https://github.com/infection/infection/pull/1982
+- Infection starts out `git` as a bare argv command without a full path so the executable is resolved through PATH.
+- If we point PATH at an empty directory then we will create an environment without git in PATH.
+- But we can't just unset or blank PATH because it will then reset to glibc's default of `/bin:/usr/bin` and `git` is there exactly.
 
+covers https://github.com/infection/infection/pull/1981 and https://github.com/infection/infection/pull/1982
```
sanmai:
```suggestion
Covers https://github.com/infection/infection/pull/1981 and https://github.com/infection/infection/pull/1982
```

tests/e2e/Without_Git/run_tests.bash
```diff
@@ -2,46 +2,19 @@
...
-fi
-
-git_branch=$(echo "${GITHUB_HEAD_REF:-$(git rev-parse --abbrev-ref HEAD)}" | sed 's/\//\\\//g')
-
-echo "git_branch: ${git_branch}"
-
-if [ "$git_branch" == "master" ]; then
-  exit 0;
-fi;
+set -eo pipefail
```
sanmai, line 5: legit bug elsewhere to, I think coderabbit mentioned it or something

theofidry: About `set -eo pipefail`, indeed. We should address it separately everything IMO.

Likewise we still use PHPUnit9 in quite a few places... A good task for Claude/Codex too.

tests/e2e/Without_Git/composer.json
```diff
@@ -1,7 +1,6 @@
 {
     "require-dev": {
-        "phpunit/phpunit": "^9.6.20",
-        "infection/infection": "dev-master"
```
theofidry, line 4: is this the only case where we have this? (requiring `infection/infection` instead of using the "parent" one).

I reckon if we did that was to take into account dependency conflicts? E.g. here if we have an incompatible package between PHPUnit 9 and `infection/infection` we now may end up in a crash instead of a Composer dependency conflict.

tests/e2e/Without_Git/run_tests.bash
```diff
@@ -2,46 +2,19 @@
...
-  if [ -n "$fork_url" ]; then
-    composer config repositories.fork vcs "$fork_url" --no-interaction
-  fi
-fi
+PHP="$(command -v php)"
+readonly PHP
 
 composer install
 
-docker run -t -v "$PWD":/opt -w /opt php:8.4-alpine vendor/bin/infection --coverage=infection-coverage
```
theofidry, line 44: Does this really need to change? I think the failure in the CI was due to requiring `infection/infection`, not the `PATH`. Because I reckon there is no `git` in `php:8.4-alpine` so there is no need to alter the `PATH`there

tests/e2e/Without_Git/composer.json
sanmai: It may be the only case but I looked into the PR history and it sounds like it was accidental, not intentional- just to get Infection to run inside Docker. And if we'd keep it, what do we test exactly, that Composer can install from VCS?  Well, we know that, and even if suddenly stops doing that, it should not be the reason to break PR builds IMO.

If we want to test Infection against PHPUnit 9, 10, and so on, we already do that in other fixtures IIRC.

tests/e2e/Without_Git/run_tests.bash
sanmai: You may want to run it locally, and you may want to test it under different versions of PHP. With Docker we run essentially the same test across the whole test matrix. I understand your desire to keep the change minimal, but Docker here is really is a hack.

tests/e2e/Without_Git/run_tests.bash
sanmai: Yep, PR is in the works.

tests/e2e/Without_Git/run_tests.bash
sanmai: What the readme for the test said:

> This ensures it works and doesn't fail because of git absence.

And that's what it tests.

tests/e2e/Without_Git/run_tests.bash
theofidry: Done in https://github.com/infection/infection/pull/3361.

tests/e2e/Without_Git/composer.json
theofidry: It may not have been clear from my previous comment, I think removing `infection/infection` there is fine, the risk should be minimal and potential conflicting symbols would be about PHPUnit stuff, which is not the SUT here.

I am wondering however if there is other cases such as this one that would cause similar problems.

theofidry (approved):
Provided the PR is updated & there is no `required-dev infection/infection` cases (I did not check)

sanmai:
Few more tweaks: removed this temp dir dance for a fixed directory, and tested that git is indeed isn't there after our PATH shenanigans.

sanmai:
FWIW `phpstan-build-infection` installs Infection from VCS like we did here

maks-rafalko:
@sanmai out of interest, what was the reason of this PR, did you face some issue with this test?

sanmai:
Our PR builds were broken.

One of the recent failing builds: https://github.com/infection/infection/actions/runs/28952149922/job/85901451305

And here's an excerpt: 

```
 ./Without_Git
 git_branch: conductor-phpbench-phpbench-77006
 No composer.lock file present. Updating dependencies to latest instead of installing from lock file. See https://getcomposer.org/install for more information.
 Loading composer repositories with package information
 Updating dependencies
 Your requirements could not be resolved to an installable set of packages.
 
   Problem 1
     - Root composer.json requires infection/infection dev-conductor-phpbench-phpbench-77006, found infection/infection[dev-feature/fail-on-non-matching-ignore-regex, ..., dev-patch-1, 0.1.0, ..., 0.34.x-dev] but it does not match the constraint.
 
 ##[error]Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires infection/infection dev-conductor-phpbench-phpbench-77006, found infection/infection[dev-feature/fail-on-non-matching-ignore-regex, ..., dev-patch-1, 0.1.0, ..., 0.34.x-dev] but it does not match the constraint.

 make: *** [Makefile:280: test-e2e] Error 255
```

It looks like Packagist stopped registering conductor-authored branches, and I can't blame them.

PR #3368, opened 2026-07-16
doc: Update the covers annotation ADR

theofidry:
## Context

The ADR 0002 which is about `@covers` PHPUnit annotations is completely obsolete:

- PHPUnit no longer has covers annotations but attributes.
- The Symfony listener `Symfony\Bridge\PhpUnit\CoverageListener` was removed in https://github.com/infection/infection/pull/1926 as, at that time, it was not compatible with PHPUnit 10/11 which we wanted to upgrade to.

There also has been some clarification about enforcing the `integration` group when `#[CoversNothing]` is being used in adr/0007-declare-phpunit-coverage-metadata.md.

## Description

This PR deletes the obsolete ADR 0002 and introduces the ADR 0008 which supersedes it based on the **current conventions**. The goal here is to capture those, not challenge them.


adr/0007-declare-phpunit-coverage-metadata.md
```diff
@@ -0,0 +1,103 @@
...
+[PHP-CS-Fixer][php-cs-fixer] can also add the `#[CoversNothing]` attribute when
+no PHPUnit coverage attribute is present.
+
+Explicit coverage attributes also make it easier to determine programmatically
+which tests cover which production code, simplifying the creation and
+enforcement of [PHPat][phpat] rules.
+
+**The question is which convention to enforce.**
+
+Historically, ADR 0002 stated that:
```
sanmai:
```suggestion
Historically, ADR 0002 (deprecated and deleted) stated that:
```

sanmai (approved):
Makes sense 👍 

Without `CoverageListener` there's not much options.

theofidry:
Disclaimer: I did leverage Codex a lot for this document. I'm not 100% happy with it tbh, there is some gaps (some that I tried to fill). That being said, this is about _existing_ conventions, not challenging them, so I did not go an extra length to try to justify things.

adr/0002-@covers-annotations.md
maks-rafalko, line 1: IMO we should keep old ADRs and add something like "Superseded by: X ADR" and in X ADR to add "Supersedes Y ADR".

This way we see the history of why we made particular decisions at some point in time. AI will also see this information so it's not an issue.

maks-rafalko: well, it's even proposed in the neighbor PR: https://github.com/infection/infection/pull/3369/changes#diff-e0e08971bafe9a53bb63abbdd077590659cf7b64be904e2143c6db2f64291464R44

adr/0007-declare-phpunit-coverage-metadata.md
```diff
@@ -19,7 +19,7 @@ enforcement of [PHPat][phpat] rules.
 
 **The question is which convention to enforce.**
 
-Historically, ADR 0002 stated that:
+Historically, ADR 0002 (now deprecated and deleted) stated that:
```
maks-rafalko, line 22: it's not actually deleted now, it's superseded

theofidry: Fixed in https://github.com/infection/infection/pull/3378

PR #3371, opened 2026-07-16
refactor(autoreview): Add PHPat rule to enforce the ADR 0001

theofidry:
This PR adds a PHPat rule to enforce the convention defined by ADR 0001.

### DIFF
--- a/tests/Architecture/PHPat/Selector/HasInheritDoc.php
+++ b/tests/Architecture/PHPat/Selector/HasInheritDoc.php
@@ -74,7 +74,7 @@ final readonly class HasInheritDoc implements SelectorInterface
         ReflectionMethod $methodReflection,
         ClassReflection $classReflection,
     ): bool {
-        return $methodReflection->getDeclaringClass()->getName() === $classReflection->getName();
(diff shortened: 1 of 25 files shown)


tests/Architecture/PHPat/Selector/HasInheritDoc.php
```diff
@@ -0,0 +1,77 @@
...
+    public function matches(ClassReflection $classReflection): bool
+    {
+        $nativeReflection = $classReflection->getNativeReflection();
+
+        if (self::containsInheritDoc($nativeReflection->getDocComment())) {
+            return true;
+        }
+
+        foreach ($nativeReflection->getMethods() as $method) {
+            if ($method->getDeclaringClass()->getName() !== $classReflection->getName()) {
```
maks-rafalko: this isn't obvious piece of code IMO. What do you think to simplify it to 

```php
if (self::isInheritedMethod($method, $classReflection->getName()) {
    continue;
}
```

or something like that

maks-rafalko (approved):

PR #3369, opened 2026-07-16
doc: Document ADRs

theofidry:
This PR:

- Adds a template to follow when creating a new ADR.
- A brief `README.md` in `adr/` to explain what they are about and when to use them.
- Update `AGENTS.md` to make it aware of the ADR template.

A later additional would be to update `AGENTS.md` to suggest the AI to write an ADR when a task reveals or introduces a durable architectural decision or project-wide convention.

PR #3372, opened 2026-07-16
doc: Define the PHP version support policy

theofidry:
## Description

Rework ADR 0005 from a record of the PHP 7.4 and 8.0 support drop into the project's ongoing PHP version support policy.

The policy now documents the established annual review cadence and the considerations used when raising the minimum PHP version. It also records the decisions that led to PHP 8.3 (the current minimum supported version) becoming the current minimum.

## Changes

- Define an annual review of the oldest supported PHP version, with a baseline increase of
  one PHP minor version per year.
- Document how PHP support status, ecosystem adoption, language benefits, dependency and
  tooling support, maintenance cost, release timing, and user impact inform each decision.
- Describe the consequences for maintainers and users.
- Record the PHP 8.1, 8.2, and 8.3 minimum-version decisions and link their supporting issues
  and pull requests.


adr/0005-Bump-PHP-versions.md
```diff
@@ -1,32 +0,0 @@
-# Bumping PHP version requirements
```
maks-rafalko, line 1: this ADR shouldn't be removed, but marked as superseded

See

- https://github.com/infection/infection/pull/3369/changes#diff-e0e08971bafe9a53bb63abbdd077590659cf7b64be904e2143c6db2f64291464R44
- https://github.com/infection/infection/pull/3368#discussion_r3598816761

adr/0008-PHP-version-support-policy.md
```diff
@@ -0,0 +1,83 @@
...
+  constraints, and contributor effort required by the oldest version.
+- **Release timing and user impact.** A floor increase must be made deliberately in a release
+  that communicates the new requirement. Users unable to upgrade PHP can remain on the
+  latest compatible Infection release. This option reduces, but does not eliminate, the
+  impact and must not replace an assessment of actual adoption.
+
+These considerations inform a documented decision rather than a scoring system. A proposal
+to raise the minimum must identify the versions being removed, their PHP support status and
+adoption where data is available, the specific maintenance or language benefits, and the
+migration path for affected users.
```
maks-rafalko, line 49: I think we also should state here that when we drop a minor PHP version support, we MUST release it as a new minor `0.X.Y` version. 

Because last time when we released it as a patch, we had issue when Testo was integrated.

Example: if `0.34.x` supports PHP 8.3, it should support it up to the latest `0.34.x` tag.

Dropping PHP 8.3 support should lead to `0.35.x` tag.

sanmai: Another reason to do this: if a bug comes up and we have to release a patched version. But if we dropped a PHP version in a patch version, we could be leaving the users of that just-recently-supported version in the cold.

adr/0005-Bump-PHP-versions.md
```diff
@@ -26,7 +26,7 @@ version is fine.
 
 ## Status
 
-Accepted ([#1760]).
+Superseded by [ADR 0008](0008-PHP-version-support-policy.md). Was accepted in Accepted [#1760].
```
maks-rafalko, line 29: > Was accepted in Accepted [#1760].

seems like a mistake here

theofidry: when the human makes more mistakes than AI :D

theofidry: Fixed in https://github.com/infection/infection/pull/3377.

PR #1192, opened 2020-03-20
Add accepted ADR entries

theofidry:
Add ADR entries for some conventions or practices that have been discussed in the course of the last few months.

This PR is not an attempt or invitation to challenge any of those decisions: those are decisions _already_ made. If you wish to challenge or change one, this should be discussed and done in a separate issue/PR.

PR #3373, opened 2026-07-16
doc: Clarify exception assertions in tests

theofidry:
## Description

ADR 0004 explains why PHPUnit tests should avoid direct `try-catch` blocks. However, it did not
cover tests that need to inspect the thrown exception beyond its type, message, or code.

The PR https://github.com/infection/infection/pull/2708 introduced the `ExpectsThrowables` trait for these cases. It returns the thrown exception
for further assertions while preserving readability and ensuring that the test fails when the
expected statement does not throw.

This PR updates ADR 0004 to document the complete convention.

## Changes

- Clarifies when to use PHPUnit's `expectException*()` API.
- Documents `ExpectsThrowables::expectToThrow()` for inspecting the thrown exception or continuing
  with further assertions.
- Adds an example that isolates the statement expected to throw from the test setup.
- Clarifies why direct `try-catch` blocks are error-prone.
- Corrects the ADR acceptance reference to PR #1192.


maks-rafalko (approved):

adr/0008-event-and-subscriber-naming.md
maks-rafalko, line 1: shouldn't it be a separate PR? Looks like committed by a mistake to this branch

theofidry: yes

PR #3375, opened 2026-07-17
doc: Add an ADR about our usage of assertEquals in our PHPUnit tests

theofidry:


sanmai (approved):
Makes sense.

adr/0010-compare-objects-with-assert-equals.md
```diff
@@ -0,0 +1,99 @@
...
+
+```php
+$this->assertEquals($expectedObject, $actualObject);
+```
+
+Use `assertSame()` when object identity is part of the contract. Continue to compare scalar
+values with `assertSame()` or a more specific assertion. Use explicit property assertions or
+a dedicated assertion helper when loose comparison could conceal a meaningful difference
+in type, or when the domain requires comparison semantics other than complete object state
+equality.
```
maks-rafalko: should we state here in a decision that `webmozarts/strict-phpunit` is a must to use `$this->assertEquals()`?

maks-rafalko (approved):

adr/0010-compare-objects-with-assert-equals.md
theofidry: That's a good idea

PR #3118, opened 2026-05-10
feat(telemetry): Add telemetry spans for source collection, artefact collection and reporting

theofidry:
## Description

Adds lifecycle events for source collection, artefact collection and reporting. Those events are also wired into OpenTelemetry tracing. Those phases were described in https://github.com/infection/infection/pull/3103 and allows us to have more insights on the trace execution.

Before:

Screenshot 2026-05-10 at 09 33 32

After:

Screenshot 2026-05-10 at 09 33 21

## Changes

This introduces:

- `infection.source_collection`,
- `infection.artefact_collection`
- `infection.reporting spans`

The initial tests and static analysis spans are now grouped under `infection.artefact_collection`.

## Reviewer Notes

The screenshot from the description section are from my local Jaeger instance, but those changes will also be observable on our Sentry instance.

## Related issues

- #3090

PR #3376, opened 2026-07-17
doc: Propose an ADR to establish event dispatching decorator conventions

theofidry:
## Description

Proposes an architecture convention for dispatching lifecycle events around operations exposed through interfaces. The ADR separates operation-boundary events, which belong in same-interface decorators, from internal algorithm milestones, which remain in the component that owns the algorithm.

This keeps core implementations independent of event infrastructure while preserving event ordering, delegation semantics, return values, and exceptions.

This is a **proposal** based on conventions used in https://github.com/infection/infection/pull/3108. On `master`, at the time of writing, there is no such cases yet.

## Changes

- Adds ADR 0011 for event-dispatching decorators.
- Defines the `EventDispatching<DecoratedType>` naming and dependency conventions.
- Requires decorators to preserve the decorated interface and delegation contract.
- Defines when completion events are dispatched after success or from `try-finally`.
- Establishes composition-root wiring and canonical PHPUnit test expectations.
- Documents why internal algorithm milestones do not follow the decorator pattern.
- Records direct dispatcher injection and universal decorator dispatch as rejected alternatives.

### DIFF
--- a/adr/0011-dispatch-events-with-decorators.md
+++ b/adr/0011-dispatch-events-with-decorators.md
@@ -1,0 +1,134 @@
+# Dispatch operation events with decorators
+
+## Context
+
+Events often describe the lifecycle of an operation already represented by an interface,
+such as parsing a file or producing a report. Dispatching these events from the
+implementation mixes the operation with event infrastructure. It also requires every
+implementation to know about the dispatcher, although event dispatch is an application-level
+concern rather than part of the operation's contract.
+
+The source-collection subsystem already uses decorators that implement the decorated
+service's interface to provide cross-cutting behaviour. For example,
+`MemoizedSourceCollector` implements `SourceCollector` and delegates `::collect()` to another
+`SourceCollector`, while keeping the underlying collector unaware of memoisation.
+Event-dispatching decorators use the same structure to keep the decorated implementation
+unaware of the event system.
+
+Not every event represents the boundary of a delegated call. Some events describe
+milestones within an algorithm, such as processing each item in a generated stream. Only
+the component that owns that algorithm can dispatch such events at the correct point. For
+example, `MutationGenerator` dispatches `MutableFileWasProcessed` only after it has yielded
+all mutations for one source file and collected their identifiers.
+
+## Decision
+
+When events describe the boundary of an operation exposed through an interface, dispatch
+them from a decorator instead of the underlying implementation. The decorator must:
+
+- implement the same interface as the decorated service;
+- be named `EventDispatching<DecoratedType>`;
+- accept the decorated service and `EventDispatcher` as explicit constructor dependencies;
+- dispatch the start event before delegating the operation;
+- delegate the original arguments exactly once and preserve its return value or exception;
+- dispatch the completion event at the point defined by that event's semantics; and
+- avoid adding behaviour unrelated to event construction and dispatch.
+
+For example, an event-dispatching source collector delegates the complete operation while
+surrounding it with source-collection events:
+
+```php
+final readonly class EventDispatchingSourceCollector implements SourceCollector
+{
+    public function __construct(
+        private SourceCollector $decoratedSourceCollector,
+        private EventDispatcher $eventDispatcher,
+    ) {
+    }
+
+    public function collect(): array
+    {
+        $this->eventDispatcher->dispatch(new SourceCollectionWasStarted());
+
+        $sources = $this->decoratedSourceCollector->collect();
+
+        $this->eventDispatcher->dispatch(
+            new SourceCollectionWasFinished(count($sources)),
+        );
+
+        return $sources;
+    }
+}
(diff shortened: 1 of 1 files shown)


adr/0011-dispatch-events-with-decorators.md
```diff
@@ -0,0 +1,134 @@
...
+at operation boundaries. Doing so couples the core behaviour to application-level event
+infrastructure and repeats the same lifecycle mechanics across implementations.
+
+This ADR also rejects dispatching every event from a decorator. A decorator can observe only
+one interface boundary; events for internal milestones belong to the component that owns
+the corresponding algorithm.
+
+## Status
+
+Proposed.
```
theofidry, line 134: to update before merge

adr/0011-dispatch-events-with-decorators.md
```diff
@@ -0,0 +1,134 @@
...
+
+Events often describe the lifecycle of an operation already represented by an interface,
+such as parsing a file or producing a report. Dispatching these events from the
+implementation mixes the operation with event infrastructure. It also requires every
+implementation to know about the dispatcher, although event dispatch is an application-level
+concern rather than part of the operation's contract.
+
+The source-collection subsystem already uses decorators that implement the decorated
+service's interface to provide cross-cutting behaviour. For example,
+`MemoizedSourceCollector` implements `SourceCollector` and delegates `::collect()` to another
```
theofidry, line 13: to re-use the example of https://github.com/infection/infection/pull/3118 and https://github.com/infection/infection/pull/3197 instead once approved

adr/0011-dispatch-events-with-decorators.md
```diff
@@ -0,0 +1,134 @@
...
+is enabled.
+
+Keep event dispatch in the component that owns the algorithm when an event describes an
+internal milestone rather than the boundary of one delegated operation. Do not extract such
+dispatch into a decorator if this would require duplicating the algorithm, buffering a lazy
+stream or exposing implementation details through the interface.
+
+## Consequences
+
+- Core implementations remain independent of the event system.
```
maks-rafalko, line 106: Generally, I'm not sold on this idea tbh.

Probably it will be better understood when I dig into the code that follows this ADR, but as for me dispatching the events in decorators makes the understanding of the code *less* clear and tbh I'm not sure what is the benefit. 

> Core implementations remain independent of the event system.

why is it bad? you read the "core" code, see the dispatched event and you immediately know that some other code can rely on this events.

With decorators, you will have to 

1. understand this particular class is decorated
2. find a decorator
3. see its code. By this time you will already forget what was the "core" code

Also, as this ADR mentions, we have cases where we must dispatch events during the algorithm (so not before, not after) - `MutationGenerator`. In this case we will have a mix.. which is again not not clear why we are doing it.

I'm not saying it's a bad idea, more like writing the first impression for the future when I get back to it and sharing my thoughts.

theofidry: >why is it bad? you read the "core" code, see the dispatched event and you immediately know that some other code can rely on this events.

It is not bad per se, we have it in a few places and it is fine.

The problem I often encounter is we regularly end up with a class doing:

- a specific well defined action
- another action done conditionally with a fallback
- caching a result
- dispatching an event

There is two schools of thoughts here: either you like having everything in one place, or split it up.

The former has the net advantage of seeing it all at once, it is immediately clear what it does and you don't have to wonder how things are hooked up – container related stuff. It means looking at the core of the code is more noisy and the immediate cognitive load higher, as there is more side-effects and concerns to take in.

The latter makes the classes much simpler, so a lot easier to test and re-use. It makes reading the core of the code simpler and reduce the cognitive load _for that part_, but makes it harder to know how all the services are wired together get an overall understanding.

It is definitely a bias of mine as I really like to be able to access the core of the code with having as little noise as possible (e.g. see how the data collection is done without caring about the caching/event dispatching/fallback part) even if that makes knowing how the final service is configured harder.

>Also, as this ADR mentions, we have cases where we must dispatch events during the algorithm (so not before, not after) - MutationGenerator

Indeed that one does not fit that pattern, and so that rule is void. It is a valid usage so worth mentioning it here.

>I'm not saying it's a bad idea, more like writing the first impression for the future when I get back to it and sharing my thoughts.

What about removing the part about the event dispatchers then? It's only a piece of the ADR, it makes sense to me in regards to https://github.com/infection/infection/pull/3108, but if we exclude this PR then all the rest does apply to the current code.

theofidry: >What about removing the part about the event dispatchers then? It's only a piece of the ADR, it makes sense to me in regards to #3108, but if we exclude this PR then all the rest does apply to the current code.

Nvm I mixed things up, that is another ADR already (0009-event-and-subscriber-naming.md).

PR #3381, opened 2026-07-17
doc: Add an ADR for the final usage

theofidry:
This PR adds an ADR for the already adopted and enforced convention of `final` and `@final`.


adr/0012-final-classes-over-final-docblock.md
```diff
@@ -0,0 +1,46 @@
...
+of the code.
+
+Some services are still mocked by PHPUnit tests. Marking those classes with the
+`final` keyword would break those tests even when the production design has not
+changed. For such classes, the `@final` docblock records the architectural
+intent while preserving PHPUnit mockability.
+
+Using `@final` for classes that are not mocked weakens the signal. It makes the
+class look intentionally mockable when there is no test suite reason for it, and
+leaves the runtime inheritance contract looser than necessary.
```
sanmai: So what? I think it lacks the punchline.

adr/0012-final-classes-over-final-docblock.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Decision
+
+Use the `final` keyword for concrete source classes by default.
+
+Use `@final` without the `final` keyword only when a source class needs to stay
+mockable by PHPUnit tests.
+
+The exception must remain tied to the test need. When a class stops being mocked,
+replace `@final` with the `final` keyword.
```
sanmai:
```suggestion
When a class stops being mocked, replace `@final` with the `final` keyword.
```

adr/0012-final-classes-over-final-docblock.md
```diff
@@ -0,0 +1,46 @@
...
+
+Exceptional non-final classes that also cannot use `@final` need a dedicated
+architecture rule with a reason. `ParallelProcessRunner` is the known exception:
+it remains non-final only to allow PHPUnit partial mocks, and no class may
+extend it.
```
sanmai: Why can't it be annotated by `@final` isn't clear here.

AGENTS.md
```diff
@@ -132,19 +132,6 @@ analysers are PHPStan and Mago.
...
-Agents often mark everything `final`. Here, hard `final` is for classes never mocked (mutators,
-visitors, value objects, leaf utilities); services that tests mock get `/** @internal
-@final */` with NO keyword (e.g. `src/Mutation/Mutation.php`,
-`src/Configuration/ConfigurationFactory.php`). The PHPat finality rule accepts either form;
-adding the keyword to a mocked class breaks the suite. A third form exists for special cases:
-`ParallelProcessRunner` is plain `@internal` with a dedicated PHPat exemption whose test name
-states the reason ("intentionally non-final only to allow PHPUnit partial mocks",
-`tests/Architecture/PHPat/ClassesShouldBeFinalTest.php`). Mockability is the only accepted
-reason for `@final`. If no test mocks the class, use the keyword. Reviewers ask "is there a
-reason this is `@final` rather than `final`?" Have the answer ("it is mocked in X").
```
sanmai: Can we keep this? We don't transclude all ADR in the file so agents would not know to pull them. And they would not know how to use `@final`.

Also, not all harnesses pull @-mentioned files.

sanmai (changes requested):

adr/0012-final-classes-over-final-docblock.md
theofidry: On one hand true, on the other hand I don't really have a punchline. Since it's an already established convention, I didn't try to search to much about justifying this.

adr/0012-final-classes-over-final-docblock.md
theofidry: And actually I think it should be final... We already have the `ProcessRunner` interface. I think at some point it was inherited but no longer.

So let's skip that entire paragraph.

AGENTS.md
theofidry: >Also, not all harnesses pull @-mentioned files.

I think it's reasonable to expect them to. 100% compatibility for all agents won't happen, otherwise you'll have to worry about what agent prefers to have some HTML mixed in, clear single intent files over everything in one or vice-versa, it's a mess we can't really keep up with.

If we have an established OSS lower-tier model that we decide as the baseline I do not mind "downgrading", but otherwise trying to accommodate the lower tier of the market does not seem reasonable: it's too much of a moving target.

My understanding is that the agent should be able to pull all of the ADR files from the following instruction L93 (at the time of writing):

>- `adr/` - Architecture Decision Records; short rationale for standing conventions. Read
  before challenging one. New ADRs must follow the template `adr/0000-template.md`. Keep jargon to a minimum (ASD-STE100) when writing.

The indexing/following the files issue aside, we can also make an argument about what is better: a document with context showing examples, or a minimalist summary? What will the agent follow better? I have a feeling this will highly depend on the model.

The issue I see if we keep such duplicates (for things for which we have a human-first document) inlined in `AGENTS.md`, then the latter becomes a blackbox and it becomes hard to know what is within the realm of providing information to make it understand the codebase better, and what is an instruction to configure it.

That being said, I don't have super strong feelings about this. If you deem all of the above is BS I'm happy to comply.

adr/0012-final-classes-over-final-docblock.md
sanmai: Well, it looks like it tries to build the argument ("weakens the signal") but fails to build it (no definition to what is signal); what if we skip the entire paragraph?

AGENTS.md
sanmai: ADRs are great when you want to see the full argument. But here we have the gist of it: mark classes as  `@final` when they need mocking. If we remove the whole fragment, agents would not know our convention. So I think it is fair to make it shorter, and also mention that we don't do BypassFinals here, but removing it completely goes against the purpose of AGENTS.md, the guide for agents.

adr/0012-final-classes-over-final-docblock.md
sanmai: Agreed. Fixing the other point in #3393

adr/0012-final-classes-over-final-docblock.md
theofidry: the signal here is that it's final hence shouldn't be extended? If that's not clear enough I can add that in parenthesis.

sanmai (approved):
💯

PR #3384, opened 2026-07-17
ai: Suggest the agent to add an ADR when appropriate

theofidry:


AGENTS.md
```diff
@@ -102,6 +102,20 @@ transcluded here:
...
+dependencies, testing strategy, or a project-wide convention, tell the contributor that it
+is a candidate for an ADR. Suggest an ADR especially when credible alternatives exist or the
+same decision is likely to recur in reviews. First search `adr/`: update or supersede an
+existing decision instead of creating a competing record. Follow the criteria and workflow
+in `adr/README.md` and the template in `adr/0000-template.md`.
+
+Suggest the ADR; do not expand the task by writing one unless the contributor asks for it.
+Do not suggest ADRs for implementation descriptions, subsystem invariants, contribution
+workflows, command lists, or one-off details. Put those in architecture documentation,
+contributor documentation, or code comments as appropriate.
```
sanmai, line 117: Do we maintain an accessible index somewhere that an agent can grasp in a single turn? Cached tokens are cheap but not entirely free.

sanmai (approved):

AGENTS.md
theofidry: We do not, but in my experience Claude goes a bit off the rails when it notices something is out of date. So if we want to keep an index here for example, I'd suggest we find a an automated solution that can be done by invoking `make` and verified in the CI, preferably without AI.

maks-rafalko (approved):

AGENTS.md
theofidry: @sanmai do you think it's something worth pursuing or we should just add an index here in the AGENTS.md?

sanmai: Yes, we can do sed replace between set delimiters with a makefile generated index; probably will need a script to generate a markdown; I imagine claude/codex will create it in no time

theofidry: Will try this in a separate PR.

Issue #3322, opened 2026-07-04
RFC: Introduce an Infection Command Runner


theofidry:
## Summary

Introduce an Infection-owned command runner for common blocking command execution on top of [Symfony Process](https://symfony.com/doc/current/components/process.html).

The runner will streamline Symfony Process usage and provide an easier-to-test API. The goal is to simplify process-related code and make it easier to verify.

## Motivations

Infection uses `Symfony\Component\Process\Process` in many places. It is used to orchestrate tools such as PHP, Composer, and Git, and to evaluate mutants in separate processes.

Over the last year or so, the current usage has exposed recurring friction:

- It requires more boilerplate than necessary for most cases (see more details below).
- It is hard to test (see more details below).

In some cases, tool orchestration can be isolated well enough for this not to be a major issue. This was the case with the [Git API](https://github.com/infection/infection/issues/2622).

However, there have been several bugs in different areas, including fixes related to https://github.com/infection/infection/issues/2622 and https://github.com/infection/infection/issues/3144. This is not an exhaustive list.

Fixing those bugs is hard, not because the code changes are difficult, but because adding regression tests that show the broken behaviour and the intended fix is currently not practical.

### Streamlining the usage

The most common usage looks as follows:

```php
$process = new Process([
    $composerExecutable,
    'config',
    'bin-dir',
]);

$process->mustRun();
$composerBinDir = trim($process->getOutput());
```

This is a non-trivial amount of boilerplate for a very basic use case. This specific case is already covered by the existing `ShellCommandLineExecutor`:

```php
$composerBinDir = $shellCommandLineExecutor->execute([
    $composerExecutable,
    'config',
    'bin-dir',
]);
```

### Improve testability

When we have the following piece of code:

```php
$process = new Process([
    ...$this->findComposerExecutable(),
    'config',
    'bin-dir',
]);

$process->mustRun();
$composerBinDir = trim($process->getOutput());
```

It is not possible to mock `$process`. One could argue that such code should not be mocked because a unit test is unlikely to capture the full runtime reality. However, there are still benefits to mocking this interaction:

- It becomes possible to check that the command used to create the process was the expected one. For example, here we do not know what `$this->findComposerExecutable()` resolved to.
- It becomes possible to establish the intended behaviour in specific scenarios. Sometimes providing a real-world reproducer is too complex, so an integration test may not be an option. While such a unit test is not a complete representation of reality, it can still capture a specific regression and document the intent.

## Proposed Solution

[...]

PR #3385, opened 2026-07-18
feat(process): Introduce the ShellCommandLineExecutor interface

theofidry:
## Description

This PR introduces an interface for `ShellCommandLineExecutor`, located in the `TestFramework` contracts (see #3269). The goal is to allow test framework adapters to depend on `ShellCommandLineExecutor` (see https://github.com/infection/infection/issues/3322#issuecomment-5007787671). Whilst the current existing external test frameworks (PhpSpec and Codeception) do not depend on it, our internal ones (PHPUnit, PHPStan, Mago) do, hence IMO should be part of the contract.

This does mean `ShellCommandLineExecutor` will eventually land in `infection/abstract-test-framework`, at the moment at least. I still have some question marks on how it will work out to allow the test frameworks to test process-related things using this interface without re-implementing it, but it is a deeper discussion we can have once we polished the `TestFramework` interface and commit to move everything back to the separate packages.

## Changes

- Introduce the `ShellCommandLineExecutor` interface, which is carbon copy of the API of the existing `ShellCommandLineExecutor` concrete class.
- Rename the concrete class `ShellCommandLineExecutor` to `SymfonyProcessShellCommandLineExecutor`.
- Introduce `FakeShellCommandLineExecutor` and `DummyShellCommandLineExecutor` as simple utility implementations and leverage them.
- Tighten the concrete class `ShellCommandLineExecutor` by making it `final` and `readonly`. Mocking or children implementation now use the interface of the concrete class.

## Reviewer Notes

I do plan a rename and further changes, but I kept the changes to a minimum to ease the review.

## Related issues


src/Process/SymfonyProcessShellCommandLineExecutor.php
```diff
@@ -0,0 +1,61 @@
...
+use Symfony\Component\Process\Process;
+use function trim;
+
+/**
+ * @internal
+ */
+final readonly class SymfonyProcessShellCommandLineExecutor implements ShellCommandLineExecutor
+{
+    /**
+     * @param string[] $command
```
theofidry: to remove the PHPDoc here

src/TestFramework/Contracts/ShellCommandLineExecutor.php
```diff
@@ -57,8 +53,5 @@ class ShellCommandLineExecutor
      * @throws ProcessFailedException
      * @throws ProcessException
      */
-    public function execute(array $command): string
-    {
-        return trim((new Process($command))->mustRun()->getOutput());
-    }
+    public function execute(array $command): string;
```
theofidry, line 58: To improve the PHPDoc:


```suggestion
    /**
     * @param string[] $command
     *
     * @throws ProcessFailedException When process didn't terminate successfully
     * @throws RuntimeException When process can't be launched
     * @throws ProcessTimedOutException When process timed out
     * @throws ProcessSignaledException When process stopped after receiving signal
     */
    public function execute(array $command): string;
```

src/Container/Container.php
```diff
@@ -605,8 +606,9 @@ public static function create(): self
                 );
             },
             MemoizedComposerExecutableFinder::class => static fn (): ComposerExecutableFinder => new MemoizedComposerExecutableFinder(new ConcreteComposerExecutableFinder()),
+            ShellCommandLineExecutor::class => static fn (): ShellCommandLineExecutor => new SymfonyProcessShellCommandLineExecutor(),
```
sanmai:
```suggestion
            ShellCommandLineExecutor::class => SymfonyProcessShellCommandLineExecutor::class,
```

sanmai (approved):
Makes sense.

PR #3388, opened 2026-07-19
refactor(process): Rename ShellCommandLineExecutor to ShellCommandRunner

theofidry:
## Description

This PR renames `ShellCommnadLineExecutor::execute()` to `ShellCommandRunner::mustRun()`.

"Runner" describes the abstraction's role more naturally than “executor” and aligns with both Symfony Process (`run()`/`mustRun()`) and Infection's existing process-runner vocabulary. Renaming `execute()` to `mustRun()` also makes the current contract explicit: it returns stdout for a successful command and throws when the command fails. 

**This name change leaves the API clear for a non-throwing `run()` operation which will be added later.**

## Changes

- Renames `ShellCommandLineExecutor` to `ShellCommandRunner`.
- Renames the Symfony-backed implementation, fake, benchmark dummy, variables, properties, and container getter consistently.
- Renames `execute()` to `mustRun()` across production code and tests.

## Reviewer Notes

The reason for the renaming is that I want to introduce a non-throwing operation later, so mirroring the Symfony `Process` – which is what we use, was a no brainer.

However, if you prefer to stick to "executor", then I will rename `ShellCommnadLineExecutor::execute()` to `::mustExecute()` and keep `::execute()` for the non-throwing variant. 

## Related issues


sanmai (approved):
spot checked the changes; the direction makes sense

PR #3389, opened 2026-07-19
feat(process): Introduce ShellCommandRunner::run()

theofidry:
## Description

Adds a non-throwing execution path to `ShellCommandRunner` for callers that need to inspect an unsuccessful command instead of throwing a `ProcessFailedException`.

The existing `mustRun()` operation remains the concise happy-path API: it returns trimmed stdout and throws when the command fails. The new `run()` operation complements it by returning a `CompletedProcess` value object containing the command, exit code, stdout, and stderr.

As the first consumer, `AdapterInstaller` now uses the runner instead of constructing Symfony `Process` directly. This preserves its existing behavior of continuing after a non-zero Composer exit while making the command and timeout straightforward to verify in a unit test.

## Changes

- Adds `ShellCommandRunner::run()` with callback and timeout support.
- Adds the immutable `CompletedProcess` result, including `isSuccessful()`.
- Implements `run()` in the Symfony-backed runner and updates the fake and benchmark dummy.
- Migrates `AdapterInstaller` from direct Symfony Process usage to `ShellCommandRunner`.
- Adds `AdapterInstaller` tests.

## Related issues


tests/phpunit/TestFramework/AdapterInstallerTest.php
```diff
@@ -0,0 +1,127 @@
...
+
+    private ShellCommandRunner&MockObject $shellCommandRunner;
+
+    private AdapterInstaller $adapterInstaller;
+
+    protected function setUp(): void
+    {
+        $this->composerExecutableFinder = $this->createMock(ComposerExecutableFinder::class);
+
+        $this->shellCommandRunner = $this->createMock(ShellCommandRunner::class);
```
theofidry, line 62: note that this test was not possible before due to using `new Process()`

src/Process/SymfonyProcessShellCommandRunner.php
```diff
@@ -48,4 +51,37 @@ public function mustRun(array $command): string
...
     }
+
+    public function run(
+        array $command,
+        ?Closure $callback = null,
+        ?float $timeout = self::DEFAULT_TIMEOUT,
+    ): CompletedProcess {
+        $process = new Process(
+            $command,
+            timeout: $timeout,
```
sanmai, line 62: I wonder if `mustRun` could also use this timeout.

src/Process/SymfonyProcessShellCommandRunner.php
```diff
@@ -48,4 +51,37 @@ public function mustRun(array $command): string
...
+        $exitCode = $process->getExitCode();
+        Assert::integer($exitCode);
+
+        return new CompletedProcess(
+            $command,
+            $exitCode,
+            trim($process->getOutput()),
+            trim($process->getErrorOutput()),
+        );
+    }
```
sanmai, line 86: Sounds like this could be inlined, or is there another PR that needs that too?

sanmai (approved):
Makes sense overall

src/Process/SymfonyProcessShellCommandRunner.php
theofidry: Yes! Will be done in a follow up PR.

src/Process/SymfonyProcessShellCommandRunner.php
theofidry: I can't recall but doesn't hurt to do it right now.

PR #3390, opened 2026-07-19
feat(process): Add execution context options to ShellCommandRunner

theofidry:
## Description

Expands both `ShellCommandRunner::mustRun()` and `ShellCommandRunner::run()` with the execution context needed by common blocking Symfony Process usages: streamed output callbacks, a working directory, environment variables, input, a total timeout, and an idle timeout.

Keeping these options on the Infection-owned abstraction allows callers to migrate away from constructing Symfony `Process` directly without losing configuration capabilities or testability.

## Changes

- Adds callback, working-directory, environment, input, timeout, and idle-timeout parameters to `mustRun()` and `run()`.

## Related issues


src/Process/SymfonyProcessShellCommandRunner.php
```diff
@@ -74,4 +99,22 @@ public function run(
...
+        array $env,
+        mixed $input,
+        ?float $timeout,
+        ?float $idleTimeout,
+    ): Process {
+        $process = new Process($command, $cwd, $env, $input, $timeout);
+        $process->setIdleTimeout($idleTimeout);
+
+        return $process;
+    }
```
sanmai, line 119: You can unit test this thoroughly with

```
protected function newProcess(...): Process {
 return new Process(...);
}
```
which you can then mock the class itself with `onlyMethods(['newProcess'])` and then making this method return a mocked process

so most of actual process launching can be avoided

sanmai (approved):
Makes sense. Left a testing suggestion, non-blocking.

src/Process/SymfonyProcessShellCommandRunner.php
sanmai: How it looks like in another place

https://github.com/infection/infection/blob/2e560ebd88e7531630b61094fc6d4b3add6ea8fe/tests/phpunit/Process/Runner/ParallelProcessRunnerTest.php#L419-L433

theofidry: I think it made sense in `ParallelProcessRunnerTest` but less so here.

Having integration tests for `ParallelProcessRunner` is bound to cover only a very limited part of the observable behaviour. Testing it further does require mocking and I think you're trick is great there in the current state of things.

Here the current tests can be integration tests yet cover all the expected behaviour, rather than having a unit test purely mocking how the Symfony `Process` behaves. So theoretically, if tomorrow we switch to another `Process` class, those tests should still hold up.

PR #3393, opened 2026-07-19
Clarify finality requirement in AGENTS.md

sanmai:
## Description

Follow-up to #3381.

## Changes

- Makes `ParallelProcessRunner` soft-final like it should be.

## Checklist


theofidry (approved):
I would rather split the two as one is a refactor and one is doc but I guess it's just my habits kicking in.

I also assume `ParallelProcessRunner` couldn't be made `final` for some reasons.

PR #3395, opened 2026-07-20
ai: Add ADR listing generation in AGENTS.md

sanmai:
## Description

As per #3384, adds a script to inject and update the listing of ADR records.

`make autoreview` fails if the list goes out of date.


Makefile
```diff
@@ -109,6 +110,16 @@ sbx-image-test:
...
 
+.PHONY: update-agents-adr-list
+update-agents-adr-list:	## Updates the generated ADR list in AGENTS.md
+update-agents-adr-list:
+	./devTools/update-agents-adr-list.sh "$(AGENTS_MD)"
+
+.PHONY: check-agents-adr-list
+check-agents-adr-list:	## Checks the generated ADR list in AGENTS.md
+check-agents-adr-list: update-agents-adr-list
+	git diff --exit-code -- "$(AGENTS_MD)"
```
sanmai, line 121: Codex really want to build a check script, but a diff check works just as well and just a single line.

devTools/update-agents-adr-list.sh
```diff
@@ -0,0 +1,35 @@
...
+            continue
+        fi
+
+        title=$(sed -n 's/^# //p; /^# /q' "$file")
+
+        printf -- "- [\`%s\`](%s) - %s\n" "$file" "$file" "$title"
+    done
+
+    sed -n '/<!-- adr-list:end -->/,$p' "$agents_md"
+} > "$tmp_file"
```
sanmai, line 35: - read the chunks before the marker
- build a new list
- read the chunks after the marker

devTools/update-agents-adr-list.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+agents_md=${1:-AGENTS.md}
+
+# Bail out before writing anything: without exactly one of each marker the
+# splice below would silently duplicate or truncate the file.
```
sanmai:
```suggestion
# There should be just one marker of each kind
```

Makefile
theofidry: I does, but it is a bit clunky especially if you're switching branches with different ADRs. But it's enough of a small edge-case to not be a blocker either.

theofidry (commented):
Ha apologies I did my PR #3396 without seeing yours.

As it is, I'm totally fine with the script although I'll admit in its current form it is the sort of scrip I'll probably never be able to touch myself: adjusting/debugging bash has never really been a nice experience IMO.

So if you prefer this over #3396, np. The latter definitely has more boilerplate.

One thing to keep in mind before deciding however is how to address the ADR "AI summary". In another thread you mentioned how the problem with the current ADRs is that they are a bit too verbose for the agent and hence having a summary here in `AGENTS.md` was useful. I think one way to solve this is a similar anchor-based replacement strategy. For example, `AGENTS.md` could have:

```
<!-- adr-summaries:start -->
<!-- adr-summaries:end -->
```

For each ADR, we add a section `Summary for AI` (and to the template too) and we could take from the ADR the title + the content of that summary and place it in that section in AGENTS.md.

This way, not only we ensure we have an up to date list of ADRs listed in `AGENTS.md`, but we have a deterministic way to keep it up to date with a more concise summary.

sanmai:
Either the titles should work as a one-line summary (the most straightforward) or we could be adding a line prefixed such as with "Summary:" and using it. That said, I would not hang on a summary very hard: the file itself should have the outcomes in a condensed form anyway, so although extended summaries should not hurt, they should not be necessary either.

theofidry:
>the file itself should have the outcomes in a condensed form anyway, so although extended summaries should not hurt, they should not be necessary either.

I'm confused, wasn't it what you were suggesting in https://github.com/infection/infection/pull/3381#discussion_r3610226738?

Alternatively, maybe `Summary` is not always necessary, e.g. if `Decision` is already sufficient. But I admit I like the ADR to be more verbose sometimes to give more context or examples, without worrying too much about whether it's a bit excessive for an AI. So we could also have that fallback as an option instead of requiring `Summary`.

maks-rafalko (approved):

sanmai:
I guess I used the word summary too liberally. But I could be wrong and agents may need a longer summary. In this case we can either use more descriptive headers, or go with a more complicated approach. What I think we must do is surface these ADRs even as a file list, because file names and titles already bring enough clues.

theofidry:
@sanmai I am confused by what should be done there then.

Currently the `AGENTS.md` has the auto-generated list of ADRs – good.

But now we have those ADRs "summaries" in there which are not and I would like to get rid of. Either we decide listing the ADRs are sufficient and we remove them, or we added a new section in the ADR for the AI* which can be copied to `AGENTS.md`.

I'd prefer the 2nd option, but then I feel like a bash script is not the best to do this.

*: this section would be optional, only necessary if the ADR is a bit verbose and the title is insufficient to convey what needs to be done.

sanmai:
I'd ask what is the reason we have to have ADR summaries in the first place there, as if we want to argue with agents. 

I understand for us humans ADRs are important because that's what we agree on. But agents, they are tools. There's no arguing and no proving points for the sake of proving. They aren't people.

theofidry:
>But agents, they are tools. There's no arguing and no proving points for the sake of proving. They aren't people.

I agree with you, but this goes to your point that a more synthesised version of the ADR is sufficient for the AI (unless the task is to challenge it).

>I'd ask what is the reason we have to have ADR summaries in the first place there, as if we want to argue with agents.
Tomorrow you add an ADR, you have two situations:

- status quo: you add the ADR, you run the make command to update the list and then either you cross your fingers that the agent picked up the instruction from `adr/README.md` to add a section in `AGENTS.md` or you need to explicitly ask it to do so.
- with a "summary" (FWIW I think the current "Decision" section is likely sufficient most of the time), this can be part of the ADR template, this will be part of the active changes.

So in the second case the (human) attention can be given to the proposed document, the rest of the changes are automation.

PR #3396, opened 2026-07-20
ai: Add the ADR index list to AGENTS.md

theofidry:
## Description

Follow up of https://github.com/infection/infection/pull/3384/changes#r3613071151.

This PR introduces two new make commands `adr-index-{update,check}` which will ensure we have an up to date list of ADRs in `AGENTS.md`. 

`adr-index-check` has been added to `autoreview` so ensuring the list is up to date will also be done in the CI.

## Changes

Since there was a bit too many regexing to do for my liking, I added a PHP script `devTools/update-adr-index.php`. To allow minimal testing, I added the used classes to `tests/phpunit/AutoReview/AI/`. It is not the most perfect location but it allows to easily bootstrap utilities with tests without setting up an entire project just for a small script.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,6 +116,19 @@ Do not suggest ADRs for implementation descriptions, subsystem invariants, contr
 workflows, command lists, or one-off details. Put those in architecture documentation,
 contributor documentation, or code comments as appropriate.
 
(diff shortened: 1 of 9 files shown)


theofidry:
Closed in favour of #3395.

Issue #2332, opened 2025-07-12
BreakContinueReturn Mutator


sanmai:
**Is your feature request related to a problem? Please describe.**

I see a persistent pattern in how Infection mutates `break` into `continue`, and if you replace it with the counterpart, it will mutate `continue` into `break` with all mutations happily escaping the tests.

For me, it means it has to be `return` here. And usually it is very easy to test.

```diff
- break;
+ return;
```

**Describe the solution you'd like**

I'd like a mutator that will replace `break;` and `continue;` with `return;`.

Initially, in methods:

- without any return type
- with `void` return type
- with `Generator` return type
- or containing `yield` / `yield from` statements (`return;` is legal there)

**Additional context**

https://github.com/infection/infection/pull/2322/files#r2202918062

PR #1483, opened 2021-02-17
WIP: Allow mutation of functions

BackEndTea:
This PR:

- [x] Adds new feature: Allows mutation of functions
- [ ] Covered by tests
- [ ] Doc PR: https://github.com/infection/site/pull/XXX

Fixes #1482

PR #3400, opened 2026-07-22
feat: Add ContinueReturn mutator

maxm86545:
## Description

`Continue_`/`Break_` mutations often "flap": `continue -> break;` escapes the tests, and once the code is refactored to the counterpart, the reverse mutation escapes as well (#2332). Replacing the jump with a `return` skips both the remaining iterations and everything after the loop, so the mutant is killed by a simple assertion on the post-loop behaviour.

The new `ContinueReturn` mutator replaces `continue;` (and `continue N;`) with a return statement matching the enclosing function signature:

| Enclosing function | Mutation |
|---|---|
| no declared return type, `void`, generators | `return;` |
| `?T`, `T\|null`, `mixed`, `null` | `return null;` |
| `int` / `float` / `string` / `bool`, `false`, `true` / `array`, `iterable` | `return 0;` / `0.0` / `''` / `false`, `true` / `[]` |
| objects, interfaces, enums, non-null unions, intersections, `never` | not mutated — no value is guaranteed to compile and run |

No mutant is produced when it would be equivalent or a duplicate:

- nothing observable executes after the target loop, or the first post-loop statement returns the very same value or is itself a jump — such a mutant would only duplicate `continue -> break;` of the `Continue_` mutator;
- the post-loop successor is resolved along the execution path, not the document order: sibling `elseif`/`else`/`catch` branches are never successors;
- `continue N;` counts enclosing switches as levels, exactly as PHP does; jumps targeting a switch are skipped.

## Changes

- Adds the `ContinueReturn` mutator to the `@operator` profile.
- Registers it in `ProfileList` and `resources/schema.json`.
- Test cases: all loop types, jump levels, the return-type ladder, generators, closures, every suppression rule.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated https://github.com/infection/site/pull/305
- [x] Appropriate labels applied (`Feature`, `Mutator`)

## Reviewer Notes

- The suppression logic lives in `duplicatesContinueToBreakMutant()`: both mutants agree up to the loop exit, so everything they can differ by happens after the loop.
- Plain functions are not mutated: the pipeline does not scope `Node\Stmt\Function_` bodies (see #1483); a dedicated test documents this.
- A loop wrapped in a block-form `declare(ticks=N) {}` is treated conservatively: the successor search does not look inside the declare block, so such a `continue` is not mutated rather than risking a duplicate; a dedicated test documents this.

## Related issues

Part of https://github.com/infection/infection/issues/2332 — the `continue` half.


src/Mutator/Operator/ContinueReturn.php
```diff
@@ -0,0 +1,457 @@
...
+            return 1;
+        }
+
+        // PHP only compiles positive integer literals as jump levels; anything
+        // else the parser may have accepted cannot run and is not worth mutating.
+        if ($node->num instanceof Node\Scalar\Int_ && $node->num->value >= 1) {
+            return $node->num->value;
+        }
+
+        return null;
```
theofidry, line 186: so when can the jump level be `null`?

src/Mutator/Operator/ContinueReturn.php
```diff
@@ -0,0 +1,457 @@
...
+    }
+
+    /**
+     * The return statement to substitute for the jump, or null when the function
+     * signature admits none.
+     */
+    private static function findReplacementReturn(Node $node): ?Node\Stmt\Return_
```
theofidry: `find*` hints that it looks up for a node, but in practice this method creates the replacement node, so I would rather name it `findReplacementReturn()`

src/Mutator/Operator/ContinueReturn.php
```diff
@@ -0,0 +1,457 @@
...
+
+        if ($functionScope === null) {
+            return null;
+        }
+
+        $returnType = $functionScope->getReturnType();
+        $attributes = NodeAttributes::getAllExceptOriginalNode($node);
+
+        if ($returnType === null) {
+            return new Node\Stmt\Return_(null, $attributes);
```
theofidry: You could group those two:

```suggestion
        if (
            $returnType === null
            || $returnType instanceof Node\Identifier && $returnType->toLowerString() === 'void'
        ) {
            return new Node\Stmt\Return_(null, $attributes);
```

theofidry (commented):
I think the idea is good overall.

Before considering a merge, there is two major issues that should be addressed:

- This mutator claims to be a circumstantial replacement for the `Continue_` and `Break_` mutators. However, currently it is purely additive. A user encountering unproductive mutators before will still get them, the only difference is that it may get a few more mutators that may not escape. I admit I am not sure what the solution is here.
- This mutator relies on _a lot_ of node search and I do not think this should be the direction of the mutators. It is an easy way to do and allows mutators to be more self-contained, so it is a natural approach. However, I think if every mutator is implemented this way then performances will be dreadful. Infection does two passes, the enrichment pass and the mutation pass. the idea is that during the enrichment pass we can add the necessary information to the nodes as attributes, or we could also collect and store state in a separate service that could be added to the mutation pass so that mutators can leverage that information.

The 2nd point in peculiar may not be easy work and I am not sure about the scope of the changes either. I recommend waiting on another review before proceeding, I don't want to throw you in a direction that will not be productive.

Alternatively we could decide it is not a problem for this singular mutator and that is a problem I could work on separately afterwards and evaluate at the same time if there is other mutators suffering from the same issue.

src/Mutator/Operator/ContinueReturn.php
maxm86545: When the level parses but cannot compile (`continue $level;`, `continue 0;`) - the "positive integer literal" rule is enforced by PHP's compiler, not by the parser, and Infection also mutates files that are never loaded during the initial test run. See the `It does not mutate a continue with a dynamic level` test case.

src/Mutator/Operator/ContinueReturn.php
maxm86545: Renamed to `createReplacementReturn()`, plus `findDefaultReturnExpression()` to
`createDefaultReturnExpression()` for consistency.

maks-rafalko:
> Continue_/Break_ mutations often "flap": continue -> break; escapes the tests, and once the code is refactored to the counterpart, the reverse mutation escapes as well (https://github.com/infection/infection/issues/2332).

could you please update the description with a self-explanatory example of where and why the current 2 mutators are bad, as I don't follow it. If the mutant forces you to change the business logic (so you update the code from `continue` to `break`) - why do we expect our tests to catch the same mutants? I would appreciate a step-by-step explanation of the problem

- minimal code of the problem and the test covering it
- mutation that happens
- what is changed and where to kill that mutation
- and then what's wrong with the escaped mutants

sanmai:
I've experienced this continue/break flapping myself, so the issue is real. But I could not find right away where I had it, so if @maxm86545 has it, it would be great to see a [demo on the playground](https://infection-php.dev/).

maks-rafalko:
yep. Just to be clear, I have no doubts there is an issue since you both experienced it. At the same time, @maxm86545 It would be cool to fully understand it and have an example. This will help to

- think about potential solutions together with the full context
- easily remind ourselves in the future what was the issue exactly and how it could be reproduced

maxm86545:
@maks-rafalko @sanmai minimal example, code and test are in the playground links.

The loop skips expired messages from a list that the class keeps sorted freshest-first. Once one message is expired, all following ones are too, so `continue` and `break` behave identically for any state the class can hold. No test can tell them apart without breaking the class' own invariant.

1. `Continue_`: `continue → break` escapes, nothing can kill it: https://infection-php.dev/r/oww0
2. Both keywords are correct, `break` is arguably more precise, so the developer switches.
3. `Break_`: `break → continue` escapes with the same test: https://infection-php.dev/r/7ddv

Whichever keyword you write, you get an unkillable mutant on this line. To your question: the switch in step 2 does not change behaviour, so there is no new behaviour for tests to catch; the mutant is just noise in either direction.

`ContinueReturn` asks a different question on the same line. `continue → return;` leaves not only the loop but the whole method, so `$this->messages = []` after the loop never runs. The existing test catches that: the second `flush()` sends the message again and `assertSame([20], ...)` fails. So the mutant is killed, and it is killed for the right reason: the test really does verify what happens after the skip, which is the thing `continue → break` could never tell us.

PR #3413, opened 2026-07-24
refactor(phpunit): Pass the PCOV directory as a raw argument

theofidry:
## Description

Pass Infection's inferred `pcov.directory` value directly in the PHP argv list rather than applying shell escaping.

Symfony Process executes the initial PHPUnit command from a `list<string>`, so each entry is already an isolated argument and does not require shell escaping. We addressed the same class of issue in, for example, https://github.com/infection/infection/pull/3147.

However, in this specific case, PHP removes the matching quotes added by `escapeshellarg()` when parsing `-d`. Therefore, this does not cause an observable bug.

This PR retains the change as a clean-up because passing the raw value better reflects the argv contract and requires less code. It also adds an end-to-end test to cover this behaviour in case PHP's handling changes in the future.

## Changes

- Remove `escapeshellarg()` from the generated `pcov.directory` PHP argument.
- Make the PHPUnit adapter expectations platform-independent for PCOV directory values.
- Add a PCOV end-to-end fixture whose source directory contains spaces.


tests/e2e/PCOV_Directory_With_Spaces/source files/SourceClass.php
```diff
@@ -0,0 +1,13 @@
...
+
+declare(strict_types=1);
+
+namespace PCOVDirectoryWithSpaces;
+
+final class SourceClass
+{
+    public function returnsTrue(): bool
+    {
+        return true;
```
maks-rafalko: we have `Example_Test` that is used with `tests/add_new_e2e`.

This e2e test is either generated by some other command by a mistake (or just AI) or for some reason has the different code comparing to `Example_Test`.

If this is generated by AI, we need to update our AGENTS.md to say AI to use `add_new_e2e` and not create anything artificially, to avoid confusion and inconsistencies.

tests/e2e/PCOV_Directory_With_Spaces/README.md
```diff
@@ -0,0 +1,16 @@
+# PCOV directory containing spaces
+
+This scenario covers a project whose source directory contains spaces. When PCOV has no
+configured `pcov.directory`, Infection provides the source directory to the initial test run
+as a PHP `-d` argument.
+
+Symfony Process receives the command as an argument list, so the directory must remain
+unescaped. Shell quotes would become part of PCOV's configured directory and prevent it from
```
maks-rafalko: so, before this patch, do we say here that `escapeshellarg()` broke cases where a folder contains spaces?

tests/e2e/PCOV_Directory_With_Spaces/source files/SourceClass.php
theofidry: I copied another existing e2e test, I admit I can't remember the last time I used `add_new_e2e` and maybe I should have another look at it.

Regarding the AI it should already be aware of it, but whether it will follow that instruction is a different story.

tests/e2e/PCOV_Directory_With_Spaces/README.md
theofidry: no it worked just fine.

The only reason I dug into this is because _normally_ it would be a problem and it's a problem I've already fixed in other places so it came under my radar.

The one and only reason it did not cause any issue is because PHP CLI parses options differently, being a lot more tolerant than most programs.

Which is why I marked it as refactor rather than bugfix. I still did this patch nonetheless because the resulting source code is simpler and I do not know if that sort of behaviour from PHP is intentional or not and may change in the future (so there is value IMO to have an e2e test for it).

tests/e2e/PCOV_Directory_With_Spaces/source files/SourceClass.php
maks-rafalko: > I can't remember the last time I used `add_new_e2e` and maybe I should have another look at it

this was even documented by you https://github.com/infection/infection/commit/ffa9a43576163bef44d5607c1987b045b941a2e0 😃

theofidry: yes, but in the end I ended more copying existing ones than creating brand new ones from scratch 😅

tests/e2e/PCOV_Directory_With_Spaces/README.md
theofidry: I updated the README.md

maks-rafalko (approved):

tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapter/PhpUnitAdapterTest.php
```diff
@@ -1080,15 +1077,13 @@ public static function initialTestRunProvider(): iterable
...
+        yield 'with PCOV directory containing spaces' => [
             $default
                 ->withPcovDirectory('/path with spaces/src')
                 ->withExpected([
                     self::PHP_EXECUTABLE,
                     '-d',
-                    OperatingSystem::isWindows()
-                        ? 'pcov.directory="/path with spaces/src"'
-                        : "pcov.directory='/path with spaces/src'",
+                    'pcov.directory=/path with spaces/src',
```
sanmai, line 1086: so since the argument itself (the whole `pcov.directory=...`) is escaped as a whole, we don't need to additionally escape the path inside of it, right?

sanmai (approved):

tests/phpunit/TestFramework/PhpUnit/Adapter/PhpUnitAdapter/PhpUnitAdapterTest.php
theofidry: correct. Otherwise you would end up with a double escape which would normally cause problems (it isn't because of PHP CLI level of tolerance)

PR #3417, opened 2026-07-26
ci: Record PHPUnit version and CPU steal in performance metrics

sanmai:
See: https://github.com/infection/infection-metrics/pull/4

## What and Why

Looking at [our metrics dashboard](https://infection.github.io/infection-metrics/), we can see that the wall clock per mutation went up semi-recently.

But before we raise the alarms about a potential regression, we have to normalize the number further against other known factors. Specifically, 
- PHPUnit version, as time spent running it adds up to the total, so if PHPUnit becomes slower in a new version, we have to be able to isolate that;
- and the [time the hypervisor gave someone else](https://unix.stackexchange.com/questions/264958/iostat-what-does-the-steal-field-mean) (`%steal` or `%st`). So if we get a certain performance in one run, and reduced performance in another, we can use this percentile to find the actual number of CPU seconds our processes used.

Sister PR that adds an option to save these metrics: 
https://github.com/infection/infection-metrics/pull/4


.github/workflows/perf-metrics.yaml
```diff
@@ -133,8 +141,18 @@ jobs:
...
+          grep '^cpu ' /proc/stat > stat-after.txt
           cat timing.json
 
+          # CPU steal over the run: jiffies the hypervisor gave to other tenants.
+          # paste joins the snapshots, so fields 2-11 are before, 13-22 after; steal is field 9.
+          steal_percent=$(paste stat-before.txt stat-after.txt | awk '{
+            for (i = 2; i <= 11; i++) total += $(i + 11) - $i
+            printf "%.2f", (total > 0 ? 100 * ($20 - $9) / total : 0)
+          }')
+          echo "steal_percent=$steal_percent" >> $GITHUB_OUTPUT
```
theofidry: what is this "steal percent"?

sanmai: steal% is time the hypervisor gave our vCPU to someone else

CPU is same among many runs, so we can normalize the perf between a pair of runs using this %%

 I'll write a proper description hopefully in a bit

.github/workflows/perf-metrics.yaml
```diff
@@ -134,8 +145,17 @@ jobs:
...
+
+          # Send SIGINT to sar so it outputs the Average summary line
+          kill -INT $sar_pid 2>/dev/null
+          wait $sar_pid 2>/dev/null || true
+
           cat timing.json
 
+          # Extract steal % from sar's Average line (second-to-last field is %steal)
+          steal_percent=$(grep "Average:" sar-output.txt | tail -1 | awk '{print $(NF-1)}')
+          echo "steal_percent=${steal_percent:-0.00}" >> $GITHUB_OUTPUT
```
sanmai, line 157: it outputs something like this:
```
18:29:18        CPU     %user     %nice   %system   %iowait    %steal     %idle
18:29:19        all      0.93      0.00      0.47      0.00      0.00     98.60
18:29:20        all      1.19      0.00      0.61      0.11      0.00     98.10

Average:        all      1.06      0.00      0.54      0.05      0.00     98.35
```
So we only need the second number from the end in the average row.

theofidry (approved):
I admit I'm still not 100% clear on how to leverage that information, neither am I too worried about minor performance degradations: the biggest possible win we can have at the moment is through better test framework integration and orchestration, which are both in the works. IIRC there was also a few changes that were degrading (very minor) performances but the original behaviour was wrong.

PR #3419, opened 2026-07-26
ci: Ignore zizimor error in the metrics collection workflow

sanmai:
## Description

The workflow runs on tags, and Zizimor assumes that it helps with released artifacts, so it flags the use of caches. But we don't touch release artifacts here.

## Changes

- Silence an irrelevant issue.

### DIFF
--- a/.github/workflows/perf-metrics.yaml
+++ b/.github/workflows/perf-metrics.yaml
@@ -67,7 +67,7 @@ jobs:
         run: echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
 
       - name: Cache dependencies
-        # zizmor: ignore[cache-poisoning] the tag trigger only collects metrics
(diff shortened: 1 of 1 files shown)


.github/workflows/perf-metrics.yaml
```diff
@@ -67,6 +67,7 @@ jobs:
         run: echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
 
       - name: Cache dependencies
+        # zizmor: ignore[cache-poisoning] This workflow only collects metrics (no publishing) and runs with read-only repo permissions.
```
theofidry, line 70:
```suggestion
        # zizmor: ignore[cache-poisoning] Tag runs publish only performance data
  to the metrics repository; they do not publish release artifacts.
```

theofidry (approved):

.github/workflows/perf-metrics.yaml
theofidry: it is otherwise not correct, we do publish something

PR #3421, opened 2026-07-26
ai: Add a handoff skill

theofidry:
## Description

Add a shared handoff skill that preserves unfinished work in a concise, self-contained brief so another agent session can resume without reconstructing conversational context.

Handoffs are written to the agent-neutral, Git-ignored `var/agent/` directory. The skill captures transient decisions, constraints, failed approaches, and next steps while pointing to repository artifacts instead of duplicating them.

## Changes

- Adds the canonical Codex handoff skill and its UI metadata.
- Defines the handoff structure, storage location, safety rules, and resume workflow.
- Exposes the same skill to Claude through a symlink, avoiding duplicate definitions.

## Reviewer Notes

The skill disables implicit invocation because creating a handoff is an explicit transition
between sessions.

### DIFF
--- a/.codex/skills/handoff/SKILL.md
+++ b/.codex/skills/handoff/SKILL.md
@@ -1,0 +1,45 @@
+---
+name: handoff
+description: Persist unfinished work in a self-contained brief so a fresh Codex session or another workspace can resume without re-deriving conversational context. Use when context has become too large or fragmented, work must move between sessions or sandboxes, or the user explicitly asks for a handoff.
+---
+
+# Handoff
+
+Create a concise handoff for unfinished work. Preserve information that exists only in the
+conversation; point to information already stored in the repository instead of copying it.
+
+## Workflow
+
+1. Inspect the current Git branch, working tree, and relevant on-disk artifacts. Do not
+   modify, commit, discard, or clean working-tree changes while preparing the handoff.
+2. If the user supplied a focus, make that the handoff's priority.
+3. Write `var/agent/handoff.md`. Use a more specific filename under `var/agent/` when the
+   user requests one or when preserving multiple handoffs. This repository ignores `var/`,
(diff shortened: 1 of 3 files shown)


sanmai (approved):
Looks like a common handoff skill.

.codex/skills/handoff/SKILL.md
```diff
@@ -0,0 +1,45 @@
+---
+name: handoff
+description: Persist unfinished work in a self-contained brief so a fresh Codex session or another workspace can resume without re-deriving conversational context. Use when context has become too large or fragmented, work must move between sessions or sandboxes, or the user explicitly asks for a handoff.
```
sanmai, line 3: I've seen this mentioned before, but it escaped my how's that materially different from the regular compaction beyond saving the summary into a file and a more condensed format. Or is it the whole idea?

(For hard or long tasks I usually ask the agent to maintain a log of what it planned/did/tried/failed/etc, starting with the initial hypothesis/problem. And then this running log works as a hand-off for a new session. It works great for Claude (it is very happy to follow along) and I can imagine it working for Codex just as well.)

theofidry: >how's that materially different from the regular compaction

The main difference is what it can be used for. If you can get away with compaction, use it.

The main reason for using it is that the sandbox is ephemeral and may be re-created at any time. So I use it when I know I need to hand over work to a fresh agent which does not have access to the previous memories.

Sometimes it ends up being the same agent, but it's at a much later time to having a file is one safer (in case I forget about it and reset the sandbox meanwhile) and second I kinda prefer to have those list of things to do at the FS.

So if you have a polluted session and you want to save a memory and restart from that, that skill is not very useful.

theofidry:
>Looks like a common handoff skill.

Yes, nothing special about it

PR #3425, opened 2026-07-27
ADR: priorities of Infection

maks-rafalko:
**The main reason** of this PR is that I completely overwhelmed by the number of RFCs / PRs / issues to review in the recent months and my inbox keeps growing, being at ~100 unread issues/PRs at peaks. This became completely unmaintainable and brings a lot of frustration.

I absolutely appreciate all the time that has been put into all the work, but we must have a maintainable process, which this ADR proposal should set.

**The goals** of this ADR are: 

- to reduce the number of unwanted changes (even AI noticed the number of "whys" I added to PRs :))
- to avoid any kind of "I like this more than that" PRs
- focus on bug fixing and features again
- do not touch the code that worked for years if there is no a real justified reason for this (read the ADR, about "Enabling refactoring")

While I always strive to be polite, I'm bad at writing such things. If something sounds rude and/or offending for any of the reviewers or future readers - apologies, let's make it better together. 

At the same time, I intentionally made it very strict, with a strong message.

Everything what is proposed here is to help shipping faster **the things that are important for the Infection's end users**.

I expect it to be tough, and that's great. Let's discuss.

### DIFF
--- a/adr/0014-priorities.md
+++ b/adr/0014-priorities.md
@@ -7,11 +7,10 @@ much the project can absorb, and every pull request spends part of that capacity
 
 A large number of pull requests neither fix a reported bug nor add a requested feature. They
 often restructure code, or add tooling that no current task needs.
-Such a change is harder to review than a bug fix. A bug fix carries its own purpose: the
-linked issue says what was wrong. A restructuring usually does not, so the reviewer has to work
+Such a change is harder to review than a bug fix. A bug fix has a purpose: the
+linked issue identifies a problem. A restructuring usually does not, so the reviewer has to work
 out why it is proposed at all, and whether it brings any bug fix or feature closer. That answer
-is frequently missing from the pull request, and asking for it costs more time than reading the
-diff.
+is frequently missing from the pull request, and the answer to this question is more important than the diff.
 The result is that reported bugs and accepted features wait behind changes that no user asked
 for.
 
--- a/adr/0014-priorities.md
+++ b/adr/0014-priorities.md
@@ -102,30 +102,6 @@ visible instead of remembered, and so that the queue can be read in one place. T
 issues as well as pull requests, because a proposal that waits for a decision, an RFC for
 example, spends the same review capacity as a diff.
 
-### Mapping onto GitHub
-
-This subsection describes the current mechanism only. It can change without a new ADR, provided
-that the two axes stay recorded and readable.
-
-A pull request states its type in the `Type of change` section of the pull request template.
-GitHub cannot validate a template, which is why enforcement is review only.
-
-An issue carries the organisation `Type` field: `Bug`, `Feature`, `Enabling Refactoring` or
-`Other`. Organisation members see it in the issue sidebar.
-
-GitHub gives a pull request no issue field, and a project board cannot use an issue field as a
-column. The `Development Board`, which already receives every pull request, therefore carries two
-single-select fields of its own: `Review type`, with the four type names, and `Review priority`,
-with the four urgency names. They are named for the review because that is what they order, and
-they apply to pull requests and to the issues that wait for review.
-
-A board view whose columns are `Review type`, sorted by `Review priority`, shows what to review
-first. Moving a card between columns sets its type. The `Status` field keeps its existing meaning,
-so a card also records how far it is through review.
-
-The board is private, so it serves maintainers only. This ADR and the pull request template
-remain the statement of priorities that contributors can read.
-
 ## Alternatives considered
 
 Reviewing every well-written pull request on its merits alone was rejected. It ignores the
--- a/adr/0014-priorities.md
+++ b/adr/0014-priorities.md
@@ -1,0 +1,148 @@
+# Prioritise bug fixes, features, and only then enabling refactoring
+
+## Context
+
+Maintainers review Infection in their own time. Review capacity, not the number of PRs, sets how
+much the project can absorb, and every pull request spends part of that capacity.
+
+A large number of pull requests neither fix a reported bug nor add a requested feature. They
+often restructure code that already works, or add tooling that no current task needs.
+Such a change is harder to review than a bug fix. A bug fix carries its own purpose: the
+linked issue says what was wrong. A restructuring usually does not, so the reviewer has to work
+out why it is proposed at all, and whether it brings any bug fix or feature closer. That answer
+is frequently missing from the pull request, and asking for it costs more time than reading the
+diff.
+The result is that reported bugs and accepted features wait behind changes that no user asked
+for.
+
+## Decision
+
+We sort review work on two axes: the type of change, and its urgency. They answer different
+questions, and mixing them is what makes a queue unreadable. This increases the review time.
+
+There are four types of change. The type decides whether we accept the work at all, and gives
+the default order in which maintainers review it.
+
+1. **Bug.** Incorrect mutants, wrong MSI, crashes, regressions, data loss, and performance
+   regressions. Compatibility fixes for supported PHP, PHPUnit and test-framework versions
+   belong here.
+2. **Feature.** User-facing capability that an issue describes and the maintainers have accepted.
+3. **Enabling Refactoring.** A structural change that **is a precondition for a specific fix or
+   feature that is impractical to implement today**. The pull request must name and link that
+   fix or feature.
+4. **Other.** Everything else: developer tooling, style sweeps, renames, reorganisation, new
+   static-analysis rules, and comparable work. We accept these only when they do not compete
+   with work of types 1 to 3.
+
+Urgency is the second axis, and it varies inside every type. Two bug fixes are not equally
+urgent: a crash or a wrong MSI in a common configuration comes before a rare edge case that has
+a workaround. A maintainer may therefore review a very urgent item ahead of a whole type, and
+say in the pull request why it goes first.
+
+Urgency cannot promote a change into work that we want. It reorders the queue; it does not turn
+a type 4 change into required work, and "this cleanup is urgent for me" is not urgency.
+
+The two axes together give the review order:
+
+|                      | Urgent  | High    | Medium  | Low     |
+|----------------------|---------|---------|---------|---------|
+| Bug                  | Now     | Next    | Queued  | Queued  |
+| Feature              | Next    | Queued  | Queued  | Backlog |
+| Enabling Refactoring | Queued  | Queued  | Backlog | Backlog |
+| Other                | Backlog | Backlog | Backlog | Backlog |
+
+- **Now** - we interrupt other work, and the fix goes out as soon as possible.
+- **Next** - it goes to the top of its type, ahead of everything queued.
+- **Queued** - normal order inside its type.
+- **Backlog** - we review it when nothing above it competes for the same hours.
+
+Two properties of this table are deliberate. The `Bug` row never falls below `Queued`, because
+every reported bug earns a review even when it is minor. The `Other` row is flat, because urgency
+reorders work but does not make a type 4 change into work that the project needs.
+
+Every pull request states its type, and the author is the one who states it. Type 3 also names
+and links the issue that the refactoring unblocks, because the type alone does not tell the
+reviewer what the change buys. A refactoring that cannot name that issue is type 4, and we
+normally decline or postpone it - it is not a priority for maintainers' review. Urgency is for
+maintainers to judge, not for the author to declare.
+
+A rationale must be verifiable by the reviewer: a reproduced defect, a named capability, a
+measured performance figure, or an existing project convention such as an ADR. Personal taste
+is not a rationale. "I prefer this shape", "this is more idiomatic", and "this is cleaner"
+do not justify changing code that has worked for years. This applies to maintainers as
+much as to first-time contributors.
+
+Infection is not a playground. Its users are other maintainers who trust its results, so a
+change to working code answers to their needs, not to our curiosity about a technique. Try
+ideas out in a fork, and bring back the ones that a bug or a feature needs.
+
+## Consequences
+
+- Reported bugs and accepted features get reviewed first, and users get fixes sooner.
+- Contributors learn early which type their work falls in, and lose less effort on changes
+  that we will not merge.
+- Refactoring keeps a place in the project, but must earn it by unblocking user-visible work.
+- Maintainers may close a type 4 pull request without a detailed review, and must say why.
+  Closing judges the type of change, not the contributor or the quality of the code.
+- An urgent bug fix can go out before a queue of older bug fixes, because the recorded urgency
+  puts it there.
+- Some genuinely good cleanup will not be merged, or will wait a long time. We accept that
+  cost, because the alternative spends review capacity that bug fixes need.
+
+## Enforcement
+
+Review only. Nothing here is automated, because no tool can decide whether a change is required.
+
+The author states the type. A maintainer confirms it before an in-depth review, and corrects it
+when the author has it wrong, which is expected for `Enabling Refactoring`. Urgency is set by a
+maintainer.
+
+Both values are recorded where every maintainer can see them, so that the order in the table is
+visible instead of remembered, and so that the queue can be read in one place. The record covers
+issues as well as pull requests, because a proposal that waits for a decision, an RFC for
+example, spends the same review capacity as a diff.
+
+### Mapping onto GitHub
+
+This subsection describes the current mechanism only. It can change without a new ADR, provided
+that the two axes stay recorded and readable.
+
+A pull request states its type in the `Type of change` section of the pull request template.
+GitHub cannot validate a template, which is why enforcement is review only.
+
+An issue carries the organisation `Type` field: `Bug`, `Feature`, `Enabling Refactoring` or
+`Other`. Organisation members see it in the issue sidebar.
+
+GitHub gives a pull request no issue field, and a project board cannot use an issue field as a
+column. The `Development Board`, which already receives every pull request, therefore carries two
+single-select fields of its own: `Review type`, with the four type names, and `Review priority`,
+with the four urgency names. They are named for the review because that is what they order, and
+they apply to pull requests and to the issues that wait for review.
+
+A board view whose columns are `Review type`, sorted by `Review priority`, shows what to review
+first. Moving a card between columns sets its type. The `Status` field keeps its existing meaning,
+so a card also records how far it is through review.
+
+The board is private, so it serves maintainers only. This ADR and the pull request template
+remain the statement of priorities that contributors can read.
+
+## Alternatives considered
+
+Reviewing every well-written pull request on its merits alone was rejected. It ignores the
+constraint that causes the problem, which is finite review capacity, and in practice it lets
+optional work displace required work.
+
+Refusing all refactoring was rejected. Parts of the codebase do block new features, and
+forbidding structural change would push contributors into workarounds that cost more later.
+
+Triaging the queue instead of publishing a priority order was rejected, though not because
+triage is wrong: the Enforcement section is triage. Triage alone is not enough because it acts
+only after the work exists. The contributor has already spent their time, a maintainer still
+has to open every pull request to sort it, and the sorting is itself review work. Without a
+written criterion, each triage decision also reads as the preference of whoever performed it,
+which is the disagreement this ADR exists to prevent. The priority order gives triage a
+criterion, and gives contributors that criterion before they open an editor.
+
+## Status
+
+Proposed.
--- a/adr/0014-priorities.md
+++ b/adr/0014-priorities.md
@@ -59,6 +59,20 @@ Two properties of this table are deliberate. The `Bug` row never falls below `Qu
 every reported bug earns a review even when it is minor. The `Other` row is flat, because urgency
 reorders work but does not make a type 4 change into work that the project needs.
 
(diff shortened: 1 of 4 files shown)


adr/0014-priorities.md
```diff
@@ -0,0 +1,148 @@
...
+The author states the type. A maintainer confirms it before an in-depth review, and corrects it
+when the author has it wrong, which is expected for `Enabling Refactoring`. Urgency is set by a
+maintainer.
+
+Both values are recorded where every maintainer can see them, so that the order in the table is
+visible instead of remembered, and so that the queue can be read in one place. The record covers
+issues as well as pull requests, because a proposal that waits for a decision, an RFC for
+example, spends the same review capacity as a diff.
+
+### Mapping onto GitHub
```
maks-rafalko: actually, this is what I started with before writing ADR - I wanted to have some board where we can prioritize "what to look at next" from review point of view. To have some order between ~100 waiting-for-review issues/PRs

It turns out, this is not easy. Issues on GH has native Type and Priority fields. But PRs don't. At the same time, it's possible to create our custom single-value dropdowns fields and use them for both Issues and PRs - that's yet to be decided.

I think this section can be removed from this ADR, but leaving it here just to show you what I started with.

adr/0014-priorities.md
```diff
@@ -0,0 +1,148 @@
+# Prioritise bug fixes, features, and only then enabling refactoring
+
+## Context
+
+Maintainers review Infection in their own time. Review capacity, not the number of PRs, sets how
+much the project can absorb, and every pull request spends part of that capacity.
+
+A large number of pull requests neither fix a reported bug nor add a requested feature. They
+often restructure code that already works, or add tooling that no current task needs.
```
theofidry:
```suggestion
often restructure code, or add tooling that no current task needs.
```

(restructuring code that does not work does not make any sense, or I don't know what that would mean)

adr/0014-priorities.md
```diff
@@ -0,0 +1,148 @@
...
+- **Queued** - normal order inside its type.
+- **Backlog** - we review it when nothing above it competes for the same hours.
+
+Two properties of this table are deliberate. The `Bug` row never falls below `Queued`, because
+every reported bug earns a review even when it is minor. The `Other` row is flat, because urgency
+reorders work but does not make a type 4 change into work that the project needs.
+
+Every pull request states its type, and the author is the one who states it. Type 3 also names
+and links the issue that the refactoring unblocks, because the type alone does not tell the
+reviewer what the change buys. A refactoring that cannot name that issue is type 4, and we
```
theofidry:
```suggestion
reviewer what the change brings. A refactoring that cannot name that issue is type 4, and we
```
?

adr/0014-priorities.md
```diff
@@ -0,0 +1,148 @@
...
+normally decline or postpone it - it is not a priority for maintainers' review. Urgency is for
```
theofidry:
```suggestion
normally decline or postpone it because it is not a priority for maintainers' review. Urgency is for
```
?

adr/0014-priorities.md
```diff
@@ -0,0 +1,148 @@
...
+maintainers to judge, not for the author to declare.
+
+A rationale must be verifiable by the reviewer: a reproduced defect, a named capability, a
+measured performance figure, or an existing project convention such as an ADR. Personal taste
```
theofidry, line 80: whilst I get the sentiment, I do have one point to raise there.

This is the cost of the readability of the code. Code ages, practices evolves, tooling evolves, quality of the reviews differs. This can result in a situation where the code is perfectly working, it does its job, but a change may require its understanding, and understanding it may be overly complicated.

I am not saying one should look for a piece of code that may be a bit obfuscated to change it for the sake of it, but equally if an obfuscated piece of code can be made much simpler (for various reasons – e.g. a new tool/dependency available making easier), there is value (IMO) in transferring this cost of learning into a change that makes it easier for new readers.

Obviously this is prone to opinions, what is readable is very often quite subjective or not in the interest of the maintainers and main contributors, or the gains are too marginal to justify it. But I felt the need to counter a bit the "it just makes it cleaner so there is no value in it".

adr/0014-priorities.md
theofidry: >It turns out, this is not easy. Issues on GH has native Type and Priority fields. But PRs don't.

Yes... It's a policy they have on a number of things, it's the same with issue types and completion progress among other things...

I don't think it's worth to add this section to this ADR as long as it's a work in progress to figure out a solution that works especially if trying it out or enforcing it is not something visible from the code (i.e. it's easy to update a rule without updating this ADR, making it outdated).

That said, I'd be happy to see any attempt to improve the Development Board. I feel lonely using it, so I barely do 😅

adr/0014-priorities.md
```diff
@@ -0,0 +1,148 @@
...
+is not a rationale. "I prefer this shape", "this is more idiomatic", and "this is cleaner"
+do not justify changing code that has worked for years. This applies to maintainers as
+much as to first-time contributors.
+
+Infection is not a playground. Its users are other maintainers who trust its results, so a
+change to working code answers to their needs, not to our curiosity about a technique. Try
+ideas out in a fork, and bring back the ones that a bug or a feature needs.
+
+## Consequences
```
theofidry, line 97: An aspect missing here (in this ADR, not sure what section exactly) IMO is the level of importance some changes. I am thinking of the recent changes done in regards to the AutoReview for example.

Indeed, the latter were about improving the existing test harnesses we have and enforcing more conventions, either explicitly stated via an ADR, or established through existing code (part of that work was sometimes adding an ADR for it).

Whilst AI shifts the load and is part of the discussion, in absolute it's not the important factor either. The point is more: there may be a lot of work there, but it has a very limited impact (maintainers and contributors), with well established boundaries (PHPStan violations or failing PHPUnit tests), and the implementation is, ultimately, not important – within reason.* This sort of work does not deserve the same level of attention than the rest, but by its nature if the assumption is that we are always under capacity, then such work can never be done, which I do not think is the correct thing to do either.

Another example is the Docker Sandbox: I experienced it, it was game changer for me personally, but if I had to wait for review on it – especially at a time where I had little experience with it and would have trouble to justify any priotization of it over anything else, then by this ADR this would have never been done. It has since then proved extremely valuable for me and if it was not already in Infection that would 100% be the very first thing I would be doing, and if that was not committed I would have to create another repo/script synchronized somewhere to be able to use it because it is that much of a necessity IMO. And if someone else is trying it and convinced this is absolutely useless and should be removed, it can be discussed and either kept or removed (after the fact).

Those are two examples, but the essence is: depending of the impact and who is affected, I feel the current document misses a bit of nuance in regards to when some changes can land and how much consideration/careful attention/review they need.

*: Obviously no one wants a giant pile of convoluted obfuscated gibberish to maintain.

theofidry (approved):
It's a bit late, so the feedback is a bit more verbose and less well thought than I would like.

LGTM overall, added my 2c on one thing or two but those are generally nitpicks in case you find a good way to express them in the document, not blockers.

adr/0014-priorities.md
```diff
@@ -0,0 +1,148 @@
+# Prioritise bug fixes, features, and only then enabling refactoring
+
+## Context
+
+Maintainers review Infection in their own time. Review capacity, not the number of PRs, sets how
+much the project can absorb, and every pull request spends part of that capacity.
+
+A large number of pull requests neither fix a reported bug nor add a requested feature. They
+often restructure code that already works, or add tooling that no current task needs.
+Such a change is harder to review than a bug fix. A bug fix carries its own purpose: the
```
sanmai:
```suggestion
Such a change is harder to review than a bug fix. A bug fix has a purpose: the
```

adr/0014-priorities.md
```diff
@@ -0,0 +1,148 @@
...
+linked issue says what was wrong. A restructuring usually does not, so the reviewer has to work
```
sanmai:
```suggestion
linked issue identifies a problem. A restructuring usually does not, so the reviewer has to work
```

adr/0014-priorities.md
```diff
@@ -0,0 +1,148 @@
...
+out why it is proposed at all, and whether it brings any bug fix or feature closer. That answer
+is frequently missing from the pull request, and asking for it costs more time than reading the
+diff.
```
sanmai:
```suggestion
is frequently missing from the pull request, and the answer to this question is more important than the diff.
```

sanmai (approved):
I feel like the overall point of this ADR is that we have to slow down. And I agree with that. It is too easy these days to make things, too easy to get overwhelmed by this avalanche as a reader, and often the root reason to do things is lagging behind.

I started to edit out AI speak, but it will probably take more time and even better to do it in a proper editor.

theofidry:
Slept a bit more on this and I would like to add one more piece to this.

There is two aspect to this ADR / issue description: the prioritisation and the pace.

I do not have a definite solution for the latter, I'm happy if us trying anything really. Currently there is no way to express any level of importance, so any PR appears to have the same weight or level of importance than any other, unless someone goes around explicitly pinging about the importance of a specific PR via other channels. So the GitHub board can be a good start.

About the pace... I am conflicted. I am also an OSS maintainers not just for infection, there is times in life where I was too overwhelmed to pay much attention to OSS, there is times where i have a crazy amount of notifications too, I am all too familiar with the situation too.

However, I don't think this ADR will change anything about the pace. Prioritisation will help any of us to know what to review first, but if the idea is, instead of a 100 issues/PRs/month you will have 2-3, I do not think this will happen. For me personally my roadmap is https://github.com/infection/infection/pull/2272 and the workload just reflects the time I have and the level of motivation – not AI (it has been an enabler, but to be transparent, except for a few unimportant pieces, I used it a lot more to articulate changes and thoughts, to better explain the why and motivation factor than outputting raw code – i.e. better explain the rational of the changes). But even if you take me out of the equation, before I "resumed" any activity, the project had 145+ issues and 30+ PRs. There will never be a ground zero, not do I believe it is a healthy objective to achieve.

What I'm trying to say is, when you log in to check your notifications, as long as there is work to do (which I believe will always be the case) and contributions, there will always be notifications. As long as you as a reviewer has a smaller capacity than the contributions (and in OSS it's almost always the case), that is inevitable.

To be clear I'm not throwing this to reject this proposal and much less the sentiment. I think wanting to prioritize to know what is important to check first is good, but expect to have nothing to check no.

sanmai:
My apologies. I used a specific language and I did not think thoroughly how it would look like from your perspective. It looks harsh, and that was never an intent to shut you down, or anyone. I think now I need to sleep on it too.

theofidry:
@sanmai you did not – sorry if my reply sounded like that. I am neither taking offence nor inferring ill or malicious-intents here no worries, no need to walk on eggshells.

I have been a bit more communicative to @maks-rafalko about my usage of AI but not as much to you so I thought it was worth a clarification.

adr/0014-priorities.md
maks-rafalko: yeah, complex topic :) I got your point. But is there anything in particular you want to rephrase here?

adr/0014-priorities.md
maks-rafalko: makes sense. Removed

adr/0014-priorities.md
maks-rafalko: while I agree that some kind of changes requires less attention (but it's hard to achieve this point), there are some statements that I disagree with:

> but if I had to wait for review on it – especially at a time where I had little experience with it and would have trouble to justify any priotization of it over anything else, then by this ADR this would have never been done.

yea, this would have never been done and this is fine IMO. If one want to play with something, expecially with no experience - do it without committing. If it's valuable - then there shouldn't be any problem to justify and introduce it to the codebase.

> and if that was not committed I would have to create another repo/script synchronized somewhere to be able to use it because it is that much of a necessity IMO

didn't get this. What's the problem to add all the files to a folder that is git-ignored?

----

> Those are two examples, but the essence is: depending of the impact and who is affected, I feel the current document misses a bit of nuance in regards to when some changes can land and how much consideration/careful attention/review they need.

would you mind suggest changes by `diff` feature or just by putting some text that you think should be added/updated?

maks-rafalko:
> Currently there is no way to express any level of importance, so any PR appears to have the same weight or level of importance than any other, unless someone goes around explicitly pinging about the importance of a specific PR via other channels. So the GitHub board can be a good start.

that's what this ADR describes by introducing Urgency/Priority field, and my idea was to use Boards where we can see tickets/PRs in different columns (for example) based on this custom fields, so for example I can quickly see what is more important to be reviewed. Example: Test Framework RFC/PRs or PHPStan out-of-memory bugfix would get `Priority: Urgent`. While adding `PHPAt` rule would get `Priority: Medium or Low`. And both author and reviewer understand this: as soon as we have Urgent items to work on - medium and low are deprioritized.

> There is two aspect to this ADR / issue description: the prioritisation and the pace.

This ADR is about prioritizing (hence its name - "ADR: priorities of Infection"), not about the pace. However I would say those have some kind of correlation, but that's another story.

Let me put it in another way: having `N` PRs a day for bugfixes, features or the things that move bugfixes/features forward is ok. Having `N` PRs a day for the things that are not or much less important (see ADR) - is not ok. Regardless of how big the `N` is.

So I do expect that this ADR will help us prioritizing more important things. Implicitly, this should reduce the pace for less important things (which is good), but may increase the pace for more important things (which is good).

> But even if you take me out of the equation, before I "resumed" any activity, the project had 145+ issues and 30+ PRs. There will never be a ground zero, not do I believe it is a healthy objective to achieve.

Not sure how we got to this point, achieving "zero" is not the goal at all. But just to reiterate again: having the number of open PRs and issues is ok. Having the increasing number of _less_ important things when we have other _more_ important things - is not ok. That should be solved by prioritization of what to work on.

> I think wanting to prioritize to know what is important to check first is good, but expect to have nothing to check no.

Absolutely. If you think that some sentence of this ADR says we need to have "nothing to check" or "zero" PRs/issues - let's update it. This is not the goal of this ADR.

adr/0014-priorities.md
```diff
@@ -0,0 +1,123 @@
...
+reviewer what the change brings. A refactoring that cannot name that issue is type 4, and we
+normally decline or postpone it because it is not a priority for maintainers' review. Urgency is for
+maintainers to judge, not for the author to declare.
+
+A rationale must be verifiable by the reviewer: a reproduced defect, a named capability, a
+measured performance figure, or an existing project convention such as an ADR. Personal taste
+is not a rationale. "I prefer this shape", "this is more idiomatic", and "this is cleaner"
+do not justify changing code that has worked for years. This applies to maintainers as
+much as to first-time contributors.
+
```
theofidry, line 84: Follow up of https://github.com/infection/infection/pull/3425#discussion_r3668601552.

```suggestion

Readability work is not inherently without value. Difficult code imposes a recurring
comprehension cost on every future change and a refactoring can transfer that repeated cost
into a single reviewed change. Because readability is subjective, the proposal must make that
cost verifiable, for example through repeated misunderstandings, defects caused by the current
structure, disproportionate effort in recent changes or a substantial and demonstrable
reduction in complexity. Without that evidence, or a specific bug or feature that the
refactoring enables, it remains type 4 work and retains the corresponding priority.

```

adr/0014-priorities.md
theofidry: >What's the problem to add all the files to a folder that is git-ignored?

As soon as you change of machine this would be a problem, which is the case for me at least. I regularly have to juggle between 1-3 machines.

Edit: kinda get why it appears here but at the same time quite weird...

adr/0014-priorities.md
```diff
@@ -0,0 +1,123 @@
...
+- **Now** - we interrupt other work, and the fix goes out as soon as possible.
+- **Next** - it goes to the top of its type, ahead of everything queued.
+- **Queued** - normal order inside its type.
+- **Backlog** - we review it when nothing above it competes for the same hours.
+
+Two properties of this table are deliberate. The `Bug` row never falls below `Queued`, because
+every reported bug earns a review even when it is minor. The `Other` row is flat, because urgency
+reorders work but does not make a type 4 change into work that the project needs.
+
```
theofidry, line 58: Follow up of https://github.com/infection/infection/pull/3425#discussion_r3668732632

```suggestion

Priority determines which work receives attention first. It does not prescribe the same review
depth for every change. Review effort is proportional to the change's exposure: who it affects,
whether its correctness has objective boundaries, how easily it can be reverted, and who will
maintain it. A small change to mutation generation may need substantial review because an error
can silently affect users. A change confined to contributor tooling, guarded by PHPUnit or
static-analysis failures, or otherwise isolated from shipped behaviour may need much less.

Low-exposure work may therefore land opportunistically, including type 4 work, when its review
does not materially delay higher-priority work. Review capacity is not fully interchangeable:
work performed and maintained by someone with a particular need or area of expertise does not
necessarily displace a bug fix that requires another maintainer. "Does not compete" refers to
the implementation and review attention actually needed by higher-priority work, not merely to
the presence of such work in the queue.

```

adr/0014-priorities.md
theofidry: Suggestion: https://github.com/infection/infection/pull/3425#discussion_r3672276025

adr/0014-priorities.md
theofidry: >https://github.com/infection/infection/pull/3425#discussion_r3672331540

Done in https://github.com/infection/infection/pull/3425#discussion_r3672331540

adr/0014-priorities.md
maks-rafalko: applied

adr/0014-priorities.md
maks-rafalko: applied 1 paragraph below

maks-rafalko:
is there anything we would like to update here @sanmai @theofidry ?

adr/0014-priorities.md
```diff
@@ -0,0 +1,145 @@
...
+   belong here.
+2. **Feature.** User-facing capability that an issue describes and the maintainers have accepted.
+3. **Enabling Refactoring.** A structural change that **is a precondition for a specific fix or
+   feature that is impractical to implement today**. The pull request must name and link that
+   fix or feature.
+4. **Other.** Everything else: developer tooling, style sweeps, renames, reorganisation, new
+   static-analysis rules, and comparable work. We accept these only when they do not compete
+   with work of types 1 to 3.
+
+Urgency is the second axis, and it varies inside every type. Two bug fixes are not equally
```
sanmai:
```suggestion
Urgency is the second independent axis. Two bug fixes are not equally
```
I still want to edit this heavy to remove ai-isms.

adr/0014-priorities.md
```diff
@@ -0,0 +1,145 @@
...
+urgent: a crash or a wrong MSI in a common configuration comes before a rare edge case that has
+a workaround. A maintainer may therefore review a very urgent item ahead of a whole type, and
+say in the pull request why it goes first.
+
+Urgency cannot promote a change into work that we want. It reorders the queue; it does not turn
+a type 4 change into required work, and "this cleanup is urgent for me" is not urgency.
```
sanmai: I feel like this paragraphs just restates the obvious; so that's another editing angle.

sanmai (commented):
There's a third axis to measure any piece of work: complexity. So if a refactor is trivial (easy to grasp and review) it should not be grouped with another enabling refactor the adds fifteen different new classes.

adr/0014-priorities.md
maks-rafalko: applied

adr/0014-priorities.md
maks-rafalko: ok, removed

maks-rafalko:
> There's a third axis to measure any piece of work: complexity. So if a refactor is trivial (easy to grasp and review) it should not be grouped with another enabling refactor the adds fifteen different new classes.

Agree. But, unfortunately, there are many other axises here

- consistency
- "the gold community standards"
- security updates, or just packages upgrades
- new PHP and libs features / versions / etc. that can be said "easy to add, let's make it"

I would stick though to the minimum, as it is now - just to show the main goal: we work on bugs and features first, and we prioritize the things by urgency ([already doing it](https://github.com/orgs/infection/projects/2/views/3)). Everything else is details and I don't think they should be explained here, as this is impossible to cover all of them. But feel free to disagree.

sanmai:
I want to definitely disagree about complexity. Even if a PR adds a feature *and* also adds twenty new classes (Codex is the usual suspect) - it should not have the same treatment as other more gentle and robust work.

maks-rafalko:
> I want to definitely disagree about complexity. Even if a PR adds a feature _and_ also adds twenty new classes (Codex is the usual suspect) - it should not have the same treatment as other more gentle and robust work.

do you have a concrete suggestion on how we can update/improve the ADR from this angle?

sanmai:
Yes, let me get my thoughts together.

theofidry:
Relevant to the discussion, some stats about the volume of PRs.

I want to add a big disclaimer: this is mostly done by deriving categorisation via the PR title, description and labels and a PR can have multiple labels. Are excluded PRs that are: closed (not merged), drafts, open and non-categorised PRs. Inevitably most non-categorised PRs tend to be old ones, but I don't really want to touch them all just for the sake of it.

Refactoring in that context is "Enabling refactoring".

[pull-requests-by-category-over-time.html](https://github.com/user-attachments/files/31300293/pull-requests-by-category-over-time.html)

PR #3433, opened 2026-08-02
feat(test-framework): Add debug test framework and static analysis adapters

theofidry:
## Context

Understanding the environment in which Infection orchestrates test frameworks and static analysers is not straightforward. A recent memory-limit issue (see https://github.com/infection/infection/pull/3358) was difficult both to diagnose and to capture in a regression test. The fix is not final and is likely to change again soon, while the test is fragile and specific to one framework. This prompted an investigation into a more resilient way to test this behaviour.

## Description

This PR adds built-in debug adapters for the test-framework and static-analysis process pipelines. The adapters record the commands and runtime state observed by initial and mutant processes, allowing process orchestration to be tested without relying on PHPUnit or static analyser output.

Diagnostics are appended as JSON [JSON Lines][JSONL] to a configurable `debug.logFile`, with a default path when none is configured. This captures the execution state of initial tests and mutant processes.

## Changes

- Registers `debug` as a test framework and static analysis tool, including schema and configuration support for the optional log path. These options are also available to users, although no specific user-facing scenario is intended. Exposing them keeps the implementation simple and is not expected to cause harm.
- Adds debug adapters and command construction for initial, mutant, and static-analysis processes.
- Records runtime state from adapter processes, including command arguments, the memory limit, PHP ini state, and relevant environment variables. This provides insight into how a test framework or static analyser may behave when using XdebugHandler.
- Adds an end-to-end fixture with subset-based JSON expectations* for Xdebug, PCOV, PCOV alongside an Xdebug restart, and phpdbg environments. Each environment has distinct expectations because the coverage driver is part of the expected and tested state.

*: we match the declared JSON against the actual one, but it is not a exact match, i.e. the expectations defines what needs to be there and matched, but the actual JSON can contain more values. This is to allow to keep a rich amount of information in the diagnosis without making testing impossible due to the dynamic nature of some values (e.g. the Xdebug version).

## Related issues

- Prompted by https://github.com/infection/infection/pull/3358.
- Blocker for https://github.com/infection/infection/issues/2863.

[JSONL]: https://jsonlines.org/


resources/schema.json
```diff
@@ -165,6 +165,17 @@
                 }
             }
         },
+        "debug": {
```
theofidry, line 168: not too happy about the name tbh but it's in line with the other test frameworks

theofidry:
I still need to iron out the results, but the gist of it is there.

resources/debug-runtime.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+parse_str(implode('&', array_slice($_SERVER['argv'], 1)), $options);
```
sanmai: `getopt` might be a more appropriate way to handle this

resources/debug-runtime.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
```
sanmai, line 3: Would be nice to have a doc comment here explaining what this does

sanmai (approved):
Haven't tried to use it but looks sane overall.

PR #3441, opened 2026-08-06
fix(reporter): Escape GitHub annotation workflow commands

theofidry:
## Description

Infection reports escaped mutants as GitHub Actions annotations by writing workflow commands to standard output:

```text
::warning file=<file path>,line=<line>::<message>
```

`GitHubAnnotationsReporter` previously interpolated the relative file path without escaping it, while escaping only line feeds in the message. GitHub defines different escaping rules for these contexts. Command data must escape `%`, carriage returns, and line feeds. Command properties must also escape `:` and `,`, which delimit the command and its properties. Without this encoding, a specially crafted path can alter an annotation's structure. If a raw line break reaches the reporter unchanged, it can also terminate the annotation and cause the following text to resemble another workflow command:

```text
src/example.php
::error file=README.md,line=1,title=Injected::Attacker-controlled annotation.php
```

This PR encodes the file path as command-property data and the annotation message as command data. The implementation references GitHub's workflow-command documentation and a pinned version of the official Actions toolkit, which defines the exact character sets.

## Potential abuse

If an attacker can inject a complete workflow command through a raw line break, the most disruptive candidate is `stop-commands`:

```text
::stop-commands::ATTACKER_TOKEN
```

The runner stops interpreting subsequent workflow commands until it encounters the matching resume marker:

```text
::ATTACKER_TOKEN::
```

This could prevent the creation of later Infection annotations or annotations from other tools in the same step. The underlying output would remain in the raw job log, so this would suppress structured annotations without deleting the evidence.

The same precondition could permit an `add-mask` command:

```text
::add-mask::attacker-chosen text
```

GitHub replaces subsequent occurrences of the selected text in the job log with `***`. This cannot reveal secrets, but it could obscure filenames, error messages, test names, or other diagnostic text.

Annotation and presentation commands could also fabricate or distort review feedback:

```text
::error file=trusted.php,line=10,title=Security failure::Fabricated finding
::warning file=src/Foo.php,line=20::Misleading warning
::notice::Fake success or review instruction
::group::Misleading section
::endgroup::
```

These are consequences of a successful workflow-command injection rather than independently reproduced exploit paths. They could attribute findings to innocent code, impersonate another analyser, or flood the checks interface. An `error` annotation does not itself cause a step to fail.

## Security impact

The confirmed defect allows reserved characters in repository-controlled values to alter the grammar of a generated annotation. The plausible impact is therefore a loss of integrity in CI reporting through malformed or false annotations, masked diagnostic output, or suppressed later annotations. No route from this defect to secret disclosure, repository writes, arbitrary command execution, or a changed job result has been demonstrated.

Historically, stdout commands such as `set-env` and `add-path` presented a greater risk because they could influence later process execution. GitHub disabled those commands and replaced them with environment files. The old `set-output` stdout command is also deprecated, so these are not treated as viable impacts on a current GitHub-hosted runner.

For these reasons, this is treated as a security-hardening fix rather than grounds for requesting a CVE. This assessment should be revisited if an affected downstream configuration demonstrates that the defect crosses a security boundary or affects a protected decision.

## Changes

Escapes the affected annotation values.

### DIFF
--- a/src/Reporter/GitHubAnnotationsReporter.php
+++ b/src/Reporter/GitHubAnnotationsReporter.php
@@ -36,6 +36,7 @@ declare(strict_types=1);
 namespace Infection\Reporter;
 
 use Infection\Metrics\ResultsCollector;
+use function sprintf;
 use function str_replace;
 use Symfony\Component\Filesystem\Path;
 
@@ -80,10 +81,27 @@ final readonly class GitHubAnnotationsReporter implements LineMutationTestingRes
      */
     private function buildAnnotation(string $filePath, array $error): string
     {
-        // new lines need to be encoded
-        // see https://github.com/actions/starter-workflows/issues/68#issuecomment-581479448
-        $message = str_replace("\n", '%0A', $error['message']);
+        // Data and properties need to be escaped to avoid a file path or error message to hijack
+        // the GitHub annotations by leveraging the annotation delimiters.
+        //
+        // See:
+        // https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#about-workflow-commands
+        // https://github.com/actions/toolkit/blob/193fa46c20fde8b0ed54194bc08b841c78c0776d/packages/core/src/command.ts#L92-L111
+        return sprintf(
+            "::warning file=%s,line=%d::%s\n",
+            self::escapeProperty($filePath),
+            $error['line'],
+            self::escapeData($error['message']),
+        );
+    }
 
-        return "::warning file={$filePath},line={$error['line']}::{$message}\n";
+    private static function escapeData(string $value): string
+    {
+        return str_replace(['%', "\r", "\n"], ['%25', '%0D', '%0A'], $value);
+    }
+
+    private static function escapeProperty(string $value): string
+    {
+        return str_replace(['%', "\r", "\n", ':', ','], ['%25', '%0D', '%0A', '%3A', '%2C'], $value);
     }
 }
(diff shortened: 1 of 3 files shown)


src/Reporter/GitHubAnnotationsReporter.php
```diff
@@ -80,10 +81,27 @@ public function getLines(): array
      */
     private function buildAnnotation(string $filePath, array $error): string
```
sanmai, line 82: consider `buildAnnotation(string $filePath, int $line, string $message)` while we are at it

sanmai (approved):
Sounds like even a simple string such as `sprintf('%s')` was bugged.

src/Reporter/GitHubAnnotationsReporter.php
theofidry: I'll gladly do it, but separately – this would be a pure refactoring and from this PR PoV pure noise.

tests/phpunit/Reporter/GitHubAnnotationsReporterTest.php
```diff
@@ -83,6 +86,37 @@ public static function metricsProvider(): iterable
...
+                self::createMutantExecutionResult(
+                    0,
+                    For_::class,
+                    DetectionStatus::ESCAPED,
+                    'unused',
+                    mutantDiff: "%\r\n::error::Injected",
+                ),
+            ),
+            [
+                "::warning file=foo/bar,line=10::Escaped Mutant for Mutator \"For_\":%0A%0A%25%0D%0A::error::Injected\n",
```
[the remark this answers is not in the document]
theofidry: that comment is wrong:

>The reporter's heredoc message always ends with a newline (which is escaped to %0A),

it is not, as evidently put:

```php
"::warning file=%s,line=%d::%s\n",
```

note `\n` there, not escaped.

src/Reporter/GitHubAnnotationsReporter.php
theofidry: See https://github.com/infection/infection/pull/3442

PR #3447, opened 2026-08-07
fix(phpunit): Use recordTestRunHistory instead of cacheResult on PHPUnit 13.3

### LINKED ISSUE #3445: bug: injected "cacheResult" attribute trips PHPUnit 13.3's deprecation and aborts the initial test run
| Question    | Answer
| Infection version | 0.34.1
| Test Framework version | PHPUnit 13.3.0
| CI | either
| PHP version | 8.5.9
| Platform    | MacOS and Ubuntu
| Github Repo | https://github.com/NexusPHP/mcp

PHPUnit 13.3 deprecated the `cacheResult` XML attribute in favour of `recordTestRunHistory`. Its XSD now carries `<xs:documentation>Deprecated, use recordTestRunHistory instead</xs:documentation>` on `cacheResult` and defines `recordTestRunHistory` alongside it, and `TextUI/Configuration/Xml/Loader.php:811-816` raises a test-runner deprecation whenever the attribute is present.

Infection writes that attribute into the PHPUnit configuration it generates, in two places in `src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php`:

- `deactivateResultCaching()` sets `cacheResult="false"` (line 100)
- `handleResultCacheAndExecutionOrder()` sets `cacheResult="true"` for PHPUnit >= 11.0 (line 107)

For a project whose PHPUnit configuration sets `failOnPhpunitDeprecation="true"`, that deprecation makes PHPUnit exit 1. Infection reads the non-zero exit from the **initial** test run as a failing suite and aborts before mutating anything, so the whole mutation gate is down on PHPUnit 13.3.

The reported diagnosis is misleading, which is what makes this expensive to track down. Infection prints `Project tests must be in a passing state before running Infection.` and suggests the suite has hidden order dependencies, while PHPUnit's own output directly above it reads `OK, but there were issues!` with 5850 passing tests, 0 failures, and one deprecation that the project's code did not cause.

<details>

 ```json5
 {
     "$schema": "./tools/vendor/infection/infection/resources/schema.json",
     "source": {
         "directories": ["src"]
     },
     "timeout": 30,
     "threads": "max",
     "tmpDir": "build",
     "minMsi": 100,
     "minCoveredMsi": 100,
     "testFramework": "phpunit"
 }
 ```
</details>

<details>

 ```xml
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
          bootstrap="vendor/autoload.php"
          cacheDirectory="build/.phpunit.cache"
          executionOrder="random"
          failOnPhpunitDeprecation="true">
     <testsuites>
         <testsuite name="Unit Tests">
             <directory>tests</directory>
         </testsuite>
     </testsuites>
     <source>
         <include><directory>src</directory></include>
     </source>
 </phpunit>
 ```
</details>

 ```shell
 vendor/bin/infection --with-uncovered --static-analysis-tool=phpstan
 ```

<details>

 ```
         Infection runs the test suite in a RANDOM order. Make sure your tests
         do not have hidden dependencies.
         ...
         '/path/to/vendor/bin/phpunit' '--configuration'
         '/path/to/build/infection/phpunitConfiguration.initial.infection.xml'
         '--exclude-source-from-xml-coverage'
         '--coverage-xml=/path/to/build/infection/coverage-xml'
         '--log-junit=/path/to/build/infection/junit.xml'
         PHPUnit reported an exit code of 1.

         STDOUT:
         PHPUnit 13.3.0 by Sebastian Bergmann and contributors.
         Runtime:       PHP 8.5.9 with Xdebug 3.5.3
         Configuration: /path/to/build/infection/phpunitConfiguration.initial.infection.xml
         Random Seed:   1786092846

         Time: 00:31.539, Memory: 191.00 MB

         There was 1 PHPUnit test runner deprecation:

         1) The "cacheResult" attribute is deprecated and will be removed in
         PHPUnit 14. Use "recordTestRunHistory" instead.

         OK, but there were issues!
         Tests: 5850, Assertions: 14894, PHPUnit Deprecations: 1.
 ```
</details>

**Workaround:** `"testFrameworkExtraArgs": "--do-not-fail-on-phpunit-deprecation"` in `infection.json5`. It restores the gate, but it also masks genuine PHPUnit deprecations in the project's own tests for the duration of the Infection run.

**Suggested fix:** version-gate the attribute the way `handleResultCacheAndExecutionOrder()` already gates on 11.0 and 7.3, emitting `recordTestRunHistory` for PHPUnit >= 13.3 and keeping `cacheResult` below it.

lazerg:
## Description

PHPUnit 13.3 deprecated the `cacheResult` attribute in favour of `recordTestRunHistory`. Its config loader reports a test runner deprecation when `cacheResult` is present and `recordTestRunHistory` is not, so on 13.3 every Infection run reports a deprecation the project did not cause. A project with `failOnPhpunitDeprecation="true"` then gets exit code 1 from the initial test run, and Infection aborts with `Project tests must be in a passing state before running Infection.` even though the suite passed.

`recordTestRunHistory` short-circuits that check, so writing it in the generated configurations is enough.

## Changes

- Disable the `defects` PHPUnit test order for 13.3+.
- Disable `recordTestRunHistory` instead of `cacheResult` in 13.3+.

## Checklist

- [x] Tests added/updated
- [x] Appropriate labels applied (e.g. `performance`, `feature`) — I cannot apply them, `Bugfix` and `TestFramework / PHPUnit` look right.

## Related issues

Fixes https://github.com/infection/infection/issues/3445.

### DIFF
--- a/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
+++ b/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
@@ -90,7 +90,7 @@ final readonly class InitialConfigBuilder implements ConfigBuilder
         $this->configManipulator->replaceWithAbsolutePaths($xPath);
         $this->configManipulator->setStopOnFailureOrDefect($version, $xPath);
         $this->configManipulator->deactivateColours($xPath);
-        $this->configManipulator->deactivateResultCaching($xPath);
+        $this->configManipulator->deactivateResultCaching($version, $xPath);
         $this->configManipulator->deactivateStderrRedirection($xPath);
         $this->configManipulator->removeExistingLoggers($xPath);
         $this->configManipulator->removeExistingPrinters($xPath);
--- a/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
+++ b/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
@@ -156,8 +156,7 @@ final readonly class InitialConfigBuilder implements ConfigBuilder
 
     private function addRandomTestsOrderAttributesIfNotSet(string $version, SafeDOMXPath $xPath): void
     {
-        // ordering by defects needs the test run history, which the initial run deactivates. PHPUnit ignored that combination silently until 13.3 turned it into a warning https://github.com/sebastianbergmann/phpunit/blob/13.3.0/src/TextUI/Application.php
-        if (PhpUnitAdapter::supportsExecutionOrderDefectsRandom($version) && version_compare($version, '13.3', '<')) {
+        if (PhpUnitAdapter::supportsExecutionOrderDefectsRandom($version)) {
             if ($this->addAttributeIfNotSet('executionOrder', 'defects,random', $xPath)) {
                 $this->addAttributeIfNotSet('resolveDependencies', 'true', $xPath);
             }
--- a/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
+++ b/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
@@ -156,7 +156,8 @@ final readonly class InitialConfigBuilder implements ConfigBuilder
 
     private function addRandomTestsOrderAttributesIfNotSet(string $version, SafeDOMXPath $xPath): void
     {
-        if (PhpUnitAdapter::supportsExecutionOrderDefectsRandom($version)) {
+        // ordering by defects needs the test run history, which the initial run deactivates. PHPUnit ignored that combination silently until 13.3 turned it into a warning https://github.com/sebastianbergmann/phpunit/blob/13.3.0/src/TextUI/Application.php
+        if (PhpUnitAdapter::supportsExecutionOrderDefectsRandom($version) && version_compare($version, '13.3', '<')) {
             if ($this->addAttributeIfNotSet('executionOrder', 'defects,random', $xPath)) {
                 $this->addAttributeIfNotSet('resolveDependencies', 'true', $xPath);
             }
--- a/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
+++ b/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
@@ -95,13 +95,29 @@ final readonly class XmlConfigurationManipulator
         }
     }
 
-    public function deactivateResultCaching(SafeDOMXPath $xPath): void
+    public function deactivateResultCaching(string $version, SafeDOMXPath $xPath): void
     {
+        // starting from PHPUnit 13.3 cacheResult is deprecated in favour of recordTestRunHistory, and using it makes PHPUnit report a test runner deprecation https://github.com/sebastianbergmann/phpunit/blob/13.3.0/phpunit.xsd
+        if (version_compare($version, '13.3', '>=')) {
+            $this->setAttributeValue($xPath, 'recordTestRunHistory', 'false');
+
+            return;
+        }
+
         $this->setAttributeValue($xPath, 'cacheResult', 'false');
     }
 
     public function handleResultCacheAndExecutionOrder(string $version, SafeDOMXPath $xPath, string $mutationHash, string $tmpDir): void
     {
+        // starting from PHPUnit 13.3 cacheResult is deprecated in favour of recordTestRunHistory https://github.com/sebastianbergmann/phpunit/blob/13.3.0/phpunit.xsd
+        if (version_compare($version, '13.3', '>=')) {
+            $this->setAttributeValue($xPath, 'recordTestRunHistory', 'true');
+            $this->setAttributeValue($xPath, 'cacheDirectory', sprintf('%s/.phpunit.result.cache.%s', $tmpDir, $mutationHash));
+            $this->setAttributeValue($xPath, 'executionOrder', 'defects');
+
+            return;
+        }
+
         // starting from PHPUnit 11.0 the cacheResultFile was removed, we now set cacheDirectory instead https://github.com/sebastianbergmann/phpunit/blob/11.0.0/phpunit.xsd
         if (version_compare($version, '11.0', '>=')) {
             $this->setAttributeValue($xPath, 'cacheResult', 'true');
--- a/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
+++ b/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
@@ -97,7 +97,7 @@ final readonly class XmlConfigurationManipulator
 
     public function deactivateResultCaching(string $version, SafeDOMXPath $xPath): void
     {
-        // starting from PHPUnit 13.3 cacheResult is deprecated in favour of recordTestRunHistory, and using it makes PHPUnit report a test runner deprecation https://github.com/sebastianbergmann/phpunit/blob/13.3.0/phpunit.xsd
+        // PHPUnit 13.3 deprecated cacheResult in favour of recordTestRunHistory
         if (version_compare($version, '13.3', '>=')) {
             $this->setAttributeValue($xPath, 'recordTestRunHistory', 'false');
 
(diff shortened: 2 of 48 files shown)


src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
```diff
@@ -95,13 +95,29 @@ public function removeExistingLoggers(SafeDOMXPath $xPath): void
         }
     }
 
-    public function deactivateResultCaching(SafeDOMXPath $xPath): void
+    public function deactivateResultCaching(string $version, SafeDOMXPath $xPath): void
     {
+        // starting from PHPUnit 13.3 cacheResult is deprecated in favour of recordTestRunHistory, and using it makes PHPUnit report a test runner deprecation https://github.com/sebastianbergmann/phpunit/blob/13.3.0/phpunit.xsd
+        if (version_compare($version, '13.3', '>=')) {
```
theofidry: in this case I think either:


```suggestion
$property = version_compare($version, '13.3', '>=') ? 'recordTestRunHistory' : 'cacheResult';
```

or a if/else would be clearer here.

theofidry (commented):
Thanks for the PR! Could you also add an e2e test for PHPUnit 13 (<13.3) and PHPUnit 13.3+? You can take the existing `PHPUnit_12-0`  and `PHPUnit_12-5` as references

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
```diff
@@ -95,13 +95,29 @@ public function removeExistingLoggers(SafeDOMXPath $xPath): void
         }
     }
 
-    public function deactivateResultCaching(SafeDOMXPath $xPath): void
+    public function deactivateResultCaching(string $version, SafeDOMXPath $xPath): void
     {
+        // starting from PHPUnit 13.3 cacheResult is deprecated in favour of recordTestRunHistory, and using it makes PHPUnit report a test runner deprecation https://github.com/sebastianbergmann/phpunit/blob/13.3.0/phpunit.xsd
```
sanmai:
```suggestion
        // PHPUnit 13.3 deprecated cacheResult in favour of recordTestRunHistory
```

sanmai (approved):
Makes sense 👍

sanmai:
@lazerg Did you use an agent to build this? I wonder how did it work out for you, as we built a very extensive AGENTS.md to help with any kind of work.

lazerg:
> @lazerg Did you use an agent to build this? I wonder how did it work out for you, as we built a very extensive AGENTS.md to help with any kind of work.

Yes, I have an agent that does everything from start to end. 
I just test and review. But in 90% of cases, it’s not even needed.

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
lazerg: Done, thanks.

theofidry:
It seems it missed my comment about the e2e tests though :)

lazerg:
Sorry, I did miss it. Both are in now: `tests/e2e/PHPUnit_13-2` pins PHPUnit below 13.3, `tests/e2e/PHPUnit_13-3` pins 13.3+. They copy `PHPUnit_12-5` and add `failOnPhpunitDeprecation="true"` to `phpunit.xml`, so a deprecation fails the run instead of passing quietly. Both skip below PHP 8.4.1, which is what PHPUnit 13 needs.

The 13.3 one found a second problem. The initial config sets `executionOrder="defects,random"`, and 13.3 warns if you ask for defects ordering while the test run history is off. Infection turns that warning into a failed initial run. Older PHPUnit accepted the same combination without a word. On 13.3+ the initial config now asks for `random` only.

Both e2e tests give the same output as `PHPUnit_12-5`. On master the 13.3 one fails with the error from #3445.

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
lazerg: I left the early return because `setStopOnFailureOrDefect()` right below has the same shape: one version gate, two attribute names, one value. Happy to switch to the ternary if you prefer it.

src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
```diff
@@ -156,7 +156,8 @@ private function addCoverageNodes(string $version, SafeDOMXPath $xPath): void
 
     private function addRandomTestsOrderAttributesIfNotSet(string $version, SafeDOMXPath $xPath): void
     {
-        if (PhpUnitAdapter::supportsExecutionOrderDefectsRandom($version)) {
+        // ordering by defects needs the test run history, which the initial run deactivates. PHPUnit ignored that combination silently until 13.3 turned it into a warning https://github.com/sebastianbergmann/phpunit/blob/13.3.0/src/TextUI/Application.php
+        if (PhpUnitAdapter::supportsExecutionOrderDefectsRandom($version) && version_compare($version, '13.3', '<')) {
```
theofidry: shouldn't this condition be moved to within `::supportsExecutionOrderDefectsRandom()` or does it has more side-effects?

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
```diff
@@ -95,13 +95,29 @@ public function removeExistingLoggers(SafeDOMXPath $xPath): void
         }
     }
 
-    public function deactivateResultCaching(SafeDOMXPath $xPath): void
+    public function deactivateResultCaching(string $version, SafeDOMXPath $xPath): void
     {
+        // PHPUnit 13.3 deprecated cacheResult in favour of recordTestRunHistory
```
theofidry, line 100: the comment was resolved but the code unchanged

tests/e2e/PHPUnit_13-3/phpunit.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/phpunit.xsd"
+         bootstrap="vendor/autoload.php"
+         colors="true"
+         testdoxSummary="true"
+         cacheDirectory="var/phpunit-cache"
```
theofidry, line 7: shouldn't those contain `cacheResult`/`recordTestRunHistory`?

src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php
lazerg: Moved it in 02964ae0. The only other caller is `getInitialTestsFailRecommendations()` and it wants the same answer: on 13.3 it now suggests `executionOrder="random"`, which is what the initial config actually sets.

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
lazerg: Done in be847021. I used the same ternary in `handleResultCacheAndExecutionOrder()`, where the 13.3 branch was just a copy of the 11.0 one.

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
lazerg: My mistake, I should not have resolved it. Unresolved now, and the ternary is in.

tests/e2e/PHPUnit_13-3/phpunit.xml
lazerg: Infection writes those itself, so the fixture does not need them. `deactivateResultCaching()` overwrites the attribute whatever the source config says, and I checked that adding `cacheResult="false"` to the fixture gives identical output. The gate is already covered: revert the src change and this e2e fails with `Tests cannot be ordered by defects because recording of the test run history is disabled` plus the `cacheResult` deprecation.

PR #3448, opened 2026-08-07
fix!: Fail when PHP configures an auto-prepend file

### LINKED ISSUE #3444: All mutants report as escaped (0% MSI) despite 100% Mutation Code Coverage — confirmed not a test-quality issue
## Summary

Running Infection against small, well-covered, dependency-free PHP classes produces **0% MSI — every single mutant reports as escaped**, including mutations that are logically impossible not to be caught by the existing (passing) test suite. Infection's own coverage analysis correctly reports 100% Mutation Code Coverage for these files, so it correctly identifies that the mutated lines are exercised by tests. The failure appears to be specifically in whether the mutated code is the code that actually executes during each mutant's test run.

I've spent a full diagnostic pass ruling out the usual suspects (see below) and have a decisive reproduction that isolates this to Infection itself rather than the test suite. Details follow.

## Environment

- **Infection**: reproduced identically on both `0.27.11` and `0.34.1` (upgraded mid-investigation via `composer require --dev infection/infection --with-all-dependencies`)
- **PHPUnit**: `10.5.64`
- **PHP**: `8.3.30`
- **OS**: macOS
- **Coverage driver**: Xdebug (`[notice] You are running Infection with Xdebug enabled.` shown on every run — have not yet tried PCOV as an alternative)
- **Composer**: `2.8.6`
- **Xdebug version**: confirmed loaded (`php -m` lists `xdebug`/`Xdebug`); bundled with the LocalWP PHP 8.3.30 build, exact Xdebug version not separately captured
- **macOS version / chip**: `27.0` (build `26A5378n`), Apple Silicon (`darwin-arm64`)
- **PHP install method**: bundled with [Local](https://localwp.com/) (a local WordPress development app), not a standalone system install. Binary path: `.../lightning-services/php-8.3.30+1/bin/darwin-arm64/bin/php`. `PHPRC` is set by the Local app's shell environment to a site-specific `php.ini` directory (`.../Local/run/<id>/conf/php/`) intended for that tool's own PHP-FPM process — this is very likely relevant context even though the specific `auto_prepend_file` directive in it was ruled out above as the direct cause. A Homebrew-installed PHP 8.5.9 is also present on this machine but fails to load extensions correctly due to the same `PHPRC` override pointing it at extension paths built for the LocalWP 8.3 binary.

## Steps to reproduce

```bash
vendor/bin/infection --filter=src/Elementor/LayoutParser.php,src/Elementor/LayoutSerializer.php --threads=4
```

(On 0.34.1, `--filter` is deprecated in favor of a positional argument, but behavior is identical either way.)

**Result:**
```
26 mutations were generated:
       0 mutants were killed
      26 covered mutants were not detected

Mutation Code Coverage: 100%
Covered Code MSI: 0%
```

Every mutant across both files — including trivially-fatal ones — is reported as escaped.

## Decisive reproduction: a mutation that is logically impossible to escape

One of the 26 escaped mutants (`PublicVisibility`) changes:

```diff
-    public function parse(array $nodes): array
+    protected function parse(array $nodes): array
```

This method is called from roughly 10 different test methods, all from outside the class. Changing `public` to `protected` is a guaranteed PHP fatal error (`Error: Call to protected method ... from global scope`) on the very first call — there is no code path under which the existing test suite could fail to catch this. Infection nonetheless reports it as escaped.

To rule out this being a fluke of Infection's own reporting, I applied one of the other escaped mutations **by hand, directly to the source file** (Infection not involved at all):

```diff
-        $elementCount = 0;
+        $elementCount = -1;
```

Then ran plain PHPUnit directly:

```bash
vendor/bin/phpunit
```

**Result: 4 test failures**, exactly as expected, including:
```
Failed asserting that 3 is identical to 4.
```
(from an `assertSame(4, $result['element_count'])` assertion that Infection had reported this exact mutation as escaping).

This confirms the test suite correctly detects the mutated behavior when it actually runs. The problem is that Infection is not causing the mutated code to be the code that executes during its own test invocation for these mutants.

## Things ruled out

| Variable | Test performed | Result |
|---|---|---|
| Threads | `--threads=1` vs `--threads=4` vs `--threads=9` | Identical (0% MSI) |
| PHPUnit result cache | `rm -rf .phpunit.cache` before rerun | Identical |
| OPcache (CLI) | `php -d opcache.enable_cli=0 vendor/bin/infection ...` | Identical |
| Symlinked project path | Project is normally reached via a symlink (`wp-content/plugins/<slug>` → real path — a common WordPress plugin dev layout). Reran from the real, non-symlinked path directly. | Identical |
| Disk / filesystem | Reproduced on an external HDD **and** on a completely fresh `composer install` on the internal disk (`~/Documents/...`, new `vendor/`, nothing shared) | Identical |
| Infection version | `0.27.11` and `0.34.1` | Identical |
| `open_basedir` / `disable_functions` | Confirmed both empty via `php -i` | Not the cause |
| `auto_prepend_file` | The PHP in use is bundled with LocalWP (a local WordPress dev tool), which sets `PHPRC` in the shell environment to a site-specific `php.ini` containing `auto_prepend_file="/Applications/Local.app/Contents/Resources/extraResources/local-bootstrap.php"`. Reproduced with a copy of that same ini with only the `auto_prepend_file` line commented out (`PHPRC=<modified-ini-dir> vendor/bin/infection ...`) | Identical |

## Expected behavior

Mutants that are trivially detectable by the existing suite (like the `PublicVisibility` example) should be killed.

## Actual behavior

100% of generated mutants report as escaped, MSI 0%, regardless of every variable above.

## Additional notes

- Coverage driver is Xdebug throughout this investigation. I haven't yet tried PCOV as an alternative — happy to test that next if it's a likely lead.
- The two files used for the minimal repro (`LayoutParser.php`, `LayoutSerializer.php`) are small, framework-agnostic, pure PHP with no WordPress/external dependencies — I can extract a minimal standalone reproduction repository if that would help, since the project itself isn't public.
- `infection.json5` in use:
  ```json5
  {
      "source": { "directories": ["src"] },
      "mutators": { "@default": true },
      "testFramework": "phpunit"
  }
  ```
- Happy to run any additional diagnostic commands maintainers would find useful (`--debug`, `-vvv`, etc.) — just let me know what to capture.

theofidry:
## Context

When a file is prepended by PHP via `auto_prepend_file`, then it is immediately autoloaded before the Composer autoloader or Infection's interceptor can kick in. Part of the problem is that what the prepended file does is completely arbitrary: it can load other files, the Composer autoloader or other.

This effectively can completely disable the Infection's interceptor, i.e. our ability to mutate the code in the mutant process.

## Description

This PR prevent Infection from reporting misleading escaped mutants when PHP's `auto_prepend_file` setting loads something, and instead fail upfront with a clear failure.

## Changes

- Add a preloaded-source checker and a dedicated `PreloadedSourceFound` domain exception (note that this is not a `RuntimeException` – it is not meant to be caught. A dedicated exception was still chosen because it is easier to understand for users or in a trace rather than a generic exception.
- Add an end-to-end reproduction for the escaped-mutant failure.

## Breaking Changes

Infection now refuses to run when `auto_prepend_file` is non-empty. Such runs previously continued but could execute original code instead of mutants and report a false mutation score.

## Related issues

Closes https://github.com/infection/infection/issues/3444.

### DIFF
--- a/tests/e2e/Preloaded_Source/composer.json
+++ b/tests/e2e/Preloaded_Source/composer.json
@@ -2,9 +2,6 @@
     "require-dev": {
         "phpunit/phpunit": "^12"
     },
-    "conflict": {
-        "phpunit/phpunit": ">=12.4"
-    },
     "autoload": {
         "psr-4": {
             "Infection\\E2ETests\\PreloadedSource\\": "src/"
@@ -16,11 +13,6 @@
         }
     },
     "config": {
-        "sort-packages": true,
-        "audit": {
-            "ignore": {
-                "CVE-2026-24765": "Testing PHPUnit 12.0 compatibility"
-            }
-        }
+        "sort-packages": true
     }
 }
--- a/tests/e2e/Preloaded_Source/composer.json
+++ b/tests/e2e/Preloaded_Source/composer.json
@@ -1,0 +1,26 @@
+{
+    "require-dev": {
+        "phpunit/phpunit": "^12"
+    },
+    "conflict": {
+        "phpunit/phpunit": ">=12.4"
+    },
+    "autoload": {
+        "psr-4": {
+            "Infection\\E2ETests\\PreloadedSource\\": "src/"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "Infection\\E2ETests\\PreloadedSource\\Tests\\": "tests/"
+        }
+    },
+    "config": {
+        "sort-packages": true,
+        "audit": {
+            "ignore": {
+                "CVE-2026-24765": "Testing PHPUnit 12.0 compatibility"
+            }
+        }
+    }
+}
--- a/devTools/mago-baseline.toml
+++ b/devTools/mago-baseline.toml
@@ -975,193 +975,193 @@ count = 1
 [[issues]]
 file = "src/Container/Container.php"
 code = "unhandled-thrown-type"
-message = 'Potentially unhandled exception `Infection\FileSystem\Locator\FileOrDirectoryNotFound` in `{closure:src/Container/Container.php:286:47}`.'
+message = 'Potentially unhandled exception `Infection\FileSystem\Locator\FileOrDirectoryNotFound` in `{closure:src/Container/Container.php:287:47}`.'
 count = 1
 
 [[issues]]
 file = "src/Container/Container.php"
 code = "unhandled-thrown-type"
-message = 'Potentially unhandled exception `Infection\FileSystem\Locator\FileOrDirectoryNotFound` in `{closure:src/Container/Container.php:294:31}`.'
+message = 'Potentially unhandled exception `Infection\FileSystem\Locator\FileOrDirectoryNotFound` in `{closure:src/Container/Container.php:295:31}`.'
 count = 1
 
 [[issues]]
 file = "src/Container/Container.php"
 code = "unhandled-thrown-type"
-message = 'Potentially unhandled exception `Infection\FileSystem\Locator\FileOrDirectoryNotFound` in `{closure:src/Container/Container.php:309:49}`.'
+message = 'Potentially unhandled exception `Infection\FileSystem\Locator\FileOrDirectoryNotFound` in `{closure:src/Container/Container.php:310:49}`.'
 count = 1
 
 [[issues]]
 file = "src/Container/Container.php"
 code = "unhandled-thrown-type"
-message = 'Potentially unhandled exception `Infection\FileSystem\Locator\FileOrDirectoryNotFound` in `{closure:src/Container/Container.php:320:37}`.'
+message = 'Potentially unhandled exception `Infection\FileSystem\Locator\FileOrDirectoryNotFound` in `{closure:src/Container/Container.php:321:37}`.'
 count = 1
 
 [[issues]]
 file = "src/Container/Container.php"
 code = "unhandled-thrown-type"
-message = 'Potentially unhandled exception `Infection\FileSystem\Locator\FileOrDirectoryNotFound` in `{closure:src/Container/Container.php:333:45}`.'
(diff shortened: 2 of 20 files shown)


tests/e2e/Preloaded_Source/composer.json
```diff
@@ -0,0 +1,26 @@
+{
+    "require-dev": {
+        "phpunit/phpunit": "^12"
+    },
+    "conflict": {
+        "phpunit/phpunit": ">=12.4"
```
sanmai: looks like a leftover

sanmai (approved):
`auto_prepend_file` could have some legitimate uses so I fully expect this could blow back at us

theofidry:
> `auto_prepend_file` could have some legitimate uses so I fully expect this could blow back at us

I'm sure there is valid cases, but I do not see a way we can deal with this at the moment.

Maybe we could try to leverage it for our interceptor and it would kick in regardless of what users do in their `auto_prepend_file` but I don't know if that would work and it requires some testing.

maks-rafalko (approved):
~~I wonder how did you find it? I see this ini setting for the first time in my life :)~~

Found the issue 🤦

theofidry:
>Maybe we could try to leverage it for our interceptor and it would kick in regardless of what users do in their auto_prepend_file but I don't know if that would work and it requires some testing.

I quickly experimented with it and it's a no go: if the mutated code has any dependency, since this kicks in before the Composer autoloader is registered, it would result in a failure.

I could imagine a way to make it work, but I feel this would only result in more magic, fragile behaviour and more unexpected side-effects. It is IMO not a viable solution.

So I stand by this patch: it is a real limitation to the users, and I do not claim to invalidate their use case, but it is simply too much of a can of worm to work it out. So either they can work around that for tests, or they will have to do without infection because we have no way to guarantee that it would work (not working could very well be a straight up failure or weird report results).

sanmai:
Yep, the one way out is probably is to make it an option to turn this new behavior off, but we have yet to see someone bump into this and it may never happen.

PR #3449, opened 2026-08-08
refactor(container): make Engine fully autowirable

sanmai:
I noticed that #3448 added another `Engine` dependency. And I traced the dependencies to find that most dependencies are injectable after minor tweaks:

- we replace `runInitialStaticAnalysis` with an injected `InitialStaticAnalysisRunner`
- to make sure it only runs when needed, we add a no-op `NullInitialStaticAnalysisRunner`

I also unit-tested a pair of catch blocks in `executeCommand` to kill some escaping mutations (extracted as #3453)

PR #3450, opened 2026-08-08
refactor(container): move ConsoleOutput instantiation into Container

sanmai:
## Description

Turns out the container already has the dependencies of the `ConsoleOutput` and we can now inject it elsewhere, such as in `MinMsiChecker`.

This opens up the possibility to make the engine fully injectable #3449 and other work to split the engine into components so testing it relies less on E2E and more on unit testing.

### DIFF
--- a/devTools/mago-baseline.toml
+++ b/devTools/mago-baseline.toml
@@ -975,193 +975,193 @@ count = 1
 [[issues]]
 file = "src/Container/Container.php"
 code = "unhandled-thrown-type"
-message = 'Potentially unhandled exception `Infection\FileSystem\Locator\FileOrDirectoryNotFound` in `{closure:src/Container/Container.php:287:47}`.'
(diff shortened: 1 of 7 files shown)


theofidry (approved):
Checking on the phone right now, do I understand correctly that we already pass the logger to the Container hence its value is automatically picked up for ConsoleOutput?

theofidry:
@sanmai I would advise against putting on auto merge when waiting for review. Often (at least me) give approval because the feedback is not a blocker and the general direction is fine but there may still be points worth addressing before merging.

Otherwise I can always just comment and say it's good to me, I suppose

sanmai:
Yes, we pass the logger already.

It is a trivial change so I did give it a shot to auto-merge. I don't think I do this for non-trivial changes, but I'll double check next time. IMO GitHub could have added a note about this somewhere where you approve it.

PR #3455, opened 2026-08-09
sa: update the list of scanned directories for Mago, drop redundant types

sanmai:
## Description

Let's limit the number of issues Mago detects in tests to what is actually worth fixing.

### DIFF
--- a/mago.toml
+++ b/mago.toml
@@ -7,22 +7,25 @@ php-version = "8.3.0"
 workspace = "."
 paths = [
     "src",
-    "tests/autoloaded",
     "tests/Architecture",
     "tests/benchmark",
     "tests/phpunit"
 ]
-includes = ["vendor"]
-excludes = [
+includes = [
+    "vendor",
     "src/FileSystem/DummyFileSystem.php",
+    "tests/autoloaded",
+    "tests/phpunit/Fixtures",
+]
+excludes = [
     "src/CustomMutator/templates",
+    "tests/phpunit/Fixtures/Locator",
     "tests/benchmark/MutationGenerator/sources",
     "tests/benchmark/Tracing/benchmark-source",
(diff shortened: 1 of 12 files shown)


mago.toml
```diff
@@ -7,22 +7,25 @@ php-version = "8.3.0"
...
     "tests/benchmark",
     "tests/phpunit"
 ]
-includes = ["vendor"]
-excludes = [
+includes = [
+    "vendor",
     "src/FileSystem/DummyFileSystem.php",
+    "tests/autoloaded",
+    "tests/phpunit/Fixtures",
```
sanmai, line 18: We need to scan these for symbols (as it reduces the number of detected issues), but we do not need Mago to report on issues there.

mago.toml
```diff
@@ -63,6 +66,7 @@ unchecked-exceptions = [
 ]
 ignore = [
     { code = "unhandled-thrown-type", in = ["tests/"] },
+    { code = "imprecise-type", in = ["tests/"] },
```
sanmai, line 69: Documenting the shape of the return value of all data providers makes no sense. And we have PHPStan ignoring these issues already.

mago.toml
```diff
@@ -7,22 +7,25 @@ php-version = "8.3.0"
 workspace = "."
 paths = [
     "src",
-    "tests/autoloaded",
```
sanmai, line 10: PHPStan never scanned these, so I assume it is the original intent since PHPStan predates Mago.

sanmai: Further, PHP CS Fixer ignores `tests/autoloaded` outright.

maks-rafalko (approved):

mago.toml
theofidry: Ah perfect I had in mind to look into that for a while, thanks for looking into it

PR #3456, opened 2026-08-09
sa: Instruct PHPStan to ignore unresolved iterable types in tests

sanmai:
## Description

Same things as https://github.com/infection/infection/pull/3455 but for PHPStan.

### DIFF
--- a/devTools/phpstan-baseline.neon
+++ b/devTools/phpstan-baseline.neon
@@ -1351,6 +1351,12 @@ parameters:
 			path: ../tests/phpunit/MockedContainer.php
 
 		-
+			rawMessage: 'Method Infection\Tests\MockedContainer::createWithServices() has parameter $values with no value type specified in iterable type array<class-string<object>, mixed>.'
+			identifier: missingType.iterableValue
+			count: 1
+			path: ../tests/phpunit/MockedContainer.php
+
+		-
 			rawMessage: 'Parameter #3 $mutators of method Infection\Mutation\FileMutationGenerator::generate() expects array<Infection\Mutator\Mutator<PhpParser\Node>>, array{Infection\Mutator\Arithmetic\Plus} given.'
 			identifier: argument.type
 			count: 1
@@ -1400,6 +1406,12 @@ parameters:
 			identifier: argument.type
 			count: 1
 			path: ../tests/phpunit/Mutator/Extensions/BCMathConfigTest.php
+
+		-
+			rawMessage: 'Method Infection\Tests\Mutator\Extensions\BCMathTest::test_it_can_mutate() has parameter $settings with no value type specified in iterable type array.'
+			identifier: missingType.iterableValue
+			count: 1
+			path: ../tests/phpunit/Mutator/Extensions/BCMathTest.php
 
 		-
 			rawMessage: 'Parameter #1 $callback of function array_map expects (callable(int<1, max>): mixed)|null, Closure(string): non-falsy-string given.'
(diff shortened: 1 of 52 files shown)


devTools/phpstan.neon
```diff
@@ -105,7 +105,12 @@ parameters:
 
         # tests
         - '#Dynamic call to static method PHPUnit\\Framework\\.*::.*#'
-        - '#Method Infection\\Tests\\.*::.*\(\) return type has no value type specified in iterable type iterable#'
+        -
+            identifier: missingType.iterableValue
+            path: ../tests/*
+        -
+            identifier: missingType.generics
```
theofidry, line 110: is this one necessary? The removed baseline entries seem legit

sanmai: What did you see?

I spot checked a few, and they looked like something nobody should ever fix:


> Method Infection\Tests\Process\Runner\MutationTestingRunnerTest::someIterable() return type with generic class PHPUnit\Framework\Constraint\Callback does not specify its types: CallbackInput

It makes no difference to document these, and we validate this code runs during CI anyway.

> Property Infection\Tests\Mutator\Util\AbstractValueToNullReturnValueTest::$testSubjectStub with generic class Infection\Mutator\Util\AbstractValueToNullReturnValue does not specify its types: TNode

There's no specific singular type, it is used as a mock.

devTools/phpstan-baseline.neon
```diff
@@ -1449,12 +1401,6 @@ parameters:
 			count: 1
 			path: ../tests/phpunit/Mutator/Extensions/BCMathConfigTest.php
 
-		-
-			rawMessage: 'Method Infection\Tests\Mutator\Extensions\BCMathTest::test_it_can_mutate() has parameter $settings with no value type specified in iterable type array.'
```
theofidry, line 1453: this one seems legit

devTools/phpstan-baseline.neon
```diff
@@ -1608,54 +1530,18 @@ parameters:
...
 			path: ../tests/phpunit/Mutator/MutatorRobustnessTest.php
 
-		-
-			rawMessage: 'Property Infection\Tests\Mutator\NoopMutatorTest::$mutatorMock with generic interface Infection\Mutator\Mutator does not specify its types: TNode'
-			identifier: missingType.generics
-			count: 1
-			path: ../tests/phpunit/Mutator/NoopMutatorTest.php
-
-		-
-			rawMessage: 'Method Infection\Tests\Mutator\ProfileListProvider::getProfiles() return type has no value type specified in iterable type array.'
```
theofidry, line 1636: this one seems legit

devTools/phpstan-baseline.neon
```diff
@@ -1608,54 +1530,18 @@ parameters:
...
-			path: ../tests/phpunit/Mutator/ProfileListProvider.php
-
 		-
 			rawMessage: 'Parameter #1 $objectOrClass of class ReflectionClass constructor expects class-string<T of object>|T of object, string given.'
 			identifier: argument.type
 			count: 1
 			path: ../tests/phpunit/Mutator/ProfileListProvider.php
 
-		-
-			rawMessage: 'Method Infection\Tests\Mutator\ProfileListTest::test_all_mutator_profiles_are_sorted_lexicographically() has parameter $profileOrMutators with no value type specified in iterable type array.'
```
theofidry, line 1648: this one seems legit

devTools/phpstan-baseline.neon
```diff
@@ -1668,12 +1554,6 @@ parameters:
 			count: 1
 			path: ../tests/phpunit/PhpParser/FakeTokenTest.php
 
-		-
-			rawMessage: 'Method Infection\Tests\PhpParser\MutatedNodeTest::test_it_can_be_instantiated() has parameter $node with no value type specified in iterable type array.'
```
theofidry, line 1672: this one seems legit

devTools/phpstan-baseline.neon
```diff
@@ -1698,24 +1578,6 @@ parameters:
 			count: 1
 			path: ../tests/phpunit/PhpParser/Visitor/VisitorTestCase/ConcreteVisitorTestCase.php
 
-		-
-			rawMessage: 'Method Infection\Tests\Process\Runner\MutationTestingRunnerTest::emptyIterable() return type with generic class PHPUnit\Framework\Constraint\Callback does not specify its types: CallbackInput'
```
theofidry, line 1702: those 3 seems legit

devTools/phpstan-baseline.neon
```diff
@@ -1824,18 +1674,6 @@ parameters:
 			count: 1
 			path: ../tests/phpunit/TestFramework/Coverage/Locator/BaseReportLocator/DemoReportLocator.php
 
-		-
-			rawMessage: 'Method Infection\Tests\TestFramework\Coverage\XmlReport\TestLocatorTest::getTestsLocations() return type has no value type specified in iterable type array.'
```
theofidry, line 1828: this one seems legit

devTools/phpstan.neon
theofidry: I commented a few but I admit I didn't look super deep into either, as in I did not try to fix them and see what was really needed. It could be that the fix is too BS and not worth addressing.

I also have the feeling I'm missing something on the Mutator and I started to look into it at some point but didn't get very far, so I don't really have an opinion about it either.

sanmai: Thanks, I'm going to revert this part as it sounds like you may want to fix an issue like this silences - your comments were mostly about `missingType.iterableValue` so I'm going to drop that instead of `missingType.generics`

theofidry (approved):

PR #3458, opened 2026-08-09
fix(phpunit): Remove the defects test order when disabling the test r…

### LINKED ISSUE #3457: bug: `recordTestRunHistory="false"` in generated initial config breaks projects using `executionOrder="defects"` on PHPUnit 13.3
| Question    | Answer
| Infection version | 0.34.2
| Test Framework version | PHPUnit 13.3.0
| PHP version | 8.5.9
| Platform    | Fedora Linux

Since PHPUnit 13.3.0, Infection's initial test run fails for any project that uses `executionOrder="defects"` (or `executionOrder="depends,defects"`) in its `phpunit.xml`.

Infection generates `phpunitConfiguration.initial.infection.xml` by copying the project's configuration and adding, among others, `recordTestRunHistory="false"`. The `executionOrder` attribute is copied verbatim. PHPUnit 13.3 combines these two into a test runner warning:

> Tests cannot be ordered by defects because recording of the test run history is disabled

Because Infection also adds `failOnWarning="true"` and `stopOnDefect="true"` to the generated configuration, that warning aborts the run before a single test is executed (`No tests executed!`), and Infection reports `Project tests must be in a passing state before running Infection.`

The relevant PHPUnit code is `PHPUnit\TextUI\Application::initializeTestRunHistory()`:

```php
if ($configuration->executionOrderDefects() === TestSuiteSorter::ORDER_DEFECTS_FIRST) {
    EventFacade::emitter()->testRunnerTriggeredPhpunitWarning(
        'Tests cannot be ordered by defects because recording of the test run history is disabled'
    );
```

On Infection's side, the attribute is written by `XmlConfigurationManipulator::deactivateResultCaching()`:

```php
public function deactivateResultCaching(string $version, SafeDOMXPath $xPath): void
{
    $attribute = version_compare($version, '13.3', '>=') ? 'recordTestRunHistory' : 'cacheResult';
    $this->setAttributeValue($xPath, $attribute, 'false');
}
```

The mapping of `cacheResult` to `recordTestRunHistory` is already in place, but the semantics changed: under PHPUnit &lt;= 13.2, `cacheResult="false"` together with `executionOrder="defects"` was silently tolerated, whereas PHPUnit 13.3 warns about it.

Suggested fixes, either of which resolves the problem (both verified locally against the generated configuration below):

1. Set `recordTestRunHistory="true"` for the initial run, mirroring what `handleResultCacheAndExecutionOrder()` already does for mutant runs.
2. Do not touch `recordTestRunHistory` at all for PHPUnit >= 13.3. PHPUnit's default is `true`, so ordering by defects keeps working, and the test run history file is written to the project's `cacheDirectory` just as it would be for a plain PHPUnit run.

If neither is desirable, the alternative would be to also override `executionOrder` in the generated configuration whenever `recordTestRunHistory` is disabled, so the two settings cannot contradict each other.

<details>

 ```json5
{
    "$schema": "https://raw.githubusercontent.com/infection/infection/0.34.0/resources/schema.json",
    "phpUnit": {
        "customPath": "tools/phpunit"
    },
    "staticAnalysisTool": "phpstan",
    "phpStan": {
        "customPath": "tools/phpstan"
    },
    "threads": "max",
    "source": {
        "directories": [
            "src"
        ]
    },
    "mutators": {
        "@default": true
    },
    "minMsi": 100,
    "minCoveredMsi": 100
}
 ```
</details>

<details>

 ```xml
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/13.3/phpunit.xsd"
         bootstrap="vendor/autoload.php"
         cacheDirectory=".phpunit.cache"
         executionOrder="depends,defects"
         beStrictAboutCoverageMetadata="true"
         beStrictAboutOutputDuringTests="true"
         displayDetailsOnAllIssues="true"
         failOnAllIssues="true"
         colors="true">
    <testsuites>
        <testsuite name="unit">
            <directory>tests/unit</directory>
        </testsuite>

        <testsuite name="integration">
            <directory>tests/integration</directory>
        </testsuite>

        <testsuite name="end-to-end">
            <directory>tests/end-to-end</directory>
        </testsuite>
    </testsuites>

    <source ignoreIndirectDeprecations="true" restrictNotices="true" restrictWarnings="true">
        <include>
            <directory>src</directory>
        </include>
    </source>
</phpunit>
 ```
</details>

<details>
 <summary>Generated phpunitConfiguration.initial.infection.xml</summary>

 ```xml
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/13.2/phpunit.xsd" bootstrap="/usr/local/src/neutrino/vendor/autoload.php" cacheDirectory=".phpunit.cache" executionOrder="depends,defects" requireCoverageMetadata="true" beStrictAboutCoverageMetadata="true" beStrictAboutOutputDuringTests="true" displayDetailsOnAllIssues="true" failOnAllIssues="true" colors="false" failOnRisky="true" failOnWarning="true" stopOnDefect="true" recordTestRunHistory="false" stderr="false">
  <testsuites>
    <testsuite name="unit">
      <directory>/usr/local/src/neutrino/tests/unit</directory>
    </testsuite>
    <testsuite name="integration">
      <directory>/usr/local/src/neutrino/tests/integration</directory>
    </testsuite>
    <testsuite name="end-to-end">
      <directory>/usr/local/src/neutrino/tests/end-to-end</directory>
    </testsuite>
  </testsuites>
  <source ignoreIndirectDeprecations="true" restrictNotices="true" restrictWarnings="true">
    <include>
      <directory>/usr/local/src/neutrino/src</directory>
    </include>
  </source>
</phpunit>
 ```
</details>

 ```shell
./tools/infection
 ```

<details>

 ```
         Infection runs the test suite in a RANDOM order. Make sure your tests
         do not have hidden dependencies.

         You can add these attributes to `phpunit.xml` to check it: <phpunit
         executionOrder="random" resolveDependencies="true" ...

         If you don't want to let Infection run tests in a random order, set the
         `executionOrder` to some value, for example <phpunit
         executionOrder="default"

         '/usr/local/src/neutrino/tools/phpunit' '--configuration'
         '/tmp/infection/phpunitConfiguration.initial.infection.xml'
         '--exclude-source-from-xml-coverage'
         '--coverage-xml=/tmp/infection/coverage-xml'
         '--log-junit=/tmp/infection/junit.xml'
         PHPUnit reported an exit code of 1.
         STDOUT:
         PHPUnit 13.3.0 by Sebastian Bergmann and contributors.

         Runtime:       PHP 8.5.9 with Xdebug 3.5.3
         Configuration:
         /tmp/infection/phpunitConfiguration.initial.infection.xml

         There was 1 PHPUnit test runner warning:

         1) Tests cannot be ordered by defects because recording of the test run
         history is disabled

         No tests executed!

         Generating code coverage report in PHPUnit XML format ... done
         [00:00.018]
 ```
</details>

sebastianbergmann:
## Description

Since PHPUnit 13.3, ordering tests by `defects` while the recording of the test run history is disabled is a test runner warning:

> Tests cannot be ordered by defects because recording of the test run history is disabled

The initial run configuration sets both `recordTestRunHistory="false"` and `failOnWarning="true"`, so on a project whose `phpunit.xml` declares `executionOrder="defects"` (or `"depends,defects"`) that warning aborts the run before a single test is executed. Infection then reports `Project tests must be in a passing state before running Infection.` even though the suite is perfectly healthy.

#3447 already anticipated this and guarded `PhpUnitAdapter::supportsExecutionOrderDefectsRandom()`, but that guard is only consulted from `InitialConfigBuilder::addRandomTestsOrderAttributesIfNotSet()`, i.e. only when Infection itself would introduce the order. When the project's own configuration already declares it, `addAttributeIfNotSet()` is a no-op, the value is copied verbatim, and `deactivateResultCaching()` still writes `recordTestRunHistory="false"` next to it — the one path that reaches the very combination the guard exists to prevent.

This PR enforces the constraint where the history is disabled, so it holds regardless of who wrote the order. Dropping the `defects` component is safe because PHPUnit's config loader parses it as a flag independent from the rest of `executionOrder` (`defects` sets `$defectsFirst`, while `random`/`duration`/… set the order and `depends`/`no-depends` set `resolveDependencies`), so the remaining components keep their meaning. It also keeps the initial run out of the project's own cache directory, which the alternative — enabling the history for the initial run — would not.

## Changes

- Fixes the initial run configuration by stripping `defects` from `executionOrder` whenever `recordTestRunHistory` is disabled, falling back to `default` when no other component remains.
- Adds unit coverage for the four order shapes (`defects`, `defects,random`, `depends,defects`, and an order without `defects`) plus a PHPUnit 13.2 case asserting the order is left untouched there.
- Updates the `PHPUnit_13-3` end-to-end fixture to order by `depends,defects`, covering the path the guard in #3447 does not.
- Updates the PHPStan baseline for the one added `SafeDOMXPath` reference counted by the package boundary rule.

## Reviewer Notes

- The main decision is *where* the constraint belongs. Setting `recordTestRunHistory="true"` for the initial run, or leaving the attribute alone on 13.3+ (its default is `true`), would also silence the warning, but both make the initial run write a run-history file into the project's configured `cacheDirectory` as a side effect of an Infection run. Normalising `executionOrder` avoids that.
- `PhpUnitAdapter::supportsExecutionOrderDefectsRandom()` is deliberately left untouched. Its `>= 13.3` guard is now redundant — an Infection-written `defects,random` would be stripped back to `random` anyway, giving an identical generated configuration — but it is kept as defence in depth, and removing it would only change the wording of the failure recommendation. Happy to drop it in a follow-up if preferred.
- One pre-existing behaviour worth noting: on PHPUnit >= 13.3, projects that do *not* set `executionOrder` get `random` for the initial run rather than `defects,random`. That is harmless, since there is no history to order by during the initial run, but it does mean 13.3 differs from 13.2 for those projects too.
- Verified against a real PHPUnit 13.3.0: with the fix reverted, the updated end-to-end fixture reproduces the reported failure; with the fix, it completes and matches `expected-output.txt`. PHPStan, Mago and PHP-CS-Fixer are clean.

## Related issues

Fixes https://github.com/infection/infection/issues/3457.


tests/phpunit/TestFramework/PhpUnit/Config/XmlConfigurationManipulatorTest.php
```diff
@@ -690,6 +690,68 @@ static function (XmlConfigurationManipulator $configManipulator, SafeDOMXPath $x
...
+
+    public static function executionOrderProvider(): iterable
+    {
+        yield 'defects only' => ['defects', 'default'];
+
+        yield 'defects first' => ['defects,random', 'random'];
+
+        yield 'defects last' => ['depends,defects', 'depends'];
+
+        yield 'no defects' => ['depends,random', 'depends,random'];
```
theofidry, line 729: I think it is missing the case of when it is just `defects` with no fallback (which I assume should end up being either entirely removed or declared with `default`)

theofidry (approved):

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
```diff
@@ -98,9 +100,16 @@ public function removeExistingLoggers(SafeDOMXPath $xPath): void
...
+        if (version_compare($version, '13.3', '<')) {
+            $this->setAttributeValue($xPath, 'cacheResult', 'false');
 
-        $this->setAttributeValue($xPath, $attribute, 'false');
+            return;
+        }
+
+        $this->setAttributeValue($xPath, 'recordTestRunHistory', 'false');
+
+        // Ordering tests by defects requires the test run history which is deactivated right above. PHPUnit ignored that combination silently until 13.3 turned it into a test runner warning, which aborts the initial run because of failOnWarning https://github.com/sebastianbergmann/phpunit/blob/13.3.0/src/TextUI/Application.php
```
maks-rafalko, line 111: We have this logic now:

https://github.com/infection/infection/blob/18d9bef39fae250f202920d1453ab96e793e6637/src/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilder.php#L159-L163

which adds `defects,random` for PHPUnit <13.3, so for PHPUnit 13.2.

But the comments says

> PHPUnit ignored that combination silently until 13.3

shouldn't we update `supportsExecutionOrderDefectsRandom()` so that it returns `false` for those PHPUnit versions that silently ignore `defects` in `defects,random` combination with disabled result cache?

What happens right now for PHPUnit 13.2

we add `defects,random`, then we set `cacheResult=false`, and PHPUnit silently ignores `defects`. So it's probably better to not set `defects,...` in the first place?

maks-rafalko: fixed

tests/phpunit/TestFramework/PhpUnit/Config/XmlConfigurationManipulatorTest.php
maks-rafalko: actually, line 723 is exactly this

https://github.com/infection/infection/pull/3458/changes#diff-2e384b6396e35a4a31560adf4a8f1fd64f111351e9fddde1517b222d43f6636cR723

maks-rafalko:
thank you @sebastianbergmann, released https://github.com/infection/infection/releases/tag/0.35.2

PR #3461, opened 2026-08-09
fix: Replace deprecated Node\Scalar\LNumber with Node\Scalar\Int_

sanmai:
## Description

php-parser deprecated `LNumber/DNumber` in favor of `Int_`/`Float_`; let's update mutators and tests to the new class names and shrink the Mago baseline.

We have nikic/php-parser at `^5.6.2`, and `Int_` was introduced in 5.0.0.

### Reviewer Notes

This is mostly a search and replace kind of thing, with a singular exception - I commented inline.

Escaping mutants are fixed by #3463

### DIFF
--- a/src/Mutator/ReturnValue/FloatNegation.php
+++ b/src/Mutator/ReturnValue/FloatNegation.php
@@ -108,7 +108,7 @@ final class FloatNegation implements Mutator
         $expression = $node->expr;
 
         if ($expression instanceof Node\Expr\UnaryMinus) {
-            /** @var Node\Scalar\LNumber $innerExpression */
+            /** @var Node\Scalar\DNumber $innerExpression */
             $innerExpression = $expression->expr;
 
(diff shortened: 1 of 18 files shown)


src/Mutator/ReturnValue/FloatNegation.php
```diff
@@ -108,7 +108,7 @@ private function getFloatValueOfNode(Node $node): float
         $expression = $node->expr;
 
         if ($expression instanceof Node\Expr\UnaryMinus) {
-            /** @var Node\Scalar\LNumber $innerExpression */
+            /** @var Node\Scalar\DNumber $innerExpression */
```
sanmai, line 111: This was a documentation bug, as `LNumber` is not a float.

theofidry (approved):

maks-rafalko (approved):

PR #3462, opened 2026-08-09
refactor(container): Switch to direct class-to-class injection

sanmai:
## Description

The DI container at 0.1.22 now autowires a concrete class name given as a value, so we can direct interfaces towards implementations without any extra closures:

```
PrettyPrinterAbstract::class => InfectionPrettyPrinter::class,
Git::class => CommandLineGit::class,
```

I also removed closures and getters that the container resolved automatically.

### DIFF
--- a/src/Container/Container.php
+++ b/src/Container/Container.php
@@ -275,7 +275,9 @@ final class Container extends DIContainer
     {
         $container = new self([
             IndexXmlCoverageParser::class => IndexXmlCoverageParserBuilder::class,
-            Tracer::class => TraceProviderAdapterTracer::class,
+            Tracer::class => static fn (self $container) => new TraceProviderAdapterTracer(
+                $container->getTraceProvider(),
+            ),
             TraceProvider::class => static fn (self $container): TraceProvider => new CoveredTraceProvider(
                 $container->getPhpUnitXmlCoverageTraceProvider(),
                 $container->getJUnitTestExecutionInfoAdder(),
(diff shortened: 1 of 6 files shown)


src/Container/Container.php
```diff
@@ -618,10 +602,7 @@ public static function create(): self
                 );
             },
             MemoizedComposerExecutableFinder::class => static fn (): ComposerExecutableFinder => new MemoizedComposerExecutableFinder(new ConcreteComposerExecutableFinder()),
-            Git::class => static fn (self $container): Git => new CommandLineGit(
-                new ShellCommandLineExecutor(),
```
theofidry, line 622: note that strictly speaking there is a difference here as (I assume) the `ShellCommandLineExecutor` will be a shared instance now, but personally I'm happy about this change

tests/phpunit/Container/ContainerTest.php
```diff
@@ -182,6 +182,27 @@ public function test_factory_is_essential(string $id): void
         // Iterate over $reflection->iterateExpectedConcreteServices(), calling getService() for each service
     }
 
+    public static function provideImplementationsWithReflection(): iterable
+    {
+        $reflection = new ContainerReflection(Container::create());
+
+        foreach ($reflection->getImplementations() as $id => $implementation) {
```
theofidry, line 189: what does `ContainerReflection::getImplementations()` really do?

theofidry:
(LGTM, just didn't approve to get a reply before the merge)

maks-rafalko (approved):

tests/phpunit/Container/ContainerTest.php
sanmai: Reads the value of the private property `implementations` so we can attempt building all things container supposed to provide.

sanmai: That's actually made me thinking that it should not do anything at all, so here's #3470 to strip it do the bare bones.

PR #3474, opened 2026-08-11
fix: Fix version detection with quiet console output

### LINKED ISSUE #3123: bug(testFramework): Retrieving the version of the test framework/static analysis fails depending of `SHELL_VERBOSITY`
| Question    | Answer
| Infection version | 0.32.7
| PhpSpecAdapter version | 0.3.1
| CodeceptionAdapter version | 0.4.6
| TestoAdapater | 0.1.3
| CI | either
| PHP version | any
| Platform    | any
| Github Repo | -

Executing:

```
bin/infection --test-framework=phpstan --quiet
```

Will cause a failure. This is because the Symfony Application will adjust `SHELL_VERBOSITY` accordingly and the process we start to get the package version inherits those environment variables, so we end up doing `SHELL_VERBOSITY=-2 phpstan --version --no-ansi` which will yield an empty string.

Fixing this should not be too hard, finding how to properly test it without looking completely artificial is a bit more annoying.

theofidry:
## Description

`SHELL_VERBOSITY` is an environment variable that some applications honour to control output verbosity. Symfony Console supports it by default, but it is not exclusive to Symfony; for example, [Python's Poetry also honours it](https://python-poetry.org/docs/cli/#global-options).

[Symfony Console's `Application` sets `SHELL_VERBOSITY`](https://github.com/symfony/symfony/blob/8204558/src/Symfony/Component/Console/Application.php#L262), while Symfony `Process` inherits environment variables, as demonstrated below. Consequently, Infection's verbosity can affect its subprocesses:

<details>
<summary>Demonstration of the `SHELL_VERBOSITY` inheritance</summary>

```php
<?php // demo.php

require __DIR__.'/vendor/autoload.php';

// Simulate SHELL_VERBOSITY having been set by the parent application.
putenv('SHELL_VERBOSITY=-1');
$_ENV['SHELL_VERBOSITY'] = '-1';
$_SERVER['SHELL_VERBOSITY'] = '-1';

$command = [
    PHP_BINARY,
    '-r',
    'echo getenv("SHELL_VERBOSITY"), PHP_EOL;',
];

// No environment explicitly provided: SHELL_VERBOSITY is inherited.
$inheritedProcess = new Process($command);
$inheritedProcess->mustRun();

echo 'Inherited: '.$inheritedProcess->getOutput();

// Explicitly reset the verbosity for the child process.
$overriddenProcess = new Process(
    command: $command,
    env: ['SHELL_VERBOSITY' => '0'],
);
$overriddenProcess->mustRun();

echo 'Overridden: '.$overriddenProcess->getOutput();
```

Then:

```
$ php demo.php
Inherited: -1
Overridden: 0
```

</details>

This interaction causes bugs such as https://github.com/infection/infection/issues/3123:

```shell
bin/infection --test-framework=phpstan --quiet
# fails
```

This PR addresses this class of problem by explicitly passing `SHELL_VERBOSITY` to processes whose output Infection reads. Not every case fixes a current bug: some applications do not honour the variable, or no longer honour it in specific contexts. Applying it consistently remains worthwhile because this behaviour may change and it makes the requested verbosity explicit. The table distinguishes confirmed bugs from safeguards.

| Place                                     | Process or use case       | Classification                     | Rationale                                |
|-------------------------------------------|---------------------------|------------------------------------|------------------------------------------|
| `SymfonyProcessShellCommandRunner`        | Shared-runner commands    | General safeguard                  | Covers all shared commands.              |
| `AbstractTestFrameworkAdapter`            | Test-framework version    | Actual bug class                   | Parses stdout.                           |
| `PHPStanAdapter`                          | PHPStan version           | **Confirmed bug**                  | Reproduced with `-1`.                    |
| `MagoAdapter`                             | Mago version              | Good measure                       | Parses stdout.                           |
| `TestFrameworkFinder`                     | `composer config bin-dir` | Good measure                       | Uses stdout as a path.                   |
| `StaticAnalysisToolExecutableFinder`      | `composer config bin-dir` | Good measure                       | Uses stdout as a path.                   |
| `CommandLineGit`                          | Git queries               | Good measure                       | Uses stdout as data.                     |
| `AdapterInstaller`                        | `composer require`        | Good measure                       | Covered by shared default.               |
| `InitialTestsRunProcessFactory`           | Initial tests             | **Confirmed bug with Codeception** | Uses output for diagnostics and metrics. |
| `MutantProcessContainerFactory`           | Mutant tests              | **Confirmed bug**                  | Uses output to classify mutants.         |
| `InitialStaticAnalysisProcessFactory`     | Initial static analysis   | Good measure                       | Retains diagnostics.                     |
| `PHPStanMutantProcessFactory`             | PHPStan mutant analysis   | Good measure                       | Retains reporting output.                |
| `MagoMutantProcessFactory`                | Mago mutant analysis      | Good measure                       | Retains reporting output.                |
| `DebugStaticAnalysisMutantProcessFactory` | Debug mutant analysis     | Good measure                       | Retains debug output.                    |

PHPUnit _no longer_ uses `symfony/console` and is therefore no longer affected, while Mago does not honour the variable. Composer previously honoured it for `config` commands but no longer does so in recent versions.

## Changes

- Set `SHELL_VERBOSITY=0` for processes for which we read the output.
- Add end-to-end tests.

## Related issues

This belongs to the class of issues reported in https://github.com/infection/infection/issues/3322.

Fixes https://github.com/infection/infection/issues/3123.


src/Process/Factory/InitialStaticAnalysisProcessFactory.php
```diff
@@ -53,6 +53,7 @@ public function createProcess(): Process
     {
         return new Process(
             command: $this->staticAnalysisToolAdapter->getInitialRunCommandLine(),
+            env: ['SHELL_VERBOSITY' => '0'],
```
sanmai: what are the chance we can make this `'0'` a constant with a description? it is not obvious that `-1` is actually non-verbose and `0` is the default verbosity

sanmai (approved):
LGTM otherwise

src/Process/Factory/InitialStaticAnalysisProcessFactory.php
theofidry: Good idea.

PR #3477, opened 2026-08-12
fix(mutator): Guard LogicalOr against non-string variable variable names

sanmai:
## Description

`${$s}` has a dynamic (Expr) name rather than a string one; so building the variable-name list used to push that `Expr` straight into `array_intersect()` is going to blow up

Suggested by PHPStan.

### DIFF
--- a/tests/phpunit/Mutator/Boolean/LogicalOrTest.php
+++ b/tests/phpunit/Mutator/Boolean/LogicalOrTest.php
@@ -213,6 +213,21 @@ final class LogicalOrTest extends BaseMutatorTestCase
                     PHP,
             ),
         ];
+
+        yield 'It mutates logical or when the same variable variable is tested against "Identical" on both sides' => [
+            self::wrapCodeInMethod(
+                <<<'PHP'
+                    $s = 'other';
+                    ${$s} === 'hello' || ${$s} === 'world';
+                    PHP,
+            ),
+            self::wrapCodeInMethod(
+                <<<'PHP'
+                    $s = 'other';
+                    ${$s} === 'hello' && ${$s} === 'world';
+                    PHP,
+            ),
+        ];
     }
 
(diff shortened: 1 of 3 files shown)


sidz (approved):

tests/phpunit/Mutator/Boolean/LogicalOrTest.php
```diff
@@ -153,6 +153,66 @@ private static function equalityMutationsProvider(): iterable
...
+                    $myVar === 'hello' || 'world' === ${$s};
+                    PHP,
+            ),
+            self::wrapCodeInMethod(
+                <<<'PHP'
+                    $s = 'other';
+                    $myVar === 'hello' && 'world' === ${$s};
+                    PHP,
+            ),
+        ];
```
theofidry, line 215: I would add the case which was originally what the `array_intersect()` was for:

```php
${$s} === 'hello' || ${$s} === 'world'
```

(the expectation being that `||` is mutated, whereas if `${$s}` was `$s` instead, it would not have `||` mutated)

theofidry (commented):
Suggested description:

<hr/>

`LogicalOr` mutates `$a || $b` into `$a && $b`. However, it suppresses that mutation for mutually exclusive identity
  checks such as `$value === 'foo' || $value === 'bar'` as changing this would effectively guarantee the expression to be `false`.

To recognise this case, `LogicalOr::canMutate()` gathers the variable names from both comparisons to look for the shared name with `array_intersect($varNameLeft, $varNameRight)`. However, if one operant is not a string but an expression, e.g. for `${$a} || $b`, then `array_intersect()` causes a crash.

This PR checks for the types of the operand to exclude the variable name check if the operand is not a string. Note that this means the following case is not detected anymore*:

```php
`${$value} === 'foo' || ${$value} === 'bar'`
```

*: in reference to the identity check, without this PR the above causes a crash.

theofidry:
LGTM otherwise 👍

PR #3473, opened 2026-08-11
refactor(process): Refactor initial static analysis to use ShellCommandRunner

theofidry:
## Description

Use the shared blocking command execution boundary for the initial static-analysis run. This removes direct Symfony Process construction from this path and makes its execution behaviour testable without launching a sub-process.

## Reviewer Notes

The initial static-analysis run deliberately passes `timeout: null`, retaining the previous unlimited timeout. The review focus is the replacement of the live `Process` result with `CompletedProcess` while preserving event dispatch and failure diagnostics.

## Related issues

https://github.com/infection/infection/issues/3322

PR #3478, opened 2026-08-12
refactor(process): Update the ShellCommandRunner types

theofidry:
## Description

This PR is a small preparatory step for https://github.com/infection/infection/pull/3473.

## Changes

- Make `$command` of `ShellCommandRunner::run()` from a `list<string>` into a `string[]` as the former is needlessly constraining and it is easier to do this update rather than making `$command` of `ShellCommandRunner::mustRun()` a `list<string>`. As a result of this change `CompletedProcess#command` also becomes a `string[]`.
- Add the callback type to `ShellCommandRunner::mustRun()`.
- Extract the callback type to a PHPStan type for readability.

## Related issues

### DIFF
--- a/.github/workflows/zizmor.yaml
+++ b/.github/workflows/zizmor.yaml
@@ -33,7 +33,7 @@ jobs:
           persist-credentials: false
 
       - name: Run zizmor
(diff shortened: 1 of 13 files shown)


theofidry:
Actually reading this again this should rather be solved by making `ShellCommandRunner::run()` take a `string[] $command` instead, there is no need to require a `list<string>` instead.

PR #3484, opened 2026-08-13
feat: allow filtering logged mutant types via --log-verbosity

### LINKED ISSUE #2445: More fine-grained settings for mutant types displayed in the logs

Currently, we have [`--log-verbosity=all|none|default`](https://infection.github.io/guide/command-line-options.html#log-verbosity)

> The verbosity of the log file, all - this mode will add “Killed mutants” into log file and add additional information, default - normal mode will skip “Killed mutants” section in the log file, none - which will disable logging to files.

But the things are now much more complex. We also have

- syntax errors
- errors
- kills by Static Analysis
- Not Covered
- Ignored 
- etc.

And, sometimes we want to add one type of mutants to the report but exclude others. Example: https://github.com/CuyZ/Valinor/pull/721#discussion_r2419671057

In addition to `all`, `default`, `none` (keeping them for BC), we can add a feature to add or exclude particular types of mutants from the log report: 

`--log-verbosity=all,-errors` - all without errors
`--log-verbosity=default,+killed,+killed-by-sa` - default ones and killed (including kills by static analysis)

So basically to support `+xxx` and `-yyy` notation.

This will be very useful to configure it for CI, so it's not actually a problem that  this will be quite verbose - we will write it once and forget.

I would not add any new flags, because we already have a very complex and hard to understand system that affects the output, like

- `--log-verbosity`
- `--debug`
- `-v|-vv`
- `--show-mutations`
- `--with-uncovered`

So, actually no new flag is needed from my POV.

mmustafasenoglu:
## Summary

Closes #2445

This extends the `--log-verbosity` option so you can choose exactly which mutant types end up in the text log, instead of only the `all` / `default` / `none` presets.

`--log-verbosity` now also accepts a comma separated list of mutant-type slugs. Each slug can be prefixed with `+` (include) or `-` (exclude) and the list is applied on top of the `default` set, for example:

- `default,-timeout` — everything except timed-out mutants
- `none,+escaped,+timeout` — only escaped and timed-out mutants
- `all,-killed,-error` — full detail minus killed-by-tests and errors

Available slugs: `escaped`, `timed-out` (alias `timeout`), `skipped`, `killed`, `killed-sa`, `error` (alias `errors`), `syntax-error`, `not-covered`, `ignored`.

The `all`, `default` and `none` shorthands keep their existing behavior, so this is fully backward compatible.

## How it works

- `LogVerbosity` gains `resolveDisplayedStatuses()` / `isValidOption()` which parse the option into a `DetectionStatus[]` set.
- `BaseTextFileReporter` renders only the statuses present in that set (the `only-covered` gate for `not-covered` is preserved).
- Process output and the "use `--log-verbosity=all`" note are still shown whenever a debug-level status (killed/error) is part of the set, matching the previous `all` behavior.

## Test plan

- [x] Unit tests for `LogVerbosity::resolveDisplayedStatuses` / `isValidOption`
- [x] `TextFileReporterTest` covers status filtering
- [x] Existing reporter/config/command suites (1338 tests) pass

I verified the change locally with PHP 8.5 + the project's PHPUnit suite.

### DIFF
--- a/src/Configuration/Configuration.php
+++ b/src/Configuration/Configuration.php
@@ -43,12 +43,14 @@ use Infection\Configuration\Entry\PhpStan;
 use Infection\Configuration\Entry\PhpUnit;
 use Infection\Configuration\Entry\Source;
 use Infection\Configuration\SourceFilter\SourceFilter;
+use Infection\Console\LogVerbosity;
 use Infection\Mutator\Mutator;
 use Infection\StaticAnalysis\StaticAnalysisToolTypes;
 use Infection\TestFramework\TestFrameworkTypes;
 use function is_string;
 use function ltrim;
 use PhpParser\Node;
+use function sprintf;
 use Webmozart\Assert\Assert;
 
 /**
@@ -57,12 +59,6 @@ use Webmozart\Assert\Assert;
  */
 readonly class Configuration
 {
-    private const array LOG_VERBOSITY = [
-        'all',
-        'none',
-        'default',
-    ];
-
     /**
      * @param array<string, Mutator<Node>> $mutators
      * @param array<string, array<int, string>> $ignoreSourceCodeMutatorsMap
@@ -113,7 +109,10 @@ readonly class Configuration
     ) {
         Assert::nullOrGreaterThanEq($processTimeout, 0);
         Assert::allIsInstanceOf($mutators, Mutator::class);
-        Assert::oneOf($logVerbosity, self::LOG_VERBOSITY);
+        Assert::true(
+            LogVerbosity::isValidOption($logVerbosity),
+            sprintf('Invalid "--log-verbosity" option "%s".', $logVerbosity),
+        );
         Assert::oneOf($testFramework, TestFrameworkTypes::getTypes());
         Assert::nullOrOneOf($staticAnalysisTool, StaticAnalysisToolTypes::getTypes());
         Assert::nullOrGreaterThanEq($minMsi, 0.);
--- a/src/Console/LogVerbosity.php
+++ b/src/Console/LogVerbosity.php
@@ -35,10 +35,18 @@ declare(strict_types=1);
 
 namespace Infection\Console;
 
+use function array_filter;
 use function array_key_exists;
+use function array_values;
+use function explode;
 use function in_array;
 use Infection\CannotBeInstantiated;
+use Infection\Mutant\DetectionStatus;
+use InvalidArgumentException;
+use function sprintf;
+use function substr;
 use Symfony\Component\Console\Input\InputInterface;
+use function trim;
 
 /**
  * @internal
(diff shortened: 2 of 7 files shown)


src/Configuration/Configuration.php
```diff
@@ -113,7 +109,10 @@ public function __construct(
     ) {
         Assert::nullOrGreaterThanEq($processTimeout, 0);
         Assert::allIsInstanceOf($mutators, Mutator::class);
-        Assert::oneOf($logVerbosity, self::LOG_VERBOSITY);
+        Assert::true(
+            LogVerbosity::isValidOption($logVerbosity),
+            sprintf('Invalid "--log-verbosity" option "%s".', $logVerbosity),
```
theofidry, line 114: The `Configuration` should not need to be aware of the option names from `RunCommand`.

src/Reporter/BaseTextFileReporter.php
```diff
@@ -49,18 +52,25 @@
...
 {
+    /**
+     * @param array<int, DetectionStatus>|null $displayedStatuses
+     */
     public function __construct(
         private ResultsCollector $resultsCollector,
         private bool $debugVerbosity,
         private bool $onlyCoveredMode,
         private bool $debugMode,
+        private ?array $displayedStatuses = null,
```
theofidry, line 63: the value is always passed so there is no need to make it nullable: it makes this class code more complicated than necessary:


```suggestion
        private ?array $displayedStatuses,
```

src/Reporter/FileReporterFactory.php
```diff
@@ -133,23 +135,29 @@ private function wrapWithFileLogger(string $filePath, LineMutationTestingResults
 
     private function createTextLogger(Logs $logConfig): LineMutationTestingResultsReporter
     {
+        $displayedStatuses = LogVerbosity::resolveDisplayedStatuses($this->logVerbosity);
```
theofidry, line 138: I do not mind having the resolution of the log verbosity within `LogVerbosity::resolveDisplayedStatuses()`, but I think it makes more sense to resolve it within `ConfigurationFactory` and pass the _resolved_ log verbosities to `Configuration`.

Currently it's not a big deal because this is only passed to the `BaseTextFileReporter` based file reporters, but that should change in the future.

src/Console/LogVerbosity.php
```diff
@@ -77,11 +85,59 @@ final class LogVerbosity
...
+     * Map of a slug (as accepted by `--log-verbosity`) to the corresponding {@see DetectionStatus}.
+     *
+     * @var array<string, DetectionStatus>
+     */
+    public const array STATUS_SLUGS = [
+        'escaped' => DetectionStatus::ESCAPED,
+        'timed-out' => DetectionStatus::TIMED_OUT,
+        'timeout' => DetectionStatus::TIMED_OUT,
+        'skipped' => DetectionStatus::SKIPPED,
+        'killed' => DetectionStatus::KILLED_BY_TESTS,
```
theofidry, line 98: shouldn't this be `killed-by-tests`? Otherwise I would assume `killed` is equivalent to `killed-by-tests + killed-by-sa`

src/Console/LogVerbosity.php
```diff
@@ -77,11 +85,59 @@ final class LogVerbosity
         self::NONE_INTEGER => self::NONE,
     ];
 
+    /**
+     * Map of a slug (as accepted by `--log-verbosity`) to the corresponding {@see DetectionStatus}.
+     *
+     * @var array<string, DetectionStatus>
+     */
+    public const array STATUS_SLUGS = [
```
theofidry, line 93: I think the current list is a bit flawed. There is two different concerns to cover:

- mapping each detection status to an option slug, e.g. `DetectionStatus::ESCAPED` -> `'escaped'`
- creating a slug that captures a family of detection statuses, e.g. `errors` -> `DetectionStatus::ERROR, DetectionStatus::SYNTAX_ERROR`

I'm not sure what's the best way to implement that though.

src/Console/LogVerbosity.php
```diff
@@ -77,11 +85,59 @@ final class LogVerbosity
...
+        'escaped' => DetectionStatus::ESCAPED,
+        'timed-out' => DetectionStatus::TIMED_OUT,
```
theofidry, line 95: I'm confused with timed-out & timeout

theofidry (commented):
I could only take a partial look, just in case it was missed I think it would be good to suggest a new option name for it: https://github.com/infection/infection/issues/2445#issuecomment-3396316259

But maybe it's best to do it separately after all

PR #3506, opened 2026-08-26
perf: Stop retaining every Mutation until the end of the run

Meacue:
## Description
In one of my projects, I upgraded Infection from 0.29.14 to 0.35.2, after that, the local run started dying with a PHP out-of-memory error **(with 128mb limit)**. Working through the diff between the two versions with Claude Code (Opus 5), looking for what could have affected memory consumption, and that search led to `MutantFactory` and its `create()` method. 

### What
`later(static fn () => yield $mutation->getOriginalFileContent())` wraps the original file content in a deferred closure that captures the whole Mutation object. This replaces it with `now($mutation->getOriginalFileContent())`, which wraps the string that is already in memory: the same `Later\Interfaces\Deferred` contract, but no closure and nothing captured.

### Why
The deferred used to wrap real work: pretty-printing the AST back into source (#1420). Since #2448 (0.32.0), the original content is stored verbatim, and Mutation::getOriginalFileContent() is a plain getter over a private readonly string. There is nothing left to defer, only a closure holding the Mutation.

## Numbers
Measured on my own project which contains 2051 mutations, runs with `--threads=6`. Using `Memory:` line in Infection's own output.

<details>

 ```json5
{
    "$schema": "vendor/infection/infection/resources/schema.json",
    "source": {
        "directories": [
            "src"
        ]
    },
    "timeout": 10,
    "mutators": {
        "@default": true
    },
    "phpUnit": {
        "configDir": "."
    }
}
 ```
</details>

### Env 1. Windows, PHP 8.4 ZTS: 

| options | 0.29.14 | 0.35.2 (`later`) | 0.35.2 patched (`now`) |
|---|---|---|---|
| **with progress** `--dry-run` | 0.11 GB. | 0.20 GB | 0.20 GB |
| `--no-progress --dry-run`| 0.04 GB | 0.19 GB | **0.13 GB** |
| **with progress** |  0.16 GB | 0.21 GB | 0.21 GB |
| `--no-progress` | 0.12 GB | 0.20 GB | **0.14 GB** |

### Env 2. WSL2 Ubuntu, PHP 8.4 NTS + pcov

I didn't run the infection without the --dry-run option to keep the experiment running quickly.

| options | 0.29.14 | 0.35.2 (`later`) | 0.35.2 patched (`now`) |
|---|---|---|---|
| **with progress** `--dry-run` | 0.13 GB. | 0.22 GB | 0.22 GB |
| `--no-progress --dry-run`| 0.06 GB | 0.19 GB | **0.14 GB.** |

Different versions given different detection statuses on the same files. I think this could have affected the memory consumption results. Anyway, limit was reached, but seems better with patch. 

## Reviewer Notes

**You will not reproduce this on Infection repository with current config.** `infection.json5` sets `"global-ignoreSourceCodeByRegex": ["Assert::.*"]`. The `global-` prefix puts every mutator into `ignoreSourceCodeMutatorsMap`, so `MutationTestingRunner::ignoredByRegex()` calls `$mutant->getDiff()->get()` for every single mutant, which resolves the deferred and releases the `Mutation` immediately.

### DIFF
--- a/src/Mutant/MutantFactory.php
+++ b/src/Mutant/MutantFactory.php
@@ -38,8 +38,8 @@ namespace Infection\Mutant;
 use Infection\Differ\Differ;
 use Infection\Mutation\Mutation;
 use Later\Interfaces\Deferred;
-use function Later\later;
 use function Later\lazy;
+use function Later\now;
 use function sprintf;
 
 /**
@@ -64,7 +64,7 @@ class MutantFactory
         );
 
         $mutatedCode = lazy($this->createMutatedCode($mutation));
-        $originalPrettyPrintedFile = later(static fn () => yield $mutation->getOriginalFileContent());
+        $originalPrettyPrintedFile = now($mutation->getOriginalFileContent());
 
         return new Mutant(
             $mutantFilePath,


src/Mutant/MutantFactory.php
```diff
@@ -64,7 +64,7 @@ public function create(Mutation $mutation): Mutant
         );
 
         $mutatedCode = lazy($this->createMutatedCode($mutation));
-        $originalPrettyPrintedFile = later(static fn () => yield $mutation->getOriginalFileContent());
+        $originalPrettyPrintedFile = now($mutation->getOriginalFileContent());
```
theofidry, line 67: if we do so sure the contract is respected but... you loose the laziness of the evaluation of the Closure, which begs the question of why not making it a property straight up.

So whilst it's a slightly bigger change, I think it is worth changing this from a `Deferred` to a `string`.

theofidry (commented):
Nice find!

Note: couldn't check yet the result on my machine.

src/Mutant/MutantFactory.php
Meacue: Thanks for the quick response.

Agreed. My first reaction when I saw this was the same as your suggestion. However, I tried to keep the change minimal so that a performance fix wouldn't turn into a refactor.

> I think it is worth changing this from a `Deferred` to a `string`.

That makes sense. I'll ping you once it's done and ready for re-review.

sanmai:
The memory usage really went up since 0.29.

sanmai (approved):
Makes sense. There's no reason to defer this part as the original file content is "work already done". I think at some point we used to (and possibly still do) pretty-print the files on our end and it was an intense operation. Then it made sense to defer it.

 If you up to make it a string, go for it.

src/Mutant/MutantFactory.php
sanmai: tbh I'm not super fond of going all in on string replacement because that looks like a bc break and we will need a major version bump, conservatively  speaking; yes, we have ADR 0013 that says this should not be a bc break

sanmai:
Asked Codex to review other uses:

> I found no other unjustified production use of `Deferred`. The only stale use is the original-source-code chain addressed by PR #3506. The other uses defer real work, ntentionally memoize a computation, or consume/pass through a legitimately deferred value. `ProjectDirProvider` is the least obvious case, but it intentionally memoizes the first `getcwd()` result and retains no large object.

src/Mutant/MutantFactory.php
theofidry: @sanmai `Mutant` is not part of the public API. Custom mutators issue a `Mutation`, or am I missing something?

Meacue:
@theofidry @sanmai 

Since we're still discussing the current changes, the fix itself and how far the refactoring should go, I might keep this PR as is and create another PR that includes the 'Deferred to a string' refactoring.

So you can decide later which changes deserve merging now, and which ones aren't worth merging later or aren't worth merging at all.

Perhaps I should open a separate issue about the observed increased memory consumption from patch 0.29.14 to 0.35.2, so as not to lose the essence of the changes.

What do you think?

theofidry:
Thank you @Meacue

PR #3399, opened 2026-07-22
fix(git): Resolve diff paths from the configuration directory

theofidry:
## Description

Fix Git diff source filtering when Infection is installed in a repository subdirectory or invoked from outside the configured PHP project.

Consider a repository containing separate frontend and backend projects:

```
repository/
├── .git/
├── frontend/
└── server/
    ├── infection.json5
    ├── src/
    │   └── SourceClass.php
    └── vendor/
        └── bin/infection
```

The Infection configuration declares `src` as its source directory. This path is relative to `server/infection.json5`, regardless of where Infection is invoked.

Git interprets pathspecs from its working directory but, by default, reports matching paths relative to the repository root. These different reference directories caused two failure modes:

- When Infection ran from `server/`, Git reported `server/src/SourceClass.php`. Infection resolved that path from `server/`, producing the invalid path `server/server/src/SourceClass.php`.
- When Infection ran from the repository root or `frontend/`, Git interpreted the configured `src` path from that directory rather than from `server/`. It therefore searched the wrong source directory and could report no changed source files.

The Git diff contract now receives the configuration directory explicitly. `CommandLineGit` executes the diff with `git -C <configuration-directory>` and `--relative`. Git therefore interprets `src` from `server/` and reports `src/SourceClass.php`, independently of the process working directory. Changed-line matching resolves returned paths against the same directory.

## Changes

- Executes Git diff commands from the Infection configuration directory.
- Reports changed file paths relative to the configuration directory.
- Resolves changed-line paths against the same directory.
- Updates Git commands, callers, test doubles, and the Git diff benchmark to use the explicit working-directory argument.
- Adds an e2e scenario with a PHP project under `server/` and a sibling `frontend/` directory.

## Related issues

Fixes https://github.com/infection/infection/issues/3397.

PR #3516, opened 2026-08-27
fix(git): Fix GitDiffSourceLineMatcher compatibility with Windows

### LINKED ISSUE #3515: bug: --git-diff-lines generates zero mutants on Windows
| Question    | Answer
| Infection version | 0.35.2
| Test Framework version | PHPUnit 11.5.56
| CI | local only
| PHP version | 8.5.9
| Platform    | Windows 11
| Github Repo | -

On Windows, `--git-diff-lines` always generates **zero** mutants, whatever the diff contains. `master` still has the defect.

It fails silently. `NoSourceFound` is not thrown, because the *file* filter works; only the *line* filter matches nothing. The run finishes normally and reports `Total: 0`, which reads as "your diff is fully covered" rather than as a failure.

## Cause

`CommandLineGit` keys the changed-lines map with `Path::join()`, which always emits forward slashes:

```php
// src/Git/CommandLineGit.php:247
$result[Path::join($projectDirectory, $path)] = $changedLines;
```

`NodeTraverserFactory` passes `SplFileInfo::getRealPath()` into the matcher, which on Windows emits backslashes:

```php
// src/PhpParser/NodeTraverserFactory.php:95
new ExcludeUnchangedLinesVisitor($this->sourceLineMatcher, $sourceFile->getRealPath()),
```

`GitDiffSourceLineMatcher` then looks the path up as a raw array key, with no normalisation:

```php
// src/Source/Matcher/GitDiffSourceLineMatcher.php:90
return $this->memoizedFilesChangedLinesMap[$fileRealPath] ?? [];
```

Every lookup misses. `ExcludeUnchangedLinesVisitor` therefore marks every node ineligible, so nothing is mutated.

`RealPathFilterIterator` already normalises for exactly this reason, which is why the *file* filter works and the *line* filter does not:

```php
// src/FileSystem/Finder/Iterator/RealPathFilterIterator.php:55
$filename = $this->current()->getRealPath();

if (OperatingSystem::isWindows()) {
    $filename = Path::normalize((string) $filename);
}
```

## Reproduction

This script needs no Infection run. It builds the map and looks each file up both ways.

```php
$git = new CommandLineGit(new SymfonyProcessShellCommandRunner());
$map = $git->getChangedLinesRangesByFilePaths('AM', $base, ['app'], $projectDir);
$matcher = new GitDiffSourceLineMatcher($git, $base, 'AM', ['app'], $projectDir);

foreach (array_keys($map) as $key) {
    var_dump($matcher->touches(realpath($key), 1, 100000)); // false on Windows
    var_dump($matcher->touches($key, 1, 100000));           // true
}
```

Output on Windows, for a four-file diff:

```
--- keys produced by CommandLineGit ---
  C:/projects/app-under-test/app/Http/Controllers/EventController.php
  C:/projects/app-under-test/app/Mail/GuestRemovedMail.php
  ...

--- lookups with SplFileInfo::getRealPath() ---
  EventController.php     realpath:false   forward-slash:true
  GuestRemovedMail.php    realpath:false   forward-slash:true
  ...

changed files: 4, matched via realpath: 0
```

A control confirms the rest of the toolchain is fine: the same four files under a plain source filter generate 926 mutants and kill 643.

## Suggested fix

Normalise in `GitDiffSourceLineMatcher::getChangedLinesRanges()`, the same way `RealPathFilterIterator` does:

```php
if (OperatingSystem::isWindows()) {
    $fileRealPath = Path::normalize($fileRealPath);
}

return $this->memoizedFilesChangedLinesMap[$fileRealPath] ?? [];
```

I am running this locally as a Composer patch and it restores normal line-level results. Happy to open a PR if the approach looks right.

<details>

 ```json5
{
    $schema: "vendor/infection/infection/resources/schema.json",
    source: {
        directories: ["app"]
    },
    logs: {
        text: "storage/infection/infection.log",
        summary: "storage/infection/summary.log",
        json: "storage/infection/infection.json"
    },
    tmpDir: "storage/infection/tmp",
    timeout: 240,
    phpUnit: {
        customPath: "vendor/phpunit/phpunit/phpunit"
    },
    mutators: {
        "@default": true,
        ArrayItemRemoval: {
            settings: {
                remove: "all"
            }
        }
    }
}
 ```
</details>

 ```shell
vendor/bin/infection --threads=4 --git-diff-lines --git-diff-base=origin/main
 ```

theofidry:
## Description

Since #3399, the paths returned by `Git::getChangedFilePaths()` have been absolute, so `GitDiffSourceLineMatcher` no longer needs to resolve them. Previously, `GitDiffSourceLineMatcher` used [`Filesystem::realpath()`](https://github.com/infection/infection/pull/3399/changes#diff-fff620a2b732d1185a52cab4447c54831ceaec8314672acfa56ee6b3ca48e9c7L104), which returns paths with native separators on Windows. By contrast, `Git::getChangedFilePaths()` [uses `Path::join()`](https://github.com/infection/infection/pull/3399/changes#diff-ac9ebefbc5789e378dc66a100cd4834967896ce96f73902d3ff656a0f1789171R247), which returns normalised paths with forward slashes.

On Windows, this mismatch prevents `GitDiffSourceLineMatcher` from matching any changed lines. `ExcludeUnchangedLinesVisitor` therefore marks every node as ineligible, and no mutations are generated.

## Changes

- Normalise the real file path in `GitDiffSourceLineMatcher` when running on Windows.

## Related issues

Fixes https://github.com/infection/infection/issues/3515.

### DIFF
--- a/src/Source/Matcher/GitDiffSourceLineMatcher.php
+++ b/src/Source/Matcher/GitDiffSourceLineMatcher.php
@@ -36,8 +36,10 @@ declare(strict_types=1);
 namespace Infection\Source\Matcher;
 
 use Infection\Differ\ChangedLinesRange;
+use Infection\Framework\OperatingSystem;
 use Infection\Git\Git;
 use Infection\Source\Exception\NoSourceFound;
+use Symfony\Component\Filesystem\Path;
 
 /**
  * @internal
@@ -86,6 +88,10 @@ final class GitDiffSourceLineMatcher implements SourceLineMatcher
             $this->sourceDirectories,
             $this->workingDirectory,
         );
+
+        if (OperatingSystem::isWindows()) {
+            $fileRealPath = Path::normalize($fileRealPath);
+        }
 
         return $this->memoizedFilesChangedLinesMap[$fileRealPath] ?? [];
     }
--- a/tests/phpunit/Source/Matcher/GitDiffSourceLineMatcherTest.php
+++ b/tests/phpunit/Source/Matcher/GitDiffSourceLineMatcherTest.php
@@ -35,6 +35,7 @@ declare(strict_types=1);
 
 namespace Infection\Tests\Source\Matcher;
 
+use const DIRECTORY_SEPARATOR;
 use Infection\Differ\ChangedLinesRange;
 use Infection\Git\Git;
 use Infection\Source\Matcher\GitDiffSourceLineMatcher;
@@ -44,6 +45,7 @@ use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 use function sprintf;
+use function str_replace;
 
 #[AllowMockObjectsWithoutExpectations]
 #[CoversClass(GitDiffSourceLineMatcher::class)]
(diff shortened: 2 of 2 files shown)


maks-rafalko (approved):

maks-rafalko:
thank you @theofidry

PR #3518, opened 2026-08-27
feat(test-framework): Introduce TestFrameworkFactory

theofidry:
## Description

Adds `TestFrameworkFactory`, the analogue of `TestFrameworkAdapterFactory` but for `TestFrameworkFactory`. For now, its API is indentical – except for `TestFrameworkAdapter` -> `TestFramework`. Adapting it will be done in future PRs.

To achieve this, there is little thing I am not too happy about which is `LegacyAdapterFactory`. Currently as we need `TestFrameworkAdapter` injected in places, we still need to instantiate a `TestFrameworkAdapter`. Having `LegacyTestFrameworkBridge` exposing its adapter is insufficient as we would still end up with a circular dependency.

In practice however, the adapters are stateless, so I do not expect it to cause issues. Removing references to `TestFrameworkAdapter` will be the next PR too so this hack should not live for long.

## Changes

- Introduced the `TestFrameworkFactory` contract for native test framework creation.
- Added `LegacyAdapterFactory` to isolate legacy adapter resolution, so that services requiring still work.
- Updated `Factory` and container registration to support native frameworks and legacy adapters.
- Updated PHPUnit integration to implement `TestFrameworkFactory`.
- Added support for extension-provided native and legacy factories.

## Related issues

- #2863


theofidry:
@coderabbitai full review

PR #3519, opened 2026-08-28
test(test-framework): Add basic tests for the test framework `Factory`

theofidry:
## Description

Adds basic tests for `Factory`. In #3518, we will introduced the `TestFrameworkFactory` interface which is the analogue of `TestFrameworkAdapterFactory` for `TestFramework`.

## Changes

- Introduce `ConfigurableTestFrameworkAdapterFactory` which is a dummy implementation of `TestFrameworkAdapterFactory`. A particularity of `Factory` is that it relies on invoking the adapter factory statically, so it requires a tiny bit of plumbing.
- Add `FactoryTest`.

## Related issues


devTools/mago-baseline.toml
```diff
@@ -4476,12 +4476,6 @@ code = "non-existent-class"
 message = 'Class `Infection\Tests\Fixtures\Console\FakeOutput` not found.'
 count = 1
 
-[[issues]]
```
theofidry: ha, seems unrelated, I probably forgot to update once more the baseline in #3513. I should look for a way to make mago fail upon a non-synced baseline

theofidry:
For some reasons I thought `FactoryTest` in https://github.com/infection/infection/pull/3518 was new, so this work may not be needed after all.

PR #3524, opened 2026-08-29
fix(test-framework): Avoid mutating PATH during lookup

### LINKED ISSUE #3423: bug: TestFrameworkFinder can break tools resolution
## Context

`TestFrameworkFinder` searches for a test-framework executable using a project-local-first policy. It prepends Composer's configured binary directory to the Infection process's `PATH` so that, for example, `vendor/bin/phpunit` takes precedence over `/usr/bin/phpunit`. This is a pragmatic use of Symfony's `ExecutableFinder`.

## Problem

The search has a process-wide side effect: `TestFrameworkFinder` permanently mutates `PATH`. Consequently, subsequent executable lookups use a different search order, which child processes inherit unless their environment overrides it.

This affects more than the requested test framework. For example, if a project contains `vendor/bin/git`, a later lookup may resolve that executable instead of the Git executable selected through the user's original environment, such as `/opt/homebrew/bin/git`.

This behaviour arose while testing a custom test framework in a private project. Although a project-local executable is unlikely to shadow a system tool, the effect is not limited to Git and depends on the contents of `vendor/bin`. An operation intended only to locate a test framework can therefore change the behaviour of otherwise unrelated commands.

The mutation also weakens reproducibility: command resolution depends on whether `TestFrameworkFinder::find()` has already run in the current process, rather than solely on the configured environment and the command being resolved.

I want to note that this is very unlikely to affect people in the wild. It is worth fixing if requires low efforts, but not a high priority.

## Desired outcome

Finding a project-local test-framework executable should retain the existing lookup precedence without modifying the process environment used by unrelated executable lookups or child processes.

davesheffer:
## Description

`TestFrameworkFinder` currently prepends Composer's bin directory to the process-wide `PATH`. That lookup side effect leaks into later executable resolution and child processes.

This keeps Composer-bin-first resolution for test-framework candidates while leaving the process environment unchanged.

## Changes

- Resolve every executable candidate in Composer's configured bin directory before consulting the original `PATH`.
- Apply Windows `PATHEXT` suffixes inside Composer's bin directory, including the normal fallback suffixes when `PATHEXT` is absent.
- Preserve Windows `.bat` handling and Unix non-executable fallback precedence.
- Make directory precedence explicit on Windows: an extensionless Composer-bin proxy now wins over a `PATH` `.bat` proxy.
- Inject Composer-bin resolution into `TestFrameworkFinder` and use Infection's filesystem abstraction for existence checks.
- Add focused regressions for `PATH` stability, Windows `.cmd`/`.exe` lookup, Windows directory precedence, and Unix non-executable fallback.
- Remove the Mago baseline entry made obsolete by deleting the unsafe environment concatenation.

## Checklist

- [x] Tests added/updated
- [ ] Documentation updated (not applicable; internal bug fix)
- [ ] CHANGELOG.md updated (not applicable; no deprecation or breaking change)
- [ ] Appropriate labels applied (maintainers can apply `bug` if desired)

## Reviewer Notes

The analogous `StaticAnalysisToolExecutableFinder` still mutates `PATH`. It is intentionally left out so this change stays scoped to issue #3423.

Local verification:

- `make cs`
- PHPStan, Mago, Composer validation, Rector, collision detection, and ADR-list checks passed
- autoreview suite: 1,762 tests / 1,722 assertions
- default unit suite: 4,837 tests / 13,347 assertions; 4 environment-dependent skips
- integration suite: 1,950 tests / 3,188 assertions; 10 environment/platform-dependent skips
- targeted finder tests: 20 tests / 42 assertions
- changed-line Infection run: 47/47 mutants killed, 100% mutation code coverage and 100% covered-code MSI
- native zizmor 1.29.0: no findings

## Related issues

Fixes https://github.com/infection/infection/issues/3423.

### DIFF
--- a/src/FileSystem/Finder/ComposerBinExecutableFinder.php
+++ b/src/FileSystem/Finder/ComposerBinExecutableFinder.php
@@ -1,0 +1,87 @@
+<?php
+/**
+ * This code is licensed under the BSD 3-Clause License.
+ *
+ * Copyright (c) 2017, Maks Rafalko
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of the copyright holder nor the names of its
+ *   contributors may be used to endorse or promote products derived from
+ *   this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+declare(strict_types=1);
+
+namespace Infection\FileSystem\Finder;
+
+use function explode;
+use Infection\CannotBeInstantiated;
+use function is_executable;
+use function is_file;
+
+/**
+ * @internal
+ */
+final class ComposerBinExecutableFinder
+{
+    use CannotBeInstantiated;
+
+    private const string WINDOWS = 'Windows';
+
+    private const string WINDOWS_PATH_SEPARATOR = ';';
+
+    private const array DEFAULT_WINDOWS_EXECUTABLE_SUFFIXES = ['.exe', '.bat', '.cmd', '.com'];
+
+    /**
+     * @param list<string> $candidates
+     */
+    public static function find(
+        array $candidates,
+        string $composerBinDir,
+        string $operatingSystemFamily,
+        string|false $pathExtension,
+    ): ?string {
+        $windowsSuffixes = $pathExtension === false || $pathExtension === ''
+            ? self::DEFAULT_WINDOWS_EXECUTABLE_SUFFIXES
+            : explode(self::WINDOWS_PATH_SEPARATOR, $pathExtension);
+        $suffixes = $operatingSystemFamily === self::WINDOWS ? $windowsSuffixes : [];
+
+        // The caller searches PATH only after this directory. Checking every local candidate
+        // here prevents a PATH .bat from outranking an extensionless Composer proxy.
+        foreach ($candidates as $name) {
+            foreach (['', ...$suffixes] as $suffix) {
+                $composerCandidate = $composerBinDir . '/' . $name . $suffix;
+
+                if (
+                    is_file($composerCandidate)
+                    && ($operatingSystemFamily === self::WINDOWS || is_executable($composerCandidate))
+                ) {
+                    return $composerCandidate;
+                }
+            }
+        }
+
+        return null;
+    }
+}
--- a/src/FileSystem/Finder/ComposerBinExecutableFinder.php
+++ b/src/FileSystem/Finder/ComposerBinExecutableFinder.php
@@ -36,17 +36,15 @@ declare(strict_types=1);
 namespace Infection\FileSystem\Finder;
 
 use function explode;
-use Infection\CannotBeInstantiated;
 use function is_executable;
 use function is_file;
 
 /**
  * @internal
+ * @final
  */
-final class ComposerBinExecutableFinder
+class ComposerBinExecutableFinder
 {
-    use CannotBeInstantiated;
-
     private const string WINDOWS = 'Windows';
 
     private const string WINDOWS_PATH_SEPARATOR = ';';
@@ -56,21 +54,24 @@ final class ComposerBinExecutableFinder
     /**
      * @param list<string> $candidates
      */
-    public static function find(
+    public function find(
         array $candidates,
         string $composerBinDir,
         string $operatingSystemFamily,
         string|false $pathExtension,
     ): ?string {
-        $windowsSuffixes = $pathExtension === false || $pathExtension === ''
-            ? self::DEFAULT_WINDOWS_EXECUTABLE_SUFFIXES
-            : explode(self::WINDOWS_PATH_SEPARATOR, $pathExtension);
-        $suffixes = $operatingSystemFamily === self::WINDOWS ? $windowsSuffixes : [];
+        $suffixes = [''];
 
-        // The caller searches PATH only after this directory. Checking every local candidate
-        // here prevents a PATH .bat from outranking an extensionless Composer proxy.
+        if ($operatingSystemFamily === self::WINDOWS) {
+            $windowsSuffixes = $pathExtension === false || $pathExtension === ''
+                ? self::DEFAULT_WINDOWS_EXECUTABLE_SUFFIXES
+                : explode(self::WINDOWS_PATH_SEPARATOR, $pathExtension);
+
+            $suffixes = ['', ...$windowsSuffixes];
+        }
+
         foreach ($candidates as $name) {
-            foreach (['', ...$suffixes] as $suffix) {
+            foreach ($suffixes as $suffix) {
                 $composerCandidate = $composerBinDir . '/' . $name . $suffix;
 
                 if (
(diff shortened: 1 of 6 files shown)


src/FileSystem/Finder/TestFrameworkFinder.php
```diff
@@ -153,13 +148,34 @@ private function findTestFramework(string $testFrameworkName, string $customPath
         $extraDirs = [$cwd, $cwd . '/bin'];
 
         foreach ($candidates as $name) {
+            if ($composerBinDir !== null) {
+                $composerCandidate = $composerBinDir . '/' . $name;
+
+                if (
+                    is_file($composerCandidate)
+                    && (DIRECTORY_SEPARATOR === '\\' || is_executable($composerCandidate))
```
theofidry:
```suggestion
                    // On Windows, executability is instead
  determined by filename extensions such as .exe, .bat, and .cmd
                    && (!OperatingSystem::isWindows() || is_executable($composerCandidate))
```

src/FileSystem/Finder/TestFrameworkFinder.php
```diff
@@ -153,13 +148,34 @@ private function findTestFramework(string $testFrameworkName, string $customPath
         $extraDirs = [$cwd, $cwd . '/bin'];
 
         foreach ($candidates as $name) {
+            if ($composerBinDir !== null) {
+                $composerCandidate = $composerBinDir . '/' . $name;
```
theofidry: Thought something like `is_file('C:\\project\\vendor\\bin/phpunit.bat')` works, I believe this would show the intend more clearly:

```suggestion
                $composerCandidate = $composerBinDir . DIRECTORY_SEPARATOR . $name;
```

Likewise L171

src/FileSystem/Finder/TestFrameworkFinder.php
```diff
@@ -153,13 +148,34 @@ private function findTestFramework(string $testFrameworkName, string $customPath
...
+
+                if (
+                    is_file($composerCandidate)
+                    && (DIRECTORY_SEPARATOR === '\\' || is_executable($composerCandidate))
+                ) {
+                    return $composerCandidate;
```
theofidry: one major difference here is that we no longer call `$path = $finder->find($name, null, $extraDirs);
`.

So before this patch, we added `/path/to/vendor/bin` to `PATH` and executed `$finder->find('phpunit');`, this meant we could find `vendor/bin/phpunit.exe` or `vendor/bin/phpunit.cmd.`. The new direct lookup checks only the explicit candidates:

- phpunit.bat
- phpunit
- phpunit.phar

But maybe it is acceptable as Composer should generate a `.bat` on Windows, at least if the executable is properly configured as such in the package.

If you have an easy fix and way to test it, it would be best to do it, but we can leave it alone otherwise.

theofidry (commented):
> Resolve executable candidates in Composer's configured bin directory before consulting the original PATH.

This is more of an FYI to any reader: this is the expected behaviour from `ExecutableFinder` hence we follow this here.

<hr/>

In the tests, now the `PATH` is expected to remain unchanged but we do not check this at any point, maybe it is worth doing so? I would normally say that testing this code does not do is non-sense, but in this case we use third-party code which we have little control over.

Additional tests that I think are worthwhile:

1. That we apply the Windows path extensions in the Composer bin directory (Windows only) – if we decide to handle this case; otherwise I would still add it to show that it does not work:

  - Composer bin contains only `phpunit.cmd` or `phpunit.exe`.
  - `PATHEXT` contains the corresponding extension.
  - Composer bin is absent from `PATH`.
  - Assert that the Composer-bin executable is found.
  - Assert that `PATH` remains unchanged.

This should fail against the current PR and exposes the `ExecutableFinder` compatibility (acceptable) regression.

2. That candidate precedence across the Composer bin directory and `PATH` is preserved on Window:

  - Composer bin contains `phpunit`.
  - `PATH` contains `phpunit.bat`.
  - Assert which one should win.

I think before `PATH/phpunit.bat` would win, because the candidates priority was higher than the directory priority, but I think that was a mistake. So if the behaviour changes here, I would document that change of behaviour however much of an edge case it is.

3. Non-executable Composer candidates are prioritised over non-executable `PATH` candidates (Unix-only):

  - Composer bin contains a non-executable `phpunit`.
  - `PATH` contains another non-executable `phpunit`.
  - No executable candidate exists.
  - Assert that the Composer-bin file is returned through the non-executable fallback.
  - Assert that PATH remains unchanged.

This locks down the fallback precedence previously obtained by prepending Composer bin to `PATH`.

<hr/>

Thanks a lot of the PR, I realise it's a lot to ask so if you feel uncomfortable going through with those changes I can submit a patch over your PR.

Out of curiosity did you encounter any actual issue or was it just a low-handing fruit you could grab?

davesheffer:
Thanks for the thorough review. I added all three cases:

- Composer-bin lookup now honors Windows `PATHEXT`, with regression coverage for `.cmd` and `.exe`.
- The whole Composer bin directory is searched before `PATH`, so an extensionless local proxy wins over a `phpunit.bat` found on `PATH`. The PR now calls out that behavior explicitly.
- On Unix, a non-executable Composer-bin candidate wins over a non-executable `PATH` candidate.

The original regression tests already checked that `PATH` stays unchanged, and the new cases check it too.

I did not hit this in production myself. I chose the existing #3423 report for a preregistered blind Hunch evaluation. That evaluation did not show a measured Hunch lift, but it did lead to this real fix.

The changed-line mutation run kills 43/43 mutants, and all 61 CI checks are green, including the native Windows job.

src/FileSystem/Finder/ComposerBinExecutableFinder.php
```diff
@@ -0,0 +1,87 @@
...
+        string $operatingSystemFamily,
+        string|false $pathExtension,
+    ): ?string {
+        $windowsSuffixes = $pathExtension === false || $pathExtension === ''
+            ? self::DEFAULT_WINDOWS_EXECUTABLE_SUFFIXES
+            : explode(self::WINDOWS_PATH_SEPARATOR, $pathExtension);
+        $suffixes = $operatingSystemFamily === self::WINDOWS ? $windowsSuffixes : [];
+
+        // The caller searches PATH only after this directory. Checking every local candidate
+        // here prevents a PATH .bat from outranking an extensionless Composer proxy.
```
sanmai: What does it even say? I understand this is claudese, but the meaning escapes me. We could just as well strip it.

src/FileSystem/Finder/ComposerBinExecutableFinder.php
```diff
@@ -0,0 +1,87 @@
...
+        foreach ($candidates as $name) {
+            foreach (['', ...$suffixes] as $suffix) {
```
sanmai: we never use $suffixes without this addition so we can prepend the empty suffix once and be done with it

tests/phpunit/FileSystem/Finder/TestFrameworkFinderTest.php
```diff
@@ -122,33 +119,150 @@ public function test_invalid_custom_path_throws_exception(): void
...
+
+        $path = self::getPath();
+
+        $frameworkFinder = $this->createFinderWithComposerBin($composerBinDir);
+
+        $this->assertSame(
+            Path::normalize($composerExecutable),
+            Path::normalize($frameworkFinder->find(TestFrameworkTypes::PHPUNIT)),
+        );
+        $this->assertSame($path, self::getPath());
```
sanmai: Much of these tests can be avoided if `ComposerBinExecutableFinder` was a injectable dependency either here or at `ComposerExecutableFinder`

src/FileSystem/Finder/TestFrameworkFinder.php
```diff
@@ -160,6 +167,16 @@ private function findTestFramework(string $testFrameworkName, string $customPath
             }
         }
 
+        if ($composerBinDir !== null) {
+            foreach ($candidates as $name) {
+                $composerCandidate = $composerBinDir . '/' . $name;
+
+                if (file_exists($composerCandidate)) {
```
sanmai: `Infection\FileSystem\FileSystem` exposes `exists()` so we can eliminate direct file operations here and end up with pure unit tests

sanmai (changes requested):
Try starting a new Claude session and tell it that "I did this and it does not look great, what can we improve? just look at the review comment"

Do not say "you did it" because Claude will get defensive and you get nothing done. Blame yourself, that costs nothing and gets it moving.

src/FileSystem/Finder/ComposerBinExecutableFinder.php
davesheffer: Agreed. I removed the comment.

src/FileSystem/Finder/ComposerBinExecutableFinder.php
davesheffer: Done. The suffix list now starts with the empty suffix, and the Windows extensions are appended once.

tests/phpunit/FileSystem/Finder/TestFrameworkFinderTest.php
davesheffer: Done. `ComposerBinExecutableFinder` is now an injected instance. `TestFrameworkFinder` tests can stub that boundary, while the platform and permission behavior stays in the focused finder tests.

src/FileSystem/Finder/TestFrameworkFinder.php
davesheffer: Done. `TestFrameworkFinder` now receives Infection's `FileSystem` and uses `exists()` instead of direct existence checks. The fallback test mocks that boundary.

davesheffer:
Thanks. I went through the review from scratch and addressed all four points:

- Removed the unclear comment.
- Built the suffix list once with the empty suffix first.
- Made `ComposerBinExecutableFinder` an injected dependency.
- Replaced direct existence checks in `TestFrameworkFinder` with Infection's `FileSystem::exists()`.

I also tightened the tests around the injected boundary. The changed-line mutation run now kills 47/47 mutants, and all 61 CI checks are green, including Windows.

This review is also useful input for Hunch. It can record the dependency-injection and filesystem-abstraction points as repository-scoped architectural constraints, so future work on these finder files receives them before editing. The suffix and comment points fit better as advisory review heuristics: build candidate lists once, and keep a comment only when it adds a concrete reason.

These lessons came from this review after the blind run, so I will record them as review-derived learning and will not count them as evidence that Hunch found them during the original experiment.

davesheffer:
Rebased on master now that #3538 is in: `ConfigureCommand` passes `$this->fileSystem` to `TestFrameworkFinder`, and the two new mock chains in `TestFrameworkFinderTest` follow the semicolon rule from #3533. No behaviour change since the last round.

@sanmai this should address your review; I cannot re-request it from my side.

theofidry:
Thanks for the update. Apologies if this is taking some time. I did have a look but I feel like I need to have a more proper look at it: it is unfortunately not such an easy piece to assess in general.

davesheffer:
No worries at all—thanks for taking the time to review it properly. I appreciate it, and I’m happy to clarify anything that would help.

PR #3525, opened 2026-08-29
feat(test-framework): Route test framework operations through the new contract

theofidry:
## Description

This PR updates the code to use the new `TestFramework` API instead of `TestFrameworkAdapter`.

The immediate goal is to enable the introduction of `TestFrameworkFactory`. This requires removing `TestFrameworkAdapter` from the container. Otherwise, we would need either to create another test framework instance or to bridge the new API to the old one, which is not feasible.

## Changes

This PR:

- Moves coverage requirement checks and the standalone initial test run behind `TestFramework`.
- Exposes JUnit report support on the contract for coverage enrichment and progress logging. This is only to facilitate the migration.
- Makes the bridge responsible for the legacy adapter and passes it explicitly to the remaining legacy process factories.
- Removes superseded adapter, coverage checker, and mutant result factory services from the container.

This also includes the following **behavioural changes**:

- `initial-test:run` now validates generated coverage artifacts.
- Coverage requirement failures occur later during a normal run. In practice, this means the user can see the threat count message before the coverage failure.

## Breaking Changes

None.

## Reviewer Notes

In the proof of concept in #3290, I retained the second instance because it was simpler. Although this should be safe because
the adapters are effectively stateless, avoiding the additional instance provides a cleaner solution.

This work also suggests the following changes for later PRs:

- ~~Add PHPat rules that report uses of the legacy `TestFrameworkAdapter` and `TestFrameworkAdapterFactory` as deprecations. This would demonstrate how PRs such as this one resolve outstanding violations.~~ Done in https://github.com/infection/infection/pull/3528.

## Related issues

- #2863


src/TestFramework/Coverage/CoverageCheckerFactory.php
```diff
@@ -0,0 +1,68 @@
...
+
+use Infection\AbstractTestFramework\TestFrameworkAdapter;
+use Infection\Configuration\Configuration;
+use Infection\TestFramework\Coverage\JUnit\JUnitReportLocator;
+use Infection\TestFramework\Coverage\XmlReport\IndexXmlCoverageLocator;
+
+/**
+ * @internal
+ */
+final readonly class CoverageCheckerFactory
```
maks-rafalko, line 46: trying to understand why do we need this new factory?

src/Container/Container.php
```diff
@@ -663,15 +641,23 @@ static function () use ($container): SourceCollector {
...
-                $container->get(ConsoleOutput::class),
-                $container->getCoverageChecker(),
-                $container->getInitialTestsRunner(),
-                $container->getConfiguration(),
-                $container->getMutantProcessContainerFactory(),
-                $container->getTestFrameworkExtraOptionsFilter(),
-            ),
+            TestFramework::class => static function (self $container): TestFramework {
+                $config = $container->getConfiguration();
+                $adapter = $container->getFactory()->create(
```
maks-rafalko, line 650: note to myself: `$container->getFactory()` is too ambiguous (it returns `TestFramework\Factory`).

src/Process/Factory/InitialTestsRunProcessFactory.php
```diff
@@ -46,17 +46,13 @@
  */
 class InitialTestsRunProcessFactory
 {
-    public function __construct(
-        private readonly TestFrameworkAdapter $testFrameworkAdapter,
-    ) {
-    }
```
maks-rafalko, line 52: is it to be able to run it N times with different adapters, or why do we need this change?

src/Process/Factory/InitialTestsRunProcessFactory.php
```diff
@@ -46,17 +46,13 @@
...
-
     /**
      * Creates process with enabled debugger as test framework is going to use in the code coverage.
      *
      * @param string[] $phpExtraOptions
      */
     public function createProcess(
+        TestFrameworkAdapter $testFrameworkAdapter,
```
maks-rafalko, line 55: do I get it right that this code eventually will be rewritten since there won't be `TestFrameworkAdapter`?

src/Process/Runner/InitialTestsRunner.php
```diff
@@ -58,11 +59,13 @@ public function __construct(
      * @param string[] $phpExtraOptions
      */
     public function run(
+        TestFrameworkAdapter $testFrameworkAdapter,
```
maks-rafalko, line 62: for understanding, will it be replaced eventually with `TestFramework`, or rewritten? what is the idea?

src/TestFramework/LegacyTestFrameworkBridge.php
```diff
@@ -122,11 +125,17 @@ public function executeInitialRun(): InitialRunResults
...
+            $this->adapter,
             $mutant,
             $this->getFilteredExtraOptionsForMutant(),
         );
     }
 
+    public function hasJUnitReport(): bool
+    {
+        return $this->adapter->hasJUnitReport();
+    }
```
maks-rafalko, line 137: > Exposes JUnit report support on the contract for coverage enrichment and progress logging. This is only to facilitate the migration.

again, just for understanding - this eventually likely will be removed, right? As it sounds like test framework (its tracer) is responsible to extend traces with JUnit info if needed.

----

Since `Contracts` is explicitly shaped as a future standalone package (per `AdapterExtractionBoundariesTest::testTestFrameworkContractsStayWithinFuturePackageBoundary`), a bare `hasJUnitReport(): bool` sitting at the end of the interface reads as permanent contract surface - a PHPStan or Psalm "test framework" has no JUnit report to speak of.

`LegacyTestFrameworkBridge` carries a `@deprecated ... To be removed.` block. Could this method carry the same marker while it exists, so the next reader of the contract knows it is scaffolding rather than design?

No PHPat violation either way - the return type is a scalar, so the package boundary holds.

_Generated as part of a joint review by Maks Rafalko and claude._

src/Command/RunCommand.php
```diff
@@ -577,7 +579,7 @@ private function startUp(
             $consoleOutput->logNotInControlOfExitCodes();
         }
 
-        $container->getCoverageChecker()->checkCoverageRequirements();
+        $container->getTestFramework()->checkRequirements();
```
maks-rafalko: `checkRequirements()` now runs twice per invocation: here, and again in `Engine::runInitialTestSuite()` (`src/Engine.php:129`).

Before this PR the two call sites did different work - `startUp()` called `checkCoverageRequirements()`, `Engine` called `checkRequirements()`. Moving `checkCoverageRequirements()` into `checkRequirements()` (`LegacyTestFrameworkBridge.php:84`) collapsed them into the same call.

The method is not idempotent: it calls `$this->consoleOutput->logSkippingInitialTests()`, which has no dedup guard. Running `tests/e2e/Skip_Initial_Tests` on both branches:

```
master:
  [notice]  Running Infection with 1 thread.
  [warning] Skipping the initial test run can be very dangerous.   <- once

this branch:
  [warning] Skipping the initial test run can be very dangerous.   <- before the thread notice
  [notice]  Running Infection with 1 thread.
  [warning] Skipping the initial test run can be very dangerous.   <- and again
```

`checkCoverageExists()` and `checkCoverageRequirements()` also re-run, which is wasted filesystem work but otherwise harmless.

One trade-off worth naming explicitly rather than absorbing silently: on `master` the coverage requirement check ran *here*, before `mkdir($config->tmpDir)`, subscriber registration and `ApplicationExecutionWasStarted`. Moving it wholly into `Engine` means those side effects now happen before a missing-coverage-driver error surfaces. That seems acceptable, but it is a deliberate choice.

Note the e2e scenario diffs `infection.log`, not stdout, so nothing in CI catches this.

_Generated as part of a joint review by Maks Rafalko and claude._

src/Command/InitialTest/InitialTestRunCommand.php
```diff
@@ -95,52 +92,11 @@ protected function executeCommand(IO $io): bool
...
-                $configuration->skipCoverage,
-            );
-
-        if (!$initialTestSuiteProcess->isSuccessful()) {
-            throw InitialTestsFailed::fromProcessAndAdapter(
-                $initialTestSuiteProcess,
-                $container->getTestFrameworkAdapter(),
-            );
-        }
+        $container->getTestFramework()->executeInitialRun();
```
maks-rafalko, line 115: This drops the `Command executed: <argv>` output. Verified against `tests/e2e/Skip_Initial_Tests` (run as `initial-test:run`):

```
master:
  Command executed:
  '.../vendor/bin/phpunit' '--configuration' '.../phpunitConfiguration.initial.infection.xml'
    '--coverage-xml=.../coverage-xml' '--log-junit=.../junit.xml'

  Running initial tests with PHPUnit version 9.6.36

this branch:
  Running initial tests with PHPUnit version 9.6.36
```

The test was updated to match, so I assume it is intentional - but the command's own description is *"Executes the initial test run as orchestrated by Infection with the debug mode enabled by default"*, and the exact argv Infection built is the single most useful thing it prints. It was also printed *before* the run, so a hanging or crashing suite still told you what had been executed; that case now prints nothing at all.

Could the output be preserved on top of it? Or why it's not needed anymore? IMO it's very useful

_Generated as part of a joint review by Maks Rafalko and claude._

src/TestFramework/LegacyTestFrameworkBridge.php
```diff
@@ -81,6 +81,8 @@ public function checkRequirements(): void
     {
         // TODO: check supported version
 
+        $this->coverageChecker->checkCoverageRequirements();
```
maks-rafalko: This new line has no assertion behind it.

Both `checkRequirements()` tests in `LegacyTestFrameworkBridgeTest` build `$coverageChecker` with `createMock()` and set expectations only on `checkCoverageExists()`, so an unexpected — or a missing — `checkCoverageRequirements()` call passes silently either way.

Given the house rule that mock expectations should be tight enough that mutants die, adding `->expects($this->once())->method('checkCoverageRequirements')` to both existing tests would pin it.

`hasJUnitReport()` (line 134) likewise has no test.

_Generated as part of a joint review by Maks Rafalko and claude._

src/TestFramework/Contracts/TestFramework.php
```diff
@@ -88,4 +88,6 @@ public function executeInitialRun(): InitialRunResults;
      * is interpreted.
      */
     public function test(Mutant $mutant): MutantProcessContainer;
+
+    public function hasJUnitReport(): bool;
```
maks-rafalko: Non-blocking, and adjacent to the question I left on `LegacyTestFrameworkBridge` — this is just the concrete follow-up.

Since `Contracts` is explicitly shaped as a future standalone package (per `AdapterExtractionBoundariesTest::testTestFrameworkContractsStayWithinFuturePackageBoundary`), a bare `hasJUnitReport(): bool` sitting at the end of the interface reads as permanent contract surface — a PHPStan or Psalm "test framework" has no JUnit report to speak of.

`LegacyTestFrameworkBridge` carries a `@deprecated ... To be removed.` block. Could this method carry the same marker while it exists, so the next reader of the contract knows it is scaffolding rather than design?

No PHPat violation either way — the return type is a scalar, so the package boundary holds.

_Generated as part of a joint review by Maks Rafalko and claude._

src/Process/Factory/MutantProcessContainerFactory.php
```diff
@@ -93,7 +94,7 @@ public function create(Mutant $mutant, string $testFrameworkExtraOptions = ''):
             new MutantProcess(
                 $process,
                 $mutant,
-                $this->mutantExecutionResultFactory,
+                new TestFrameworkMutantExecutionResultFactory($testFrameworkAdapter),
```
maks-rafalko, line 97: The allocation is now per-mutant rather than per-run. It is a small readonly object, so the cost is modest - but this is the hottest path in the application and the factory is invariant for the whole run - why do we need this change?

_Generated as part of a joint review by Maks Rafalko and claude._

maks-rafalko (commented):
makes sense in general. Found a couple of blockers though, and majority of the comments are questions to understand the future plans.

src/TestFramework/Coverage/CoverageCheckerFactory.php
theofidry: `CoverageChecker` depends on `TestFrameworkAdapter` (for its name and whether it has a JUnit support).

We can migrate it to `TestFramework` – and will, but we need to keep such a copy for the non-migrated `TestFrameworkAdapter` to preserve their behaviour.

src/Container/Container.php
theofidry: I'll add a todo for now. I wouldn't change it right this instant since out of scope of the PR and to avoid further conflicts, but we can do so at some point.

src/Process/Factory/InitialTestsRunProcessFactory.php
theofidry: No it is purely to be able to keep an implementation that depends on `TestFrameworkAdapter` around for the existing adapters.

Maybe to re-explain it a bit as it's easy to loose track of the current state. We need to be able to:

- Adapt the domain code to no longer use `TestFrameworkAdapter` and use `TestFramework` instead.
- Migrate the adapters living in `infection/infection` to `TestFramework`.

But because we don't want to break compatibility with the existing external adapters which at the moment are Testo, PhpSpec and Codeception, we still need to keep all the code that depends on `TestFrameworkAdapter` somehow.

So if we are to keep a service such as `CoverageChecker` in the new adapter, we will also have a copy using the old adapter (which I will move to the Legacy directory in a sub-sequent PR for clarity).

Maybe an un-phrased question you had is why do we need to do this. From the description:

>The immediate goal is to enable the introduction of TestFrameworkFactory. This requires removing TestFrameworkAdapter from the container. Otherwise, we would need either to create another test framework instance or to bridge the new API to the old one, which is not feasible.

Reading it again I appreciate it does a poor job at going into the details, I can roughly try to answer this.

In the POC, I had services depending on both the legacy and new API, for the sake of the example let's take `CoverageChecker` (depending on the new API) and `LegacyCoverageChecker` (depending on the old one). If we keep things in the `Container` (which I believe is less overall changes), then you need to be able to do something like:

```php
            CoverageChecker::class => static function (self $container): CoverageChecker {

                // The problematic statement
                $testFrameworkAdapter = $container->getTestFrameworkAdapter();

                return new CoverageChecker(...);
            },
```

`$testFrameworkAdapter = $container->getTestFrameworkAdapter()` is a problematic statement because it means, whenever `CoverageChecker` is injected, hence instantiated, we need a `TestFrameworkAdapter` instance around. Currently, it's not a problem, but once `PHPUnitAdapter` is migrated to `TestFramework` without using the legacy bridge, there is just no `TestFrameworkAdapter` instance to have anymore.

In the POC I solved it by having another `Factory` creating the legacy adapter which mean in practice, for PHPUnit, we always had 2 adapter instances around (one legacy and one new) and whilst it was fine for the POC, I didn't really like the idea either.

So this PR is a different direction: we make the adapter injectable at runtime so that we can remove `TestFrameworkAdapter` from the `Container`. This is purely for those services though, I am not suggesting that all services using a `TestFramework` instance must have it as a method parameter instead of injected in the constructor.

src/Process/Factory/InitialTestsRunProcessFactory.php
theofidry: I hope the previous replies clarified that

src/TestFramework/LegacyTestFrameworkBridge.php
theofidry: Correct. It's purely to reduce the scope of the changes, to not change everything at once as it's otherwise too easily overwhelming and easy to miss things. In the POC it mattered less, but this code otherwise needs to not introduce BC breaks or undesired changes at runtime so we need to be careful about it

src/Command/RunCommand.php
theofidry: Good catch, forgot about `RunCommand` in this PR

src/Process/Factory/MutantProcessContainerFactory.php
theofidry: I hope the previous answers helped.

To note that we will mitigate this tiny performance degradation for adapters within `infection/infection` very soon – so only third-party adapters will remain affected for a while.

src/TestFramework/Coverage/CoverageCheckerFactory.php
maks-rafalko: still not sure I got you. I mean, why do we need a factory class in the first place?

src/Process/Factory/InitialTestsRunProcessFactory.php
maks-rafalko: well, it's hard to understand it, tbh

> So this PR is a different direction: we make the adapter injectable at runtime so that we can remove TestFrameworkAdapter from the Container. This is purely for those services though, I am not suggesting that all services using a TestFramework instance must have it as a method parameter instead of injected in the constructor.

- where we will get it from, if container won't have it?
- if many places need `TestFrameworkAdapter`, does it mean we would create it multiple times? or how would/will it work?

src/TestFramework/LegacyTestFrameworkBridge.php
maks-rafalko: so do we need to mark such methods somehow so that in the future we now they are **not** needed?

src/Process/Factory/MutantProcessContainerFactory.php
maks-rafalko: > I hope the previous answers helped.

not in this case

> To note that we will mitigate this tiny performance degradation for adapters within infection/infection very soon – so only third-party adapters will remain affected for a while.

how? what does it mean?

src/TestFramework/Contracts/TestFramework.php
theofidry, line 95: missing `@deprecated`

src/TestFramework/LegacyTestFrameworkBridge.php
theofidry: I added a note to mark it as `@deprecated`, I forgot about it

src/Process/Factory/MutantProcessContainerFactory.php
theofidry: >how? what does it mean?

As we migrate the `PHPUnitAdapter` to `TestFramework`, it won't create a such a `TestFrameworkMutantExecutionResultFactory` service for every mutant process container.

I don't recall exactly the exact change done, but you can see a more final state in `PHPUnitAdapter::test()` in the POC

But for example for Codeception, which uses the legacy bridge even after we migrate PHPUnit, this (tiny) performance impact won't go away (until we migrate Codeception).

src/Process/Factory/InitialTestsRunProcessFactory.php
theofidry: >where we will get it from, if container won't have it?

It is observable from this PR: in the legacy bridge we have the old adapter instance and we pass it around from there.

>if many places need TestFrameworkAdapter, does it mean we would create it multiple times? or how would/will it work?

No, with this PR we create only one instance of `TestFrameworkAdapter` at most and pass it around, so we preserve the previous behaviour we had with the `Container` which was giving a shared service.

<hr/>

I did double check the POC just in case, so no surprise once the migration is done, `Container::getTestFrameworkAdapter()` is never called (unless you use a legacy adapter like Codeception or PhpSpec).

The trouble happens in-between: for a little while, you have services instantiated that require `TestFramework` and others `TestFrameworkAdapter`. This PR does two things:

1. Use `TestFramework` for the "domain" code (i.e. anything that is not specific to to the test framework. Notable, `CoverageChecker` and the `Tracer` _are_, so they are left alone here).
2. Find a solution so what we can have an instance of `TestFramework` and `TestFrameworkAdapter` at the same time*

*: this is necessary to do the migration progressively. Otherwise, it's too much change at once.

For 2. there is several solutions that I see:

- Create two test framework instances: one for the new API and one for the old. This is what I did in the POC by having another `Factory` (see the previous explanation).
- Instead of having `TestFrameworkAdapter` in the Container, services that require it take it as a runtime parameter instead of a construction parameter. This way, they can still live in the Container as usual, but don't require a `TestFrameworkAdapter` to be instantiated.

I want to re-iterate that it is a very artificial problem stemming from trying to migrate `PHPUnitAdapter` progressively. It is technically possible to do without, but I think it's too many changes at once.

If it's not clear I think a video call would be more appropriate because besides a demo or checking the code together I don't see much more of a written explanation

src/Command/InitialTest/InitialTestRunCommand.php
maks-rafalko: what about this one, are we going to fix it or accept?

src/Process/Factory/InitialTestsRunProcessFactory.php
maks-rafalko: > Instead of having TestFrameworkAdapter in the Container, services that require it take it as a runtime parameter instead of a construction parameter. This way, they can still live in the Container as usual, but don't require a TestFrameworkAdapter to be instantiated.

ok, I think I got it, but let's see in the next PRs, let me resolve this one for now.

Please rebase github.com/infection/infection/pull/3526, probably it will make it a bit more clear to follow looking into the next steps.

maks-rafalko (approved):
Other comments are non-blocking, so approving it. Thank you @theofidry

src/Command/InitialTest/InitialTestRunCommand.php
theofidry: Hm no I agree with you. Down the line it is the sort of things that I hope will be fixed by using the `Logger` that we can pass to the test frameworks, but right now it is a regression...

It's a bit ugly but I found a fix. I'm adding a TODO to make it clear this is not the final form.

src/TestFramework/Coverage/CoverageCheckerFactory.php
theofidry: Should be answered in that thread https://github.com/infection/infection/pull/3525#discussion_r3943419953

PR #3529, opened 2026-08-29
fix!(test-framework): Make the mutant memory limiting a test framework implementation detail

theofidry:
Refactor mutant memory limiting into the PHPUnit adapter

## Description

`MemoryUsageAware` allows a test-framework adapter to report the memory used by its initial test run. The engine then turns that value into a global limit for every mutant process, regardless of which tool runs it. This conflicts with static-analysis tools, which must opt out with `memory_limit=-1`, and prevents test frameworks from having independent limits.

This PR moves that behaviour into the PHPUnit adapter. After the initial PHPUnit run, the adapter records the reported memory usage and adds `-d memory_limit=<N>M` directly to PHPUnit mutant commands. As before, the limit is twice the observed usage, is applied only when that usage is known, and does not override a memory limit set by the user.

The memory limit does not need to be applied globally, as already demonstrated by static-analysis tools removing it from their own processes. This PR therefore moves that responsibility out of `Engine`. However, the new behaviour is not equivalent:

- On Windows, the executable is a wrapper (`phpunit.bat`) which drops the PHP `-d` arguments added by Infection, so mutant processes receive no automatic memory limit. This is a regression compared with changing `php.ini`. However, the previous implementation changed the temporary `php.ini` only when XdebugHandler had restarted Infection. With PCOV, for example, the memory limit was not applied to mutant processes.
- The current `TestFrameworkAdapter` API does not provide an easy way for third-party adapters to add these PHP arguments. As a result, third-party adapters can no longer use this feature. In practice, Codeception is the only existing integration affected. This also prevents test frameworks such as Testo from supporting it unless `TestFrameworkAdapter` is changed.

## Changes

- Move `MemoryLimiter` and `MemoryLimiterEnvironment` from the general resource namespace to the PHPUnit integration.
- Remove memory-limit orchestration and its container service from `Engine`.
- Record initial PHPUnit memory usage through `LegacyTestFrameworkBridge` while PHPUnit still uses the legacy adapter contract.
- Allow adapters to provide extra PHP arguments for mutant commands, and use this in PHPUnit to add its derived memory limit.
- Leave the shared `php.ini` unchanged so static-analysis and other follow-up processes retain their own memory configuration.
- Update unit and end-to-end coverage for PHPUnit memory limiting and the PHPStan inheritance regression.
- Exercise the PHPStan memory-limit regression with both PCOV and Xdebug instead of requiring the XdebugHandler temporary-ini path.

## Breaking Changes

- On Windows, PHPUnit is usually invoked through `phpunit.bat`, which drops the PHP `-d` arguments added by Infection. PHPUnit mutants therefore no longer receive an automatically derived limit there. Previously, this worked only when XdebugHandler had restarted Infection with a temporary `php.ini`; PCOV runs did not receive the limit.
- The engine previously allowed any third-party adapter implementing `MemoryUsageAware` to set the limit for all mutant processes. That path is removed. Codeception is the existing integration affected by this change. Adapters such as PhpSpec and Testo cannot opt into PHPUnit's new mechanism through the current `TestFrameworkAdapter` API.

## Reviewer Notes

The main point to review is that a limit derived from PHPUnit now applies only to PHPUnit mutant processes, rather than being written to shared PHP configuration. In particular, this comes with the Windows and third-party-adapter limitations described above.

For now, `LegacyTestFrameworkBridge` passes the initial memory usage to `PhpUnitAdapter`. This can be removed once PHPUnit has been migrated to the newer `TestFramework` contract.

## Related issues

Refs https://github.com/infection/infection/issues/2863.


theofidry:
@coderabbitai full review

theofidry:
@coderabbitai full review

PR #3535, opened 2026-08-30
chore: Fix a PHPStan deprecation

theofidry:


tests/Architecture/PHPat/Selector/Support/PHPUnitTestClassAnalysis.php
```diff
@@ -64,7 +64,7 @@ public static function isPHPUnitTestCase(ClassReflection $classReflection): bool
     {
         return str_ends_with($classReflection->getName(), 'Test')
             && ClassReflectionPredicates::isConcreteClass($classReflection)
-            && $classReflection->isSubclassOf(TestCase::class);
+            && $classReflection->getAncestorWithClassName(TestCase::class) !== null;
```
sanmai, line 67: The error suggested `isSubclassOfClass`

theofidry: It does but we have:

- `::getAncestorWithClassName(string $className): ?ClassReflection`
- `::isSubclassOfClass(ClassReflection $classReflection): bool`

So the 2nd requires a `ClassReflection`, which we do not have here. We could get one, but I don't think it would be more terse or expressive.

sanmai (approved):

PR #3536, opened 2026-08-30
doc: Add an ADR for naming exceptions

theofidry:


theofidry:
Putting this to not forget it but it's not high priority at all.

adr/0014-name-exceptions-after-domain-failures.md
```diff
@@ -0,0 +1,90 @@
...
+throws it. When one class would need a generic component name because it represents several
+unrelated failures, prefer separate, failure-specific exception classes. A single class may
+still expose several named constructors when they represent variants of the same failure.
+
+Give each exception a domain-specific concrete class, and extend the SPL exception type that
+best categorises the failure. For example, `InvalidSchema` extends
+`UnexpectedValueException`. The concrete name describes what failed; the parent type
+describes the general kind of failure.
+
+When callers need to catch several domain failures together, introduce a marker interface
```
sanmai, line 40: It looks like a leaky abstraction and if you have to click through to see what is caught, we could just as well catch the full list.

theofidry: I think that could be rephrased. The example is correct, but the conclusion is not.

The grouping of domain failures is there because for the caller it may be interesting to catch a kind of failure, but the actual details of the failure are not interesting and only relevant for, for example, reporting, communicating with the user. It has more to do with the domain modeling than a technical convenience.

PR #3537, opened 2026-08-30
cs: Add a Rector rule to enforce `@param` over `@var`

theofidry:
In PHP we now have the choice between:

```php
    public function __construct(
        /** @var list<string> Source paths. */
        public array $sourcePaths,
        /** @var list<string> */
        public array $testPaths,
    ) {
    }
```

And:

```php
    /**
     * @param list<string> $sourcePaths Source paths.
     * @param list<string> $testPaths
     */
    public function __construct(
        public array $sourcePaths,
        public array $testPaths,
    ) {
    }
```

I tend to prefer the latter simply because it means keeping one style does not mean to update all existing code. To try to solve this in an automatic way, this PR introduces a Rector rule.

That being said, I do not feel strongly about which form is preferred, I would simply like _one_ form and have a tool to automatically enforce it.

I did not bother creating an ADR because I don't think it is a worthwhile issue to capture in it.

### DIFF
--- a/mago.toml
+++ b/mago.toml
@@ -32,6 +32,7 @@ excludes = [
     "tests/phpunit/PhpParser/Visitor/EnrichmentTraverse/Fixtures",
     # Rector or PHPStan stubs can pollute Mago's Reflection inference
     # https://github.com/carthage-software/mago/issues/1676
+    "tests/phpunit/AutoReview/Rector",
     "vendor/rector/rector",
     "vendor/phpstan",
 ]
--- a/rector.php
+++ b/rector.php
@@ -33,6 +33,7 @@
 
 declare(strict_types=1);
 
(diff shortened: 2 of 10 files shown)


sanmai (approved):
If this works for PHPStan/Mago, that's much more readable

theofidry:
@sanmai both are well understood by PHPStan & Mago, so it is really a matter of what we prefer to read or have in a diff.

sanmai:
Let's ship this then.

maks-rafalko:
@theofidry out of interest, did you propose this to be in a rector core? Looks useful not only for us

theofidry:
@maks-rafalko I did not. Would be interesting to do indeed, but couldn't motivate myself to do so yesterday

PR #3538, opened 2026-08-30
chore: Replace direct file access with FileSystem abstraction

sanmai:
## Description

During the review of #3524 I noticed that we do not use the existing file system abstractions in a few places.

I left the changes to `TestFrameworkFinder` to introduce the same abstraction to PR #3524, so here we change everything else to use the abstraction.

## Reviewer Notes

Since the `FileSystem` class does not have a state, I chose to skip wiring it through the container to reduce the amount of scaffold for no gain.

The slight downside of this approach: `IoCodeDetectorVisitor` treats `new FileSystem()` as IO, so we can't untag the existing tests as integration, but they also do other direct file work - so we couldn't have untagged them anyway.

### DIFF
--- a/mago.toml
+++ b/mago.toml
@@ -63,6 +63,7 @@ unchecked-exceptions = [
     "PHPUnit\\Exception",
     "Psr\\Container\\ContainerExceptionInterface",
     "Safe\\Exceptions\\SafeExceptionInterface",
+    "Symfony\\Component\\Filesystem\\Exception\\ExceptionInterface",
 ]
 ignore = [
     { code = "unhandled-thrown-type", in = ["tests/"] },
--- a/src/Command/BaseCommand.php
+++ b/src/Command/BaseCommand.php
@@ -41,7 +41,6 @@ use Override;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
-use Throwable;
 use Webmozart\Assert\Assert;
 
 /**
@@ -73,9 +72,6 @@ abstract class BaseCommand extends Command
         $this->io = new IO($input, $output);
     }
 
-    /**
-     * @throws Throwable
-     */
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         return $this->executeCommand($this->getIO())
@@ -83,9 +79,6 @@ abstract class BaseCommand extends Command
             : self::FAILURE;
     }
 
-    /**
-     * @throws Throwable
-     */
     abstract protected function executeCommand(IO $io): bool;
 
     final protected function getIO(): IO
--- a/src/Command/BaseCommand.php
+++ b/src/Command/BaseCommand.php
@@ -41,6 +41,7 @@ use Override;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Throwable;
 use Webmozart\Assert\Assert;
 
 /**
@@ -72,6 +73,9 @@ abstract class BaseCommand extends Command
         $this->io = new IO($input, $output);
     }
 
+    /**
+     * @throws Throwable
+     */
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         return $this->executeCommand($this->getIO())
@@ -79,6 +83,9 @@ abstract class BaseCommand extends Command
             : self::FAILURE;
     }
 
+    /**
+     * @throws Throwable
+     */
     abstract protected function executeCommand(IO $io): bool;
 
     final protected function getIO(): IO
--- a/src/Command/ConfigureCommand.php
+++ b/src/Command/ConfigureCommand.php
@@ -66,7 +66,6 @@ use function str_starts_with;
 use Symfony\Component\Console\Helper\FormatterHelper;
 use Symfony\Component\Console\Helper\QuestionHelper;
 use Symfony\Component\Console\Input\InputOption;
-use Symfony\Component\Filesystem\Exception\IOException;
 
 /**
  * @internal
@@ -100,9 +99,6 @@ final class ConfigureCommand extends BaseCommand
             );
     }
 
-    /**
-     * @throws IOException
-     */
     protected function executeCommand(IO $io): bool
     {
         if (!$io->isInteractive()) {
--- a/src/Command/ConfigureCommand.php
+++ b/src/Command/ConfigureCommand.php
@@ -57,7 +57,6 @@ use const JSON_PRETTY_PRINT;
 use const JSON_UNESCAPED_SLASHES;
 use OutOfBoundsException;
 use RuntimeException;
-use function Safe\file_get_contents;
 use function Safe\glob;
 use function Safe\json_decode;
 use function Safe\json_encode;
@@ -67,6 +66,7 @@ use function str_starts_with;
 use Symfony\Component\Console\Helper\FormatterHelper;
 use Symfony\Component\Console\Helper\QuestionHelper;
 use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Filesystem\Exception\IOException;
 
 /**
  * @internal
@@ -100,6 +100,9 @@ final class ConfigureCommand extends BaseCommand
             );
     }
 
+    /**
+     * @throws IOException
+     */
     protected function executeCommand(IO $io): bool
     {
         if (!$io->isInteractive()) {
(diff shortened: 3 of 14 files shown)


theofidry:
@sanmai before I dive in a review, is it a worthwhile case?

`TestFrameworkFinder` exercises `ExecutableFinder` and the filesystem and environment variable manipulation there can't be mocked, so we still need an integration test that uses a real filesystem in the end.

sanmai:
@theofidry As far as I can tell the only remotely related thing changed here is `PhpUnitCustomExecutablePathProviderTest` - but we only untied it from the `FileSystem` used for validation, everything else left intact.

src/Command/ConfigureCommand.php
```diff
@@ -95,6 +100,9 @@ protected function configure(): void
             );
     }
 
+    /**
+     * @throws IOException
+     */
```
theofidry: the agents like to add this exception but it's a case we do not want as we do not react on `::executeCommand()` (like we do not react on the native `::execute()`). I did not look/find a neat way to disable Mago about this case though so meanwhile we need to stay vigilant during reviews

src/Command/ConfigureCommand.php
```diff
@@ -139,13 +147,10 @@ protected function executeCommand(IO $io): bool
             $this->abort();
         }
 
-        $container = $this->getApplication()->getContainer();
```
theofidry, line 143: Whilst harmless here (I think?), it's not equivalent.

I don't remember if it's something we already have (quick current lookup suggests not) but I mentioned it in #2572 so maybe it's a later plan – but a FS could have a read-only implementation.

In this case, this implementation would be registered at the container level, so the code here would bypass that configuration.

I don't think it's a problem right now at least, but to keep in mind with this sort of changes

src/Config/ValueProvider/PhpUnitCustomExecutablePathProvider.php
```diff
@@ -59,6 +59,7 @@ public function __construct(
         private TestFrameworkFinder $phpUnitExecutableFinder,
         private ConsoleHelper $consoleHelper,
         private QuestionHelper $questionHelper,
+        private FileSystem $fileSystem = new FileSystem(),
```
theofidry, line 62: For such cases where this is a service instantiated by the `Container`, it should be the injected shared filesystem instance; Otherwise we are introducing back the problem that we can configure the `Container` but there is multiple places that still need to be updated because they decided to instantiate their own services.

The statelessness of `Filesystem` does not guarantee that two implementations behave the same way (e.g. if there is a readonly one).

Making those optional values also makes it harder to guarantee that tests don't suddenly use an actual instance and end up having a side-effects whilst executing the tests.

tests/phpunit/Config/ValueProvider/PhpUnitCustomExecutablePathProviderTest.php
```diff
@@ -46,26 +47,29 @@
 use PHPUnit\Framework\MockObject\MockObject;
 use Symfony\Component\Console\Exception\RuntimeException as SymfonyRuntimeException;
 use Symfony\Component\Console\Input\StringInput;
-use Symfony\Component\Filesystem\Path;
 
 #[Group('integration')]
```
theofidry, line 51: is this still necessary now then? Likewise for the other affected tests

tests/phpunit/StaticAnalysis/Config/StaticAnalysisConfigLocatorTest.php
```diff
@@ -49,6 +50,27 @@ final class StaticAnalysisConfigLocatorTest extends TestCase
...
+            ]);
+        $fileSystemMock
+            ->expects($this->once())
+            ->method('realPath')
+            ->with('/custom-dir/phpstan.neon.dist')
+            ->willReturn('/resolved/phpstan.neon.dist');
+
+        $locator = new StaticAnalysisConfigLocator('/config-dir', $fileSystemMock);
+
+        $this->assertSame('/resolved/phpstan.neon.dist', $locator->locate('phpstan', '/custom-dir'));
```
theofidry: not critical but I personally prefer having the actual result assigned to a variable (e.g. `$actual`). It doesn't make much difference but when you need to dig in to step debug, this avoids a few annoying steps as you don't need to go through `::assertSame()`

theofidry (commented):
LGTM overall 👍

src/Config/ValueProvider/PhpUnitCustomExecutablePathProvider.php
sanmai: That is correct: if the `Filesystem` changes in the future, it could bite us. OTOH this increases the scope of this work hedging against the future that may never come, and if it comes, `git grep new\ FileSystem` will find all violations at once.

tests/phpunit/Config/ValueProvider/PhpUnitCustomExecutablePathProviderTest.php
sanmai: PHPat rejects all 7: `IoCodeDetectorVisitor` treats `new FileSystem()`  as I/O

src/Command/BaseCommand.php
```diff
@@ -72,13 +73,19 @@ protected function initialize(InputInterface $input, OutputInterface $output): v
         $this->io = new IO($input, $output);
     }
 
+    /**
+     * @throws Throwable
```
sanmai: If we tell Mago that the outermost function throws, it no longer requires inner functions to declare that they throw. So with this change we no longer going to need to care about `@throws Foo` popping elsewhere.

sanmai: Never mind, it does not.

mago.toml
```diff
@@ -63,6 +63,7 @@ unchecked-exceptions = [
     "PHPUnit\\Exception",
     "Psr\\Container\\ContainerExceptionInterface",
     "Safe\\Exceptions\\SafeExceptionInterface",
+    "Symfony\\Component\\Filesystem\\Exception\\ExceptionInterface",
```
sanmai, line 67: To silence that `unhandled-thrown-type` in bulk

tests/phpunit/StaticAnalysis/Config/StaticAnalysisConfigLocatorTest.php
sanmai: That's done

tests/phpunit/Config/ValueProvider/TestFrameworkConfigPathProviderTest.php
```diff
@@ -57,14 +58,18 @@ final class TestFrameworkConfigPathProviderTest extends BaseProviderTestCase
...
     private MockObject&ConsoleHelper $consoleMock;
 
+    private MockObject&FileSystem $fileSystemMock;
+
     protected function setUp(): void
     {
         $this->locatorMock = $this->createMock(TestFrameworkConfigLocatorInterface::class);
         $this->consoleMock = $this->createMock(ConsoleHelper::class);
+        $this->fileSystemMock = $this->createMock(FileSystem::class);
         $this->provider = new TestFrameworkConfigPathProvider(
```
theofidry, line 69:
```suggestion

        $this->provider = new TestFrameworkConfigPathProvider(
```

tests/phpunit/StaticAnalysis/Config/StaticAnalysisConfigLocatorTest.php
theofidry, line 81: This is more an FYI and is out of scope of this PR, but what I had in mind for example, to avoid having tests having potential side-effects, was here to inject a `ReadonlyFileSystem` which could either do nothing or bail out on write operations, allowing to guarantee that this test will not modify anything on the FS.

tests/phpunit/Config/ValueProvider/PhpUnitCustomExecutablePathProviderTest.php
theofidry: Which is legit I suppose, having it being able to trace if you override the parameter or something is probably asking too much from it.

src/Config/ValueProvider/PhpUnitCustomExecutablePathProvider.php
theofidry: is introducing this wiring such a big deal/noise though? – I did not look, but I thought it would barely affect some services in `Container` which is not much to look at and for tests you already explicitly inject it in almost all cases

theofidry (approved):

src/Command/ConfigureCommand.php
sanmai: That's a nice idea :100:

PR #3539, opened 2026-08-31
doc: Add an ADR for the decorators naming conventions

theofidry:
Add an ADR to establish the pattern followed when creating a decorator.

### DIFF
--- a/adr/0014-decorator-naming.md
+++ b/adr/0014-decorator-naming.md
@@ -1,0 +1,88 @@
+# Decorator naming conventions
+
+## Context
+
+A decorator implements the same contract as another object, delegates to that object and
+adds behaviour. Without a shared naming convention, decorator classes may use a generic
+`Decorator` suffix, while the decorated dependency may be called `$inner`, `$wrapped`,
+`$delegate` or merely repeat the contract name.
+
+Infection names symbols after their domain role and behaviour, not their technical kind or
+the pattern used to implement them. An interface does not receive an `Interface` suffix and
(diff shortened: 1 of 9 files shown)


adr/0014-decorator-naming.md
```diff
@@ -0,0 +1,88 @@
...
+when they are changed; adopting this convention does not by itself require unrelated
+renames.
+
+## Consequences
+
+Decorator names communicate both their contract and the behaviour they add. The decorated
+dependency is unambiguous among the class's collaborators, and code search for `decorated`
+can find decorator relationships.
+
+Some existing decorators do not follow this convention. Aligning them may cause internal
```
theofidry, line 64: actually changed them here, it's not many so easy enough to do, but I would do it in a separate PR.

adr/0014-decorator-naming.md
```diff
@@ -0,0 +1,88 @@
...
+
+Name a decorator `<Behaviour><Contract>`, where `<Behaviour>` describes what the decorator
+adds and `<Contract>` is the short name of the contract it implements. Do not add a
+`Decorator` suffix.
+
+For example, a source collector that memoizes its result is named
+`MemoizedSourceCollector`, not `SourceCollectorDecorator` or
+`MemoizedSourceCollectorDecorator`.
+
+Name the constructor-injected decorated object `$decorated<Role>`, where `<Role>` is the
```
theofidry, line 35: I picked `$decorated` but I'd be fine with `$inner` too, your pick.

PR #3533, opened 2026-08-30
cs: Enforce the semicolon position via a PHP-CS-Fixer rule

theofidry:
A PHP-CS-Fixer rule is now available for the semicolon placement, so we can enforce this to have one style.

The current PR proposes:

```php
$configuration
    ->setRules([...])     
    ->setFinder($finder);
```

But we can use `new_line_for_chained_calls` to have the form:

```php
$configuration
    ->setRules([...])     
    ->setFinder($finder)
;
```

I don't have a preference at this point, so up to you.

PR #3542, opened 2026-09-01
tests: Nail down mutations in IncrementInteger/DecrementInteger

sanmai:
## Description

To fix the broken build in https://github.com/infection/infection/pull/3533

### DIFF
--- a/tests/phpunit/Mutator/Number/IncrementIntegerTest.php
+++ b/tests/phpunit/Mutator/Number/IncrementIntegerTest.php
@@ -239,6 +239,19 @@ final class IncrementIntegerTest extends BaseMutatorTestCase
             ),
         ];
 
+        yield 'It does increment a minus one limit argument of a function other than preg_split' => [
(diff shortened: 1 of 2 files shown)


tests/phpunit/Mutator/Number/IncrementIntegerTest.php
```diff
@@ -239,6 +239,19 @@ public static function mutationsProvider(): iterable
...
 
+        yield 'It does increment a minus one limit argument of a function other than preg_split' => [
+            self::wrapCodeInMethod(
+                <<<'PHP'
+                    explode('//', 'string', -1);
+                    PHP,
+            ),
+            self::wrapCodeInMethod(
+                <<<'PHP'
+                    explode('//', 'string', -0);
```
sanmai, line 250: that's what it does apparently

theofidry (approved):
I'm confused by the description – I don't really get what it's fixing specifically, but if this covers branches that were not covered previously 👍

PR #3545, opened 2026-09-02
fix(phpunit): Remove the duration test orders when disabling the test run history

sebastianbergmann:
## Description

Since PHPUnit 13.3, ordering tests by duration while the recording of the test run history is disabled is a test runner warning, exactly like ordering by defects:

> Tests cannot be ordered by duration because recording of the test run history is disabled

The initial run configuration sets `recordTestRunHistory="false"` and `failOnWarning="true"`, so on a project whose `phpunit.xml` declares `executionOrder="duration-ascending"` (or `"depends,defects,duration-ascending"`) that warning aborts the run before a single test is executed. Infection then reports `Project tests must be in a passing state before running Infection.` even though the suite is perfectly healthy.

#3458 fixed this for `defects` by stripping that component from `executionOrder` when the history is disabled, but only for `defects`. PHPUnit emits the warning for two things: the defects-first flag and the duration orders (`duration-ascending`, `duration-descending`, and the deprecated `duration` alias, see `TextUI/Application.php` in PHPUnit 13.3). A configuration such as `executionOrder="depends,defects,duration-ascending"` therefore still fails: `defects` is removed, `duration-ascending` is copied verbatim, and PHPUnit refuses to run.

This PR makes the initial run strip every `executionOrder` component that requires the test run history, so the fix from #3458 holds regardless of which of the two orders a project uses. Dropping a component is safe because PHPUnit's config loader parses them independently: `defects` sets the defects-first flag, `depends`/`no-depends` set the dependency resolution, and the remaining values set the order itself. Whatever is left keeps its meaning, and without an order component PHPUnit uses its default order. The initial run does not benefit from any ordering anyway, since there is no history to order by.

## Changes

- Replaces the `defects`-only filter in `XmlConfigurationManipulator::deactivateResultCaching()` with a filter for all `executionOrder` components that require the test run history (`defects`, `duration`, `duration-ascending`, `duration-descending`), listed in a named constant with a link to the PHPUnit check that motivates it. The fallback to `default` when nothing remains is unchanged.
- Extends the `XmlConfigurationManipulatorTest` data provider with the duration cases (each duration value on its own, and `depends,defects,duration-ascending` which is the shape from the report) and renames the test to match.
- Updates the `InitialConfigBuilderTest` fixture to `depends,defects,duration-ascending` (renamed to `phpunit_with_order_requiring_test_run_history_set.xml`) so the builder-level test covers both orders in one go; the 13.2 case still asserts the order is left untouched there.
- Updates the `PHPUnit_13-3` end-to-end scenario to order by `depends,defects,duration-ascending` and its README accordingly. With the source change reverted, this scenario reproduces the reported failure; with it, it completes and matches `expected-output.txt`.

## Checklist

- [x] Tests added/updated
- [ ] Documentation updated (not needed, no user-facing behaviour change beyond the fix)
- [ ] CHANGELOG.md updated (no deprecations or breaking changes)
- [ ] Appropriate labels applied (`bug`)

## Reviewer Notes

- The list of orders lives in a constant rather than in the filter closure so that the PHPUnit reference sits next to the values, and so that a future PHPUnit change only needs a new entry.
- `duration` is included even though PHPUnit 13.3 deprecates it: the loader maps it to `duration-ascending` before the warning check, so it triggers the same warning.
- Verified against PHPUnit 13.3.2: `make cs`, `make phpstan`, `make mago`, `make rector-check` and the two affected unit test files are clean, and the updated e2e scenario passes. Also verified on the real project from the report (`executionOrder="depends,defects,duration-ascending"`, `failOnPhpunitWarning="true"`): the generated initial configuration now contains `executionOrder="depends"` and the initial test run passes.

## Related issues

Follow-up to https://github.com/infection/infection/pull/3458 and https://github.com/infection/infection/issues/3457

### DIFF
--- a/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
+++ b/src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
@@ -43,6 +43,7 @@ use function explode;
 use const FILTER_VALIDATE_URL;
 use function filter_var;
 use function implode;
+use function in_array;
 use Infection\TestFramework\PhpUnit\Config\Path\PathReplacer;
 use Infection\TestFramework\XML\SafeDOMXPath;
 use const LIBXML_ERR_ERROR;
@@ -61,6 +62,20 @@ use Webmozart\Assert\Assert;
  */
 final readonly class XmlConfigurationManipulator
 {
+    /**
+     * The `executionOrder` components PHPUnit 13.3+ reports a test runner warning for when the
+     * recording of the test run history is disabled https://github.com/sebastianbergmann/phpunit/blob/13.3.0/src/TextUI/Application.php.
+     * `duration` is the deprecated alias of `duration-ascending`.
+     *
+     * @var list<non-empty-string>
+     */
+    private const array EXECUTION_ORDERS_REQUIRING_TEST_RUN_HISTORY = [
+        'defects',
+        'duration',
+        'duration-ascending',
+        'duration-descending',
+    ];
+
     public function __construct(
         private PathReplacer $pathReplacer,
         private string $phpUnitConfigDir,
@@ -108,8 +123,8 @@ final readonly class XmlConfigurationManipulator
 
         $this->setAttributeValue($xPath, 'recordTestRunHistory', 'false');
 
-        // Ordering tests by defects requires the test run history which is deactivated right above. PHPUnit ignored that combination silently until 13.3 turned it into a test runner warning, which aborts the initial run because of failOnWarning https://github.com/sebastianbergmann/phpunit/blob/13.3.0/src/TextUI/Application.php
-        $this->removeDefectsFromExecutionOrder($xPath);
+        // Ordering tests by defects or by duration requires the test run history which is deactivated right above. PHPUnit ignored that combination silently until 13.3 turned it into a test runner warning, which aborts the initial run because of failOnWarning https://github.com/sebastianbergmann/phpunit/blob/13.3.0/src/TextUI/Application.php
+        $this->removeExecutionOrdersRequiringTestRunHistory($xPath);
     }
 
     public function handleResultCacheAndExecutionOrder(string $version, SafeDOMXPath $xPath, string $mutationHash, string $tmpDir): void
@@ -345,10 +360,12 @@ final readonly class XmlConfigurationManipulator
     }
 
     /**
-     * The `defects` component of `executionOrder` is orthogonal to the other ones, so dropping it
-     * leaves the meaning of the remaining components untouched.
+     * PHPUnit parses the components of `executionOrder` independently from each other: `defects` only
+     * sets the defects-first flag, `depends`/`no-depends` only set the dependency resolution and the
+     * remaining ones set the order itself. Dropping a component hence leaves the meaning of the other
+     * ones untouched; without an order component PHPUnit falls back to its default order.
      */
-    private function removeDefectsFromExecutionOrder(SafeDOMXPath $xPath): void
+    private function removeExecutionOrdersRequiringTestRunHistory(SafeDOMXPath $xPath): void
     {
         $node = $xPath->queryAttribute('/phpunit/@executionOrder');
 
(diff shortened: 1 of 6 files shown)


sanmai (approved):
Makes sense, thanks

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
```diff
@@ -345,10 +360,12 @@ private function buildSchemaPath(string $nodeValue): string
...
     /**
-     * The `defects` component of `executionOrder` is orthogonal to the other ones, so dropping it
-     * leaves the meaning of the remaining components untouched.
+     * PHPUnit parses the components of `executionOrder` independently from each other: `defects` only
+     * sets the defects-first flag, `depends`/`no-depends` only set the dependency resolution and the
+     * remaining ones set the order itself. Dropping a component hence leaves the meaning of the other
+     * ones untouched; without an order component PHPUnit falls back to its default order.
      */
-    private function removeDefectsFromExecutionOrder(SafeDOMXPath $xPath): void
+    private function removeExecutionOrdersRequiringTestRunHistory(SafeDOMXPath $xPath): void
```
theofidry, line 373: I quite like this new method name.

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
```diff
@@ -61,6 +62,20 @@
  */
 final readonly class XmlConfigurationManipulator
 {
+    /**
+     * The `executionOrder` components PHPUnit 13.3+ reports a test runner warning for when the
```
theofidry, line 66: I think this comment should rather be in `::removeExecutionOrdersRequiringTestRunHistory()`. The current comment of that method explains _why it's a safe change_, but does not explain _why we are doing this_, which this comment does.

I would also phrase it in a more generic way, i.e. not PHPUnit 13.3+ specific – the problem is older after all no?

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
```diff
@@ -61,6 +62,20 @@
...
+     * recording of the test run history is disabled https://github.com/sebastianbergmann/phpunit/blob/13.3.0/src/TextUI/Application.php.
+     * `duration` is the deprecated alias of `duration-ascending`.
+     *
+     * @var list<non-empty-string>
+     */
+    private const array EXECUTION_ORDERS_REQUIRING_TEST_RUN_HISTORY = [
+        'defects',
+        'duration',
```
theofidry, line 74: So if I understand correctly we now additionally strips `duration`, `duration-ascending` and `duration-descending`.

Just to make sure, the description and comment claims this is fine in PHPUnit 13.3+, but is this the case for older versions of PHPUnit? Technically Infection can still be run with PHPUnit 9.

theofidry (approved):
Left a few nitpicks/questions, LGTM otherwise, thanks for the PR!

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
sebastianbergmann: Good point, moved it into the docblock of `removeExecutionOrdersRequiringTestRunHistory()`, which now explains both why we strip these components and why doing so is safe.

I kept the reference to PHPUnit 13.3 on purpose though: the requirement on the test run history is indeed older, but before 13.3 PHPUnit silently ignored the combination, so nothing broke.

PHPUnit 13.3 turned it into a test runner warning, which `failOnWarning` promotes to an aborted initial run. That is also why the stripping is gated to PHPUnit >= 13.3, so the version is worth mentioning next to it.

src/TestFramework/PhpUnit/Config/XmlConfigurationManipulator.php
sebastianbergmann: Older versions are not affected. `deactivateResultCaching()` returns early for anything below 13.3 after setting `cacheResult="false"`, so on PHPUnit 9 through 13.2 `executionOrder` is copied verbatim into the initial run configuration.

`InitialConfigBuilderTest::test_it_keeps_the_orders_requiring_the_test_run_history_already_set_for_phpunit_13_2()` asserts exactly that with `depends,defects,duration-ascending`. And even if it were applied there, those versions silently ignore the defects and duration orders when the result cache is disabled, so dropping the components would not change the order of the initial run.

maks-rafalko (approved):
👍

PR #3549, opened 2026-09-02
feat(source): Add source symbol selection

### LINKED ISSUE #2237: Reduce mutation scope to class method or line

theofidry:
## Checklist

- [ ] Documentation updated (provide link to PR: https://github.com/infection/infection.github.io/pull/XXX)

## Description

Allow `infection run` to target source symbols in addition to source and test paths. This provides a precise way to run Infection against the code currently being worked on without changing `infection.json5`.

Selectors support short or fully qualified class names, with an optional method and absolute source line:

| Selector | Mutation scope |
| --- | --- |
| `Differ` | Every eligible node in the class |
| `Differ::diff` | A method |
| `Differ::32` | Eligible nodes covering line 32 in the class |
| `Differ::diff::32` | Eligible nodes covering line 32 in the method |
| `Infection\Differ\Differ::diff` | A fully qualified class and method |

Multiple source selectors are additive. Source paths, source selectors, and test paths can also be combined in one command:

```shell
infection run \
    src/Differ/Differ.php \
    Differ::diff \
    tests/phpunit/Differ/DifferTest.php
```

Here the source path limits which files are considered for mutation, the selector narrows the eligible code within those files, and the test path independently limits the initial test suite. Existing positional source and test paths continue to work as before.

The same selection can be previewed on an enriched AST, including node eligibility and line numbers:

```shell
infection debug:dump-ast \
    src/Differ/Differ.php \
    --source-selector='Differ::diff::32'
```

## Changes

- Generalises positional path classification into positional argument classification. Readable files and directories retain precedence, while other values are parsed as source selectors using the `Class[::method|line][::line]` grammar without consulting the autoloader.
- Introduces a source filter that keeps file and symbol selection separate. File filters reduce the collected source files; a matcher then compares enriched AST nodes against the selected class, method, and line ranges.
- Applies symbol selection in the existing AST enrichment stack by marking non-selected nodes as ineligible. The complete AST is preserved, and filtering occurs before covering tests are attached and untested nodes are excluded.
- Tracks matches across all processed source files and reports all selectors that matched no source symbol together. This avoids silently running with a misspelt or out-of-scope selector.
- Adds `debug:dump-ast --source-selector` so the selected eligibility can be inspected without starting a mutation run.

## Related issues

Closes #2237.


theofidry:
@coderabbitai full review

src/Command/Debug/DumpAstCommand.php
```diff
@@ -111,6 +115,12 @@ protected function configure(): void
             InputOption::VALUE_OPTIONAL,
             'List of changed line ranges. E.g. "10:12,24:30" will indicate that the lines 10, 11, 12, 24, 25, ..., 30 changed.',
         );
+        $this->addOption(
```
theofidry, line 118: to review if this option is really necessary, I would otherwise think we can do like we do in the `RunCommand`

src/Command/Option/PathsArgument.php
```diff
@@ -55,13 +55,20 @@ final class PathsArgument
 
     public static function addArgument(Command $command): Command
     {
-        return $command->addArgument(
-            self::NAME,
-            InputArgument::IS_ARRAY | InputArgument::OPTIONAL,
+        return self::addArgumentWithDescription(
```
theofidry, line 58: may need to revert that, depends on the DumpAstCommand

src/Command/RunCommand.php
```diff
@@ -372,6 +373,9 @@ protected function configure(): void
             );
     }
 
+    /**
+     * @throws SourceSymbolNotFound
```
theofidry, line 385: to remove

src/Configuration/SourceFilter/PositionalPathsFilter.php
theofidry, line 39: May need a refresh

src/Configuration/ConfigurationFactory.php
```diff
@@ -479,7 +482,7 @@ private function refineFilterIfNecessary(
             );
         }
 
-        return [$sourceFilter, $testFrameworkExtraArgs];
+        return [new SourceFilter($sourceFilter, $sourceSymbolSelectors), $testFrameworkExtraArgs];
```
theofidry, line 486: line return

src/Configuration/PositionalArgumentsClassifier.php
```diff
@@ -69,18 +69,33 @@ public function __construct(
...
 
         $configDir = dirname($schema->pathname);
         $absoluteSourceDirs = self::resolveAbsoluteSourceDirectories($schema, $configDir);
 
         $sourcePaths = [];
         $testPaths = [];
+        $sourceSelectors = [];
 
         foreach ($paths as $path) {
+            $absolutePath = Path::isAbsolute($path)
```
theofidry, line 85: TODO: to extract this – maybe another PR before it

src/Configuration/PositionalArgumentsClassifier.php
```diff
@@ -90,10 +105,18 @@ public function classify(
...
-        return new ClassifiedPaths($sourcePaths, $testPaths);
+        return new ClassifiedPositionalArguments(
+            $sourcePaths,
+            $testPaths,
+            $sourceSelectors,
+        );
     }
 
     /**
+     * TODO: Consolidate this with BasicSourceCollector::makePathsAbsolute()
```
theofidry, line 116: Won't address this TODO in this story I think

src/Container/Container.php
```diff
@@ -582,7 +582,7 @@ public static function create(): self
                     $container->getSourceCollector(),
                     $config->mutators,
                     $container->getEventDispatcher(),
-                    $container->getFileMutationGenerator(),
+                    $container->getFileMutationGenerator(), $container->getNodeTraverserFactory()->getSourceSymbolMatcher(),
```
theofidry, line 589: line return

src/Mutation/MutationGenerator.php
```diff
@@ -117,6 +122,12 @@ public function generate(bool $onlyCovered): iterable
             );
         }
 
+        $unmatchedSelectors = $this->sourceSymbolMatcher->getUnmatchedSelectors();
+
+        if ($unmatchedSelectors !== []) {
+            throw SourceSymbolNotFound::forSelectors($unmatchedSelectors);
```
theofidry, line 128: My original thought was, if the user uses a symbol which is invalid (does not select anything), then we want to fail with an exception `SourceSymbolNotFound` here.

However, the timing is a bit awkward. Indeed, we can't do this earlier, because we did not necessarily parse all files before this point, but at the same time it also means we started the mutation evaluation process, so you get some of the output already:

```
$ bin/infection Differ::diff tests/phpunit/Git/CommandLineGitTest.php

#StandWithUkraine

Infection - PHP Mutation Testing Framework version 0.35.x-dev

[notice] Running Infection with 7 threads.

Running initial tests with PHPUnit version 12.5.34

   40 [============================]    9 s

Generate mutants...

Processing source code files: 626/626
In SourceSymbolNotFound.php line 54:

  The following source selectors did not match any source symbol: "Differ::diff".

$ bin/infection Differ::diff tests/phpunit/Git/CommandLineGitTest.php --no-progress

Infection - PHP Mutation Testing Framework version 0.35.x-dev

[notice] Running Infection with 7 threads.

Running initial tests with PHPUnit version 12.5.34

.: killed by tests, A: killed by SA, M: escaped, U: uncovered
E: fatal error, X: syntax error, T: timed out, S: skipped, I: ignored


Generate mutants...

Processing source code files...

In SourceSymbolNotFound.php line 54:

  The following source selectors did not match any source symbol: "Differ::diff".


```

I would like to also point out that currently if you have something like:

```shell
bin/infection src/Differ tests/phpunit/Git/CommandLineGitTest.php
```

You will not get any failure, despite having no mutation. This may be confusing at first glance, but those are two different scenarios: the user asked to mutate some specific files and restricted the tests. That that file has no coverage (in this case because the tests cover something else), is a legitimate outcome.

On the other hand with the symbol selection, we are validating that it _selected something_, regardless of coverage. Now part of the issue is... this is not reliable. Indeed, the selection could point to a legitimate symbol in the codebase, but it may either no be found in the current selected file subset, or, it is marked as ineligible for other reasons (e.g. not part of the git diff).

So I am really questioning if we should validate this... Maybe a better idea would be to have at debug level of verbosity a log message logging the matched symbols?

src/Source/Exception/SourceSymbolNotFound.php
```diff
@@ -0,0 +1,79 @@
...
+     * @param non-empty-list<SourceSymbolSelector> $selectors
+     */
+    public static function forSelectors(array $selectors): self
+    {
+        return new self(
+            sprintf(
+                'The following source selectors did not match any source symbol: %s.',
+                implode(
+                    ', ',
+                    array_map(self::formatSelector(...), $selectors),
```
theofidry, line 59: line return

src/PhpParser/Visitor/ExcludeNonSelectedSourceNodesVisitor.php
```diff
@@ -0,0 +1,130 @@
...
+            $this->classStack[] = $node;
+        }
+
+        if ($node instanceof Node\Stmt\ClassMethod) {
+            $this->methodStack[] = $node;
+        }
+
+        if (
+            LabelNodesAsEligibleVisitor::isEligible($node)
+            && !$this->matches($node)
```
theofidry, line 84: subtle but important part: we only match against eligible nodes. This means the information we get (checked symbols) does not match the complete list of symbols present in those files (see my previous comment as to what impact it has)

tests/e2e/Source_Symbol_Selection/run_tests.bash
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+readonly INFECTION="../../../${1:-bin/infection}"
```
theofidry, line 5: I'm not happy with this test. I would rather have a `Source_Selection` end-to-end test where we consolidate various variants but maybe I'm too demanding and it won't be pratical

tests/phpunit/Mutation/MutationGeneratorTest.php
```diff
@@ -150,10 +155,82 @@ public function test_it_dispatches_events(): void
...
             $fileMutationGeneratorMock,
+            new SourceSymbolMatcher([]),
         );
 
         foreach ($mutationGenerator->generate(false) as $_) {
             // We just want to iterate here to trigger the generator
         }
     }
+
+    public function test_it_rejects_unmatched_source_selectors_after_processing_every_source(): void
```
theofidry, line 167: I didn't look at the tests of this file yet

theofidry:
I think this is good enough for a first round of a review despite having a few points to polish still.

src/Command/Debug/DumpAstCommand.php
maks-rafalko: IIRC, we deprecated `--filter` in favor of positional paths. So with this logic we can do the same with this option (=it's not needed).

Or what is the use case where it would be needed?

src/Configuration/SourceFilter/PositionalPathsFilter.php
```diff
@@ -42,7 +42,7 @@
  *
  * @internal
  */
-final readonly class PositionalPathsFilter implements SourceFileFilter
```
maks-rafalko, line 45: not sure I got why it's changed this way

src/Mutation/MutationGenerator.php
```diff
@@ -89,6 +93,7 @@ public function __construct(
      */
     public function generate(bool $onlyCovered): iterable
     {
+        $this->sourceSymbolMatcher->reset();
```
maks-rafalko, line 96: why is it needed here, in the beginning of `::generate()`?

don't we call it once?

src/PhpParser/NodeTraverserFactory.php
```diff
@@ -62,11 +65,18 @@
...
+    /**
+     * @param list<SourceSymbolSelector> $sourceSymbolSelectors
+     */
     public function __construct(
         private SourceLineMatcher $sourceLineMatcher,
         private LineRangeCalculator $lineRangeCalculator,
         private bool $onlyCovered,
+        array $sourceSymbolSelectors,
     ) {
+        $this->sourceSymbolMatcher = new SourceSymbolMatcher($sourceSymbolSelectors);
```
maks-rafalko: why not via container?

src/PhpParser/NodeTraverserFactory.php
```diff
@@ -94,12 +105,23 @@ public function createEnrichmentTraverser(
...
             ),
-            new AddTestsVisitor(
-                $trace,
-                $this->lineRangeCalculator,
-            ),
         ];
 
+        $sourceSymbolMatcher = $sourceSymbolSelector === null
+            ? $this->sourceSymbolMatcher
+            : new SourceSymbolMatcher([$sourceSymbolSelector]);
```
maks-rafalko: to me: not sure I understand this code

tests/e2e/Source_Symbol_Selection/infection.json5
```diff
@@ -0,0 +1,22 @@
+{
+    "$schema": "../../../resources/schema.json",
+    "timeout": 25,
+    "threads": 1,
+    "source": {
+        "directories": [
+            "../../../src"
```
maks-rafalko, line 7: why do we use Infection's `src` in this e2e tests when all the other e2e tests are self-contained? Can't we just create a couple of files here, in `./src`?

src/PhpParser/NodeTraverserFactory.php
maks-rafalko, line 112: can't DumpAstCommand use constructor's `array $sourceSymbolSelectors,` to avoid this code and nullable parameter `?SourceSymbolSelector $sourceSymbolSelector = null,`?

src/Command/RunCommand.php
```diff
@@ -380,6 +381,9 @@ protected function configure(): void
         ;
     }
 
+    /**
+     * @throws SourceSymbolNotFound
```
maks-rafalko, line 385: The `@throws` is accurate, and that is the problem: `SourceSymbolNotFound` is the only new user-facing failure this feature introduces, and it is the one exception here that escapes `executeCommand()` uncaught.

`NoSourceFound`, `InitialTestsFailed`, `MinMsiCheckFailed` and `MaxTimeoutCountReached` all get a clean `$io->error()`. A mistyped selector — by far the most likely user error for this feature — gets a raw Symfony exception dump with a stack-trace hint instead.

Adding it to the existing arm at line 416 would do it:

```php
} catch (InitialTestsFailed|MinMsiCheckFailed|MaxTimeoutCountReached|SourceSymbolNotFound $exception) {
```

_Generated as part of a joint review by Maks and claude._

src/Configuration/PositionalArgumentsClassifier.php
```diff
@@ -69,18 +69,33 @@ public function __construct(
...
         $testPaths = [];
+        $sourceSelectors = [];
 
         foreach ($paths as $path) {
+            $absolutePath = Path::isAbsolute($path)
+                ? $path
+                : Path::join($configDir, $path);
+
+            if (!$this->isValidPath($absolutePath)) {
+                $selector = $this->sourceSymbolSelectorParser->parse($path);
```
maks-rafalko, line 90: This moves the selector parse *ahead* of `looksLikeClassOrFileName()`, which silently changes what a bare PascalCase argument means.

On `master`, `infection run SomeClass` hit the `// like SomeFile or SomeFile.php - bare values behave as --filter values` branch and became a `PlainFilter` — a file filter. After this PR it parses as a source symbol selector instead. The new tests encode it: `'short class selector' => ['Plus_']` yields `([], [], [selector])`, while `'symbolic php source token' => ['Plus_.php']` still routes to `sourcePaths`.

Two consequences for anyone relying on the bare form today:

- No file filter is produced, so `SourceFilter::filtersFiles()` returns `false`, `SourceCollectorFactory::create()` gets `null`, and the whole source tree is collected and enrichment-traversed.
- If the file's class name does not match the argument, or the file declares no class at all, the run now hard-fails with `SourceSymbolNotFound` where it previously mutated the file.

Per `AGENTS.md` ("Any BC risk upgrades the prefix to `feat!:`/`fix!:` and gets a `BC break` label plus an enumerated Breaking Changes section"), this needs the title bumped to `feat!(source):` with that section — or the bare-name case kept on the old file-filter path, reserving selectors for arguments containing `::` or a namespace separator.

_Generated as part of a joint review by Maks and claude._

src/Container/Container.php
```diff
@@ -586,7 +586,7 @@ public static function create(): self
...
```
maks-rafalko, line 589: `$container->getNodeTraverserFactory()->getSourceSymbolMatcher()` reaches through one service to fetch another, and `getSourceSymbolMatcher()` exists solely to hand out a mutable object owned by a `readonly` factory.

Per invariant 6 ("Container - one registry, typed getters"), registering `SourceSymbolMatcher::class` and injecting it into both `NodeTraverserFactory` and `MutationGenerator` would be flatter, and would drop both the accessor and the private-field construction in the factory.

Also, minor: two constructor arguments ended up on one line here, and again at line 471 where an existing line break was removed. The rest of this file is one argument per line.

_Generated as part of a joint review by Maks and claude._

src/PhpParser/Visitor/ExcludeNonSelectedSourceNodesVisitor.php
```diff
@@ -0,0 +1,130 @@
...
+        $this->classStack = [];
+        $this->methodStack = [];
+
+        return null;
+    }
+
+    public function enterNode(Node $node): null
+    {
+        if ($node instanceof Node\Stmt\ClassLike) {
+            $this->classStack[] = $node;
```
maks-rafalko, line 75: Pushing every `ClassLike` — anonymous classes included — means `Differ::diff` skips everything inside a `new class {...}` declared in `diff()`, and `Differ::32` can never target a line inside one.

The test `'an anonymous class does not inherit the enclosing symbol context'` shows this is deliberate, but nothing in the source says why, and a user selecting a method would reasonably expect code lexically inside it to be mutated.

Per the "explain WHY" rule, a short rationale comment here — why the enclosing context is deliberately dropped rather than inherited — would save the next reader the trip to the test.

_Generated as part of a joint review by Maks and claude._

maks-rafalko (commented):
thanks for working on this one, great feature.

Did the first round of testing and review

src/Command/Debug/DumpAstCommand.php
theofidry: I'm did not verify it yet, but if we can remove it, it would be better

src/Mutation/MutationGenerator.php
theofidry: @maks-rafalko open question

src/Configuration/SourceFilter/PositionalPathsFilter.php
theofidry: can it be it's from an obsolete diff? I can't make sense of this one

src/Mutation/MutationGenerator.php
theofidry: In practice not, but the service is stateful atm so if feels better to reset it. I could also reset it at the end if that helps.

But the whole thing is tight to that discussion which I would aim to resolve before deciding on this reset: https://github.com/infection/infection/pull/3549#discussion_r3916947034

tests/e2e/Source_Symbol_Selection/infection.json5
theofidry: Some slops which don't make sense, we should do like the others indeed

src/PhpParser/NodeTraverserFactory.php
theofidry: It should, we create a container with the input so there is no reasons we can't do it in the dump:ast and can in the run command.

src/Command/RunCommand.php
theofidry: it should indeed be addressed, if we throw such an exception (which is a pending point, see https://github.com/infection/infection/pull/3549#discussion_r3916947034)

src/Configuration/PositionalArgumentsClassifier.php
theofidry: It is without a doubt a BC and should be properly documented if it actually lands.

I'm not quite sure about the desired behaviour in this scenario.

src/PhpParser/Visitor/ExcludeNonSelectedSourceNodesVisitor.php
theofidry: it is a limitation of the current implementation, but to verify if it is one to stay or can be fixed

PR #3555, opened 2026-09-03
test(phpunit): Refactor PHPUnit configuration builder tests

theofidry:
## Description

Refactors the PHPUnit configuration builder tests so that each scenario displays the complete generated configuration beside its inputs. This makes behavioural differences between supported PHPUnit versions easier to review and reduces the boilerplate needed to add scenarios.

This address a long-standing issue that I have been having with the PHPUnit Adapter tests which is hard to tell what the actual output of the created config looks like for a given input. For example:

```php
public function test_it_disables_caching(): void
{
    $xml = $this->filesystem->readFile($this->builder->build('6.5'));

    $cacheResult = $this->queryXpath($xml, '/phpunit/@cacheResult')[0]->nodeValue;

    $this->assertSame('false', $cacheResult);
}
```

It is simple and does show more explicitly what we want the result to be, but not what it actually looks like. Now this would look like this:

```php
yield 'result caching is disabled' => self::createScenario(
    version: '6.5',
    xml: $loadFixture('phpunit.xml'),
    filteredSourceFilesToMutate: [],
    <<<XML
        <phpunit backupGlobals="false" backupStaticAttributes="false" bootstrap="$projectPath/app/autoload2.php" colors="false" convertErrorsToExceptions="true" convertNoticesToExceptions="true" convertWarningsToExceptions="true" processIsolation="false" syntaxCheck="false" defaultTestSuite="unit" failOnRisky="true" failOnWarning="true" stopOnFailure="true" cacheResult="false" stderr="false">
        ...
        XML,
);
```

Additionally, it is now possible to directly pass a string without creating a file:

```php
yield 'result caching is disabled' => self::createScenario(
    version: '6.5',
    fixture: <<<XML
        ...
        XML,
    filteredSourceFilesToMutate: [],
    <<<XML
        ...
        XML,
);
```

I did not remove the existing ones as there is no value in doing that though.

Last but not least, we now have a `FileSystem` instance that does not write to the filesystem. So no more risk of accidental writes.

## Changes

- Consolidates the `InitialConfigBuilder` scenarios into a data provider with complete expected XML documents.
- Reworks the `MutationConfigBuilder` tests around complete configuration and autoload outputs while retaining coverage for repeated builds, test ordering, PHPUnit version boundaries, and existing attribute values.
- Allows the in-memory filesystem to resolve absolute paths through Symfony's filesystem implementation, matching the operation required by the builders.


src/FileSystem/InMemoryFileSystem.php
```diff
@@ -69,6 +69,15 @@ final class InMemoryFileSystem extends FileSystem
      */
     private array $directories = [];
 
+    // An original instance we keep around for operations that are safe to delegate to keep the
+    // behaviour as close as the original.
```
sanmai: Look like something Claude would leave 🤔

src/FileSystem/InMemoryFileSystem.php
```diff
@@ -69,6 +69,15 @@ final class InMemoryFileSystem extends FileSystem
...
+    private FileSystem $fileSystem;
+
+    public function __construct()
```
sanmai:
```suggestion
    public function __construct(private FileSystem $fileSystem = new FileSystem())
```

sanmai (approved):
The approach makes sense. Some questions.

I wish we extended the old test so it would be easier to review 🤔

tests/phpunit/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilderTest.php
```diff
@@ -151,558 +131,735 @@ public function test_the_original_xml_config_must_be_a_valid_phpunit_config_file
...
+        $path = $builder->build($version);
+        $actual = $this->filesystem->readFile($path);
+
+        $this->assertSame($expected, $actual);
+    }
+
+    public static function configurationProvider(): iterable
+    {
+        $projectPath = Path::canonicalize(self::FIXTURES . '/project-path');
+        $loadFixture = static fn (string $fixture): string => file_get_contents(self::FIXTURES . '/' . $fixture);
```
sanmai, line 158: what is the reason we can't let the test itself handle this?

phpunit loads all data providers from the start, even before running any tests, so all file content will be in memory until at least the start of this test

src/FileSystem/InMemoryFileSystem.php
theofidry: I was actually me because I had to pause for a second whether having a `FileSystem` within another instance of `FileSystem` was really the way (as opposed in this case to call `Path::isAbsolute()` directly in the implementation)

sanmai: Ah, it was obvious that we use it like this looking at `isAbsolutePath`

tests/phpunit/TestFramework/PhpUnit/Config/Builder/InitialConfigBuilderTest.php
theofidry: I guess it's possible too.

My thinking is that I would like to be able to pass the XML string rather than having to add a file and then pass that file path, but then you cannot distinguish which one is which from the type, you need to do a path existence check. Doing it in the provider avoids that, at that cost. That said I couldn't observe a memory usage increase.

theofidry:
>I wish we extended the old test so it would be easier to review 🤔

How do you mean?

sanmai:
I meant adding these cases on top of existing cases. So for a time we will have both sets. And after a review you could strip the old cases. In any case, I think I get it so we shouldn't do that just here.

PR #3562, opened 2026-09-09
feat(test-framework): Make test framework factory dependencies explicit

theofidry:
## Description

Extend the internal `TestFrameworkFactory::create()` contract with the configuration and services required by the PHPUnit adapter. This lets factories using the new contract receive those dependencies through the common factory path and replaces optional parameters checked at runtime with required, typed arguments.

## Changes

- Add the parameters to the factory contract and forward them from `Factory`.
- Align `PhpUnitAdapterFactory` with the contract, removing defaults and redundant null assertions.
- Temporarily exempt `TestFrameworkFactory` from the future package dependency boundary while it constructs the legacy PHPUnit bridge.

## Related issues

- #2863

### DIFF
--- a/src/TestFramework/Contracts/TestFrameworkFactory.php
+++ b/src/TestFramework/Contracts/TestFrameworkFactory.php
@@ -35,6 +35,15 @@ declare(strict_types=1);
 
 namespace Infection\TestFramework\Contracts;
 
+use Infection\Configuration\Configuration;
+use Infection\Console\ConsoleOutput;
+use Infection\Process\Factory\MutantProcessContainerFactory;
+use Infection\Process\Runner\InitialTestsRunner;
+use Infection\TestFramework\Coverage\CoverageCheckerFactory;
+use Infection\TestFramework\TestFrameworkExtraOptionsFilter;
+use SplFileInfo;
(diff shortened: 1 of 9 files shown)


src/TestFramework/Contracts/TestFrameworkFactory.php
```diff
@@ -44,6 +53,7 @@ interface TestFrameworkFactory
 {
     /**
      * @param string[] $sourceDirectories
+     * @param SplFileInfo[] $filteredSourceFilesToMutate
      */
     public static function create(
```
sanmai, line 58: I wish this wasn't a static method but instead a [builder type](https://github.com/sanmai/di-container#builder-objects) that pulls the typed dependencies (Filesystem, ConsoleOutput, ...) through the constructor injection.

sanmai (approved):
Suggestion is non-blocking.

src/TestFramework/Contracts/TestFrameworkFactory.php
theofidry: I may be possible if we make the DI a requirement, but I would suggest we hold off onto that idea until the API is more refined. There's going to be a lot of changes to that factory method still

PR #3577, opened 2026-09-19
Instruct AI to use docker

maks-rafalko:
Almost every claude session fails running tests on my local setup because I have a strict memory limit, not saying that other settings can be just wrong for Infection.

Instruct AI to use our prepared docker contains where all the checks are green, so that AI doesn't spend tokens running the same checks locally and failing.


AGENTS.md
```diff
@@ -164,6 +164,51 @@ vendor/phpunit/phpunit/phpunit --filter test_method_name           # one method
...
+make test-unit-docker       # = make test-unit
+make test-e2e-docker        # e2e with a real coverage driver
+make test-infection-docker  # dogfood with a real coverage driver
+make test-docker            # all of the above
+```
+
+The image pins the settings the suite needs, so the `-docker` twin runs a known-good
+configuration regardless of the host. Reach for the `-docker` twin when:
+
+- **the host `php.ini` drifts from `devTools/`.** The image sets `memory_limit = 512M`
```
sanmai:
```suggestion
- **the host `php.ini` may differ from `devTools/`.** The image sets `memory_limit = 512M`
```

"drifts" is confusing; host's php.ini could have memory_limit=128M by default

sanmai (approved):

PR #3579, opened 2026-09-19
Add PHP 8.6 docker service for compose

maks-rafalko:
Pins the PHP RC base image and Xdebug `3.6.0alpha1`, as 8.6 is not released and no stable Xdebug supports it. 

The 8.6 base image no longer bundles `mbstring`, which PHPUnit requires, so it is compiled when missing - conditionally, leaving the 8.3-8.5 images unchanged.


devTools/Dockerfile
```diff
@@ -8,20 +8,24 @@ RUN apk add --update --no-cache \
 		linux-headers \
 	;
 
+# mbstring is compiled below only when the base image lacks it: PHP >= 8.6 no
+# longer bundles it, and PHPUnit requires it.
```
sanmai:
```suggestion
# PHPUnit requires mbstring, so we have to install it if missing
```

docker-compose.yml
```diff
@@ -36,3 +36,13 @@ services:
     volumes:
       - .:/opt/infection
 
+  # PHP 8.6 is not released yet: the base image is still a release candidate and
+  # no stable Xdebug supports it. Switch to '8.6'/a stable Xdebug once 8.6 is out.
```
sanmai:
```suggestion
```
I think we can drop this comment entirely - it appears to me rather obvious we use these specific versions due to 8.6 being RC.

sanmai (approved):

docker-compose.yml
maks-rafalko: hm.. agree, it wasn't here after the first run of claude, then I missed it. Thanks

